Issue #561, opened 2013-12-28
Attribute to always inline a function


klickverbot:
It should map to the `alwaysinline` IR attribute, and also cause such functions from imported modules to be emitted as `available_externally` to support cross-module inlining (similar to the extra semantic3s we run with inlining enabled).

We might want to implement this as a "blessed" UDA in a runtime module, e.g. `ldc.attributes`.

JohanEngelen:
A relevant PR: #1172.

klickverbot:
#1577 has been merged.

Issue #547, opened 2013-11-24
Add __attribute__ ((section(".interrupt_vector"))) -like syntax


JinShil:
The ```__attribute__ ((section("name")))``` syntax is often used for bare-metal firmware to place symbols at a specific memory location at link time. 

It is specifically useful to place the interrupt vector table and initial stack pointer at the appropriate place when programming for the ARM Cortex-M MCUs.  Here's an example using C and a GNU LD linker script.
```C
/***************************
* start.c
****************************/
// defined in linker script
extern unsigned long _stack_end;

void handler_reset(void)
{
  //Print hello world using SWI
}

__attribute__ ((section(".interrupt_vector")))
void (* const table_interrupt_vector[])(void) =
{
  (void *) &_stack_end,
  handler_reset
};
```

```
/***************************
* linkerscript.ld
****************************/
MEMORY
{
  CCRAM    (rxw) : ORIGIN = 0x10000000, LENGTH =   64k
  SRAM     (rxw) : ORIGIN = 0x20000000, LENGTH =  128k
  FLASH    (rx)  : ORIGIN = 0x08000000, LENGTH = 1024k
}

_stack_end = ORIGIN(CCRAM) + LENGTH(CCRAM);

SECTIONS
{
  .interrupt_vector :  
  {
    . = ALIGN(4);
    KEEP(*.interrupt_vector))       
    . = ALIGN(4);
   } >FLASH	
}
```

klickverbot:
Done in master, as `@ldc.attributes.section("…")`.

JohanEngelen:
@klickverbot I have trouble finding how to do `@ldc.attributes.section` with master. Can you help out?

klickverbot:
@JohanEngelen: Exactly what is the issue? Unless we have a regression (IR-level tests could find it easily now, of course), it should just work as expected. Import the module, apply the UDA and the variable should be emitted to the given section.

JohanEngelen:
@klickverbot 
Before, I was trying @ldc.attribute("section",...). Now I've found that `@section("...")` does work, but:
```D
import ldc.attributes;
@ldc.attributes.section(".mySection") int myGlobal;
```
gives the compile error: 
```
/Users/johan/ldc/johan/tests/ir/attributes.d(14): Error: unexpected ( in declarator
/Users/johan/ldc/johan/tests/ir/attributes.d(14): Error: basic type expected, not ".mySection"
[...]

PR #563, opened 2013-12-30
Add UDAs for LDC attributes.

redstar:
Introduces an attribute @ldc.attribute which can be used to annotate functions.
Example:

    @ldc.attribute("alwaysinline") void foo() {}

Valid attribute names can be found in the LLVM manual at http://www.llvm.org/docs/LangRef.html#function-attributes.

This fixes issue #561 and creates a base for issue #547.

### DIFF
--- a/gen/attribute.cpp
+++ b/gen/attribute.cpp
@@ -1,0 +1,213 @@
+//===-- gen/attribute.h - Handling of @ldc.attribute ------------*- C++ -*-===//
+//
+//                         LDC – the LLVM D compiler
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+//
+// Contains helpers for handling of @ldc.attribute.
+//
+//===----------------------------------------------------------------------===//
+
+#include "gen/attribute.h"
+#include "llvm/ADT/StringSwitch.h"
+#if LDC_LLVM_VER >= 303
+#include "llvm/IR/Function.h"
+#else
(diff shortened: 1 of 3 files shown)


JohanEngelen:
I need this! :)

JohanEngelen:
in the near future, I'd like to try and add function multiversioning in LDC, but the first step will be to allow people to specify e.g. "+sse" on one function (without enabling SSE on everything else through the cmdline). Something like this: https://gcc.gnu.org/wiki/FunctionMultiVersioning

JohanEngelen:
@klickverbot I am not sure if it is a good idea to add a new struct name for each LDC attribute we introduce.

Note that with the current solution, simple code like this is broken (silently) due to a naming conflict that I think can easily happen in practice:
```D
import ldc.attributes;

struct section {
  string name;
}

@section(".mySection") int myGlobal; // will *not* be in .mySection
```
And this does not compile:
```D
@ldc.attributes.section(".mySection") int myGlobal;
```

JohanEngelen:
btw, the IR-testsuite is perfect for this stuff, here is an example file that you could add:
```D
// Tests LDC-specific attributes

import ldc.attributes;

version(ThisWillBreakTheCode) {
  struct section {
    string name;
  }
}

// CHECK-DAG: @_D10attributes8myGlobali ={{.*}} section ".mySection"
@section(".mySection") int myGlobal;
```

JohanEngelen:
For reference, relevant PR = #1172

JohanEngelen:
Done: #1243

Issue #259, opened 2012-12-28
Data type real not implemented correctly on architectures different from x86 / x86_64


redstar:
TDPL states that `real` is the floating point type "largest in hardware". On x86 / x86_64 this is 80bit. On PPC/PPC64 there are only 32bit and 64bit floating point types. Such `real` is equal to `double`. There is also a `long double` type with 128bit but this is a software implementation using 2 `double` values.

The LDC IR code gets it right. Sadly, the front end uses hard-coded constants for floating point attributes like mant_dig. This has the following bad effects on PPC64:
- LDC chooses double as the LLVM type (64bit)
- the frontend uses the compile time constants (default with gcc: 128bit; default with xlC: 64bit) 
- modules like `std.math`do not compile (no support for `doubledouble` type with 106bit mantissa)

A quick fix for a native compiler is to change the definitions in `longdouble.h`.
Clearly, this prevents cross-compiling floating point code across different cpu architectures.

kinke:
The main remaining obstacle here is a software `real_t` type for hosts with a native `real` (exception for Windows hosts: x87), with less precision than the target. The constants in `dmd.root.ctfloat` should be good for all formats; the `real` format selection still needs to be adapted for exotic targets (e.g., IBM z, and PowerPC with their various formats), by someone who knows or reads up on these archs.

PR #643, opened 2014-06-26
Implement longdouble as a wrapper around llvm::APFloat.

redstar:
This commit introduces a new class `ldc::longdouble` which provides `real_t`
semantics based on the target. It uses `llvm::APFloat` as base implementation.

This fixes issue #259.


redstar:
Now based on 0.16.0 master.

PR #654, opened 2014-06-29
Use --gc-sections on Linux.

dnadlinger:
This was painful.

<del>Before merging (or at least releasing), we'll have to decide on a flag to disable `--gc-sections`. If we had a way users to specify linker options to be passed _after_ the builtin/ldc.conf ones (which might also be interesting for the curl issues, #652), `--no-gc-sections` would do the trick.</del>

There is `-disable-linker-strip-dead` now. Choosing a name was not easy – `-no-gc-sections` would make no sense to the casual user, and e.g. on OS X the linker option is called `-dead_strip`.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -377,6 +377,10 @@ cl::opt<bool> linkonceTemplates("linkonce-templates",
     cl::desc("Use linkonce_odr linkage for template symbols instead of weak_odr"),
     cl::ZeroOrMore);
 
+cl::opt<bool> disableLinkerStripDead("disable-linker-strip-dead",
+    cl::desc("Do not try to remove unused symbols during linking"),
+    cl::init(false));
+
 cl::opt<bool, true> allinst("allinst",
     cl::desc("generate code for all template instantiations"),
     cl::location(global.params.allInst));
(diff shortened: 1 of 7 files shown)


smolt:
Hi all.  Is there a reason this -dead_strip was not added for OSX link path?  I always use when linking D code for iOS (Xcode has it on by default).  I just added it in my local driver/linker.cpp to see how it goes of OS X builds.

dmakarov:
Just lack of testing. -- speaking of which, shouldn't there be at least an open issue in the tracker about setting up CI testing on OS X?

smolt:
Ok.  Including during integration of 2.067 (now perhaps?) might be a good time because folks testing pre-release should be more forgiving if something does break.

smolt:
Hmm, I seem to have a case in which -dead_strip may break something on OS X.  When testing merge-2.067 branch, std.variant unittest release build is failing and not finding the exception handler.

```
$ runtime/phobos2-test-runner std.variant
_Unwind_RaiseException failed with reason code: 5
Uncaught exception - terminating
Abort trap: 6
```

If I remove -dead_strip, or lower optimization to -O1, or add -g flag, then it is ok and passes.  I'll have to investigate more.

PR #688, opened 2014-07-19
Remove magic constants from Target::critsecsize().

redstar:
The type of a critical section is returned by DtoMutexType() and can be used to derive the size.
This is required for correct cross-compilation.


redstar:
`DtoSize_t` requires a context, too. I did not decide if I wanted to use the same approach as in the `target.cpp` class.

PR #723, opened 2014-09-23
some fixes

Trass3r:
collection of local patches


redstar:
I picked the assembly annotator: see #881.

Trass3r:
Ok but please set the author correctly in the future for better tracking.

Issue #702, opened 2014-08-26
Vararg functions does not compile on x86_64


redstar:
The following module

    import core.vararg;
    import core.stdc.stdio;
    
    void log(const(char)* str, ...)
    {
        va_list args;
        version(LDC)
            va_start(args, str);
        else
            va_start(args, __va_argsave);
        vprintf(str, args);
        va_end(args);
    }
    
    void main()
    {
        log("%d %d %d\n", 1, 2, 3);
    }

compiles with dmd but gives an error with ldc:

    ldc_github.d(11): Error: function core.stdc.stdio.vprintf (const(char*) format, __va_list* arg) is not callable using argument types (const(char)*, void*)

Only on x86_64. The implementation of `core.vararg` is wrong.

kinke:
Fixed by https://github.com/ldc-developers/ldc/pull/768 and https://github.com/ldc-developers/druntime/issues/14.

Issue #172, opened 2012-09-29
ABI transformations are never applied for variadic arguments


klickverbot:
There is a big problem with the current ABI code: ABI rewrites (such as for turning a struct into an integer so it gets passed in a single register by LLVM) are only ever applied for formal parameters, but not for variadic parameters. This is a problem on x86_64, where there is no difference between variadic and formal paremeters with regard to the calling convention.

The classification scheme needs to happen at least partly at per-call, not per-declaration basis.

For an example of this, see DMD testcase `test12`.

kinke:
Fixed by https://github.com/ldc-developers/ldc/pull/768.

Issue #73, opened 2012-02-03
core.stdc.stdarg doesn't work on 64 bits


SiegeLord:
```D
import stdarg = core.stdc.stdarg;
import vararg = core.vararg;
import std.stdio;

version(DigitalMars) version(X86_64)
	import core.stdc.stdarg : __va_argsave_t;

// Use core.vararg
void test1(...)
{
	foreach(type; _arguments)
	{
		writeln(vararg.va_arg!(int)(_argptr));
	}
}

// Use core.stdc.stdarg #1
void test2(...)
{
	foreach(type; _arguments)
	{
		writeln(stdarg.va_arg!(int)(cast(stdarg.va_list)_argptr));
	}
}

// Use core.stdc.stdarg #2
void test3(...)
{
	foreach(type; _arguments)
	{
		void[] buffer = new void[type.tsize];
		stdarg.va_arg(cast(stdarg.va_list)_argptr, type, buffer.ptr);
		writeln(*(cast(int*)buffer.ptr));
	}
}

void main()
{
	test1(42); //Works ok
	test2(42); //Prints out junk
	test3(42); //SIGSEGV
}
```

The ```va_arg``` functions in ```core.stdc.stdarg``` don't seem to work. They require a cast (see code) to even compile properly, and after that they don't work. All 3 functions compile and function fine on DMD 2.057 32 and 64 bits.

kinke:
Fixed by https://github.com/ldc-developers/ldc/pull/768 and https://github.com/ldc-developers/druntime/issues/14.

PR #768, opened 2014-10-26
Varargs fix & ABI refactoring

kinke:
This fixes the varargs horror (see https://github.com/ldc-developers/ldc/issues/702, https://github.com/ldc-developers/ldc/issues/172, https://github.com/ldc-developers/ldc/issues/73 and https://github.com/ldc-developers/druntime/pull/14) which proved to be an important obstacle for Win64.
In the course of this PR, things got slightly out of hand ;) and I ended up reworking big parts of LDC's ABI system, mainly the System V AMD64 ABI used on all x86_64 platforms except for Windows.


kinke:
Okay, some updates over the weekend.

Current status with LLVM 3.6 and https://github.com/ldc-developers/druntime/pull/14:
```
98% tests passed, 9 tests failed out of 590

Total Test time (real) = 199.42 sec

The following tests FAILED:
	211 - std.stream (SEGFAULT)
	225 - std.datetime (Failed)
	233 - std.range (Failed)
	268 - std.internal.math.errorfunction (Failed)
	269 - std.internal.math.gammafunction (Failed)
	500 - std.stream-debug (Failed)
	522 - std.range-debug (Failed)
	588 - dmd-testsuite-debug (Failed)
	590 - dmd-testsuite (Failed)
```

kinke:
Yep, Linux x86_64. Thanks for the link, that fuzzer will be very useful indeed. It'd be great if you could take over as soon as you find some time.

kinke:
Update, this time with LLVM 3.5 (although that shouldn't make a huge difference):
```
99% tests passed, 5 tests failed out of 590

Total Test time (real) = 249.28 sec

The following tests FAILED:
	225 - std.datetime (Failed)
	233 - std.range (Failed)
	522 - std.range-debug (Failed)
	588 - dmd-testsuite-debug (Failed)
	590 - dmd-testsuite (Failed)
```

kinke:
I've adapted the fuzzer a lil' bit: https://gist.github.com/kinke/3ce30ba28da30400aba8

kinke:
This PR currently doesn't handle empty structs correctly (they are passed in memory). This has recently been fixed for DMD by https://github.com/D-Programming-Language/dmd/pull/4266.
Clang doesn't pass empty structs at all, they have no corresponding LL function arguments. I hope this can be fixed properly in the DMD front-end, so that we don't have to take care of this in LDC. Dropping the arguments could be useful for 0-length static arrays, static arrays of 0-sized elements and void arguments, too.

kinke:
Another thing the fuzzer reveals immediately is a bug in DMD's toArgTypes() when it comes to small structs (> 8 bytes and <= 16 bytes) containing floating-point values. DMD passes the whole struct in memory if either exactly one eightbyte is a fp number or if both eightbytes are fp numbers and they are not both doubles.
The official docs (and Clang) don't have such a restriction. A simple ```struct { double a; int b; }``` should be passed as if it were two separate arguments; a in a FP register if available, b in a GP register if available.

Related DMD bugs:
* https://issues.dlang.org/show_bug.cgi?id=13955, fixed by https://github.com/D-Programming-Language/dmd/pull/4265
* https://issues.dlang.org/show_bug.cgi?id=13957

kinke:
Okay, I think I figured out the main issue bugging this PR atm. Some struct comparisons fail unexpectedly if they contain padding bytes.

```D
import std.stdio;

struct Small {
    bool a;
    long b;
}

extern(C) void normal(int arg0, Small arg1)
{
    printSmall(arg1, "normal_arg1");

    assert(arg1.a == true);
    assert(arg1.b == 2);
}

extern(C) void structCmp(int arg0, Small arg1)
{
    printSmall(arg1, "structCmp_arg1");

    assert(arg1 == Small(true, 2));
}

void printSmall(ref Small small, string prefix)
{
    long* ptr = cast(long*) &small;
    writeln(prefix, "[0] = ", ptr[0]);
    writeln(prefix, "[1] = ", ptr[1]);
}

void main()
{
    Small arg1 = Small(true, 2);

    printSmall(arg1, "arg1");

    normal(0, arg1);
    structCmp(0, arg1);
}
```
yields something like:
```
arg1[0] = 1
arg1[1] = 2
normal_arg1[0] = 140734108926977
normal_arg1[1] = 2
structCmp_arg1[0] = 140734108926977
structCmp_arg1[1] = 2
core.exception.AssertError@test_reduced.d(20): Assertion failure
```
Relevant LL:
```
  %arg1 = alloca %test_reduced.Small, align 8
  %.structliteral = alloca %test_reduced.Small
  %1 = getelementptr %test_reduced.Small* %.structliteral, i32 0, i32 0
  %2 = bitcast i1* %1 to i8*
  store i8 1, i8* %2
  %3 = load i8* %2
  %4 = bitcast %test_reduced.Small* %.structliteral to i8*
  %.padding = getelementptr i8* %4, i32 1
  call void @llvm.memset.p0i8.i64(i8* %.padding, i8 0, i64 7, i32 1, i1 false)
  %5 = getelementptr %test_reduced.Small* %.structliteral, i32 0, i32 1
  store i64 2, i64* %5
  %6 = load i64* %5
  %7 = bitcast %test_reduced.Small* %arg1 to i8*
  %8 = bitcast %test_reduced.Small* %.structliteral to i8*
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %7, i8* %8, i64 16, i32 1, i1 false)
  %9 = load %test_reduced.Small* %arg1
  call void @_D12test_reduced10printSmallFKS12test_reduced5SmallAyaZv({ i64, i8* } { i64 4, i8* getelementptr inbounds ([5 x i8]* @.str6, i32 0, i32 0) }, %test_reduced.Small* %arg1)
  %10 = load %test_reduced.Small* %arg1
  call void @normal(i32 0, %test_reduced.Small %10)
  %11 = load %test_reduced.Small* %arg1
  call void @structCmp(i32 0, %test_reduced.Small %11)
  ret i32 0
}

define void @structCmp(i32 %arg0_arg, %test_reduced.Small %arg1_arg) #0 {
  %arg0 = alloca i32
  %arg1 = alloca %test_reduced.Small
  %.structliteral = alloca %test_reduced.Small
  store i32 %arg0_arg, i32* %arg0
  store %test_reduced.Small %arg1_arg, %test_reduced.Small* %arg1
  call void @_D12test_reduced10printSmallFKS12test_reduced5SmallAyaZv({ i64, i8* } { i64 14, i8* getelementptr inbounds ([15 x i8]* @.str3, i32 0, i32 0) }, %test_reduced.Small* %arg1)
  %1 = getelementptr %test_reduced.Small* %.structliteral, i32 0, i32 0
  %2 = bitcast i1* %1 to i8*
  store i8 1, i8* %2
  %3 = load i8* %2
  %4 = bitcast %test_reduced.Small* %.structliteral to i8*
  %.padding = getelementptr i8* %4, i32 1
  call void @llvm.memset.p0i8.i64(i8* %.padding, i8 0, i64 7, i32 1, i1 false)
  %5 = getelementptr %test_reduced.Small* %.structliteral, i32 0, i32 1
  store i64 2, i64* %5
  %6 = load i64* %5
  %7 = bitcast %test_reduced.Small* %arg1 to i8*
  %8 = bitcast %test_reduced.Small* %.structliteral to i8*
  %9 = call i32 @memcmp(i8* %7, i8* %8, i64 16)
  %10 = icmp eq i32 %9, 0
  ...
}
```

kinke:
Another, newly introduced issue concerns vararg structs passed in registers. Since I got rid of the unnecessary rewrites to layout-compatible unnamed structs or scalars (for structs with exactly 1 element), these structs are passed by reference:
```D
struct Small {
    bool a;
    long b;
}

extern(C) void vararg(int arg0, ...) {}

extern(C) void normal(int arg0, Small arg1) {}
void main()
{
    Small arg1 = Small(true, 2);
    vararg(0, arg1);
    normal(0, arg1);
}
```
produces these LL calls in main():
```LL
%9 = load %minitest.Small* %arg1
call void (i32, ...)* @vararg(i32 0, %minitest.Small* %arg1)
%10 = load %minitest.Small* %arg1
call void @normal(i32 0, %minitest.Small %10)
```
I don't know where that originates from yet.

kinke:
Regarding the struct comparison issue: ```struct Small { bool a; long b; }``` is passed in two registers; ```a``` (i1) in RSI, ```b``` (i64) in RDX:
```
movq	104(%rsp), %rdx
movzbl	96(%rsp), %esi
```
and stored by the callee at RSP+16 via
```
movb	%sil, %al
...
movq	%rdx, 24(%rsp)
andb	$1, %al
movb	%al, 16(%rsp)
```
I.e., the caller zero-extends the bool's byte to RSI's lower 32-bit half (ESI). The callee then only stores the lowest byte (masked via a seemingly obsolete AND 1) to memory, leaving the 7 padding bytes undefined.

@klickverbot @redstar I could really use some help here.

kinke:
The last commit, while apparently fixing the struct comparison issue (verified using the fuzzer), has sadly introduced 4 new failing unit tests (release version only):
```
98% tests passed, 9 tests failed out of 590

Total Test time (real) = 286.73 sec

The following tests FAILED:
	202 - std.encoding (Failed)
	210 - std.uni (Failed)
	212 - std.regex (Failed)
	225 - std.datetime (Failed)
	233 - std.range (Failed)
	235 - std.typecons (Failed)
	522 - std.range-debug (Failed)
	588 - dmd-testsuite-debug (Failed)
	590 - dmd-testsuite (Failed)
```

kinke:
Hey David,

the last commit fixes the std.range-debug failure, so that finally all debug unit tests pass. Still enough failing release tests though:
```
98% tests passed, 9 tests failed out of 590

Total Test time (real) = 287.88 sec

The following tests FAILED:
	202 - std.encoding (Failed)
	210 - std.uni (Failed)
	212 - std.regex (Failed)
	225 - std.datetime (Failed)
	233 - std.range (Failed)
	235 - std.typecons (Failed)
	242 - std.functional (Failed)
	588 - dmd-testsuite-debug (Failed)
	590 - dmd-testsuite (Failed)
```
I'll try to get dmd-testsuite to compile again next and see if there's some hint as to what goes wrong for optimized builds.

kinke:
A reduced testcase from dmd-testsuite's testabi.d (which fails in release mode):
```D
import core.stdc.stdio;

void print(T)(string prefix, ref T t)
{
    printf(prefix.ptr);
    foreach(ref e; t.tupleof) printf(" %d,", e);
    printf("\n");
}

T test1_out(T)( )
{
        T t;
        print("test1_out pre:", t);
        foreach( i, ref e; t.tupleof )  e = i+1;
        print("test1_out post:", t);
        return t;
}
T test1_inout(T)( T t)
{
        print("test1_inout pre:", t);
        foreach( i, ref e; t.tupleof )  e += 10;
        print("test1_inout post:", t);
        return t;
}

bool test1_call_out(T)()
{
        T t1;
        print("test1_call_out t1 pre:", t1);
        foreach( i, ref e; t1.tupleof ) e = i+1;
        print("test1_call_out t1 post:", t1);
        T t2 = test1_out!(T)();
        print("test1_call_out t2:", t2);

        return t1 == t2;
}
bool test1_call_inout(T)()
{
        T t1;
        print("test1_call_inout t1 pre:", t1);
        foreach( i, ref e; t1.tupleof ) e = i+1;
        print("test1_call_inout t1 post:", t1);
        T t2 = test1_inout!(T)( t1 );
        print("test1_call_inout t2:", t2);
        print("test1_call_inout t1 pre #2:", t1);
        foreach( i, ref e; t1.tupleof ) e += 10;
        print("test1_call_inout t1 post #2:", t1);

        return t1 == t2;
}

struct Test { byte a,b,c; }

void main()
{
    bool resout = test1_call_out!Test();
    bool resinout = test1_call_inout!Test();

    if (!resout)
        printf("out failed\n");
    if (!resinout)
        printf("inout failed\n");    
}
```
Debug yields:
```
test1_call_out t1 pre: 0, 0, 0,
test1_call_out t1 post: 1, 2, 3,
test1_out pre: 0, 0, 0,
test1_out post: 1, 2, 3,
test1_call_out t2: 1, 2, 3,
test1_call_inout t1 pre: 0, 0, 0,
test1_call_inout t1 post: 1, 2, 3,
test1_inout pre: 1, 2, 3,
test1_inout post: 11, 12, 13,
test1_call_inout t2: 11, 12, 13,
test1_call_inout t1 pre #2: 1, 2, 3,
test1_call_inout t1 post #2: 11, 12, 13,
```
Release yields:
```
test1_call_out t1 pre: 0, 0, 0,
test1_call_out t1 post: 1, 2, 3,
test1_out pre: 0, 0, 0,
test1_out post: 1, 2, 3,
test1_call_out t2: -1, -1, -1,   <<< WRONG!
test1_call_inout t1 pre: 0, 0, 0,
test1_call_inout t1 post: 1, 2, 3,
test1_inout pre: 0, 0, 0,   <<< WRONG!
test1_inout post: 0, 0, 0,   <<< WRONG!
test1_call_inout t2: -1, -1, -1,   <<< WRONG!
test1_call_inout t1 pre #2: 1, 2, 3,
test1_call_inout t1 post #2: 11, 12, 13,
out failed
inout failed
```

This fails for structs T of sizes which are not a power of two: 3, 5-7, 9-15 bytes.

kinke:
```
100% tests passed, 0 tests failed out of 590
```
FINALLY! :]

kinke:
There are surely tons of ABI mismatches compared to our previous implementation; I haven't checked those in a while. The new toArgTypes() solution works well with the core.stdc.stdarg implementation in druntime and gets rid of the LDC vararg issues, but obviously shares the bugs with DMD (empty structs, static arrays of size 0, static arrays of empty elements, structs > 8 bytes and <= 16 bytes with exactly one fp eightbyte etc.).
The fuzzer doesn't find any D-only (I've excluded the C++ interop tests since there are simply too many bugs) issues in more than 1,000 iterations (using varargs too).
DMD's toArgTypes() is essentially a cheap hack; something like our LDC implementation would make much more sense. But I don't wanna open a PR for DMD, all my other PRs have been sitting there for months - plus I'm sure the guys would require me to fix core.stdc.stdarg too. ;(

One question which is still open is what types we wanna allow to be passed partially in registers (currently, delegates only). It's an ABI decision and we may use DMD as reference. I just haven't gotten around to test that yet.
Another question mark is the real/imaginary order for complex numbers.

redstar:
This is really great work! I am eager to merge this!
Could you please
- add your druntime change to this PR. Then we can see the test results in Travis-Ci.
- address David's comment about the end.of-line comments.
- check if you can remove the `VALIDATE_AGAINST_OLD_LDC_VERSION` code. This looks like development/dead code.

Trass3r:
> But I don't wanna open a PR for DMD, all my other PRs have been sitting there for months - plus I'm sure the guys would require me to fix core.stdc.stdarg too. ;(

Same here, yep that's how you scare away contributors.

yebblies:
> However, the issue is that we can't really regress DMD, so if we fix an actual ABI bug in toArgTypes, we might need to get somebody (e.g. @yebblies) to fix the DMD backend too.

If you're talking about the float+nonfloat in a struct bug, I don't know how to fix it.  I think all the other struct passing abi bugs have been fixed in master, except maybe one varargs empty struct one.

If it helps with motivation, LDC won't be able to compile DDMD until most of the varargs issues are fixed.

Trass3r:
Is the parameter order fixed? Haven't checked for a while.
https://github.com/D-Programming-Language/dmd/pull/4092

yebblies:
I have no idea.  It does match the parameter order for `extern(C)` and `extern(C++)` functions.

kinke:
The DMD and LDC parameter order is of course still reversed for non-variadic extern(D) functions; e.g., inline assembly in std.math depends on it.
I guess @yebblies meant that DMD doesn't reverse it for extern(C), and well, that's to be expected.
This reversing for extern(D), a legacy from DMD's x86 ABI, obviously clearly contradicts the docs stating that the D calling convention should match C's on all other platforms. But this issue seems to be neglected too, not too surprisingly I may add.

A proper fix for toArgTypes(), incl. float+nonfloat struct, involves a proper implementation of the algorithm in the SysV ABI specs, which is more or less what the code in gen/abi-x86-64.cpp guarded by VALIDATE_AGAINST_OLD_LDC_VERSION does.
And as I already pointed out, toArgTypes() is also broken for static arrays > 1 element but <= 16 bytes (toArgTypes() always passes those in memory instead of trying to fit them into registers as the specs demand).
Other interesting cases are static arrays with 0 elements (according to toArgTypes(), passed in memory) and static arrays of empty types (passed in memory too except if length is 1).
Your fuzzer would need to be enhanced by static arrays to check these issues. I'm not sure how DMD now handles empty structs.

@redstar:
* VALIDATE_AGAINST_OLD_LDC_VERSION is removed. The code may still prove useful someday and can then be found in git history.
* The end-of-line comments are obsolete now after refactoring the rewrites.
* As David pointed out, my PRs https://github.com/ldc-developers/druntime/pull/14 and https://github.com/ldc-developers/dmd-testsuite/pull/3 need to be merged to temporary branches.

yebblies:
> I guess @yebblies meant that DMD doesn't reverse it for extern(C), and well, that's to be expected.
This reversing for extern(D), a legacy from DMD's x86 ABI, obviously clearly contradicts the docs stating that the D calling convention should match C's on all other platforms. But this issue seems to be neglected too, not too surprisingly I may add.

I'm happy to support pull requests for DMD that make extern(D) behave like extern(C), even on win32.  The sooner we normalize the ABIs the better.

Trass3r:
@yebblies well https://github.com/D-Programming-Language/dmd/pull/4092 is a quick sketch. I don't know how to fix it properly.

redstar:
@kinke Thanks. I created branches ldc-vararg (druntime and dmd-testsuite) and merged your commits into these branches.

kinke:
@redstar Thank you, I've rebased this PR and the druntime one, so that you'll be able to merge it painlessly next time (I've just noticed there were merge conflicts). The rebased druntime PR includes your MIPS64 additions which are missing in the current ldc-vararg branch.

redstar:
Ok, first test result is here. There is a problem with LLVM 3.1 and 3.2 - but also with certain LLVM 3.4 builds. LLVM 3.3 and 3.5 are fine.

redstar:
The problem with LLVMM 3.4 is the 32bit build.

kinke:
LLVM 3.1 is fixed, 3.2 still exhibits strange issues with structs whose size are not a power of 2, and the LLVM 3.4 32-bit builds should be fixed with the updated druntime PR. @redstar Please update the temporary ldc-vararg branch.

redstar:
@kinke I updated the ldc-vararg branch with a forced push.

kinke:
Thanks Kai. I can reproduce the runnable/testabi.d dmd-testsuite issue with LLVM 3.2, but only when using -O3. All lower optimization levels (and other LLVM versions) pass that test. Could that be an LLVM 3.2 bug?

kinke:
Alright, the LLVM 3.2 issue is fixed too now. LLVM 3.1-3.5 all green except for std.algorithm kills.

redstar:
I triggered Travis-CI to give you the joy of an all-green build. :-)

redstar:
@kinke  Could you please contact me by mail (kai@redstar.de)? Thanks!!!!

Issue #722, opened 2014-09-22
Offer -ffast-math option(s)


Trass3r:
Important for auto-vectorization for example.

kinke:
LDC 1.5 uses the default LLVM codegen options, among which there are `-fp-contract`, `-enable-unsafe-fp-math`, `-enable-fp-mad`, `-enable-no-infs-fp-math`, `-enable-no-nans-fp-math`, `-enable-no-trapping-fp-math`, `-denormal-fp-math`, `-recip`.

PR #789, opened 2014-11-13
Add various math options.

redstar:
New options are: -enable-fp-mad, -enable-unsafe-fp-math, -enable-no-infs-fp-math,
-enable-no-nans-fp-math and -enable-sign-dependent-rounding-fp-math.

This fixes issue ???. It does not solve the question about good command line interfacing.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -315,6 +315,27 @@ cl::opt<FloatABI::Type> mFloatABI("float-abi",
         clEnumValN(FloatABI::Hard, "hard", "Hardware floating-point ABI and instructions"),
         clEnumValEnd));
 
+cl::opt<bool> enableFPMAD("enable-fp-mad",
+    cl::desc("Enable less precise MAD instructions to be generated"),
+    cl::init(false));
+
+cl::opt<bool> enableUnsafeFPMath("enable-unsafe-fp-math",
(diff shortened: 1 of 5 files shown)


9il:
Please, add `ftp-contract` option.
https://github.com/llvm-mirror/clang/blob/master/test/CodeGen/ffp-contract-option.c

This option is very useful in scientific libs. It allows to use FMA instructions without enabling associative floating point optimisations.

P.S.
What is `-enable-fp-mad` ?

9il:
```C
  /// LessPreciseFPMAD - This flag return true when -enable-fp-mad option
  /// is specified on the command line.  When this flag is off(default), the
  /// code generator is not allowed to generate mad (multiply add) if the
  /// result is "less precise" than doing those operations individually.
``` what the difference between `ffp-contract`?

9il:
O_O, `enable-unsafe-fp-math` do exactly what I want: default FMA without associative floating point optimizations.

JohanEngelen:
Another needed math option is emitting `fast` attribute on e.g. `fmul` and `fadd`. Clang does this with the -ffast-math commandline option.

For all these options, it'd be very good if they can be enabled/disabled per function (e.g. by UDAs) in addition to the commandline switches.

thewilsonator:
Fix according to https://github.com/ldc-developers/ldc/issues/722#issuecomment-338499775

PR #835, opened 2015-01-31
Store IR metadata in hash table (i.e. remove IrDsymbol from Dsymbol)

Safety0ff:
I re-created my old code for removing ```IrDsymbol``` from ```Dsymbol``` and I thought I'd post it here in case it can be used (I think it something similar will be needed for DDMD.)

Some unnecessary code churn seeped in, I'll clean it up if there's interest in actually merging something like this.

I haven't re-benchmarked since https://github.com/ldc-developers/ldc/pull/707 has been merged, so I don't know the impact on memory/performance compared to current master.


redstar:
What was the reason to close this PR?

PR #846, opened 2015-02-20
Updated to Doxygen v1.8.9.1 and changed path 'dmd' to 'dmd2'.

archfrog:
The Doxyfile file hadn't been touched for 6 years so the paths referred to the old DMD project, effectively making the Doxyfile unusable.  Not that it is worth a lot as it is, but even a little help to grok the project is valuable, IMHO.


archfrog:
1. In-lined C++ source in Doxygen HTML docs for off-line study on a tablet.
2. Increased DOT_GRAPH_MAX_NODES to 1,000 because Doxygen complained about it being too small.  This may be ridiculously large, but it works well and doesn't take all that long to build the docs.
3. Added driver and gen/passes subdirectories to make all sources available to Doxygen.

PR #849, opened 2015-02-22
Fix issue 825: Correct size for TypeVector.

Safety0ff:
Since ```sizeof(TypeVector)``` < ```sizeof(TypeBasic)```.
When using ```Type::nullAttributes``` to create a new ```TypeVector```, it will read (```sizeof(TypeBasic)``` - ```sizeof(TypeVector)``` ) past the end of the source ```TypeVector``` if it was allocated with the actual size e.g. with ```new TypeVector```.

Do I create a PR for both LDC and DMD, or just for DMD?


Safety0ff:
Ok, gotcha.

PR #852, opened 2015-02-26
Fix forward-declared LL signature of function _d_eh_personality.

kinke:
This caught my eye while investigating why EH doesn't work on Win64 with current LLVM 3.6 head. It used to work a couple of months ago. This PR doesn't fix the actual problem, it just fixes the LL signature used when invoking the LL ```landingpad``` instruction for Win64 and ARM.
The actual problem seems to be the native exception raised via ```RtlRaiseException()``` in ```ldc.eh2._d_throw_exception()``` ending up unhandled. I got the impression that LLVM is trying to natively support Windows SEH but I have no idea what the status quo is. @redstar Could you possibly have a look at this?

/edit: This may have something to do with it: http://reviews.llvm.org/rL228733

### DIFF
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -919,11 +919,32 @@ static void LLVM_D_BuildRuntimeModule()
     /////////////////////////////////////////////////////////////////////////////////////
     /////////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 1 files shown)


redstar:
Yes, the signature is different. Thanks for noticing this.
The Google guys are working on native SEH support in LLVM, see http://reviews.llvm.org/D7886.

Issue #830, opened 2015-01-21
Why is LDC so much slower than dmd...


MartinNowak:
...and how can we make it faster?

The difference in compilation time is almost 2x, see [dub Travis-CI](https://travis-ci.org/D-Programming-Language/dub/jobs/47783953).
I'm a little fed up by dmd's bad optimizer (and codegen bugs), but switching the compiler would be a much simpler choice if LDC compiled debug builds almost as fast as dmd.
I guess you guys know what the problems are. Have you already tried to optimize compile times?

Trass3r:
https://github.com/ldc-developers/ldc/issues/539
Almost all of the compile time is spent in llvm.

rsw0x:
FWIW, using a parallel makefile I find ldc to be -much- faster than DMD - LDC's optimized compile times can often beat DMD without optimization when using make. DMD seems to choke on input that isn't grouped together in one pass so it doesn't see much of a speedup at all from parallel make.

MartinNowak:
Really?
That sounds weird because LDC uses the same frontend and also suffers from the overhead of separate compilation. With D's module system, it's inherently more efficient to compile multiple modules at once, because you common imports only get parsed once.

Temtaime:
Maybe there's some unoptimized IR is passed to llvm ?
There should be a way to reduce compilation time.
I'm new with llvm but'll try to look.

Issue #539, opened 2013-11-05
Increase performance and decrease memory footprint


redstar:
As reported on the newsgroup, LDC is around 4 times slower then DMD.

My personal experience is that LDC is really fast for small files but speed decreases dramatically with larger files, e.g. `std.datetime` and `std.array`. At the same time, LDC consumes a huge amount of memory (compare with issue #438)!
A possible explanation for the decreased speed could be that the system has to swap out memory to satisfies LDC requirements.

Memory and performance profiling must be performed to find the root cause(s).

PR #856, opened 2015-02-28
Win64 ABI fix: return non-POD structs via sret.

kinke:
And get rid of obsolete integer rewrites on Win64 and PPC and some general rewrites on System V x86_64 platforms.

### DIFF
--- a/gen/abi-generic.h
+++ b/gen/abi-generic.h
@@ -21,6 +21,64 @@
 #include "gen/structs.h"
 #include "gen/tollvm.h"
 
+struct LLTypeMemoryLayout
+{
+    // Structs and static arrays are folded recursively to scalars or anonymous structs.
+    // Pointer types are folded to an integer type.
+    static LLType* fold(LLType* type)
+    {
+        // T* => integer
+        if (type->isPointerTy())
+            return LLIntegerType::get(gIR->context(), getTypeBitSize(type));
+
+        if (LLStructType* structType = isaStruct(type))
+        {
+            unsigned numElements = structType->getNumElements();
+
+            // fold each element
+            std::vector<LLType*> elements;
+            elements.reserve(numElements);
+            for (unsigned i = 0; i < numElements; ++i)
+                elements.push_back(fold(structType->getElementType(i)));
+
+            // single element? then discard wrapping struct
+            if (numElements == 1)
+                return elements[0];
+
+            return LLStructType::get(gIR->context(), elements, structType->isPacked());
+        }
+
+        if (LLArrayType* arrayType = isaArray(type))
+        {
+            unsigned numElements = arrayType->getNumElements();
+            LLType* foldedElementType = fold(arrayType->getElementType());
+
+            // single element? then fold to scalar
+            if (numElements == 1)
+                return foldedElementType;
+
+            // otherwise: convert to struct of N folded elements
+            std::vector<LLType*> elements(numElements, foldedElementType);
+            return LLStructType::get(gIR->context(), elements);
+        }
+
+        return type;
+    }
+
+    // Checks two LLVM types for memory-layout equivalency.
+    static bool typesAreEquivalent(LLType* a, LLType* b)
+    {
+        if (a == b)
+            return true;
+        if (!a || !b)
+            return false;
+
+        return fold(a) == fold(b);
+    }
+};
+
+//////////////////////////////////////////////////////////////////////////////
+
 /// Removes padding fields for (non-union-containing!) structs
 struct RemoveStructPadding : ABIRewrite {
     /// get a rewritten value back to its original form
@@ -58,6 +116,44 @@ struct RemoveStructPadding : ABIRewrite {
  */
 struct IntegerRewrite : ABIRewrite
 {
+    static LLType* getIntegerType(unsigned minSizeInBytes)
+    {
+        if (minSizeInBytes > 8)
+            return NULL;
+
+        unsigned size = minSizeInBytes;
+        switch (minSizeInBytes) {
+          case 0:
+            size = 1;
+            break;
+          case 3:
+            size = 4;
+            break;
+          case 5:
+          case 6:
+          case 7:
+            size = 8;
+            break;
+          default:
+            break;
+        }
+
+        return LLIntegerType::get(gIR->context(), size * 8);
+    }
+
+    static bool isObsoleteFor(LLType* llType)
+    {
+        if (!llType->isSized()) // e.g., opaque types
+        {
+            IF_LOG Logger::cout() << "IntegerRewrite: not rewriting non-sized type "
+                << *llType << '\n';
+            return true;
+        }
+
+        LLType* integerType = getIntegerType(getTypeStoreSize(llType));
+        return LLTypeMemoryLayout::typesAreEquivalent(llType, integerType);
+    }
+
     LLValue* get(Type* dty, DValue* dv)
     {
         LLValue* integer = dv->getRVal();
@@ -77,29 +173,13 @@ struct IntegerRewrite : ABIRewrite
     {
         assert(dty == dv->getType());
         LLValue* address = getAddressOf(dv);
-        LLType* integerType = type(dty, NULL);
+        LLType* integerType = getIntegerType(dty->size());
         return loadFromMemory(address, integerType, ".IntegerRewrite_putResult");
     }
 
     LLType* type(Type* t, LLType*)
     {
-        unsigned size = t->size();
-        switch (size) {
-          case 0:
-            size = 1;
-            break;
-          case 3:
-            size = 4;
-            break;
-          case 5:
-          case 6:
-          case 7:
-            size = 8;
-            break;
-          default:
-            break;
-        }
-        return LLIntegerType::get(gIR->context(), size * 8);
+        return getIntegerType(t->size());
     }
 };
 
--- a/gen/abi-ppc64.cpp
+++ b/gen/abi-ppc64.cpp
@@ -66,15 +66,12 @@ struct PPC64TargetABI : TargetABI {
     void rewriteFunctionType(TypeFunction* tf, IrFuncTy &fty)
     {
         // EXPLICIT PARAMETERS
-
         for (IrFuncTy::ArgIter I = fty.args.begin(), E = fty.args.end(); I != E; ++I)
         {
             IrFuncTyArg& arg = **I;
 
-            if (arg.byref)
-                continue;
-
-            rewriteArgument(fty, arg);
+            if (!arg.byref)
+                rewriteArgument(fty, arg);
         }
     }
 
@@ -86,8 +83,11 @@ struct PPC64TargetABI : TargetABI {
         {
             if (canRewriteAsInt(ty))
             {
-                arg.rewrite = &integerRewrite;
-                arg.ltype = integerRewrite.type(arg.type, arg.ltype);
+                if (!IntegerRewrite::isObsoleteFor(arg.ltype))
+                {
+                    arg.rewrite = &integerRewrite;
+                    arg.ltype = integerRewrite.type(arg.type, arg.ltype);
+                }
             }
             else
             {
--- a/gen/abi-win64.cpp
+++ b/gen/abi-win64.cpp
@@ -47,11 +47,16 @@ struct Win64TargetABI : TargetABI
     void rewriteArgument(IrFuncTy& fty, IrFuncTyArg& arg);
 
 private:
-    // Returns true if the D type is a composite (struct/static array/complex number).
-    bool isComposite(Type* t)
+    // Returns true if the D type is an aggregate:
+    // * struct
+    // * static/dynamic array
+    // * delegate
+    // * complex number
+    bool isAggregate(Type* t)
     {
-        return t->ty == Tstruct || t->ty == Tsarray
-            || t->iscomplex(); // treat complex numbers as structs too
+        TY ty = t->ty;
+        return ty == Tstruct || ty == Tsarray || ty == Tarray || ty == Tdelegate
+            || t->iscomplex();
     }
 
     // Returns true if the D type can be bit-cast to an integer of the same size.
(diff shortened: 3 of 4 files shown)


redstar:
Does experimental mean that you are still working on it?

kinke:
I wanted to check whether that introduces any regressions. Seems like it doesn't for VS 2015, so I guess it doesn't for VS2013 too. And apparently the more thorough typesAreEuivalent() check also works nicely for the System V ABI.
I'm not perfectly happy with that LLTypeMemoryLayout struct in abi-generic.h. So in case someone has a better name and/or location for it, pls tell me.

PS: We're now down to 50 failing test modules for VS2015.

kinke:
@redstar: This one's good to go unless you have a better name and/or location for the LLTypeMemoryLayout struct.

redstar:
I am still unsure about the `LLTypeMemorylayout` solution. This adds a (possible recursive) call to every struct parameter. What is the performance impact? (See #830 and #539.)

gen/abi-generic.h
```diff
@@ -21,6 +21,64 @@
...
+
+            // fold each element
+            std::vector<LLType*> elements;
+            elements.reserve(numElements);
+            for (unsigned i = 0; i < numElements; ++i)
+                elements.push_back(fold(structType->getElementType(i)));
+
+            // single element? then discard wrapping struct
+            if (numElements == 1)
+                return elements[0];
```
redstar: Maybe it is better to check for the one-element case first?
This is really a question - are there one member structs in Phobos and druntime?

kinke:
Subjectively, I haven't noticed any slowdown, but I don't have any numbers (yet). It's not invoked for every struct function parameter; on Win64, only for structs of power-of-2 sizes <= 8 bytes, on PowerPC similar iirc, and for System V only structs which can be passed in registers, i.e., generally <= 16 bytes, so that the recursion depth isn't deep. But currently the results aren't cached for each already visited type, so this could be added if there's really a significant slowdown.

Of course there are single-member structs - in upstream druntime, every associative array is a struct containing an opaque pointer, just to name a popular example. And besides druntime and phobos there's always user code too ;). The many unnecessary rewrites cluttering the LL code produced by LDC are the reason for this more thorough (and hence recursive) check.

kinke:
Alright, some numbers for a rough estimate:
* Clean build directory
* `make ldc2`
* `time make -j3 all druntime-ldc-unittest-debug phobos2-ldc-unittest-debug`, i.e., build druntime and phobos in debug & release + druntime and phobos unittests, debug only

Current LDC head, Linux x64, LLVM 3.5.2 in my Xubuntu VM using 3 of my 4 physical cores:
```
real	11m36.532s
user	33m10.192s
sys	0m36.870s
```
With this PR:
```
real	11m33.704s
user	33m0.260s
sys	0m34.958s
```

JohanEngelen:
@kinke With these changes, I can no longer build certain files in druntime and phobos. For example `runtime\druntime\src\core\sys\windows\stacktrace.d` fails. They all fail with a ICE: 
`Assertion failed: Ty && "Invalid GetElementPtrInst indices for type!", file E:\ldc_msvc\LLVM36\include\llvm/IR/Instructions.h, line 782`

I am using MSVC 2013 and LLVM 3.6. If I revert the changes from this PR, the problem is gone.

Part of the backtrace (line numbers from ldc-merge-2.067 branch):
```
0x000007FED9007642 (0x000000000000000A 0x0000000000000000 0x000000000012C4B8 0x000000000012C5C0), _initptd() + 0x912 bytes(s)
0x000007FED9132044 (0x000000000012CBD0 0x0000000000000240 0x000007FED91F8430 0x00000001420FFA10), abort() + 0x24 bytes(s)
0x000007FED911BF78 (0x00000001420FFA10 0x00000001420F2BD0 0xCCCCCCCC0000030E 0xCCCCCCCCCCCCCCCC), _wassert() + 0x108 bytes(s)
0x000000013FF8D00C (0x0000000000000000 0x000000000012D0D0 0xCCCCCCCCCCCCCCCC 0xCCCCCCCCCCCCCCCC), llvm::checkGEPType() + 0x4C bytes(s), e:\ldc_msvc\llvm36\include\llvm\ir\instructions.h, line 782 + 0x2E byte(s)
0x00000001400ACF14 (0x0000000002F3F598 0x0000000002F387C8 0x000000000012D150 0x000000000012D1B8), llvm::ExtractValueInst::ExtractValueInst() + 0x64 bytes(s), e:\ldc_msvc\llvm36\include\llvm\ir\instructions.h, line 1938 + 0x64 byte(s)
0x00000001400AD70B (0x0000000002F387C8 0x000000000012D1E0 0x000000000012D1B8 0x0000000000000000), llvm::ExtractValueInst::Create() + 0x8B bytes(s), e:\ldc_msvc\llvm36\include\llvm\ir\instructions.h, line 1883 + 0x5A byte(s)
0x00000001400AD9BC (0x0000000002D42F40 0x0000000002F387C8 0x000000000012D330 0x000000000012D250), llvm::IRBuilder<1,llvm::ConstantFolder,llvm::IRBuilderDefaultInserter<1> >::CreateExtractValue() + 0xDC bytes(s), e:\ldc_msvc\llvm36\include\llvm\ir\irbuilder.h, line 1488 + 0x42 byte(s)
0x00000001400BCB59 (0x0000000002F387C8 0x0000000002B37C30 0x0000000002F1B8A0 0x000007FED9000000), DtoAggrPaint() + 0xD9bytes(s), e:\ldc_msvc\ldc\gen\tollvm.cpp, line 890 + 0x79 byte(s)
0x000000014020C848 (0x0000000002D057F8 0x0000000000364CC0 0x0000000002EF6770 0x0000000002BCF290), DtoCallFunction() + 0x11B8 bytes(s), e:\ldc_msvc\ldc\gen\tocall.cpp, line 510 + 0x1D byte(s)
0x00000001401F79DA (0x000000000012DE88 0x0000000002D057F0 0xCCCCCCCCCCCCCCCC 0xCCCCCCCCCCCCCCCC), ToElemVisitor::visit() + 0x1DDA bytes(s), e:\ldc_msvc\ldc\gen\toir.cpp, line 1113 + 0x3A byte(s)
0x000000014001409B (0x0000000002D057F0 0x000000000012DE88 0xCCCCCCCCCCCCCCCC 0xCCCCCCCCCCCCCCCC), CallExp::accept() + 0x3B bytes(s), e:\ldc_msvc\ldc\dmd2\expression.h, line 933 + 0x3B byte(s)
0x00000001401E8737 (0x0000000002D057F0 0x0000000002ECDFA0 0x0000000002F1AEE0 0x0000000002F1AE80), toElem() + 0x47 bytes(s), e:\ldc_msvc\ldc\gen\toir.cpp, line 3091
```

Verbose compilation output right before the crash:
```
* * * * * * * * * DtoCallFunction()
* * * * * * * * * * Building type: pure nothrow @system inout(ulong)[](inout(ulong)[] a)
* * * * * * * * * * * DtoFunctionType(pure nothrow @system inout(ulong)[](inout(ulong)[] a))
* * * * * * * * * * * * Rewriting argument type inout(ulong)[]
* * * * * * * * * * * * * { i64, i64* } => { i64, i64* }*
* * * * * * * * * * * * Final function type: void ({ i64, i64* }*, { i64, i64* }*)
* * * * * * * * * * doing normal arguments
* * * * * * * * * * Arguments so far: (1)
* * * * * * * * * * *   %.rettmp = alloca { i64, i64* }, align 8
* * * * * * * * * * Function type: pure nothrow @system inout(ulong)[](inout(ulong)[] a)
* * * * * * * * * * DtoArgument
* * * * * * * * * * * VarExp::toElem: a @ const(ulong)[]
* * * * * * * * * * * * DtoSymbolAddress ('a' of type 'const(ulong)[]')
* * * * * * * * * * * * * function param
* * * * * * * * * * * * * type: const(ulong)[]
* * * * * * * * * * Rewrite: putParam
* * * * * * * * * * repainting return value from 'inout(ulong)[]' to 'const(ulong)[]'
```

So it seems there is a problem with `inout(T)[] _rawDup(T)(inout(T)[] a)` in `object.di`.

In `tollvm.cpp`:
```
LLValue* DtoAggrPaint(LLValue* aggr, LLType* as)
{
    if (aggr->getType() == as)
        return aggr;

    LLValue* res = llvm::UndefValue::get(as);

    LLValue* V = gIR->ir->CreateExtractValue(aggr, 0);   <-- ASSERTION CRASH
```
called from `tocall.cpp` line 511:
```
    // repaint the type if necessary
    if (resulttype)
    {
        Type* rbase = stripModifiers(resulttype->toBasetype());
        Type* nextbase = stripModifiers(tf->nextOf()->toBasetype());
        if (!rbase->equals(nextbase))
        {
            IF_LOG Logger::println("repainting return value from '%s' to '%s'", tf->nextOf()->toChars(), rbase->toChars());
            retllval->getType()->dump();
            DtoType(rbase)->dump();
            std::cout << std::endl;
            switch(rbase->ty)
            {
            case Tarray:
                if (tf->isref)
                    retllval = DtoBitCast(retllval, DtoType(rbase->pointerTo()));
                else
                retllval = DtoAggrPaint(retllval, DtoType(rbase));  <---------
                break;
```

I do not understand enough of the code to help you fix this :(

redstar:
@kinke Should we revert this PR?

JohanEngelen:
Reverting is OK, but I'd like to understand what is going on too :)
`retllval` is of type `%c = alloca { i64, i64* }, align 8`
which is a pointer (LLVM tells me), and thus `CreateExtractValue` cannot be used on it.
Overall `DtoCallFunction` is very long, and hard to understand :'(

redstar:
Only speculating: the value for `retllval` is a memory allocation. The pointer must be loaded with `DtoLoad`.
e.g. the alloc: https://github.com/ldc-developers/ldc/blob/eadefdc676d9bdca14f4b117f945b251357d45bd/gen/arrays.cpp#L828
and the load: https://github.com/ldc-developers/ldc/blob/eadefdc676d9bdca14f4b117f945b251357d45bd/gen/arrays.cpp#L831

kinke:
Thanks Johan.
> but I'd like to understand what is going on too :)
Heh me too. ;) I've never liked this repainting in case the types differ. In your example, the result (pointer to an array returned via sret) should obviously not be repainted at all, as the formal `inout` array result is simply converted to a `const` array as the single inout parameter is a const array (passed by ref and hence rewritten).
Based on your output, `tf->nextOf()` is `inout(ulong)[]` while `
So applying `stripModifiers()` on the function's formal return type `tf->nextOf()->toBasetype() = inout(ulong)[]` and `rbase = stripModifiers(resulttype->toBasetype())` is `const(ulong)[]`. Stripping the modifiers of the array type itself doesn't help; both types are mutable. The types only differ in the qualifiers of their element type.
I.e., returning `inout(ulong[])` would work, but the LDC code fails for `inout(ulong)[]`.

JohanEngelen:
@kinke Yep, I thought the same and had tried to figure out how to strip all qualifiers (recursively) but couldn't figure it out so quickly.

JohanEngelen:
@kinke @redstar Got it! 
In:
```
Type * stripModifiers( Type * type )
{
    if (type->ty == Tfunction)
        return type;
    
-    return type->castMod(0);
+    return type->unqualify(MODimmutable | MODconst | MODwild);
}
```

`unqualify` works transitively.

JohanEngelen:
@kinke Now the next error (probably related to this PR) is:
```
[12/497] Generating src/core/time.obj
FAILED: cmd.exe /c cd /D E:\ldc_msvc\ldc && E:\ldc_msvc\ldc\build\bin\ldc2.exe --output-o -c -IE:/ldc_msvc/ldc/runtime/druntime/src -IE:/ldc_msvc/ldc/runtime/druntime/src/gc E:/ldc_msvc/ldc/runtime/druntime/src/core/time.d -ofE:/ldc_msvc/ldc/build/runtime/src/core/time.obj -w -d -O3 -release -disable-invariants
E:/ldc_msvc/ldc/build/import\object.di(681): Error: Function type does not match previously declared function with the same mangled name: _d_newarrayU
```

The error is generated in `functions.cpp`. Dumping the LLVM types of the two unequal function types 
```C++
        func->getFunctionType()->dump();
        functype->dump();
```
gives
```LLVM
{ i64, i8* } (%object.TypeInfo*, i64)
void ({ i64, i8* }*, %object.TypeInfo*, i64)
```

Thanks for any hints.

kinke:
I've come up with a +5 lines solution, yours is way more elegant. ;)

Your function signature mismatch appears to be related to sret. `func->getFunctionType()` returns a dynamic array as normal 16-bytes LL struct. `functype` returns it via sret though, i.e., the LL struct is allocated by the caller and the callee gets its address as implicit first parameter.

JohanEngelen:
@kinke OK, I'll look into this tomorrow. 
Note that the problematic function is `extern (C)`. I do not know how C-calling convention handles struct return types. But I'm guessing it is not sret? So perhaps `functype` is the problem? (our `gABI` wrongly applied on `extern (C)` functions?)

kinke:
There's not a lot of diffs between D and C calling conventions, apart from the reversed parameter order (to be fixed sometime) and variadic functions. So on Win64, a 16-bytes struct (dynamic D array) is to be passed by ref and to be returned via sret.

JohanEngelen:
@kinke  OK, I've fixed the function fwd decl, but again hitting on another LLVM assert. I'm starting to wonder if this ever actually worked on Win64!  ;-)

:'(

JohanEngelen:
@kinke In `DtoNewDynArray`, the `CreateCallOrInvoke2` is calling `_d_newarrayU` runtime function, which has an sret parameter (I think because of these changes https://github.com/kinke/ldc/commit/d24fc71a81ccfd04bd5989ba3339bd08f9f60b83, judging from the commit message)
I don't see how that `CreateCallOrInvoke2` call can succeed, with allocating memory and passing the sret argument.

kinke:
@JohanEngelen To be honest, I'm not sure whether I've properly tested the 2 last commits on Win64 (with LLVM 3.7) as they appeared pretty minuscule. I'll have to check that when I get home.
And yep, Win64 is currently the only target which returns dynamic arrays via sret. It doesn't really have to be this way, I just found it appropriate given that the Win64 C calling convention returns all structs > 8 bytes via sret and I think we should treat a dynamic D array as normal struct.

JohanEngelen:
@kinke  I bet a bottle of whisky that it's not going to work with any version of LLVM ;)
I will change the code so that it does not use sret for arrays for now. So then at least I can develop again! 

I agree it is good to follow the Win64 C calling convention, so I will try later.
To get the runtime functions to work with sret is some work, and will add some ugly code; but I'll write it so we can see the result. If there is a way to add the runtime functions properly to the AST as fwd decls (instead of directly as LLVM fwd decls), that would allow us to solve the problem in a nicer way I think (instead of special calling code we could just do a DtoCall?)

JohanEngelen:
The work continues in https://github.com/ldc-developers/ldc/pull/901

PR #858, opened 2015-03-01
CMake fix: Phobos test runners depend on druntime lib.

kinke:
There's no existing test dependency, so druntime may either be missing or not up-to-date when building the Phobos test runners.

Lastly, use debug instead of release druntime for debug Phobos test runners.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -692,6 +692,10 @@ macro(build_test_runner name_suffix d_flags c_flags)
             set(phobos2_bc "")
             compile_phobos2("${flags}" "-unittest${name_suffix}" "" phobos2_o phobos2_bc)
 
+            # We need an actual druntime library for the Phobos test runner.
+            add_test(build-druntime-ldc${name_suffix} "${CMAKE_COMMAND}"
+                --build ${CMAKE_BINARY_DIR} --target druntime-ldc${name_suffix})
+
             set(phobos2-casm phobos2-ldc-casm${name_suffix})
             add_library(${phobos2-casm} STATIC ${ZLIB_C})
             set_target_properties(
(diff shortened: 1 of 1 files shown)


kinke:
From what I've seen, the thing is that there's no druntime-ldc-unittest(-debug).lib, we only build the object files. Otherwise I would have simply linked the Phobos test runners to that (that's what we do for the shared libs).
And only building druntime-ldc-unittest(-debug) + phobos2-ldc-unittest(-debug) doesn't build druntime-ldc(-debug); their only requirement is ldc2. So that's what this PR fixes.

kinke:
> At this point in the build process, bin/ldc2 (.exe, or whatever) already contains a working LDC instance that is correctly configured to build a D program, runtime library included.

What makes you think that? Why should I have to run ```make``` before being able to run ```ctest``` without phobos-test-runner* failing to link, leading to hundreds of tests not being run? This is especially frustrating if that's your very first experience with the LDC tests after building for more than an hour (Win64).
Just to clarify: druntime-ldc* and phobos-ldc* are NOT (re-)built when running the tests. The ldc2 target is (re-)built as both druntime-ldc-unittest* and phobos-ldc-unittest* targets depend on it, and these targets are built via the existing build-druntime-ldc* and build-phobos2-ldc* tests, which the test-runner tests depend on. But druntime-ldc* and phobos2-ldc* are NOT built as prerequisite for any make target built during the tests!
We don't care about Phobos, but we need a druntime library for the phobos test runners, and as there's no static druntime-ldc-unittest(-debug) library, we have to make sure druntime-ldc is there and up-to-date. This is what this little change does, probably saving people quite a bit of frustration.

I found myself in strange situations with outdated druntime-ldc builds after making changes to LDC itself or to druntime parts used by Phobos and then directly running the tests. If you're lucky, you get a linking error, if you're unlucky, the druntime tests work but Phobos fails mysteriously due to an outdated druntime-ldc.

I'd prefer this solution over a single root test making the default target as that default target is not entirely needed - we don't need Phobos for the tests, and building phobos2-ldc on Win64 takes a couple of minutes I'd rather not waste on a continuous basis.

But the real reason for this PR is of course the round number of 500 CTest tests it yields. ;)

kinke:
Okay now I understand. ;)

Yeah corresponding CMake targets for the runners would definitely be more desirable. If I find the time... ;)

PR #860, opened 2015-03-09
Fix cmake bug when ldc is a submodule

smolt:
This change fixes a cmake error that can happen when ldc is included as a git submodule.

    CMake Error at CMakeLists.txt:215 (string):
      string end index: 6 is out of range -1 - 0

The change is a selection of updates from a more recent version of GetGitRevisionDescription.cmake at https://github.com/rpavlik/cmake-modules.

Note for error verification - the error doesn't happen if there is a tag on the submodule commit.


smolt:
Looks like Travis failures are something unrelated to this change.  A test taking too long and getting killed?

redstar:
Yes, `std.algorithm` is often killed.

smolt:
Thanks Kai!  By the way, are the "cherry picked ..." messages in the topic branch commit useful or just noise, or a don't care?

Issue #812, opened 2014-12-17
Segfault when casting a class with a 0 length array


hberntsen:
The following code runs fine with rdmd and gdc, but with ldc2 it generates a segfault after the first printline:
```d
import std.stdio;

interface A {}

class B : A {
    ubyte[0] test;
}

void main() {
    A a = new B();
    writeln("Got new B!");
    B b = cast(B) a;
    writeln("all good!");
}
```
Simply compile this program with ldc2, run it and you get the segfault. I used the prebuilt 0.15.1 binary on Ubuntu 14.10 x86_64. After I changed the length of the ubyte array to 1 it runs fine.

JohanEngelen:
For ubyte[1], a snippet from the outputted LLVM IR:

    %test.B_init = type { %test.B.__vtbl*, i8*, [1 x i8], { %object.Interface* }* }
    %test.B = type { %test.B.__vtbl*, i8*, [1 x i8], { %object.Interface* }* }

and for ubyte[0]:

    %test.B_init = type { %test.B.__vtbl*, i8*, [0 x i8]*, { %object.Interface* }* }
    %test.B = type { %test.B.__vtbl*, i8*, { %object.Interface* }* }

In the case of a zero-length array, no space should be allocated. The LLVM struct type of B is correctly generated, but the initializer is not correct. The bug is fixed by skipping zero-size fields when generating the class initializer (in IrAggr::addFieldInitializers()). Submitting a PR soon.

Thanks for this bug report! I am very new to LDC, and it was a fun bug to fix, just at the right level of difficulty for me without knowledge of LDC's codebase :))

PR #861, opened 2015-03-14
Skip erroneous creation of implicit initializers for zero-length array in aggregates

JohanEngelen:
Fixes issue #812 .

I will create a separate PR that adds a testcase to the testsuite. I do not know how to combine PRs across submodules (I actually have to figure out how to effectively work with submodule branches at all...).


redstar:
Thanks! Looks good to me. Could you please rebase your patch to create only one commit? (E.g. `git rebase -i HEAD~2`).

JohanEngelen:
@redstar , it's one commit now. (thanks for the git command)

PR #864, opened 2015-03-15
Fix #863: Add a linker option -rdynamic when debug info is requested

### LINKED ISSUE #863: Debug symbols not shown in stacktrace on Linux/x64
An assertion failure, e.g. forced by the following `main4.d` application:

    int fun(int x){ assert(x<0); return x+x;}
    void main(){int x=fun(10);}

results in a stack trace without symbols:

    core.exception.AssertError@main4.d(3): Assertion failure
    ----------------
    ./main4() [0x401ddc]
    ./main4() [0x401e02]
    ./main4() [0x40f194]
    ./main4() [0x40f069]
    ./main4() [0x40f105]
    ./main4() [0x40f069]
    ./main4() [0x40efc8]
    ./main4() [0x401f28]
    /lib64/libc.so.6(__libc_start_main+0xf5) [0x7f66b960edc5]

Application is compiled with `-g`.

This bug was uncovered by the fix for issue #115.

dmakarov:
This fixes issue #863.  The -rdynamic is only used when debug info is generated too.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -428,7 +428,11 @@ llvm::DIType ldc::DIBuilder::CreateCompositeType(Type *type)
         );
     }
 
+#if LDC_LLVM_VER >= 307
+    ir->diCompositeType = DBuilder.replaceTemporary(llvm::TempMDType(ir->diCompositeType), static_cast<llvm::MDCompositeType*>(ret.get()));
+#else
     ir->diCompositeType.replaceAllUsesWith(ret);
+#endif
     ir->diCompositeType = ret;
 
     return ret;


driver/linker.cpp
```diff
@@ -177,6 +177,11 @@ static int linkObjToBinaryGcc(bool sharedLib)
     bool addSoname = false;
     switch (global.params.targetTriple.getOS()) {
     case llvm::Triple::Linux:
+        if (opts::debugInfo) {
```
klickverbot: Just use `global.params.symdebug` here and drop all the `cl_options` changes.

dmakarov:
The undefined symbol is used in _D7imports10test11931a6Engine8__mixin23fooMFZv:

```
callq   _D7imports10test11931d13__T6SignalTvZ6Signal3addMFNbNfDFZvZS7imports10test11931d15ConnectionPoint
```

However, it's not defined anywhere. Because of -rdynamic, all symbols are added to the dynamic symbol table, not only the ones that are used. This particular one cannot be resolved, but apparently is not used, since the code links without -rdynamic. Do you have an idea how this ```callq ...``` was generated?

dmakarov:
The same problem exists on OS X:
```
ldmd2 -m64  -g -od. -oflink11931_0 link11931.o test11931a.o test11931b.o test11931c.o test11931d.o
  "__D7imports10test11931d13__T6SignalTvZ6Signal3addMFNbNfDFZvZS7imports10test11931d15ConnectionPoint", referenced from:
      __D7imports10test11931a6Engine8__mixin23fooMFZv in test11931a.o
```
It hasn't been uncovered until now because LDC is not tested on OS X.

dmakarov:
@klickverbot what's the word on this PR? Clearly, it doesn't introduce a new problem, rather it reveals another existing bug.  Shouldn't the issue be closed and a new one open for the actual bug?

dmakarov:
@klickverbot How often do you merge the updates in the frontend? Specifically, will ldc/dmd2 be updated when dmd 2.067 is released? This particular problem of undefined symbol comes from the frontend, and appears to be fixed in the current dmd source base.

PR #865, opened 2015-03-15
Experimental: Don't reverse parameter order for extern(D) on x86_64.

kinke:
Requires https://github.com/ldc-developers/druntime/pull/19 and https://github.com/ldc-developers/phobos/pull/9.
Debug unittests pass on Linux x64 with LLVM 3.5, but dmd-testsuite reveals a serious bug:

``` D
import core.stdc.stdio;
import std.algorithm;

struct Shell
{
    string str;

    const int opCmp(ref const Shell s)
    {
    return std.algorithm.cmp(this.str, s.str);
    }
}

void main()
{
    Shell[3] a;

    a[0].str = "hello";
    a[1].str = "betty";
    a[2].str = "fred";

    a.sort;

    foreach (Shell s; a)
    {
    printf("%.*s\n", s.str.length, s.str.ptr);
    }

    assert(a[0].str == "betty");
    assert(a[1].str == "fred");
    assert(a[2].str == "hello");
}
```

a.sort() sorts the array in reversed order.

### DIFF
--- a/dmd2/clone.c
+++ b/dmd2/clone.c
@@ -507,10 +507,12 @@ FuncDeclaration *buildXopEquals(StructDeclaration *sd, Scope *sc)
     Loc declLoc = Loc();    // loc is unnecessary so __xopEquals is never called directly
     Loc loc = Loc();        // loc is unnecessary so errors are gagged
 
+    // Use C linkage: we need to make sure the arguments order (p, q) isn't reversed, in
+    // order to obtain a signature identical to that of method `S.opEquals(ref const S)`.
     Parameters *parameters = new Parameters;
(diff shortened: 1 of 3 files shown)


kinke:
I've just updated the druntime patch - I've previously missed `TypeInfo_Struct.compare()` which reverses the arguments when invoking `_xopCmp()`. The test case above now works.

kinke:
It finally seems to be working now. The front-end hack is ugly and can't be incorporated as-is, but I guess the proper location to fix it is the upstream front-end.

kinke:
Upstream PR: https://github.com/D-Programming-Language/dmd/pull/5232

kinke:
Closing because this has finally been implemented in v1.29.

PR #878, opened 2015-03-30
Add Coverage Analysis (à la DMD)

JohanEngelen:
This PR adds coverage analysis to LDC2 (cmdline option -cov[=n%]), equal to DMD's coverage analysis (http://dlang.org/code_coverage.html).

Two data structures are added to the module object file for coverage analysis: _d_cover_data and _d_cover_valid. _d_cover_data is an array of the execution count for each line of code (the execution count is atomically incremented at the start of each statement). _d_cover_valid is a bit-array that for each line says whether there is a statement on that line for which the execution count must be printed.
Each module compiled with -cov is registered for coverage output by calling _d_cover_register2() in druntime/rt/cover.d, passing _d_cover_data and _d_cover_valid. Then, druntime will take care of outputting the coverage listing. _d_cover_register2() is called by a shared static constructor.

There is a bug in druntime's interface of _d_cover_register2(). The _d_cover_valid bit array must be passed as a "size_t[] valid", however, the length field of the slice must be the number of bits in the array (number of lines). I will submit a PR to druntime to fix this, but do not know if it will be accepted. The code in this PR is very conservative and creates an oversized array for _d_cover_valid, such that the slice length field is equal to (at least) number of lines.

Tested using MSVC, x64.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -23,6 +23,34 @@
 
 namespace opts {
 
+/* Option parser that defaults to zero when no explicit number is given.
+ * i.e.:  -cov    --> value = 0
+ *        -cov=9  --> value = 9
+ *        -cov=101 --> error, value must be in range [0..100]
+ */ 
+struct CoverageParser : public cl::parser<unsigned char> {
+#if LDC_LLVM_VER >= 307
+    CoverageParser(cl::Option &O) : cl::parser<unsigned char>(O) {}
+#endif
+
+    bool parse(cl::Option &O, llvm::StringRef ArgName, llvm::StringRef Arg, unsigned char &Val)
+    {
+        if (Arg == "") {
+            Val = 0;
+            return false;
+        }
+
+        if (Arg.getAsInteger(0, Val))
+            return O.error("'" + Arg + "' value invalid for required coverage percentage");
+
+        if (Val > 100) {
+            return O.error("Required coverage percentage must be <= 100");
+        }
+        return false;
+    }
+};
+
+
 // Positional options first, in order:
 cl::list<std::string> fileList(
     cl::Positional, cl::desc("files"));
@@ -411,6 +439,12 @@ cl::opt<bool, true> vgc("vgc",
 cl::opt<bool, true, FlagParser> color("color",
     cl::desc("Force colored console output"),
     cl::location(global.params.color));
+
+cl::opt<unsigned char, true, CoverageParser> coverageAnalysis("cov",
+    cl::desc("Compile-in code coverage analysis\n(use -cov=n for n% minimum required coverage)"),
+    cl::location(global.params.covPercent),
+    cl::ValueOptional,
+    cl::init(127));
 
 static cl::extrahelp footer("\n"
 "-d-debug can also be specified without options, in which case it enables all\n"
--- a/gen/coverage.cpp
+++ b/gen/coverage.cpp
@@ -1,0 +1,51 @@
+//===-- gen/coverage.h - Code Coverage Analysis -----------------*- C++ -*-===//
+//
+//                         LDC – the LLVM D compiler
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "gen/coverage.h"
+
+#include "mars.h"
+#include "module.h"
+#include "gen/irstate.h"
+#include "gen/logger.h"
+
+void emitCoverageLinecountInc(Loc &loc) {
+    // Only emit coverage increment for locations in the source of the current module
+    // (for example, 'inlined' methods from other source files should be skipped).
+    if ( global.params.cov && (loc.linnum != 0) && loc.filename
+         && (gIR->module->getModuleIdentifier().compare(loc.filename) == 0) )
+    {
+        unsigned line = loc.linnum-1; // convert to 0-based line# index
+        assert(line < gIR->dmodule->numlines);
+        {
+            IF_LOG Logger::println("Coverage: increment _d_cover_data[%d]", line);
+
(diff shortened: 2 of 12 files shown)


redstar:
Looks nice. I try to do a review this evening.

JohanEngelen:
@klickverbot I was thinking about another solution for the druntime bug, preserving the function's signature. But I think yours is simpler and better.

redstar:
In general, the coding looks good. Could you please check the Travis build? There are compile errors with the different LLVM versions we support.
A PR requires a green Travis build. (At least, a "virtual" green build. There is an out-of-memory problem in the `std.algorithm` unit test which sometimes kills the build. You can ignore this failure.)

gen/coverage.h
```diff
@@ -0,0 +1,22 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef LDC_GEN_COVERAGE_H
+#define LDC_GEN_COVERAGE_H
+
+struct Loc;
+
+void emitCoverageLinecountInc(Loc &loc);
+
+#endif LDC_GEN_COVERAGE_H
```
redstar: Please remove the `LDC_GEN_COVERAGE`. The file fits on a screen so this does not add information.

gen/coverage.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "mars.h"
+#include "module.h"
+#include "gen/irstate.h"
+#include "gen/logger.h"
+
+void emitCoverageLinecountInc(Loc &loc) {
+    // Only emit coverage increment for locations in the source of the current module
+    // (for example, 'inlined' methods from other source files should be skipped).
+    if ( global.params.cov && (loc.linnum != 0) && loc.filename && gIR->module->getName().equals(loc.filename) )
```
redstar: `Module::getName()` is called `Module::getModuleIdentifier` in LLVM 3.1+.

gen/coverage.cpp
```diff
@@ -0,0 +1,50 @@
...
+        assert(line < gIR->dmodule->numlines);
+        {
+            IF_LOG Logger::println("Coverage: increment _d_cover_data[%d]", line);
+
+            // Get GEP into _d_cover_data array
+            LLConstant* idxs[] = { DtoConstUint(0), DtoConstUint(line) };
+            LLValue* ptr = llvm::ConstantExpr::getGetElementPtr(gIR->dmodule->d_cover_data, idxs, true);
+
+            // Do an atomic increment, so this works when multiple threads are executed.
+            gIR->ir->CreateAtomicRMW(
```
redstar: I do not see why the compiler does not like the statement. Please compare with https://github.com/ldc-developers/ldc/blob/master/gen/toir.cpp#L1005.

redstar: It's to late - just remove the `BinOp` in the line below.

driver/cl_options.cpp
```diff
@@ -23,6 +23,32 @@
 
 namespace opts {
 
+/* Option parser that defaults to zero when no explicit number is given.
+ * i.e.:  -cov    --> value = 0
+ *        -cov=9  --> value = 9
+ *        -cov=101 --> error, value must be in range [0..100]
+ */ 
+struct CoverageParser : public cl::parser<unsigned char> {
```
redstar: The build with LLVM 3.7 does not like this parser.

driver/cl_options.cpp
```diff
@@ -23,6 +23,32 @@
...
+    bool parse(cl::Option &O, llvm::StringRef ArgName, llvm::StringRef Arg, unsigned char &Val)
```
redstar: How do I specify a value of `42` with a `char`?

driver/cl_options.cpp
```diff
@@ -23,6 +23,32 @@
...
+        if (Arg.getAsInteger(0, Val))
+            return O.error("'" + Arg + "' value invalid for required coverage percentage");
+
+        if (Val > 100) {
+            return O.error("Required coverage percentage must be <= 100");
+        }
+        return false;
+    }
+
+    const char *getValueName() const  { return "hoi"; }
```
klickverbot: Huh?

driver/cl_options.cpp
klickverbot: 0 <= 42 <= 255?

driver/cl_options.cpp
JohanEngelen: oops!

driver/cl_options.cpp
JohanEngelen: Frankly, LLVM's commandline code is a real pain to work with. Now with the 3.7 changes, and outdated documentation online, I do not know how to make this work.

JohanEngelen: OK, adding a constructor for LLVM 3.7 did the trick. It now builds for me on LLVM 3.4, 3.5, 3.6, and 3.7. Awaiting Travis results.

redstar:
Looks good known. Could you please rebase your commits into one? Then I merge this PR.

redstar:
BTW: Could you please check `ldmd2`? I think the `-cov` switch is parsed but I don`t know if it passed to `ldc2`. The help text is currently disabled. You can provide another PR for this.

JohanEngelen:
@redstar ldmd2 should now correctly handle -cov. I've squashed all commits into a single one.
Thanks!

JohanEngelen:
@klickverbot See https://issues.dlang.org/show_bug.cgi?id=14417 .

JohanEngelen:
@redstar "all" Travis builds are green now.
I've also put in a PR to enable the coverage analysis tests: https://github.com/ldc-developers/dmd-testsuite/pull/7

redstar:
@JohanEngelen Great work! Thank you!

PR #883, opened 2015-04-03
Try the new Travis docker environment

redstar:
Advantages: more CPU, more memory.
Disadvantages: currently not all LLVM versions are available

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,20 +31,12 @@ addons:
 install:
   - if [[ "${LLVM_CONFIG}" == *3\.[567]* ]]; then export CC="gcc-4.9"; export CXX="g++-4.9"; fi
 env:
-  - LLVM_CONFIG="llvm-config-3.1" TEST_DEBUG=1
   - LLVM_CONFIG="llvm-config-3.1"
-  - LLVM_CONFIG="llvm-config-3.2" TEST_DEBUG=1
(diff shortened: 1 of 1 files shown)


redstar:
Read here: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
`sudo` is forbidden but you can download files and uses non-root folders to install.

I am also for dropping support for older LLVM releases.

smolt:
Custom LLVM? does that open door for including TLS support for iOS?

redstar:
@klickverbot  I think we can use the docker environment now. Current limitations:

 - I have not yet figured out the required packages for the multilib 32-bit build
 - LLVM 3.7 only as soon as the release is build
 - no LLVM snapshot builds

redstar:
BTW I did not encounter a failure due to out of memory. :-) I also changed `-j2` to `-j3` because there are now 2 cores instead of 1.5.

redstar:
I am still trying to get the best setup here. The hint with `linux-libc-dev:i386` was good. `libcurl` is still missing, too.

redstar:
With `-j1` the OOM is gone.

redstar:
Not yet perfect but the build times are good: LLVM 3.2 build is 16min for release only in old environment and 19min for release+debug in docker environment. For LLVM 3.3 it is 25min vs. 20min.

PR #884, opened 2015-04-03
Remove most uses of FlagParser class.

redstar:
This should un-break the LLVM 3.7 build.
The deprecated switches -enable-boundscheck and -disable-boundscheck
are not available in LLVM 3.7 builds and will be removed in 0.16.0.


redstar:
I am working on a `FlagParser` based on the new LLVM 3.7 code.

PR #886, opened 2015-04-04
Allocate structs on the heap via runtime function _d_newitem.

kinke:
This is required for the GC to be able to call their dtors, if need be (by appending a `TypeInfo` pointer after the actual struct in case the struct has a dtor).

`_d_newitemT()` / `_d_newitemiT()` also perform the initialization, so we don't need to take care of that in `toir.cpp` anymore.


kinke:
This fixes the `rt.lifetime` unittests in lines 2567 and 2671. The one for classes in line 2633 still fails.

PR #890, opened 2015-04-06
Add new dmd -verrors option

smolt:
Implement dmd `-verrors` option in ldc2 and ldmd2.  This also fixes the dmd-testsuite failures that use `-verrors` missing.


redstar:
Looks good to me. Please merge.

PR #891, opened 2015-04-06
[WIP] Implement cent/ucent data types.

redstar:
enabled if compiled with gcc.


redstar:
A general uint128_t class is missing (implementation is gcc/clang only). DMD also lacks code generation.

yebblies:
So what's the plan for the cent stuff?

redstar:
See https://github.com/ldc-developers/ldc/pull/1120.

PR #895, opened 2015-04-12
Cache global variables for string literals.

dmakarov:
The problem this PR should solve is to avoid multiple copies of the same string literal in a single module.  The general idea is to cache the global variables and `result`s of `ToElemVisitor`.  If the same string literal is found try to reuse the `result` unless the type is different (e.g. `Tsarray` v `Tarray`). If the type is different try to reuse previously created gvar, unless again the type is different (e.g. char v. wchar). Create new GlobalVariable, if everything else failed.  This fixes externmangle.d test case.


smolt:
I think this is the the right idea.  I've been poking at the dmd front end code that handles string literals and it performs little optimizations such that identical constant strings must be merged otherwise you will get pathological results.

toconstelem.cpp needs similar treatment (StringExp::toConstElem) and tollvm.cpp (DtoConstString) probably too.

I see your use of PrivateLinkage instead of InternalLinkage is nice too on OS X because it allows linker to additionally merge identical strings from different object files.  I wonder how universal that behavior is on other operating systems?

dmakarov:
Should I extend the patch to toconstelem and DtoConstString?

I looked at what clang generates for string literals and noticed it's private linkage (both on OSX and on Linux), instead of internal. So I figured it should be ok to do that in LDC as well.

gen/irstate.h
```diff
@@ -125,6 +127,7 @@ struct IRState
     // module
     Module* dmodule;
     llvm::Module* module;
+    std::map<std::string, std::pair<llvm::GlobalVariable*, DValue*> > stringLiteralCache;
```
klickverbot: I think an `unordered_map` would be preferable here.

gen/toir.cpp
```diff
@@ -370,52 +370,66 @@ class ToElemVisitor : public Visitor
 
         LLType* ct = voidToI8(DtoType(cty));
         LLArrayType* at = LLArrayType::get(ct, e->len+1);
+        llvm::GlobalVariable* gvar = 0;
+        bool cached = gIR->stringLiteralCache.find(e->toChars()) != gIR->stringLiteralCache.end();
```
redstar: I think you should save the value of e->toChars(). It's used in the next line or at the again.
(Maybe someone can look at the toChars() function, too. I found an instance which looked like an use-after-free. In this case saving the value is not save.)

dmakarov: @redstar this is a good point. I was thinking maybe to use some sort of a digest as a key instead of the entire string literal. It's difficult to tell whether that would save space in the map as well, since most string literals are very short.
@klickverbot David and Kai, what do you think about storing the `result` in the map value along with the `gvar`. Is it worth it?

klickverbot: Regarding "some sort of digest", just use `llvm::StringMap`. It is optimized for more or less exactly this use case. Whether to cache the `DValue` depends entirely on the size of the string table for a typical module vs. the typical number of re-uses. I don't have a clear feeling what is better without doing some profiling either.

dmakarov:
Ok, I'll do that. However, I compared the test case failures with and without the PR, and there were no new ones.

PR #896, opened 2015-04-13
Cache global variables for string literals.

dmakarov:
This is redone PR for caching the string literal global variables. Added caching to `DtoConstString` and `ToConstElemVisitor`. Got rid of pair<GlovalVariable*, DValue*>. Switched to use StringRef as the map key, and StringMap as the global variable cache.  One thing I'm not sure about is whether it is necessary to make a copy of e->string.  Since StringExp objects are never destroyed, it seems ok to store e->string pointer in `StringRef key`.  Please, review.

### DIFF
--- a/dmd2/module.h
+++ b/dmd2/module.h
@@ -151,6 +151,11 @@ public:
     void parse();       // syntactic parse
 #endif
     void importAll(Scope *sc);
+#if IN_LLVM
+    using Package::semantic;
+    using Dsymbol::semantic2;
+    using Dsymbol::semantic3;
+#endif
     void semantic();    // semantic analysis
     void semantic2();   // pass 2 semantic analysis
     void semantic3();   // pass 3 semantic analysis
--- a/gen/irstate.h
+++ b/gen/irstate.h
@@ -30,6 +30,7 @@
 #if LDC_LLVM_VER >= 305
 #include "llvm/IR/CallSite.h"
 #else
+#include <map>
 #include "llvm/Support/CallSite.h"
 #endif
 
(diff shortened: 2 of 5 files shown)


dmakarov:
I reworked the PR one more time.  `llvm::StringMap` does not exist prior to LLVM 3.5. `std::unordered_map` does not exist prior to C++ 2011. So I had to use regular `std::map` for LDC_LLVM_VER < 305.  Moreover the API for StringMap seems to be changed in LLVM 3.6 so this PR doesn't build for llvm 3.6, 3.7. For master branch and LLVM 3.5 it works and passes the tests.  I'm sorry, but I can't spend more time on this.

dmakarov:
Builds and passes the tests, occasionally failing on std/algorithm.d compilation probably due to running out of memory
```
7: Generating std/algorithm-unittest-debug.o
7: make[3]: *** [runtime/std/algorithm-unittest-debug.o] Killed
7: make[3]: *** Deleting file `runtime/std/algorithm-unittest-debug.o'
7: make[2]: *** [runtime/CMakeFiles/phobos2-ldc-unittest-debug.dir/all] Error 2
7: make[1]: *** [runtime/CMakeFiles/phobos2-ldc-unittest-debug.dir/rule] Error 2
7: make: *** [phobos2-ldc-unittest-debug] Error 2
```

redstar:
Really nice work!

dmakarov:
Thank you. I'm happy to help.

dmakarov:
no worries. When you drop support for LLVMs before version 3.5, the ifdes that contain std::map for gvar caching can and should be removed. It's good to keep performance in mind, but at this stage correctness issues and huge memory consumption do more harm to LDC, in my opinion. In the long run LDC has a chance to attract more users than DMD. I even have this crazy idea of adapting clang frontend to parse D, so there's no dependence on DMD frontend. Perhaps, eventually ddmd frontend may be better, but it may be tough to outperfm clang anyway.

PR #899, opened 2015-04-15
LLVM 3.7: More metadata and output stream changes.

redstar:


driver/toobj.cpp
```diff
@@ -110,7 +110,11 @@ static void codegenModule(llvm::TargetMachine &Target, llvm::Module& m,
     Target.addAnalysisPasses(Passes);
 #endif
 
+#if LDC_LLVM_VER >= 307
+    llvm::raw_fd_ostream fout(out);
```
kinke: Do you intentionally copy the stream here?

PR #901, opened 2015-04-15
WIP: enable sret for dynamic arrays

JohanEngelen:
Work In Progress!

The goal is to enable dynarray sret for Win64 and other platforms that wish it.
Making a PR to have Travis check the code on other platforms.
All comments are welcome.

### DIFF
--- a/gen/abi-x86-64.cpp
+++ b/gen/abi-x86-64.cpp
@@ -221,7 +221,7 @@ struct X86_64TargetABI : TargetABI {
 
     llvm::CallingConv::ID callingConv(LINK l);
 
-    bool returnInArg(Type* tf, LINK linkage);
+    bool returnInArg(Type* rt, LINK linkage);
 
     bool passByVal(Type* t);
 
--- a/gen/abi-x86-64.cpp
+++ b/gen/abi-x86-64.cpp
@@ -221,7 +221,7 @@ struct X86_64TargetABI : TargetABI {
 
     llvm::CallingConv::ID callingConv(LINK l);
 
-    bool returnInArg(TypeFunction* tf);
+    bool returnInArg(Type* tf, LINK linkage);
 
     bool passByVal(Type* t);
 
@@ -262,11 +262,7 @@ llvm::CallingConv::ID X86_64TargetABI::callingConv(LINK l) {
     }
 }
 
-bool X86_64TargetABI::returnInArg(TypeFunction* tf) {
-    if (tf->isref)
-        return false;
-
-    Type* rt = tf->next;
+bool X86_64TargetABI::returnInArg(Type* rt, LINK /*linkage*/) {
     return passByVal(rt);
 }
 
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -139,6 +139,9 @@ llvm::Function* LLVM_D_GetRuntimeFunction(const Loc &loc, llvm::Module* target, 
         return fn;
 
     fn = M->getFunction(name);
+
+    std::cout << name << std::endl;
+
     assert(fn && "Runtime function not found.");
 
     LLFunctionType* fnty = fn->getFunctionType();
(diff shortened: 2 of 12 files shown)


kinke:
Oh yes, I'm already loving this `LLVM_D_BuildFunctionFwdDecl()` getting rid of more than 200 lines. :]

gen/abi-x86-64.cpp
```diff
@@ -221,7 +221,7 @@ struct X86_64TargetABI : TargetABI {
 
     llvm::CallingConv::ID callingConv(LINK l);
 
-    bool returnInArg(TypeFunction* tf);
+    bool returnInArg(Type* tf, LINK linkage);
```
kinke: tf => rt

gen/runtime.cpp
```diff
@@ -226,6 +229,40 @@ static void ensureDecl(DECL *decl, const char *msg)
...
+
+    LLFunction *fn = LLFunction::Create(fty, LLGlobalValue::ExternalLinkage, fname, M);
+    if (sret) {
+        fn->addAttribute(1, LDC_ATTRIBUTE(StructRet));
+        fn->addAttribute(1, LDC_ATTRIBUTE(NoAlias));
+    }
+    fn->setCallingConv(gABI->callingConv(linkage));
+    return fn;
+}
+
```
kinke: Alright, this injects the sret parameter.
We additionally need to take care of non-sret dynamic array parameters which will be passed byref. This can probably be achieved in `rt_array()` by returning a pointer to the LL struct.

redstar:
@JohanEngelen You should make the `LLVM_D_BuildFunctionFwdDecl()` part a separate PR. It is useful on its own.

redstar:
@JohanEngelen Could you please write me an email? I am not sure if I have a valid mail address from you.

JohanEngelen:
@redstar I've set my email to public on github.

JohanEngelen:
@redstar :'( I thought C++11 is available generally (it apparently is on win64). But it isn't... I guess it is not possible to require C++11? (LLVM 3.6+ does require it I thought?)

JohanEngelen:
Yeah, ok. I will rewrite things to work with C-arrays, like the LLVM functions do. (maybe faster too, preventing dyn allocs...)

gen/runtime.cpp
JohanEngelen: @kinke Not sure what you mean here. Can you give a D code example?

JohanEngelen: @kinke OK, I now understand what you mean. Basically, much more work is needed. If we treat arrays as aggregates, not only will they be returned by sret, but also *passed* by ref, and so all runtime fwd decls that have array arguments will need their arguments to be 'fixed' have to be modified. Overall, I feel this is the wrong approach to the problem. Can we not just parse a bit of D code with fwd declarations of all these runtime functions? So then the runtime fwd decls can never go out of sync with the rest of the code gen.

kinke: Yes parsing some D code would make more sense, but I have no idea whether that's feasible. https://github.com/D-Programming-Language/druntime/pull/1222 could be useful in that regard, too. Pinging @redstar and @klickverbot.

JohanEngelen: In tollvm.cpp, `DtoIsReturnInArg()` correctly uses gABI to check if sret should be used, however, `DtoIsPassedByRef()` does not. The two are not the same ofcourse, but adding Tarray to `isAggregate` does mean that both returning and passing arrays becomes byref. So `DtoIsPassedByRef()` should be changed to use gABI to check for it?

JohanEngelen:
OK, I got things to work with sret arrays now on Win64. Note that dyn arrays are not passed byref still. For that to work, more work is needed in defining the runtime functions, and also in the calling code. (the helperfunctions should be extended in functionality, and the calling code should be made less ABI aware).

gen/runtime.cpp
kinke: No, I'd use `gABI->passByVal()` (iirc) directly as afaik `DtoIsPassedByRef()` is used in a lot of places. Unfortunately, `Win64TargetAbi::passByVal()` returns false for all types as LLVM's ByVal isn't Win64's ByVal (we use `ExplicitByvalRewrite` for that when rewriting the function type).
As a work-around, we could make `Win64TargetAbi::passByVal()` return true for all types later rewritten by `ExplicitByvalRewrite` and handle this stuff then appropriately when rewriting the function...

JohanEngelen: @kinke  I am kind of demotivated about this undertaking. There is no clear benefit that I see, and this whole 'rewriting' business seems very messy.

JohanEngelen:
trying to pick this up again.
First: refactor how we fwd declare druntime functions, and also how druntime calls are generated. What I remember is that that is somewhat of a mess.

JohanEngelen:
closing PR, going to start from scratch.

kinke:
Great news!

> As a work-around, we could make Win64TargetAbi::passByVal() return true for all types later rewritten by ExplicitByvalRewrite. We would then need to take care of actually rewriting these (then byref) args when rewriting the function type...

I'll take care of that, so that you're free to ask `gABI->passByVal()` whether a type is passed in memory and the LL function parameter is a pointer - if that's needed at all. A clean solution would probably let `gABI->rewriteFunctionType()` rewrite the runtime function as every other function.

PR #902, opened 2015-04-15
Inline assembly fix for MSVC: treat real ptr as 64-bit double.

kinke:
This should allow some inline assembly originally crafted for 80-bit x87 reals to work with 64-bit doubles too, although it would still use the x87 coprocessor instead of trying to exploit enhanced x86 instruction sets.


kinke:
I can't really test it as std.format-unittest-debug fails to compile (`Value does not fit in the Fixup field` in LLVM's `X86AsmBackend.cpp`, haven't investigated further yet).

Issue #855, opened 2015-02-28
LLVM 3.6/3.7: d_do_test fails with "Enforcement failed"


redstar:
The `d_do_test` test runner fails with

    object.Exception@/home/travis/build/ldc-developers/ldc/runtime/phobos/std/stdio.d(2180): Enforcement failed

if compiled with LLVM 3.6 and LLVM 3.7. The dmd test suite is not run in Travis because of this.

redstar:
The exception is caused by the first call to `writef`, independent of the parameters. If you put a `writef("Hello\n");` in the first line of `main()` then the unittest run fails, too.

redstar:
Finally found the culprit: if you use the `-singleobj` command line switch then even a simple hello world type application dies with the exception. This does not happen if you do not use the switch.

klickverbot:
This suggests that the behavior of the LLVM IR module linker changed somehow. It is probably somehow related to static constructors or TLS initialization.

MartinNowak:
FYI, I also get this for any other program I tested with 0.15.2-beta1.

redstar:
Commit 13fb3f2acee8dbf9feaf8128b1a7b0f5eb295ca0 hopefully fixes this error.

redstar:
Test `compilable/test4003.d` still fails because of linking.

klickverbot:
We are not using llvm::Linker anymore.

PR #904, opened 2015-04-19
2nd try at issue #855.

redstar:
I cannot reproduce this issue on my Gentoo Linux. :-(
The fix adds another empty module. It does not crash on Windows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ before_install:
   - echo yes | sudo add-apt-repository ppa:h-rayflood/llvm
 
   # LLVM 3.4 / 3.5 / 3.6 / snapshots
-  - sudo sh -c "echo 'deb http://llvm.org/apt/precise/ llvm-toolchain-precise-3.4 main' >> /etc/apt/sources.list"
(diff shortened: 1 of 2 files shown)


redstar:
Great. I also spent time on this... This makes LLVM 3.6 a no-go for a vanilla install.

redstar:
reat!

You should add Bill Wendling (isanbard@gmail.com) as reviewer. He is the code owner. I think the first comment will be "Test case?". This should go into unittest/Linker/LinkModulesTest.cpp.
If you get the LGTM this week then we can propose this fix for LLVM 3.6.2.

redstar:
reat!

You should add Bill Wendling (isanbard@gmail.com) as reviewer. He is the code owner. I think the first comment will be "Test case?". This should go into unittest/Linker/LinkModulesTest.cpp.
If you get the LGTM this week then we can propose this fix for LLVM 3.6.2.

redstar:
Great!

You should add Bill Wendling (isanbard@gmail.com) as reviewer. He is the code owner. I think the first comment will be "Test case?". This should go into unittest/Linker/LinkModulesTest.cpp.
If you get the LGTM this week then we can propose this fix for LLVM 3.6.2.

redstar:
Obsolete. See #984 for master and #974 for merge-2.067.

PR #911, opened 2015-04-23
Enable passing binAssign expressions as ref/out arguments.

kinke:
Fixes `runnable/testassign.d`:
```D
void test12211()
{
    int a = 0;
    void foo(ref int x)
    {
        assert(x == 10);
        assert(&x == &a);
        x = 3;
    }
    foo(a = 10);
    assert(a == 3);
    foo(a += 7);
    assert(a == 3);
...
}
```
Previously, in `foo(a += 7)` a new lvalue based on the rvalue binAssign result was passed to `foo()` instead of the nested lvalue `a`.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -232,11 +232,14 @@ static DValue* toElemAndCacheLvalue(Expression* e)
     return value;
 }
 
-// Evaluates e and returns
-// * the (casted) nested lvalue if one is found, otherwise
-// * the expression's result.
-static DValue* toElemAndTryGetLvalue(Expression* e)
+// Evaluates e and, if tryGetLvalue is true, returns the
+// (casted) nested lvalue if one is found.
+// Otherwise simply returns the expression's result.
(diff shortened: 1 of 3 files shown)


gen/toir.cpp
```diff
@@ -278,7 +281,8 @@ class ToElemVisitor : public Visitor
 
     void visit(DeclarationExp *e)
     {
-        IF_LOG Logger::print("DeclarationExp::toElem: %s | T=%s\n", e->toChars(), e->type->toChars());
+        IF_LOG Logger::print("DeclarationExp::toElem: %s | T=%s\n", e->toChars(),
+            e->type ? e->type->toChars() : "(null)");
```
kinke: This is unrelated, but caused a segfault while debugging. `DeclarationExp::type` may be null for current front-end (see https://github.com/ldc-developers/ldc/blob/merge-2.067/dmd2/expression.c#L1556).

PR #915, opened 2015-04-25
Fix alias declaration type resolution

### LINKED ISSUE #893: Crash: "Error: size of type <T> is not known"
This happens in ldc 0.15.1
This happens even though the same code builds fine in previous ldc and the last three dmd's

In this case: `Error: size of type ImGuiCol is not known`
where ImGuiCol is simply an `alias int ImGuiCol`.

Repository to reproduce: 
https://github.com/Extrawurst/DerelictImgui

failing travis build:
https://travis-ci.org/Extrawurst/DerelictImgui/jobs/57739655

JohanEngelen:
Fix a bug where the mutable version of a Type is not properly resolved if it contains alias identifiers. This results in a bug later during compilation where stripModifiers() will return a not-fully-resolved type.
Fixes #893

### DIFF
--- a/dmd2/declaration.c
+++ b/dmd2/declaration.c
@@ -557,7 +557,7 @@ void AliasDeclaration::semantic(Scope *sc)
     if (s && ((s->getType() && type->equals(s->getType())) || s->isEnumMember()))
         goto L2;                        // it's a symbolic alias
 
-    type = type->addStorageClass(storage_class);
+    type = type->addSTC(storage_class);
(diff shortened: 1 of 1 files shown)


JohanEngelen:
OK, this is obviously not the correct fix. I don't know how to fix it. DMD is too opaque for me regarding this type business / cached types for different modifiers, etc. 
See https://issues.dlang.org/show_bug.cgi?id=14500
Hoping that a DMD dev can fix it.

PR #916, opened 2015-04-25
Don't initialize a new class instance a 2nd time after _d_newclass().

kinke:


redstar:
I merged this too fast. It adds a couple of new failures.

kinke:
Yep, apparently only works for debug builds. Haven't investigated further; `_d_newclass()` initializes the allocated block with `ClassInfo.init` so it appeared superfluous doing it in LDC again afterwards.

kinke:
I've tracked this down to be related to the `GarbageCollect2Stack` optimization pass.

Test case:
```D
import core.stdc.stdio;

struct S
{
    int store;
    this(int n)
    {
        store = n << 3;
    }
}

void main()
{
    static class C
    {
        S nestedS = 4;
    }

    auto c = new C;
    printf("%d\n", c.nestedS.store);
//    assert(c.nestedS == S(4) && c.nestedS.store == 32);
}
```

This works in debug and with -O1, but not with -O2 unless I use `--disable-gc2stack`.
With -O2, main() is reduced to:
```LL
  %tmp = tail call i32 (i8*, ...)* @printf(i8* getelementptr inbounds ([4 x i8]* @.str2, i64 0, i64 0), i32 undef) ; [#uses = 0 type = i32]
  ret i32 0
```
vs. -O1:
```LL
  %.newclass_gc_alloc = tail call noalias %object.Object* @_d_newclass(%object.TypeInfo_Class* @_D2ci4mainFZ1C7__ClassZ) ; [#uses = 1 type = %object.Object*]
  %1 = getelementptr %object.Object* %.newclass_gc_alloc, i64 1 ; [#uses = 1 type = %object.Object*]
  %2 = bitcast %object.Object* %1 to i32*         ; [#uses = 1 type = i32*]
  %3 = load i32* %2, align 4                      ; [#uses = 1 type = i32]
  %tmp = tail call i32 (i8*, ...)* @printf(i8* getelementptr inbounds ([4 x i8]* @.str2, i64 0, i64 0), i32 %3) ; [#uses = 0 type = i32]
  ret i32 0
```
I particularly love this `i32 undef` as 2nd printf arg in the -O2 version. %)

PR #917, opened 2015-04-25
Don't set default ctor in ClassInfo if it's @disabled.

kinke:
Fixes an issue in `runnable/xtest46.d`.


redstar:
This commit is required - but now the shared lib build is broken in `rt.util.typeinfo`. Any idea?

kinke:
The LLVM 3.4 shared lib builds have always failed in the last few days - I've never tried to find out why though. You sure there's a new issue with this particular commit?

PR #919, opened 2015-04-26
Don't memcmp() empty struct types.

kinke:
Although they occupy 1 byte, D now apparently doesn't care about the actual data when comparing instances for (in)equality.

Fixes `runnable/xtest46.d` and `runnable/cppa.d`.


kinke:
The 3rd commit fixes `runnable/testdt.d`.

PR #920, opened 2015-04-26
Fix a problem with debug info for nested local variables.

redstar:
Debug info for nested local variables must be created for each
function (with different address expression). The resulting
debug info was not stored per function which resulted in the
use of the last created debug info even if it was wrong.

This commit fixes the bug and the current assertion with LLVM 3.7.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -20,6 +20,7 @@
 #include "enum.h"
 #include "module.h"
 #include "mtype.h"
+#include <map>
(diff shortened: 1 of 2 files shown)


redstar:
Yes, I also think that `IrVar` is not the best choice for the mapping. But it was the simple way to fix the bug. I try a refactoring soon. Some kind of `hashmap` would be better than `std::map`. I'll have a look at `llvm::DenseMap`.

PR #946, opened 2015-05-28
Remove extra load on assignment generated only for debug info

klickverbot:
This leads to serious problems for values we really should
not be loading from, such as 100k element static arrays.

It was hard to figure out exactly why we need to emit DWARF
info here, so the easiest solution is to just have a look
at assignments during the next big debug info push.

@liranz


redstar:
During last debug session I noted that ldc sometimes generates unnecessary loads. E.g. storing a value and then loading it. I did not have a closer look. Maybe this is caused by this, too.

PR #970, opened 2015-06-07
Fail when LLVM 3.6 is detected; disable Travis builds accordingly

klickverbot:


redstar:
The linker class is in a single file. We could include the fixed class and use it if we detect LLVM 3.6.

redstar:
Yes, I proposed to include the modified `llvm::Linker` class. But removing the dependency on `llvm::Linker` is another good solution. :-)

PR #976, opened 2015-06-14
Compile debug druntime/Phobos with optimizations on

klickverbot:
We still need to discuss whether it really makes sense to slow
down user code just because -g is passes (as the debug libraries
are still built without -release). But there is no reason to forgo
optimization of them completely, since the instances where you
actually want to debug into the standard library as a user are
very rare. People actually specifying -O3 -g is much more common.


smolt:
I say go for it and add -release to debug build too.  I don't think -g should change nature of code, should just instrument.  I wonder if could build debug libs once with -O3 -release -g and then strip to get release version.  Or generate separate debug info for systems that support.

Unittests for runtime should still have a version without -release to catch library errors.  Maybe still a run without optimization to help catch compiler bugs?

redstar:
There is sometimes different behavior between compiling with optimization enabled or disabled. I think it is good to ship optimized libraries with debug info but I also think that our Travis build should include a library and test run with disabled optimization.

Issue #978, opened 2015-06-14
duplicate symbol when using -lib (works with dmd)


timotheecour:
Please see simple test case here: https://github.com/timotheecour/dsnippet/blob/master/static_lib/

with dmd, this works:
make clean && make lib2 && make reexe2

with ldc, the same thing fails:
make clean && make lib2 && make reexe2

I need to be able to recompile a single file and update my static library/executable with the modified file.
dmd allows this but right now I'm getting errors with ldc:

duplicate symbol _main in:
    /tmp/d31//fun1.o
    /tmp/d31//libfun.a(libfun.o)

This is a major blocker for efficient partial recompilation.

NOTE:
I tried many alternatives for ldc without success, eg:
* http://www.manpages.info/macosx/ld.1.html (-multiply_defined suppress)
* using shared library instead with '-undefined supress -flat_namespace' (works great in simple cases; but in many cases gives obscure undefined errors when building the final executable without -undefined supress, eg on _aaApply)
* using 'ar -rcs libmylib.a *.o' followed after recompiling some object (eg fun1.o) by 'ar -rs libmylib.a fun1.o' => ignores the newly modified object files

timotheecour:
friendly ping? is there any thought on this?
That's a blocker for efficient partial recompilation, which prevents me from switching to ldc.

redstar:
Sorry for the delay - I was busy the last 2 weeks.
The problem is here: https://github.com/ldc-developers/ldc/blob/master/driver/main.cpp#L499
If `-of` and `-lib` are specified then this implies `-singleobj`. dmd does not make this implication but creates a lot of object files.

redstar:
Fixed in master now. (I want happy LDC users!)

timotheecour:
thanks!!

PR #983, opened 2015-06-25
Do not imply ` -singleObj` if creating lib.

redstar:
If a target file name is given (`-of`) and a library is created then
`-singleObj` is implied. This results in behaviour not compatible
with dmd. This commit checks for this situation and does not set
`-singleObj`. This fixes issue #978.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -496,11 +496,7 @@ static void parseCommandLine(int argc, char **argv, Strings &sourceFiles, bool &
         error(Loc(), "flags conflict with -run");
     }
     else if (global.params.objname && sourceFiles.dim > 1) {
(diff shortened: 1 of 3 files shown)


driver/main.cpp
```diff
@@ -1246,7 +1242,7 @@ int main(int argc, char **argv)
         }
 
         m->parse(global.params.doDocComments);
-        m->buildTargetFiles(singleObj);
+        m->buildTargetFiles(singleObj, createSharedLib | createStaticLib);
```
klickverbot: `|`?

redstar: || !

klickverbot: Now that I call efficient communication. ;)

PR #984, opened 2015-06-27
Add modified version of LinkModules.cpp for LLVM 3.6.

redstar:
This fixes a serious bug in the linker code. See
http://reviews.llvm.org/D10301 for the applied diff.


redstar:
Ok, I try it.

redstar:
See #985.

PR #985, opened 2015-06-27
Backport of "Avoid using llvm::Linker" to master.

redstar:
See pull request #974 for the original code.


redstar:
Funny: dmd 2.066.1 crashes when you run `dmd -unittest runnable/traits_getUnitTests.d`.

redstar:
There was a problem with the module ctor/dtor in the original code. It's fixed now, too.

Issue #442, opened 2013-08-16
LDC only compiles with /LARGEADDRESSAWARE:NO on Win64


dymk:
This makes it impossible to use of 2gb of memory, which is very unfortunate given one of the main reasons for compiling 64 bit is to be able to use more than the 32 bit address space. 

I've tried passing the flag -L"/LARGEADDRESSAWARE" and -L"-LARGEADDRESSAWARE", however LDC turns the first into a file to link with, and the second into a /LIBPATH argument, so I've had no luck overriding by passing in an additional flag. 

Is there a way to get around this limitation?

JohanEngelen:
Reopening this issue.
LDC needs `-L"/LARGEADDRESSAWARE"` for debug `-g` builds.
See LLVM bug: https://llvm.org/bugs/show_bug.cgi?id=15393

kinke:
Finally obsolete due to CodeView debug infos and fixed by #1880.

PR #1002, opened 2015-07-19
MSVC: pass /LARGEADDRESSAWARE:NO to MS linker.

kinke:
A work-around for https://github.com/ldc-developers/ldc/issues/442 regarding debug infos.


kinke:
In case there are no objections regarding `ADDITIONAL_DEFAULT_LDC_SWITCHES`, I suggest merging this to facilitate MSVC builds.

Issue #988, opened 2015-07-06
MSVC build, lots and lots of warnings


rikkimax:
Compiler: 0.15.2 beta2 win64
Test program:

```D
import std.stdio:writeln;
void main() {writeln("Hi!");}
```

Compiles + runs fine.
Only issue is well, umm https://gist.github.com/rikkimax/2db36ccd7ef65daeef84
Lots and lots of warnings.

So, given how simple that program is, any chance for a significant decrease in number of second definitions?
Or at the very least, make a note of it somewhere?

redstar:
I noted this in the announcement: http://forum.dlang.org/post/ccuhfwbuzzqvpqpnhfhf@forum.dlang.org
I am investigating the issue.

p0nce:
I have the exact problem :), also compilation is very slow.

redstar:
I am wondering if the root cause is this change mentioned in the LLVM release notes:

    Comdats are now ortogonal to the linkage. LLVM will not create comdats for weak linkage globals and the frontends are responsible for explicitly adding them.

redstar:
Fixed in merge-2.067 branch.

kinke:
The situation has improved a lot, thanks Kai @redstar, but it's still present - 460 warnings left, see
https://ci.appveyor.com/project/kinke/ldc/build/job/etqtastpn8ppi47c

redstar:
I see. I missed some `enum`s or so.

kinke:
Perfect, thanks a lot Kai.

kinke:
Damn, there are still some remains, I should have checked more carefully. :/
Only Phobos test runner fails to link, see https://ci.appveyor.com/project/kinke/ldc/build/job/1qmk2thhqxv5cdsf (https://ci.appveyor.com/project/kinke/ldc/build/1.0.138/job/1qmk2thhqxv5cdsf#L1355 till line 1561).

kinke:
Finally fixed by https://github.com/ldc-developers/ldc/pull/1087 - at least there are no more linking errors when building the runtimes and their test runners.

PR #1015, opened 2015-08-05
Revise CMake scripts some more for MSVC

kinke:
These changes allow for LDC to be built with only a handful of warnings using VS 2015, without the need for any explicit CMake compiler flags.
We make it through warning level 3 by disabling a dozen specific warnings.
I've tested the Debug and RelWithDebInfo CMake build types.

For reference, here are the default CMake 3.3 variables for the Ninja generator and MSVC 2015:
```
-- CMAKE_C_FLAGS:  /DWIN32 /D_WINDOWS /W3
-- CMAKE_C_FLAGS_DEBUG: /D_DEBUG /MDd /Zi /Ob0 /Od /RTC1
-- CMAKE_C_FLAGS_MINSIZEREL: /MD /O1 /Ob1 /D NDEBUG
-- CMAKE_C_FLAGS_RELEASE: /MD /O2 /Ob2 /D NDEBUG
-- CMAKE_C_FLAGS_RELWITHDEBINFO: /MD /Zi /O2 /Ob1 /D NDEBUG
-- CMAKE_CXX_FLAGS:  /DWIN32 /D_WINDOWS /W3 /GR /EHsc
-- CMAKE_CXX_FLAGS_DEBUG: /D_DEBUG /MDd /Zi /Ob0 /Od /RTC1
-- CMAKE_CXX_FLAGS_MINSIZEREL: /MD /O1 /Ob1 /D NDEBUG
-- CMAKE_CXX_FLAGS_RELEASE: /MD /O2 /Ob2 /D NDEBUG
-- CMAKE_CXX_FLAGS_RELWITHDEBINFO: /MD /Zi /O2 /Ob1 /D NDEBUG
```

Issue https://github.com/ldc-developers/ldc/issues/988 about multiple definitions when linking the runtimes is unfortunately unaffected. Interestingly, this issue doesn't arise when using LLVM 3.6 (but does for 3.7).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,24 @@ if(CMAKE_COMPILER_IS_GNUCXX OR (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang"))
     if(CMAKE_COMPILER_IS_GNUCXX AND NOT CMAKE_C_COMPILER_VERSION VERSION_LESS "4.7.0")
         append("-Wno-pedantic" LDC_CXXFLAGS)
     endif()
+elseif(MSVC)
+    # Remove flags here, for exceptions and RTTI.
+    # CL.EXE complains to override flags like "/GR /GR-".
+    string(REGEX REPLACE "(^| )[/-]EH[-cs]*( |$)" "\\2" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    string(REGEX REPLACE "(^| )[/-]GR-?( |$)" "\\2" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    append("/GR- /EHs-c-" CMAKE_CXX_FLAGS)
+    add_definitions(/D_HAS_EXCEPTIONS=0)
+
+    # warning C4018: signed/unsigned mismatch
+    # warning C4101: unreferenced local variable
+    # warning C4102: unreferenced label
+    # warning C4146: unary minus operator applied to unsigned type, result still unsigned
+    # warnings C4244 and C4267: conversion from '...' to '...', possible loss of data
+    # warnings C4456-4459: declaration of '...' hides ...
+    # warning C4624: destructor was implicitly defined as deleted because a base class destructor is inaccessible or deleted
+    # warning C4800: forcing value to bool 'true' or 'false' (performance warning)
+    # warning C4996: we're not using Microsoft's secure stringOp_s() functions
+    append("/wd4018 /wd4101 /wd4102 /wd4146 /wd4244 /wd4267 /wd4456 /wd4457 /wd4458 /wd4459 /wd4624 /wd4800 /wd4996" DMD_CXXFLAGS LDC_CXXFLAGS)
 endif()
 # Append -mminimal-toc for gcc 4.0.x - 4.5.x on ppc64
 if( CMAKE_COMPILER_IS_GNUCXX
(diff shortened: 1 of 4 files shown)


runtime/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set(D_FLAGS               -w;-d                                     CACHE STRING
 set(D_FLAGS_DEBUG         -g                                        CACHE STRING  "Runtime build flags (debug libraries), separated by ;")
 set(D_FLAGS_RELEASE       -O3;-release                              CACHE STRING  "Runtime build flags (release libraries), separated by ;")
 if(MSVC)
-    set(LINK_WITH_MSVCRT  OFF                                       CACHE BOOL    "Link with MSVCRT.LIB instead of LIBCMT.LIB")
+    set(LINK_WITH_MSVCRT  ON                                        CACHE BOOL    "Link with MSVCRT.lib (DLL) instead of LIBCMT.lib (static)")
```
kinke: Default to MSVCRT. Let's not shock beginners by a huge size of a hello-world binary.

runtime/CMakeLists.txt
```diff
@@ -200,11 +199,7 @@ endif()
 #
 # Create configuration files.
 #
-if(MSVC)
-    set(LIB_PATH "\"-L-L${CMAKE_BINARY_DIR}/lib${LIB_SUFFIX}/${CMAKE_BUILD_TYPE}\",")
-else()
-    set(LIB_PATH "\"-L-L${CMAKE_BINARY_DIR}/lib${LIB_SUFFIX}\",")
-endif()
```
kinke: That didn't work for RelWithDebInfo, as the libs aren't in that subdirectory, but directly in `lib`.

runtime/CMakeLists.txt
```diff
@@ -232,38 +227,28 @@ add_custom_command(
...
 endif()
-
-# Always build zlib and other C parts of the runtime in release mode, regardless
-# of what the user chose for LDC itself.
-set(cfiles "${CORE_C};${DCRT_C}")
-if(PHOBOS2_DIR)
-    list(APPEND cfiles "${ZLIB_C}")
-endif()
-set_source_files_properties(${cfiles} PROPERTIES
-    COMPILE_FLAGS "${CMAKE_C_FLAGS} ${CMAKE_C_FLAGS_RELEASE}"
```
kinke: The problem is that these `COMPILE_FLAGS` are appended to the normal compile flags (`CMAKE_C_FLAGS CMAKE_C_FLAGS_<BuildType>`) and don't override them, as `set`ting them would suggest.
That's why I took a different approach by overwriting all `CMAKE_C_FLAGS_<BuildType>` with `CMAKE_C_FLAGS_RELEASE`.

runtime/CMakeLists.txt
```diff
@@ -45,7 +45,6 @@ if(BUILD_SHARED_LIBS)
     set(D_LIBRARY_TYPE SHARED)
 else()
     set(D_LIBRARY_TYPE STATIC)
-    set(CXX_COMPILE_FLAGS " ")
```
kinke: Apparently superfluous, there's no C++ involved in the runtime.

CMakeLists.txt
```diff
@@ -120,6 +120,24 @@ if(CMAKE_COMPILER_IS_GNUCXX OR (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang"))
     if(CMAKE_COMPILER_IS_GNUCXX AND NOT CMAKE_C_COMPILER_VERSION VERSION_LESS "4.7.0")
         append("-Wno-pedantic" LDC_CXXFLAGS)
     endif()
+elseif(MSVC)
+    # Remove flags here, for exceptions and RTTI.
+    # CL.EXE complains to override flags like "/GR /GR-".
+    string(REGEX REPLACE "(^| )[/-]EH[-cs]*( |$)" "\\2" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    string(REGEX REPLACE "(^| )[/-]GR-?( |$)" "\\2" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    append("/GR- /EHs-c-" CMAKE_CXX_FLAGS)
```
kinke: We may have removed `/EHsc` and `/GR` from `CMAKE_CXX_FLAGS`, so make sure to put the inverted flags back into `CMAKE_CXX_FLAGS`, as `EXTRA_CXXFLAGS` isn't used for all C++ files.
I've omitted `/MP` and `/GF`, I'd prefer letting the user pass those `EXTRA_CXXFLAGS` in explicitly via the CMake command line. E.g., `/MP` is useless if a new process is spawned for each C++ source file.

CMakeLists.txt
```diff
@@ -320,23 +338,14 @@ if(GENERATE_OFFTI)
...
 if(LLVM_ENABLE_ASSERTIONS)
-    set(EXTRA_CXXFLAGS "${EXTRA_CXXFLAGS} -UNDEBUG")
-    # avoid msvc warning D9025
-    string(REGEX REPLACE "(^| )[/-]D *NDEBUG( |$)" " " CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
-    string(REGEX REPLACE "(^| )[/-]D *NDEBUG( |$)" " " CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
+    append("-UNDEBUG" EXTRA_CXXFLAGS)
+    # avoid MSVC warning D9025 about "-DNDEBUG ... -UNDEBUG"
+    string(REGEX REPLACE "(^| )[/-]D *NDEBUG( |$)" "\\1-UNDEBUG\\2" CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
+    string(REGEX REPLACE "(^| )[/-]D *NDEBUG( |$)" "\\1-UNDEBUG\\2" CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL}")
+    string(REGEX REPLACE "(^| )[/-]D *NDEBUG( |$)" "\\1-UNDEBUG\\2" CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
```
kinke: I prefer an explicit `-UNDEBUG` in the command line (if NDEBUG was defined in the first place) over simply dropping `-DNDEBUG` silently, which may get easily overlooked.
There are too many C++ flags around (the CMake ones, EXTRA_CXXFLAGS, DMD_CXXFLAGS, LLVM_CXXFLAGS, LDC_CXXFLAGS, various combinations thereof...).

runtime/CMakeLists.txt
```diff
@@ -200,11 +199,7 @@ endif()
 #
 # Create configuration files.
 #
-if(MSVC)
-    set(LIB_PATH "\"-L-L${CMAKE_BINARY_DIR}/lib${LIB_SUFFIX}/${CMAKE_BUILD_TYPE}\",")
-else()
-    set(LIB_PATH "\"-L-L${CMAKE_BINARY_DIR}/lib${LIB_SUFFIX}\",")
-endif()
+set(LIB_PATH "\"-L-L${CMAKE_BINARY_DIR}/lib${LIB_SUFFIX}\",")
```
klickverbot: Could you just move that back into the config file then? (The reason I added it is that the extra build type paths caused a slew of linker warnings on OS X, where they do not exist.)

kinke: Done.

kinke:
> On the surface, the changes look reasonable.

They are ;), I've been using this for weeks and it really pays off to simplify and clean up MSVC builds (and suppress warnings, although atm that's kind of cynical, as due to https://github.com/ldc-developers/ldc/issues/988 there are myriads of warnings when linking any binary).

redstar:
I have not tried this PR. The disabled warnings are something I wanted to do since we dropped `/W3`. Just go on, please.

Issue #995, opened 2015-07-12
Assertion `(T->isOpaque() || V[i]->getType() == T->getElementType(i)) && "Initializer for struct element doesn't match struct element type!"' failed.


redstar:
LDC 0.15.2-beta2 on Linux x86_64.

Happens with current vibe.d master, compiling the auth_digest example.

    ldc2: /xtra/llvm-git/llvm-3.6.0.src/lib/IR/Constants.cpp:954: llvm::ConstantStruct::ConstantStruct(llvm::StructType*, llvm::ArrayRef<llvm::Constant*>): Assertion `(T->isOpaque() || V[i]->getType() == T->getElementType(i)) && "Initializer for struct element doesn't match struct element type!"' failed.
    #0 0x2556da8 llvm::sys::PrintStackTrace(_IO_FILE*) /xtra/llvm-git/llvm-3.6.0.src/lib/Support/Unix/Signals.inc:423:0
    #1 0x2557043 PrintStackTraceSignalHandler(void*) /xtra/llvm-git/llvm-3.6.0.src/lib/Support/Unix/Signals.inc:481:0
    #2 0x2555c65 SignalHandler(int) /xtra/llvm-git/llvm-3.6.0.src/lib/Support/Unix/Signals.inc:198:0
    #3 0x7f5dec3f7d30 (/lib64/libpthread.so.0+0xfd30)
    #4 0x7f5dead69e67 gsignal (/lib64/libc.so.6+0x34e67)
    #5 0x7f5dead6b1ca abort (/lib64/libc.so.6+0x361ca)
    #6 0x7f5dead63026 (/lib64/libc.so.6+0x2e026)
    #7 0x7f5dead630d2 (/lib64/libc.so.6+0x2e0d2)
    #8 0x223f71c llvm::ConstantStruct::ConstantStruct(llvm::StructType*, llvm::ArrayRef<llvm::Constant*>) /xtra/llvm-git/llvm-3.6.0.src/lib/IR/Constants.cpp:952:0
    #9 0x224dfcb llvm::ConstantAggrKeyType<llvm::ConstantStruct>::create(llvm::StructType*) const /xtra/llvm-git/llvm-3.6.0.src/lib/IR/ConstantsContext.h:371:0
[...]

PR #1021, opened 2015-08-15
Workaround for issue #995.

redstar:
An opaque and a non-opaque type may be rewritten differently. The fix is to always rewrite the types.

Also backport of the valist change from merge-2.067.

### DIFF
--- a/gen/abi-aarch64.cpp
+++ b/gen/abi-aarch64.cpp
@@ -1,0 +1,134 @@
+//===-- abi-aarch64.cpp ---------------------------------------------------===//
+//
+//                         LDC – the LLVM D compiler
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+//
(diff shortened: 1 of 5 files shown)


redstar:
I am not happy with this, too. But IMHO vibe.d is too important to not be able to compile it...

kinke:
That's certainly true. But at least please keep it as comment and add a TODO about opaque types, so that we don't forget about this.

redstar:
Glad you asked. I currently suspect that this is an issue with type resolution. The error occurs when I compile 2 modules (`core` and `connectionpool`). Both modules import another module (`driver`) which then `public import`s the module with the offending type (`task`).
From the language point of view, the type is not opaque. The type is known during compilation of both modules. But at vtbl construction time, the type is opaque.... Looks like the front end is confused.

redstar:
Yes, I get 2 different type instances.

redstar:
I am closing this PR as 0.15.2-alpha2 is now obsolete and I did not have encountered this issue in 0.16.0-alpha3.

PR #1030, opened 2015-08-19
EH/cleanup codegen rewrite

### LINKED ISSUE #1010: 2.067 branch: Enormous compile time increase after dtor rework
For example, the std.uni debug mode unit tests take several times longer to compile now (probably more than 10x). 70% of the time is spent in `llvm::MachineModuleInfo::getOrCreateLandingPadInfo(llvm::MachineBasicBlock*)`.

klickverbot:
Fixes #1010.

<del>Compiles all the tests without crashing, but there is a myriad of small runtime issues left to debug.</del> Should be good to go.


kinke:
It's a pretty big diff, and I'm not really familiar with all that EH stuff, so I'm surely not suited as reviewer. ;)
Have you checked LLVM 3.7? I did a couple of days ago on Linux x64, and there's 2 failing assertions in the Phobos tests (1 is std.typecons, the other one std.container.array iirc) due to lifetime issues (RefCounted!T etc.)...

kinke:
Congratz, no regressions on Linux x64! :+1:

PR #1044, opened 2015-08-27
Fix DMD bugzilla 14903 (proper args destruction)

kinke:
A backport of my upstream PR https://github.com/D-Programming-Language/dmd/pull/4885, which won't get merged due to DDMD.
Let's not wait until LDC supports the front-end in D to fix this serious bug. The additional diff won't be relevant anymore anyway.

I've extended the upstream PR by proper evaluation order for array ops, compatible with our args evaluation order in tocall.cpp.

### DIFF
--- a/dmd2/expression.c
+++ b/dmd2/expression.c
@@ -1829,128 +1829,149 @@ bool functionParameters(Loc loc, Scope *sc, TypeFunction *tf,
      */
     if (1)
     {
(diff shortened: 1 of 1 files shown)


kinke:
I still need to add the tests from the upstream PR (+ new ones with throwing dtors) to dmd-testsuite. @klickverbot : glad to say that all 5 original test cases work flawlessly on Windows too with this and your EH rewrite!

PR #1046, opened 2015-08-27
Travis: reverse job order.

kinke:
Currently, the one job taking nearly twice as long as the others, i.e., LLVM 3.6 multilib (~53 mins vs. others averaging in ~29 mins), is started after the first batch of 5 parallel jobs.
Letting it start at the beginning will decrease the overall feedback time by roughly one third!


redstar:
Thanks! That's really good. If somebody knows how to change the result folder of the dmd testsuite for the debug build then we can also run the release and the debug version in parallel. This should reduce this step, too.

PR #1048, opened 2015-08-29
Add Appveyor support for Windows CI

kinke:
https://ci.appveyor.com/project/kinke/ldc/history

2 jobs now (debug & release): Win64, Visual Studio 2015, `RelWithDebInfo` configuration for LLVM and LDC, preferring Ninja over msbuild and using a pre-built LLVM 3.7 RC (assertions enabled) from my DropBox. ;)
The unit tests are run; release ones are allowed to fail (core.thread fails in release). dmd-testsuite is excluded for now.
The resulting 'installed' LDC should be published as artifact, that somehow doesn't work yet.

We have at most 40 minutes before a job is killed and just 1 job at a time.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,6 +23,15 @@ cache:
   # Appveyor configuration isn't changed
   - c:\projects\llvm-x64.7z -> appveyor.yml
 
+environment:
+  matrix:
+    - APPVEYOR_JOB_CONFIG: Debug
+    - APPVEYOR_JOB_CONFIG: Release
+
+matrix:
+  allow_failures:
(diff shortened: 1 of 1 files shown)


kinke:
I've just cancelled the wrong job - when checking things twice (push to official branch + PR), the GitHub integration associates the PR with the Appveyor job checking the push to the official branch.

kinke:
Building the tests serially brings about 10 minutes for the debug job (!), while the release job is slower, probably by just 2 minutes. So release currently fails to finish in 40 minutes. Anyway, merging this as Windows CI is definitely worth it. :)

PR #1050, opened 2015-08-29
Merge 2.067 into master

klickverbot:
Should be in better shape than master now, even though the x86 32 bit tests still don't pass on Travis (they are allowed to fail on master).


smolt:
after this gets into master, will merge-2.067 branch end?  Just thinking about unfinished PRs based on merge-2.067.

Issue #1031, opened 2015-08-20
OS X: Corrupted EH tables (_Unwind_RaiseException) due to system linker bug


klickverbot:
This happens only with LLVM 3.7 (Git commit 8113832 from the current release_37 branch, should be close to final release) on OS X, using the latest merge-2.067 branch (6215acc15a). The symptom is that the `core.time-debug` and `std.variant-debug` tests fail with

```
Fatal error in EH code: _Unwind_RaiseException failed with reason code: 5.
```

This implies that the unwinder reaches the top of the stack without finding any matching catches – which of course is a problem, since druntime makes sure that there is always a top-level try/catch.

Breaking in the `_d_throw_exception` call in the debugger, it seems that either the stack is corrupted in some way or the unwind info is incorrect, as it also cannot find any stack frames beyond the unit test block.

klickverbot:
I'm hoping that this is not #362.

What makes the issue rather annoying to debug is that it does not occur when just compiling and running the unit tests for the modules in question in isolation.

klickverbot:
Only occurs when generating debug info (-g).

klickverbot:
There are some suspicious sizes, but I'm not sure whether this is the problem yet:
```
0x00000BD0, no registers saved LSDA) __D9crashtest18__unittestL1412_20FNeZv
0x000010B0, no registers saved LSDA) __D4core7runtime19defaultTraceHandlerFPvZ16DefaultTraceInfo7opApplyMxFMDFKmKxAaZiZi
0x00001670, no registers saved LSDA) __D9crashtest18__unittestL1507_21FNeZv
```

Note: `unwinddump` is no longer shipped with Xcode, need to build it from https://opensource.apple.com/source/ld64/ld64-136/src/other/unwinddump.cpp.

klickverbot:
Similarly for the druntime test runner:
```
0x00000820, no registers saved) __D4core4time12TickDuration18__unittestL2692_56FZv
0x000009D0, no registers saved) __D4core4time8Duration18__unittestL1512_32FNaNfZv
0x00000AC0, no registers saved) __D4core4time12TickDuration18__unittestL2829_60FZv
0x00000B60, no registers saved) __D4core4time8MonoTime18__unittestL1925_38FNfZv
0x00001040, rbx              ) __D2rt5minfo16__unittestL378_6FZv
[...]

PR #1051, opened 2015-08-29
OS X x86 ABI - fix struct return

smolt:
OS X varies from IA-32 by returning structs of 1 or 2 bytes in EAX and 4 or 8 bytes in EAX:EDX; other structs still use an sret arg.  The return value also is rewritten as an int type for LLVM to use the correct return registers (this matches clang).

Now the resulting IR for
```d
struct PointF {float a, b;}
extern(C) PointF pfret();
```
Matches clang:
```
declare i64 @pfret()
```

Note - also changed so D static arrays are always returned by sret arg.  It doesn't make sense to return a array by value and generates lots of code for bigger arrays.

Test runnable/cabi1.d checks most struct return types and now passes for OS X 32-bit along with runnable/cppa.d.  I have an additional test which checks all struct bytes sizes from 1-8 which should probably be added to the test suite.

The only drawback is that std.datetime release unittest is now failing with `Fatal error in EH code: _Unwind_RaiseException failed with reason code: 5` and appears to be related to #1031.  I am using llvm 3.6.1 to build LDC.

### DIFF
--- a/gen/abi-x86.cpp
+++ b/gen/abi-x86.cpp
@@ -22,7 +22,13 @@
 
 struct X86TargetABI : TargetABI
 {
+    const bool isOSX;
     IntegerRewrite integerRewrite;
+
+    X86TargetABI()
+        : isOSX(global.params.isOSX)
+    {
+    }
 
     llvm::CallingConv::ID callingConv(llvm::FunctionType* ft, LINK l)
     {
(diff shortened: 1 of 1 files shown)


smolt:
Hmm, looking at the build results, runnable/cabi1.d fails for 32-bit linux.  Looks like it already had a problem before this change, which is unfortunate because that is the test I cared about most on linux.

smolt:
@klickverbot - do you think my assessment that linking with your patched ld was enough to show the std.datetime unittest failure was not directly related to this change?  I read through your play-by-play in #1031 (very interesting I might add) enough to get a high-level understanding, but there is still much I don't understand on that issue.

P.S. I think it would be nice to have a way to channel bounties into the new D foundation.

PR #1052, opened 2015-08-30
Travis: try to fix multilib errors with CC=g++

kinke:


kinke:
Okay this seems to work now, `-lstdc++` errors etc. are gone.
Still failing in 32-bit: `runnable/cppa.d` (debug + release, assertion in line 253), `runnable/mars1.d` (segfault in debug).

smolt:
I looked at a -m32 runnable/mars1.d failure on OS X, and if it is the same, it is `test8658` which makes 1000 function calls with a large structure.  Each invocation keeps pushing the stack deeper until SEGV.  Something to do with x86_stdcallcc not cleaning stack.  For grins, I changed the called function `foo8658` to extern(C) and then it passes.  The same error is present in 0.15.2 beta.

kinke:
I don't have a 32-bit Linux set-up right now, so I can't dig into these last merge-2.067 issues. The `runnable/cppa.d` issue in line 253 has to do with empty structs. @klickverbot has changed it in commit `ee50259dfd22c34024c5db6982ac92dd84e3ddb6` so that empty struct parameters are completely ignored before rewriting the function type by the TargetABI implementation, apparently explicitly for 32-bit x86 code.
The tested function uses `extern(C++)` and takes an empty struct as first parameter. So I guess that empty structs cannot be ignored on all platforms (OSX only and not on Linux?). In that case, it should probably be implemented by something like `bool TargetABI::ignoreEmptyStructParams()`.

smolt:
For comparison.  On OS X merge-2.067 -m32, cppa.d passes.  However, these 3 runnables fail:
- runnable/mars1.d debug, as already noted, pre-dates merge-2.067
- runnable/testabi.d release fails because TLS data isn't properly aligned for movaps
- runnable/ldc_extern_weak.d - PR ldc-developers/dmd-testsuite#8 (needs to be updated though)



testabi.d fails for 32-bits -O because:
```
  results_1[0..5] = 0;
```
gets these instructions:
```asm
	movl	__D7testabi9results_1G85i@TLVP-L70$pb(%eax), %eax
	calll	*(%eax)
	xorps	%xmm0, %xmm0
	movaps	%xmm0, (%eax)   <-- expects 128-bit alignment
```
and eax is not 16-byte aligned, it is only 8-byte aligned (64-bit),
yet the .tbss directive has alignment of 2^4 -> 16.

`results_1` is properly aligned when dumping addresses with nm, but gets shifted by 8-bytes after initialized thread_data section is prepended to tbss section.

PR #1059, opened 2015-09-02
OS X - fix so MULTILIB test_runner will link

smolt:
On OS X, test_runners were failing link because various casm libs were ending up in the wrong place.  This puts them back in lib.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -722,6 +722,13 @@ macro(build_test_runner name_suffix d_flags c_flags)
         endif()
     endif()
 endmacro()
+
+# output_path was last left to an arch specific lib by build_runtime, need to
+# change so all lib casms ends up in correct lib dir for linking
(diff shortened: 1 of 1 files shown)


smolt:
I am not sure when it started to fail for me.  These are commands I use:
```
cmake -DLLVM_ROOT_DIR=~/llvm-3.6.1/Release/ -DMULTILIB=1 ../ldc
make -j3 && make -k -j3 test
```
but first test_runner fails to link because libdruntime-ldc-casm.a is in lib32 instead of lib.
```
Output:
----------------------------------------------------------
ld: library not found for -ldruntime-ldc-casm
<end of output>
Test time =   0.25 sec
----------------------------------------------------------
Test Failed.
"build-druntime-test-runner" end time: Sep 02 18:52 PDT
"build-druntime-test-runner" time elapsed: 00:00:00
----------------------------------------------------------
```

PR #1066, opened 2015-09-03
Use thunks instead of calling _d_toObject for D interfaces

### LINKED ISSUE #1065: Calling members of large (>64 kiB) classes through interface reference fails
Still need to get a minimal test case, but the culprit is that we (all versions of LDC) use `_d_toObject` to adjust the interface pointer to the object base pointer before calling the method in the caller, instead of doing the adjustment in a thunk like DMD (and probably also GDC) does.

PR incoming.

klickverbot:
This allows us avoid the problems stemming from the 64 kiB
size limit hack in _d_toObject.

GitHub: Fix #1065.

### DIFF
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -365,32 +365,6 @@ DValue* DtoDynamicCastObject(Loc& loc, DValue* val, Type* _to)
 
 //////////////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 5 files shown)


redstar:
There are some 32bit related errors. E.g. `runnable/closure.d` fails now.

Issue #1061, opened 2015-09-02
Since 0.16.0, can't build any project with DUB whenever it has a dependency


p0nce:
This seriously limits the ability to find more bugs since none of my projects build.

For an example, try to build d-idioms:

```
$ dub --compiler=ldc2
Fetching dmarkdown 0.1.0 (getting selected version)...
Placing dmarkdown 0.1.0 to /Users/ponce/.dub/packages/...
Performing "debug" build using ldc2 for x86_64.
Building dmarkdown 0.1.0 configuration "library"...
Error: failed to create path to file: .dub/obj/../../.dub/packages/dmarkdown-0.1.0/.dub/build/library-debug-posix.osx-x86_64-ldc_0-9D24AF7ACC3A312273AFCBB86B148109/libdmarkdown.a
No such file or directory
ldc2 failed with exit code 1.
```
What changed? It used to work with 0.15.x

(BTW I maintain the DUB PR that allows to use `--compiler=ldc2` on Mac: https://github.com/D-Programming-Language/dub/pull/638 
It would be merged more easily if LDC accepted multiple `-march` flags like DMD (yes, DUB sends many -march flags for some reason)).

**OS**: Mac OS X

p0nce:
Same problem on Windows.

Maybe the problem is the discrepancy between output files:

```
ldc2 -march=x86-64 -lib -of..\..\..\AppData\Roaming\dub\packages\gfm-3.0.3\.dub\
build\library-debug-windows-x86_64-ldc_0-CC1AE8AC3A572E41539A953C6EC83851\libgfm
_core.a -d-debug -g -w -oq -od=.dub/obj -d-version=Have_gfm_core -I..\..\..\AppD
ata\Roaming\dub\packages\gfm-3.0.3\core ..\..\..\AppData\Roaming\dub\packages\gf
m-3.0.3\core\gfm\core\memory.d ..\..\..\AppData\Roaming\dub\packages\gfm-3.0.3\c
ore\gfm\core\package.d ..\..\..\AppData\Roaming\dub\packages\gfm-3.0.3\core\gfm\
core\queue.d
Error: failed to create path to file: .dub/obj\..\..\..\AppData\Roaming\dub\pack
ages\gfm-3.0.3\.dub\build\library-debug-windows-x86_64-ldc_0-CC1AE8AC3A572E41539
A953C6EC83851\libgfm_core.a.lib
no such file or directory
```

schuetzm:
[...]

PR #1069, opened 2015-09-09
Fix for issue #1061.

redstar:
Starting with LLVM 3.7, linkage and COMDAT are 2 different concepts.
This means that LinkageODROnce does not put the object into a COMDAT.
On Windows this resulted in linker error messages.

This PR places all template functions, TypeInfo objects and other
objects into a COMDAT.


redstar:
Well, this fixes issue #988.

PR #1074, opened 2015-09-11
GitHub: Closes #863.

### LINKED ISSUE #863: Debug symbols not shown in stacktrace on Linux/x64

dmakarov:


dmakarov:
What's the problem with merging this PR? Why is it ignored?

dmakarov:
DMD uses ```-rdynamic``` on all platforms, and LDC uses it on OS X anyway, because this is the default for OS X linker.  The patch makes things consistent for Linux platform.

kinke:
@klickverbot: Did you get a chance to evaluate ld's `--export-dynamic`? It's required on Ubuntu 16.04 and really shouldn't be. DMD apparently specifies it in its default config file (for Linux, Solaris and BSD), so it's also active for non-debug builds. That might be a viable option for LDC too as the user can always remove it if need be.

PR #1082, opened 2015-09-16
x86: Only skip empty structs on OS X

klickverbot:
Need to do further tests, but it seems like GCC does expect empty structs to be passed. If this is correct, it would be a Clang++/G++ ABI incompatibility on Linux.


redstar:
I'm merging this because it uncovers the next ABI error.

PR #1083, opened 2015-09-16
Fix more of issue #988.

redstar:
Initializers and thunk code were still missing in COMDAT section.

### DIFF
--- a/gen/declarations.cpp
+++ b/gen/declarations.cpp
@@ -251,6 +251,7 @@ public:
             assert(gvar && "DtoResolveVariable should have created value");
 
             const llvm::GlobalValue::LinkageTypes llLinkage = DtoLinkage(decl);
+            const bool isTemplateInstance = DtoIsTemplateInstance(decl);
 
             // Check if we are defining or just declaring the global in this module.
             if (!(decl->storage_class & STCextern))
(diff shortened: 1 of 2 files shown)


redstar:
@kinke  With this fix there are no more warnings during release build of druntime/phobos.

gen/declarations.cpp
```diff
@@ -266,6 +267,7 @@ class CodegenVisitor : public Visitor {
                         llLinkage, 0,
                         "", // We take on the name of the old global below.
                         gvar->isThreadLocal());
+                    if (isTemplateInstance) SET_COMDAT(newGvar, gIR->module);
```
klickverbot: Can't you just set this based on `llLinkage`?

redstar: At one place I used `if (func->hasLinkOnceLinkage() || func->hasWeakLinkage()) SET_COMDAT(func, gIR->module);`.
I dislike that it is currently easy to not think about COMDATs. What do you think about `DtoLinkage()` and `lowerFuncLinkage()` returning a "need COMDAT flag" as `std::pair<llvm::GlobalValue::LinkageTypes, bool>`?

klickverbot: Yes, this seems more like it.

PR #1087, opened 2015-09-17
Refactor new COMDAT code.

redstar:
Linkage and the COMDAT flag is now returned as std::pair<>.
Also adds more COMDATs.


kinke:
This fixes #988 - apparently for good. ;)

redstar:
@klickverbot Thunk are also in COMDAT but do not use `DtoLinkage()`.

Issue #1084, opened 2015-09-17
ldc2 does not respect -of on Windows


p0nce:
Easy to see with a DUB project with one static depedency.
For example: https://github.com/p0nce/scheme-d/tree/master/examples/repl

**OS:** Win64
**Compiler:** Version: LDC 0.16.0-alpha3

**To reproduce:**
`$ dub build --build=debug --arch=x86_64 --compiler=ldc2 --force -v`

Here this produces the command: 
```
ldc2 -march=x86-64 -lib -of..\..\.dub\build\library-debug-windows-x86_64-ldc_0-9D41B5ED4C8AD317C9307A7BAF5A80A1\libscheme-d.a -d-debug -g -w -oq -od=.dub/obj -d
-version=Have_scheme_d -I..\..\source ..\..\source\schemed\environment.d ..\..\source\schemed\eval.d ..\..\source\schemed\package.d ..\..\source\schemed\parser.d ..\..\source\schemed\types.d
```

However the file produced is named `libscheme-d.a.lib`and not `libscheme-d.a` so DUB fails next.

redstar:
`.a` is not a file extension which is recognized by the MS tool chain. I argue that this is a dub issue.

p0nce:
I'm completely OK for fixing this in DUB, but ldc takes a .a as -of parameter and silently renames it to .lib
I think it should instead do what it is told, and let the link fail.

p0nce:
If DUB pass a .lib path, will another .lib be appended?

redstar:
Ok, I got it. This is a misunderstanding of dmd behaviour. I'll fix it.

redstar:
The issue is fixed in master now. For the naming I have created a dub issue: https://github.com/D-Programming-Language/dub/issues/687

PR #1088, opened 2015-09-17
Do not append extension if library filee name already has one.

redstar:
This is the same behavior as dmd. Fixes issue #1084.


redstar:
I like to keep the comment because it documents a deviation from DMD file handling.

PR #1089, opened 2015-09-17
Treat atomic instructions as intrinsics

kinke:
Whose types aren't rewritten by the regular TargetABI.

Fixes the undefined reference to `llvm_atomic_load()` for MSVC builds.

There's a couple more comparisons against `LLVMintrinsic` alone. And I'm not sure whether we can add some more instructions (fence, bitops) as intrinsics in `DtoIsIntrinsic()`.


redstar:
I would expect that at least the bit operation intrinsics needs to be added. (`LLVMbitop_bt`, `LLVMbitop_btr`, etc.)

PR #1091, opened 2015-09-18
Add predefined version for S390.

redstar:


redstar:
`SystemZ` is indeed better. But there is already some code using `version(S390X)`. I'll try to change this.

redstar:
See https://github.com/D-Programming-Language/dmd/pull/5101 and https://github.com/D-Programming-Language/dlang.org/pull/1100.

PR #1096, opened 2015-09-19
SysV ABI uses %eay for sret argument.

redstar:
This is documented in chapter 3, page 14/15. In contrast, OSX places
the sret argument on the stack.


redstar:
This fixes one part of the failing test `cppa.d`, function `test16()`.

redstar:
@klickverbot The Apple doc says for register `EAX`:

    Used to return integral and pointer values. The caller may also place the address to storage where the callee places its return value in this register.

Maybe you can check this?

redstar:
Hm, after re-reading I think this is wrong.

PR #1106, opened 2015-09-22
Use a more robust way of dumping a LL value to memory.

kinke:
This fixes a previous ICE in https://github.com/ldc-developers/dmd-testsuite/blob/ldc/runnable/sdtor.d#L3331 on Win64. The temporary is rewritten to `(auto x = HasDtortest11763(), x)`, and `x` is somehow already an alloca (most likely due it being an instance of an empty struct with a dummy size of 1 byte).

Not sure if this is enough or whether we should actually make sure these empty structs get represented by a dummy i8 value. Ideally, empty structs wouldn't require any memory at all.

### DIFF
--- a/gen/tocall.cpp
+++ b/gen/tocall.cpp
@@ -966,8 +966,9 @@ DValue* DtoCallFunction(Loc& loc, Type* resulttype, DValue* fnval, Expressions* 
 
     // if we are returning through a pointer arg
     // or if we are returning a reference
+    // or if we are returning a struct or static array
     // make sure we provide a lvalue back!
-    if (retinptr || (tf->isref && returnTy != Tvoid))
+    if (retinptr || (tf->isref && returnTy != Tvoid) || DtoIsPassedByRef(dReturnType))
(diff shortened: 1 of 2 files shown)


kinke:
Yes, that's the issue. So `i8` for all empty structs it shall be.

kinke:
@klickverbot: It is this piece of code which doesn't work correctly; the following yields the same ICE on Linux x64 with LLVM 3.7:
```
struct NoDtortest11763 { int a; }

struct HasDtortest11763
{
    NoDtortest11763 func()
    {
        return NoDtortest11763();
    }
    ~this() {}
}

void main()
{
    HasDtortest11763().func();
}
```
It works if you remove the `int` field from `NoDtortest11763`, as then the empty struct is returned via sret (not sure if that's really the way it should be!).
On Win64, an empty `NoDtortest11763` is returned directly as struct (`type { [1 x i8] }`) and so already exposes the ICE.

kinke:
Agreed. I've also worked on this and came up with something similar, and we seem to share the same problems. ;)

PR #1108, opened 2015-09-23
driver: Only use slow -debuglib libraries if explicitly asked to 

klickverbot:
Should finally settle the long-lasting dilemma of abysmal performance with `-g`. I recently discovered that there really is no justification for automatically enabling the debug libraries at all, because some of the druntime contracts e.g. change AA lookup to be O(N) and similar stuff that is not just a constant overhead.

@redstar: This is my last big change/fix for 0.16.0.


kinke:
Oh wow, Travis is green. :)
Seems like the `build.ninja` file wasn't patched properly for AppVeyor (removing `-g` for `std.string-unittest-debug` via `runtime/std/string-unittest-debug.obj -w -d -g -unittest` => `runtime/std/string-unittest-debug.obj -w -d -unittest`, so the 2 strings need to be adapted - probably an additional `-link-debuglib` after `-g`.

PR #1111, opened 2015-09-23
Revise DtoRawAlloca() alignments

kinke:


gen/statements.cpp
```diff
@@ -1150,7 +1150,7 @@ class ToIRVisitor : public Visitor {
         if (stmt->key)
             keyvar = DtoRawVarDeclaration(stmt->key);
         else
-            keyvar = DtoRawAlloca(keytype, 0, "foreachkey"); // FIXME: align?
```
klickverbot: ?

gen/abi-aarch64.cpp
```diff
@@ -110,14 +110,13 @@ struct AArch64TargetABI : TargetABI
...
     void vaCopy(LLValue* pDest, LLValue* src) {
-        // Analog to va_start, we need to allocate a __va_list struct on the stack first
-        // and set the passed 'dest' char* pointer to its address.
-        LLValue* valistmem = DtoRawAlloca(getValistType(), 0, "__va_list_mem");
-        DtoStore(DtoBitCast(valistmem, getVoidPtrType()), pDest);
-
-        // Now bitcopy the source struct over the destination struct.
-        src = DtoBitCast(src, valistmem->getType());
-        DtoStore(DtoLoad(src), valistmem); // *(__va_list*)dest = *(__va_list*)src
+    // Analog to va_start, we need to allocate a new __va_list struct on the stack,
```
klickverbot: What happened to the indentation here?

kinke: Oh, just copy'n paste from the x86_64 ABI, haven't noticed the additional indent. Will be fixed.

gen/tocall.cpp
```diff
@@ -924,6 +915,8 @@ DValue* DtoCallFunction(Loc& loc, Type* resulttype, DValue* fnval, Expressions*
             IF_LOG Logger::cout() << "final return value: " << *retllval << '\n';
         }
     }
+    else
+        resulttype = dReturnType;
```
kinke: This is just to make sure the returned DValue has a D type, as `resulttype` (param) is NULL in some cases. Do we really need that explicit D result type? Is it used to skip this repainting by providing NULL? Please advise.

gen/statements.cpp
kinke: If `stmt->key` is NULL, `keytype` is `DtoSize_t()` (line 1148). I was pretty sure size_t's get properly aligned by LLVM..

gen/tocall.cpp
klickverbot: I didn't ever quite get around to cleaning up the whole `DValue` mess, so I'm really not sure. However, my guess from quickly skimming the calling code would be that whoever wrote this simply didn't think about this, and your patch is correct.

gen/statements.cpp
klickverbot: Ah, okay, I got confused by the code (why is the keytype assignment not just part of the if statement?). You are right, size_t is aligned correctly.

kinke:
Yeah I like these refactorings too ;) - not sure about the 6 overloads neither (I chose int over size_t for the alignment in order not to confuse the compiler..), but I still think it's better than having to do `->getRVal()', `->getType()` etc. over and over again in a gazillion call sites. ;)

Currently it cannot be used for D structs and static arrays though (`DVarValue::getRVal()` returns a pointer, so it would only dump the pointer and not memcpy the struct/sarray itself). That logic (although a bit flakey if the D type isn't available) is still in `ABIRewrite::storeToMemory()` and could be moved to `DtoAllocaDump()`.

kinke:
I've edited the 2nd commit - it fixes the indentation and handles a missing `resulttype` in DtoCallFunction() more gracefully by setting it early to the function's `returntype`. They can differ, especially wrt. modifiers (`inout`...), so having an explicit `resulttype` is useful (so I've backed-out my original 3rd commit which removed it).
The new 3rd commit clarifies an interface I've just come across (again). ;)

kinke:
Yep, seen it. I have a rough idea -- I tried replacing 2 `DtoAllocaDump() + loadFromMemory()` combos with a single `DtoAllocaDump()` (with desired `asType`).

kinke:
Yep, that was the problem - we're green again.

Issue #1110, opened 2015-09-23
0.16_alpha4: No rule to make target '/usr/lib64/libLLVMTableGen.a'


mleise:
I was building against LLVM 3.7.0 and got the message "No rule to make target '/usr/lib64/libLLVMTableGen.a', needed by 'bin/gen_gccbuiltins'.  Stop."
Note that I am using ccache, so some object file targets are "done" in an instant, but I did not have any issue when I provided the gen_gccbuiltins patch.

klickverbot:
Does the file exist on your system? Otherwise, it might be a bug in the LLVM package you are using.

redstar:
It happens if the LLVM libraries are built as shared libraries (`-DBUILD_SHARED_LIBS=ON´). Then `llvm-config --libs` returns the non-existant static library names which we use for linking.

mleise:
There isn't really an option on the Gentoo ebuilds for shared or static libraries. So this is an upstream bug in llvm 3.7, right?

mleise:
Reported 2014-06-03: https://llvm.org/bugs/show_bug.cgi?id=19937

klickverbot:
A simply workaround would be to add an option like `USE_SHARED_LLVM` to our CMake scripts that people need to set in order to link to a shared LLVM despite this problem. We'd then simple replace the static library paths with the shared ones. I haven't checked whether we can infer this somehow from the `llvm-config` output without the need for an extra flag.

redstar:
Finally fixed with #1155.

PR #1116, opened 2015-09-24
Use `llvm-config --libs` to retrieve list of libraries.

redstar:
Requires to add `-L` switch, too. Also removes some clang options
not recognized by gcc.

Fixes issue #1110.


redstar:
This idea is good but not perfect. Now we are always linking against `libLLVMTableGen.so` and getting duplicate defined command line switches...

PR #1120, opened 2015-09-26
Implement cent/ucent data types.

redstar:
enabled if compiled with gcc.


redstar:
This one was accidental closed because it was against merge-2.067 branch.

redstar:
Now updated for 0.17.0 (v2.068.2).

PR #1124, opened 2015-09-26
Enable OS X build in Travis.

redstar:


kinke:
Very cool. The matrix could look like this (white-list):
```
matrix:
  include:
    - env: TRAVIS_OS_NAME=linux # all LLVM versions
    - env: TRAVIS_OS_NAME=osx   # LLVM 3.6 only
      env: LLVM_CONFIG="llvm-config-3.6"
```

redstar:
The matrix looks better now. Linux uses LLVM 3.7 and 3.6, OS X only LLVM 3.6.

redstar:
It compiles now on OS X but there is still a problem with executing the DMD testsuite.

redstar:
Error is:

    640: Building d_do_test tool
    640: OS: osx
    640: Assertion failed: (!Name.empty() && "Must have a name!"), function findProgramByName, file /private/tmp/llvm3620150726-9734-18sfvps/llvm-3.6.2.src/lib/Support/Unix/Program.inc, line 60.
    640: Error: Error executing /Users/travis/build/ldc-developers/ldc/bin/ldc2: Illegal instruction: 4

jacob-carlborg:
Related: http://stackoverflow.com/questions/14268887/what-is-the-illegal-instruction-4-error-and-why-does-mmacosx-version-min-10 ?

kinke:
https://github.com/ldc-developers/ldc/commit/43ff4373d662377cd46ccdf0c57e47060a968ec6 hopefully fixes this assertion.

kinke:
It does - I've retriggered a Travis build: https://travis-ci.org/ldc-developers/ldc/jobs/84793285
There was a single failure in `runnable/testabi.d` before the job was killed (time-out). If the Travis OSX boxes are underpowered, we may have to split debug & release into 2 jobs again (as for Win64, which isn't even running dmd-testsuite).

PR #1129, opened 2015-09-26
MSVC: Detect and set up MSVC environment for linker via batch file helper

kinke:
A simple approach for setting up a proper environment for the MS linker/librarian. If no MSVC environment is set up, Visual Studio (2013/2015) will be detected automatically, and the MSVC batch file will be invoked before forwarding to link.exe/lib.exe.

This greatly simplifies using LDC on Win64, as one doesn't need an MSVC environment anymore. While it's straight-forward to use LDC in a `VS Native Tools Command Prompt`, using it from GUI IDEs should now be possible without having to touch the environment variables. #1130 further eliminates the need to patch the .conf file whenever the install location changes.

As the latest detected Visual Studio version will be used, a new linker may be used when installing a newer VS. To stick with a fixed version, the user can enforce it by setting the `LDC_VSDIR` environment variable.


driver/exe_path.cpp
```diff
@@ -0,0 +1,47 @@
+#include "exe_path.h"
```
redstar: Could you please add the usual copyright header.

driver/exe_path.h
```diff
@@ -0,0 +1,16 @@
+#ifndef LDC_DRIVER_EXE_PATH_H
```
redstar: Same here.

kinke:
@redstar: (c) headers added.

kinke:
The sooner we merge this, the easier we make it for LDC users on Win64 and myself (wrt. hopping between master and merge-2.068). ;)
In case you're wondering: I've tested the batch file quite extensively, and I'm pretty confident it's doing what it should. I could generalize it for x86 too, as preparation for a possible 32-bit MSVC build in the future; otherwise, I regard this PR as complete.

kinke:
Sure, but the biggest change isn't related to Windows: https://github.com/kinke/ldc/commit/1670323ebb4f9672de75479936925c24f38a6402.
Wrt. the Windows-related changes, it's definitely better than what we currently have and something users demand, see [this exploding forum thread](http://forum.dlang.org/thread/mailman.434.1443140754.22025.digitalmars-d@puremagic.com).

Daniel-N:
This patch is a big improvement, but wouldn't it be worth the extra effort to check the real install location in the registry. I for instance never use the default for any application.

HKEY_CURRENT_USER\Software\Microsoft\VisualStudio\12.0_Config
HKEY_CURRENT_USER\Software\Microsoft\VisualStudio\14.0_Config

kinke:
Thanks for participating, sounds like a good idea.
> I for instance never use the default for any application.

:D

kinke:
Alrighty, this should detect VS properly now (VS 2013 and 2015 only) and is prepared for x86 support. :)

Daniel-N:
Awesome, thanks for the quick update! 
LGTM

kinke:
I don't wanna wait any longer :P - Kai seems busy, so I'll just merge this.

PR #1130, opened 2015-09-27
Strive for a portable LDC installation

kinke:


runtime/CMakeLists.txt
```diff
@@ -200,16 +198,22 @@ endif()
...
+    set(MULTILIB_ADDITIONAL_INSTALL_PATH "\n        \"-L-L%%LDC_BASE_DIR%%/lib${MULTILIB_SUFFIX}\",\n        \"-L--no-warn-search-mismatch\",")
 endif()
 
 if(BUILD_SHARED_LIBS AND (${CMAKE_SYSTEM_NAME} MATCHES "Linux"))
     set(SHARED_LIBS_RPATH "\n        \"-L-rpath=${CMAKE_BINARY_DIR}/lib\",")
 endif()
 
+if(${INCLUDE_INSTALL_DIR} STREQUAL "${CMAKE_INSTALL_PREFIX}/include/d")
+    set(INCLUDE_INSTALL_DIR "%%LDC_BASE_DIR%%/include/d")
+endif()
```
kinke: We wouldn't need to worry about this if we opted for a fixed include dir, not configurable by the user.

driver/configfile.cpp
```diff
@@ -226,19 +226,20 @@ bool ConfigFile::read(const char* argv0, void* mainAddr, const char* filename)
     // handle switches
     if (config_setting_t *sw = config_setting_get_member(root, "switches"))
     {
-        std::string binpathkey = "%%ldcbinarypath%%";
+        std::string baseDirKey = "%%LDC_BASE_DIR%%";
 
-        std::string binpath = sys::path::parent_path(getMainExecutable(argv0, mainAddr));
+        std::string baseDir = sys::path::parent_path(getMainExecutable(argv0, mainAddr));
+        baseDir += "/..";
```
kinke: This will be fixed when https://github.com/kinke/ldc/commit/f340b0fb3ab0664b52446abdf5626a351cae131a makes it in.

kinke:
Are we okay with this direction? Or should we restrict defaulting to %%LDC_BASE_DIR%% in the _install.conf file on Windows only?

kinke:
Closing this as the AppVeyor job now simply performs a find-and-replace in the .conf file to make it portable. I guess it makes sense to adapt the .conf file separately before releasing something.

Issue #1118, opened 2015-09-25
0.16.0-beta1: Test failures on OpenSolaris 11.2 (x86)


redstar:
ldc does run on OpenSolaris (x86), too! Here is my collection of test failures. I suspect that some/most of the failures are also present in Linux/x86.

`runnable/aliasthis.d`, options `-gc -O`:

    piece covers entire variable
      call void @llvm.dbg.value(metadata %aliasthis.C1* null, i64 0, metadata !210, metadata !1724), !dbg !1807
    !210 = !DILocalVariable(tag: DW_TAG_auto_variable, name: "s3", scope: !193, file: !4, line: 290, type: !166)
    !1724 = !DIExpression(DW_OP_bit_piece, 0, 32)
    piece is larger than or outside of variable
      call void @llvm.dbg.value(metadata %aliasthis.C3* null, i64 0, metadata !210, metadata !1809), !dbg !1807
    !210 = !DILocalVariable(tag: DW_TAG_auto_variable, name: "s3", scope: !193, file: !4, line: 290, type: !166)
    !1809 = !DIExpression(DW_OP_bit_piece, 32, 32)
    piece is larger than or outside of variable
      call void @llvm.dbg.value(metadata %aliasthis.S2* null, i64 0, metadata !210, metadata !1811), !dbg !1807
    !210 = !DILocalVariable(tag: DW_TAG_auto_variable, name: "s3", scope: !193, file: !4, line: 290, type: !166)
    !1811 = !DIExpression(DW_OP_bit_piece, 64, 32)
    piece is larger than or outside of variable
      call void @llvm.dbg.value(metadata %aliasthis.S2* null, i64 0, metadata !210, metadata !1812), !dbg !1807
    !210 = !DILocalVariable(tag: DW_TAG_auto_variable, name: "s3", scope: !193, file: !4, line: 290, type: !166)
    !1812 = !DIExpression(DW_OP_bit_piece, 96, 32)
    LLVM ERROR: Broken module found, compilation aborted!

`runnable/variadic.d`: `core.exception.AssertError@runnable/variadic.d(585): Assertion failure`

`runnable/nested.d`, options `-release -gc -O`:

    piece is larger than or outside of variable
      call void @llvm.dbg.value(metadata i32 4, i64 0, metadata !689, metadata !1881), !dbg !2359
    !689 = !DILocalVariable(tag: DW_TAG_auto_variable, name: "y", scope: !687, file: !4, line: 1122, type: $
[...]

PR #1131, opened 2015-09-27
Use another Travis-CI build order.

redstar:
This PR re-arranges the build parameters a bit. Basically, I moved
every option to a higher LLVM version. I expect that this will
show some more failures. It also prepares for the drop of support
of LLVM 3.1-3.4.


kinke:
No failures. :+1:

redstar:
A bit surprising as I can reproduce the debug info failure from #1118 on my Gentoo system.

PR #1137, opened 2015-10-03
Watch out for special case when initializing ref variables

kinke:
It may be the initialization of the variable's hidden pointer, or the initialization of the pointee.

This fixes bugzilla 14944 for us (tested in `runnable/sctor.d`), see https://github.com/D-Programming-Language/dmd/pull/5013.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -517,31 +517,29 @@ public:
             return;
         }
 
+        // Initialization of ref variable?
         // Can't just override ConstructExp::toElem because not all TOKconstruct
         // operations are actually instances of ConstructExp... Long live the DMD
         // coding style!
-        if (e->op == TOKconstruct)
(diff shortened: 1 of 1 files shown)


kinke:
I've just fixed that one, see https://github.com/ldc-developers/ldc/pull/1138. Took me a while to find out the root cause.

kinke:
Oh and `ismemset & 1` isn't *that*  helpful, as it merely specifies whether the elements can be bitcopied or not. The check where you added the comment about using `ismemset` isn't affected by that, as that one checks whether it's an array-to-array or an element-wise assignment. The complexity stems from the fact that the element type isn't trivial for multi-dimensional arrays (i.e., `T[1][2][3][4]` can be filled with `T[1][2][3]`, `T[1][2]`, `T[1]` or `T` elements).

PR #1142, opened 2015-10-05
Exclude monitor pointer in initializers of C++ classes

kinke:


kinke:
Test case added.

kinke:
The offset (GEP index) was already correct when memcpying, only the initializer still contained a null-initialized monitor pointer as 2nd field between vtable pointer and first explicit field.

Issue #972, opened 2015-06-09
equivalent of dmd -conf=path for ldc and ldmd ?


timotheecour:
is there an equivalent flag?
what's the recommended way to set it ?

it would be a lot more convenient than the instructions given at http://wiki.dlang.org/Using_LDC (too many search paths for this IMO)

redstar:
`ldmd2 -conf=` is currently a dummy, implementation is pending.

kinke:
Implemented by #1143, analogous to DMD.

Issue #879, opened 2015-04-01
Reminder: -conf switch for ldmd2


mleise:
If possible blanking the configuration with this switch should work for 2.067 ldmd2. That would allow me to have the *dman* tool build with ldc2 on Gentoo. I'm trying to keep everything as compiler agnostic as possible for what it's worth.
Setting and parsing a specific configuration file is not required at this time. It seems like the primary use case for -conf= is to not specify a configuration at all.

kinke:
> It seems like the primary use case for -conf= is to not specify a configuration at all.

That's what I was thinking too. But DMD treats it as if there was no -conf switch at all, defaulting to an auto-located configuration file. In case you really need a switch to disable loading a config file, we could use some magic string like `null` for that.

mleise:
What? That changes everything. I'll leave dman out then. Thanks for the info.

PR #1143, opened 2015-10-06
Implement -conf command-line switch for LDC2 and LDMD2

kinke:
One more step towards the 0.16 release. :)

Addresses https://github.com/ldc-developers/ldc/issues/879 and https://github.com/ldc-developers/ldc/issues/972.

### DIFF
--- a/driver/configfile.cpp
+++ b/driver/configfile.cpp
@@ -11,6 +11,7 @@
 #include "driver/exe_path.h"
 #include "mars.h"
 #include "libconfig.h"
+#include "llvm/Support/CommandLine.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 #include <cassert>
@@ -26,6 +27,12 @@
 #endif
 
 namespace sys = llvm::sys;
+
+// dummy only; needs to be parsed manually earlier as the switches contained in
+// the config file are injected into the command line options fed to the parser
+llvm::cl::opt<std::string> clConf("conf",
+    llvm::cl::desc("Use configuration file <filename>"),
+    llvm::cl::value_desc("filename"));
 
 #if LDC_LLVM_VER >= 304
 #if _WIN32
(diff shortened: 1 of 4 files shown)


kinke:
This reveals an issue in the dmd-testsuite Makefile:
```
DRUNTIME_PATH=../../druntime
PHOBOS_PATH=../../phobos
```
This should most likely include another `..`, as the Makefile resides 3 dirs deep in the LDC repo. Why not get rid of the obsolete `d2` dir? `tests/d2/dmd-testsuite` => `tests/dmd-testsuite`. Or directly a top-level dir in the LDC repo...

kinke:
Ah, DMD treats an explicitly empty `-conf=` as if it wasn't specified, falling back to a default config file it finds via `findConfFile(argv[0], "sc.ini"/"dmd.conf")`.

redstar:
@kinke That's really great! Thanks for the work. (I'm still busy with my book... think that it will take 'til end of month to finish it.)

driver/main.cpp
```diff
@@ -247,6 +247,17 @@ static void hideLLVMOptions() {
 
 int main(int argc, char **argv);
 
+static const char* tryGetExplicitConfFile(int argc, char** argv)
```
redstar: Is there a reason not to use a `llvm::cl:opt<>`?

driver/configfile.cpp
```diff
@@ -27,6 +28,12 @@
 
 namespace sys = llvm::sys;
 
+// dummy only; needs to be parsed manually earlier as the switches contained in
+// the config file are injected into the command line options fed to the parser
```
kinke: @redstar: see this comment

redstar: We could call `cl::ParseCommandLineOptions()` several times: first with the `argc`, `argv` parameter from `main()`, then locate and read the config file and then with the parameters from the config file. This should not change the option parsing order but avoids a workaround for `-conf=`.

This would also help if I ever implement the new config scheme I proposed once. See here: http://forum.dlang.org/post/bexmafwrzktfprexzejp@forum.dlang.org
(This requires some more code moving...)

kinke: I was looking into this, but I didn't find any reset() function, and I was afraid parsing the command-line options twice would lead to the options being duplicated...

redstar:
I merged this because the use of  `cl::ParseCommandLineOptions()` would result in a change of the parse order: first the config options are parsed and then the `argv[]` options. Don't know yet how to avoid this.

PR #1144, opened 2015-10-08
Nested static functions and function literals do not inherit the parent context

### LINKED ISSUE #1127: [2.068] std.datetime unittest compile error
With https://github.com/ldc-developers/ldc/pull/1123, this is the only remaining compile error (LDC + runtimes + runtime unittests, excluding the std.string-unittest-debug issue #930) on Win64 (in both debug and release variants):
```
C:\LDC\ldc\runtime\phobos\std\datetime.d(31362): Error: function std.datetime.__unittestL31340_428.__funcliteral4.map!((b) => cast(char)b).map!(immutable(ubyte)[]).map is a nested function and cannot be accessed from std.datetime.__unittestL31340_428.__funcliteral4
```
I don't get why, as we have exactly the same line (and imports) in line 31097.

kinke:
Should fix #1127.

*Edit*: would probably be better suited for master; let's check the tests at least first here on merge-2.068 where the issue surfaces.

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -1789,12 +1789,7 @@ FuncDeclaration* getParentFunc(Dsymbol* sym, bool stopOnStatic)
         if (stopOnStatic)
         {
             // Fun fact: AggregateDeclarations are not Declarations.
-            if (FuncDeclaration* decl = parent->isFuncDeclaration())
-            {
-                if (decl->isStatic())
-                    return NULL;
-            }
-            else if (AggregateDeclaration* decl = parent->isAggregateDeclaration())
+            if (AggregateDeclaration* decl = parent->isAggregateDeclaration())
             {
                 if (!decl->isNested())
                     return NULL;
@@ -1803,7 +1798,11 @@ FuncDeclaration* getParentFunc(Dsymbol* sym, bool stopOnStatic)
         parent = parent->parent;
     }
 
-    return parent ? parent->isFuncDeclaration() : NULL;
+    if (!parent)
+        return NULL;
+
+    FuncDeclaration* fd = parent->isFuncDeclaration();
+    return fd;//(stopOnStatic && fd->isStatic()) ? NULL : fd;
 }
 
 LLValue* DtoIndexAggregate(LLValue* src, AggregateDeclaration* ad, VarDeclaration* vd)
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -1781,28 +1781,43 @@ FuncDeclaration* getParentFunc(Dsymbol* sym, bool stopOnStatic)
     if (!sym)
         return NULL;
 
-    Dsymbol* parent = sym->parent;
-    assert(parent);
-
-    while (parent && !parent->isFuncDeclaration())
+    // check if symbol is itself a static function/aggregate
+    if (stopOnStatic)
     {
+        // Static functions and function (not delegate) literals don't allow
+        // access to a parent context, even if they are nested.
+        if (FuncDeclaration* fd = sym->isFuncDeclaration())
+        {
+            bool certainlyNewRoot = fd->isStatic() ||
+                (fd->isFuncLiteralDeclaration() &&
+                    static_cast<FuncLiteralDeclaration*>(fd)->tok == TOKfunction);
+            if (certainlyNewRoot)
+                return NULL;
+        }
+        // Fun fact: AggregateDeclarations are not Declarations.
+        else if (AggregateDeclaration* ad = sym->isAggregateDeclaration())
+        {
+            if (!ad->isNested())
+                return NULL;
+        }
+    }
+
+    for (Dsymbol* parent = sym->parent; parent; parent = parent->parent)
+    {
+        if (FuncDeclaration* fd = parent->isFuncDeclaration())
+            return fd;
+
         if (stopOnStatic)
         {
-            // Fun fact: AggregateDeclarations are not Declarations.
-            if (AggregateDeclaration* decl = parent->isAggregateDeclaration())
+            if (AggregateDeclaration* ad = parent->isAggregateDeclaration())
             {
-                if (!decl->isNested())
+                if (!ad->isNested())
                     return NULL;
             }
         }
-        parent = parent->parent;
     }
 
-    if (!parent)
-        return NULL;
-
-    FuncDeclaration* fd = parent->isFuncDeclaration();
-    return fd;//(stopOnStatic && fd->isStatic()) ? NULL : fd;
+    return NULL;
 }
 
 LLValue* DtoIndexAggregate(LLValue* src, AggregateDeclaration* ad, VarDeclaration* vd)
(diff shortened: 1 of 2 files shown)


gen/llvmhelpers.cpp
```diff
@@ -1789,12 +1789,7 @@ FuncDeclaration* getParentFunc(Dsymbol* sym, bool stopOnStatic)
         if (stopOnStatic)
         {
             // Fun fact: AggregateDeclarations are not Declarations.
-            if (FuncDeclaration* decl = parent->isFuncDeclaration())
```
kinke: This is never true, see the while expression a couple of lines above.

gen/llvmhelpers.cpp
```diff
@@ -1803,7 +1798,11 @@ FuncDeclaration* getParentFunc(Dsymbol* sym, bool stopOnStatic)
         parent = parent->parent;
     }
 
-    return parent ? parent->isFuncDeclaration() : NULL;
+    if (!parent)
+        return NULL;
+
+    FuncDeclaration* fd = parent->isFuncDeclaration();
+    return fd;//(stopOnStatic && fd->isStatic()) ? NULL : fd;
```
kinke: Here would be the proper place to check whether the parent function is static and return NULL in that case if `stopOnStatic` is set.
That triggers an LLVM assertion (`Can't get a pointer to <null> type!`) in `DtoNestedVariable()` though. Needs to be further investigated. :/

kinke:
I've overlooked the fact that `getParentFunc()` takes an arbitrary symbol. The fix is to check whether the symbol itself is a static function/function literal/static aggregate first, and return null in that case if `stopOnStatic` is set. If the symbol is enclosed by a static aggregate before coming across the innermost parent function, also return null for `stopOnStatic`, but don't care about the static-ness of a potential parent function.

kinke:
merge-2.068 is really in excellent shape already. Based on https://travis-ci.org/ldc-developers/ldc/jobs/84218182: 4 minor issues left, ignoring 4 dmd-testsuite tests complaining about not finding `gdb`:
* unittest `std-variant` fails in both debug and release
* compilable/test9057.d
* fail_compilation/fail39.d
* fail_compilation/pragmainline2.d

kinke:
The undefined references (debug only, most likely due to inlining in release) are:
```
_D3std5range18__T4iotaTxdTxdTxdZ4iotaFxdxdxdZ6Result6__initZ
_D3std5range14__T6RepeatTAiZ6Repeat11__xopEqualsFKxS3std5range14__T6RepeatTAiZ6RepeatKxS3std5range14__T6RepeatTAiZ6RepeatZb
_D3std8typecons12__T5TupleTeZ5Tuple8__xopCmpFKxS3std8typecons12__T5TupleTeZ5TupleKxS3std8typecons12__T5TupleTeZ5TupleZi
_D3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImpl11__xopEqualsFKxS3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImplKxS3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImplZb
_D3std5range14__T6RepeatTAiZ6Repeat9__xtoHashFNbNeKxS3std5range14__T6RepeatTAiZ6RepeatZm
_D3std5range13__T5CycleTAkZ5Cycle9__xtoHashFNbNeKxS3std5range13__T5CycleTAkZ5CycleZm
_D3std5range18__T3ZipTAAyaTAAyaZ3Zip9__xtoHashFNbNeKxS3std5range18__T3ZipTAAyaTAAyaZ3ZipZm
_D3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImpl9__xtoHashFNbNeKxS3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImplZm
_D3std9container5slist12__T5SListTfZ5SList4Node11__xopEqualsFKxS3std9container5slist12__T5SListTfZ5SList4NodeKxS3std9container5slist12__T5SListTfZ5SList4NodeZb
_D3std5range18__T3ZipTAAyaTAAyaZ3Zip11__xopEqualsFKxS3std5range18__T3ZipTAAyaTAAyaZ3ZipKxS3std5range18__T3ZipTAAyaTAAyaZ3ZipZb
_D3std5range13__T5CycleTAkZ5Cycle11__xopEqualsFKxS3std5range13__T5CycleTAkZ5CycleKxS3std5range13__T5CycleTAkZ5CycleZb
_D3std8typecons12__T5TupleTeZ5Tuple11__xopEqualsFKxS3std8typecons12__T5TupleTeZ5TupleKxS3std8typecons12__T5TupleTeZ5TupleZb
_D3std9container5slist12__T5SListTeZ5SList4Node6__initZ
_D3std8typecons118__T10RefCountedTS3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImplVE3std8typecons24RefCountedAutoInitializei0Z10RefCounted10__postblitMFNaNbNiNfZv
_D3std5range18__T4iotaTxdTxdTxdZ4iotaFxdxdxdZ6Result11__xopEqualsFKxS3std5range18__T4iotaTxdTxdTxdZ4iotaFxdxdxdZ6ResultKxS3std5range18__T4iotaTxdTxdTxdZ4iotaFxdxdxdZ6ResultZb
_D3std8typecons12__T5TupleTeZ5Tuple6__initZ
_D3std9container5slist12__T5SListTeZ5SList4Node9__xtoHashFNbNeKxS3std9container5slist12__T5SListTeZ5SList4NodeZm
_D3std9container5slist12__T5SListTfZ5SList4Node6__initZ
_D3std9container5slist12__T5SListTfZ5SList4Node9__xtoHashFNbNeKxS3std9container5slist12__T5SListTfZ5SList4NodeZm
_D3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImpl15__fieldPostblitMFNbNeZv
_D3std8typecons118__T10RefCountedTS3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImplVE3std8typecons24RefCountedAutoInitializei0Z10RefCounted6__dtorMFZv
_D3std8typecons12__T5TupleTeZ5Tuple6toHashMxFNbNeZm
_D3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImpl11__fieldDtorMFNeZv
_D3std5range18__T4iotaTxdTxdTxdZ4iotaFxdxdxdZ6Result9__xtoHashFNbNeKxS3std5range18__T4iotaTxdTxdTxdZ4iotaFxdxdxdZ6ResultZm
_D3std5range18__T3ZipTAAyaTAAyaZ3Zip6__initZ
_D3std5stdio4File24__T14ByLineCopyImplTaTaZ14ByLineCopyImpl6__initZ
_D3std9container5slist12__T5SListTeZ5SList4Node11__xopEqualsFKxS3std9container5slist12__T5SListTeZ5SList4NodeKxS3std9container5slist12__T5SListTeZ5SList4NodeZb
```
Some initializers and some special  __xopCmp, __xopEquals,__xtoHash, __postblit, __fieldPostblit, __fieldDtor, __dtor and toHash functions.

PR #1145, opened 2015-10-08
Backport getParentFunc() fix from merge-2.068

kinke:


kinke:
My intention was to just test it on merge-2.068 and then commit it on master. Since it's been merged already, do we want to wait or backport it?
Btw, given how promising merge-2.068 already appears, will LDC 0.16 be based on D2.067 or D2.068? Or is there going to be a 0.17 release 2 weeks later? ;)

kinke:
I'll just go ahead and merge it, no reason not to have it fixed in 0.16 already.

kinke:
Yeah, makes sense. master may still benefit from a couple of fixes for issues surfacing in merge-2.068, but that's not a good reason to hold 0.16 back any longer - 0.17 is already around the corner, so bug reports for 0.16 *could* ;) make it directly into 0.17 finalization phase.

redstar:
I create 0.16.0-beta2 now. Maybe I also create next week the first 0.17.0-alpha release.

Issue #1156, opened 2015-10-11
MSVC: extern(C++) ABI issue


kinke:
An extended snippet from `runnable/cppa.d`:

cppb.cpp:
```C++
#include <cstdio>

struct foo { int i; int j; void* p; };

class bar
{
public:
  virtual foo getFoo(foo& f){
    printf("getFoo: this = %p, f = %p\n", this, &f);
    f.i = 1;
    f.j = 2;
    f.p = (void*)this;
    return f;
  }
};

bar* newBar()
{
  bar* b = new bar();
  printf("newBar: %p\n", b);
  return b;
}
```
cppa.d:
```D
import core.stdc.stdio;

extern(C++)
{
  struct foo { int i; int j; void* p; }
  interface bar { foo getFoo(ref foo); }
  bar newBar();
}

void main()
{
  bar b = newBar();
  foo f;
  printf("b = %p, f = %p\n", cast(void*)b, &f);
  foo r = b.getFoo(f);
  printf("f: %d, %d, %p\n", f.i, f.j, f.p);
  printf("r: %d, %d, %p\n", r.i, r.j, r.p);
}
```
Example output:
```
newBar: 00000026DB8C3410
b = 00000026DB8C3410, f = 00000026DB85F9C0
getFoo: this = 00000026DB85F9B0, f = 00000026DB85F9C0
f: 1, 2, 00000026DB85F9B0
r: -611976512, 38, 000007FAAF6617B2
```
So the return value `r` is way off, and `this` in `getFoo()` is wrong as well. I highly suspect that the order of the `sret` and `this` pointer arguments needs to be reversed (this before sret).

kinke:
I'm working on it.

PR #1157, opened 2015-10-12
Win64 ABI: pass 'this' before 'sret' for extern (C++)

kinke:
Fixes https://github.com/ldc-developers/ldc/issues/1156.

### DIFF
--- a/gen/abi-win64.cpp
+++ b/gen/abi-win64.cpp
@@ -40,6 +40,8 @@ struct Win64TargetABI : TargetABI
 
     bool passByVal(Type* t);
 
+    bool passThisBeforeSret(TypeFunction* tf);
+
     void rewriteFunctionType(TypeFunction* tf, IrFuncTy &fty);
 
     void rewriteArgument(IrFuncTy& fty, IrFuncTyArg& arg);
@@ -115,6 +117,11 @@ bool Win64TargetABI::returnInArg(TypeFunction* tf)
 bool Win64TargetABI::passByVal(Type* t)
 {
     return false;
+}
+
+bool Win64TargetABI::passThisBeforeSret(TypeFunction* tf)
+{
+    return tf->linkage == LINKcpp;
 }
 
 void Win64TargetABI::rewriteFunctionType(TypeFunction* tf, IrFuncTy &fty)
--- a/gen/abi.h
+++ b/gen/abi.h
@@ -106,6 +106,9 @@ struct TargetABI
     /// Returns true if the type is passed by value
     virtual bool passByVal(Type* t) = 0;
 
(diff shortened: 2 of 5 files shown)


kinke:
This gets `runnable/cppa.d` to work on Win64 (at least with `merge-2.068`) - except for a minor linking issue (public static extern(C++) function defined in both .cpp and .d source files) and an apparently wrong test for MSVC-based builds, verifying an apparent DMD ABI bug. ;)

kinke:
Not sure what to do about `extern(D)`. Currently, I know of these diffs compared to the Win64 C calling convention:

* `extern (D)` uses a reversed arguments order because DMD does and some runtime functions depend on this
* we pass slices (dynamic arrays) as LL aggregate (i.e., in 2 registers) and not as normal 16-bytes struct (byref)
* we pass sret before this (with this PR, only for `extern (D)`)

Needless to say, the D docs mention something about the D calling convention basically matching C's on that system.

yebblies:
> `extern (D)` uses a reversed arguments order because DMD does and some runtime functions depend on this

IMO we (all) should match the C calling convention argument order.  If we really need to keep the same argument for array ops we can just switch the parameter order around.

> we pass slices (dynamic arrays) as LL aggregate (i.e., in 2 registers) and not as normal 16-bytes struct (byref)

They should be passed the same as the equivalent struct, I think DMD does this...

> we pass sret before this (with this PR, only for extern (D))

I don't know about this.  The only thing I can think of is that it would be nice to set the parameter order so that a member function can be called via a function pointer with `this` passed as the last argument.  I guess putting sret before this means this will always be last, so it works?

kinke:
> IMO we (all) should match the C calling convention argument order.

We've been saying this for years, right? ;) I tried it once, there's still a PR for LDC, druntime and Phobos, but then interesting bugs appeared, e.g., sorting a range would sort it in descending order. So there are definitely some runtime functions relying on Walter's favorite arguments order. ;)

> They should be passed the same as the equivalent struct, I think DMD does this...

@JohanEngelen had a go at this, but it's quite tedious as we'd need to apply the TargetABI rewrites to all runtime function calls as well. Well maybe it wouldn't be *that* bad, but I'm not keen on doing this in the near future. ;)

> I guess putting sret before this means this will always be last, so it works?

Well right now, on all non-Win64 platforms, LDC passes the optional `sret` pointer, then the optional `this` or context pointer, then the optional `_arguments` array and only then the explicit args in either reversed (non-variadic `extern (D)` functions only) or regular order. So `this` is either the first or 2nd argument (and only ever the last one if the function takes no parameters at all). For Visual C++, it's always the first arg.

PR #1159, opened 2015-10-16
Introduce DtoAlignment() and overload DtoAlloca() for VarDeclarations

kinke:
Trying to get the alignment right by using the first non-default one in the following order of descending priority:

`VarDeclaration::alignment` [variables only of course]
`Type::alignment()`
`Type::alignsize()`

This fixes the alignment of `align(x) struct S { ... }` instances on the stack.
Local variables with an explicit align attribute are still default-aligned though. :/

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -876,7 +876,7 @@ void DtoDefineFunction(FuncDeclaration* fd)
                 // Create the param here and set it to a "dummy" alloca that
                 // we do not store to here.
                 irparam = getIrParameter(vd, true);
-                irparam->value = DtoAlloca(vd->type, vd->ident->toChars());
+                irparam->value = DtoAlloca(vd, vd->ident->toChars());
             }
             else
             {
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -136,12 +136,35 @@ void DtoDeleteArray(Loc& loc, DValue* arr)
 
 /****************************************************************************************/
 /*////////////////////////////////////////////////////////////////////////////////////////
+// ALIGNMENT HELPERS
+////////////////////////////////////////////////////////////////////////////////////////*/
+
+unsigned DtoAlignment(Type* type)
+{
+    structalign_t alignment = type->alignment();
+    if (alignment == STRUCTALIGN_DEFAULT)
+        alignment = type->alignsize();
+    return (alignment == STRUCTALIGN_DEFAULT ? 0 : alignment);
+}
+
+unsigned DtoAlignment(VarDeclaration* vd)
+{
+    return vd->alignment == STRUCTALIGN_DEFAULT ? DtoAlignment(vd->type) : vd->alignment;
+}
+
+/****************************************************************************************/
+/*////////////////////////////////////////////////////////////////////////////////////////
 // ALLOCA HELPERS
 ////////////////////////////////////////////////////////////////////////////////////////*/
 
 llvm::AllocaInst* DtoAlloca(Type* type, const char* name)
 {
-    return DtoRawAlloca(DtoMemType(type), type->alignsize(), name);
+    return DtoRawAlloca(DtoMemType(type), DtoAlignment(type), name);
+}
+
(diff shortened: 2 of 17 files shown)


kinke:
An error in `runnable/ldc_github_726.d` seems to be caused by a wrongly determined size for a packed struct:
```D
struct Buggy
{
    align(1):
    uint a = 0x0a0a0a0a;
    ulong b = 0x0b0b0b0b0b0b0b0b;
}
```
The offset is computed correctly in
```D
Buggy packed;
ulong raw = *cast(ulong*)(cast(ubyte*)&packed + packed.b.offsetof);
```
and the LL type ```type { i32, i64, [4 x i8] }``` is the same as upstream. I think it's a bug that the front-end determines a size of 16 bytes; the max member alignment is 4, so I think the struct's should be 4 as well, leading to a size of 12 bytes. Apparently setting the struct alignment to 8, just because a member's alignment would be 8 if it wasn't packed, but is actually 4, seems wrong.

Anyway, with this PR, the size used when initializing `packed` by memcpying the (correct) initializer is **24** bytes.

Oh, I've just noted that the upstream IR uses a slightly different type: `type <{ i32, i64, [4 x i8] }>`, i.e., it encloses it in `<>`. Let me guess - it means packed, right? ;)

kinke:
To verify the PR, it's useful to inspect the generated .ll for:
```D
align(16) struct Struct { int[2] a; }
struct S { align(16) int a; }

void main()
{
    align(32) int[2] x;
    align(32) Struct s;
    S s2;
    //void foo() { x[0] = 666; s.a[0] = 123; }
    //foo();
}
```
=>
```LL
%x = alloca [2 x i32], align 4                  ; [#uses = 1, size/byte = 8]
%s = alloca %align.Struct, align 16             ; [#uses = 2, size/byte = 16]
%s2 = alloca %align.S, align 16                 ; [#uses = 1, size/byte = 16]
```
When uncommenting `foo()` and its invokation, `x` and `s` are captured and so placed in a nested context IR struct. LDC now adds an 8-bytes padding inbetween to make sure `s` is 16-bytes aligned, aligns the context struct itself at a 16-bytes boundary and uses the correct GEP index for the nested variable:
```LL
%.frame = alloca %nest.main, align 16           ; [#uses = 3, size/byte = 32]
%s2 = alloca %align.S, align 16                 ; [#uses = 1, size/byte = 16]
%x = getelementptr %nest.main, %nest.main* %.frame, i32 0, i32 0 ; [#uses = 1, type = [2 x i32]*]
%s = getelementptr %nest.main, %nest.main* %.frame, i32 0, i32 2 ; [#uses = 1, type = %align.Struct*]
... in foo():
%s = getelementptr %nest.main, %nest.main* %5, i32 0, i32 2 ; [#uses = 1, type = %align.Struct*]
```
What's still not working is the explicit alignment for local variables.

kinke:
https://github.com/ldc-developers/phobos/blob/ldc/std/conv.d#L5105 fails due to an insufficiently aligned (1-bytes on IR level, 4 bytes on the Travis boxes) static byte[] array when trying to emplace a class instance in it (8-bytes aligned). I need to fix the alignment of these globals.

kinke:
Wrt. the locals, I assumed their `VarDeclaration::alignment` would be set by the front-end if there was an `align(x)` attribute, but apparently that's not the case.

MartinNowak:
No tests?

I just tried to translate [_mm_loadu_ps](http://clang.llvm.org/doxygen/xmmintrin_8h.html#afb07ee3c546b8d6c59ecd195ec4b3643) to D, but the compiler keeps using an aligned load.

```d
import core.simd, ldc.simd;

align(1) struct Float4
{
    align(1) float4 v;
}

float4 test(float* ptr)
{
    return (cast(Float4*)ptr).v;
}
```

JohanEngelen:
@MartinNowak Tests are here: https://github.com/ldc-developers/ldc/blob/master/tests/codegen/align.d
But your case is missing.

(there are some "C<space>HECK" things in that testfile (disabled tests), about which there should be a D language discussion)

JohanEngelen:
@MartinNowak Could you report a new issue for this?

Issue #1161, opened 2015-10-17
Win64: no error status when creating a library fails


rainers:
Here's a repro:
```
C:\tmp\d\ldc>ldc2 -lib test.d
C:\tmp\d\ldc>attrib +R test.lib
C:\tmp\d\ldc>ldc2 -lib test.d
LINK : fatal error LNK1104: cannot open file 'test.lib'
Error: C:\l\vs14\VC\BIN\lib.exe failed with status: 1104

C:\tmp\d\ldc>echo %ERRORLEVEL%
0
```
ldc sould return an error here, probably the status of the failing lib call.

klickverbot:
Fixed by #1162, thanks!

(By the way, if you include something to the effect of `closes #<issue_id>` or `fixes #<issue_id>` in the commit message, the PR will be linked with the issue automatically.)

PR #1162, opened 2015-10-17
issue #1161: return status if lib invocation failed

rainers:
Add missing forward of execution result.


kinke:
Thanks, I stumbled upon this too today.

Issue #1160, opened 2015-10-17
Win64: link fails with lots of object files


rainers:
Building druntime with a Visual D project puts all modules on the commandline (through a response file), but linking the unittest executable or creating the library fails with
```
The command line is too long.
Error: C:/s/d/ldc/build-ldc2-x64/bin/Debug/x86.bat failed with status: 1
```

This happens because the list of arguments is longer than the limit for batch files (8192 characters). LDC should switch to a response file in this case, too.

kinke:
Thanks for the report. I've noted that you seem to be cross-compiling to 32-bit code; please note that this is not officially supported yet. There are probably some MSVC-specific branches which are guarded by Win64 checks instead of checking for MSVC; probably not just in LDC, but in druntime and Phobos too...

rainers:
> I've noted that you seem to be cross-compiling to 32-bit code; please note that this is not officially supported yet.

I'm actually trying to get LDC working for Win32, too. I succeeded to some extent with LLVM 3.7, until I noticed that clang also failed to access caught objects correctly. That's why I switched to LLVM 3.8 and now have C++ compatible exception handling, but it still fails in complex situations.

klickverbot:
Marking as closed as per your comment in @kinke's PR.

kinke:
> I'm actually trying to get LDC working for Win32, too.

Oh, cool - all the best! :+1: 

> I succeeded to some extent with LLVM 3.7, until I noticed that clang also failed to access caught objects correctly. That's why I switched to LLVM 3.8 and now have C++ compatible exception handling, but it still fails in complex situations.

Are you already using [LLVM's new EH instructions](http://llvm.org/docs/ExceptionHandling.html#new-exception-handling-instructions)?

rainers:
> Are you already using LLVM's new EH instructions?

Yes. I started with the beginCatch/endCatch intrinsics of LLVM 3.7, but had to switch to the new instructions for 3.8.

kinke:
Interesting. I suggest you open a work-in-progress PR soon, so that others see that there's movement on that front - Manu was here ~2 weeks ago and wanted to hack on LLVM too, so he may be working on the same stuff. And Kai is interested in getting a working MSVC Win32 build too once he's finished his book...

rainers:
> I suggest you open a work-in-progress PR soon, so that others see that there's movement on that front (and may chime in)

[...]

PR #1163, opened 2015-10-17
issue #1160: windows: pack linker/lib arguments into a response file if too long

rainers:
Not sure if this copy and paste is ok, but the needed calls are not accessible in llvm/lib/Support/Program.cpp.

https://github.com/ldc-developers/ldc/issues/1160

### DIFF
--- a/driver/linker.cpp
+++ b/driver/linker.cpp
@@ -22,9 +22,7 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Program.h"
 #include "llvm/Support/Path.h"
-#if _WIN32
 #include "llvm/Support/SystemUtils.h"
-#endif
 
 //////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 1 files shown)


kinke:
Thanks Rainer, I'll incorporate the response file in an upcoming PR.

kinke:
Incorporated by https://github.com/ldc-developers/ldc/pull/1165, commit https://github.com/kinke/ldc/commit/4514d7e342b10ccf36dfff08bc13f6c28f1e12b9. I've only changed from `createUniqueFile()` to `createTemporaryFile()`, as there's the possibility that the current working directory isn't writable. Visual Studio uses the user's temp folder as well iirc.

rainers:
> I've only moved from createUniqueFile() to createTemporaryFile()

Yeah, that's better. I'm not too familiar with llvm code, so just grepped for similar code and only found createUniqueFile.

PR #1164, opened 2015-10-17
fix build against llvm master (llvm 3.8)

rainers:
My attempts to build LDC against current LLVM master.

Not sure if there is a replacement for initializeIPA.


redstar:
Thanks for the help!

rainers:
Thanks. I updated again today and noticed another change: https://github.com/ldc-developers/ldc/pull/1170

PR #1165, opened 2015-10-17
Windows: fix command line when invoking linker batch file

kinke:
This command line sent to CreateProcess():

```
"C:\LDC\bin\amd64.bat" 1 2 "3" 4
```

is internally transformed to:

```
C:\Windows\system32\cmd.exe /c "C:\LDC\bin\amd64.bat" 1 2 "3" 4
```

Now guess what - cmd.exe treats the command string after /c or /k in a very special way if it begins with double quotes and is followed by other quoted args. In this example, it tries to invoke:

```
C:\LDC\bin\amd64.bat" 1 2 "3
```

Another example: `C:\Windows\system32\cmd.exe /c "C:\L D C\bin\amd64.bat" 1 2 "3" 4` => `C:\L`

The fix seems to be enclosing the whole command with additional double quotes (but no additional escaping of the command!) and using `/s` for cmd.exe:

```
C:\Windows\system32\cmd.exe /s /c ""C:\L D C\bin\amd64.bat" 1 2 "3" 4"
```

So the command is quoted, but not escaped, i.e., no regular argument to cmd.exe, so we can't use LLVM's executeAndWait() to launch the cmd.exe process. :/

Additionally, LLVM's arg-quoting code is not public, so we can't use it to quote & escape the regular args inside the command.

I therefore implemented a simple `quoteArgs()` variant which should suffice for our use cases.

### DIFF
--- a/driver/linker.cpp
+++ b/driver/linker.cpp
@@ -375,11 +375,29 @@ int executeMsvcToolAndWait(const std::string& tool, const std::vector<std::strin
         commandLine.append(windows::quoteArg(toolPath));
     }
 
+    const size_t commandLineLengthAfterTool = commandLine.size();
+
     // append (quoted) args
     for (size_t i = 0; i < args.size(); ++i)
     {
         commandLine.push_back(' ');
         commandLine.append(windows::quoteArg(args[i]));
+    }
+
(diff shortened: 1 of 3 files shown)


rainers:
LG, fixes both issues 1160 and 1152 for me.

PR #1172, opened 2015-10-27
Implement @ldc.attributes.section(...) for variables

klickverbot:
<del>I'm still not sure whether we really want to go the `@attribute` route, or if we should prefer individual types (e.g. `@section(...)`).

The latter seems cleaner and more flexible (regarding aliasing different UDAs for different compilers). It might make sense to have string arguments for an UDA that just sets the corresponding LLVM function attribute, but using them for this seems somewhat off. GDC uses a single magic attribute struct, but that's probably just to map directly to the C `__attribute__` syntax.</del>

This now uses actual struct literals for attributes that can be generated by arbitrary CTFE expressions. This should allow for quite a bit of flexibility when targeting multiple compilers.

### DIFF
--- a/gen/declarations.cpp
+++ b/gen/declarations.cpp
@@ -21,6 +21,7 @@
 #include "gen/llvmhelpers.h"
 #include "gen/logger.h"
 #include "gen/tollvm.h"
+#include "gen/uda.h"
 #include "ir/irtype.h"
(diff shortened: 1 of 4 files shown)


redstar:
Looks good to me. I like the typed approach. :-)

PR #1179, opened 2015-10-31
driver: Remove gratuitous differences to DMD main()

klickverbot:
This actually adds a "fatal() point" after running deferred
semantic(), but apparently its absence didn't cause any issues
so far.


kinke:
Are you going to rebase this at some point? Getting rid of these superfluous casts etc. is nice...

PR #1180, opened 2015-11-01
Make use of C++11 range-based for

kinke:
Should be available now that support for LLVM < 3.5 has been dropped.

**Warning**: Another bigger patch I haven't double-checked. ;)

### DIFF
--- a/utils/gen_gccbuiltins.cpp
+++ b/utils/gen_gccbuiltins.cpp
@@ -183,17 +183,8 @@ bool emit(raw_ostream& os, RecordKeeper& records)
     map<string, Record*> defs = records.getDefs();
 #endif
 
-    for(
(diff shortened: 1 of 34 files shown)


utils/gen_gccbuiltins.cpp
```diff
@@ -183,17 +183,8 @@ bool emit(raw_ostream& os, RecordKeeper& records)
...
-#if LDC_LLVM_VER >= 306
-        auto it = defs.cbegin();
-#else
-        map<string, Record* >::iterator it = defs.begin();
-#endif
-        it != defs.end();
-        it++)
-    {
-        processRecord(os, *(*it).second, arch);
-    }
```
kinke: @klickverbot: I'd be deeply impressed if the clang tools were able to get rid of such constructs... ;)

PR #1182, opened 2015-11-01
driver: Add -static option for creating fully static binaries

klickverbot:


kinke:
Looks fine to me, and this option may really be appropriate to select the static/dynamic MSVCRT libs in the foreseeable future. So if you tested it locally, feel free to merge. :)

PR #1185, opened 2015-11-01
driver/gen/ir: clang-format the world

klickverbot:
This uses the LLVM style, which makes sense for sharing code
with other LLVM projects. The DMD code we use will soon all
be in D anyway.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,1 @@
+BasedOnStyle: LLVM
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -24,218 +24,177 @@ namespace opts {
  */
 struct CoverageParser : public cl::parser<unsigned char> {
 #if LDC_LLVM_VER >= 307
-    CoverageParser(cl::Option &O) : cl::parser<unsigned char>(O) {}
+  CoverageParser(cl::Option &O) : cl::parser<unsigned char>(O) {}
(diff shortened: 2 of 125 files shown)


redstar:
I'm ok with reformatting the source. I found the inconsistent style in the source always confusing.

kinke:
Puh... ;)
The first impression is (surprisingly) pretty good, although I'm really no friend of `T *` and `T &`, and very much prefer `T*` instead. Other than that, I guess 'll have to get used to a mini-indentation of 2 spaces, and the compact curly braces syntax (opening `{` on same line).

redstar:
@kinke It's not so tought. I did some LLVM development and this style is pretty ok.

redstar:
Is there an example for an Git hook using clang-format available? (Windows,Linux,OS X?) This could help with further development?

kinke:
I'd also ask @smolt and @joakim-noah before merging this, as rebasing their presumably big ARM branches could easily turn out to be a nightmare.

kinke:
They can reformat their current head, but not simply rebase their patches as nearly every source line would be formatted differently, or am I missing something?

kinke:
Yeah, the approach of pre-comitting the re-formatted patches before merging the re-formatted master makes sense. I've never done it this way, but I'm also no git expert. ;)

PR #1186, opened 2015-11-01
Use clang-tidy for modernization and cleanups

klickverbot:
Also contains a few manual fixes.


kinke:
Oh yeah, loving this one! :+1:

PR #1193, opened 2015-11-05
Fix DtoConstFP()

kinke:
1. For 32-bit float targets: don't convert host `longdouble` to double and then to float - do the conversion in a single step to possibly improve precision.

2. For host `longdouble` > 64 bits: represent value as llvm::APFloat, and let LLVM convert it to the target type.
The previous code assumed an 80-bit x87 host `longdouble` source for `real` targets, and copied the 80 bits (zero-extended) into the target `real`.
This has got to be wrong for 128-bit PowerPC doubledoubles (apparently an exact sum of 2 separate doubles) as target and/or source.
So this should fix x86 -> PowerPC, PowerPC -> PowerPC and PowerPC -> x86 cross compilations wrt. `reals`.

### DIFF
--- a/gen/tollvm.cpp
+++ b/gen/tollvm.cpp
@@ -410,28 +410,39 @@ LLConstant *DtoConstFP(Type *t, longdouble value) {
   LLType *llty = DtoType(t);
   assert(llty->isFloatingPointTy());
 
-  if (llty == LLType::getFloatTy(gIR->context()) ||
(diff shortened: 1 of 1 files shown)


kinke:
@redstar: This should be primarily of interest to you, I accidentally stumbled upon this. I'm still unsure whether endian-ness matters here.

kinke:
That will render this absolutely obsolete.

Issue #1192, opened 2015-11-04
Testsuite with LLVM IR codegen tests


JohanEngelen:
Would it make sense to add tests for LLVM IR codegen?
I recently set up such a testsuite for SDC, so it will be very easy for me to do so too for LDC.
Using LLVM's tools (lit and FileCheck), we can do cool things like easily check name mangling, LLVM attributes. The test files are very readable too (I find, perhaps it takes a few minutes to get familiar with FileCheck's syntax).
(also, I have already figured out Travis automation)

What do you think?
And, if you think: yes! Where should I put it?

kinke:
I like the idea. It'd have allowed me to add test cases for the latest alignment fixes, which aren't easy to test in a robust fashion at runtime. I'd suggest `tests/ir` as appropriate location.

klickverbot:
Yep, seems like it would definitely be a good idea. Things like checking debug info (though I'm not sure how well FileCheck handles this), verifying that constants are actually emitted as read-only data, etc. could also immediately benefit from this. Regarding the location, just pick a somewhat relevant name and create a directory in `tests/`.

JohanEngelen:
Great. Working on it, almost done!
I'm trying to make it easy to run the tests individually, and enable out-of-tree building (this is how I work).

klickverbot:
Do we even support in-tree builds? ;)

JohanEngelen:
See PR #1194 
One nasty issue left about LLVM's FileCheck  :(

klickverbot:
Closing this, as the PR is coming along nicely.

PR #1194, opened 2015-11-05
[WIP] Add IR test suite (using LLVM tools: lit and FileCheck)

JohanEngelen:
PR for issue #1192 
Work in progress, created PR so that I can see what Travis makes of it

I got things to work on my end, but for one issue: FileCheck is not installed by LLVM... So one has to manually copy FileCheck tool to LLVM's installed bin folder.  (FileCheck *does* come standard with MacPorts's version of LLVM). Perhaps someone can help me with this? How do we get FileCheck on other devs machine? (I have to check if Travis's LLVM install has it)


JohanEngelen:
Hmm found this... LLVM >= 3.7  has option to also install FileCheck:
https://codereview.chromium.org/1019043002/

JohanEngelen:
I was thinking about downloading FileCheck in the Travis script and compiling it there :), it's just one cpp file, but with LLVM dependencies.

JohanEngelen:
I think having FileCheck.cpp in our repo is not a good idea, because compilation success may depend on a certain version of LLVM (changes to functionnames for example in its LLVM dependencies). I don't think we care in particular about which FileCheck version we use (we could say: we pick LLVM 3.5's version as minimum requirement), so a developer can locally build it and copy it into LLVM's bin dir, or in the tests/ir dir (the lit configuration will search both locations). If you build LLVM yourself, from LLVM 3.7, setting LLVM_INSTALL_UTILS in cmake will install FileCheck automatically.

JohanEngelen:
@klickverbot @kinke I got the test suite working locally (Mac) and on Travis. For AppVeyor, I am not sure: I see that the script is downloading a pre-built LLVM package. Could we add FileCheck to the bin directory in that package?

I can add a small script that downloads and compiles FileCheck, or what do you think is best to do for developers locally?

kinke:
If it's only a matter of adding `-DLLVM_INSTALL_UTILS=ON` to the LLVM CMake command line for LLVM 3.7+, that's good enough for my local environment, I won't use earlier LLVM versions anymore.

I'm currently rebuilding LLVM (current release_37 branch) and will then update the pre-built LLVM used by AppVeyor. The FileCheck tool will then be in its bin directory as you suggested.

kinke:
But they'll probably already have installed the LLVM binary packages (as well as the required dev ones obviously), which should include FileCheck since 3.7, right?

JohanEngelen:
How about adding a small script that can be used (or slightly modified) to build FileCheck?

JohanEngelen:
But then how do we solve the incompatibilities between FileCheck.cpp from a different LLVM version than the one installed? I just tested compiling FileCheck.cpp from LLVM3.5 with LLVM3.6:  they are not compatible (function signature changes).

JohanEngelen:
OK, and then for LLVM trunk, do nothing?

JohanEngelen:
@klickverbot Something like in this last commit?

JohanEngelen:
OK, I've cleanup the changes and rebased.
For LLVM 3.5, 3.6, and 3.7, it builds FileCheck. To support future LLVM versions, we just have to copy the new FileCheck version to `/utils` and name it according to its LLVM verison.

@klickverbot @kinke  Can you try it with your local setup?  `ctest -R llvm` should only run the IR tests.

What I don't like is that the "llvm-ir-testsuite" is now run as part of the stdlib tests, so it does not stand out very much. Could be fixed at a later point if it proves to be annoying.

kinke:
Took me some steps to get it right:
* Install `pip` for my PortablePython 2.7.5.1 installation
  * Download `get-pip.py` from https://pip.pypa.io/en/latest/installing/
  * Install by invoking `python <path to get-pip.py>`
* Install `lit`: `python -m pip install lit`
  I got an error `invalid command 'bdist_wheel'`, but it still seems to be working.
* Extend the command in `tests/ir/CMakeLists.txt` by an explicit `python ` prefix (I haven't set any .py file associations)

Now:
```
577: Test command: "C:\LDC\Tools\Portable Python 2.7.5.1\App\python.exe" "./runlit.py" "-v" "."
577: Test timeout computed to be: 9.99988e+006
577: -- Testing: 1 tests, 1 threads --
577: PASS: LLVM IR codegen :: asm_output.d (1 of 1)
577: Testing Time: 0.04s
577:   Expected Passes    : 1
1/1 Test #577: llvm_ir_testsuite ................   Passed    0.11 sec
```
:+1:

kinke:
Oh and AppVeyor doesn't run it; it excludes all `.*testsuite.*` tests in lines 139-140. So please either change that to exclude to `dmd-testsuite` or run `llvm_ir_testsuite` as separate command if you want it to stand out (I'd prefer that).

kinke:
> Regarding the executable mapping problem, perhaps there is a FindPython CMake module or similar?

We may just keep things simple and use a `python ` prefix on Windows. I guess it just works on all other platforms. ;)

JohanEngelen:
Further changes:
- add error message to runlit.py to help solve problems
- add python to the test command (it worked on my machine because of the #! at start of the file)
- run these tests before stdlib tests (my own priority ;-), and separately.

(I wonder how these lit command lines work on Windows: `// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s --check-prefix LLVM < %t.ll`   We should be careful not to make them too fancy, I think.

PR #1196, opened 2015-11-07
druntime: Make use of atomic read-modify-write LLVM intrinsics

kinke:


kinke:
According to the Travis LLVM 3.7 multilib job, it's working on Linux x86 and x86_64, and according to AppVeyor, on Win64 too. The OSX release unit tests pass; 2 debug ones fail with `Fatal error in EH code: _Unwind_RaiseException failed with reason code: 5` though, but I guess that's unrelated.

kinke:
See https://github.com/ldc-developers/druntime/pull/38 about my concerns wrt. the stored value potentially diverging from the returned value due to potentially different precisions for the operation.

kinke:
```D
import core.atomic, core.stdc.stdio;

void main() {
  shared ubyte x = 3;
  ubyte r;

  r = atomicOp!"+="(x, 257);
  assert(x == r);
  printf("%d\n", r);

  r = atomicOp!"+="(x, -263);
  assert(x == r);
  printf("%d\n", r);

  r = atomicOp!"-="(x, 257);
  assert(x == r);
  printf("%d\n", r);

  r = atomicOp!"-="(x, -263);
  assert(x == r);
  printf("%d\n", r);
}
```
=>
```
4
253
252
3
```
for both LDC with this PR and DMD 2.068.0.

kinke:
Now it's the new llvm_ir_testsuite failing because it cannot find lit. Oh well...

JohanEngelen:
I don't fully understand, but perhaps the llvm-ir-testsuite will work if you rebase this branch onto master?

kinke:
Yep, I don't understand it either, but rebasing was a good idea.

JohanEngelen:
Could you try adding a testcase in the ir testsuite for this, and see if you like it / what could be improved?

kinke:
Done. I must say I'm already loving it after playing with it for 5 minutes. :+1:
FileCheck's error messages are nice too:
```
 // LLVM: =^ atomicrmw xor i8*
          ^
<stdin>:268:24: note: scanning from here
 %4 = atomicrmw or i8* %val_arg, i8 %3 seq_cst ; [#uses = 1]
                       ^
<stdin>:287:5: note: possible intended match here
 %4 = atomicrmw xor i8* %val_arg, i8 %3 seq_cst ; [#uses = 1]
    ^
```

tests/ir/atomicrmw.d
```diff
@@ -0,0 +1,40 @@
+// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s --check-prefix LLVM < %t.ll
+
+import core.atomic;
+
+void main() {
+  shared ubyte x = 3;
+  ubyte r;
+
+  r = atomicOp!"+="(x, uint(257));
+  assert(x == r);
```
JohanEngelen: I'm not sure if you want to have asserts in these tests, as it will never be executed. Or, I guess you could add a second LIT run directive:
`// RUN: %ldc -of=%t %s && ./%t`
to have it executed.

kinke: Yeah that's what I originally wanted, but that won't work on Windows (need a .exe suffix). I didn't get to looking up the lit docs yet.

JohanEngelen: @kinke Just had an idea: why don't you simply add an .exe suffix? :) (Linux&Mac don't care about that I think?)
```
// RUN: %ldc -of=%t.exe %s && ./%t.exe
```
But it is perhaps better to skim through LLVM/Clang's testcases to see if they ever execute files. (this was part of my journey to learning LIT and how to use it)

kinke: %t is an absolute path to a temporary file on Windows. So `./%t.exe` can't be executed.

JohanEngelen: @kinke In a PGO test, I use `%ldc -run %s` to compile and run a file. (works on Mac)

JohanEngelen: @kinke Now %ldc -run %s also works on Windows! (bee7c95d3410e1c035134c1ee8c62db69210a907)

PR #1199, opened 2015-11-07
Base implicit executable filename on first object file name

kinke:
And only then on an optional root module name. This should mimic DMD behavior (I haven't checked its source though) and so makes sure LDMD is compatible in this regard.

E.g., when compiling a module `a.d` stating `module tst;` to an executable without providing an explicit output file name, the executable will now be called `a[.exe]` instead of `tst[.exe]`.

### DIFF
--- a/driver/linker.cpp
+++ b/driver/linker.cpp
@@ -62,26 +62,21 @@ static std::string getOutputName(bool const sharedLib) {
   // Output name is inferred.
   std::string result;
 
-  // try root module name
-  if (Module::rootModule) {
+  if (global.params.objfiles->dim) {
+    result = FileName::removeExt(global.params.objfiles->data[0]);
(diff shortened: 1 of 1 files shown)


driver/linker.cpp
```diff
@@ -62,26 +62,21 @@ static std::string getOutputName(bool const sharedLib) {
...
+  } else if (Module::rootModule) {
     result = Module::rootModule->toChars();
-  } else if (global.params.objfiles->dim) {
-    result = FileName::removeExt(
-        static_cast<const char *>(global.params.objfiles->data[0]));
   } else {
     result = "a.out";
   }
+
+  bool isWindows = global.params.targetTriple.isOSWindows();
```
klickverbot: `const` ;)

kinke:
Superseded by #1726.

PR #1202, opened 2015-11-08
propagate "nothrow" to LLVM IR

rainers:
I noticed that "nothrow" was never propageted to the LLVM function object, causing these function to still be "invoked" instead of "called". I guess adding this can reduce code bloat quite a bit for functions with cleanup code.

Maybe omitting it was a deliberate choice? It can change the way exceptions are caught, because no landing pads are generated if the try block is known to be "nothrow". This shows in the unittest of lifetime.d: https://github.com/ldc-developers/druntime/blob/ldc/src/rt/lifetime.d#L2439 which now fails. I think the test is wrong here.

### DIFF
--- a/ir/irfunction.h
+++ b/ir/irfunction.h
@@ -357,8 +357,11 @@ llvm::CallSite ScopeStack::callOrInvoke(llvm::Value *callee, const T &args,
   llvm::Function *calleeFn = llvm::dyn_cast<llvm::Function>(callee);
 
   // Intrinsics don't support invoking and 'nounwind' functions don't need it.
-  const bool doesNotThrow =
-      calleeFn && (calleeFn->isIntrinsic() || calleeFn->doesNotThrow());
+  const bool isIntrinsic = calleeFn && calleeFn->isIntrinsic();
+  const bool isNothrow = calleeFn && calleeFn->doesNotThrow();
+
+  // ignore "nothrow" inside a try/catch block to allow catching Errors
+  const bool doesNotThrow = isIntrinsic || (isNothrow && catchScopes.empty());
 
   if (doesNotThrow || (cleanupScopes.empty() && catchScopes.empty())) {
     llvm::CallInst *call = irs->ir->CreateCall(callee, args, name);
@@ -386,7 +389,16 @@ llvm::CallSite ScopeStack::callOrInvoke(llvm::Value *callee, const T &args,
       irs->ir->CreateInvoke(callee, postinvoke, landingPad, args, name);
 
   if (calleeFn) {
-    invoke->setAttributes(calleeFn->getAttributes());
+    auto attr = calleeFn->getAttributes();
+    if (isNothrow) {
+      attr = attr.removeAttribute(irs->context(),
+                                  llvm::AttributeSet::FunctionIndex,
+                                  llvm::Attribute::NoUnwind);
+      attr =
+          attr.addAttribute(irs->context(), llvm::AttributeSet::FunctionIndex,
+                            llvm::Attribute::NoInline);
+    }
+    invoke->setAttributes(attr);
   }
 
   irs->scope() = IRScope(postinvoke);
--- a/ir/irfunction.h
+++ b/ir/irfunction.h
@@ -262,7 +262,7 @@ public:
   /// are catches/cleanups active or not.
   template <typename T>
   llvm::CallSite callOrInvoke(llvm::Value *callee, const T &args,
-                              const char *name = "");
+                              const char *name = "", bool isNothrow = false);
 
   /// Terminates the current basic block with an unconditional branch to the
   /// given label, along with the cleanups to execute on the way there.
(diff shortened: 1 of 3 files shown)


rainers:
IIRC dmd takes advantage of nothrow for RAII, but not for scope(exit) and similar.

Maybe a compromise is to force "invoke" inside try blocks, but not in general. That would allow catching errors, but cleanup might be skipped inside called functions.

deadalnix:
@klickverbot These assert are only useful for unitests. Why not optimizing based on nothrow only in release mode ?

JohanEngelen:
This seems like a good issue to add a test for in the IR testsuite ;-)

kinke:
> I think we reached the compromise that it would only be applicable in release mode, and throwing Errors would immediately abort the program then.

So would that simply entail extending the if-condition to `f->isnothrow && !global.params.useAssert` or does it involve more work?

rainers:
> Maybe a compromise is to force "invoke" inside try blocks, but not in general. That would allow catching errors, but cleanup might be skipped inside called functions.

I tried this and the tests pass for win32/64 (can't build win64 phobos unittests right now due to assertions for large fixups during compilation). I had to patch LLVM though, because it's impossible to overload the non-existence of an attribute in the instruction. Without the patch, the optimizer still converts invoke instructions to calls.

```
 lib/IR/Instructions.cpp | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/lib/IR/Instructions.cpp b/lib/IR/Instructions.cpp
index 82eb1e0..65539d0 100644
--- a/lib/IR/Instructions.cpp
+++ b/lib/IR/Instructions.cpp
@@ -616,6 +616,9 @@ bool InvokeInst::hasFnAttrImpl(Attribute::AttrKind A) const {
   if (isFnAttrDisallowedByOpBundle(A))
     return false;
 
+  if (A == Attribute::NoUnwind)
+    return false;
+
   if (const Function *F = getCalledFunction())
     return F->getAttributes().hasAttribute(AttributeSet::FunctionIndex, A);
   return false;
```

rainers:
Updated to make nothrow explicite in callOrInvoke instead of setting the function attribute, so no change necessary in LLVM. The druntime functions can still be marked as nothrow in runtime.cpp.

ir/irfunction.h
```diff
@@ -351,14 +351,18 @@ class ScopeStack {
...
 
   // Intrinsics don't support invoking and 'nounwind' functions don't need it.
-  const bool doesNotThrow =
-      calleeFn && (calleeFn->isIntrinsic() || calleeFn->doesNotThrow());
+  const bool isIntrinsic = calleeFn && calleeFn->isIntrinsic();
+  if (!isNothrow)
+    isNothrow = (calleeFn && calleeFn->doesNotThrow());
+
+  // ignore "nothrow" inside a try/catch block to allow catching Errors
+  const bool doesNotThrow = isIntrinsic || (isNothrow && catchScopes.empty());
```
kinke: So if there's a catch scope, you even invoke NoUnwind functions? You probably want something like
```C++
if (isNothrow) {
  // ignore "nothrow" inside a try/catch block to allow catching Errors
  isNothrow = catchScopes.empty();
} else {
  // intrinsics don't support invoking and 'nounwind' functions don't need it
  isNothrow = calleeFn && (calleeFn->isIntrinsic() || calleeFn->doesNotThrow());
}
// no doesNotThrow and no isIntrinsic anymore...
```

rainers: I think we cannot trust 'nounwind' any more than 'nothrow'. If an Error is generated while executing some C function marked 'NoUnwind', we still have to catch it.

BTW: while implementing the Windows exception handling I've come to the conclusion that Errors going through the exception mechanism was a bad design choice. Errors should call a termination function that can be hooked, just like std::terminate. That would make semantics much clearer and implementations and optimizations simpler.

rainers: One optimization to this PR could be to only force invoking when a catch scope handles Errors, not if it is restricted to Exceptions.

kinke: Yeah that's what deadalnix suggested. Not sure if it's really worth it, or maybe I'm just too lazy. ;)

> not if it is restricted to Exceptions

That's the right condition. Any catch handling a non-Exception Throwable should prevent that optimization.

kinke: > I think we cannot trust 'nounwind' any more than 'nothrow'. If an Error is generated while executing some C function marked 'NoUnwind', we still have to catch it.

Then I think the proper fix is to remove the NoUnwind attributes from those runtime functions. As Errors are modeled as special exceptions, I don't think LLVM's NoUnwind attribute semantics (`If the function does raise an exception, its runtime behavior is undefined. However, functions marked nounwind may still trap or generate asynchronous exceptions.`) apply for nothrow functions in general. It should probably be reserved for functions which really never throw.

klickverbot: > I think we cannot trust 'nounwind' any more than 'nothrow'. If an Error is generated while executing some C function marked 'NoUnwind', we still have to catch it.

Oh, but `nounwind` on the IR level literally means that the function does not unwind, and it unwinding is undefined behaviour. The attribute shouldn't be applied automatically either (i.e. just manually to some of our runtime functions, plus by LLVM functions for which the complete IR is available).

rainers: I think we never hit a function with attribute "NoUnwind" as long as LDC doesn't add it itself. So I'm not sure we really have to deal with it when deciding call/invoke. Where would the functions' IR come from?

kinke: Yes, I think the hardcoded druntime functions should be the only ones featuring `nounwind`. So let's just assume the attribute is set appropriately, i.e., only for functions which really never throw. We've been calling them directly for quite a while now.

rainers: Ok, so treating 'nounwind' functions just like intrinsics as in your PR should be fine.

rainers:
Closing in favor of #1656

ir/irfunction.h
klickverbot: (Sometimes, you encounter `nounwind` functions in optimized IR after LLVM's attribute inference, but that of course is irrelevant here.)

PR #1204, opened 2015-11-09
Fix Travis script

kinke:
And re-arrange order of LLVM IR tests (after dmd-testsuite, before unittests).


.travis.yml
```diff
@@ -69,8 +69,6 @@ script:
   # Outputs some environment info, plus makes sure we only run the test suite
   # if we could actually build the executable.
   - bin/ldc2 -version || exit 1
-  # Run LLVM IR testsuite
-  ctest --output-on-failure -R "llvm-ir-testsuite";
```
kinke: Note the missing `-`. This led to Travis not doing anything.

PR #1205, opened 2015-11-10
OS X - fix so MULTILIB test_runner will link (#1059 again)

smolt:
On OS X, test_runners fail link because various casm libs end up in the wrong lib directories. This puts them back in lib.

This is a redo of #1059

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -572,7 +572,7 @@ endforeach()
 # enabled. When linking statically, we have to directly pass the object files
 # to the linking command instead so that all tests are pulled in.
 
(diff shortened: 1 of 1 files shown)


smolt:
I know next to nothing about cmake, so just found the first thing to work.  cmake docs say functions and dirs have scope, and cmake macros are like C macros, so my guess is macros use the scope they are expanded in.  Perhaps build_test_runner should not use output_path but its own var.

smolt:
Well, new commit converts `build_test_runner` to a function.  Not sure if `lib_suffix` is correct for other MULTILIB builds besides OS X.  I guess we'll find out in a bit for linux...

smolt:
hah

PR #1208, opened 2015-11-12
Clean up druntime fwd decls

JohanEngelen:
WORK IN PROGRESS

Makes the code much prettier I hope.
The main motivation is to enable sret for dynamic arrays, see closed PR #901.

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -378,43 +378,10 @@ void DtoResolveFunction(FuncDeclaration *fdecl) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-static void set_param_attrs(TypeFunction *f, llvm::Function *func,
-                            FuncDeclaration *fdecl) {
-  IrFuncTy &irFty = getIrFunc(fdecl)->irFty;
+void applyParamAttrsToLLFunc(TypeFunction *f, IrFuncTy &irFty,
+                             llvm::Function *func) {
   AttrSet newAttrs = AttrSet::extractFunctionAndReturnAttributes(func);
-
-  int idx = 0;
-
-// handle implicit args
-#define ADD_PA(X)                                                              \
-  if (irFty.X) {                                                               \
-    newAttrs.add(idx, irFty.X->attrs);                                         \
-    idx++;                                                                     \
-  }
-
-  ADD_PA(ret)
-
-  if (irFty.arg_sret && irFty.arg_this && gABI->passThisBeforeSret(f)) {
-    ADD_PA(arg_this)
-    ADD_PA(arg_sret)
-  } else {
-    ADD_PA(arg_sret)
-    ADD_PA(arg_this)
-  }
-
-  ADD_PA(arg_nest)
-  ADD_PA(arg_arguments)
-
-#undef ADD_PA
-
-  // Set attributes on the explicit parameters.
-  const size_t n = irFty.args.size();
-  for (size_t k = 0; k < n; k++) {
-    const size_t i = idx + (irFty.reverseParams ? (n - k - 1) : k);
-    newAttrs.add(i, irFty.args[k]->attrs);
-  }
-
-  // Store the final attribute set
+  newAttrs.merge(irFty.getParamAttrs(gABI->passThisBeforeSret(f)));
   func->setAttributes(newAttrs);
 }
 
@@ -499,7 +466,7 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
 
   // parameter attributes
   if (!DtoIsIntrinsic(fdecl)) {
-    set_param_attrs(f, func, fdecl);
+    applyParamAttrsToLLFunc(f, getIrFunc(fdecl)->irFty, func);
     if (global.params.disableRedZone) {
       func->addFnAttr(LLAttribute::NoRedZone);
     }
@@ -928,8 +895,7 @@ void DtoDefineFunction(FuncDeclaration *fd) {
     bb->eraseFromParent();
   } else if (!gIR->scopereturned()) {
     // llvm requires all basic blocks to end with a TerminatorInst but DMD does
-    // not put a return statement
-    // in automatically, so we do it here.
+    // not put a return statement in automatically, so we do it here.
 
     // pass the previous block into this block
     gIR->DBuilder.EmitStopPoint(fd->endloc);
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -16,6 +16,8 @@
 #include "mtype.h"
 #include "root.h"
 #include "gen/abi.h"
+#include "gen/attributes.h"
+#include "gen/functions.h"
 #include "gen/irstate.h"
 #include "gen/llvm.h"
 #include "gen/llvmhelpers.h"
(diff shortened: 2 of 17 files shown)


gen/runtime.cpp
```diff
@@ -207,6 +210,36 @@ template <typename DECL> static void ensureDecl(DECL *decl, const char *msg) {
...
+  for (auto fname : fnames) {
+    llvm::Function *fn =
+        llvm::Function::Create(llvm::cast<llvm::FunctionType>(DtoType(dty)),
+                               llvm::GlobalValue::ExternalLinkage, fname, M);
+    assert(dty->ctype);
+    IrFuncTy &irFty = dty->ctype->getIrFuncTy();
+    gABI->rewriteFunctionType(dty, irFty);
+    fn->setAttributes(attribset.addAttributes(
+        gIR->context(), 1,
+        llvm::AttributeSet::get(gIR->context(), 1, irFty.args[0]->attrs)));
```
kinke: Attributes index 1 is most likely wrong (1st LL parameter). You'll probably want `llvm::AttributeSet::AttrIndex::FunctionIndex = ~0U` for that.

kinke:
> The main motivation is to enable sret for dynamic arrays

More precisely: passing (incl. sret) dynamic arrays (slices) as regular 16-bytes struct for the Win64 ABI. ;)

JohanEngelen:
(I actually only have half a clue what "sret" means, lol :D)

kinke:
https://github.com/ldc-developers/ldc/blob/master/gen/abi.h#L100 hopefully explains it, and https://github.com/ldc-developers/ldc/blob/master/gen/abi-generic.h#L180 should clarify why we need an ABIRewrite for dynamic arrays on Win64.

JohanEngelen:
OK, will do. Saving symbol name cleanup for last (note all the dumb "DD" prefixing for now).

JohanEngelen:
Unfortunately, LLVM3.5 has initializer list constructor of ArrayRef behind an `#ifdef LLVM_HAS_INITIALIZER_LISTS`. That define is really only set for the clang compiler it seems (for gcc it is not enabled): http://lists.llvm.org/pipermail/llvm-dev/2015-February/082266.html
That's pretty annoying... what shall I do... just -DLLVM_HAS_INITIALIZER_LISTS ;)?

kinke:
> -DLLVM_HAS_INITIALIZER_LISTS

Or std::vector<llvm::StringRef>.

JohanEngelen:
I added a semi-overkill solution for ArrayRef  :)

JohanEngelen:
(rebased to hide embarrassing coding error ;))

JohanEngelen:
@kinke I'm sorry I am sometimes flooding the AppVeyor build server with a string of new commits when trying to debug an issue that perhaps I don't see locally. Is there a way to cancel or remove jobs from the queue? (On Travis, I can click the "cancel job" button)

kinke:
Hehe no worries. Unfortunately I have no idea how to give you guys proper access to the AppVeyor project. I just set some explicit permissions for the `user` role, but I haven't figured how to map GitHub accounts to roles. We may need an official ldc-developer account for AppVeyor.

JohanEngelen:
@kinke @klickverbot I think the clean-up is ready. What do you think?
(Next step is to move to Windows and see what breaks when I enable passing of dyn arrays as regular structs. What I remember from last time is that the codegen code assumes too much about the function signatures and will break when the Windows ABI changes.)

gen/runtime.cpp
```diff
@@ -207,23 +212,72 @@ template <typename DECL> static void ensureDecl(DECL *decl, const char *msg) {
...
+        llfunctype, llvm::GlobalValue::ExternalLinkage, fname, M);
+
+    applyParamAttrsToLLFunc(dty, dty->ctype->getIrFuncTy(), fn);
+
+    fn->addAttributes(~0, attribset.getFnAttributes());
+    fn->addAttributes(0, attribset.getRetAttributes());
+    for (size_t i = 0, e = paramtypes.size(); i < e; ++i) {
+      fn->addAttributes(i + 1, attribset.getParamAttributes(
+                                   i + 1)); // func param index starts at 1
+    }
```
kinke: This should be replaceable by
```D
AttrSet attrs = fn->getAttributes();
attrs.merge(attribset);
fn->setAttributes(attrs);
```
with https://github.com/ldc-developers/ldc/pull/1217.

gen/runtime.cpp
```diff
@@ -207,23 +212,72 @@ template <typename DECL> static void ensureDecl(DECL *decl, const char *msg) {
...
+  }
+  int varargs = 0;
+  auto dty = new TypeFunction(params, returntype, varargs, linkage);
+
+  for (auto fname : fnames) {
+    // the call to DtoType performs many actions such as rewriting the function
+    // type and storing it in dty
+    auto llfunctype = llvm::cast<llvm::FunctionType>(DtoType(dty));
+    assert(dty->ctype);
```
kinke: No need to do this for every fname, right?

gen/runtime.cpp
```diff
@@ -207,23 +212,72 @@ template <typename DECL> static void ensureDecl(DECL *decl, const char *msg) {
...
+
+    llvm::Function *fn = llvm::Function::Create(
+        llfunctype, llvm::GlobalValue::ExternalLinkage, fname, M);
+
+    applyParamAttrsToLLFunc(dty, dty->ctype->getIrFuncTy(), fn);
```
kinke: Perhaps it'd be better if this returned an `AttrSet` (could even be a public static function in AttrSet) instead of setting fn's attributes directly. Then the final attributes (based on this + `attribset` below) could be computed once for all `fnames` too.

JohanEngelen: Yep, done :)

gen/runtime.cpp
JohanEngelen: fixed

gen/runtime.cpp
JohanEngelen: thanks!

gen/functions.cpp
```diff
@@ -378,10 +378,8 @@ void DtoResolveFunction(FuncDeclaration *fdecl) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-static void set_param_attrs(TypeFunction *f, llvm::Function *func,
-                            FuncDeclaration *fdecl) {
-  IrFuncTy &irFty = getIrFunc(fdecl)->irFty;
-  AttrSet newAttrs = AttrSet::extractFunctionAndReturnAttributes(func);
+AttrSet getParamAttrs(TypeFunction *f, IrFuncTy &irFty) {
```
kinke: Just one more thing - please convert this to a method of `IrFuncTy`, as it nearly exclusively depends upon it.

JohanEngelen:
It's still not super pretty. It's because IrFuncTy apparently does not know whether to passThisBeforeSret. Fixing it would require more surgery though :(

kinke:
Lgtm, thanks Johan.

So the ABI rewrites are now applied when forward declaring all runtime functions [except for `_d_dso_registry()` on Linux, which is declared directly in IR].
That doesn't seem to have any effect so far, otherwise we'd already have the problem that we are facing next: applying the ABI rewrites for return value and parameters when calling a runtime function. For regular functions, that's all in `DtoCall()` and mainly depends on `IrFuncTy`. So we can't call the `LLFunction` returned by `getRuntimeFunction()` direcly anymore.

JohanEngelen:
@thanks kinke.
Indeed, now we have to add a "callRuntimeFunction()" method I think.

Issue #666, opened 2014-07-05
Migrating fibers across threads vs. TLS


dnadlinger:
`core.thread` fibers are supposed to be safe to migrate across threads, i.e. resume them from a different thread than they `yield`ed from. There is also a [unit test](https://github.com/ldc-developers/druntime/blob/c3253ad45f5966e791484fa8626afde719211a6c/src/core/thread.d#L4372) to verify that this works.

However, there is a problem: LLVM currently (version 3.4/3.5 SVN) always assumes that the address of thread-local variables don't change while on a given stack. This assumption hasn't really caused any problems with non-PIC code so far: TLS accesses are made by addressing off `fs`, `gs` or so in most Windows/Linux x86 ABIs, and as a result there is nothing to gain by caching. However, when using the ELF general dynamic TLS model (default for PIC on Linux), the address is determined via a call to `__tls_get_addr`. If there now are multiple accesses to a TLS variable in the same function, LLVM indeed caches the result of that call on the stack. This is a problem if some code in between those accesses ends up calling `Fiber.yield`, as the thread the code is running on, and thus the address of the TLS symbol, might have changed in the meantime.

In consequence, migrating fibers between threads currently is an unsafe thing to do if the code they execute relies on TLS at all. See [`core.thread.Fiber.switchOut`](https://github.com/ldc-developers/druntime/blob/db7f8ee0295478e1bc67fcde1aa17e1616b6371c/src/core/thread.d#L4251) for an example of how this play out – in user code, the context switch can be hidden behind many layers of calls, of course.

This is not a bug in druntime, but poses a problem for pretty much all userspace fiber/coroutine/green-thread implementations out there. The correct solution is to have an option in the compiler backend to disable caching of TLS addresses across (opaque) function calls. [MSVC has a switch for this](http://msdn.microsoft.com/en-us/library/6e298fy4.aspx), while [the GCC devs refuse to acknowledge the problem altogether](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=26461). [The LLVM/Clang devs are aware of the issue](http://llvm.org/bugs/show_bug.cgi?id=19177), although I don't think I agree with the proposed solution. This should likely be handled by just providing an option to force the target-specific lowering code not to keep the address across function calls.

[...]

PR #1209, opened 2015-11-12
OS X - default OS to macosx instead of darwin

smolt:
Not mean to be the end all as other darwin targets are coming along.  But this makes everything better with Xcode 7.  In particular, OS X ld in Xcode 7 warns when default LLVM triple with OS darwin is used because the darwin default version (x86_64-apple-darwin14.5.0 on my box) is greater than OS X version.  More work has to be done for specifying OS min versions and such.  Hope this is ok for now.

For background: http://forum.dlang.org/post/avnefmjkfgqbfiapibfv@forum.dlang.org

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -741,6 +741,7 @@ static void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("Posix");
     break;
   case llvm::Triple::Darwin:
+  case llvm::Triple::MacOSX:
     VersionCondition::addPredefinedGlobalIdent("OSX");
     VersionCondition::addPredefinedGlobalIdent(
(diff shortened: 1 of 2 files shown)


smolt:
Why doesn't core.thread fail in Travis CI for OS X RELEASE?  It must not be using multiple real CPUs?  I still always get a SEGV due to #666 when I run tests against master.  BTW, I'd like to fix that with ldc-developers/druntime#30 if somebody has the time to review. The PR is old and may need some updates. It would probably also work for Windows and the `TlsGetValue()` `Thread.getThis()` workaround could be removed.

kinke:
Oh, looks like this also got rid of the linker bug for Travis OSX?

Issue #1154, opened 2015-10-11
align() not respected for local variable declarations


redstar:
This application

    // LDC - the LLVM D compiler (0.16.0-beta2):
    //  based on DMD v2.067.1 and LLVM 3.7.0
    //  Default target: x86_64-pc-windows-msvc
    //
    // ldc2 main.d -w -g -unittest -m64 -oq
    
    void main(string[] args)
    {
    }
    
    align(16) struct some_buf
    {
    align(16):
        long[32] data;
    }
    
    int somefun(ref some_buf) { return 0; }
    
    unittest
    {
        import std.stdio;
        some_buf buf;
        writeln(&buf);
        assert((cast(size_t)&buf & cast(size_t)0b1111) == 0);
        void foo()
        {
            writeln("foo");
            somefun(buf);
        }
        foo();
    }

asserts because variable `some_buf` is only `align(8)`, not `align(16)`.

kinke:
Seems to work for struct fields only:
```D
struct Struct { align(16) int[2] a; }
```
=> struct is padded to a size of 16 bytes, and gets properly aligned on stack:
```LL
%align.Struct = type { [2 x i32], [8 x i8] }
%s = alloca %align.Struct, align 16             ; [#uses = 1, size/byte = 16]
```

Struct itself:
```D
align(16) struct Struct { int[2] a; }
```
=> struct is padded to 16 bytes too, but alignment = 4:
```LL
%s = alloca %align.Struct, align 4              ; [#uses = 1, size/byte = 16]
```

A local variable `align(16) int x;` is falsely 4-bytes aligned too.

kinke:
I'll work on this tomorrow. I'll probably come up with an `LLStructBuilder`, which allows aligned fields by inserting padding if need be and providing a field index -> real GEP index mapping (need to skip over padding i8-arrays) as well as a required overall struct alignment (the max of all fields' individual alignments if I'm right) when later allocating such a struct on the stack. Some of this (aligned fields) is already implemented by the `RTTIBuilder` class.

kinke:
The issue wrt. the `align` attribute for structs can be mitigated by using `Type::alignment()` instead of `Type::alignsize()` in `DtoAlloca()` (if != STRUCTALIGN_DEFAULT).

kinke:
The original test case is fixed by https://github.com/ldc-developers/ldc/pull/1159. What still needs to be fixed is the alignment of local variables.

klickverbot:
[...]

PR #1214, opened 2015-11-17
Fix explicit alignment of local variables

JohanEngelen:
This fixes the remaining problem of issue #1154, explicit (overriding) alignment of local variables.

It's the first time I look at the parser code, so please review carefully.
What it looks like to me is that the attributes found on line 3804 are forgotten and not applied to normal local variables without initializers (the last part of the function). So I added code that applies the AlignDeclaration to the declaration, following a pattern found elsewhere in this function. Perhaps we also need to add the other attributes (except linkage?):
```C++
        if (storage_class)
        {
            Dsymbol *s = new StorageClassDeclaration(storage_class, a);
            a = new Dsymbols();
            a->push(s);
        }
        if (structalign != 0)
        {
            Dsymbol *s = new AlignDeclaration(structalign, a);
            a = new Dsymbols();
            a->push(s);
        }
        if (udas)
        {
            Dsymbol *s = new UserAttributeDeclaration(udas, a);
            a = new Dsymbols();
            a->push(s);
        }
``` 

@kinke: this was fun to figure out :) hope you haven't been working on it too.

### DIFF
--- a/dmd2/parse.c
+++ b/dmd2/parse.c
@@ -4067,6 +4067,12 @@ L2:
                     new TemplateDeclaration(loc, ident, tpl, NULL, a2, 0);
                 s = tempdecl;
             }
+            if (structalign != 0)
+            {
+                Dsymbols *ax = new Dsymbols();
+                ax->push(s);
+                s = new AlignDeclaration(structalign, ax);
+            }
             if (link != linkage)
             {
                 Dsymbols *ax = new Dsymbols();
--- a/tests/ir/align.d
+++ b/tests/ir/align.d
@@ -1,4 +1,4 @@
-// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s --check-prefix LLVM < %t.ll
+// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
 
 align(32) struct Outer { int a; }
 struct Inner { align(32) int a; }
@@ -6,11 +6,29 @@ struct Inner { align(32) int a; }
 void main() {
   static Outer globalOuter;
   static Inner globalInner;
-  // LLVM: constant %align.Outer_init zeroinitializer, align 32
-  // LLVM: constant %align.Inner_init zeroinitializer, align 32
+  // CHECK: constant %align.Outer_init zeroinitializer, align 32
+  // CHECK: constant %align.Inner_init zeroinitializer, align 32
 
   Outer outer;
   Inner inner;
-  // LLVM: %outer = alloca %align.Outer, align 32
-  // LLVM: %inner = alloca %align.Inner, align 32
+  // CHECK: %outer = alloca %align.Outer, align 32
+  // CHECK: %inner = alloca %align.Inner, align 32
+
+  align(16) byte byte16;
+  // CHECK: %byte16 = alloca i8, align 16
+  align(64) Outer outer64;
+  // CHECK: %outer64 = alloca %align.Outer, align 64
+  align(128) Inner inner128;
+  // CHECK: %inner128 = alloca %align.Inner, align 128
+
+  alias Byte8 = align(8) byte;
+  Byte8 byte8;
+  // Can aliases contain align(x) ?
+  // C HECK: %byte8 = alloca i8, align 8
+  // C HECK: %byte8 = alloca i8, align 1
+
+  align(16) Outer outeroverride;
+  // Yet undecided if align() should override type alignment:
+  // C HECK: %outeroverride = alloca %align.Outer, align 16
+  // C HECK: %outeroverride = alloca %align.Outer, align 32
 }


kinke:
Cool! :+1:
That's all front-end territory though (so you'd need to either guard it by an `#ifdef IN_LLVM` or fix it directly upstream) - so does DMD not respect these alignments as well?

JohanEngelen:
Looking at the code, DMD has the bug too.
https://github.com/D-Programming-Language/dmd/blob/master/src/parse.d#L4214

(DMD being in D now, how do we keep our code in C??)

kinke:
Oh, I forgot. :D
I guess we'll link against a static dmd-frontend lib. No idea with what compiler that lib is supposed to be generated though. @klickverbot?

JohanEngelen:
Will be some more work than just the simple fix similar to this PR to get this into DMD. (working on it)

JohanEngelen:
Judging from comment 10 here: https://issues.dlang.org/show_bug.cgi?id=2278#c10
This is a new feature, not a bug ?

kinke:
I don't think so, it looks more like a feedback post after some testing. Judging by the other comments, DMD doesn't support any explicit alignment at all. Wow.
Anyway, your fix will be important for GDC and DMD too once they implement it in their backend. Not sure about UDAs etc., but you'll hopefully more information there - although it may take a while.

JohanEngelen:
@kinke How about this one?
```
  align(16) Outer outer16; // largest alignment should win
  // CHECK: %outer16 = alloca %align.Outer, align 32
```

kinke:
> Wouldn't that break the type system?

So would doing this:

```D
struct Container {
  align(1) byte packed;
  align(1) Outer outer;
}
```

outer is stored immediately after the byte at offset 1, and Container padded by 31 bytes:

```LL
%packed.Container = type <{ i8, %packed.Outer, [31 x i8] }>
```

When additionally specifying `align(1) struct Container { ... }`, the padding is removed, and a Container's size is 33 bytes.

JohanEngelen:
My concern is that if we are not more strict on alignment issues, codegen cannot optimize for alignments it knows about. What actually is the use case for `align(32) struct Outer {...}`, if it is so easily defeated?

The case for the Alignedbyte alias is this perhaps:
```
alias vectorizable = align(16) int[4];
```

kinke:
> Not respecting the struct alignment at runtime would probably still be undefined behavior, right?

Yep, that's how I would see it too. I can always cast a random address to a pointer to my aligned type and then access an unaligned object.
If I understand correctly, Johan's asking whether we want to assume all pointers to explicitly aligned types are actually aligned for internal compiler stuff like memcpy() in `DtoAggrCopy()` etc. LLVM should already know about the alignment if the pointer is from a local stack alloca(), but pointer parameters are tricky. Note that this is probably the reason for Temtaime's recent complaint about unaligned store of an `align(16) struct` iirc.

JohanEngelen:
I think this should be discussed upstream too.
For lack of a place to put it, I will put this here. A list of things that need to be clarified (hope you can add more):

```D
import std.typecons;

void main() {
  byte dummy; //increase (?) chances of a non-aligned stack

  // Two 32-byte aligned structs used throughout
  align(32) struct Outer { int a; }
  struct Inner { align(32) int a; }

  static byte dummdumm1;
  static Outer globalOuter;
  static byte dummdumm2;
  static Inner globalInner;
  assert(cast(size_t)&globalOuter % 32 == 0);
  assert(cast(size_t)&globalInner % 32 == 0);

  Outer outer; assert(cast(size_t)&outer % 32 == 0);
  byte dummy1;
  Inner inner; assert(cast(size_t)&inner % 32 == 0);
  byte dummy2;

  // Type aliasing
  alias OuterAlias = Outer;
  OuterAlias outeralias; assert(cast(size_t)&outeralias % 32 == 0);
  byte dummy3;
  alias OuterTypedef = Typedef!(Outer);
  OuterTypedef outertypedef; assert(cast(size_t)&outertypedef % 32 == 0);
  byte dummy4;

  // Explicit variable alignment
  align(16) byte byte16; assert(cast(size_t)&byte16 % 16 == 0);
  byte dummy5;
  align(64) Outer outer64; assert(cast(size_t)&outer64 % 64 == 0);
  byte dummy6;
  align(128) Inner inner128; assert(cast(size_t)&inner128 % 128 == 0);
  byte dummy7;

  // Explicit alignment overrules prior alignment, or should the largest alignment win?
  align(16) Outer outer16; assert(cast(size_t)&outer16 % 32 /+ or 16? +/ == 0);
  byte dummy8;

  struct A {
    align(32) int a;
    void foo() {
        // some function with awesome optimization that assumes 32 byte alignment
        assert(cast(size_t)&a % 32 == 0);
    }
  }
  align(16) A a;
  a.foo(); // oops?
  // therefore:
  assert(cast(size_t)&a % 32 = 0); // or 16?
  // Where this may accidentally happen is in a template:
  // what about an align(16) applied to T in template code, and T is passed an align(32) type?

  byte dummy9;

  Outer outerarray[15];
  static assert(typeof(outerarray).sizeof == 15*32);
  assert(cast(size_t)&outerarray[1] % 32 = 0);
  byte dummy10;

  // Alignment of array acts only on full array
  align(16) int alignedarray[4];
  assert(cast(size_t)&alignedarray[0] % 32 == 0);
  assert(cast(size_t)&alignedarray[1] % 32 == 1); // elements may be unaligned
  byte dummy11;

  // Type alias containing alignment directive
  alias AlignedVector = align(16) int[4];
  AlignedVector alignedvector; assert(cast(size_t)&alignedvector[0] % 16 == 0);
  byte dummy12;
  alias AlignedVectorTD = Typedef!(align(16) int[4]);
  AlignedVectorTD alignedvectorTD; assert(cast(size_t)&alignedvectorTD[0] % 16 == 0);
  byte dummy13;
}
```

kinke:
LL function parameters passed with the ByVal attribute (i.e., for most ABIs: most D structs and static arrays passed by value) probably cannot be aligned, as LLVM chooses the function parameters stack slot. That may really be a problem.

JohanEngelen:
Yeah, I think you are right. 
```D
void foo(AlignedClass param) {
   param.method();  // method cannot assume this is aligned
}
```
So that greatly simplifies things :)
I will try and get the extended alignment.d testfile in our testsuite.

JohanEngelen:
@klickverbot @kinke I am not sure how to proceed with this. What do you want to do? Open a discussion/page upstream about this? In the meanwhile, what to do about the bug in DMD (parser eats the "align" attribute but discards it, the fix is straightforward but then the backend also needs to use it (LDC already does))?

kinke:
Not sure about the rest too, i.e. a proper alignment discussion, but wrt. to

> what to do about the bug in DMD (parser eats the "align" attribute but discards it, the fix is straightforward but then the backend also needs to use it (LDC already does))?

Please open a DMD PR. Then the backend **can** use it; we already do, GDC probably will soon and DMD... well, who knows.

JohanEngelen:
Rebase + extra changes:  only apply the parsed align() to identifiers (stack variables) and type aliases. Before, it was also applied to function declarations.

JohanEngelen:
DMD PR : https://github.com/D-Programming-Language/dmd/pull/5284

kinke:
Perfect, Walter already merged it upstream. Could you please update this commit here as well?

JohanEngelen:
I have removed the parse tree fix for `alias Int16 = align(16) int`, because I can't test it and so I did not submit that upstream. The PR is now identical to the merged upstream PR.
It also adds disabled testcases for future alignment issues.

kinke:
Thanks, merging.

JohanEngelen:
To support `alias Integer16 = align(16) int;', I think we have to be able to assign an alignment to a type. Which has many implications and so should be discussed upstream.

PR #1216, opened 2015-11-17
Misc. druntime updates

kinke:
* sections_elf_shared: Do not depend on __tls_get_addr for static builds
* MSVCx86: use pretty names instead of mangled names for exceptions
* Clean up LDC-specifics in core.math
* Fiber migration and safety checks


kinke:
Closing as Kai already merged in the stdlib updates.

PR #1219, opened 2015-11-22
Profile-guided optimizations (PGO)

JohanEngelen:
See this page on the wiki that documents the work: http://wiki.dlang.org/LDC_LLVM_profiling_instrumentation

This PR implements PGO as Clang/LLVM does it:
1.  Compile with instrumentation:  `ldc2 -fprofile-instr-generate test.d -of=test1`
2.  Run executable:  `./test1`
     This generates a "default.profraw" file.
3.  Run llvm-profdata tool:  `llvm-profdata merge default.profraw -output test.profdata`
4.  Compile again, now using profile data: `ldc2 -profile-instr-use=test.profdata test.d -of=test2`

Instead of adding the LLVM profiler runtime to druntime like the PR does now, I think it should be a separate library that can be linked to in case of `-fprofile-instr-generate`.


JohanEngelen:
@klickverbot Thanks for the testcase. I get a different error though: assert in PGO code. Traversing DMD's AST is non-trivial :/

gen/pgo.cpp
```diff
@@ -0,0 +1,1065 @@
+//===-- gen/pgo.cpp ---------------------------------------------*- C++ -*-===//
+//
+//                         LDC – the LLVM D compiler
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+
+//===--- CodeGenPGO.cpp - PGO Instrumentation for LLVM CodeGen --*- C++ -*-===//
```
klickverbot: Could you merge this into the preceding LDC header comment? Maybe just note that it is adapted from `CodeGenPGO.cpp`, and that it is under LLVM's license. The reference to LICENSE.TXT also doesn't make a lot of sense in our repository.

JohanEngelen:
Johan, take a look at this too at some point: http://reviews.llvm.org/D15829

smolt:
I talk to myself often too, but not yet on github.

JohanEngelen:
Rebased onto master!

JohanEngelen:
Now also working for LLVM 3.9

JohanEngelen:
@klickverbot @redstar The work is done! Please review, and merge ;-)

(This PR is becoming larger and larger. By now, so much time has gone into this that i am very biased towards it. It's been tested to safely compile Phobos/druntime. The frontend changes are not there for the fun of it. Some frontend changes have been submitted upstream, but https://github.com/D-Programming-Language/dmd/pull/5501 was met with disappointing review, especially considering [time investment PGO] vs [apparent review time].)

etcimon:
Just curious, has anyone benchmarked this yet in favorable conditions? What are the potential gains?

JohanEngelen:
@etcimon I did some simple tests, but have not been able to create testcases with big performance improvements. I too am very curious to see what the gains can be (and for what kind of code).

JohanEngelen:
@klickverbot https://github.com/D-Programming-Language/dmd/pull/5501 is not going anywhere, what do you think?

JohanEngelen:
@etcimon PGO is relatively new in LLVM, and I think that LLVM 3.8 actually does not use much of the profiling information during optimization (e.g. afaik it is not used for inlining). Development of PGO is pretty active in trunk though.

smolt:
Curious if it could make LDC faster, optimizing D front-end, and clang w/ PGO make a faster LLVM?

JohanEngelen:
@klickverbot  Rebased!

What needs some work in the future is PGO for exception handling statements (try/catch/finally/scope). It is not so easy, and I don't expect to work on it soon. Hope this can be merged without it. @rainers I will need your help to implement PGO for MSVC EH!

JohanEngelen:
This is currently broken on Windows because of COMDAT "any" selection for all functions. I will have to ask on LLVMdev whether this can be worked around or not.

JohanEngelen:
CircleCI is broken because it uses latest LLVM. I was waiting for https://github.com/ldc-developers/ldc/pull/1420 but now I see it is for merge-2.071

JohanEngelen:
Still fails on Windows. It needs LLVM >= r270596 (#1513).

JohanEngelen:
Yeah I had seen it. It is sad: perhaps front-end PGO in LDC is doomed now (so much work.... luckily some of the code is still relevant for coverage). I still think front-end PGO has use (e.g. the virtual call optimization is much harder in IR, but could perhaps be solved with some extra front-end IR annotations), but the case for it is a lot weaker when Clang front-end PGO is stopped. But some people argue that they want to use front-end pgo (because it also gives coverage at the same time). Not a good argument I think: coverage tests corner-cases, not the common behavior of your program that you want to PGO for.

JohanEngelen:
GREEN, also on AppVeyor! :-)

kinke:
Congratz! :+1

JohanEngelen:
Btw, for using IRPGO we also need compiler-rt. So if later on, IRPGO starts to rock more, we can pull out a bit of the PGO code and only keep it for coverage (LLVM coverage annotation is on my todo list, but it's too long already...).

JohanEngelen:
> Would it be a good idea to put up a CircleCI job that builds everything with instrumentation enabled?

Maybe? Build LDC with instrumentation -> build phobos -> rebuild LDC using instrumentation?

JohanEngelen:
Really happy about this :)  Now I need to write that promised blog post.

JohanEngelen:
It's a good point. I think not all bugs I found during development made it into the testcases.

What we could do is add a test job that builds Phobos's unittests with instrumentation enabled. That should give decent coverage of complex ASTs with `-profile-instr-generate`. Then for testing `-profile-instr-use` with complex ASTs we could add a profile file to the repo, that is then used while building something (perhaps again the phobos unittests). All this to say that: we don't have to actually generate the profile on the tester.

MaskRay:
Interesting. e8e28b33b2cfd51400e03cfe64232e1e818a9db0 added -fprofile-generate

clang supports -fprofile-generate (IR based; the interface is the same as GCC but the underlying implementation is very different) and -fprofile-instr-generate (AST-based; used by -fcoverage-mapping). IMHO -fprofile-instr-generate is rarely used outside of  -fcoverage-mapping. Most PGO users use -fprofile-generate. Has anyone compared ldc's  -fprofile-generate and  -fprofile-instr-generate?

JohanEngelen:
@jondegenhardt Perhaps has tested the difference.

jondegenhardt:
Can't use IR based PGO with my tools, as IR based PGO doesn't work with LTO, and I combine PGO and LTO in my build scripts. The bug report: #2582 .

I could setup tests of AST-based PGO vs IR-based PGO, without LTO, but that would involve it's own bias, at least in my tools. The tools and benchmark suite I have setup makes significant use of druntime/phobos in tight loops, and those facilities need LTO to make the IR available to the optimizer.

PR #1220, opened 2015-11-23
Update druntime with disabled Fiber migration test

smolt:
Don't run the Fiber migration unittest unless version FiberMigrationTest
is set.  Otherwise, just print a message without failing.


smolt:
Rolled back to druntime to before all Fiber check commits.  Just to see what happens with tests, see if there is more green.

smolt:
Same red.  It is something in the earlier druntime commits that is causing the other failures.

smolt:
experiment over...

PR #1221, opened 2015-11-25
Avoiding directives that split up parts of statements.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

I noticed that the project uses this types of directives only in a few places by running my tool. I'm willing to contribute by removing those issues if it is important to the project.

### DIFF
--- a/dmd2/statement.c
+++ b/dmd2/statement.c
@@ -5117,6 +5117,7 @@ Statement *GotoStatement::semantic(Scope *sc)
 
 bool GotoStatement::checkLabel()
 {
+    int error_msg;
     if (!label->statement)
     {
         error("label '%s' is undefined", label->toChars());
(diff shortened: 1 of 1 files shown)


RomeroMalaquias:
Hello @klickverbot i'm from the same research team of flaviommedeiros and i need to ask if you prefer this code before or after the commit. What do you think about this kind of change?
I mean, would you accept if it was not for the fact we submitted the pull request to a deprecated code and if the declaration of the variable was did in the right place?
Thanks in advance.

PR #1168, opened 2015-10-24
[WIP] support building for x86 with MSVC runtime

rainers:
Work in progress, but made public to show there is some progress.

Needs LLVM master (3.8) and ~~https://github.com/ldc-developers/druntime/pull/35~~ https://github.com/ldc-developers/druntime/pull/60
~~Exceptions work in simple cases, but combinations still cause trouble:~~

Open for comments...

PR #1222, opened 2015-11-29
Win64 ABI: Rewrite magic __c_long_double struct as double

kinke:
Another positive side-effect of #1168.

To be complemented by https://github.com/D-Programming-Language/druntime/pull/1443.

### DIFF
--- a/gen/abi-win64.cpp
+++ b/gen/abi-win64.cpp
@@ -38,26 +38,9 @@ private:
   IntegerRewrite integerRewrite;
   MSVCLongDoubleRewrite longDoubleRewrite;
 
-  // Returns true if the D type is an aggregate:
-  // * struct
-  // * static/dynamic array
-  // * delegate
-  // * complex number
-  static bool isAggregate(Type *t) {
-    TY ty = t->ty;
-    return ty == Tstruct || ty == Tsarray ||
-           /*ty == Tarray ||*/ ty == Tdelegate || t->iscomplex();
-  }
-
   static bool isMagicCppLongDoubleStruct(Type *t) {
     return t->ty == Tstruct &&
            static_cast<TypeStruct *>(t)->sym->ident == Id::__c_long_double;
-  }
-
-  // Returns true if the D type can be bit-cast to an integer of the same size.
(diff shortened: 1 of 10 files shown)


gen/abi-win64.cpp
```diff
@@ -129,27 +131,32 @@ void Win64TargetABI::rewriteFunctionType(TypeFunction *tf, IrFuncTy &fty) {
...
 
     // the copy is treated as a local variable of the callee
     // hence add the NoAlias and NoCapture attributes
     arg.attrs.clear().add(LLAttribute::NoAlias).add(LLAttribute::NoCapture);
+  } else if (t->ty == Tstruct &&
+             static_cast<TypeStruct *>(t)->sym->ident == Id::__c_long_double) {
+    arg.rewrite = &longDoubleRewrite;
   } else if (isAggregate(t) && canRewriteAsInt(t) &&
-             !IntegerRewrite::isObsoleteFor(originalLType)) {
+             !IntegerRewrite::isObsoleteFor(arg.ltype)) {
```
kinke: We land here for `cpp_(u)long`, but the IntegerRewrite is obsolete (passing `{ i32 }` vs. `i32` makes no difference).

gen/abi-win64.cpp
```diff
@@ -129,27 +131,32 @@ void Win64TargetABI::rewriteFunctionType(TypeFunction *tf, IrFuncTy &fty) {
...
     // these types are passed byval:
     // the caller allocates a copy and then passes a pointer to the copy
     arg.rewrite = &byvalRewrite;
-    arg.ltype = byvalRewrite.type(arg.type, arg.ltype);
 
     // the copy is treated as a local variable of the callee
     // hence add the NoAlias and NoCapture attributes
     arg.attrs.clear().add(LLAttribute::NoAlias).add(LLAttribute::NoCapture);
+  } else if (t->ty == Tstruct &&
+             static_cast<TypeStruct *>(t)->sym->ident == Id::__c_long_double) {
```
kinke: We only land here if `__c_long_double` is <= 64 bits. I.e., a x87 MinGW `__c_long_double` is passed in memory as any other x87 value.

PR #1227, opened 2015-12-05
Streamline generic x86 ABI incl. some fixes

kinke:


rainers:
LGTM.
Is there an issue with adding the patch https://github.com/ldc-developers/ldc/pull/1168#issuecomment-162257833 to this PR?

kinke:
> Is there an issue with adding the patch #1168 (comment) to this PR?

Not really, it would just need the `callingConv()` signature changes from your commit https://github.com/rainers/ldc/commit/3e36652c2cc2016d1c7dd8ff7229678671fa1706. I thought I'd better leave this to you in #1168, but I can obviously also put it in here.

kinke:
It'd be nice if someone could verify the OSX x86 tests with this (I don't have an OSX VM, and no Apple products at all ;)).

smolt:
@kinke - I'll give it a test drive on OSX

kinke:
Thx Dan!

smolt:
I am getting a Segmentation fault now for -m32 runnable/cppa.d.  I'll take a look...

gen/abi-aarch64.cpp
```diff
@@ -17,7 +17,8 @@
 #include "gen/abi-aarch64.h"
 
 struct AArch64TargetABI : TargetABI {
-  llvm::CallingConv::ID callingConv(LINK l) {
+  llvm::CallingConv::ID callingConv(llvm::FunctionType *ft, LINK l,
```
smolt: I think callingConv() here is a hangover from before it was virtual.  I think it can be removed because it returns CallingConv::C like everybody else now.  Would it be ok to do in this PR (kinda unrelated)?

kinke:
@smolt: It's been a continuous WIP, so I don't know which commit you actually tested. ;) The last one now works for me with Linux x86. Note that we expect Clang to be used on OSX, that's what makes OSX interesting.
I'll remove the superfluous AArch64 override once it's green.

smolt:
It was the second to last: e7f26a9

smolt:
I'll update to last commit 54ceb2490adae2992b23e1560acdbb7c358c2ace and run tests again before I start debugging.

kinke:
Thx, that sounds good. My Win64 dmd-testsuite tests are also successful with that commit.

kinke:
Dammit, I missed the complex numbers. Will take another round at least ;)

smolt:
The segfault in runnable/cppa.d happens when Dmain returns.  Stack must be off.  I'll track it down later.

kinke:
Thx, I'll head off to bed. Looking good otherwise.

smolt:
The failed test in runnable/cpp.d is D using sret on a C++ struct, while clang++ says to return directly in registers.  Its this function:
```
struct S13707
{
    void* a;
    void* b;
    this(void* a, void* b)
    {
        this.a = a;
        this.b = b;
    }
}

extern(C++) S13707 func13707();
```
I read clang source, peeked at ItaniumCXXAB.cpp and TargetInfo.cpp.  According to `ItaniumCXXABI::classifyReturnType()`, the struct return is indirect if there is either a nonTrivialDestructor or nonTrivialCopyConstructor.  The test struct `S13707` has neither of these.  Our `TargetABI::isPOD()` almost works for OSX except for the hasCtor clause.  A little edit and OSX -m32 runnable/cppa will pass:

```D
bool TargetABI::isPOD(Type *t, LINK l) {
  t = t->toBasetype();
  assert(t->ty == Tstruct);
  StructDeclaration *sd = static_cast<TypeStruct *>(t)->sym;
  // the C++ definition of PODs excludes structs with constructors
  //return sd->isPOD() && !(l == LINKcpp && hasCtor(sd));
  return sd->isPOD();
}
```

I would guess Linux would have the same issue, but CI shows it passing.  Hmmm.  Maybe this needs an OSX unique change.

kinke:
Thx Dan. I thought it'd be a common C++ definition, but apparently it isn't. The constructors are needed for the MSVC++ definition. gcc/g++ on Linux apparently returns **all** aggregates (except for cfloat, according to our code) via sret, so it doesn't make a difference on Linux.

smolt:
clang++ I think is default on FreeBSD now.  I targeted some clang supported OSes and looked at IR to see what happens on `func13707()`:

OS     | IR
--------|---
darwin | `define i64 @_Z9func13707v()`
linux    | `define void @_Z9func13707v(%struct.S13707* noalias sret %agg.result)`
freebsd | `define i64 @_Z9func13707v()`
netbsd | `define void @_Z9func13707v(%struct.S13707* noalias sret %agg.result)`
openbsd | `define i64 @_Z9func13707v()`
dragonflybsd | `define i64 @_Z9func13707v()`

smolt:
I think we will need some sort of C++ abi thingie based on targeted OS.

kinke:
Horrible. One more reason to hate 32-bit code, and love the 2 clearly defined ABIs with no additional calling convention nuances for x86_64 (Win64, System V). ;) Thanks for checking!

kinke:
I've slightly extended your table (clang is indeed very helpful here):

OS     | POD | With ctor
--------|---|---
darwin | directly | directly
linux    | sret | sret
windows-msvc | directly | sret
freebsd | directly | directly
netbsd | sret | sret
openbsd | directly | directly
dragonflybsd | directly | directly

So MSVC (same for x64) is a special case here wrt. C++ structs with ctors. Linux and NetBSD seem pretty conservative by not using regs at all when returnings structs.

kinke:
Merging this eagerly to pave the way for #1168.

smolt:
We forgot all about Solaris.  It should be handled like Linux.

kinke:
Confirmed, thanks for pointing that out - will be fixed shortly.

PR #1229, opened 2015-12-08
Replace ubiquitous gIR->context by an implicit conversion of *gIR

kinke:


kinke:
Well, I love it, I've really come to hate `gIR->context()`. But we don't need to agree on everything. ;)

kinke:
And it's not like I absolutely want it either. ;) I just did some counting:

* `*gIR`: used in 177 lines
* `gIR->ir`: 150 lines
* `gIR->module`: 114
* `gIR->[^(ir|module)]`: 149 lines

`gIR->context()` is used almost exclusively as argument, that's what made the implicit conversion handy. I agree that it doesn't feel 100% correct conceptually in that it is implicitly convertible to its module's context instead of the module itself; it just so happens that the context is used in 60% more lines (I thought the number would be even much higher).

PR #1230, opened 2015-12-11
x86 ABI: Fix Solaris regression and work around MSVC byval alignment issue

kinke:


smolt:
I inspected the solaris IR generated for cppa.d and retested OSX 32-bit, passes.  Looks good.

PR #1232, opened 2015-12-13
Mark most GEPs `inbounds` and add additional param attributes

kinke:


gen/tocall.cpp
```diff
@@ -531,7 +531,7 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
     assert(bitmask == 31 || bitmask == 63);
     // auto q = cast(size_t*)ptr + (bitnum >> (64bit ? 6 : 5));
     LLValue *q = DtoBitCast(ptr, DtoSize_t()->getPointerTo());
-    q = DtoGEP1(q, p->ir->CreateLShr(bitnum, bitmask == 63 ? 6 : 5), "bitop.q");
+    q = DtoGEP1(q, p->ir->CreateLShr(bitnum, bitmask == 63 ? 6 : 5), true, "bitop.q");
```
kinke: Not sure about this being a legal `inbounds`.

ir/irclass.cpp
```diff
@@ -373,7 +373,7 @@ llvm::GlobalVariable *IrAggr::getInterfaceVtbl(BaseClass *b, bool new_instance,
                                    : 1];
       LLType *targetThisType = thisArg->getType();
       thisArg = DtoBitCast(thisArg, getVoidPtrType());
-      thisArg = DtoGEP1(thisArg, DtoConstInt(-b->offset));
+      thisArg = DtoGEP1(thisArg, DtoConstInt(-b->offset), true);
```
kinke: Not sure about this one either.

klickverbot: Should be fine, this is the same as http://llvm.org/docs/LangRef.html#prefix-data, right?

gen/tocall.cpp
klickverbot: The intended semantics for the intrinsics are not really defined in any detail in the upstream docs, but I guess it should be fine.

kinke:
The 2nd commit adds `nonnull` attributes for the `this`/`nestedContext` implicit params, and `dereferenceable<N>` for D `ref`/`out` params.

gen/functions.cpp
```diff
@@ -141,27 +144,30 @@ llvm::FunctionType *DtoFunctionType(Type *type, IrFuncTy &irFty, Type *thistype,
...
     if (arg->storageClass & STClazy) {
       // Lazy arguments are lowered to delegates.
       Logger::println("lazy param");
       auto ltf = new TypeFunction(nullptr, arg->type, 0, LINKd);
       auto ltd = new TypeDelegate(ltf);
       loweredDType = ltd;
-    } else if (!passPointer) {
+    } else if (passPointer) {
+      // ref/out
+      attrs.addDereferenceable(loweredDType->size());
```
kinke: `getTypeAllocSize(DtoType(loweredDType))` doesn't work here as some LL types seem opaque.

ir/irclass.cpp
kinke: Yep, looks similar. The docs also state that the pointer must only point somewhere into the 'object' (whatever that exactly is), not exclusively to its start, so this should be fine.

gen/functions.cpp
klickverbot: That's strange, but I guess we'll roll with it for now.

PR #1233, opened 2015-12-14
Update druntime and remove additional VS2015 .lib from config file

kinke:
The required lib is now referenced as defaultlib in druntime's `msvc.c` if druntime is built with VS 2015, see https://github.com/ldc-developers/druntime/pull/47.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -193,11 +193,6 @@ endif()
 # Create configuration files.
 #
 
-# Add a lib required by VS 2015+ when building LDC with VS 2015+.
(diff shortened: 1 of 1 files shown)


smolt:
Looks pretty straightforward.

kinke:
It is. ;) Just wasn't sure whether I should mention the druntime commits in this commit's msg.

smolt:
I do like when submodule changes are summarized, but it is also easy enough to look at the submodule commits diffs (which I did).  And I almost always build against latest ldc branches anyway unless I am doing a "cleanroom" build.

kinke:
Oh I've just figured that your ARM commits have already been merged, so it's really just Rainer's commit. Merging.

PR #1238, opened 2015-12-20
Prepare DtoCallFunction() for runtime function calls

kinke:
By also taking `DValue*` args, not just `Expression*`s.

### DIFF
--- a/gen/tocall.cpp
+++ b/gen/tocall.cpp
@@ -208,18 +208,16 @@ static void addExplicitArguments(std::vector<LLValue *> &args, AttrSet &attrs,
 
 ////////////////////////////////////////////////////////////////////////////////
 
-static LLValue *getTypeinfoArrayArgumentForDVarArg(Expressions *arguments,
-                                                   int begin) {
+static LLValue *
+getTypeinfoArrayArgumentForDVarArg(const std::vector<DValue *> &argvals,
+                                   int begin) {
   IF_LOG Logger::println("doing d-style variadic arguments");
   LOG_SCOPE
 
   // number of non variadic args
   IF_LOG Logger::println("num non vararg params = %d", begin);
 
-  // get n args in arguments list
-  size_t n_arguments = arguments ? arguments->dim : 0;
-
-  const size_t numVariadicArgs = n_arguments - begin;
+  const size_t numVariadicArgs = argvals.size() - begin;
 
   // build type info array
   LLType *typeinfotype = DtoType(Type::dtypeinfo->type);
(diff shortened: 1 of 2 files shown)


kinke:
A first step towards calling runtime functions with the `DtoCallFunction()` machinery. What's still missing:

1. Right now, we only have a LL function for each runtime function. We'll need to keep the associated `TypeFunction*` around too in order to create the `DValue* fnval` argument for `DtoCallFunction()`.
2. Replace all direct LL function calls by `DtoCallFunction()` calls, using `DValue*` args instead of LL args etc.

JohanEngelen:
Nice that you are working on the follow-up to #1208 !
I was thinking about adding a map to gen/runtime.h from functionstringname -> (llvm::Function, TypeFunction).

gen/tocall.cpp
```diff
@@ -1009,3 +980,34 @@ DValue *DtoCallFunction(Loc &loc, Type *resulttype, DValue *fnval,
 
   return new DImValue(resulttype, retllval);
 }
+
+DValue *DtoCallFunction(Loc &loc, Type *resulttype, DValue *fnval,
+                        Expressions *arguments, llvm::Value *retvar) {
+  TypeFunction *const tf = DtoTypeFunction(fnval);
+
+  const size_t numArguments = arguments ? arguments->dim : 0;
+  std::vector<DValue *> argvals(numArguments);
```
JohanEngelen: Perhaps you can use llvm::SmallVector here (number of arguments is usually small I'd think).

gen/tocall.cpp
```diff
@@ -204,18 +204,16 @@ static void addExplicitArguments(std::vector<LLValue *> &args, AttrSet &attrs,
 
 ////////////////////////////////////////////////////////////////////////////////
 
-static LLValue *getTypeinfoArrayArgumentForDVarArg(Expressions *arguments,
-                                                   int begin) {
+static LLValue *
+getTypeinfoArrayArgumentForDVarArg(const std::vector<DValue *> &argvals,
```
JohanEngelen: Instead of taking a std::vector&, you can use llvm::ArrayRef.

kinke:
I'm not a huge fan of `llvm::ArrayRef` due its missing support for initializer lists in LLVM 3.5 (and 3.6?) and I doubt a `llvm::SmallVector` provides any measurable performance benefit over a single small heap allocation (size known a priori, elements are pointers only, each `DValue` arg allocated on the heap as well iirc), but I can still be persuaded. ;)

I still need to figure out how to structure the logging, as the args are currently evaluated outside the `DtoCallFunction` logger scope.

JohanEngelen:
Yeah, forget about my comments :)

redstar:
Rebase on ltsmaster and merge?

kinke:
Finally rebased incl. taking care of logging.

PR #1239, opened 2015-12-22
Fix a few compiler warnings.

JohanEngelen:
This fixes a few compiler warnings, and tries to simplify the structure of a few #if constructs.

### DIFF
--- a/gen/abi-x86.cpp
+++ b/gen/abi-x86.cpp
@@ -49,9 +49,8 @@ struct X86TargetABI : TargetABI {
     case LINKwindows:
       return ft->isVarArg() ? llvm::CallingConv::C
                             : llvm::CallingConv::X86_StdCall;
-    default:
-      llvm_unreachable("Unhandled D linkage type.");
     }
+    llvm_unreachable("Unhandled D linkage type.");
   }
 
   std::string mangleForLLVM(llvm::StringRef name, LINK l) override {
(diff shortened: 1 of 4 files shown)


gen/abi-x86.cpp
```diff
@@ -49,9 +49,8 @@ struct X86TargetABI : TargetABI {
     case LINKwindows:
       return ft->isVarArg() ? llvm::CallingConv::C
                             : llvm::CallingConv::X86_StdCall;
-    default:
-      llvm_unreachable("Unhandled D linkage type.");
```
smolt: it doesn't seem there should be a warning for this as all return paths are covered.  Which compiler issues a warning?

JohanEngelen: clang on OS X.
It warns that all enum cases are covered, and so you don't need/want a default statement. My reasoning for my fix is that, if something is added to the LINK enum, we would get a compiler warning instead of a runtime error.

smolt: I too am using OS X, clang with Xcode 7.  I don't get that warning, hmmm..
But I like your reasoning.

smolt:
There are a bunch of unused variable/function warnings too for stuff only used in assert.  Like `checkVarValueType()` in value.cpp.  Does it make sense to wrap some of these like:

```c++
#if !defined(NDEBUG)
static bool checkVarValueType(LLType *t, bool extraDeref) {
...
}
#endif
```

JohanEngelen:
I think we should try to aim for -Werror, within reason. It's a pity that the compiler sees "assert" as unused stuff though. I'm not sure about removing those functions in release builds. I always build with debug on (it turns out), and so I may accidentally commit code that will not build in release mode if I happen to start using that function. Don't know a better solution though.

smolt:
Me too on -Werror.  There are enough warnings to sift through that it is sometimes hard to see when I make new ones.

JohanEngelen:
Merged this.

smolt:
Good note on -Werror.

PR #1240, opened 2015-12-22
A char may not be signed, use signed char

antonblanchard:
On PowerPC and ARM a char is unsigned. Fix a few places that
relied on it being signed.


smolt:
Since `istream::get()` returns an int, wouldn't that be better than signed char?  Or perhaps `auto`?

antonblanchard:
Sure, I'll re-spin the patch to use auto.

smolt:
thanks!

PR #1244, opened 2016-01-03
Implement rudimentary support for @ldc.attributes.target("...")

JohanEngelen:
The implementation is rudimentary, but I think it works well enough.
The string passed to target(...) is not checked, and so the user has to be careful in specifying valid options. LLVM does warn on invalid options.
One significant difference with clang is that for the string "sse4.1", this PR will set "target-features" to "+sse4.1", whereas clang will set a long list of features "sse,sse2,sse3,ssse3,sse4.1" and more. From simple testcases, it appears this is not needed for LLVM to do the right thing.
Clang has an elaborate implementation that deals with x86's options, see clang/lib/basic/Targets.cpp. Perhaps that functionality can be moved into LLVM; if not, we will have to copy that code (and keep it up-to-date with the latest microarchitecture...).

### DIFF
--- a/gen/uda.cpp
+++ b/gen/uda.cpp
@@ -7,11 +7,14 @@
 #include "expression.h"
 #include "module.h"
 
+#include "llvm/ADT/StringExtras.h"
+
 namespace {
 
 /// Names of the attribute structs we recognize.
 namespace attr {
 const std::string section = "section";
+const std::string target  = "target";
 }
 
 bool isFromLdcAttibutes(StructLiteralExp *e) {
@@ -28,6 +31,28 @@ bool isFromLdcAttibutes(StructLiteralExp *e) {
     return false;
   }
   return true;
+}
+
+StructLiteralExp *getLdcAttributesStruct(Expression *attr) {
+  // See whether we can evaluate the attribute at compile-time. All the LDC
+  // attributes are struct literals that may be constructed using a CTFE
+  // function.
+  unsigned prevErrors = global.startGagging();
+  auto e = ctfeInterpret(attr);
+  if (global.endGagging(prevErrors)) {
+    return nullptr;
+  }
+
+  if (e->op != TOKstructliteral) {
+    return nullptr;
+  }
+
+  auto sle = static_cast<StructLiteralExp *>(e);
+  if (isFromLdcAttibutes(sle)) {
+    return sle;
+  }
+
+  return nullptr;
 }
 
 void checkStructElems(StructLiteralExp *sle, llvm::ArrayRef<Type *> elemTypes) {
@@ -50,7 +75,6 @@ void checkStructElems(StructLiteralExp *sle, llvm::ArrayRef<Type *> elemTypes) {
 }
 
 const char *getFirstElemString(StructLiteralExp *sle) {
-  checkStructElems(sle, {Type::tstring});
   auto arg = (*sle->elements)[0];
   assert(arg->op == TOKstring);
   auto strexp = static_cast<StringExp *>(arg);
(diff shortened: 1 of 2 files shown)


JohanEngelen:
Don't merge this, I have to fix the druntime reference.

9il:
Amazing! I have some thought about this feature that require a little bit different syntax. I will write them today.

JohanEngelen:
Rebased onto master: the "target-features" attribute seems to work from LLVM 3.7, so I updated the codegen test to require at least llvm 3.7.

JohanEngelen:
@9il This is not yet the function multiversioning that you want!

JohanEngelen:
@9il You are thinking about a slightly different syntax for this PR (adding special target options to a function)  or for function multiversioning?

9il:
> @9il You are thinking about a slightly different syntax for this PR (adding special target options to a function) or for function multiversioning?

Both.
About syntax:
1. It should be strict (with comma and without spaces and tabs)
2. `fpmath` should be implemented as `pragma`, IMO.

> I think we could also implement this as a library solution, instead of compiler-internally. Would that make more sense?

No, I will use it as library solution only for DMD/GDC. Multiversioning with dispatching is preferable because it reduces an executable size.

In the same time multiversioning is not so important right now and can be implemented later.

The `__target` feature and new `pragma`s are much more important:

```D
pragma(LDC_fpmath, "fast"); //<--- most important along with `pragma(inline, xxx)`
pragma(LDC_loop_vectorize, true);
pragma(LDC_fp_contract, "fast");
pragma(LDC_loop_vectorize_width, 16);
//other loop pragmas ...
````

```D
version(LDC)
{
	enum bool a = __target(has, "avx2");
	enum bool b = __target(compatible, "core-avx2");
	enum bool c = __target("broadwell");
}
else version(GNU)
{
	...
}
```

How it can help?
An engineer can check target and instruction set in compile time and choose appropriate `__vector` type and intrinsic.

The `__target` feature, the new `pragma`s, and `pragma(inline, false/true/)` should be implemented before target attributes.

JohanEngelen:
> About syntax:
> 1. It should be strict (with comma and without spaces and tabs)
Why? Clang allows whitespace. What's the downside?

> 2. fpmath should be implemented as pragma, IMO.
So you can set it at file scope?

> Multiversioning with dispatching is preferable because it reduces an executable size.
I think we can implement the exact thing you would get with compiler support, using D library code. The syntax to get multiversioning will be different, but perhaps not too awkward.

For the __target thing. How about only providing functionality to get the target string (if __target() == "broadwell"; this would just be a simple read of a global variable, really easy to implement), and implementing the rest as a library? This would enable someone to reuse some of it, potentially.
At the moment, unfortunately, all x86 target info stuff is implemented in Clang instead of LLVM (it appears). So it is kind of painful to redo all that in LDC, and it might be nice to have some hope of someone being able to use it for something else in D (when it is a library). (also easy sharing between GDC and LDC, I suppose)

9il:
> About syntax:
> 1. It should be strict (with comma and without spaces and tabs)
> Why? Clang allows whitespace. What's the downside?

It is better to have strong syntax constraints. We can relax them in the future (but probably we would not it). I hope to have multiversioning as part of D spec in the future.

> fpmath should be implemented as pragma, IMO.
> So you can set it at file scope?

The reason is clear syntax. `fpmath` is instruction to a compiler like` inline`. There is no need to have this as attribute.

> Multiversioning with dispatching is preferable because it reduces an executable size.
> I think we can implement the exact thing you would get with compiler support, using D library code. The syntax to get multiversioning will be different, but perhaps not too awkward.
> 
> For the __target thing. How about only providing functionality to get the target string (if __target() == "broadwell"; this would just be a simple read of a global variable, really easy to implement), and implementing the rest as a library? This would enable someone to reuse some of it, potentially.
> At the moment, unfortunately, all x86 target info stuff is implemented in Clang instead of LLVM (it appears). So it is kind of painful to redo all that in LDC, and it might be nice to have some hope of someone being able to use it for something else in D (when it is a library). (also easy sharing between GDC and LDC, I suppose)

Agreed, but with a little bit more functionality:

http://clang.llvm.org/docs/CrossCompilation.html#target-triple
```
The triple has the general format <arch><sub>-<vendor>-<sys>-<abi>, where:
arch = x86, arm, thumb, mips, etc.
sub = for ex. on ARM: v5, v6m, v7a, v7m, etc.
vendor = pc, apple, nvidia, ibm, etc.
sys = none, linux, win32, darwin, cuda, etc.
abi = eabi, gnu, android, macho, elf, etc.
```

and CPU string like in command line.

Examples:
```
__target(arch) //arm
__target(sub) //v6m
__target(vendor) //apple
__target(sys)//darwin
__target(abi) //eabi
```

and for CPU

```
__target(cpu) //"broadwell" // if user passes -mcpu=broadwell or if broadwell is native
```

```
__target(attr) //"avx,-avx2" // if user passes -mattr=avx,-avx2 (or empty string "" otherwise)
```

The `cpu` and `attr` are more important.

JohanEngelen:
Merged!

>>> About syntax:
>>> 1. It should be strict (with comma and without spaces and tabs)

>>Why? Clang allows whitespace. What's the downside?

>It is better to have strong syntax constraints. We can relax them in the future (but probably we would not it). I hope to have multiversioning as part of D spec in the future.

For lack of a downside, I left the whitespace trimming in.

JohanEngelen:
@9il Could you have a look in the old forum thread again? http://forum.dlang.org/post/fyckroqmeuluuwytocvn@forum.dlang.org

9il:
@JohanEngelen  __traits(targetHasFeature, "sse4") looks good

Issue #1119, opened 2015-09-25
0.16.0-beta1: Trivial program fails on FreeBSD


redstar:
Compiled ldc2 on FreeBSD 10 (PC-BSD 10.2) with statically linked LLVM 3.7.

A `Hello world` program fails with `object.Exception@/usr/home/kai/Build/LDC-BUILD/src/ldc/runtime/phobos/std/stdio.d(2390): Enforcement failed`.

Happens with LLVM 3.6, too.

    [kai@pcbsd-10] ~% ldc2 -version
    LDC - the LLVM D compiler (136253):
      based on DMD v2.067.1 and LLVM 3.7.0
      Default target: x86_64-unknown-freebsd10.2
      Host CPU: nehalem
      http://dlang.org - http://wiki.dlang.org/LDC
    
      Registered Targets:
        x86    - 32-bit X86: Pentium-Pro and above
        x86-64 - 64-bit X86: EM64T and AMD64

TheDharc:
same here FreeBSD 11-CURRENT, LLVM 3.7. the build works perfectly, i think the problem is building phobos.

PR #1248, opened 2016-01-09
FreeBSD: generate calls to _d_dso_registry (as expected by druntime).

redstar:
This fixes issue #1119.


redstar:
These are the only 2 places where the `global.params.is<OS>` variables are used. IMHO we should use the triple here and kill the `global.params.is<OS>` variables. Any other opinion?

PR #1249, opened 2016-01-10
Do not use global.params.is<OS>

redstar:
This PR replaces the few uses of `global.params.is<OS>` with
`global.params.targetTriple.isOS<OS>()`. This avoids adding a new
boolean for each supported OS.
It also defines all xBSD-type OS as using the dso-registry.

Only `global.params.isWindows` is not replaced because it is used by the frontend code.

### DIFF
--- a/dmd2/globals.h
+++ b/dmd2/globals.h
@@ -80,13 +80,15 @@ struct Param
 #endif
     bool is64bit;       // generate 64 bit code
     bool isLP64;        // generate code for LP64
+#if !IN_LLVM
     bool isLinux;       // generate code for linux
     bool isOSX;         // generate code for Mac OSX
+#endif
     bool isWindows;     // generate code for Windows
+#if !IN_LLVM
     bool isFreeBSD;     // generate code for FreeBSD
     bool isOpenBSD;     // generate code for OpenBSD
     bool isSolaris;     // generate code for Solaris
-#if !IN_LLVM
     bool mscoff;        // for Win32: write COFF object files instead of OMF
     char useDeprecated; // 0: don't allow use of deprecated features
                         // 1: silently allow use of deprecated features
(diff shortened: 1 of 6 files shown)


dmd2/mtype.c
```diff
@@ -2355,7 +2355,7 @@ Identifier *Type::getTypeInfoIdent(int internal)
 
     size_t off = 0;
 #if !IN_LLVM
-    if (global.params.isOSX || global.params.isWindows && !global.params.is64bit)
+    if (global.params.targetTriple.isMacOSX() || global.params.targetTriple.isOSWindows() && !global.params.is64bit)
```
klickverbot: !IN_LLVM anyway?

smolt:
Just a heads up that all the `isMacOSX()` here will eventually be `isOSDarwin()` to support all the other Apple OSes.

yebblies:
Shouldn't we be trying to hide this inside `Target`?

dmd2/mtype.c
redstar: Fixed. Thanks for the hint.

redstar:
@smolt `isMacOSX()` is used in exactly one place. Should I change it already?

redstar:
@yebblies I think this would be good. I have not yet checked why the frontend needs the distinction between Windows and other OS's. I hope it's easy to add.

smolt:
@redstar - if you feel like it, go ahead and change `isMacOSX()` to `isOSDarwin()`, but might be confusing without changing comments referring to OSX and `const bool isOSX` in abi-x86.cpp.

If you don't, it will get updated when (sometime, I hope) iOS stuff is merged.  Still waiting on upstream.

For some reason I thought `isMacOSX()` showed up in more places.

redstar:
@smolt Please make this change together with the iOS stuff. This gives a better context.

smolt:
yup!

PR #1258, opened 2016-01-17
Cmake: Remove trailing space from LLVM_LDFLAGS.

redstar:
This happens on my ARM device and causes a cmake error.


JohanEngelen:
@redstar There is also the string(STRIP ...) cmake command that you can use to trim whitespace.

redstar:
@JohanEngelen Thanks. I will change it.

Issue #1252, opened 2016-01-16
OS X: runnable/testabi.d segmentation fault due to unaligned data


smolt:
This is the crash plaguing the OS X release build on Travis right now (commit d3da239).

First of all to leave no surprises, the problem lies outside of LDC as the same crash can be crafted with C code (I'll show it later).  So what to do?  Document the problem, suggest a workaround, and file a bug report with Apple as I think it is a linker problem.

This problem comes and goes based on the alignment of data being linked.  At one time the crash happened when running the 32-bit OS X release version of testabi.d.  Now it has moved over to the 64-bit release.  It might only happen when LDC has optimizer enabled, but this is not an optimizer error.  With clang, you don't even need `-O` to get the same problem.

Let's look at a session showing runnable/testabi.d crash. 
```
$ ldmd2 ldc/tests/d2/dmd-testsuite/runnable/testabi.d -O -g
$ ./testabi 
Segmentation fault: 11
$ lldb testabi
(lldb) target create "testabi"
Current executable set to 'testabi' (x86_64).
(lldb) r
Process 46379 launched: '/Users/dan/projects/ldc-master/testabi' (x86_64)
Process 46379 stopped
* thread #1: tid = 0x27d5c1, 0x0000000100004f4d testabi`_D7testabi7D_test1FZv + 11981 at testabi.d:245, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x0000000100004f4d testabi`_D7testabi7D_test1FZv + 11981 at testabi.d:245
   242 	        }
   243 	        assert( pass );
   244 	
-> 245 	        results_1[0..5] = 0;
   246 	        foreach( n, T; R_T )
   247 	        {
   248 	                test1_call_out!(T)(n);
(lldb) disass --line
testabi`_D7testabi7D_test1FZv + 11968 at testabi.d:245
   244 	
   245 	        results_1[0..5] = 0;
   246 	        foreach( n, T; R_T )
testabi`_D7testabi7D_test1FZv:
    0x100004f40 <+11968>: leaq   0x58c79(%rip), %rdi       ; _D7testabi9results_1G85i
    0x100004f47 <+11975>: callq  *(%rdi)
    0x100004f49 <+11977>: pxor   %xmm0, %xmm0
->  0x100004f4d <+11981>: movdqa %xmm0, (%rax)
(lldb) p/x $rax
[...]

PR #1260, opened 2016-01-17
Update druntime with ldc-developers/druntime#53

smolt:
Workaround issue #1252 runnable/testabi.d segmentation fault on OS X.


smolt:
I'll post Apple's response once I get it.

Issue #1250, opened 2016-01-14
Can't compile a release build on OS X


John-Colvin:
```
johncolvin@thermo ~/Git/ldc/build (git)-[master] % mkdir build && cd build
johncolvin@thermo ~/Git/ldc/build (git)-[master] % cmake .. -DCMAKE_BUILD_TYPE=Release
-- The C compiler identification is AppleClang 7.0.2.7000181
-- The CXX compiler identification is AppleClang 7.0.2.7000181
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found LLVM: /usr/local/Cellar/llvm/3.6.2 (found suitable version "3.6.2", minimum required is "3.5") 
-- Found LibConfig: /usr/local/include  
-- The ASM compiler identification is Clang
-- Found assembler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- LDC version identifier: 04c6c4
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of unsigned short
-- Check size of unsigned short - done
-- Using unsigned short
-- Check if the system is big endian - little endian
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Looking for setupterm in tinfo
[...]

PR #1247, opened 2016-01-08
NetBSD support

nrTQgc:

PR #1262, opened 2016-01-21
cmake: Detect if llvm assertions enabled

smolt:
Set cmake var LLVM_ENABLE_ASSERTIONS based on llvm-config --assertion-mode.

This is a solution to #1250.

This option isn't in a build I have of llvm 3.4 but it is in my 3.5.1 build.  Does LDC support older LLVM than 3.5.1?

### DIFF
--- a/cmake/Modules/FindLLVM.cmake
+++ b/cmake/Modules/FindLLVM.cmake
@@ -151,6 +151,7 @@ else()
     llvm_set(HOST_TARGET host-target)
     llvm_set(INCLUDE_DIRS includedir true)
     llvm_set(ROOT_DIR prefix true)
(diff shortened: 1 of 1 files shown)


smolt:
The AppVeyor std.datetime unittest failure was in an earlier build too.  Yeah it is unrelated, but not clear where it came from.  Pull request #1247 passed but the merge had this same unittest failure on AppVeyor.

https://ci.appveyor.com/project/kinke/ldc/build/1.0.766/job/ibefvkc49c1l2ewd

kinke:
The AppVeyor failure should be fixed by https://github.com/D-Programming-Language/phobos/pull/3945.

PR #1263, opened 2016-01-24
buildRuntimeModule: Always create _d_dso_registry forward decl (small cleanup)

klickverbot:
The added overhead for Windows et al. is utterly insignificant
compared to all the other typically unused declarations that
will be created by this, and copy/pasting the long condition
(see d072bdf) is error-prone. The alternative would have been
to factor the check out into a predicate function.


JohanEngelen:
perhaps you could add a small comment that it is unused on Windows

Issue #1211, opened 2015-11-15
Invalid bitcast error


SiegeLord:
While compiling Tango on 64 bits, LDC 0.16.1 gives me an error (see log here https://travis-ci.org/SiegeLord/Tango-D2/jobs/86535484):

```
Invalid bitcast

  %6 = bitcast %tango.core.Time.seconds.DurationClone %5 to %core.time.Duration

LLVM ERROR: Broken function found, compilation aborted!
```

This code compiled fine with LDC 0.15.1.

redstar:
Confirmed. Happens with 0.17.0-beta2, too.

redstar:
Reduced test case:

    struct Foo
    {
        long baz;
    }
    
    Foo foo(long x)
    {
        struct Bar
        {
            long y;
        }
    
        return cast(Foo)Bar(x);
    }

PR #1270, opened 2016-01-30
Extend DtoBitCast() to support aggregate values.

redstar:
The LLVM instruction bitcast works only on non-aggregate values and types.
As long as struct are stored in memory there is no problem. But if the
struct has integral size (e.g. only an int member) then the value is not
necessarily stored in memory. Nevertheless there may be the need to cast
this value to a different type.
DtoBitCast() is extended to store and reload the value with differnt types
in this case.

This functionality is used to fix issue #1211.

### DIFF
--- a/gen/statements.cpp
+++ b/gen/statements.cpp
@@ -204,8 +204,7 @@ public:
             returnValue =
                 LLConstant::getNullValue(irs->mainFunc->getReturnType());
           } else {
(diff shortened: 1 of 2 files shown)


redstar:
I had this thought, too. Do we have a general repainting function somewhere?

redstar:
`DtoCast` works with the frontend types. But here we have already lowered to the LLVM types.

PR #1271, opened 2016-01-30
irtype: Align static array handling with dynamic arrays and pointers

klickverbot:
The only intended change is to restore the `assert(!dt->ctype);` check. I still don't trust my handling of forward references in the type translation system…


redstar:
We do the same work currently.... `IrTypeFunction` and `IrTypeDelegate` have the same pattern (and the same bug), too.

Issue #1112, opened 2015-09-24
Can't build gtk-d 3.1.4


beannaich:
Trying to build a project through dub with gtk-d 3.1.4 listed as a dependency results in the following error:

```
Assertion failed: !dt->ctype && "already has IrType", file C:\projects\ldc\ir\irtype.cpp, line 35
0x00007FF6735E2EE5 (0x0000000000000016 0x00007FF6729E39D5 0x0000000000000023 0x00007FFB32449594)
0x00007FFB324A4A68 (0x0000000000000001 0x00007FF600000000 0x0000000000000000 0x000000AA6B83EE50), raise() + 0x1E8 bytes(s)
0x00007FFB324A5A31 (0x00007FFB00000003 0x0000000000000003 0x00007FF674097D20 0x00007FF674097C68), abort() + 0x31 bytes(s)
0x00007FFB324A7A81 (0x0000000000000023 0x00007FF674097D20 0x000000AA6F2F2760 0x0000000000000000), _get_wide_winmain_command_line() + 0x1171 bytes(s)
0x00007FFB324A7D8F (0x000000AA16808A60 0x000000AA6B83F1A0 0x000000AA161990A8 0x0000000000000002), _wassert() + 0x3F bytes(s)
0x00007FF6729E39D5 (0x000000AA16808A60 0x0000000000000000 0x0000000000000000 0x000000AA6F2F2760)
0x00007FF6729E9605 (0x000000AA161990A8 0x000000AA6B83F1A0 0x000000AA16199050 0x0000000000000002)
0x00007FF6729E99B2 (0x000000AA6F2F2760 0x0000000000000003 0x000000AA6F2F2AC0 0x00007FF6728CE577)
0x00007FF6728F4B82 (0x0000000000000004 0x000000AA6B83F1A0 0x0000000000000003 0x0000000000000004)
0x00007FF6729E4012 (0x000000AA6F2F2AC0 0x000000AA16798FA0 0x000000AA16184080 0x000000AA6B9C3940)
0x00007FF6728F4965 (0x0000000000000002 0x000000AA6B83F1A0 0x0000000000000003 0x0000000000000004)
0x00007FF672A1F054 (0x0000000000000100 0x000000AA6B83F3B0 0x000000AA16809A60 0x0000000000000010)
0x00007FF6729E7384 (0x000000AA6B9AE738 0x000000AA6B83F3B0 0x000000AA16198B68 0x0000000000000020)
[...]

PR #1272, opened 2016-01-30
Do not construct Ir-Types twice for the same type.

redstar:
This is the same as in PR #1269. The Ir-Types may be initialized
twice (due to pointers, forward references etc.).
This commit fixes hopefully all instances of this problem, using
the same approach in each factory function.

This fixes issue #1112.


redstar:
It looks like as soon as a wrapper function is called (e.g. `DtoMemType()`) then Pandora's box is opened.

redstar:
Reworked to use the style from `IrTypeArray` / `IrTypeSArray`.

PR #1273, opened 2016-01-30
DtoCast: Handle struct repainting

### LINKED ISSUE #1211: Invalid bitcast error

klickverbot:
Previously, we would just ignore such casts altogether, leading to problems down the road when we would actually load the alloca/… for the source struct.

Fixes #1211.

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -580,6 +580,23 @@ DValue *DtoCastVector(Loc &loc, DValue *val, Type *to) {
   fatal();
 }
 
+DValue *DtoCastStruct(Loc &loc, DValue *val, Type *to) {
+  Type *const totype = to->toBasetype();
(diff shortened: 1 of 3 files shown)


redstar:
It fixes the bug in #1211.

kinke:
Completely off-topic: glad to see an explosion of GitHub activity here, sure looks as if Kai has finished his book and time to spend on LDC again. :)

redstar:
Right - my book is now published. :-)

PR #1276, opened 2016-02-01
fix tilde expansion

markos:
While building the debian package for 0.17.0~beta1 (for various reasons beta versions need to be prefixed with ~ so that they get superceeded by normal releases) the build failed. Tracking down the issue revealed that tilde expansion is done wrong in dmd, breaking the build. According the POSIX spec, tilde expansion can occur only in the beginning of the path and only if the next char is a path separator, so I took the liberty of fixing a patch, which fixes the debian package build -which is now being uploaded in unstable.

Feel free to make any modifications in case I have missed something.

### DIFF
--- a/dmd2/root/filename.c
+++ b/dmd2/root/filename.c
@@ -96,6 +96,21 @@ Strings *FileName::splitPath(const char *path)
             while (isspace((utf8_t)*p))         // skip leading whitespace
                 p++;
             buf.reserve(strlen(p) + 1); // guess size of path
+
+#if POSIX
+           if (*p == '~') // Only replace tilde on first char
+           {
+               if (*(p+1) == '/') // and only if next char is path separator
+               {
+                   char *home = getenv("HOME");
+                   if (home)
+                       buf.writestring(home);
+               } else
+                   buf.writestring("~");
+           }
+#endif
+
+
             for (; ; p++)
             {
                 c = *p;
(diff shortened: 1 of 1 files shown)


redstar:
IMHO this go upstream, too. There is no sense to insert path to $HOME in the middle of another path.

dmd2/root/filename.c
```diff
@@ -96,6 +96,21 @@ Strings *FileName::splitPath(const char *path)
             while (isspace((utf8_t)*p))         // skip leading whitespace
                 p++;
             buf.reserve(strlen(p) + 1); // guess size of path
+
+#if POSIX
```
redstar: Please add #if IN_LLVM before to mark as LDC specific.

dmd2/root/filename.c
```diff
@@ -125,18 +140,6 @@ Strings *FileName::splitPath(const char *path)
                     case '\r':
                         continue;       // ignore carriage returns
 
-#if POSIX
```
redstar: Just do not deleted. Please surround by #if !IN_LLVM.

markos:
Ok, will create a new pull request with the suggestions

redstar:
Thanks. We mark every change to frontend files with this style. This makes merging much more easier.

markos:
Ok, just did that, thanks for the tip.

JohanEngelen:
@markos I think something went wrong: the diff does not show any "IN_LLVM".

JohanEngelen:
@markos Has this gone upstream?

thewilsonator:
Fixed by https://github.com/dlang/dmd/pull/10284

PR #1280, opened 2016-02-05
Implement ARM AAPCS ABI and other ARM fixes

smolt:
This is the bulk of the AAPCS (Procedure Call Standard for the ARM Architecture) used for C interop.  This will be used for both arm and thumb target triples.  The extern(D) ABI tries to follows the AAPCS with a few exceptions.

### DIFF
--- a/gen/abi.cpp
+++ b/gen/abi.cpp
@@ -12,6 +12,7 @@
 #include "id.h"
 #include "gen/abi-generic.h"
 #include "gen/abi-aarch64.h"
+#include "gen/abi-arm.h"
 #include "gen/abi-mips64.h"
 #include "gen/abi-ppc64.h"
 #include "gen/abi-win64.h"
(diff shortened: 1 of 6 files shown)


smolt:
Expanding with fix for relocation model.  Now -O0 works.

Clang makes it an error if dynamic-no-pic is used with something other than darwin.  Should LDC do that too?  What do you think.

gen/abi.cpp
```diff
@@ -235,6 +343,9 @@ TargetABI *TargetABI::getTarget() {
   case llvm::Triple::aarch64:
   case llvm::Triple::aarch64_be:
     return getAArch64TargetABI();
+  case llvm::Triple::arm:
+  case llvm::Triple::thumb:
```
redstar: What about `armeb` and `thumbeb`?

redstar:
Never saw this many unit tests pass on my IFC6410. Really great!

gen/abi.cpp
smolt: Yeah, I almost always forget them.

smolt:
@redstar, do you have list of what failed?  I am going to open an issue later to work off remaining arm-linux failures (not many) unless you beat me to it.  We can cross check our failure lists.

Issue #1286, opened 2016-02-09
When a class contains a union, other fields are not statically initialized


WalterBright:
Described here:

https://github.com/D-Programming-Language/dmd/pull/5422

klickverbot:
The link should have gone to D-Programming-Language/dmd#5424, I suppose.

Anyway, a self-contained test case:

```d
class Foo {
    union {
        char* us;
        wchar* uw;
    }
    ubyte sz = 0xcc;
}

extern(C) int printf(const char*, ...);
void main() {
    auto f = new Foo;
    printf("%x\n", f.sz);
}
```

klickverbot:
Well…
```llvm
@_D4test3Foo6__initZ = constant %test.Foo_init { %test.Foo.__vtbl* @_D4test3Foo6__vtblZ, i8* null, i8* null, i16* null, i8 -52 }, align 8 ; [#uses = 2]
```

We emit both the union members!

Interestingly, this doesn't seem to be covered by the DMD testsuite at all.

WalterBright:
If we add it to the dmd test suite, then ldc will always fail in the travis build/test.

JohanEngelen:
I think the bug arises because the union is lowered into class member with a certain offset (overlapping). Currently, LDC assumes that only structs can have overlapping fields. For classes, LDC assumes no overlapping fields, and treats the union fields as normal fields.
Working on a fix.

JohanEngelen:
These union problems are tested in tag2.069 of the dmd-testsuite.

klickverbot:
@WalterBright: Just to clear up the confusion: We use the DMD test suite (i.e. `test`) as part of our CI testing. I was surprised to find out that the test suite from 2.068.2 (which is what the current LDC release is based of) apparently doesn't cover that case at all, since the bug went by unnoticed until your DMD change.

The DMD Travis tests only use LDC as part of building the DDMD frontend to make sure it stays easy to bootstrap from the last C++-only version of all the compilers. LDC never comes in contact with test suite itself (at least until we merge the corresponding frontend version).

ZombineDev:
@klickverbot I'm almost certain that I hit the same bug with DMD 2.069, so it's not suprising that it's not part of the test suite.

JohanEngelen:
Closing, merged #1287

PR #1287, opened 2016-02-10
Fix union initialization. Fixes issue 1286.

JohanEngelen:
Field overlap checking was only done for structs, and only for implicit initializers. This PR fixes that.

The `AggregateDeclaration *decl` argument to `IrAggr::addFieldInitializers(...)` is never a `UnionDeclaration`. Instead, a union's fields are merged with the encompassing class's/struct's fields, with offsets such that the fields overlap. 

I have not included codegen testcases, but opened an issue to do so. I want to focus on the merge 2.069 branch, and this issue popped up there too (`runnable/interpret.d`).
It appears that explicit initializers are given for cases where I do not see any in the code. I believe the function will need more work, hopefully uncovered while try to construct tough codegen testcases.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -98,26 +98,17 @@ static void write_struct_literal(Loc loc, LLValue *mem, StructDeclaration *sd,
 
     // get initializer expression
     Expression *expr = (index < nexprs) ? exprs[index] : nullptr;
-    if (!expr) {
-      // In case of an union, we can't simply use the default initializer.
+
+    if (vd->overlapped && !expr) {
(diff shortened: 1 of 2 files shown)


JohanEngelen:
Fixes #1286 .

JohanEngelen:
I just discovered VarDeclaration::overlapped that could be used for solving this.

JohanEngelen:
The issue fixed in DMD with this PR remains in LDC:
https://github.com/D-Programming-Language/dmd/pull/4863/files

JohanEngelen:
I have fixed another issue with union initialization and updated this PR.

PR #1303, opened 2016-02-18
AppVeyor: Use LDC 0.17.0 instead of DMD.

JohanEngelen:


appveyor.yml
```diff
@@ -82,8 +82,20 @@ install:
...
   - set PATH=%PATH%;c:\projects\dmd2\windows\bin
-  - set DC=c:\projects\dmd2\windows\bin\dmd.exe
   - dmd --version
+  # Download & extract LDC 0.17.0
+  - ps: Start-FileDownload 'https://github.com/ldc-developers/ldc/releases/download/v0.17.0/ldc2-0.17.0-win64-msvc.zip' -FileName 'ldc2.zip'
+  - md ldc_0_17_0
+  - cd ldc_0_17_0
+  - 7z x ..\ldc2.zip > nul
+  - cd ..
+  - ldc2 --version
```
kinke: `c:\projects\ldc_0_17_0\bin\ldc2.exe` as that dir has not been added to the PATH (which is probably fine, to prevent the build scripts from accidentally using that LDC later on).

redstar:
The script still downloads dmd. I think this can be removed.

JohanEngelen:
Linking now works when built with LDC, but ldc2 crashes upon the simplest of tasks.

PR #1304, opened 2016-02-19
Proof-of-concept: Use real_t in front-end for target real type

kinke:
A modified version without hacks should go upstream.
This is just to prove that the front-end works properly with 80-bit `real_t` for the Travis jobs and with 64-bit `real_t` for the AppVeyor jobs (the latter using dmd to compile the front-end, which therefore features 80-bit reals).

There are only 2 remaining occurrences of `real` in the front-end - as potential type for the `real_t` alias and as the return type of the host C function `strtold()` in `ddmd.root.real_t`.


ddmd/root/port.d
```diff
@@ -42,31 +44,19 @@ extern (C++) struct Port
...
+    enum ldbl_max = real_t.max;
+    enum ldbl_nan = real_t.nan;
+    enum ldbl_infinity = real_t.infinity;
+    enum ldbl_dig = real_t.dig;
+    enum ldbl_epsilon = real_t.epsilon;
+    enum ldbl_mant_dig = real_t.mant_dig;
+    enum ldbl_max_10_exp = real_t.max_10_exp;
+    enum ldbl_max_exp = real_t.max_exp;
+    enum ldbl_min_10_exp = real_t.min_10_exp;
+    enum ldbl_min_exp = real_t.min_exp;
```
kinke: Easily replaceable for custom `real_t` types (software emulations), if need be even as globals instead of compile-time constants.

ddmd/tokens.d
```diff
@@ -735,7 +735,7 @@ extern (C++) struct Token
 
     extern (C++) const(char)* toChars()
     {
-        __gshared char[3 + 3 * float80value.sizeof + 1] buffer;
+        __gshared char[3 + 3 * floatvalue.sizeof + 1] buffer;
```
kinke: I hope this holds for other types as well, not just 80-bit x87 values.

ddmd/root/port.d
ibuclaw: real_t_ just needs to be defined in its own module.

ddmd/tokens.d
ibuclaw: Yes, it should be.  The buffer even on x86 is very lenient.

ibuclaw:
> and as the return type of the host C function strtold() in ddmd.root.port.

ddmd.root.port.strtold should return ``real_t``.  If you are providing software emulation floating point, you are writing your own implementation of ddmd.root.port anyway.

kinke:
> ddmd.root.port.strtold should return real_t.

I've been thinking about it as well. It **should** return a `real_t`, but almost certainly returns a host `real` since it's the `strtold()` from the C runtime linked against the host compiler, and that's how `strtold()` is defined in druntime as well. My additional motivation was that `real` may help with a warning if it's not implicitly convertible to the used `real_t` type.

ddmd/root/port.d
kinke: Do you prefer a separate module analogous to the previous `longdouble` one, with a dmd implementation of ~2 lines? GDC/LDC could also use a separate module and import the custom type here before setting the `real_t` alias.
Well as you're not using this `port.d` at all, it's probably easier to factor `real_t` out into its own module. That should also be the place for all `real_t` related stuff currently in `port.d`, right?

ibuclaw:
> but almost certainly returns a host real since it's the strtold() from the C runtime linked against the host compiler, and that's how strtold() is defined in druntime as well.

Again, you'd implement your own version of ``Port::strtold``.  https://github.com/D-Programming-GDC/GDC/blob/16a15523c01e996660a0ad8b8043dd97d36535ee/gcc/d/d-port.cc#L147-L185

Whatever C ``strtold()`` does (if it even exists) (if it even supports the correct ISO standard for valid float literals in D to be interpreted https://sourceware.org/bugzilla/show_bug.cgi?id=18669 :-) - ignore it.  There are so many levels of wrong in even depending on it. https://issues.dlang.org/show_bug.cgi?id=14718

kinke:
The `real_t` stuff is now in its own module, and I've tried to resync the C++ headers with the D sources (they were already out of sync).

ddmd/constfold.d
```diff
@@ -544,12 +543,12 @@ extern (C++) UnionExp Pow(Loc loc, Type type, Expression e1, Expression e2)
...
         {
             emplaceExp!(ComplexExp)(&ur, loc, e1.toComplex(), e1.type);
-            emplaceExp!(ComplexExp)(&uv, loc, complex_t(1.0, 0.0), e1.type);
+            emplaceExp!(ComplexExp)(&uv, loc, complex_t(1, 0), e1.type);
         }
         else if (e1.type.isfloating())
         {
             emplaceExp!(RealExp)(&ur, loc, e1.toReal(), e1.type);
-            emplaceExp!(RealExp)(&uv, loc, ldouble(1.0), e1.type);
+            emplaceExp!(RealExp)(&uv, loc, cast(real_t)1, e1.type);
```
yebblies: Use `real_t(1)` instead of a cast?

ddmd/builtin.d
```diff
@@ -449,7 +449,7 @@ else
     add_builtin("_D4core4math4sqrtFNaNbNiNffZf", &eval_sqrt);
     // @safe @nogc pure nothrow real function(real, real)
     add_builtin("_D4core4math5atan2FNaNbNiNfeeZe", &eval_unimp);
-    if (Port.yl2x_supported)
+    if (TargetReal.yl2x_supported)
```
ibuclaw: What has yl2x to do with the real type?

ibuclaw:
> The real_t stuff is now in its own module, 

I'm not sure about this approach of moving things out of Port to it.  Seems like an unnecessary and incomplete (missing strtof and strtod) abstraction, with a couple others that should be agnostic to whatever Target you are supporting (yl2x).

As proof of concept I don't think it needs anything more than an alias to real, a template for explicit casting (was ldouble), and a template for formating (was ld_sprint).

ddmd/mtype.d
```diff
@@ -3709,10 +3703,7 @@ else
             case Tcomplex80:
             case Timaginary80:
             case Tfloat80:
-version(IN_LLVM)
-                ivalue = Port.ldbl_dig;
-else
-                ivalue = LDBL_DIG;
+                ivalue = real_t.dig;
```
ibuclaw: Yes, this is the correct way to go about it.

These are not necessarily compile time constants.  So for me, dig and all others are implemented as property functions.

ddmd/builtin.d
kinke: Its signature uses `real_t` only.

kinke:
> I'm not sure about this approach of moving things out of Port to it. Seems like an unnecessary and incomplete (missing strtof and strtod) abstraction, with a couple others that should be agnostic to whatever Target you are supporting (yl2x).

`strtof` and `strtod` are each only used once in the D files, and only to potentially set `errno` to `ERANGE` if the literal is out-of-range. To me, this has nothing to do with the target real type, and the functions in `port.d` should clearly be modified to something like `bool isFloat/DoubleStringLiteralOutOfRange(string)`.
With the separation into 2 modules, the `real_t` stuff is all in a single module, and GDC/LDC/future backends might not have to touch `port.d` anymore, or at least not have to keep `port.d` and `real_t.d` in sync anymore.

ddmd/constfold.d
kinke: Thanks, done.

ibuclaw:
> strtof and strtod are each only used once in the D files, and only to potentially set errno to ERANGE if the literal is out-of-range. To me, this has nothing to do with the target real type

No, it has everything to do with it.  Supporting real_t and strtod is not a target (Target) problem, it is a host (Port) problem.

The host must be able to represent all float, double and real data in the format of the target, but also must parse and run CTFE operations on said data in a way that is consistent regardless of host/target combination too.

The idea of real_t is not to encapsulate real, but all fp types.

kinke:
> The idea of real_t is not to encapsulate real, but all fp types.

Would it be enough to rename the `TargetReal` struct to `TargetFP` and add 2 `bool isFloat32/64LiteralOutOfRange(string)` functions, or would we need `float_t` and `double_t` types as well?

ibuclaw:
> functions, or would we need float_t and double_t types as well?

If it makes up stream happy.  But gdc has been using a one-type-fits-all implementation for at least 8 years.

> I hoped float and double were consistent across platforms in D...

You can't strtof or strtod to be implemented correctly.  And once you have data stored in it's synthetic representation, it's only adding more work to convert it to native.

ibuclaw:
Hope there's no offense taken if I seem to be putting on a fierce argument against.  Upstream DMD and Druntime have a frustrating habit of inventing new wheels to solve problems gdc has never had.  :-)

kinke:
`float` and `double` are very seldomly used by ddmd. They are used as compile-time results of implicit conversions from integers (don't tell me that is host-dependent too! :)) and in `ddmd.target.encodeReal()/decodeReal()`. For encoding, the `real_t` value is cast to a `double` (i.e., customizable by `real_t`) before writing its bytes, and for decoding, the bytes are interpreted as `double` and then promoted to a `real_t`.
Apart from that, there's only a single global `double ddmd.root.stringtable.loadFactor = 0.8` and the dmd internals in `port.d` (to check the literals for out-of-rangeness, to be moved to `real_t.d`).

> Hope there's no offense taken

Absolutely not. Let's strive for a proper solution, where cross-compilers just need to re-implement `ddmd.root.real_t`. LDC will soon enough feature a FP software emulation too (one of the last blockers for proper cross-compilation besides some missing tweaks for the config file) and face the same issues you've had, so your input is very much appreciated.

ibuclaw:
Updated my previous comment to read: You can't trust strtof...

kinke:
I've updated the last commit with `isFloat32/64LiteralOutOfRange()` functions now in `real_t.d`...

smolt:
thanks for working on this.  It is important for iOS, Android.  Just wanted to make sure are aware of #643 that uses llvm target float emulation and string/float conversions at compile time.

ibuclaw:
There are some nice things done here, I may reuse them when rebasing my upstream PR.  I think the only problem I have is with the I'm introduction of TargetFP.  there should be nothing that can't be done with the existing namespaces.  Any new actions can be made a property or method of real_t.

kinke:
Alright, I've gotten rid of the TargetFP struct and now use free-standing functions inside a C++ namespace instead. I've also renamed a few things and fixed an (upstream) bug, as `real` literals were never checked for out-of-rangeness.
The latter bugfix will lead to AppVeyor failures, as there are some x87 literals which cannot be represented by 64-bit reals. Sadly, these errors are also triggered in disabled blocks, e.g., inside `static if (real.mant_dig == 64) // 80-bit reals` blocks, as they are still parsed. So I guess I'll have to undo that fix.

kinke:
Upstream PR: https://github.com/D-Programming-Language/dmd/pull/5471

PR #1305, opened 2016-02-21
Implement --use-shared-defaultlib switch

### LINKED ISSUE #1282: Using/building both static and shared library
Recently I got a bug report for ldc Arch package which indicates that I need to provide a static library after all: https://bugs.archlinux.org/task/48028

And of course removing shared one now isn't something I'd be happy about :) So I tried to package both but can't see any obvious way to configure which one gets used. With dmd default is static and shared one can picked via `-defaultlib=libphobos2.so` - but `-defaultlib=libphobos-ldc2.a` doesn't work. Is it possible with LDC and if yes, how?

Dicebot:
Partially fixes #1282

Allows to force linking to shared version of standard
library if both lphobos2.so and lphobos2.a are present

Sadly, my cmake knowledge is below zero so I can't tweak build to
generate both targets at the same time but I can workaround it by
building LDC twice in package script.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -105,6 +105,14 @@ static cl::opt<std::string> debugLib(
     cl::desc("Debug versions of default libraries (overrides previous)"),
     cl::value_desc("lib1,lib2,..."), cl::ZeroOrMore);
 
+#if LDC_POSIX
+static cl::opt<bool> linkSharedDefaultLib(
+    "link-shared-defaultlib",
+    cl::desc("If both static and shared builds of default/debug libraries are present,"
+             " use the shared one. Ignored if -static is supplied."),
+    cl::ZeroOrMore);
+#endif
+
 static cl::opt<bool> linkDebugLib(
     "link-debuglib",
     cl::desc("Link with libraries specified in -debuglib, not -defaultlib"),
@@ -448,18 +456,33 @@ static void parseCommandLine(int argc, char **argv, Strings &sourceFiles,
   } else {
     // Parse comma-separated default library list.
     std::stringstream libNames(linkDebugLib ? debugLib : defaultLib);
+
+#if LDC_POSIX
+    // link either libphobos-ldc.so or libphobos-ldc.a
+    if (linkSharedDefaultLib && !staticFlag)
+        global.params.linkswitches->push("-Bdynamic");
+    else
+        global.params.linkswitches->push("-Bstatic");
+#endif        
+
     while (libNames.good()) {
       std::string lib;
       std::getline(libNames, lib, ',');
       if (lib.empty()) {
         continue;
       }
-
       char *arg = static_cast<char *>(mem.xmalloc(lib.size() + 3));
+
       strcpy(arg, "-l");
       strcpy(arg + 2, lib.c_str());
       global.params.linkswitches->push(arg);
     }
+
+#if LDC_POSIX
+    // ensure user-supplied libraries are still linked dynamically
+    // by default
+    global.params.linkswitches->push("-Bdynamic");
+#endif
   }
 
   if (global.params.useUnitTests) {


kinke:
This surely doesn't work with the MS linker...

kinke:
The command-line options reference is on MSDN. ;)
I've just mentioned this as that whole approach will probably not work for the MS linker. I guess we'll have to use a suffix for the shared Windows libs.
What also doesn't work for the MS linker is simply specifying additional lib directories for multilib builds - the linker is extremely dumb and doesn't check the architecture, it just uses the first lib it finds, so the order of the lib directories is crucial.
We'll have to think this through carefully and probably come up with an extended config file...

JohanEngelen:
For me, it would be OK to add this without Windows support for now. It'd be a shame if this is lost and perhaps someone in the future will add Windows support. It'd be good to `#ifdef` the cmdline option out of Windows builds though.

JohanEngelen:
@Dicebot I'll see if I can spend some time in updating the CMake scrips for this, so we can automate testing of this PR

JohanEngelen:
@Dicebot Can you rebase onto master? (There is a problem with Travis+LLVM38 that makes all Linux tests fail)

JohanEngelen:
> btw travis build seems to fail with ld: unknown option: -Bstatic, any idea about it?

`ld` on Mac is different from GNU ld and does not know `-Bstatic`

JohanEngelen:
Perhaps the GNU equivalent option `-static` does the same on Mac.

JohanEngelen:
If it also doesn't work on Mac, I think @kinke is right and it is better solved with an extended config file, with different names for the shared and static libraries so that no special linker flag is needed.

JohanEngelen:
I don't know. :/

kinke:
Iirc, @redstar proposed something in the past. Something like a `default` section and sections for shared/static stdlibs (druntime & phobos)  which override (or specify) the `-defaultlib` and `-debuglib` switches. And let's not forget about multilib (`-m32`/`-m64`).
We may reduce the config overhead by going with a fixed `-debug` suffix for the debug libs (and dumping the `-debuglib` switch). We may also simplify the multilib part by using 2 fixed directories (`lib32` and `lib64`)...

kinke:
Do we really need the flexibility to specify the names of the stdlibs? I'd rather tend towards using a fixed, hardcoded scheme.
A `-debug` filename suffix for the debug variants would be fine by me.
For dynamic vs. static, another `-shared` suffix seems sufficient to me. On POSIX, we'd have `libphobos2-ldc[-debug][-shared].(a|so)`, and on Windows `phobos2-ldc[-debug][-shared].lib`.
 We could use a `lib` For Windows I'd suggest using a `lib` prefix for static ones (static = `libphobos2-ldc[-debug].lib`, dynamic = `phobos2-ldc[-debug].lib`; this resembles the MS scheme a little bit).

kinke:
This in combination with a new commandline switch to select 1 of the 4 stdlib flavours.

kinke:
The 32/64 bit issue on Windows (can't specify both 32-bit and 64-bit lib directories/both lib sets and have the linker ignore incompatible libs) has been solved by an extended config file, allowing to specify the default commandline options on a per triple basis. E.g., the Windows CI multilib config looks like this:
```
// This configuration file uses libconfig.
// See http://www.hyperrealm.com/libconfig/ for syntax details.

// The default group is required
default:
{
    // 'switches' holds array of string that are appends to the command line
    // arguments before they are parsed.
    switches = [
        "-I%%ldcbinarypath%%/../include/d/ldc",
        "-I%%ldcbinarypath%%/../include/d",
        "-L-L%%ldcbinarypath%%/../lib64", 
        "-defaultlib=phobos2-ldc,druntime-ldc",
        "-debuglib=phobos2-ldc-debug,druntime-ldc-debug"
    ];
};

i686-pc-windows-msvc:
{
    // 'switches' holds array of string that are appends to the command line
    // arguments before they are parsed.
    switches = [
        "-I%%ldcbinarypath%%/../include/d/ldc",
        "-I%%ldcbinarypath%%/../include/d",
        "-L-L%%ldcbinarypath%%/../lib32", 
        "-defaultlib=phobos2-ldc,druntime-ldc",
        "-debuglib=phobos2-ldc-debug,druntime-ldc-debug"
    ];
};
```

This could obviously be further extended by pattern-matching/allowing wildcards in the triples, but it's sufficient for now.

kinke:
> How would that interact with existing -debuglib switch?

It could be deprecated (and ignored).

ximion:
@Dicebot 
> Not sure. I wouldn't be surprised if some of more strict distributions like Debian require specific naming style for all packaged libraries but I don't know certainly. Maybe @ximion can advise.

We usually don't require any specific naming (unless of course there are name clashes between two projects). Since D doesn't have an ABI GDC and LDC both obey, we do need different names for all D libraries depending on whether they are compiled with LDC or GDC (which really sucks a lot).
Fortunately, LDC names its Phobos correctly already, so, at least on the library side, there is no issue here.

Issue #1311, opened 2016-02-21
Windows: LDC cannot build itself (DDMD front-end)


JohanEngelen:
On Windows, LDC cannot be used to build itself with DDMD front-end (merge-2.069 branch).

The first problem is a linking problem: see #1310 .
This can be circumvented by removing the function definitions in the C++ header files, and adding `/O2 /Zc:inline` to the MSVC compile flags.

However, then `ldc2` immediately crashes on line `  global.params.color = isConsoleColorSupported();` inside `parseCommandLine(...)`. It is the first call into D code, I believe. So perhaps there is also a C++-D interop issue on windows.

kinke:
> So perhaps there is also a C++-D interop issue on windows.

The dmd-testsuite results would suggest that's basically okay.
But we may run into ABI issues, e.g., dynamic arrays aren't rewritten yet (having a size of 16 bytes, they need to be returned via sret). That's still the old runtime-functions-aren't-rewritten yet (when calling them) issue, which is blocking the ABI fix...

kinke:
#1317 will take care of this. Its first commit (built with DMD) was able to build and link itself and then build the stdlibs (tested yesterday on 32-bit MSVC).

JohanEngelen:
Awesome

kinke:
Well LDC can build itself now for quite some time.

PR #1314, opened 2016-02-25
Windows: Build with both LDC0.17 and DMD on AppVeyor.

JohanEngelen:
Build with LDC on AppVeyor, but also build with DMD because currently the LDC build is broken.
(The LDC is allowed to be broken for a green build)

To fix linking for the MSVC+LDC combination, the implementations of Visitor's virtual methods must be removed from C++ object files.
The current solution (#if's) is not ideal, with the danger of the two branches of the #if going out of sync. But I did not find a better solution.

Please merge if you agree with this partial solution for #1311 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,8 @@ elseif(MSVC)
     string(REGEX REPLACE "(^| )[/-]GR-?( |$)" "\\2" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
     append("/GR- /EHs-c-" CMAKE_CXX_FLAGS)
     append("/D_HAS_EXCEPTIONS=0" CMAKE_CXX_FLAGS)
+    # See github issue #1311
+    append("/O2 /Zc:inline" CMAKE_CXX_FLAGS)
 
     # warning C4018: signed/unsigned mismatch
     # warning C4101: unreferenced local variable
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,11 +16,13 @@ os: Visual Studio 2015
 
 environment:
   matrix:
+    - APPVEYOR_JOB_CONFIG: Debug LDC
     - APPVEYOR_JOB_CONFIG: Debug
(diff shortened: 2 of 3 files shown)


kinke:
Have you tried simply passing `/FORCE:MULTIPLE` to the MS linker, i.e., using `-L/FORCE:MULTIPLE` when linking with LDC? That was my work-around for the failing dmd-testsuite test to get rid of the errors about multiply defined symbols...

JohanEngelen:
Thanks for the suggestion. That may solve the Visitor problem. (/Zc:inline looks like a "make MSVC behave like C++-spec" flag, and less blunt than /force perhaps)

JohanEngelen:
_I'm not even sure whether this is supposed to work._
Yeah, me neither. For my part, the C++ definitions should all be removed. (De-virtualization optimizations...? meh)

JohanEngelen:
Removing the C++ definitions of Visitor breaks linking on Mac OS, because of missing vtable.....

JohanEngelen:
Perhaps with /FORCE we no longer need to change Visitor. Need to test it. (not on Windows machine now)

JohanEngelen:
`/FORCE:MULTIPLE` works but gives so many warnings... I'm opting for `/Zc:inline` + `#if` in `visitor.h`

JohanEngelen:
In part, yes.

kinke:
>  I thought the /force:multiple that kinke added was only for the testsuite.

Yep, it is only for dmd-testsuite, as there are multiple cpp+d test pairs requiring it.

rainers:
The duplicate symbols in cppa.d/cppb.cpp were just copy and paste errors and have been removed updstream. I suspect it is the same for visitor.h. @yebblies ?

JohanEngelen:
Removing the `visitor.h` definitions helps on Windows, but broke the build on Mac last time I tried (issue with vtable definitions).

yebblies:
It's possible there are lurking vtable emission problems when none of the virtual functions have bodies.

Issue #1312, opened 2016-02-21
extern(C++) namespaces not codegen'd


kinke:
Just noticed this while checking the Travis results of my recent upstream PR - dmd and GDC compile it, LDC fails: https://travis-ci.org/D-Programming-Language/dmd/jobs/110808138

kinke:
Well I think it might not have anything to do with `real/long double`, but with `extern(C++, namespace)` instead (`namespace` ignored).

kinke:
`cpp.cpp`:
```C++
#include <stdio.h>

namespace ns {

struct S {
    int *a = nullptr;

    S() { printf("S::ctor()\n"); }
    S(int a) : a(new int(a)) { printf("S::ctor(int)\n"); }
    S(const S &r) : a(r.a ? new int(*r.a) : nullptr) { printf("S::ctor(const S&)\n"); }
    S(S &&r) : a(r.a) { printf("S::ctor(S&&)\n"); r.a = nullptr; }
    ~S() { printf("S::dtor()\n"); delete a; }

    S &operator=(const S &r) { printf("S::operator=(const S&)\n"); delete a; a = r.a ? new int(*r.a) : nullptr; return *this; }
    S &operator=(S &&r) { printf("S::operator=(S&&)\n"); delete a; a = r.a; r.a = nullptr; return *this; }

    int set(int value) { if (!a) a = new int(value); else *a = value; }

    static S make(int a) { return S(a); }
};

struct Wrapper {
    S s;

    Wrapper() = default;
    Wrapper(int a) : s(a) {}
};

int valFunction(int a) { return a; }
int constRefFunction(const int &a) { return a; }
int mutableRefFunction(int &a) { return a; }
int constPointerFunction(const int *a) { return *a; }
int mutablePointerFunction(int *a) { return *a; }

S makeS(int a) { return S::make(a); }

} // namespace
```

`d.d`:
```D
import core.stdc.stdio;

extern (C++, ns) {

struct S {
    int* a;

    //this();
    this(int a);
    this(this) { printf("S.postblit()\n"); if (a) a = new int(*a); }
    ~this();

    // no explicit opAssign

    int set(int value);

    static S make(int a);
};

struct Wrapper {
    S s;

    //this();
    this(int a);
};

int valFunction(int a);
int constRefFunction(const ref int a);
int mutableRefFunction(ref int a);
int constPointerFunction(const(int)* a);
int mutablePointerFunction(int* a);

S makeS(int a);

}

[...]

PR #1317, opened 2016-02-26
[WIP] real_t for cross-compilation

kinke:


kinke:
Currently doesn't link due to https://github.com/ldc-developers/ldc/issues/1312.

kinke:
Nope, doesn't seem that way. We may opt to use the custom `real_t` only if cross-compilability is desired for the LDC build.

ddmd/tokens.h
```diff
@@ -205,7 +205,7 @@ struct Token
         d_uns64 uns64value;
 
         // Floats
-        d_float80 float80value;
+        real_t floatvalue;
```
kinke: A custom `real_t` with destructor is an issue here inside a union. We may just let it sit outside the union as separate field instead.

kinke:
The linking errors are quite interesting, from the Travis 3.8 job:
```
/home/travis/build/ldc-developers/ldc/ddmd/complex.d:(.text._D4ddmd7complex9complex_t8opEqualsMFS4ddmd7complex9complex_tZi+0x71): undefined reference to `_D4ddmd4root6real_t3ldc6real_t10__postblitMFZv'
/home/travis/build/ldc-developers/ldc/ddmd/complex.d:(.text._D4ddmd7complex9complex_t8opEqualsMFS4ddmd7complex9complex_tZi+0x90): undefined reference to `ldc::real_t::opEquals(ldc::real_t) const'
/home/travis/build/ldc-developers/ldc/ddmd/complex.d:(.text._D4ddmd7complex9complex_t8opEqualsMFS4ddmd7complex9complex_tZi+0xac): undefined reference to `_D4ddmd4root6real_t3ldc6real_t6__dtorMFZv'
/home/travis/build/ldc-developers/ldc/ddmd/complex.d:(.text._D4ddmd7complex9complex_t8opEqualsMFS4ddmd7complex9complex_tZi+0xb4): undefined reference to `_D4ddmd4root6real_t3ldc6real_t6__dtorMFZv'
lib/libldcD.a(complex.d.o): In function `_D4ddmd7complex9complex_t8opAssignMFNcNjS4ddmd7complex9complex_tZS4ddmd7complex9complex_t':
/home/travis/build/ldc-developers/ldc/ddmd/complex.d:(.text._D4ddmd7complex9complex_t8opAssignMFNcNjS4ddmd7complex9complex_tZS4ddmd7complex9complex_t+0x41): undefined reference to `_D4ddmd4root6real_t3ldc6real_t6__dtorMFZv'
```
We seem to have a mix of C++ and D mangling here, should be C++ only. Auto-generated `__dtor` and `__postblit` seem to make up for most of the undefined references.

yebblies:
Special functions don't work across the boundary in most cases - or at least they don't in dmd.  What does work is implementing them in both languages and forwarding to common code.

kinke:
Yep, that's exactly what I did. D postblit and dtor forward to C++ functions which aren't necessary for C++ itself (default dtor, copy/move ctor etc.).

ddmd/expression.d
```diff
@@ -1931,7 +1930,10 @@ private:
         char[__traits(classInstanceSize, IndexExp)] indexexp;
         char[__traits(classInstanceSize, SliceExp)] sliceexp;
         // Ensure that the union is suitably aligned.
-        real for_alignment_only;
+      version(IN_LLVM)
+        align(/*real_t.alignof*/ 8) byte[real_t.sizeof] for_alignment_only;
```
kinke: This really ought to work, @smolt would have had another use-case already. This stuff already works in C++11 with `alignas(alignof(real_t))`, so it obviously needs to work in D too. ;)

smolt: you mean using `align(real_t.alignof)`?  yes, I tried that and expected it to work.

kinke:
This now links on my Win64 box with LLVM 3.8. There are immediate crashes in `real_t.opAssign(real_t)` though when trying to build the stdlibs...

kinke:
Looking much better now on Win64, just hits some assertions when trying to parse some decimal string literals in the stdlibs...

kinke:
Now seems to work properly on my Win64 box (based on the unittests - single failure remains `core.thread` [release only]). Kind of unexpected, I must admit. ;)

kinke:
Supports LLVM 3.5+ now. Linking works for
* Win64 + DMD (AppVeyor)
* Linux x64 + DMD (Travis)

Using LDC to compile the front-end leads to the linking errors, see #1312.
On OSX with DMD, there are only 2 undefined references involving D `long` and `ulong` args (C++ `long` vs. `long long` or something like that?).

If linking works, we crash very soon for x87 target reals. That's due to `llvm::APFloat` printing infinity as `infinity` and the C runtime expecting `inf` and stuff like that. I currently use string representations to convert the `APFloat` in target precision to a C++ host `long double` and vice-versa, in order to offer simple and fast builtin math. We'll probably need a library like MPFR for proper compile-time math in any precision. That precision doesn't necessarily have to be the target precision; GDC uses a single big type for all host/target combinations.

kinke:
Squashed, somewhat refactored, some bugfixes (I've previously rounded when converting to integers instead of truncating, and forgot to `safeInit()` some no-properly initialized `real_t`s. This may actually solve the lifetime issues - let's see how the DMD Travis jobs fares.

gen/real_t.cpp
```diff
@@ -0,0 +1,198 @@
...
+    case llvm::Triple::x86:
+    case llvm::Triple::x86_64:
+      semantics = &APFloat::x87DoubleExtended;
+      break;
+    case llvm::Triple::ppc:
+    case llvm::Triple::ppc64:
+    case llvm::Triple::ppc64le:
+      semantics = &APFloat::PPCDoubleDouble;
+      break;
+    default:
```
redstar: AArch64 (`llvm::Triple:aarch64`/`llvm::Triple:aarch64_be`) uses `APFloat::IEEEquad`.

joakim-noah:
Nice work.  I'm using slightly updated versions of this PR with the ddmd 2.070 and 2.071 builds for Android/ARM, and it works well.

kinke:
Thanks @joakim-noah. I've rebased it and added 128-bit semantics for AArch64 (thanks @redstar).

The current state seems unchanged: older LDC builds used to compile the front-end still exhibit the mangling bug (missing C++ namespace), hence the linking issues. When using any semantic > 64 bits, for which LLVM uses dynamic memory, we run into lifetime issues which need to be debugged at some point.

joakim-noah:
Latest master ldc built with this PR now segfaults when cross-compiling `std.numeric` for Android/ARM, though it works fine with the 1.0.0 release.  Here's the last lines of the resulting backtrace:
```
[28%] Generating std/numeric.o
*** Error in `ldc/build71/bin/ldc2': double free or corruption (fasttop): 0x00000000028d8d60 ***
#0 0x00000000015dcff8 llvm::sys::PrintStackTrace(llvm::raw_ostream&) (ldc/build71/bin/ldc2+0x15dcff8)
#1 0x00000000015dba66 llvm::sys::RunSignalHandlers() (ldc/build71/bin/ldc2+0x15dba66)
#2 0x00000000015dd615 SignalHandler(int) (ldc/build71/bin/ldc2+0x15dd615)
#3 0x00007f84334c9f00 __restore_rt (/usr/lib/libpthread.so.0+0x10f00)
#4 0x00007f8432684295 __GI_raise (/usr/lib/libc.so.6+0x33295)
#5 0x00007f84326856da __GI_abort (/usr/lib/libc.so.6+0x346da)
#6 0x00007f84326bfd50 __libc_message (/usr/lib/libc.so.6+0x6ed50)
#7 0x00007f84326c5546 malloc_printerr (/usr/lib/libc.so.6+0x74546)
#8 0x00007f84326c5d1e _int_free (/usr/lib/libc.so.6+0x74d1e)
#9 0x0000000001560a98 llvm::APFloat::operator=(llvm::APFloat&&) (ldc/build71/bin/ldc2+0x1560a98)
#10 0x00000000006dc1b1 _D4ddmd4root6real_t3ldc6real_t8opAssignMFS4ddmd4root6real_t3ldc6real_tZv (ldc/build71/bin/ldc2+0x6dc1b1)
#11 0x000000000081fbc3 _D4ddmd5lexer5Lexer6inrealMFPS4ddmd6tokens5TokenZE4ddmd6tokens3TOK (ldc/build71/bin/ldc2+0x81fbc3)
#12 0x000000000081f4ac _D4ddmd5lexer5Lexer6numberMFPS4ddmd6tokens5TokenZE4ddmd6tokens3TOK (ldc/build71/bin/ldc2+0x81f4ac)
```
Let me know if you need more info.

kinke:
These are the lifetime issues I mentioned. They should only occur when using an FP semantic > 64 bits though, since `llvm::APFloat` uses dynamic memory only then. So I'd firstly check if it's using 64-bit doubles or something else (128-bit for AArch64, introduced in the last commit?).

joakim-noah:
Never mind, turns out the problem is that I was using my modified version of this PR from a month ago, that had worked with master at that time.  However, when applied to recent master it produced this error, but simply taking your current patch and making one modification for latest HEAD makes it work again.

I had compared `ddmd.root.real_t` and some of the other changes to make sure they hadn't changed, but I guess something else changed in your updated PR to make this work.  Ignore my last comment.

kinke:
Rebased.

kinke:
The upstream PR has finally been merged! :)

joakim-noah:
Where is this at?  Would be nice to get this and #1340 in, so I don't have to keep updating them for the latest Android cross-compiler builds.

kinke:
This depends on the next front-end (the upstream PR is included at least in 2.072.1), and then the lifetime issues need to be ironed out when using a precision > 64 bits.
The Travis runs for #1340 don't look as if it's working yet for `ltsmaster` (math-related failures for x87 reals).

kinke:
See https://github.com/ldc-developers/ldc/pull/1987#issuecomment-279536293 for context on why this was abandoned.

PR #1318, opened 2016-02-27
NetBSD fix: exclude libdl

nrTQgc:
NetBSD has not libdl. Dynamic link interface (dlopen, dlclose, ...) is part of libc.


JohanEngelen:
Thanks for the PR!
With this change, does the build succeed?

JohanEngelen:
Could you also add "OpenBSD", "FreeBSD" and "DragonFly" to that list?
(for convenience, I see that CMake accepts a regexp as MATCH string)

(Not to discourage you but I think you should know: in the near future, we want to switch to linking with the D compiler, and so we will no longer need the list of linker flags there.)

nrTQgc:
Use matching: .*BSD (+DragonFly) for excluding

PR #1319, opened 2016-02-28
Change initialization of IrDsymbol.irData.

redstar:
This fixes an assertion failure on Linux/AArch64 (Ubuntu) and on
Linux/PPC64 (RedHat).

### DIFF
--- a/ir/irdsymbol.cpp
+++ b/ir/irdsymbol.cpp
@@ -23,7 +23,9 @@ void IrDsymbol::resetAll() {
   }
 }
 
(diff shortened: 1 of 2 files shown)


kinke:
That's a C++ compiler bug, right?

redstar:
@kinke I have not investigated it deeply but I guess it is kind of compiler bug (or implementation dependent). E.g. on Linux/PPC64: I encountered it on a RedHat based installation but not on a Ubuntu based installation. The compiler versions differ there.

PR #1321, opened 2016-02-28
Run semantic on Type::tvalist.

redstar:
This is required to resolve the type for AAPCS/AAPCS64.
Up to now it was not necessary because only char* was used.
AAPCS/AAPCS64 use a special extern(C++) struct __va_list.

### DIFF
--- a/dmd2/func.c
+++ b/dmd2/func.c
@@ -1497,7 +1497,11 @@ void FuncDeclaration::semantic3(Scope *sc)
             if (f->linkage == LINKd || (f->parameters && Parameter::dim(f->parameters)))
             {
                 // Declare _argptr
+#if IN_LVVM
+                Type *t = Type::tvalist->semantic(loc, sc);
+#else
                 Type *t = Type::tvalist;
+#endif
                 argptr = new VarDeclaration(Loc(), t, Id::_argptr, NULL);
                 argptr->storage_class |= STCtemp;
                 argptr->semantic(sc2);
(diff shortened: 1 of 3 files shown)


smolt:
Thanks.  I was trying to figure out how to do that last week.  Trying it out locally for AAPCS, seems to work for runnable/cppa.d test.

Having std namespace show up in object causes an error in phobos std.zip
```
[ 67%] Generating std/zip-debug.o
/ldc/runtime/phobos/std/zip.d(420): Error: no property 'zlib' for type 'void'
/ldc/runtime/phobos/std/zip.d(429): Error: no property 'zlib' for type 'void'
/ldc/runtime/phobos/std/zip.d(783): Error: no property 'zlib' for type 'void'
```

where line 420 is:
```D
de._compressedData = cast(ubyte[])std.zlib.compress(cast(void[])de._expandedData);
```

I haven't played much with D C++ namespaces, don't know if this can be handled better by the compiler.

smolt:
I found one way to hide C++ std namespace is to put __va_list definition in separate module that normally is not imported by anybody, say ldc.internal.varargs?  object.d does `import ldc.internal.varargs : __va_list` and nobody gets polluted with C++ std namespace.

redstar:
Thanks! I try it!

redstar:
See http://wiki.dlang.org/DIP61 for namespace implementation.

redstar:
@klickverbot Do you know a better place the `resolve()`? The logical place would be after load of `object.d` but it is not really obvious what is happen in the frontend...

PR #1325, opened 2016-02-29
WIP - Add Objective C support

smolt:
Far, far, far from a real pull request, but I though I'd get this up so others start helping.

I had more complete changes last Summer using the full objc D implementation, but those were with a very different LDC code based and need to redo most of it anyway.  Here I hacked in the relevant pieces to get the objc tests to pass.


JohanEngelen:
These two switches need LINKobjc paths: 
https://github.com/ldc-developers/ldc/blob/master/gen/abi-x86.cpp#L39
https://github.com/ldc-developers/ldc/blob/master/gen/abi-x86.cpp#L57

jacob-carlborg:
* I see that `L_OBJC_LABEL_CLASS_$` is missing
* Does LLVM handle prefixing symbols with `L_` and suffixing with an incrementing number?

smolt:
> These two switches need LINKobjc paths

thanks

> I see that L_OBJC_LABEL_CLASS_$ is missing

I looked at clang and it appears that is only emitted with a list of implemented classes.  I did not see it used where there was just an @interface declaration.

> Does LLVM handle prefixing symbols with L_ and suffixing with an incrementing number?

Yes, it does that for PrivateLinkage based on OS being targeted.

smolt:
When I have time tomorrow, I'll create an branch here so anybody can work on it.  It is probably easier for you all to just improve it without having me update a PR.  Or if anybody else wants to start the branch, go for it.  I have a pretty full schedule this week.

jacob-carlborg:
> I looked at clang and it appears that is only emitted with a list of implemented classes. I did not see it used where there was just an `@interface` declaration.

Looks like you're right, my bad.

BTW, the implementation for x86 32bit is completely different, if someone cares. Don't think anyone cares about PPC. ARM should have the same implementation. Although it might have different usage of the `objc_msgSend` functions.

JohanEngelen:
@smolt To start a branch that we can contribute to, I think you can just push this one to
https://github.com/ldc-developers/ldc.git , close this PR and open a new one with the ldc branch.

smolt:
If I did it right, objc-wip branch can have PRs made against it.  Have at it.

My local build passes the extra objc tests in dmd-testsuite.  Going to close this now.

PR #1334, opened 2016-03-03
More (or less) Objective-C Stuff

smolt:
Experiment with letting LLVM generate image info based on module flags.  Probably the way to go.

I am not sure if doing pull requests or pushing directly to objc-wip branch is best.  I imagine incremental improvements over the next week or so.

Are we supposed to make -m32 x86 work too?  I noticed test suite Makefile only enables objc tests for 64-bit.

edit:  dmd -m32 doesn't make a working objc call.  In ldc, I got it to work by using the fragile ABI section names with -m32.  Is there any reason not to support x86?

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -289,8 +289,12 @@ llvm::FunctionType *DtoFunctionType(FuncDeclaration *fdecl) {
     dnest = Type::tvoid->pointerTo();
   }
 
-  if (fdecl->objc.selector) {
+  if (fdecl->linkage == LINKobjc && dthis) {
+    if (fdecl->objc.selector) {
       hasSel = true;
+    } else if (ClassDeclaration *cd = fdecl->parent->isClassDeclaration()) {
+      fdecl->error("Objective-C @selector is missing");
+    }
   }
 
   LLFunctionType *functype = DtoFunctionType(
--- a/gen/objcgen.cpp
+++ b/gen/objcgen.cpp
@@ -1,7 +1,18 @@
+//===-- objcgen.cpp -------------------------------------------------------===//
+//
+//                         LDC – the LLVM D compiler
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+
 #include "mtype.h"
 #include "objc.h"
 #include "gen/irstate.h"
 #include "gen/objcgen.h"
+
+// Currently generates symbols for the non-fragile ABI
 
 namespace {
 bool hasSymbols;
@@ -37,11 +48,23 @@ void retainSymbols() {
 }
 
 void genImageInfo() {
-  // Note: this would normally be produced by LLMV
+  // First option uses LLMV to produce image info ala module flags and
   // TargetLoweringObjectFileMachO::emitModuleFlags()
-  // if module flags are added ala CGObjCMac.cpp
-  // Check into it and then can eliminate this func
+#if 1
+  // Use LLVM to generate image info
+  const char *section = "__DATA,__objc_imageinfo,regular,no_dead_strip";
 
+  gIR->module.addModuleFlag(llvm::Module::Error,
+                            "Objective-C Version", 2); //  unused?
+  gIR->module.addModuleFlag(llvm::Module::Error,
+                            "Objective-C Image Info Version", 0u); // version
+  gIR->module.addModuleFlag(llvm::Module::Error,
+                            "Objective-C Image Info Section",
+                            llvm::MDString::get(gIR->context(), section));
+  gIR->module.addModuleFlag(llvm::Module::Override,
+                            "Objective-C Garbage Collection", 0u); // flags
+#else
+  // Second option just does it directly
   LLType* intType = LLType::getInt32Ty(gIR->context());
   LLConstant* values[2] = {
     LLConstantInt::get(intType, 0),  // version
@@ -56,6 +79,7 @@ void genImageInfo() {
      "OBJC_IMAGE_INFO");
   var->setSection("__DATA,__objc_imageinfo,regular,no_dead_strip");
   use(var);
+#endif
 }
 
 LLGlobalVariable *getCStringVar(const char *symbol,
--- a/gen/objcgen.cpp
+++ b/gen/objcgen.cpp
@@ -17,6 +17,13 @@
 namespace {
 bool hasSymbols;
 
+enum ABI {
+  fragile = 1,
+  nonFragile = 2
+};
+
+ABI abi = nonFragile;
+
 std::vector<LLConstant *> usedSymbols;
 
 llvm::StringMap<LLGlobalVariable *> methVarNameMap;
@@ -27,6 +34,8 @@ void initSymbols() {
   usedSymbols.clear();
   methVarNameMap.clear();
   methVarRefMap.clear();
+
+  abi = (global.params.targetTriple->isArch64Bit()) ? nonFragile : fragile;
 }
 
 void use(LLConstant *sym)
@@ -52,10 +61,11 @@ void genImageInfo() {
   // TargetLoweringObjectFileMachO::emitModuleFlags()
 #if 1
   // Use LLVM to generate image info
-  const char *section = "__DATA,__objc_imageinfo,regular,no_dead_strip";
-
+  const char *section = (abi == nonFragile ?
+                         "__DATA,__objc_imageinfo,regular,no_dead_strip" :
+                         "__OBJC,__image_info");
   gIR->module.addModuleFlag(llvm::Module::Error,
-                            "Objective-C Version", 2); //  unused?
+                            "Objective-C Version", abi); //  unused?
   gIR->module.addModuleFlag(llvm::Module::Error,
                             "Objective-C Image Info Version", 0u); // version
   gIR->module.addModuleFlag(llvm::Module::Error,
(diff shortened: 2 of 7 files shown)


smolt:
@jacob-carlborg Do you have more extensive tests for objc?

What should happen if `@selector` is not provided?  In your full implementation branch of dmd, it generated a default selector matching the function name if there were no arguments.  Right now it compiles without error but program doesn't work.

What should happen on unsupported targets like Linux?  Right now it gets a compile error like this:
```
objc_call.d(7): Error: undefined identifier 'selector'
```
Maybe it can be better?

jacob-carlborg:
> Is there any reason not to support x86?

It's a dead platform on OS X.

> Do you have more extensive tests for objc?

No, not anymore than what's in the DMD test suite.

> Right now it compiles without error but program doesn't work.

I think in the final implementation you're supposed to be able to have a method implemented in D without an explicit selector.

> Maybe it can be better?

Adding an `else` and a static assert?

smolt:
> It's a dead platform on OS X. ARM 32bit uses the non-fraglie ABI.

I think LDC has some users that target OS X x86.  Is the problem that when more Objective-C features get in to dmd that the cost of maintaining the fragile ABI is too much?  For just method calls its seems simple to support the fragile ABI.  @p0nce, do you target 32-bit?  Do you have an opinion?

If 32-bit OS X is not supported, then it seems a compile error should be generated at first use of `extern (Objective-C)`.  I think this should be true of any target where Objective-C is not yet supported.

> I think in the final implementation you're supposed to be able to have a method implemented in D without an explicit selector.

Should we add an error in LDC if the selector is not defined in the meantime?  Am I thinking right that in this implementation, every function in an Objective-C class should have a selector attribute and it is an error if not?

If anyone has time, it would be good to make a more extensive test of Objective-C calls.

@jacob-carlborg is more Objective-C trickling into dmd in the next version?  I am looking forward to it.

p0nce:
@smolt 
In my domain a majority of OSX users are indeed using 32-bit software. I made on reddit and it appears it's a shrinking but very much existing population.

Now, specifically I don't need or use extern(Objective-C) because I use the Obj-C runtime directly. At this time the support wasn't available in compilers.

Though the fragmentation of features on OSX between 32-bit and 64-bit is a bit worrying (shared libraries, TLS, etc it's a kind of hard to remember what is supported in which compilers), no reason to sweat over extern(Objective-C) for 32-bit x86 at all.

jacob-carlborg:
> I think LDC has some users that target OS X x86. Is the problem that when more Objective-C features get in to dmd that the cost of maintaining the fragile ABI is too much? For just method calls its seems simple to support the fragile ABI

The big difference is the symbols. The non-fragile ABI uses more symbols when referencing Objective-C types, forcing linker errors instead of runtime errors when a framework has not been linked. The original implementation was for 32bit, so it should be available, somewhere.

> If 32-bit OS X is not supported, then it seems a compile error should be generated at first use of extern (Objective-C). I think this should be true of any target where Objective-C is not yet supported.

Yes, I agree. As long as it possible to version out Objective-C declarations. Currently in DMD, this is the error you get when compiling on Linux while using  `extern(Objective-C)`.

```
main.d(3): Error: class main.NSObject Objective-C classes not supported
```

But I noticed you don't get this error message on 32bit OS X.

> Should we add an error in LDC if the selector is not defined in the meantime? Am I thinking right that in this implementation, every function in an Objective-C class should have a selector attribute and it is an error if not?

Yeah, that might be a good idea to do in DMD as well.

> is more Objective-C trickling into dmd in the next version? I am looking forward to it.

Unfortunately no. As long as it's taking this long time to get something merged I'm not very enthusiastic about doing any more work. The native TLS implementation has not even been merged yet.

smolt:
Maybe I wasn't totally clear, but the x86 -m32 path does work for Objective-C with the latest commit in this pull.  It could be enabled in LDC even if dmd does not, just like all the other targets that LDC supports.  I was just wondering if there was a reason to back off and delete this code.

> But I noticed you don't get this error message on 32bit OS X.

I see the code for that error is in objc_stubs.d, which is picked instead of objc.d by make of dmd.  Since LDC is told target at compile time, it needs to fold similar code into objc.d code that checks a flag to see if objc is available.  I tried doing a similar check in LDC glue code layer, but the check happens too late and you get `Error: undefined identifier 'selector'` instead, not as helpful.

>> Is more Objective-C trickling into dmd in the next version? I am looking forward to it.
> Unfortunately no.

Once D for iOS gets out there, and everybody switches to D for iOS dev, the demand will be great for more :smiling_imp:.  In the meantime, it is just fun to write code.

smolt:
I found the reason to support Objective-C -m32 fragile ABI for x86:  the iOS Simulator.

jacob-carlborg:
> Maybe I wasn't totally clear, but the x86 -m32 path does work for Objective-C with the latest commit in this pull

Aha, I see know that you've updated the code.

> I was just wondering if there was a reason to back off and delete this code.

No, no reason that I can think of.

> I see the code for that error is in objc_stubs.d, which is picked instead of objc.d by make of dmd. Since LDC is told target at compile time, it needs to fold similar code into objc.d code that checks a flag to see if objc is available. I tried doing a similar check in LDC glue code layer, but the check happens too late and you get Error: undefined identifier 'selector' instead, not as helpful.

LDC is not using `objc_stubs.d` at all?

> I found the reason to support Objective-C -m32 fragile ABI for x86: the iOS Simulator.

That's a good point. Is it still using 32bit?

smolt:
> LDC is not using objc_stubs.d at all?

No.  It could but LDC and clang make it real easy to cross-compile to anything LLVM knows about via a cmdline option.  Having real objc.d is important to make this work when LDC build host is other than OS X.  Not that it will be used often, but today I can use ldc2 on linux to cross-compile to iOS.

> Is it still using 32bit?

Yeah, the sim for armv6, armv7, and armv7s is 32-bit i386 based.

jacob-carlborg:
> No. It could but LDC and clang make it real easy to cross-compile to anything LLVM knows about via a cmdline option. Having real objc.d is important to make this work when LDC build host is other than OS X. Not that it will be used often, but today I can use ldc2 on linux to cross-compile to iOS.

Yes, right. You can actually cross compile in DMD as well, at least between 32 and 64bit. It might not have been the best idea to use the makefile to decide if the Objective-C support should be available or not. Perhaps this should be fixed in DMD.

BTW, I guess you could include the Objective-C support for all targets because GNUStep is available for non Apple platforms. Not sure how compatible the Apple and GNUStep implementation is though.

gen/functions.cpp
```diff
@@ -289,8 +289,12 @@ llvm::FunctionType *DtoFunctionType(FuncDeclaration *fdecl) {
     dnest = Type::tvoid->pointerTo();
   }
 
-  if (fdecl->objc.selector) {
+  if (fdecl->linkage == LINKobjc && dthis) {
+    if (fdecl->objc.selector) {
       hasSel = true;
+    } else if (ClassDeclaration *cd = fdecl->parent->isClassDeclaration()) {
+      fdecl->error("Objective-C @selector is missing");
```
smolt: This is now a compile time error:
```
extern (Objective-C)
interface Class
{
    final NSObject alloc(); // no @selector!  I will crash when called
}
```

but also make adding a final function an error.  Should it be an error?
```
extern (Objective-C)
interface Class
{
    final void extra() {}
}
```

JohanEngelen:
Edited title _Objection_ haha :D -> _Objective_

kinke:
Pls leave stuff like that to the PR author. I thought the amusing name was fully deliberate.

JohanEngelen:
fixed :) sorry.

smolt:
@kinke @JohanEngelen ha, no it wasn't a deliberate title, but it is funny.  My brain doesn't seem to notice such errors.

smolt:
I think functionally the code is ok.  Needs some cleanup of debug code, alternate experiments.  Also, clang adds some function attributes that have not been looked into yet.

For any of you that worked in tocall.cpp, if you know a better organization for the selector processing code, please fix it.  I'd start by moving all selector logic in one place in `ImplicitArgumentsBuilder`, but having it keep state `sel` as state for later processing seems wrong.

Also, comment on:
- Is name for gen/objcgen.{h,cpp} ok?
- how about naming pattern `objc_foo()` in that compilation unit?
- use of another global.params?
- what to do with changes in ddmd/objc.d.  Should they go upstream or be put in an `IN_LLVM` block?

JohanEngelen:
@smolt Do you think we should add IR testcases for this? For example checking function attributes, checking that cross-compiling works, etc. It is easy to add to the lit-based testsuite, and I am happy to help with it.

yebblies:
I think it makes sense to upstream the run-time selection.

gen/objcgen.cpp
```diff
@@ -37,11 +55,24 @@ void retainSymbols() {
...
+  gIR->module.addModuleFlag(llvm::Module::Error,
+                            "Objective-C Version", abi); //  unused?
+  gIR->module.addModuleFlag(llvm::Module::Error,
+                            "Objective-C Image Info Version", 0u); // version
+  gIR->module.addModuleFlag(llvm::Module::Error,
+                            "Objective-C Image Info Section",
+                            llvm::MDString::get(gIR->context(), section));
+  gIR->module.addModuleFlag(llvm::Module::Override,
+                            "Objective-C Garbage Collection", 0u); // flags
+#else
```
klickverbot: Can this be removed then?

smolt: Yes, the else path using LLVM to gen image info is way to go.

smolt: ... I meant else path can go away!

PR #1347, opened 2016-03-08
Add execinfo library for *BSD systems

nrTQgc:
Need -lexecinfo for runtime/druntime/src/core/sys/*bsd/execinfo.d
E.g see
[backtrace](http://www.freebsd.org/cgi/man.cgi?query=backtrace&apropos=0&sektion=3&manpath=FreeBSD+11-current&format=html)


kinke:
Thanks. Required for NetBSD too, so I guess it's safe to require it for all BSD flavours. @redstar ltsmaster or master?

redstar:
IMHO ltsmaster is better. I merge this.

JohanEngelen:
I don't quite understand. This PR adds the flags to the D compiler command that builds of LDC itself, not to the compiled LDC (so it does not apply to ltsmaster). Is LDC's D source using execinfo and is this needed when building LDC?

Shouldn't this be added to LDC's linker.cpp `linkObjToBinaryGcc(...)` instead?  Or perhaps better is to add the library as a `pragma(lib, ...)` in `execinfo.d`?

nrTQgc:
Some strange issue. I had problem with backtrace call from execinfo module. I believe it was inside some test. But now everything is OK. I suppose problem disappears after `make install` for master LDC, Early I used old LDC for compiling new LDC.

I suppose this pull request should be rejected. Thanks @JohanEngelen

Issue #930, opened 2015-05-10
Debug info: Fixup value out of range


kinke:
Environment: Win64, VS 2015, LDC merge-2.067 (+ my VS15 patches + https://github.com/kinke/druntime/commit/1add4f0d401717acc42d92600f9e85ca7d0efe11), LLVM master (`5de99601367aaf09ba1d17cd51478bde5e1b74fe`).

All druntime debug unittests compile fine and all Phobos ones too, except for `std.string` and only when  using the `-g` switch to generate debug infos. This snippet:

```D
import std.conv, std.string, std.typetuple;

void invoke(alias dg)() { dg(); }

void main()
{
    invoke!({
    foreach (S; TypeTuple!( char[], const( char)[], immutable( char)[],
                           wchar[], const(wchar)[], immutable(wchar)[],
                           dchar[], const(dchar)[], immutable(dchar)[]))
    {
        foreach (T; TypeTuple!( char[], const( char)[], immutable( char)[],
                               wchar[], const(wchar)[], immutable(wchar)[],
                               dchar[], const(dchar)[], immutable(dchar)[]))
        (){
            foreach(R; TypeTuple!(dchar[dchar], const dchar[dchar],
                        immutable dchar[dchar]))
            {
                R tt = ['h' : 'q', 'l' : '5'];
                assert(translate(to!S("hello world"), tt, to!T("r"))
                    == to!S("qe55o wo5d"));
                assert(translate(to!S("hello world"), tt, to!T("helo"))
                    == to!S(" wrd"));
                assert(translate(to!S("hello world"), tt, to!T("q5"))
                    == to!S("qe55o wor5d"));
            }
        }();
    }
    });
}
```
yields the following when using the `-g` switch, and otherwise compiles fine:
```
Assertion failed: isIntN(Size * 8 + 1, Value) && "Value does not fit in the Fixup field", file C:\LDC\llvm\lib\Target\X86\MCTargetDesc\X86AsmBackend.cpp, line 120
 	ldc2.exe!`anonymous namespace'::X86AsmBackend::applyFixup(const llvm::MCFixup & Fixup, char * Data, unsigned int DataSize, unsigned __int64 Value, bool IsPCRel) Line 119	C++
 	ldc2.exe!llvm::MCAssembler::Finish() Line 962	C++
 	ldc2.exe!llvm::MCObjectStreamer::FinishImpl() Line 433	C++
[...]

PR #1348, opened 2016-03-10
MSVC: Limit debug info symbol names to ~64k

kinke:
A limitation of the MS linker/COFF format and enforced by LLVM.
Fixes issue #930.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -105,8 +105,6 @@ build_script:
   - md ninja-ldc
   - cd ninja-ldc
   - cmake -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=c:\projects\ldc-x64 -DLLVM_ROOT_DIR=c:/projects/llvm-x64 -DLIBCONFIG_INCLUDE_DIR=c:/projects/libconfig/lib -DLIBCONFIG_LIBRARY=c:/projects/libconfig/lib/x64/ReleaseStatic/libconfig.lib ..\ldc
-  # Work around LDC issue #930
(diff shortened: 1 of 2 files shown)


rainers:
LGTM. As this affects only the "pretty" name displayed during debugging, not the actual symbol name, I'd be fine with cutting it down even fürther...

JohanEngelen:
Would it make sense to turn this into a commandline option? (And set it per default to 64k on Windows)

Does this affect compilation speed? (if you set it to, say, 20 characters)

kinke:
> Would it make sense to turn this into a commandline option?

Imho, not really.

> Does this affect compilation speed?

It's firstly only relevant when generating debug infos. And I don't assume it has any significant impact except maybe for a bit more RAM usage and debug info file size.
I guess it's fine the way it is, and such huge display names should be extremely rare. The problem for #930 is an anonymous function whose body consisting of a 3-level loop is unrolled at compile-time, giving 9x9x3=243 innermost loop iterations...

Merging as that bug has been bugging me for way too long, and now with improved LLVM assertions the root cause could finally be tracked down.

JohanEngelen:
@kinke I asked out of interest, not to further complicate this :)   The Weka guys told me that they have some >100kB symbol names :S 

Btw, Gitter says: "kinke is not on Gitter yet." How come? Come join us! ;-)

kinke:
> "kinke is not on Gitter yet."

Truth be told, I think LDC is way too addictive already. Instant messaging would make matters worse for me. ;)
I also prefer to keep communication overhead to a minimum and have it documented for future references. I think there's a vast and very useful pool of in-depth knowledge spread around GitHub PRs, issues etc. which Google should be able to index...

smolt:
I use Gitter as a way to ask general LDC questions, not necessarily related to any particular PR or Issue (like this discussion).  I rarely see anybody else on it, but if someone does drop in, sometimes they can get something resolved quickly by quick question answer.  I usually just wait until email reminds me there are new message, then check.  It is pretty low bandwidth.

PR #1349, opened 2016-03-11
Use build host D real type [was ARM/AArch64 updates to master (support 64-bit/128-bit real type)]

smolt:
master builds on ARM and phobos hello world works.  Now need to get ARM tests working, starting by putting 64-bit real support back in root/port.d.

Note there is more stuff in previous port.c C++ version that needs to move up to D version, such as supporting big endian.

edit: this has morphed into something simpler - just use real type of the build host D compiler.  It is needed on all the non-x86 hosts until real type cross-compiling is supported.

### DIFF
--- a/ddmd/root/port.d
+++ b/ddmd/root/port.d
@@ -84,20 +84,6 @@ extern (C++) struct Port
         static __gshared bool yl2xp1_supported = false;
     }
     static __gshared real snan;
-    static this()
-    {
-        /*
-         * Use a payload which is different from the machine NaN,
-         * so that uninitialised variables can be
-         * detected even if exceptions are disabled.
-         */
-        ushort* us = cast(ushort*)&snan;
-        us[0] = 0;
-        us[1] = 0;
-        us[2] = 0;
-        us[3] = 0xA000;
-        us[4] = 0x7FFF;
-    }
 
     static bool isNan(double r)
     {
@@ -114,9 +100,11 @@ extern (C++) struct Port
         return a % b;
     }
 
-    static real fequal(real a, real b)
+    static bool fequal(real a, real b)
     {
-        return memcmp(&a, &b, 10) == 0;
+        // don't compare pad bytes in extended precision
+        enum sz = (real.mant_dig == 64) ? 10 : real.sizeof;
+        return memcmp(&a, &b, sz) == 0;
     }
 
     static int memicmp(const char* s1, const char* s2, size_t n)


smolt:
hmm, need another way to break the import cycle.

JohanEngelen:
@smolt  For endianness, I recently found `llvm/Support/Endian.h` that might be of help.

yebblies:
Maybe it would be better to move to an explicit init function?

smolt:
> Maybe it would be better to move to an explicit init function?

Could you outline what you are thinking?

kinke:
That little fix for 64-bit front-end reals is fine by me. It's all transitional until we have a proper solution.

ddmd/root/port.d
```diff
@@ -38,6 +38,15 @@ version(CRuntime_Microsoft)
...
 }
 
+// Forward declare these to to avoid cycle
+// import ddmd.target : Target
+version (none) // not working!  hard code below for now
+private extern (C++) struct Target
+{
+    extern static __gshared int realsize;
+    extern static __gshared int realpad;
+}
```
kinke: Not needed, right? Using those was my first approach too... :)

yebblies:
> Could you outline what you are thinking?

If module constructors are causing problems with import cycles, they can be replaced with a regular function that gets called from main.

smolt:
Hey - good news!  Only this change needed for ddmd on ARM and a `version(ARM)` in phobos `std.experimental.allocator.building_blocks.region`.  A couple test failures but they are the two outstanding and understood ARM test failures with ltsmaster: `std.internal.math.errorfunction` and runnable/constfold.d.

For fun I will try another round trip and see if this ldc2 can also compile master.

> If module constructors are causing problems with import cycles, they can be replaced with a regular function that gets called from main

Ok.

smolt:
I just got rid of the Port's `static this()` to break the cycle, replacing it with -- nothing.  This strange loops way of thinking is probably flawed but why initialize `snan` when the D compiler compiling port.d already knows its value and will set it.

smolt:
Using `Target.realsize` for `fequal()` doesn't work on windows because it is the compiler's `real` type, not target's, that matters here.  dmd `real` size is 10, but target size is 8.  Since this is a transitional change, I think I will dump most of this and go for a minimal change to support 64-bit `real`.

kinke:
> I think I will dump most of this and go for a minimal change to support 64-bit real.

Sounds good. That would also be needed when building LDC for Windows and choosing LDC to compile the front-end (64-bit reals too).

kinke:
I've just cancelled the AppVeyor jobs for your 2nd last commit (I'd like to get some feedback wrt. the introduction of 32-bit AppVeyor jobs before going to bed :grin:).
```printf("real.sizeof %zu realsize %d realpad %d \n", real.sizeof, Target.realsize, Target.realpad);```
Just in case: `real.sizeof` would be 16 (64-bit front-end compiled by DMD) or 8 (compiled by LDC), and `Target.realsize` would always be 8 (and no padding obviously). For MSVC targets that is. We'd still use 80-bit reals for MinGW targets.

**Comment by smolt** [comment]
Perfectly fine.

**Comment by smolt** [comment]
> For fun I will try another round trip and see if this ldc2 can also compile master.

LDC compiles itself fine!  I cleaned up all the experiments and did minimum for ARM little endian 64-bit reals.  Targeting a different real type than the build compiler is left as an exercise for the reader.

**Comment by redstar** [comment]
Maybe you could also add an else clause for 128-bit reals? This helps on AArch64.

**Comment by smolt** [comment]
You timing is ------


------ perfect!

I was just making an upstream PR.  The bit pattern is in the old ports.c?  I'll look.

**Comment by smolt** [comment]
@redstar - take a look, I don't have AArch64 test environment with 128-bit real, so just went by wikipedia for format (https://en.wikipedia.org/wiki/Quadruple-precision_floating-point_format).  If figured the `snan` would be similar as other cases: a quiet nan with MSB of payload set.  If you like, I'll add it to the upstream PR.

**Comment by kinke** [comment]
> Have you submitted this upstream? This should be uncontroversial, and I'd be happy to merge it.

As I already mentioned, this is all transitional. A proper way of tackling this in a somewhat greater context is https://github.com/D-Programming-Language/dmd/pull/5471 and implementing a custom `real_t` with accompanying `CTFloat` helper.
As soon as D <-> C++ works (#1312 + some lifetime issues), we can merge an elaborated #1317 and then begin using an arbitrary precision library like MPFR for compile-time math in `CTFloat`. That will render LDC decoupled from the `real` type used by the D compiler for compiling the front-end, AND fully solve the floating-point cross-compilability issues.

**Comment by smolt** [comment]
@kinke - I had just typed something like that up as a reason not to do this change upstream, but hesitated before clicking "comment".  Decided to just do it, then tacked on Kai's aarch64 stuff.  I could easily close it.  One thing is for sure, something needs to be in 1.0.0 so ARM doesn't regress.

@redstar - I re-read IEEE 754 NaN formats and it looks to me like 80-bit extended `snan` is a signalling NaN (makes sense by the name), but the 64-bit double `snan` pattern I snagged from port.d is a quiet NaN.  I did the same quiet NaN payload for 128-bit quad.  Was this an error in previous port.c?

**Comment by kinke** [comment]
I think the 64-bit `snan` should be DMD's `double.init`. This `snan` represents the init value for compile-time reals.

**Comment by smolt** [comment]
> I think the 64-bit snan should be DMD's double.init. This snan represents the init value for compile-time reals.

It actually works to just declare `real snan;` and use it because the D compiler already has it builtin and real is already 64-bits (not cross-compiling, the compiler is already hosted on ARM).  Even the original `snan` initialization before this PR is not needed since you have a D compiler that already knows it.  I had that for while, then I thought the `snan` pattern becomes lost from the source code, too much mystery.

**Comment by kinke** [comment]
> I had that for while

Of course, now I remember. Yeah, that works if we trust the D compiler's bitpattern, which sounds feasible to me.

**Comment by redstar** [comment]
It is like the 80-bit case with four additional `0`:

            else static if (real.mant_dig == 113)
            {
                us[0] = 0;
                us[1] = 0;
                us[2] = 0;
                us[3] = 0;
                us[4] = 0;
                us[5] = 0;
                us[6] = 0;
                us[7] = 0xA000;
                us[8] = 0x7FFF;
            }

**Comment by smolt** [comment]
> us[7] = 0xA000;
> us[8] = 0x7FFF;

Ok - updated.

That is still a quiet NaN.  Shouldn't the double 64-bit and quadruple 128-bit cases be a signaling NaN too, like the extended precision 80-bit?

I made this table of +NaN bit patterns based on various Internet sources (set MSB sign-bit for -NaN bit patterns).

Double
- s NaN 7FF0_0000_0000_0001 - 7FF7_FFFF_FFFF_FFFF
- q NaN 7FF8_0000_0000_0000 - 7FFF_FFFF_FFFF_FFFF

Extended
- s NaN 7FFF_8000_0000_0000_0001 - 7FFF_BFFF_ ... _FFFF
- ind   7FFF_C000_0000_0000_0000 (special case of quiet NaN)
- q NaN 7FFF_C000_0000_0000_0001 - 7FFF_FFFF_ ... _FFFF

Quadruple
- s NaN 7FFF_0000_0000_0000_0000_0000_0000_0001 - 7FFF_7FFF_ ... _FFFF
- q NaN 7FFF_8000_0000_0000_0000_0000_0000_0000 - 7FFF_FFFF_ ... _FFFF

**Comment by kinke** [comment]
Seriously guys, is there any reason to discuss bit patterns at this stage? There'll be no LDC (based on master, not on ltsmaster) and no DMD compiler supporting 128-bit reals anyway until we introduce a software `real_t` for LDC (needed to cross-compile ddmd for the desired AArch64 target, before being able to build the 1st native AArch64 LDC with 128-bit reals). So why bother?
I'd simply leave it as `real snan;` for the time being, and leaving the pattern to the D compiler compiling the front-end. LLVM's APFloat obviously has a SNaN for all supported formats, so we'll most likely never need these bitpatterns anymore.

**Comment by redstar** [comment]
@kinke See my native AArch64 buildbot at http://redstarbuildbot.mynetgear.com:12080/waterfall. I am porting 0.17.0 to AArch64.

**Comment by kinke** [comment]
Cool! But why v0.17?

I mean I do see the beauty/simplicity of not having to rely on an existing D compiler. On the other hand, given how close we are to proper cross-compilation, I don't see a huge disadvantage by simply requiring the previous LDC version or something like that to compile the front-end.

For new platforms for which there's no compatible D compiler yet, but which LLVM is able to target, cross-compiling the ddmd modules on a dev box and then copying them over to a native box for C++ compiling and linking, to build the first LDC of its kind, doesn't seem that bad to me. Keeping on using the C++ linker instead of the D one as in Johan's PR would simplify things a bit, but that's only required as long as there's no reasonably working D compiler for the target platform.

**Comment by smolt** [comment]
I'm fine with eliminating the `snan` initialization.  I'll remove it.  Since this is temporary, maybe enclose changes with IN_LLVM version and I then close the upstream PR.  I added another commit to show this.  I'll squash later if this is a good.  I've tested on my ARM, seems good to me.  I mean, LDC compiling LDC compiling LDC on ARM is still pretty neat.

Yes to this concept?
Yes to no upstream PR?

**Comment by redstar** [comment]
For now it is much easier to use the C++ version. As long as the `real` support is not finished it is a constant source of errors. E.g. cross-compiling and running druntime tests on qemu has less errors than on real hardware because the different real type is incorrectly compiled.
But I agree that we should aim for an easier solution than maintaining 0.17 forever.

**Comment by kinke** [comment]
@smolt: I'd be in favour of trying to get the current state of this PR upstream - i.e., no explicit initialization (just dropping the static ctor altogether) and `fequal()` as you implemented it here (but without `version(LDC)` guard obviously).
Btw, how does the `is` operator work? I.e., `static bool fequal(real a, real b) { return a is b; }`...

**Comment by smolt** [comment]
> static bool fequal(real a, real b) { return a is b; }

good idea, seems to me it should work. I'll try it.

File: ddmd/root/port.d [hunk 2]
```diff
@@ -116,6 +117,13 @@ extern (C++) struct Port
 
     static real fequal(real a, real b)
```
smolt: Never noticed until now.  Why return a `real` when used only for boolean result?  I plan to change as @kinke suggested.

kinke: Yep, I fixed that one in #1317 / https://github.com/D-Programming-Language/dmd/pull/5471 too.

kinke:
I've seen the `is` operator being used in `isInfinity()`, but I don't actually know its semantics. ;)
What's also affected by the size of the `real` type is `isSignallingNan()`. IIRC, we don't use it in LDC, but it's used by the dmd backend, so you'll have to fix it for the upstream PR. Maybe `is real.init` works here too.

smolt:
Fails with dmd as compiler 🤔 Why would that be?

kinke:
Yep, very interesting, especially the dmd-testsuite failures. `runnable/constfold.d:321`:
```D
static assert((5.1 is 5.1) == 1);
```
`0 == 1`. So DMD's `is` in `Port.fequal()` seems to differ from LDC's. Even so, why would 2 identical double literals differ? `is` won't compare the addresses (for value types), so does DMD somehow compare the compile-time `real` padding?

smolt:
Does this seem acceptable and something worth pushing upstream as a transitional change until proper real_t is implemented?  @redstar, can you see if it works for quadruple precision?  If so, I will squash it and update the dmd PR.

smolt:
squashed and new upstream PR D-Programming-Language/dmd#5646 created.

smolt:
The upstream PR was merged.  Does it help with future front end merges if I now cherry-pick the dmd commit in place of the commit in this PR?  The deltas are identical.

smolt:
Ah, yeah I forgot ldc and dmd were disjoint.  Most my upstream changes have been in runtime which can be git cherry-picked.

PR #1353, opened 2016-03-12
Remove PassNameParser (may be breaking some command lines!)

redstar:
This PR removes the command line option which adds the pass names as additional
command line options. This code prevents compiling for AArch64 (since LLVM 3.7)
and ARM (since LLVM 3.8). Recommendation on LLVM list was to use the PassNameParser
only in test tools but not in a production compiler.

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -40,12 +40,6 @@
 extern llvm::TargetMachine *gTargetMachine;
 using namespace llvm;
 
-// Allow the user to specify specific optimizations to run.
-static cl::list<const PassInfo *, bool, PassNameParser>
-    passList(cl::desc("Running specific optimizations:"),
-             cl::Hidden // to clean up --help output
-             );
-
 static cl::opt<signed char> optimizeLevel(
     cl::desc("Setting the optimization level:"), cl::ZeroOrMore,
     cl::values(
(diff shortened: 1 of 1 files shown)


redstar:
See http://forum.dlang.org/post/tlziomjsfptuevvpnrjh@forum.dlang.org

redstar:
In general there are command line options which have the same name as the pass. This results in an error during runtime.

smolt:
Once of the switches is `-arm-load-store-opt`.  Somehow llc exposes it to the command line without conflict.

redstar:
@smolt: `-arm-load-store-opt` is a command line option (defined in lib/Target/ARM/ARMTargetMachine.cpp).
But there is a pass of the same name. The `PassNameParser` takes the name of the pass and generates an new option with the pass name at runtime. The offending pass is defined in lib/Target/ARM/ARMLoadStoreOptimizer.cpp.

smolt:
Yeah, I was in that territory a couple months ago: https://github.com/smolt/llvm/commit/958404db5ce77c2843a04e2f6fd8e032b01b394c

I just needed to get work done so temporarily did the rename for the command line options.

But - I noticed that llc didn't seem to have the same problem.  `-arm-load-store-opt` is available as a command line option.

smolt:
BTW - I am fetching your change to try it out.

redstar:
`llc` does not use the `PassNameParser`. The option you see is the one added in ARMTargetMachine.cpp.
`opt` uses the `PassNameParser` but does not add the backend passes. Therefore there is no conflict.

smolt:
Ah, I didn't look carefully at the option lists.  That fills in the missing piece for me.  And your change works fine for me.  My test is building against llvm 3.8, use an arm or aarch64 triple, and specify -O.  It would fail before with something like:
```
$ bin/ldc2 -mtriple=arm-linux-gnueabihf -O -c foo.d
ldc2: CommandLine Error: Option 'arm-load-store-opt' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```
LGTM

joakim-noah:
Verified that this PR fixes the issue on Android/ARM, thanks.

PR #1355, opened 2016-03-13
Implement cent/ucent data types.

redstar:
enabled if compiled with gcc.

### DIFF
--- a/dmd2/builtin.c
+++ b/dmd2/builtin.c
@@ -120,8 +120,12 @@ static inline int getBitsizeOfType(Loc loc, Type *type)
       case Tuns16: return 16;
       case Tint128:
       case Tuns128:
+#if WANT_CENT
+        return 128;
+#else
           error(loc, "cent/ucent not supported");
           break;
+#endif
       default:
           error(loc, "unsupported type");
           break;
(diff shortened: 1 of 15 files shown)


redstar:
This supercedes #1120.

smolt:
would it be better to just implement in master and ddmd front end instead of going into older C++ front end?

redstar:
Yes, for sure. I need this as my template. :-)

deadalnix:
Just planting my flag here. I would love to have cent/ucent available. Is there something I can do to help this move forward ?

redstar:
I currently try to port this PR to the ddmd frontend. This will be good for LDC but code generation for DMD is then still missing.

deadalnix:
Is there a reason not to move forward with in LDC ? Adding backend support for DMD may prove to take some time, and right now, there is incentive to do it as nobody else supports it.

redstar:
The only obstacle right now is amount of free time. :-(
If you like to help: a type supporting 128bit arithmetic (`struct uint128_t`) is required for bootstrapping the compiler. I have not yet implemented such a supporting type - this PR still relies on GCC support.

deadalnix:
Would this be reusable ? https://github.com/d-gamedev-team/gfm/blob/master/integers/gfm/integers/wideint.d

redstar:
I updated this PR to be based on latest ltsmaster branch.
A port to master branch is underway (requiring this PR for bootstrapping).
First upstream patches: https://github.com/dlang/dmd/pull/5984, https://github.com/dlang/druntime/pull/1618

@deadalnix The class looks nice. Thanks for the link.

redstar:
I did some bug fixing. The failure in `runnable/xtest46.d` is a wrong test case (see https://github.com/dlang/dmd/pull/5987 for a fix).
There are four more failures in the `fail_compilation` folder.
Not yet visible is a serious failure in `core.checkedint` and a problem on PPC64.

aliak00:
Ping pong ding dong. This still alive?

JohanEngelen:
Oh man, now I've found a use for `ucent` too...
@redstar I also think it's better to target master instead of ltsmaster. We can ask one of the `wideint` devs to help / make a lighter-weight version to be included in the frontend?

thewilsonator:
@JohanEngelen It should be pretty easy to sit it down now that we have `-vcg-ast`, just instantiate `wideint!128` and `uwideint!128` and pretty it up a bit.

Imperatorn:
Status?

PR #1357, opened 2016-03-16
Change mangling of extern(C++) ctor.

redstar:
This is a part of issue #1312.

### DIFF
--- a/dmd2/cppmangle.c
+++ b/dmd2/cppmangle.c
@@ -257,7 +257,15 @@ class CppMangleVisitor : public Visitor
         }
         else
         {
+#if IN_LLVM
+            // Issue #1312: Mangle constructor with class/struct
+            // name instead of __ctor.
+            const char *name = s->isCtorDeclaration()
+                               ? s->parent->ident->toChars()
+                               : s->ident->toChars();
+#else
             const char *name = s->ident->toChars();
+#endif
             buf.printf("%d%s", strlen(name), name);
         }
     }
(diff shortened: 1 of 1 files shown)


redstar:
@kinke Does this help for ctor mangling?

kinke:
Thx Kai, I'll check it over the weekend.
So this should allow using C++ ctors from D by just declaring them in D. But this doesn't fix the actual mangling bug in #1312.

yebblies:
Without also matching the ABI, matching mangling is a little dangerous.

kinke:
That's what I thought too. You mean the different destruction rules, right? Does the front-end support argument destruction by the caller for `extern(C++)` functions?

yebblies:
I'm not sure.  The issue I've had trouble with is that D will blit `init[]` over before calling the constructor, while C++ won't.  We should (hopefully) be matching argument destruction correctly for normal C++ calls, or is it different for ctors?

kinke:
> D will blit init[] over before calling the constructor, while C++ won't.

I'd think that a C++ ctor cannot make any assumptions about the initial payload, so I'd consider blitting to be useless (and affecting runtime a bit) but not troublesome.

> We should (hopefully) be matching argument destruction correctly for normal C++ calls

That's what I wanted to hear - very good.

yebblies:
> I'd think that a C++ ctor cannot make any assumptions about the initial payload, so I'd consider blitting to be useless (and affecting runtime a bit when calling it from D) but not troublesome.

When calling a ctor written in C++ from D, yes.  But when going the other way the ctor may not init all fields.

kinke:
> But when going the other way the ctor may not init all fields.

True. That could be compensated for when generating an `extern(C++)` ctor implemented in D by having the front-end inject the blitting at the beginning of the ctor (and omitting the pre-init when calling it from D). We could also simply define the semantics for `extern(C++)` ctors implemented in D as fully C++ compatible, i.e., no implicit pre-init.

redstar:
Closing in favor of https://github.com/dlang/dmd/pull/5884.

Issue #1278, opened 2016-02-03
DDMD: Use D compiler for linking


klickverbot:
The current `merge-2.069` build system (thanks to @JohanEngelen for getting up us and running!) uses `g++` for linking directly, and has D standard library names and paths hard-coded in CMakeLists.txt. This is very brittle and already broke on two of my systems with a somewhat custom configuration.

We need to either use the D compiler driver for linking, as the upstream DMD makefile does, or at least extract the linker arguments to use from a `$DC -v` call. The former way is certainly simpler, but we need to translate the LLVM linker flags accordingly.

JohanEngelen:
I think using the D compiler for linking is best.
Translating the LLVM flags should be relatively easy, I think.

JohanEngelen:
I think this might do it for the linker flag translation:
```Cmake
string(REPLACE " -" " -L-" DMD_LLVM_LDFLAGS " ${LLVM_LIBRARIES} ${PTHREAD_LIBS} ${TERMINFO_LIBS} ${LLVM_LDFLAGS}")
string(REPLACE ";" " -L" DMD_LLVM_LDFLAGS "${DMD_LLVM_LDFLAGS}")
string(REPLACE ",-" " -L-" DMD_LLVM_LDFLAGS "${DMD_LLVM_LDFLAGS}")
```

JohanEngelen:
I will no longer work on this. With how DMD functions on Windows wrt. linking, there is no gain over what we currently do. See #1361 for the changes to CMakeScript.txt that made things work on Mac and Linux.

klickverbot:
Isn't the obvious workaround just to add a D main() that forwards to the C++ one?

JohanEngelen:
@klickverbot Yes I tried to change things without touching frontend code. But I think there is no other way.

klickverbot:
This wouldn't touch frontend code?

If there is a specific issue with using DMD to link on Win32, we could just move the `main()` from `driver/main.cpp` to `driver/main.d`, which would just forward to `driver/main.cpp`.

klickverbot:
Also, you might be right that using the C++ compiler to link could end up being less work, but we'd still need to switch to extracting the linker invocation from `dmd -v`. The current solution is an awesome first step, but so fragile that I would be seriously worried about releasing it as part of a non-test version.

JohanEngelen:
doh!  Incredible how dumb one can be sometimes. I blame the ultimate frustration.

Do you want to give it a spin?

[...]

PR #1354, opened 2016-03-12
Enable MSVC EH for Win64

kinke:
Enable @rainers #1168 for Win64 too and add x86 AppVeyor jobs.
Requires https://github.com/ldc-developers/druntime/pull/65 and a post-3.8 LLVM (Rainer's work has uncovered a few LLVM bugs whose near-instant fixes didn't make it into 3.8 final).

PR #1361, opened 2016-03-18
Instead of using the C++-linker, use the D compiler to link LDC2 and LDMD

JohanEngelen:
This removes the need for the CMake logic to figure out what linker flags to pass the C++linker to link D code (50 lines of flaky cmake script).

This works on my Mac. Let's see what Travis and AppVeyor think of it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,67 +116,21 @@ if(CMAKE_BUILD_TYPE MATCHES Debug)
 else()
     append("-O -release" DDMD_DFLAGS)
 endif()
-set(Dcode_LDFLAGS)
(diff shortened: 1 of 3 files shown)


JohanEngelen:
Fixes #1278

kinke:
For AppVeyor, it somehow uses DMD's link.exe, although DMD's bin directory is the last entry in the PATH environment variable. In #1354, DMD isn't in PATH at all anymore.

JohanEngelen:
I will quit the work on this.
I have not found a solution for making this work on Windows. DMD cannot link a D program when "main()" is in C++, because it will not link with phobos/druntime and will not add a few exception handling symbols. It then degrades into the same problem we have now where we have to manually specify these things on the commandline.

PR #1362, opened 2016-03-18
LLVM: Add -lLLVMTableGen if linking against shared LLVM library

redstar:
The problem is that `llvm-config --libs tablegen` returns `-lLLVM-3.8.0` but the
code of the tablegen component is not included in this library.
Current solution is to add `-lLLVMTableGen` to the list of libraries.


redstar:
Ì have not yet checked if this is a known LLVM bug.

redstar:
Merged in ltsmaster

Issue #1310, opened 2016-02-21
Different function linkage type in object file for DMD vs. LDC


JohanEngelen:
This concerns Windows, I did not check the other OS'es.

``` D
extern (C++) class Expression : RootObject
{
public:
//...
    final Expression optimize(int result, bool keepLvalue = false)
    {
        return Expression_optimize(this, result, keepLvalue);
    }
//...
}
```

When compiled with DMD2.069.2, and running dumpbin, the `Expression::optimize` symbol is put in a section with `selection    2 (pick any)`:

```
1EE 00000000 SECTF8 notype       Static       | .text
    Section length   13, #relocs    1, #linenums    0, checksum        0, selection    2 (pick any)
```

When compiled with LDC0.17, running dumpbin, the `Expression::optimize` symbol is put in a section with `selection 1 (pick no duplicates)`:

```
0AE 00000000 SECT3D notype       Static       | .text
    Section length    6, #relocs    1, #linenums    0, checksum        0, selection    1 (pick no duplicates)
```

The difference in linkage properties shows up in trying to build LDC (with DDMD front-end) using LDC as compiler. Because the C++ source contains a second implementation of `Expression::optimize` as inlined member function:

``` C++
class Expression : public RootObject
{
public:
//...
    Expression *optimize(int result, bool keepLvalue = false)
    {
        return Expression_optimize(this, result, keepLvalue);
    }
```

This then results in a linker error (LNK2005) of multiple defined symbols.

I don't know which compiler is correct: DMD or LDC.

If this needs fixing in LDC, LDC can be made to output all functions in `selection    2 (pick any)` sections by putting the functions in comdat sections: 

``` diff
LinkageWithCOMDAT DtoLinkage(Dsymbol *sym) {
  if (DtoIsTemplateInstance(sym)) {
    return LinkageWithCOMDAT(templateLinkage, supportsCOMDAT());
  }
-  return LinkageWithCOMDAT(llvm::GlobalValue::ExternalLinkage, false);
+  return LinkageWithCOMDAT(llvm::GlobalValue::ExternalLinkage, true);
}
```

kinke:
Oh nice that you found the location to be fixed, this is also the reason for a dmd-testsuite failure on Win64. It's just unclear to me as to why that is no problem for all other platforms.

JohanEngelen:
Yeah, probably changing the linkage for all functions is not good and not needed.
[...]

PR #1366, opened 2016-03-20
Linkage type and COMDAT

kinke:
This tackles https://github.com/ldc-developers/ldc/issues/1310.


kinke:
Re-targeting branch `ltsmaster`.

PR #1368, opened 2016-03-20
Instead of using the C++-linker, use the D compiler to link LDC2 and LDMD

JohanEngelen:
This removes the need for the CMake logic to figure out what linker flags to pass the C++linker to link D code (50 lines of flaky cmake script).

Continuation from #1361


driver/ldmd.d
```diff
@@ -0,0 +1,30 @@
...
+//===----------------------------------------------------------------------===//
+//
+// Startup code for driver/ldmd.cpp
+//
+//===----------------------------------------------------------------------===//
+
+// In driver/ldmd.cpp
+extern(C++) int cppmain(int argc, char **argv);
+
+/+ Having a main() in D-source solves a few issues with building/linking with D compiler.
```
klickverbot: Could you just mention DMD/Windows here explicitly?

JohanEngelen:
I need to beat this thing

JohanEngelen:
I am not sure if it was a good idea to compile all D source in one go. It makes compilation significantly slower when you just change one cpp file...

JohanEngelen:
> Regarding .d files, DMD compiles all at once and it seems to work out fine for them.

Yes, but it's just a little slower when only one file changed.
So perhaps in future, someone may want to spend some time to split D compilation from the linking step again.

driver/main.d
```diff
@@ -32,4 +26,22 @@ extern (C++) void writeModuleDependencyFile()
         deps.setbuffer(cast(void*)ob.data, ob.offset);
         deps.write();
     }
-}
\ No newline at end of file
+}
+
+// In driver/main.cpp or driver/ldmd.cpp
```
klickverbot: "or driver/ldmd.cpp?"

JohanEngelen:
I wonder if compiling all D source at once improves performance when built with LDC. LDC becomes much faster (tested on Weka source) when built with `dmd --inline`, but `ldmd2 --inline` results in an LDC that is about 20% slower.

JohanEngelen:
Green on AppVeyor and Travis.
Does this PR also work on everybody's build setup?

JohanEngelen:
Don't merge yet.

make install does not mark ldc2 and ldmd2 as executables. Working on a fix.

JohanEngelen:
Results from a "quick" test on some code that takes a while to compile (CMAKE_BUILD_TYPE=Release):
pre-PR:  91 sec
post-PR:  77 sec
Boom!

JohanEngelen:
Install of executables is fixed.

Please merge!

JohanEngelen:
Hmm the Windows x86 build fails, because it cannot find a pdb file. For x86, DMD uses its own linker which probably cannot generate pdb's?

kinke:
Hmm, I thought DMD supports PDB, but maybe only if the full source is in D, and not for COFF objects produced by the MSVC++ compiler. Anyway, I only include the PDB to enable easier troubleshooting for people experimenting with CI builds. We can also just omit it for the x86 build if we don't find a better solution.

JohanEngelen:
@kinke Could be a missing flag?

JohanEngelen:
I am going to try and reproduce this on Windows, by following the appveyor.yml file for 32bit compilation.

JohanEngelen:
This PR fixes issue #1379

kinke:
@klickverbot: merging this just killed our 32-bit MSVC build. Please don't discard AppVeyor issues.

kinke:
I'll look into a fix. The issue seems to be `std.math.lrint()` only.

kinke:
This snippet demonstrates the problem:
```D
enum MANTISSA_LSB = 0;
enum MANTISSA_MSB = 1;

long lrint(real x)
{
    long result;

    // Rounding limit when casting from real(double) to ulong.
    enum real OF = 4.50359962737049600000E15L;

    uint* vi = cast(uint*)(&x);

    // Find the exponent and sign
    uint msb = vi[MANTISSA_MSB];
    uint lsb = vi[MANTISSA_LSB];
    int exp = ((msb >> 20) & 0x7ff) - 0x3ff;
    int sign = msb >> 31;
    msb &= 0xfffff;
    msb |= 0x100000;

    if (exp < 63)
    {
        if (exp >= 52)
            result = (cast(long) msb << (exp - 20)) | (lsb << (exp - 52));
        else
        {
            // Adjust x and check result.
            real j = sign ? -OF : OF;
            x = (j + x) - j;
            msb = vi[MANTISSA_MSB];
            lsb = vi[MANTISSA_LSB];
            exp = ((msb >> 20) & 0x7ff) - 0x3ff;
            msb &= 0xfffff;
            msb |= 0x100000;

            if (exp < 0)
                result = 0;
            else if (exp < 20)
                result = cast(long) msb >> (20 - exp);
            else if (exp == 20)
                result = cast(long) msb;
            else
                result = (cast(long) msb << (exp - 20)) | (lsb >> (52 - exp));
        }
    }
    else
    {
        // It is left implementation defined when the number is too large.
        return cast(long) x;
    }

    return sign ? -result : result;
}

void main()
{
    import core.stdc.stdio;
    printf("actual = %lld, expected = %lld\n", lrint(int.max - 0.5), 2147483646L);
    printf("actual = %lld, expected = %lld\n", lrint(int.max + 0.5), 2147483648L);
    printf("actual = %lld, expected = %lld\n", lrint(int.min - 0.5), -2147483648L);
    printf("actual = %lld, expected = %lld\n", lrint(int.min + 0.5), -2147483648L);
}
```
=>
```
actual = 2147483648, expected = 2147483646
actual = 2147483648, expected = 2147483648
actual = -2147483648, expected = -2147483648
actual = -2147483648, expected = -2147483648
```

The output is correct if `lrint()` takes a `double` argument instead of a 64-bit `real`.
```

**Comment by kinke** [comment]
I've just compared the IR for above `lrint(real)` vs. `lrint(double)`. The result is very interesting: the constant `int.max - 0.5` results in different double-precision constants: `0x41DFFFFFFFE00000` (real) vs. `0x41DFFFFFFFA00000` (double). This is most likely related to @JohanEngelen's recent Phobos workaround which also involved floating-point literals and mixed types...

**Comment by kinke** [comment]
```
RealExp::toElem: 2.14748e+09L @ real
* RealExp::toConstElem: 2.14748e+09L @ real | 0x1.fffffffe00000p+30
```
vs.
```
RealExp::toElem: 2.14748e+09 @ double
* RealExp::toConstElem: 2.14748e+09 @ double | 0x1.fffffffa00000p+30
```
Apparently `DtoType()` doesn't return a LLVM double-precision type in this case, and the 64-bit compile-time real is loaded as 80-bit x87 value to initialize the LLVM constant...

**Comment by kinke** [comment]
No idea. Maybe we just got lucky and found a lurking bug which hasn't shown up until now. Sadly with that DMD-linked build, I don't have any PDBs so I cannot debug what's going on in LDC. I'll continue investigating.

**Comment by JohanEngelen** [comment]
@kinke So this PR means we lose a lot of debugging convenience on Windows? :(
I think LDC still cannot build itself on Windows, so we could really use the debugging capability...

**Comment by kinke** [comment]
Okay so it's not `DtoConstFP()` - the front-end's compile-time real stored in `RealExp` already diverges. So the front-end somehow computes `int.max - 0.5` differently if it's desired type is a 64-bit `real` (2147483647.5!) and not `double` (2147483646.5, correct). At least when linking LDC with whatever DMD is using for 32-bit COFF objects.

**Comment by kinke** [comment]
Oh wait - when using DMD to compile the front-end, the front-end uses 80-bit compile-time reals. LDC's glue layer for MSVC targets only uses 64-bit `real`s. I'd bet we've missed a conversion somewhere...

**Comment by JohanEngelen** [comment]
> At least when linking LDC with whatever DMD is using for 32-bit COFF objects.

When DMD builds the executable, is there any difference between `-m32` and `-m32mscoff` ?

**Comment by kinke** [comment]
I'm 99.9% sure that this is because we miss most and not just a few ddmd `real` (80-bit when compiling with DMD, 64-bit with LDC) vs. LDC C++ `longdouble` (64-bit for MSVC builds) conversions.
Unfortunately, fixing it would require the first commit of https://github.com/ldc-developers/ldc/pull/1317, a backport of https://github.com/D-Programming-Language/dmd/pull/5471. So we'll have to wait for it to get merged.

**Comment by kinke** [comment]
I'm also convinced this is the reason why the `is` operator didn't work to compare compile-time `real`s for ARM in `Port.fequal()` (@smolt - the front-end probably memcmp()ing 10 bytes instead of 8).

**Comment by kinke** [comment]
> it could/should get rid of the issue

Nope, same issue. :/

**Comment by kinke** [comment]
... but it's at least able to build the front-end without crashing, and the resulting LDC is able to build druntime+Phobos (debug+release). Tested with 32-bit MSVC.

**Comment by kinke** [comment]
All commits of https://github.com/ldc-developers/ldc/pull/1317 (built with DMD) gets rid of this issue, i.e., `int.max - 0.5` for 64-bit target `real` (represented by `llvm::APFloat` based compile-time `real_t` with double-precision semantics) now finally gives 2147483646.5.

**Comment by smolt** [comment]
> Does this PR also work on everybody's build setup?

It seems that runtime is not rebuilt if ldc2 was rebuild.  Did a dependency on ldc2 get dropped?

**Comment by smolt** [comment]
> I'm also convinced this is the reason why the is operator didn't work to compare compile-time reals for ARM in Port.fequal() (@smolt - the front-end probably memcmp()ing 10 bytes instead of 8).

I think that was a different, mysterious reason.  The `is` operator worked fine for ARM 64-bit real.  The failure was on X86 hosts (e.g. OS X) when LDC was compiled by DMD.  LDC compiled by LDC was fine to.

**Comment by JohanEngelen** [comment]
> It seems that runtime is not rebuilt if ldc2 was rebuild. Did a dependency on ldc2 get dropped?

Noticed it too. I checked the dependencies, and they _look_ correct but obviously aren't.

**Comment by smolt** [comment]
Probably totally unrelated but while building LDC master today on Pi to make sure this PR works in that environment, hit a malloc assertion during unittest build:
```
[ 81%] Generating std/range/package-unittest.o                                  
ldc2: malloc.c:2905: __libc_malloc: Assertion `!victim || ((((mchunkptr)((char*\
)(victim) - 2*(sizeof(size_t)))))->size & 0x2) || ar_ptr == (((((mchunkptr)((ch\
ar*)(victim) - 2*(sizeof(size_t)))))->size & 0x4) ? ((heap_info *) ((unsigned l\
ong) (((mchunkptr)((char*)(victim) - 2*(sizeof(size_t))))) & ~((2 * (512 * 1024\
)) - 1)))->ar_ptr : &main_arena)' failed.
Aborted
```
I don't have core dump enabled so nothing to point to.  Malloc assertions often point to memory corruption.  Nothing to do now, just leaving behind a clue.

Second time worked.

PR #1381, opened 2016-03-24
WIP - Cleanup and refactor Objective-C support

smolt:
Another go at the Objective-C functions.  I think this is close to what could be put into 1.0.0.  Plumbed in support for arm and arm64 too.  Test suite is now enabled for x86 32-bit too.

### DIFF
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -683,7 +683,7 @@ static void buildRuntimeModule() {
   }
 
   if (global.params.hasObjectiveC) {
-    assert(global.param.targetTriple->isOSDarwin());
+    assert(global.params.targetTriple->isOSDarwin());
 
     // The types of these functions don't really matter because they are always
     // bitcast to correct signature before calling.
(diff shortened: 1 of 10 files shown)


gen/runtime.cpp
```diff
@@ -683,71 +682,42 @@ static void buildRuntimeModule() {
...
-    {
-        llvm::StringRef fname("objc_msgSend_stret");
-        LLType *types[] = { objectPtrTy, selectorPtrTy };
-        LLFunctionType* fty = llvm::FunctionType::get(voidTy, types, true);
-        llvm::Function::Create(fty, llvm::GlobalValue::ExternalLinkage, fname, M);
+  if (global.params.hasObjectiveC) {
+    assert(global.params.targetTriple->isOSDarwin());
+
+    // The types of these functions don't really matter because they are always
+    // bitcast to correct signature before calling.
```
JohanEngelen: Don't they have to be somewhat correct because of linking?

smolt: yeah.  I saw the conditional generation for `_d_cover_register2` and thought I'd make objc runtime conditional too.  Are you thinking it would be better to just declare these unconditionally?

JohanEngelen: I think it is nicer to not put them in the generated IR, like you do now.
I was commenting on the "The types of these functions don't really matter because they are always bitcast to correct signature before calling.", because I thought they would have to mangle correctly for linking. And only just now do I realize that these are extern(C)... doh!

Issue #1327, opened 2016-03-01
Wrong evaluation order for binary operators


trikko:
```
import std.stdio;

void main()
{

	size_t inc(size_t* v)
	{
   		(*v)++;
   		return 10;
	}


	size_t pos = 10;
	
	immutable tst = pos + inc(&pos); // <-- this
	writeln(tst);
	writeln(pos);
}
```

This code outputs:
21
11

Instead of:
20
11

With lcd 1.0.0 alpha

kinke:
Compile-time only, right?

trikko:
I don't think so.

klickverbot:
It's not compile-time only. Also, from this it seems like we codegen the arguments in LTR order, but fail to immediately load from the `pos` lvalue:

```
* * * * * * * DtoVarDeclaration(vdtype = immutable(int))
* * * * * * * * llvm value for decl:   %tst = alloca i32, align 4
* * * * * * * * expression initializer
* * * * * * * * AssignExp::toElem: tst = pos + inc(& pos) | (immutable(int))(immutable(int) = immutable(int))
* * * * * * * * * VarExp::toElem: tst @ immutable(int)
* * * * * * * * * * DtoSymbolAddress ('tst' of type 'immutable(int)')
* * * * * * * * * * * a normal variable
* * * * * * * * * AddExp::toElem: pos + inc(& pos) @ immutable(int)
* * * * * * * * * * VarExp::toElem: pos @ int
* * * * * * * * * * * DtoSymbolAddress ('pos' of type 'int')
* * * * * * * * * * * * a normal variable
* * * * * * * * * * CallExp::toElem: inc(& pos) @ int
* * * * * * * * * * * VarExp::toElem: inc @ int(int* v)
* * * * * * * * * * * * DtoSymbolAddress ('inc' of type 'int(int* v)')
* * * * * * * * * * * * * FuncDeclaration
* * * * * * * * * * * DtoCallFunction()
* * * * * * * * * * * * Building type: int(int* v)
* * * * * * * * * * * * * DtoFunctionType(int(int* v))
* * * * * * * * * * * * * * x86-64 ABI: Transforming return type
* * * * * * * * * * * * * * x86-64 ABI: Transforming argument types
* * * * * * * * * * * * * * Final function type: i32 (i32*)
* * * * * * * * * * * * DtoNestedContext for test.main.inc
* * * * * * * * * * * * * DtoCreateNestedContextType for test.main.inc
[...]

PR #1382, opened 2016-03-24
Ensure strict evaluation order in slice expressions.

JohanEngelen:
Fixes a 2.070 testsuite bug that only manifests itself for 32bit compilation.
Code like this is bugged without this PR:
```D
uint[] values = [0,1,2,3,4];
size_t offset = 0;
auto a = values[offset .. offset+=2];
assert (a == [0,1] && offset == 2);
```
The type of `offset` is important. In the testsuite, the type is `uint`, such that for 64bit compilation the lowerbound expression `32bit offset` is converted to a 64bit value before evaluating the upperbound, hence it all works out OK for 64bit. For 32bit, no conversion is needed and the lowerbound expression is only fully evaluated after offset+=2.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -1331,12 +1331,14 @@ public:
       assert(e->upr);
 
       // get bounds (make sure $ works)
+      // The lower bound expression must be fully evaluated to an RVal before
+      // evaluating the upper bound expression, because the lower bound
+      // expression might change value after evaluating the upper bound, e.g. in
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Submitted a PR for the testsuite: https://github.com/D-Programming-Language/dmd/pull/5575

kinke:
Same cause as for https://github.com/ldc-developers/ldc/issues/1327.

PR #1385, opened 2016-03-26
[WIP] Ensure that the sync object is mutable in a `synchronize(obj) {...}` statement.

JohanEngelen:
See https://github.com/D-Programming-Language/dmd/pull/5564

Also see LTS PR: https://github.com/ldc-developers/ldc/pull/1384

Depends on Phobos PR: https://github.com/ldc-developers/phobos/pull/30

### DIFF
--- a/ddmd/statement.d
+++ b/ddmd/statement.d
@@ -4831,6 +4831,29 @@ public:
                 exp = new CastExp(loc, exp, t);
                 exp = exp.semantic(sc);
             }
+
+            version(IN_LLVM)
+            {
+                /+ Semantic checking is skipped for the calls to _d_monitorenter/exit below,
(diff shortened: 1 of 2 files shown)


JohanEngelen:
Rebased and added diagnostic test for it.
I think it is important to get this in 1.0.0.

JohanEngelen:
It turns out that `synchronized invariant() { }` is allowed, where synchronization happens on `this` with type `shared(const(Klass))`... I am not sure whether this could result in an optimization error or not (like what happened with the const TypeInfo). I will change the PR to not error on this special case.

JohanEngelen:
After some discussion with Walter and Andrei at DConf, this needs more study and discussion. Will work on that.

PR #1386, opened 2016-03-27
Add 'cold' attribute to _d_assert, _d_arraybounds and _d_assert_msg.

redstar:
These functions are only called at most once: in case of a failure.
Marking them as 'cold' may help the (not yet added) IRCE pass.

Also adds the missing 'uwtable' attribute for x86_64.

### DIFF
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -10,11 +10,6 @@
 #include "gen/runtime.h"
 #include "aggregate.h"
 #include "dsymbol.h"
-#include "lexer.h"
-#include "mars.h"
-#include "module.h"
-#include "mtype.h"
-#include "root.h"
 #include "gen/abi.h"
 #include "gen/attributes.h"
 #include "gen/functions.h"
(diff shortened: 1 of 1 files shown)


smolt:
The IRCE pass looks interesting for D.

gen/runtime.cpp
```diff
@@ -299,6 +306,7 @@ static void buildRuntimeModule() {
   AttrSet NoAttrs, Attr_NoAlias(NoAttrs, 0, llvm::Attribute::NoAlias),
       Attr_NoUnwind(NoAttrs, ~0U, llvm::Attribute::NoUnwind),
       Attr_ReadOnly(NoAttrs, ~0U, llvm::Attribute::ReadOnly),
+      Attr_Cold(NoAttrs, ~0U, llvm::Attribute::Cold),
```
smolt: I noticed there is a enum to avoid these `~0U` magic numbers: ` llvm::AttributeSet::FunctionIndex = ~0U` and `llvm::AttributeSet::ReturnIndex = 0U`.

redstar:
The defintion in the IRCE pass says: An inductive range check is conditional branch in a loop with

1. a very cold successor (i.e. the branch jumps to that successor very rarely)

and

2. a condition that is provably true for some contiguous range of values taken by the containing loop's induction variable.

PR #1388, opened 2016-03-27
Tackle issue #1312

kinke:


kinke:
But for the extern(D) functions inside the C++ struct, i.e., ctor, dtor, postblit, opAssign etc.

kinke:
It is irrelevant for C++ interfacing.

It doesn't work yet, e.g., I now get this when compiling the stdlibs:
```
C:\LDC\ldc\runtime\druntime\src\core\internal\convert.d(14): Error: size of type T is not known
Unknown class of D Type!
UNREACHABLE executed at C:\LDC\ldc\gen\tollvm.cpp:196!
```
Please advise on how to make the 2nd commit work.

kinke:
Btw there are some more ignored symbols during codegen for my little test module `d`:
```
Ignoring Dsymbol::codegen for d.object
Ignoring Dsymbol::codegen for d.core
Ignoring Dsymbol::codegen for d.ns.S.__xdtor
Ignoring Dsymbol::codegen for d.ns.S.__xpostblit
Ignoring Dsymbol::codegen for d.DS.__xdtor
Ignoring Dsymbol::codegen for d.DS.__xpostblit
Ignoring Dsymbol::codegen for __entrypoint.object
```
`d.DS` is an `extern(D)` struct. I have a bad feeling about these ignores and think they should probably be an error instead of only showing up in the log.

kinke:
Thanks, that seems to do the trick.

kinke:
I've also reverted the underscore removal as druntime seems to rely on it (MSVC EH code).

kinke:
Amended.

kinke:
Damn, this needs to go to `ltsmaster`.

kinke:
I've already pushed the commit to `ltsmaster`.

Issue #894, opened 2015-04-10
Comparing array to null


d-random-contributor:
Comparing array to null arr==null should be converted to arr.length==0 instead of calling array comparison function.

[Related pull](https://github.com/D-Programming-Language/tools/pull/166)
[dmd bug](https://issues.dlang.org/show_bug.cgi?id=14436) - [fix](https://github.com/D-Programming-Language/dmd/pull/4574)

d-random-contributor:
LDC 0.17 still generates comparison with typeinfo.

kinke:
Verified by:
```D
bool isNull(T)(T[] arg) { return arg == null; }

void main()
{
    int[3] i3 = [ 0, 1, 2 ];
    assert(!isNull(i3[0..1]));
    assert(isNull(i3[0..0]));
}
```
Relevant IR:
```LL
  %2 = getelementptr inbounds { i64, i32* }, { i64, i32* }* %arg, i32 0, i32 0 ; [#uses = 1, type = i64*]
  %.len = load i64, i64* %2                       ; [#uses = 1]
  %3 = icmp eq i64 %.len, 0                       ; [#uses = 1]
  ret i1 %3
```
The `icmp` predicate is `ne` for `arg != null`.

PR #1393, opened 2016-03-28
Optimize array comparisons against null

kinke:
This fixes issue #894 and includes some refactorings.


kinke:
Thinking about it, this case should actually be rewritten by the front-end.

JohanEngelen:
Can you add a test for this? (perhaps a check that the IR does not contain any memory comparison call?)

PR #1396, opened 2016-03-29
Temporarily fix Travis builds while waiting for new LLVM sources to be whitelisted

JohanEngelen:


redstar:
I think this is a problem with locating the package on llvm.org and not with whitelisting. llvm-3.9 needs to be whitelisted. E.g. the Circle-CI build with llvm-3.8 is fine.

JohanEngelen:
@redstar APT source `llvm-toolchain-precise-3.8` has to be whitelisted for llvm 3.8.  For llvm3.9, packages still need whitelisting indeed.  I've submitted all whitelisting request to Travis.

redstar:
Ok - precise is already updated but trusty is not.... As soon as trusty is also updated I will move the Circle CI builder to llvm-3.9.

PR #1404, opened 2016-04-02
Merge 2.070 into master

klickverbot:
We are tantalisingly close to a working 2.070 merge now – for all practical purposes, it is ready.

This is to keep track of sorting out the remaining CI issues.


JohanEngelen:
FileCheck:   I was hoping this would start shipping standard with LLVM-dev packages, but alas. I'll copy FileCheck-3.9 into ltsmaster and master.

Mem management:  I'll put up PRs for this with CI enabled (right now there is a PR for druntime, but no testing done on it)

PR #1413, opened 2016-04-05
Mark memory management functions as @weak, so that user code can override them

JohanEngelen:
druntime diff can be seen here: https://github.com/ldc-developers/druntime/pull/68


JohanEngelen:
All green: merging.

JohanEngelen:
Now merging into master and merge-2.071

PR #1414, opened 2016-04-05
Enable the DDMD overides for memory allocation when LDC+@weak is detected

JohanEngelen:
~~Important: this should be merged *after* druntime has been updated with @weak for the memory allocation functions (otherwise LDC-master will not be able to build itself).
See https://github.com/ldc-developers/ldc/pull/1413~~

The change will become effective once LDC 0.17.2 (yet unreleased, assuming #1413 will be merged) is used.
Edit: so this means that our current CI systems do not use the `OVERRIDE_MEMALLOC=true` path.

### DIFF
--- a/ddmd/root/rmem.d
+++ b/ddmd/root/rmem.d
@@ -163,7 +163,22 @@ else
         goto L1;
     }
 
-    version(DigitalMars)
+    version (DigitalMars)
+    {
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Safe to merge now. The necessary changes are merged into druntime.

PR #1415, opened 2016-04-05
Replace dmd/port.c with our own version.

redstar:
Takes advantage of CMake symbol detection. This reduces the #ifdef hell.
Also includes the C++ headers. After all LDC is written in C++..


thewilsonator:
That file is written in D now, most of the floating point stuff has been moved to `target.d`

PR #1419, opened 2016-04-09
Add Objective-C Support

smolt:
Merge in Objective-C support.  Related pull ldc-developers/dmd-testsuite#22 should be merged first and branch objc-wip updated with its submodule commit.

A few upstream changes are necessary and a pull is being prepared.


smolt:
I'll merge this into master this evening (-7GMT) if I see no comments.

gen/tocall.cpp
```diff
@@ -772,14 +784,15 @@ DValue *DtoCallFunction(Loc &loc, Type *resulttype, DValue *fnval,
   }
 
   // get callee llvm value
-  LLValue *const callable = DtoCallableValue(fnval);
+  //LLValue *const callable = DtoCallableValue(fnval);
```
klickverbot: Huh?

smolt: Ah, garbage collection.  I'll make the comment go away.

gen/tocall.cpp
```diff
@@ -772,14 +784,15 @@ DValue *DtoCallFunction(Loc &loc, Type *resulttype, DValue *fnval,
...
+  LLValue *callable = DtoCallableValue(fnval);
   LLFunctionType *const callableTy =
       DtoExtractFunctionType(callable->getType());
   assert(callableTy);
   const auto callconv = gABI->callingConv(callableTy, tf->linkage,
                                           dfnval ? dfnval->func : nullptr);
 
-  //     IF_LOG Logger::cout() << "callable: " << *callable << '\n';
+  IF_LOG Logger::cout() << "callable: " << *callable << '\n';
```
smolt: I can comment these back out (this line and 822 below).  Were useful during dev, but someone thought they were better as a comment.

TheDharc:
May you tell me if this is just for OS X or for any OS? Because DMD has this feature just on OS X and I'm running FreeBSD here and really would like to get this feature too, so I can use it in my GNUstep programs.
Thanks.

smolt:
Just OS X support.  I haven't looked at the GNUstep ABI.  There is an example in clang CGObjCGNU.cpp, the `ObjCRuntime::GNUstep` path in `CreateGNUObjCRuntime()`.  It sounds like fun but I've got other priorities.  Want to give it a try?

TheDharc:
> Want to give it a try?
I'd like, but have no compiler programming knowledge to do something like that :P.

PR #1421, opened 2016-04-10
Fix C++ declarations causing link errors

rainers:
These are the symbols that caused actual link errors on Windows, but there are actually a lot more places in the headers that are wrong (i.e. adding const to an override does not override in C++, but creates a new vtable slot (if virtual)). This should be fixed in general upstream.

### DIFF
--- a/ddmd/dsymbol.h
+++ b/ddmd/dsymbol.h
@@ -190,7 +190,7 @@ public:
     char *toChars();
     virtual char *toPrettyCharsHelper(); // helper to print fully qualified (template) arguments
     Loc& getLoc();
-    char *locToChars();
+    const char *locToChars();
     bool equals(RootObject *o);
     bool isAnonymous();
     void error(Loc loc, const char *format, ...);
--- a/ddmd/mtype.h
+++ b/ddmd/mtype.h
@@ -788,7 +788,7 @@ public:
     structalign_t alignment();
     Expression *defaultInit(Loc loc);
     Expression *defaultInitLiteral(Loc loc);
-    bool isZeroInit(Loc loc) const;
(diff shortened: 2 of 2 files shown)


JohanEngelen:
( It'd be so nice if someone would write a C++-D interop checker tool, to check that the C++ headers and D declarations are the same. :)   As a start, just a tool to check class and vtable sizes would catch these tricky to find errors. )

rainers:
> As a start, just a tool to check class and vtable sizes would catch these tricky to find errors.

This would not find the issue in this PR as it introduced a non-virtual member function.
IIRC Daniel thought about generating the C++ headers from D code. That could help, too.

JohanEngelen:
> but creates a new vtable slot (if virtual)

Was commenting on that.

yebblies:
> IIRC Daniel thought about generating the C++ headers from D code. That could help, too.

It pretty much works: https://github.com/D-Programming-Language/dmd/pull/5082

rainers:
> It pretty much works: D-Programming-Language/dmd#5082

Looks pretty good. From reading funcToBuffer, it seems const modifiers on member functions are not emitted, so overrides are ok, but mangling is not.
How far do you think it is from being used? Does it make sense to still fix the headers regarding const overrides?

rainers:
I created a PR for dmd: https://github.com/dlang/dmd/pull/5673

rainers:
> I created a PR for dmd: dlang/dmd#5673

This has been merged. Is it ok to cherry-pick this into the merge-2.071 branch?

JohanEngelen:
> Is it ok to cherry-pick this into the merge-2.071 branch?
Yes!
Nice work on getting merge-2.071 closer to green.

PR #1422, opened 2016-04-10
When emitting a symbol in a COMDAT "any" group, set LLVM linkage to LinkOnceODR

JohanEngelen:
The mismatch between COMDAT "any" semantics and "ExternalLinkage" semantics becomes an issue with LLVM's PGO functionality; this PR attempts to fix that.

### DIFF
--- a/gen/tollvm.cpp
+++ b/gen/tollvm.cpp
@@ -237,15 +237,18 @@ LLValue *DtoDelegateEquals(TOK op, LLValue *lhs, LLValue *rhs) {
 ////////////////////////////////////////////////////////////////////////////////
 
 LinkageWithCOMDAT DtoLinkage(Dsymbol *sym) {
-  auto linkage = (DtoIsTemplateInstance(sym) ? templateLinkage
-                                             : LLGlobalValue::ExternalLinkage);
-
-  // If @(ldc.attributes.weak) is applied, override the linkage to WeakAny
-  if (hasWeakUDA(sym)) {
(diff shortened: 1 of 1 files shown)


JohanEngelen:
The discussion is here: https://github.com/ldc-developers/ldc/issues/1310

My opinion is also that emitting all as comdat "any" is not good. But kinke mentioned dmd-testsuite breaks for Win64 otherwise.

JohanEngelen:
My goal was to move fully to comdat "any" linkage, with consistent LLVM linkage.

But now, I'm not sure what to do anymore. I can fix the tests. But ... I turned on my Windows machine now and will see if I can move back to non-comdat "any" linkage.

JohanEngelen:
If we no longer emit functions in comdat "any", weak linkage does not work on Windows and so the rmem.d memory "optimization" won't work. I am now leaning towards this PR, instead of removing comdat "any" for all symbols.

JohanEngelen:
Re-re-reading LLVM documentation, I think linkonce_odr is also not the right semantics: "Unreferenced linkonce globals are allowed to be discarded".

:(
Perhaps I will just have to live with no PGO on Windows.

JohanEngelen:
> The function overriding a @weak function needs to be in a COMDAT "any" section too?

Yes. As far as I understand, Windows does not have a "weak linkage" type. Symbols are put in comdat groups regardless of whether we specify comdat or not  (by the linker or LLVM, I don't know), but with "no duplicate" selection. But a comdat "no duplicates" symbol cannot override a comdat "any" symbol.

PR #1431, opened 2016-04-14
fix building against LLVM master. Patch by @rainers

JohanEngelen:


JohanEngelen:
The LLVM version on AppVeyor is a little behind the Ubuntu APT version, so LLVM3.9 build is either broken for CicleCI or AppVeyor.
@kinke Could you update the binaries for AppVeyor? Does it make sense to ask the LLVM community to provide zipped versions of trunk built with VS2015?  (they do have regularly updated snapshot builds, but those are installer exe and I don't know which version of VS is used to build them)

rainers:
This might allow compilation for both versions of LLVM:

```
// intermediate mapping to either RF_IgnoreMissingEntries or RF_IgnoreMissingLocals
// that has been renamed during LLVM 3.9 development
struct RF_IgnoreMissing {
  // For the compile time comparison.
  template <typename C> struct _entries { static const auto value = C::RF_IgnoreMissingEntries; };
  template <typename C> struct _locals  { static const auto value = C::RF_IgnoreMissingLocals; };

  template <typename C> static _entries<C> test(decltype(C::RF_IgnoreMissingEntries)) {}
  template <typename C> static _locals<C> test(...) {}

  static const auto value = decltype(test<llvm::RemapFlags>(0))::value;
};

// remap values in all instructions of all blocks
void remapBlocks(std::vector<llvm::BasicBlock *> &blocks,
                 llvm::ValueToValueMapTy &VMap, llvm::BasicBlock *unwindTo,
                 llvm::Value *funclet) {
  for (llvm::BasicBlock *bb : blocks)
    for (llvm::BasicBlock::iterator I = bb->begin(); I != bb->end(); ++I) {
      llvm::RemapInstruction(&*I, VMap, RF_IgnoreMissing::value |
                                            llvm::RF_NoModuleLevelChanges);
    }
}
```

rainers:
Hmmm, yesterday llvm::getGlobalContext has been removed from LLVM master. It seems we have to switch to the C API LLVGetGlobalContext instead, but the returned type is incompatible with `llvm::LLVMContext &`. Should we replace this with `LLVMContextRef` everywhere or wrap this somehow and cast it back?

rainers:
> We could just create our own global static llvm::Context.

Here's another stab at fixing the build including our own global context: https://github.com/ldc-developers/ldc/pull/1435

JohanEngelen:
Closing in favor of #1435

PR #1434, opened 2016-04-15
Add LDC-specific traits for CTFE information about the target machine.

JohanEngelen:
Edit: updated to current state of the PR.  (removed `__traits(targetFeatures)`)

This adds the following LDC-specific traits:
```D
__traits(targetCPU) == "broadwell"
__traits(targetHasFeature, "sse2") == bool
```

Example usage:
```D
import std.stdio;
void main() {
  writeln("CPU = ", __traits(targetCPU));
  writeln("Has 'sse3' = ", __traits(targetHasFeature, "sse3"));
  writeln("Has 'sse4' = ", __traits(targetHasFeature, "sse4"));
  writeln("Has 'sse4.1' = ", __traits(targetHasFeature, "sse4.1"));
}
```
Outputs on my machine:
```
❯ ../bin/ldc2 -wi -mcpu=native -run ../../tests/codegen/target_traits.d
CPU = haswell
Has 'sse3' = true
Has 'sse4' = false
Has 'sse4.1' = true
```

### DIFF
--- a/ddmd/hooks.d
+++ b/ddmd/hooks.d
@@ -10,7 +10,10 @@ module ddmd.hooks;
 import ddmd.dscope;
 import ddmd.expression;
 
+import gen.ldctraits;
+
+/// Returns `null` when the __trait was not recognized.
 Expression semanticTraitsHook(TraitsExp e, Scope* sc)
 {
-    return null;
+    return semanticTraitsLDC(e, sc);
 }
--- a/ddmd/hooks.d
+++ b/ddmd/hooks.d
@@ -1,0 +1,16 @@
+// Compiler implementation of the D programming language
+// Copyright (c) 1999-2016 by Digital Mars
+// All Rights Reserved
+// http://www.digitalmars.com
+// Distributed under the Boost Software License, Version 1.0.
+// http://www.boost.org/LICENSE_1_0.txt
+
+module ddmd.hooks;
+
+import ddmd.dscope;
+import ddmd.expression;
+
+Expression semanticTraitsHook(TraitsExp e, Scope* sc)
+{
+    return null;
+}
--- a/ddmd/idgen.d
+++ b/ddmd/idgen.d
@@ -393,6 +393,10 @@ Msgtable[] msgtable =
     { "LDC_global_crt_ctor" },
     { "LDC_global_crt_dtor" },
     { "LDC_extern_weak" },
+
+    // IN_LLVM: LDC-specific traits.
+    { "targetCPU" },
+    { "targetHasFeature" },
 ];
 
 
--- a/ddmd/traits.d
+++ b/ddmd/traits.d
@@ -27,6 +27,7 @@ import ddmd.expression;
 import ddmd.func;
 import ddmd.globals;
 import ddmd.hdrgen;
+import ddmd.hooks;
 import ddmd.id;
 import ddmd.identifier;
 import ddmd.mtype;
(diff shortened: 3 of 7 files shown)


JohanEngelen:
@9il what do you think?

9il:
Awesome! Thank you a lot!

WalterBright:
```
__traits(targetFeatures) == Tuple("sse", "sse2", ...)
__traits(targetHasFeature, "sse2") == bool
```
The latter is redundant. We should try to avoid redundant core language features.

UplinkCoder:
so is hasMember

JohanEngelen:
> The latter is redundant. We should try to avoid redundant core language features.

I left targetHasFeature for convenience.
`any!q{a = "sse4.1"}([__traits(targetFeatures)])` is quite ugly I think for functionality that is probably most used out of the three traits.

ibuclaw:
I can't say that gcc readily provides this information, but it seems reasonable.

WalterBright:
> so is hasMember

True, but we learn from our mistakes rather than repeat them.

> is quite ugly

`__traits` is always ugly. It was intended to be buried in library code. Things that are easily done in the library should not be in the core language.

JohanEngelen:
> I can't say that gcc readily provides this information

Neither does LLVM :(
For -mcpu=native the information is easy to obtain, for -mcpu=haswell it is not...

ibuclaw:
> Neither does LLVM :(

Well at least you have `getTargetFeatureString` and `getTargetCPU` exposed to the frontend. ;-)

JohanEngelen:
OK, I've been able to squeeze the feature info out of LLVM with a small LLVM patch :)
I'll submit that patch to LLVM, and see how that goes over. Then for LLVM<3.9, __traits(targetHasFeature, ...) will only work when either explicitly mentioning the feature on the cmdline, or with cpu=native.

ibuclaw:
I guess you could cache this information in `Target` also.  That may be the best way for me to do it without adding a new gcc target "hook".

JohanEngelen:
http://reviews.llvm.org/D19189

mleise:
Nice! I hope this makes it into all compilers. There is always the option to define a safe fall-back like "i386" for dmd on x86. `targetHasFeature` would be the most frequently used trait in the set. The list version is really only good for "compiled with \<features\>" output ... unless someone writes a nice wrapper.

@ibuclaw Maybe checking the implementation of this helps:
```
gcc -march=native -dM -E - < /dev/null | egrep "MMX|SSE|AVX" | sort
#define __AVX__ 1
#define __AVX2__ 1
#define __MMX__ 1
#define __SSE__ 1
…
```

ibuclaw:
@mleise - I'm aware of how gcc generates macros, gdc uses the same mechanism for generating pre-defined version conditions. ;-)

All I said was that there is no "ready to use" way of getting the information.

JohanEngelen:
I removed `__traits(targetFeatures)` as this list is not obtainable without hacking LLVM. `__traits(targetHasFeature)` is more useful and this can be obtained from LLVM with some effort.
Updated the PR.

JohanEngelen:
For LLVM version < 3.7, `hasTargetFeature` will always return the safe `false` default. I don't know how to obtain the necessary information with LLVM < 3.7.

JohanEngelen:
Ping! :)

9il:
LGTM

This would be very helpful for [Mir's BLAS Micro Kernel](https://github.com/libmir/mir/pull/210/files) parameter estimation. LDC can generates [(see gist)](https://gist.github.com/9il/58c1b80110de2db5f2eff6999346a928) efficient micro kernel code, and #1472 will allow to optimize it even more :-)

JohanEngelen:
@redstar Can I merge this?  The "hook" is the same as #5771 so it is easily modified once that is merged.

JohanEngelen:
Will merge when green.

JohanEngelen:
@9il Can you update us on how you are using this in your code? :)

9il:
Will do with BLAS release :)

PR #1435, opened 2016-04-16
Fix master for recent LLVM master changes

rainers:
another PR to build with LLVM master, this time also adding our own context.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -30,6 +30,7 @@
 #include "gen/irstate.h"
 #include "gen/linkage.h"
 #include "gen/llvm.h"
+#include "gen/llvmhelpers.h"
 #include "gen/logger.h"
 #include "gen/metadata.h"
 #include "gen/objcgen.h"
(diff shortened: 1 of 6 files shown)


JohanEngelen:
Do we really want to support several versions of LLVM trunk? :)
Anyway, nice work. I am in favor of merging so we have LLVM3.9 CI testing again, and hope we remember to make the code a little prettier after updating appveyor's LLVM.

rainers:
If we still support LLVM 3.8 building the ms-cxx-helper.cpp, the usual #if/#else/#endif needs 5 lines of code instead of 7 now ;-) It might not need as many explanations, though.

Any objections to merging this? I'd like to cherry-pick it to the merge-2.071 branch, too.

JohanEngelen:
> If we still support LLVM 3.8 building the ms-cxx-helper.cpp

Don't we? My Mac is building ms-cxx-helper.cpp. I assumed this was to support cross-compiling.

> the usual #if/#else/#endif needs 5 lines of code instead of 7 now ;-)

:)

rainers:
>> If we still support LLVM 3.8 building the ms-cxx-helper.cpp
 
> Don't we?

The MSVC exception handling had some serious bugs that were fixed in 3.9 only. It was unclear whether the fixes would make it into 3.8.1, too.

> Merged #1435.

Thanks.

Issue #1408, opened 2016-04-03
Cannot build MS VC projects for LDC


rainers:
Running cmake (version 3.3.2) to build projects for Visual Studio fails:

```
>cmake -G "Visual Studio 14 Win64"
CMake Error at CMakeLists.txt:588 (add_custom_command):
  Error evaluating generator expression:

    $<TARGET_FILE_DIR:ldc2>

  Target "ldc2" is not an executable or library.
```

The offending line is specific to MSVC_IDE:

```
if(MSVC_IDE)
    # the IDE generator is a multi-config one
    # so copy the config file into the correct bin subfolder
    add_custom_command(TARGET ${LDC_EXE} POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy ${PROJECT_BINARY_DIR}/bin/${LDC_EXE}.conf $<TARGET_FILE_DIR:${LDC_EXE}> COMMENT "Copy config file ${LDC_EXE}.conf")
endif()
```

My cmake-expertise is non-existent, but it seems to me that the target is assumed to be an executable name.

JohanEngelen:
I don't understand what this is supposed to do. To me it looks like it is trying to copy
${PROJECT_BINARY_DIR}/bin/ldc2.conf to the same location.
$<TARGET_FILE_DIR:${LDC_EXE}> used to point to ${PROJECT_BINARY_DIR}/bin (the output folder is explicitly set to that)

If you comment-out that offending line, does anything not work for Visual Studio generator?

rainers:
VS builds used to generate the exe into bin/Release, bin/Debug or bin/RelWithDebInfo, but does not do now. 
I guess it is trying to copy ldc2.conf into these directories. It builds just fine with the line commented.

JohanEngelen:
Should we try to restore the bin/Release, bin/Debug, etc, behavior? Does the ltsmaster cmake script still work like that?

rainers:
I currently only build RelWithDebInfo, so it doesn't make a big difference for me. Debugging is sometimes a bit bad with optimizations, though. Before, I used Debug and Release configurations, but that caused trouble at some point, too (IIRC undefining NDEBUG didn't always match what LLVM does).

> Does the ltsmaster cmake script still work like that?

I haven't tried that yet.

kinke:
[...]

PR #1437, opened 2016-04-16
Add extra dependency on running idgen, so it is actually re-run after rebuilding it

JohanEngelen:
This in turn triggers rebuilding of LDC with new idgen.d identifiers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -292,7 +292,7 @@ macro(Dcompilelib  input_d  output_dir  extra_d_flags  outlist_o)
 endmacro()
 
 # Build (generate executable) of D source
-macro(Dbuild  target_id  input_d  output_exec  extra_d_flags  link_flags  extra_deps)
+macro(build_idgen  input_d  output_exec  extra_d_flags  link_flags  extra_deps)
     separate_arguments(FLAG_LIST WINDOWS_COMMAND "${D_COMPILER_FLAGS} ${extra_d_flags} ${link_flags}")
     add_custom_command(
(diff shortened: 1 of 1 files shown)


rainers:
For windows you should append `${CMAKE_EXECUTABLE_SUFFIX}`, too.

Slightly off-topic: The generated VS projects are a bit strange with regard to idgen:
- idgen.vcxproj contains a rule `idgen.rule` that doesn't contain any useful commands but declares an output. This causes any dependent project to be considered out-of-date. Can we disable generating this rule?
- the projects `LDCShared` and `ldc2` contain the same idgen.exe.rule again. Isn't this redundant with the project reference?

JohanEngelen:
@rainers Thanks for the Windows fix.
I don't really understand why the "idgen" target didn't work and needed the extra dependency. I've now completely removed the idgen custom target, does that work better with VS projects?

rainers:
> does that work better with VS projects?

Yes, looks better. There is still a rule `ldc2.rule` in the ldc2 project in addition to `ldc2.exe.rule`. It declares output `CMakefiles/ldc2` that seems to trigger rebuilds, but it's never built.

JohanEngelen:
I don't remember exactly why I added the ldc2 custom target, but I remember there being some issues with the runtime libraries' rules for different versions of cmake regarding the ldc2 target.

rainers:
Ok, ldc2 target is a different story (see also https://github.com/ldc-developers/ldc/issues/1408), so let's get this in.

PR #1439, opened 2016-04-17
fix runnable/interface2

rainers:
Trying to fix the C++ interface casting I stumbled across a number of issues:
- casting didn't take full advantage of static type information
- I didn't really grasp the logic in DtoCastClass, so I changed it the to mimick the simpler one in dmd.
- caching vtbls didn't take the thunk offset into account

### DIFF
--- a/ir/iraggr.h
+++ b/ir/iraggr.h
@@ -107,10 +107,11 @@ protected:
   /// ClassInfo initializer constant.
   llvm::Constant *constClassInfo = nullptr;
 
-  /// Map for mapping ClassDeclaration* to LLVM GlobalVariable.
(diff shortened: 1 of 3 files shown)


ir/iraggr.h
```diff
@@ -108,7 +108,7 @@ struct IrAggr {
   llvm::Constant *constClassInfo = nullptr;
 
   /// Map for mapping ClassDeclaration* to LLVM GlobalVariable.
```
klickverbot: Could you please update the comments mentioning the new index? (Probably delete this rather vacuous comment, and update the one for interfaceVtblMap).

rainers:
> Should be fine – feel free to just push the commit to merge-2.071 after amending the comment.

Ok, done.

Issue #431, opened 2013-07-25
Inline llvm IR


John-Colvin:
I'm not sure how useful this would be, but I can't see it being hard to implement.

__IR
{
//some llvm IR
}

PR #1441, opened 2016-04-17
Add `@ldc.attributes.llvmAttr`

### LINKED ISSUE #1438: FMA instruction
Hello,

I am looking for a *portable* way of vectorized FMA operations for BLAS.

### Variant 1
`fast` attribute
```
double dot(double[] a, double[] b)
{
	typeof(return) s = 0;
	
	foreach(size_t i; 0..c.length)
	{
		s = inlineIR!(`
		%p = fmul fast double %0, %1
		%r = fadd fast double %p, %2
		ret double %r`, double)(a[i], b[i], s);
	}

	return s;
}
```
Result: No FMA
```
LBB0_8:
	vmovupd	-224(%rax), %ymm4
	vmovupd	-192(%rax), %ymm5
	vmovupd	-160(%rax), %ymm6
	vmovupd	-128(%rax), %ymm7
	vmulpd	-224(%rsi), %ymm4, %ymm4
	vmulpd	-192(%rsi), %ymm5, %ymm5
	vmulpd	-160(%rsi), %ymm6, %ymm6
	vmulpd	-128(%rsi), %ymm7, %ymm7
	vaddpd	%ymm0, %ymm4, %ymm0
	vaddpd	%ymm1, %ymm5, %ymm1
	vaddpd	%ymm2, %ymm6, %ymm2
	vaddpd	%ymm3, %ymm7, %ymm3
	vmovupd	-96(%rax), %ymm4
	vmovupd	-64(%rax), %ymm5
	vmovupd	-32(%rax), %ymm6
	vmovupd	(%rax), %ymm7
	vmulpd	-96(%rsi), %ymm4, %ymm4
	vmulpd	-64(%rsi), %ymm5, %ymm5
	vmulpd	-32(%rsi), %ymm6, %ymm6
	vmulpd	(%rsi), %ymm7, %ymm7
	vaddpd	%ymm0, %ymm4, %ymm0
	vaddpd	%ymm1, %ymm5, %ymm1
	vaddpd	%ymm2, %ymm6, %ymm2
	vaddpd	%ymm3, %ymm7, %ymm3
	addq	$256, %rsi
	addq	$256, %rax
	addq	$-32, %rdx
	jne	LBB0_8
```

### Variant 2
`llvm_fmuladd` function
```
double dot(double[] a, double[] b)
{
	typeof(return) s = 0;
	
	foreach(size_t i; 0..a.length)
	{
		s = llvm_fmuladd(a[i], b[i], s);
	}

	return s;
}
```
Result: No Vectorization
```
LBB0_7:
	vmovsd	-56(%rcx), %xmm1
	vmovsd	-48(%rcx), %xmm2
	vfmadd132sd	-56(%rax), %xmm0, %xmm1
	vfmadd231sd	-48(%rax), %xmm2, %xmm1
	vmovsd	-40(%rcx), %xmm0
	vfmadd132sd	-40(%rax), %xmm1, %xmm0
	vmovsd	-32(%rcx), %xmm1
	vfmadd132sd	-32(%rax), %xmm0, %xmm1
	vmovsd	-24(%rcx), %xmm0
	vfmadd132sd	-24(%rax), %xmm1, %xmm0
	vmovsd	-16(%rcx), %xmm1
	vfmadd132sd	-16(%rax), %xmm0, %xmm1
	vmovsd	-8(%rcx), %xmm2
	vfmadd132sd	-8(%rax), %xmm1, %xmm2
	vmovsd	(%rcx), %xmm0
	vfmadd132sd	(%rax), %xmm2, %xmm0
	addq	$64, %rax
	addq	$64, %rcx
	addq	$-8, %rdx
	jne	LBB0_7
```

JohanEngelen:
~~Resolves #1438~~
Partly resolves #1438

### DIFF
--- a/gen/llvmhelpers.h
+++ b/gen/llvmhelpers.h
@@ -24,6 +24,13 @@
 
 struct IRState;
 
+// An arrayreference type with initializer_list support (C++11):
+#if LDC_LLVM_VER >= 307
+template <class T> using ArrayParam = llvm::ArrayRef<T>;
+#else
+template <class T> using ArrayParam = std::vector<T>;
+#endif
+
 llvm::LLVMContext& getGlobalContext();
 
 // dynamic memory helpers
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -215,13 +215,6 @@ template <typename DECL> static void ensureDecl(DECL *decl, const char *msg) {
   }
 }
 
-// We need a vector type with initializer_list support:
-#if LDC_LLVM_VER >= 307
-template <class T> using ArrayParam = llvm::ArrayRef<T>;
-#else
-template <class T> using ArrayParam = std::vector<T>;
-#endif
-
 // Parameters fnames are assumed to be already mangled!
 static void createFwdDecl(LINK linkage, Type *returntype,
                           ArrayParam<llvm::StringRef> fnames,
--- a/gen/uda.cpp
+++ b/gen/uda.cpp
@@ -1,6 +1,7 @@
 #include "gen/uda.h"
 
 #include "gen/llvm.h"
+#include "gen/llvmhelpers.h"
 #include "aggregate.h"
 #include "attrib.h"
 #include "declaration.h"
@@ -13,9 +14,10 @@ namespace {
 
 /// Names of the attribute structs we recognize.
 namespace attr {
+const std::string llvmAttr = "llvmAttr";
 const std::string section = "section";
-const std::string target  = "target";
-const std::string weak    = "_weak";
+const std::string target = "target";
+const std::string weak = "_weak";
 }
 
 /// Checks whether `moduleDecl` is the ldc.attributes module.
@@ -62,7 +64,7 @@ StructLiteralExp *getLdcAttributesStruct(Expression *attr) {
   return nullptr;
 }
 
-void checkStructElems(StructLiteralExp *sle, llvm::ArrayRef<Type *> elemTypes) {
+void checkStructElems(StructLiteralExp *sle, ArrayParam<Type *> elemTypes) {
   if (sle->elements->dim != elemTypes.size()) {
     sle->error(
         "unexpected field count in 'ldc.attributes.%s'; does druntime not "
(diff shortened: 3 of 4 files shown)


JohanEngelen:
> so the code in an alwaysinline function would always end up being compiled with the parent context attributes.

Nope.

JohanEngelen:
Agreed that it is a hack. I think it works because we start defining the function upon when it is first called. I first tried walking the FuncDecl->TemplInst->  path does not work, because it is instantiated outside the calling function (I couldn't get it to work, with as far as I understand the AST). Adding template parameters is not easy without breaking things I think, because the last template parameter is variadic?

JohanEngelen:
> How does it work now?

Without the "inheriting attributes" thing, LLVM IR calls the "inlineIR" function, inlines it, and sees that the `dot()` function has an attribute that is not compatible with the inlineIR function and explicitly sets the attribute to the safest option: `"unsafe-fp-math"="false"`. That's my explanation of what I see in the resulting IR. (try it, it doesn't work without the attribute inheritance)
I think the real problem here is that "inlineIR" does not inline the IR. A misunderstanding of me perhaps.

The testsuite made me realize that it is actually possible to define an inlineIR alias outside a function, and call the alias. In that case, you have to use that second (new) test case UDA construct, because the hack doesn't work: codegen starts outside the calling function. Also, like you say, you would get different functions with different attributes calling the same inlineIR function...

JohanEngelen:
> This seems to couple the semantics of the template instantiation to when it is codegen'd

Perhaps we have reached the limits of what you can do with the trick of a templated function with a magic pragma on top of it...

JohanEngelen:
*inline* the IR :)

JohanEngelen:
> Of course, if somebody specified the parameter types explicitly, things will still break.

When using an alias to define an LLVM IR function-like thing, you have to specify parameter types explicitly.

JohanEngelen:
Shall I rip out the inlineIR business? Take things one step at a time.
Perhaps all that is needed is beef up the inlineIR end of things. Try to do an actual inline, instead of a function call.

JohanEngelen:
(also see https://github.com/ldc-developers/ldc/issues/431)

JohanEngelen:
> I'd rip out the attribute inheritance part for now and merge the rest.

Will do.

JohanEngelen:
I just looked at the inline ASM stuff: it delays codegen until `visit(CallExp *e)`, where it really tries to inline things at that moment. That may be a path for the inline IR stuff too (as you mention above: instantiate a new function at each call site and inline that).

JohanEngelen:
(thanks for careful review and keeping me in check btw :))

PR #1443, opened 2016-04-17
Cherry-picked changes from master to fix build with recent LLVM master

rainers:
Fix master for recent LLVM master changes

This is commit 99fbd197a316f5b70d9c00e3aec50aeedca5fda1 on master, see also #1435

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -30,6 +30,7 @@
 #include "gen/irstate.h"
 #include "gen/linkage.h"
 #include "gen/llvm.h"
+#include "gen/llvmhelpers.h"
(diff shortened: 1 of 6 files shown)


JohanEngelen:
@rainers I think it is easier to apply this kind of fix from master directly to merge-2.071 without PRs. It saves some time and annoyance. (unless you could reasonably expect that it would break the branch very badly). Another way that I think works well is merging master into merge-2.071 (without PR).

rainers:
@JohanEngelen Ok, a merge from master seems a good idea. I'll try that, but I'm currently having trouble with the submodules...

PR #1447, opened 2016-04-19
Add a handful of fixes for Android

joakim-noah:
I finally matched the way dmd packs emulated TLS data with ldc, see below for more comments.

### DIFF
--- a/dmd2/root/port.c
+++ b/dmd2/root/port.c
@@ -654,10 +654,6 @@ longdouble Port::strtold(const char *p, char **endp)
 #if __linux__ || __APPLE__ || __FreeBSD__ || __OpenBSD__ || __NetBSD__ || __DragonFly__ || __HAIKU__
 
 #include <math.h>
-#if __linux__
-#include <bits/nan.h>
-#include <bits/mathdef.h>
-#endif
 #if __FreeBSD__ && __i386__
 #include <ieeefp.h>
 #endif
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -216,6 +216,24 @@ void CodeGenerator::emit(Module *m) {
       emitSymbolAddrGlobal(ir_->module, "__bss_start", "_d_execBssBegAddr");
       emitSymbolAddrGlobal(ir_->module, "_end", "_d_execBssEndAddr");
     }
+
+    // On Android, bracket TLS data with the symbols _tlsstart and _tlsend, as
+    // done with dmd
+    if (global.params.targetTriple.getEnvironment() == llvm::Triple::Android) {
+      auto startSymbol = new llvm::GlobalVariable(
+          ir_->module, llvm::Type::getInt32Ty(ir_->module.getContext()), false,
+          llvm::GlobalValue::ExternalLinkage,
+          llvm::ConstantInt::get(ir_->module.getContext(), APInt(32,0)),
+          "_tlsstart", &*(ir_->module.global_begin()));
+      startSymbol->setSection(".tdata");
+
+      auto endSymbol = new llvm::GlobalVariable(
+          ir_->module, llvm::Type::getInt32Ty(ir_->module.getContext()), false,
+          llvm::GlobalValue::ExternalLinkage,
+          llvm::ConstantInt::get(ir_->module.getContext(), APInt(32,0)),
+          "_tlsend");
+      endSymbol->setSection(".tcommon");
+    }
   }
 
   finishLLModule(m);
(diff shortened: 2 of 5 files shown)


dmd2/root/port.c
```diff
@@ -654,10 +654,6 @@ longdouble Port::strtold(const char *p, char **endp)
 #if __linux__ || __APPLE__ || __FreeBSD__ || __OpenBSD__ || __NetBSD__ || __DragonFly__ || __HAIKU__
 
 #include <math.h>
-#if __linux__
-#include <bits/nan.h>
-#include <bits/mathdef.h>
```
joakim-noah: These headers don't exist in bionic, and the glibc versions say they should never be included directly.

driver/codegenerator.cpp
```diff
@@ -216,6 +216,24 @@ void CodeGenerator::emit(Module *m) {
...
+	  "_tlsstart", &*(ir_->module.global_begin()));
+      startSymbol->setSection(".tdata");
+
+      auto endSymbol = new llvm::GlobalVariable(
+          ir_->module, llvm::Type::getInt32Ty(ir_->module.getContext()), false,
+	  llvm::GlobalValue::ExternalLinkage,
+          llvm::ConstantInt::get(ir_->module.getContext(), APInt(32,0)),
+	  "_tlsend");
+      endSymbol->setSection(".tcommon");
+    }
```
joakim-noah: [Match the way dmd brackets TLS data](https://github.com/dlang/dmd/blob/master/src/backend/elfobj.c#L1900), as [my small modifications to llvm](https://gist.github.com/joakim-noah/1fb23fba1ba5b7e87e1a) and the current `rt.sections_android` use this scheme.

However, [llvm recently added its own emulated TLS scheme similar to gcc](http://reviews.llvm.org/D10522) and we'll want to move over to that eventually.  I've tried that new scheme from llvm on Android/ARM and it works, with the only issue passing the emulated TLS data to the GC, which I've not looked into.

We can use this dmd scheme for now, which requires a patched llvm, or not merge this one block and just wait till that new scheme from llvm gets integrated.

smolt: My vote would be to get in what you have now.

smolt:
Looks good to me.

driver/codegenerator.cpp
```diff
@@ -216,6 +216,24 @@ void CodeGenerator::emit(Module *m) {
...
+          "_tlsstart", &*(ir_->module.global_begin()));
+      startSymbol->setSection(".tdata");
+
+      auto endSymbol = new llvm::GlobalVariable(
+          ir_->module, llvm::Type::getInt32Ty(ir_->module.getContext()), false,
+          llvm::GlobalValue::ExternalLinkage,
+          llvm::ConstantInt::get(ir_->module.getContext(), APInt(32,0)),
+          "_tlsend");
+      endSymbol->setSection(".tcommon");
+    }
```
joakim-noah: Resurrecting this comment, which only got hidden because I corrected some whitespace:

[Match the way dmd brackets TLS data](https://github.com/dlang/dmd/blob/master/src/backend/elfobj.c#L1900), as [my small modifications to llvm](https://gist.github.com/joakim-noah/1fb23fba1ba5b7e87e1a) and the current `rt.sections_android` use this scheme for emulated TLS.

However, [llvm recently added its own emulated TLS scheme similar to gcc](http://reviews.llvm.org/D10522), which shipped with llvm 3.8, and we'll want to move over to that eventually.  I've tried that new scheme from llvm on Android/ARM and it works, with the only issue passing the emulated TLS data to the GC, which I've not looked into.

We can use this dmd scheme for now, which requires a patched llvm, or not merge this one block and just wait till that new emulated TLS scheme from llvm gets integrated with the GC.

joakim-noah:
The Circle CI build seems broken only because it compiles against llvm 3.9 trunk, which ltsmaster doesn't appear to support yet?  It breaks against llvm 3.9 on Travis too, though that looks like some configuration error.

driver/codegenerator.cpp
klickverbot: Is this guaranteed to work? That is, are _tlsstart and _tlsend guaranteed to end up in the correct place even after linking multiple LLVM IR modules together? Since we no longer use llvm::Linker to implement `-singleobj`, it's probably not as critical, though.

joakim-noah: No and yes, `_tlsstart` depends on the ELF object with the main function being passed first to the linker, which, for example, the scripts that produce the test runners don't do.  I don't know about linking at the IR level.  `_tlsend` should always be at the end because of the ld.bfd linker script that puts `.tcommon` last.

PR #1448, opened 2016-04-19
For `LDC_inline_ir` functions: instantiate a new "alwaysinline" function for each call

JohanEngelen:
This is the current behavior for inline ASM too.

When certain attributes are applied to the calling function, like "unsafe-fp-math", the inlined inlineIR function has to have the same attributes otherwise the calling function attribute will be reset to the safe merge of the two: "false". Because the same inlineIR function can be called in different functions using an alias definition, a new function (with possibly different attributes) has to be instantiated upon every call.

Related GH issue #1438

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -384,9 +384,9 @@ void DtoResolveFunction(FuncDeclaration *fdecl) {
           static_cast<TypeFunction *>(type)->linkage = LINKc;
 
           DtoFunctionType(fdecl);
-          DtoDeclareFunction(fdecl);
           fdecl->ir->setDefined();
-          return;
+          return; // this gets mapped to a special inline IR call, no point in
+                  // going on.
         }
       }
     }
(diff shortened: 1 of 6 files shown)


JohanEngelen:
Hmm.. having to deal with the old versions of FileCheck is a little annoying, perhaps I'll need to spend some time in backporting recent FileCheck version to older LLVM versions...

JohanEngelen:
Updated FileChecks.
Will have to restart CI builds after #1449 is merged.

JohanEngelen:
CI is green now after updating the FileCheck's.

@klickverbot @redstar What do you think?

JohanEngelen:
I agree the hard-coded list is brittle, but I don't know we can do anything better at the moment... :/
I tried to copy all attributes at first, but that doesn't work (I also added a testcase for when it fails): Function attributes include attributes on the function parameters. Those of the parent of course don't apply to the child, and so we can't just copy _all_ attributes. Now, in LLVM there is mention of "string attributes (target dependent)". I am not sure what that means, but perhaps that _does_ exclude all attributes related to parameters and return type. So I could loop over all attributes, check if they are string attributes and apply those to the inline IR function. I'll try some more things; for example I did not test yet if ""target-features"="+sse" needs to be copied to the inline IR function...
(_super_ happy with the IR testsuite btw!)

JohanEngelen:
Thanks, missed that!  Updated PR.

JohanEngelen:
Great, thanks.

JohanEngelen:
@klickverbot I think I will change the inheriting of all function attributes and go back to the hard-coded list of attributes that I am sure we _should_ inherit. Reading the list of possible function attributes, there are a bunch that I think the inlineIR should not inherit. http://llvm.org/docs/LangRef.html#function-attributes

JohanEngelen:
(or I'll limit the inherit attributes to "string attributes")

JohanEngelen:
Indeed, but "behaving as if inlined" is a little different from "call a function with all the same attributes". I am not sure if the pessimisation really happens for many attributes. I've seen it for "unsafe-fp-math" and I guess it will happen for other similar attributes. 
Probably I'm just being paranoid :) I thought e.g. the memory access attributes may not apply to the inline IR, but on second thought I guess we either don't emit them, or we shouldn't emit them for functions containing inline IR/ASM.

Let's give it the benefit of the doubt and not change it to something else. Sorry for the fuss.

PR #1451, opened 2016-04-21
install from package instructions

wilzbach:
Hi all,

I think it's quite convenient to show the users for which distribution a package is available.
Moreover it would be quite nice if on the [dlang download page](http://dlang.org/download.html) a user would be redirected to these installation instructions (e.g. then
https://github.com/ldc-developers/ldc/README.md#using-a-package) instead of the Github releases page.

PS: Did I miss a distro (you probably know this better)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,24 @@ version supporting it.
 Installation
 ------------
 
+### From a pre-built package
+
+Some Linux and OS X package managers include recent versions of LDC, so
+manually installing it might not be necessary. For several platforms,
+there are also stand-alone binary builds available at the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +25,24 @@ version supporting it.
 Installation
 ------------
 
+### Using a package
```
JohanEngelen: This confused me at first. Perhaps "binary packages" ?

JohanEngelen: Or "From a pre-built package" ?

wilzbach: changed to the later - thanks for the feedback ;-)

JohanEngelen:
Thanks!

PR #1453, opened 2016-04-21
LLVM 3.9: Function getGlobalContext() was removed.

redstar:
There are only two files in LDC which used the global context.
This PR is a hack to compile LDC with LLVM 3.9.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -67,6 +67,10 @@
 // in traits.c
 void initTraitsStringTable();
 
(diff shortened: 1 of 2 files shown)


redstar:
IMHO the right (long term) way to fix this is to create a `CodeGen` class which contains most of our LLVM related functions and has access to a local context. This class needs to be instantiated before parsing.

redstar:
Thanks. I missed that. Then I cherry-pick this commit to ltsmaster.

PR #1455, opened 2016-04-22
C++11 a few loops [NFC]

JohanEngelen:
A simple C++ification of a few loops.
(this relaxes the requirements on the `members` data structure: random indexed access is not needed, just iteration)


JohanEngelen:
Huh?
How does this break Travis only in the multilib case??

PR #1458, opened 2016-04-22
Do not squelch deprecation messages when building druntime/Phobos

klickverbot:
Upstream does not use -d (anymore) either.


d-random-contributor:
Does it make any difference? The compiler just prints the deprecation messages, right?

rainers:
It happens because the appveyor script patches the command line for std.string unittests to avoid the asserts. This fails with the modified options: https://github.com/ldc-developers/ldc/blob/ltsmaster/appveyor.yml#L103

PR #1461, opened 2016-04-26
Fix AppVeyor std.string unit test argument patching

klickverbot:
This fixes up #1458, sorry for the mess.


kinke:
Oh is that workaround still in `ltsmaster` - darn, I really hate these 2 main branches.

Issue #1460, opened 2016-04-25
libconfig.so.8 not found (ubuntu 14.04)


ikod:
Hello,

After extracting binaries from ldc2-1.0.0-beta1-linux-x86_64.tar.xz, any atempt to run ldc2 give error:
`./bin/ldc2: error while loading shared libraries: libconfig.so.8: cannot open shared object file: No such file or directory
`
workaround: use libconfig.so.8 from ldc2-0.17.1

Thanks!

joakim-brannstrom:
Same problem for me. Workaround by setting LD_LIBRARY_PATH to same as the ldc2 binary is in.
But I think the root of the problem is some change to the build system.

```sh
ldd ldc2-0.17.1-linux-x86_64/bin/ldc2
        linux-vdso.so.1 =>  (0x00007ffd7952a000)
        libconfig.so.8 => /home/user/dlang/ldc2-0.17.1-linux-x86_64/bin/libconfig.so.8 (0x00007f97be378000)

ldd ldc2-1.0.0-alpha1-linux-x86_64/bin/ldc2
        linux-vdso.so.1 =>  (0x00007fffffe08000)
        libconfig.so.8 => /home/user/dlang/ldc2-1.0.0-alpha1-linux-x86_64/bin/libconfig.so.8 (0x00007f73f731e000)

ldd ldc2-1.0.0-beta1-linux-x86_64/bin/ldc2
        linux-vdso.so.1 =>  (0x00007ffd1a2f2000)
        libconfig.so.8 => not found
```

wilzbach:
> But I think the root of the problem is some change to the build system.

Yep - Travis builds fail [too](https://github.com/travis-ci/travis-ci/issues/5952)

redstar:
Ok, I think I found the root cause. The following link flags from our build script are not used: `-DCMAKE_EXE_LINKER_FLAGS='-static-libstdc++ -Wl,-rpath,\$ORIGIN'`. This explains the dependency to g++4.8 and the need for LD_LIBRARY_PATH.

joakim-brannstrom:
Thank you :+1:

Extrawurst:
I also have this issue on travis-ci can we get a beta2 to test again with this fix ?

klickverbot:
@Extrawurst: The quickest way to get the beta would probably be to figure out a good solution to the remaining issues in #1462. ;)

Extrawurst:
@klickverbot I see, I believe in you guys :)

[...]

PR #1462, opened 2016-04-26
Append CMAKE_EXE_LINKER_FLAGS to the LDC_FLAG_LIST.

redstar:
This fixes issue #1460. It allows to pass linker flags from the commandline.


wilzbach:
Could you release a "beta2" after this is merged, so that not that many people run into this problem and testing on Travis works. 
Thanks in advance :)

redstar:
This is the right approach but only half of the way. We add `-Xlinker` to most linker flags. This bypasses the compiler driver. Unfortunately `-static-libstdc++` is handled by the compiler driver...
@wilzbach I do as soon as I have a good solution.

PR #1466, opened 2016-04-28
fix cmake generation of VS projects:

rainers:
- there is no need to select the C runtime on the command line, the C++ files already select it
- disable copying ldc2.conf to itself
- set optimization options depending on configuration

I hope this doesn't break other platforms...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,7 @@ matrix:
   allow_failures:
     - env: LLVM_CONFIG="llvm-config-3.9"
 script:
+  - cmake --version
   - cmake -DLLVM_CONFIG=$(which ${LLVM_CONFIG}) $OPTS .
   - make -j3
   # Outputs some environment info, plus makes sure we only run the test suite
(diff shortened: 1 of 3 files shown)


rainers:
> I hope this doesn't break other platforms...

Unfortunately, it does. There are failures for some builds on Travis, but not all !? What version of cmake is installed? Are they different for different LLVM versions?

redstar:
No, the cmake version is the same. Only LLVM versions varies.

rainers:
> No, the cmake version is the same. Only LLVM versions varies.

Hmmm, then I'm even more confused why some build stumble over `$<CONFIG:Debug>` while others do not.
I've added printing the cmake version to the build scripts.

rainers:
The linux machines have cmake 2.8.7, while the OSX machines have 3.0.2 (I wasn't aware that travis also builds OSX). Appveyor runs cmake 3.5.0.
Can we move requirements to cmake 3.0 or should we restrict the multiple configuration code to VS project generation?

JohanEngelen:
I also ran into trouble with the different cmake versions. Perhaps it's worth trying to up Travis's version. How common is CMake 3.0? I am using CMake 3.4 on my Mac.

redstar:
LLVM 3.5/3.6 require at least CMake 2.8.8. IMHO we should not require a later CMake version than LLVM because it could create problems with distributions.

PR #1469, opened 2016-04-30
Allow different configurations in the ldc2.conf

rainers:
This allows selecting the configuration group by target triple extracted from the command line.

This allows using the same installation for both 32-bit and 64-bit builds on Windows by just having different lib folders. Not sure how this is done on other platforms...

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -297,6 +297,23 @@ static const char *tryGetExplicitConfFile(int argc, char **argv) {
   return nullptr;
 }
 
+static llvm::Triple tryGetExplicitTriple(int argc, char **argv) {
+  llvm::Triple triple(llvm::sys::getDefaultTargetTriple());
+  for (int i = 1; i < argc; ++i) {
+    if (strcmp(argv[i], "-m32") == 0) {
+      triple = triple.get32BitArchVariant();
+    } else if (strcmp(argv[i], "-m64") == 0) {
+      triple = triple.get64BitArchVariant();
+    } else if (strcmp(argv[i], "-mtriple=") == 0) {
+      triple = llvm::Triple(llvm::Triple::normalize(argv[i] + 9));
+    } else if (strcmp(argv[i], "-march=") == 0) {
+      std::string errorMsg; // ignore error, will show up later anyway
+      lookupTarget(argv[i] + 7, triple, errorMsg); 
+    }
+  }
+  return triple;
+}
+
 /// Parses switches from the command line, any response files and the global
 /// config file and sets up global.params accordingly.
 ///
@@ -323,8 +340,9 @@ static void parseCommandLine(int argc, char **argv, Strings &sourceFiles,
 
   ConfigFile cfg_file;
   const char *explicitConfFile = tryGetExplicitConfFile(argc, argv);
+  std::string cfg_triple = tryGetExplicitTriple(argc, argv).getTriple();
   // just ignore errors for now, they are still printed
-  cfg_file.read(explicitConfFile);
+  cfg_file.read(explicitConfFile, cfg_triple.c_str());
   final_args.insert(final_args.end(), cfg_file.switches_begin(),
                     cfg_file.switches_end());
 
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -298,18 +298,26 @@ static const char *tryGetExplicitConfFile(int argc, char **argv) {
 }
 
 static llvm::Triple tryGetExplicitTriple(int argc, char **argv) {
+  // most combinations of flags are illegal, this mimicks command line
+  //  behaviour for legal ones only
   llvm::Triple triple(llvm::sys::getDefaultTargetTriple());
+  const char* mtriple = nullptr;
+  const char* march = nullptr;
   for (int i = 1; i < argc; ++i) {
-    if (strcmp(argv[i], "-m32") == 0) {
-      triple = triple.get32BitArchVariant();
-    } else if (strcmp(argv[i], "-m64") == 0) {
-      triple = triple.get64BitArchVariant();
-    } else if (strcmp(argv[i], "-mtriple=") == 0) {
-      triple = llvm::Triple(llvm::Triple::normalize(argv[i] + 9));
-    } else if (strcmp(argv[i], "-march=") == 0) {
+    if (strcmp(argv[i], "-m32") == 0)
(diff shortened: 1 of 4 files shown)


redstar:
Nice. Looks like a reasonable subset of my proposal: http://forum.dlang.org/post/bexmafwrzktfprexzejp@forum.dlang.org

rainers:
I didn't know your proposal, but this PR is a tiny step in the same direction. With just a single option I think explicitly setting it for each supported triple is not too bad for a start.

driver/main.cpp
```diff
@@ -297,6 +297,23 @@ static const char *tryGetExplicitConfFile(int argc, char **argv) {
   return nullptr;
 }
 
+static llvm::Triple tryGetExplicitTriple(int argc, char **argv) {
+  llvm::Triple triple(llvm::sys::getDefaultTargetTriple());
+  for (int i = 1; i < argc; ++i) {
```
JohanEngelen: This looks like it will return different triples for `-m32 -mtriple=johan` and `-mtriple=johan -m32` (when johan defaults to 64bit).

rainers: Yes, but it should later fail anyway with message "-m32 and -m64 switches cannot be used together with -march and -mtriple switches".

JohanEngelen: Ah thanks for the explanation.
Could you change the loop to early return on matching any of the strings? So it's a bit more self-documented.

rainers: It is actually slightly more complicated than early returns. Updated to better mimmick command line behaviour for a combination of triple and arch (though likely not too common).

kinke:
I've retriggered the 2 failed Travis jobs (LLVM 3.5 - strange file timestamp assertion in std.file:3800; the OSX Release job timed out again).

rainers:
> Hm, it just occurred to me that we should really make sure this design also works for cross-compilation use cases. Different standard libraries (e.g. compiling for ARM "bare metal"/musl as well as Linux) wouldn't necessarily be reflected in the triple, would they?

What options would select these different targets? Please note that there is also always `-conf=` to redirect to another configuration or disable all defaults.

rainers:
> the OSX Release job timed out again).

This time it failed to download dmd ;-(

kinke:
> `-conf=` to [...] disable all defaults

IIRC, it doesn't work like that - not specifying it simply uses the default config file. I tampered with or at least looked at that code some months ago. Just as a side note. ;)

rainers:
> IIRC, it doesn't work like that - specifying an empty one simply uses the default config file. I tampered with or at least looked at that code some months ago. Just as a side note. ;)

You are right, that wouldn't work ATM. But contrary to the comment in configfile.cpp this should be the behaviour to be compatible with dmd (at least for ldmd).

JohanEngelen:
@rainers Are you going to work on making one windows 32/64 package?

rainers:
> @rainers Are you going to work on making one windows 32/64 package?

I haven't looked into the release build process, but that would be the expected installation.

I usually build both libraries with MULTILIB=ON, but have to patch the generated projects or ninja rules. This is in my build batch:
```
copy rules.ninja rules.ninja.bak
sed -e "/C_COMPILER.*32/,/description/s/amd64\\cl/cl/" rules.ninja.bak >rules.ninja
```
I guess it could also be done within cmake, but it would probably need a wizard to do...

One problem with building this on AppVeyor is that it will likely exceed the build timeout to test both platforms. Maybe we can use one job to build, but test in another?

kinke:
That would entail quite some CMake and AppVeyor hacking. I think the current way isn't that bad; I don't consider not providing a smooth out-of-the-box multilib build for Windows to be a big issue, as most (current + potential) Windows users don't build LDC themselves and/or don't need the other platform.

What matters is that we are able to provide a prebuilt 32/64-bit combi package. All that's basically needed is adding the x64 lib folder under another name to the x86 archive (or the other way around if we assume everyone interested is on Win64) and hacking the lib paths in the default .conf file. That could easily be accomplished by an additional AppVeyor job (a deployment job), which downloads the 2 GitHub archives (stable download links), merges them and publishes the combi archive, or something like that.

kinke:
There's still a minor issue:

default triple for LDC x64: x86_64-pc-windows-msvc
default triple for LDC x86: i**6**86-pc-windows-msvc
triple for LDC x64 with `-m32`: i**3**86-pc-windows-msvc
triple for LDC x86 with `-m64`: x86_64-pc-windows-msvc

jamonahn:
>  Hm, it just occurred to me that we should really make sure this design also works for cross-compilation use cases. Different standard libraries (e.g. compiling for ARM "bare metal"/musl as well as Linux) wouldn't necessarily be reflected in the triple, would they?

>     What options would select these different targets? Please note that there is also always -conf= to redirect to another configuration or disable all defaults.


Don't know if this is the best place for this:

I am trying to bootstrap LDC2 onto a Docker Alpine image.  I have been able to build the ltsmaster branch and install it.  Now when I "git checkout master" or even 'git clone' to a seperate dir and build I get the following:
bash-4.3# pwd
/root/ldc
bash-4.3# git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
bash-4.3# mkdir build && cd build
bash-4.3# cmake ..
-- The C compiler identification is GNU 5.3.0
-- The CXX compiler identification is GNU 5.3.0
... ... ... ... ... ...
-- Looking for setupterm in curses
-- Looking for setupterm in curses - found

/usr/lib/gcc/x86_64-alpine-linux-musl/5.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: cmakeExtractDMDSystemLinker.o: relocation R_X86_64_32S against `ldc.dso_initialized' can not be used when making a shared object; recompile with -fPIC
cmakeExtractDMDSystemLinker.o: error adding symbols: Bad value
CMake Error at cmake/Modules/ExtractDMDSystemLinker.cmake:40 (message):
  Failed to compile empty program using D compiler '/usr/local/bin/ldmd2'
  CMakeLists.txt:546 (include)


-- Configuring incomplete, errors occurred!
See also "/root/ldc/build/CMakeFiles/CMakeOutput.log".
See also "/root/ldc/build/CMakeFiles/CMakeError.log".

Is there a '-conf=' setting that will work with musl under Alpine or is my initial 0.17 ldc2 build at fault?

PR #1470, opened 2016-04-30
Use random output filenames for `-run`.

JohanEngelen:
The temporary files created may clash with parallel processes (e.g. in a test suite where LDC is called in parallel with perhaps same input).
These files will be deleted after execution so their name is not relevant. For better linking errors, the new filenames are created by appending random characters to their name.

Note that there is no check whether the random executable file exists (same as for the normal infered executable filename). We don't want to overwrite the user's files, but I think the chances of a user having a file named "infered name + '_ldctemp' + random 8 digit hexadecimal number" are very slim.
For the object files, there is a (naive) existance check.

### DIFF
--- a/gen/module.cpp
+++ b/gen/module.cpp
@@ -42,6 +42,8 @@
 #include "ir/irtype.h"
 #include "ir/irvar.h"
 #include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
 #include "llvm/IR/Verifier.h"
 #include "llvm/LinkAllPasses.h"
 #include "llvm/Transforms/Utils/ModuleUtils.h"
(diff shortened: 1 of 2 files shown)


JohanEngelen:
(The reason for this work was issues with parallel lit-based tests that would sometimes break because of files being deleted by another thread)

d-random-contributor:
Maybe append the ldc's pid instead of a random string? Pids are guaranteed to not clash.

JohanEngelen:
> Maybe append the ldc's pid

Yeah I thought about that too. I didn't implement it because of extra platform portability mess... Let me check again if LLVM provides a portable function for PID.

d-random-contributor:
Maybe store it somewhere on the D side?

JohanEngelen:
>  I thought there should be some compiler driver that would determine file names and other code parts would just take those names and blindly use them.

The diff is in those places where that happens ;-)

ddmd/dmodule.d
```diff
@@ -1343,6 +1343,40 @@ else
...
+                        rndstr[i] = "0123456789ABCDEF"[uniform(0, 15)];
+                    }
+                    char* s = cast(char *)alloca(len + rndlen + 1 + extlen);
+                    memcpy(s, argobj, len);
+                    memcpy(s + len, rndstr.ptr, rndlen);
+                    // add ext, otherwise forceExt will remove all after first '.':
+                    s[len+rndlen] = '.';
+                    memcpy(s + len + rndlen + 1, ext, extlen + 1);
+                    // null terminate:
+                    s[len + rndlen + 1 + extlen] = 0;
```
klickverbot: Isn't this out of bounds?

JohanEngelen: Excellent catch, line 1363 needs an extra `+1`.

smolt:
I know I am late here - but is it possible to use `llvm::sys::fs::createUniqueFile` or related?

JohanEngelen:
LLVM has nice functionality for temporary files.  `createUniqueFile`, `createTemporaryFile`, and (very nice) `RemoveFileOnSignal`. I don't really remember why I did not use `createTemporaryFile`; I ended up implementing my own version of it. I think I was worried about confusing linker errors, but the first function argument specifies a prefix, so that's solved then... I'll have a look at it again.

JohanEngelen:
Different approach now using LLVM's functionality.

(I decided to keep the clang-format changes in module.cpp)

gen/module.cpp
```diff
@@ -76,43 +78,62 @@ static void check_and_add_output_file(Module *NewMod, const std::string &str) {
...
+      llvm::SmallString<128> tempFilename; // must outlive buildFilePath call
+      if (global.params.run) {
+        using namespace llvm;
+        // If `-run` is passed, the obj file is temporary and is removed
+        // after execution. Make sure the name does not collide with other files
+        // from other processes by creating a unique filename.
+        auto tmpname = objname ? objname : "tmp";
+        auto tmppath =
+            global.params.objdir
+                ? Twine(global.params.objdir) + sys::path::get_separator()
```
klickverbot: Seeing `Twine` stored on the stack like this makes me nervous. Are you sure that `Twine(global.params.objdir)` being destroyed at the end of the statement doesn't mess things up?

JohanEngelen: doesn't `Twine(...) + ""` create a new `Twine`?

klickverbot: I've always interpreted this:
```c++
  /// A Twine is not intended for use directly and should not be stored, its
  /// implementation relies on the ability to store pointers to temporary stack
  /// objects which may be deallocated at the end of a statement.
```
to mean that concatenation and so on might keep pointers to their parameters.

JohanEngelen: ah I think I know what you mean here.  The things that the Twine points to outlive the statement, so it should be fine.

klickverbot: And a brief look at the implementation seems to confirm that, although you might of course end up being lucky here with this particular order of arguments.

JohanEngelen: ahhh. hmm. maybe you are right. So this _is_ bad. Will fix. Thanks.

klickverbot: Sorry, our comments were reordered on the wire. Unless I'm mistaken, your code just works because C strings happen to be special-cased in `operator+` for efficiency. If a regular Twine instance were constructed for `sys::path::get_separator()`, it would be out of scope.

In other words, this seems to rely on an implementation detail, and is anyway confusing enough to verify that I'd rather avoid it altogether. Even just pasting it into the below expression to make a single long on would be better, I think.

JohanEngelen:
> untangling the twines

haha

[OT] Does Phobos have a "twine" type?

JohanEngelen:
Twine business is fixed. Waiting for green on CI testers.

JohanEngelen:
Green: merging.

PR #1472, opened 2016-05-02
Add @fastmath magic UDA

JohanEngelen:
Adds `@fastMath` that should enable aggressive math optimizations.

The attribute sets "fast" attributes on all floating point operations, ~~but does not set "unsafe-fp-math". I think it _should_ set "unsafe-fp-math" though.~~ and it sets unsafe-fp-math=true. This will also enable reassociation optimizations.  ~~(`@fastMath(false)` will do the opposite)~~

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -822,6 +822,13 @@ void DtoDefineFunction(FuncDeclaration *fd) {
   // assert(gIR->scopes.empty());
   gIR->scopes.push_back(IRScope(beginbb));
 
+  // Set the FastMath options for this function scope.
+#if LDC_LLVM_VER >= 308
+  gIR->scopes.back().builder.setFastMathFlags(irFunc->FMF);
+#else
+  gIR->scopes.back().builder.SetFastMathFlags(irFunc->FMF);
+#endif
+
   // create alloca point
   // this gets erased when the function is complete, so alignment etc does not
   // matter at all
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -406,11 +406,26 @@ void DtoResolveFunction(FuncDeclaration *fdecl) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+namespace {
+
 void applyParamAttrsToLLFunc(TypeFunction *f, IrFuncTy &irFty,
                              llvm::Function *func) {
   AttrSet newAttrs = AttrSet::extractFunctionAndReturnAttributes(func);
   newAttrs.merge(irFty.getParamAttrs(gABI->passThisBeforeSret(f)));
   func->setAttributes(newAttrs);
+}
+
+void applyDefaultMathAttributes(IrFunction *irFunc) {
+  // TODO: implement commandline switches to change the default values.
+
+  // "unsafe-fp-math" is not properly reset in LLVM between function definitions,
+  // i.e. if a function does not define a value for "unsafe-fp-math" it will be
+  // compiled using the value of the previous function. Therefore, each function
+  // must explicitly define the value (clang does the same).
+  // See https://llvm.org/bugs/show_bug.cgi?id=23172
+  irFunc->func->addFnAttr("unsafe-fp-math", "false");
+}
+
 }
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -496,6 +511,9 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
     }
   }
 
+  // Set default math function attributes here, such that they can be overridden
+  // by UDAs.
+  applyDefaultMathAttributes(irFunc);
   applyFuncDeclUDAs(fdecl, irFunc);
 
   // main
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -496,7 +496,7 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
     }
   }
 
-  applyFuncDeclUDAs(fdecl, func);
+  applyFuncDeclUDAs(fdecl, irFunc);
 
   // main
   if (fdecl->isMain()) {
--- a/tests/codegen/attr_fastmath.d
+++ b/tests/codegen/attr_fastmath.d
@@ -1,0 +1,53 @@
+// Test @fastmath
+
+// RUN: %ldc -O0 -release -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+import ldc.attributes;
+
+// CHECK-LABEL: define{{.*}} @notfast
+// CHECK-SAME: #[[ATTR_NOTFAST:[0-9]+]]
+extern (C) double notfast(double a, double b)
+{
+    @fastmath
+    double nested_fast(double a, double b)
+    {
+        return a * b;
+    }
+
+// CHECK-NOT: fmul fast
+    return a * b;
+}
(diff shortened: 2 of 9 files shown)


JohanEngelen:
Related issue: #1438 

@9il What do you think?

9il:
LGTM

tests/codegen/attr_fastmath.d
```diff
@@ -0,0 +1,51 @@
+// Test vectorization of simple constructs
+
+// RUN: %ldc -O0 -release -c -output-ll -of=%t.ll %s && FileCheck %s --check-prefix LLVM < %t.ll
+// RUN: %ldc -mtriple x86_64-linux-gnu -mattr=+avx2,+fma -O3 -release -c -output-s -of=%t.s %s && FileCheck %s --check-prefix ASM < %t.s
```
9il: `fma` should be already available for avx2

tests/codegen/attr_fastmath.d
```diff
@@ -0,0 +1,51 @@
...
+    }
+// LLVM: fmul fast
+    return a * b;
+}
+
+
+// LLVM-LABEL: define{{.*}} @dot
+// ASM-LABEL: dot:
+@fastMath(true)
+@target("avx2,fma")
```
9il: `fma` should be already available for `avx2`

tests/codegen/attr_fastmath.d
```diff
@@ -0,0 +1,51 @@
+// Test vectorization of simple constructs
+
+// RUN: %ldc -O0 -release -c -output-ll -of=%t.ll %s && FileCheck %s --check-prefix LLVM < %t.ll
+// RUN: %ldc -mtriple x86_64-linux-gnu -mattr=+avx2,+fma -O3 -release -c -output-s -of=%t.s %s && FileCheck %s --check-prefix ASM < %t.s
+
+import ldc.attributes;
+pragma(LDC_inline_ir) R inlineIR(string s, R, P...)(P);
```
9il: unused

JohanEngelen:
@9il Thanks for the comments, but the tests are still a work-in-progress :-)

Is the attribute useable if I have it enable all math optimizations? LLVM will reassociate freely, which I think is necessary for FMA+vectorization of `dot`.

Also, what about the name? "fastMath", is that ok?

9il:
> Is the attribute useable if I have it enable all math optimizations?

Yes



> Also, what about the name? "fastMath", is that ok?

`@fastmath` looks better for me. Lower case is clear for basic function attributes, IMHO

JohanEngelen:
@redstar @klickverbot `fastMath` or `fastmath` ?

smolt:
fastmath looks better to me too.

JohanEngelen:
Another thought I had is to perhaps namespace math attribs, such for future other math-related attribs?
e.g. `@ldc.attributes.math.fast`.  Or perhaps just prefix all math attribs with math: `@mathFast`, `@mathNoNaN`. There are not many math attribs I can think of right now though.

gen/functions.cpp
```diff
@@ -406,13 +406,27 @@ void DtoResolveFunction(FuncDeclaration *fdecl) {
...
                              llvm::Function *func) {
   AttrSet newAttrs = AttrSet::extractFunctionAndReturnAttributes(func);
   newAttrs.merge(irFty.getParamAttrs(gABI->passThisBeforeSret(f)));
   func->setAttributes(newAttrs);
 }
 
+void applyDefaultMathAttributes(IrFunction *irFunc) {
+  // TODO: implement commandline switches to change the default values.
+
+  // "unsafe-fp-math" is not properly reset in LLVM between function definitions,
```
klickverbot: Do you have an LLVM PR number for this?

JohanEngelen: I do: https://llvm.org/bugs/show_bug.cgi?id=23172

Add it to the code comments?

klickverbot: Yes, please – that or the commit message. ;)

JohanEngelen: Will do. Good point.

JohanEngelen:
Any more opinions on meaning? :)
2 votes for `fastmath`
I am leaning towards `mathFast` at the moment.

9il:
> 2 votes for fastmath

This is something that can be a part of Dspec

 

> I am leaning towards mathFast at the moment.

For llvm specific attributes: `fp("fast")`

JohanEngelen:
@ibuclaw Does `@fastmath` sound OK for possibly future GDC too?

ibuclaw:
@JohanEngelen - you can have whatever backend-specific attributes you want. ;-)

JohanEngelen:
Changed it to `@fastmath`. (lowercase)

Temtaime:
I'm against that @fastmath because of attribute hell and because it's not a solution, but a workaround.
There is(?) a flag for fast math for entire compilation unit. Why have that be function dependent ?

Temtaime:
Alors there should be an one :)
GCC, Clang have --fast-math, why not ldc ?

9il:
> Alors there should be an one :)
> GCC, Clang have --fast-math, why not ldc ?

This is deprecated approach. Mast flags should be a part of the language because in the same unit you may need both types of math. Furthermore this allows to write fast generic BLAS without assembler

Temtaime:
Yes, it solves the problem, but it's strange that one program in some parts of it violates IEEE and in some parts does not. IMO command line switch is more consistent. Maybe -Ofast as it's done in clang for aggressive optimizations.

JohanEngelen:
Very much agree with having this as a localized attribute instead of a compiler flag.

One comment:
> (you can still just apply @fastMath to everything if you want).

Unfortunately, it's not that easy (or perhaps it is and I don't know how!). I found out it's the way UDAs work: they don't apply to nested functions (see the test case). So a `@fastmath():` at the top of your program is not going to result in `@fastmath` nested functions :/

JohanEngelen:
@9il @klickverbot Good to merge?

JohanEngelen:
I was wondering about how to make a UDA that combines a set of UDAs, thanks for teaching me :)
Keep it for future work?

JohanEngelen:
Ah, I see, indeed.    I am not sure if `@fastmath(false)` is useful enough. I could simply remove the bool param for now. Let me think about it a little.
(I could add another attribute to only set the FMF for float ops, and then `alias fastmath = AliasSeq!(llvmAttr("unsafe-fp-math","true"), llvmFMF("fast"))`  but it is quite specific to current LLVM stuff that may change in future? What I mean is: llvmAttr and FMF should not be used by the user directly I think)

9il:
maybe `@normalmath` ?

btw, how can I set align [parameter attribute](http://llvm.org/docs/LangRef.html#parameter-attributes) for pointer alignment? I need this for BLAS too

9il:
> btw, how can I set align parameter attribute for pointer alignment? I need this for BLAS too

looks like  `__vector` types do what i need

JohanEngelen:
Changed the PR: it now adds `@llvmFMF` that sets LLVM's fast-math flags on floating point ops. And `@fastmath = @llvmAttr("unsafe-fp-math","true") + @llvmFMF("fast")`.

JohanEngelen:
Ping. Mergable?

kinke:
Yep, AppVeyor issues are due to the LLVM regression.

p0nce:
Is there a way to have fastmath as a global compiler switch?

PR #1475, opened 2016-05-03
[AppVeyor] Use LLVM snapshot build

JohanEngelen:
Created a PR to see if this works.

Use LLVM snapshot build installers on AppVeyor, so LLVM is up-to-date with LLVM trunk without the need for us to build it ourselves.


JohanEngelen:
Never mind, the LLVM package appears to only have clang binaries + a few other things in it. It is not an LLVM-dev package...

PR #1477, opened 2016-05-06
[Travis] Make sure we test at least one build with an C++-buildable LDC

JohanEngelen:
To test that there is a bootstrapping sequence with 1 intermediate C++-based D compiler
(instead of having to hop several compiler versions due to D features used that are not available in the C++-based version).


JohanEngelen:
The LDC 0.17 name was wrong in travis.yml, doh!

PR #1479, opened 2016-05-06
Remove TupleDeclaration.semantic3

klickverbot:
It was added ages ago to work around an issue with tuples not being
expanded in function arguments and/or variable declarations, but
it doesn't seem to be required anymore.


JohanEngelen:
Nice

PR #1480, opened 2016-05-06
Update frontend to the final 2.071 release

klickverbot:
There were no druntime/Phobos changes between b1 and the final version.


kinke:
So only `runnable/inner.d` to be fixed for all platforms? Niiice.

Issue #1473, opened 2016-05-02
ICE with classes derived from IUnknown


rainers:
If a class is derived from a class that implement IUnknown:
```
/*extern(Windows)*/ interface IUnknown {}

class ComObject : IUnknown {}

class DComObject : ComObject {}
```
LDC aborts with an ICE:
```
llvm::GlobalValue::isDeclaration() + 0x4 bytes(s), c:\s\d\ldc\llvm\lib\ir\globals.cpp, line 158
llvm::GlobalVariable::setInitializer() + 0x68 bytes(s), c:\s\d\ldc\llvm\lib\ir\globals.cpp, line 282 + 0x5 byte(s)
IrAggr::getClassInfoInterfaces() + 0x33F bytes(s), c:\s\d\ldc\ldc\ir\irclass.cpp, line 537
DtoDefineClassInfo() + 0x19B bytes(s), c:\s\d\ldc\ldc\gen\classes.cpp, line 628 + 0x8 byte(s)
IrAggr::getClassInfoInit() + 0x26 bytes(s), c:\s\d\ldc\ldc\ir\irclass.cpp, line 267 + 0x14 byte(s)
CodegenVisitor::visit() + 0x18E bytes(s), c:\s\d\ldc\ldc\gen\declarations.cpp, line 270 + 0xB byte(s)
ddmd.dclass.ClassDeclaration.accept() + 0x1D bytes(s), C:\s\d\ldc\ldc\ddmd\dclass.d, line 1431 + 0x1 byte(s)
Declaration_codegen() + 0x2A bytes(s), c:\s\d\ldc\ldc\gen\declarations.cpp, line 567
codegenModule() + 0x13D bytes(s), c:\s\d\ldc\ldc\gen\module.cpp, line 701 + 0xD byte(s)
```

Happens with both LDC 1.0.0-beta1 and merge-2.071.

kinke:
Uh that's quite crucial. :)

rainers:
https://github.com/ldc-developers/ldc/pull/1481

JohanEngelen:
@rainers Did #1481 fix this? Could you add a test for it and then close the issue?

lit-based test like this should do it:
```D
// RUN: %ldc -c %s

extern(Windows) interface IUnknown {}
class ComObject : IUnknown {}
class DComObject : ComObject {}
```

rainers:
> @rainers Did #1481 fix this? 

Yes. 

> Could you add a test for it and then close the issue?

[...]

PR #1481, opened 2016-05-08
fix issue #1473: ICE when implementing extern(C++) interfaces

rainers:
The vtbl still contained the class info entry for C++ interfaces.

### DIFF
--- a/ir/irclass.cpp
+++ b/ir/irclass.cpp
@@ -548,8 +548,9 @@ LLConstant *IrAggr::getClassInfoInterfaces() {
 
   // create and apply initializer
   LLConstant *arr = LLConstantArray::get(array_type, constants);
-  classInterfacesArray->setInitializer(arr);
-  setLinkage(cd, classInterfacesArray);
+  auto ciarr = getInterfaceArraySymbol();
+  ciarr->setInitializer(arr);
+  setLinkage(cd, ciarr);
 
   // return null, only baseclass provide interfaces
   if (cd->vtblInterfaces->dim == 0) {
@@ -563,9 +564,9 @@ LLConstant *IrAggr::getClassInfoInterfaces() {
 
   LLConstant *ptr = llvm::ConstantExpr::getGetElementPtr(
 #if LDC_LLVM_VER >= 307
-      isaPointer(classInterfacesArray)->getElementType(),
+      isaPointer(ciarr)->getElementType(),
 #endif
-      classInterfacesArray, idxs, true);
+      ciarr, idxs, true);
 
   // return as a slice
   return DtoConstSlice(DtoConstSize_t(cd->vtblInterfaces->dim), ptr);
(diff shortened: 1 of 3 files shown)


ir/irclass.cpp
```diff
@@ -548,8 +548,9 @@ LLConstant *IrAggr::getClassInfoInterfaces() {
 
   // create and apply initializer
   LLConstant *arr = LLConstantArray::get(array_type, constants);
-  classInterfacesArray->setInitializer(arr);
-  setLinkage(cd, classInterfacesArray);
+  auto ciarr = getInterfaceArraySymbol();
```
JohanEngelen: The changes in this function look like they are only because `classInterfacesArray` may still be `nullptr` at this point? Can you add a comment about that?
(you could also keep the classInterfacesArray name or another longer name; a little clearer than `ciarr`, but that's perhaps just me)

ir/irtypeclass.cpp
```diff
@@ -56,9 +56,9 @@ void IrTypeClass::addClassData(AggrTypeBuilder &builder,
 
       // add to the interface map
       addInterfaceToMap(b->sym, builder.currentFieldIndex());
-
-      const auto ivtblType = llvm::StructType::get(
-          gIR->context(), buildVtblType(interfacePtrType, &arr));
+      Type* first = b->sym->isCPPinterface() ? nullptr : interfacePtrType;
```
JohanEngelen: can you add on line 43 a comment about what the first pointer will be in case of extern(C++)?

ir/irtypeclass.cpp
```diff
@@ -141,13 +143,15 @@ IrTypeClass::buildVtblType(Type *first, FuncDeclarations *vtbl_array) {
   std::vector<llvm::Type *> types;
   types.reserve(vtbl_array->dim);
 
-  // first comes the classinfo
-  if (!cd->isCPPclass() && !cd->isCPPinterface()) {
+  auto I = vtbl_array->begin();
+  // first comes the classinfo for D interfaces
+  if (first) {
```
JohanEngelen: can you add some function documentation on line 137 about that `first` is allowed to be `nullptr` and what happens in that case?

ir/irclass.cpp
rainers: Yes, it can still be nullptr for C++ classes. I rather think not using the accessor function needs a comment why classInterfacesArray is supposed to be initialized ;-)
Just calling a getter function and then using the initialized cached member is a bit strange, so I preferred using a local variable. I agree `ciarr` isn't the best name, though.

ir/irtypeclass.cpp
rainers: Done.

ir/irtypeclass.cpp
rainers: I added a short info to the declaration in the header.

ir/irclass.cpp
JohanEngelen: > Just calling a getter function and then using the initialized cached member is a bit strange

Yes, very :-)

rainers:
Hmmm, this fails for the OSX testers. Do C++ interfaces have a different layout there?

rainers:
> Hmmm, this fails for the OSX testers. Do C++ interfaces have a different layout there?

It failed because the OSX and Windows builds have assertions enabled, the others seem to run without. Now passes but one of the OSX builds hit the build timeout.

Issue #1482, opened 2016-05-10
operations on integer SIMD vectors


9il:
LDC allows to multiply  vectors composed of `ushort`, but `uint` and `ulong` are not allowed. AVX-512 has `vpmullq` and `vpaddq` instructions, which are supported in LLVM.

---

``` d
void ok(ref __vector(ushort[16]) a)
{
    a = a * a;
}
```

``` asm
    .cfi_startproc
    vmovdqu16   (%rdi), %ymm0
    vpmullw %ymm0, %ymm0, %ymm0
    vmovdqu16   %ymm0, (%rdi)
    retq
```

---

``` d
void fail(ref __vector(ulong[4]) a)
{
    a = a * a;
}
```

```
  avx                   - Enable AVX instructions.
  avx2                  - Enable AVX2 instructions.
  avx512bw              - Enable AVX-512 Byte and Word Instructions.
  avx512cd              - Enable AVX-512 Conflict Detection Instructions.
  avx512dq              - Enable AVX-512 Doubleword and Quadword Instructions.
  avx512er              - Enable AVX-512 Exponential and Reciprocal Instructions.
  avx512f               - Enable AVX-512 instructions.
  avx512pf              - Enable AVX-512 PreFetch Instructions.
  avx512vl              - Enable AVX-512 Vector Length eXtensions.
```

JohanEngelen:
may sound stupid but: small code example please! ;)

9il:
done)

JohanEngelen:
There is a hardcoded error upon anything other than multiplying vectors of ushort or short... Is there any reason why you'd want an error on code like this?
```D
void ok(ref __vector(uint[64]) a)
{
    a = a * a;
}
```
Sure, it won't be a nice single instruction multiply but it will work fine. 
Working on a PR.

JohanEngelen:
There are some more hardcoded restrictions on operations with __vector, so you may run into more troubles. Just open a new issue for them ;)

redstar:
Please be aware: the restrictions are by design! We should not change the language in an uncontrolled way. See my comment in the PR. We need a generic solution here, not just LDC hacks.

JohanEngelen:
What was the design rationale? (is this one other case of "DMD is the spec" ?)
I would expect the code to run on any processor, regardless of whether it has special instructions for it. GCC spec was what I was expecting:
[...]

PR #1485, opened 2016-05-11
[WIP] Disable DMD-specific limitation on __vector multiplication.

JohanEngelen:
Resolves issue #1482

### DIFF
--- a/tests/codegen/vector_multiply_gh1482.d
+++ b/tests/codegen/vector_multiply_gh1482.d
@@ -1,0 +1,24 @@
+// Tests integer SIMD vectorization
+
+// RUN: %ldc -mcpu=haswell -c -output-ll %s -of=%t.ll && FileCheck %s --check-prefix LLVM < %t.ll
+// RUN: %ldc -mcpu=haswell -c -output-s  %s -of=%t.s  && FileCheck %s --check-prefix ASM  < %t.s
+
+void mulshorts(ref __vector(ushort[16]) a)
+{
+    // LLVM: mul <16 x i16>
+    // ASM: vpmullw
+    a = a * a;
+}
+
+void mulints(ref __vector(uint[8]) a)
+{
+    // LLVM: mul <8 x i32>
+    // ASM: vpmulld
+    a = a * a;
+}
+
+void mullongs(ref __vector(ulong[4]) a)
+{
+    // LLVM: mul <4 x i64>
+    a = a * a;
+}
--- a/ddmd/expression.d
+++ b/ddmd/expression.d
@@ -13931,8 +13931,15 @@ public:
         }
         else if (tb.ty == Tvector && (cast(TypeVector)tb).elementType().size(loc) != 2)
         {
+version(IN_LLVM)
+{
+            // LLVM supports arbitrary vector sizes, so all is good here.
+}
+else
+{
             // Only short[8] and ushort[8] work with multiply
             return incompatibleTypes();
+}
         }
         return this;
     }


tests/codegen/vector_multiply_gh1482.d
```diff
@@ -0,0 +1,24 @@
...
+{
+    // LLVM: mul <8 x i32>
+    // ASM: vpmulld
+    a = a * a;
+}
+
+void mullongs(ref __vector(ulong[4]) a)
+{
+    // LLVM: mul <4 x i64>
+    a = a * a;
```
JohanEngelen: my LLVM 3.8 did not optimize this to single instruction multiply (with -mcpu=knl), so not check on the ASM output.

redstar:
This is not a solution which will be accepted upstream.
The problem here is that the DMD SIMD facility is designed to allow only those operators which are supported by the host CPU (*). DMD does not support AVX2 and therefore disallows these operations.
What we really need here is a customized way to define which vector types are supported by which operation. For LDC we would allow most types (because we don't know what the target supports), DMD would disallow a lot of them.

(*) This was the goal. Some operations were missed, e.g. left and right shifting an integer vector by a scalar is disallowed despite the fact that there is an SSE instruction for it.

JohanEngelen:
> This is not a solution which will be accepted upstream.

It wasn't intended for that.

It is very disappointing to be so limited by what DMD supports.

redstar:
What about my proposed solution in the issue?

JohanEngelen:
> What about my proposed solution in the issue?

Let's go with that.

MartinNowak:
Also see https://github.com/dlang/dmd/pull/6723 @redstar.

MartinNowak:
BTW, clang/llvm and gcc do a lot of emulation for missing SIMD ops, we maybe should decide where to draw the line, as it's not to nice if a seemingly supported vector ops are emulated with scalar instructions.
```c++
typedef int v4si __attribute__ ((vector_size (16)));
v4si test(v4si a, v4si b)
{
    return a / b;
}
```
```
0000000000000000 <_Z4testDv4_iS_>:
   0:	55                   	push   %rbp
   1:	48 89 e5             	mov    %rsp,%rbp
   4:	c5 f8 29 45 f0       	vmovaps %xmm0,-0x10(%rbp)
   9:	c5 f8 29 4d e0       	vmovaps %xmm1,-0x20(%rbp)
   e:	c5 f8 28 45 f0       	vmovaps -0x10(%rbp),%xmm0
  13:	c5 f8 28 d1          	vmovaps %xmm1,%xmm2
  17:	c4 e3 79 16 c8 01    	vpextrd $0x1,%xmm1,%eax
  1d:	c4 e3 79 16 c1 01    	vpextrd $0x1,%xmm0,%ecx
  23:	89 45 dc             	mov    %eax,-0x24(%rbp)
  26:	89 c8                	mov    %ecx,%eax
  28:	99                   	cltd   
  29:	8b 4d dc             	mov    -0x24(%rbp),%ecx
  2c:	f7 f9                	idiv   %ecx
  2e:	c5 f9 7e ca          	vmovd  %xmm1,%edx
  32:	c5 f9 7e c6          	vmovd  %xmm0,%esi
  36:	89 45 d8             	mov    %eax,-0x28(%rbp)
  39:	89 f0                	mov    %esi,%eax
  3b:	89 55 d4             	mov    %edx,-0x2c(%rbp)
  3e:	99                   	cltd   
  3f:	8b 75 d4             	mov    -0x2c(%rbp),%esi
  42:	f7 fe                	idiv   %esi
  44:	c5 f9 6e d8          	vmovd  %eax,%xmm3
  48:	8b 45 d8             	mov    -0x28(%rbp),%eax
  4b:	c4 e3 61 22 d8 01    	vpinsrd $0x1,%eax,%xmm3,%xmm3
  51:	c4 e3 79 16 ca 02    	vpextrd $0x2,%xmm1,%edx
  57:	c4 e3 79 16 c0 02    	vpextrd $0x2,%xmm0,%eax
  5d:	89 55 d0             	mov    %edx,-0x30(%rbp)
  60:	99                   	cltd   
  61:	8b 7d d0             	mov    -0x30(%rbp),%edi
  64:	f7 ff                	idiv   %edi
  66:	c4 e3 61 22 d8 02    	vpinsrd $0x2,%eax,%xmm3,%xmm3
  6c:	c4 e3 79 16 c8 03    	vpextrd $0x3,%xmm1,%eax
  72:	c4 e3 79 16 c2 03    	vpextrd $0x3,%xmm0,%edx
  78:	89 45 cc             	mov    %eax,-0x34(%rbp)
  7b:	89 d0                	mov    %edx,%eax
  7d:	99                   	cltd   
  7e:	44 8b 45 cc          	mov    -0x34(%rbp),%r8d
  82:	41 f7 f8             	idiv   %r8d
  85:	c4 e3 61 22 c0 03    	vpinsrd $0x3,%eax,%xmm3,%xmm0
  8b:	c5 f8 29 55 b0       	vmovaps %xmm2,-0x50(%rbp)
  90:	5d                   	pop    %rbp
  91:	c3                   	retq   
```

thewilsonator:
Well it depends on what view you take. 
DMD takes the view that 'only support what vector operation are natively supported because efficiency'. 
 Whereas LLVM and GCC take the view that you're doing that anyway no point in preventing you from expressing that. If it can be expressed in a single vector operation, good, otherwise recursively break the vector down from `__vector(T[N])` into `__vector(T[N/2])[2]`.
 Particularly for dcompute being able to express all of the vector operations is crucial some of which may not be supported by the host, which doesn't matter because it's for the target, and we do all in one compilation not separate host + target.

MartinNowak:
> DMD takes the view that 'only support what vector operation are natively supported because efficiency'.
Whereas LLVM and GCC take the view that you're doing that anyway no point in preventing you from expressing that.

Well doing an unsupported op like that is a major (performance) bug.
If it were `ubyte[16] *= ubyte[16]` which requires 2 or 3 SIMD ops, that'd be fine. But just allowing anything, even if it's horribly slow, means we don't have a static introspection to use fallback code.
https://github.com/MartinNowak/druntime/blob/3701b4758904341f2c94073fd065e82c70ff1880/src/core/internal/arrayop.d#L22

```d
void foo(int[] a, int[] b)
{
    auto va = cast(int4[]) a, vb = cast(int4[]) b;
    foreach (i, ref v; va)
        v /= vb[i];
}
```

> Particularly for dcompute being able to express all of the vector operations is crucial some of which may not be supported by the host

Host support doesn't matter at all, that's why the code is in Target ;).

thewilsonator:
> Well doing an unsupported op like that is a major (performance) bug.

I think you underestimate the ability of LLVM's and GCC's vectorisers.

> Host support doesn't matter at all, that's why the code is in Target ;).

Sorry I meant host as in Target and Target as in the set of arch's (possibly plural) that we are building for i.e. {host, NVPTX (multiple arch versions),SPIR-V (multiple OpenCL versions)}.
I have a very simple compiler hack for detecting which target is being targeted (essentially as "AST -> IR" time if), but it can't be used for frontend `static if` decisions. Thus the need to support all vector operations, exotic cross-type operation excluded.

JohanEngelen:
> as it's not to nice if a seemingly supported vector ops are emulated with scalar instructions.

This doesn't make sense to me.  You are proposing to disable (normal) division when the target CPU does not support it, or to disable `ulong` if the CPU cannot d0 64 bit operations. And we should probably remove `cent` from the language altogether.

If you want to check whether the target supports vectorized instruction or not, a compile error is not the way imho.

> But just allowing anything, even if it's horribly slow, 

What does the fast solution look like?

PR #1487, opened 2016-05-12
trying to figure the failure on AppVeyor for x86

rainers:
Please do not merge!

Is this how a phobos PR can be tested on the build servers?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -68,7 +68,7 @@ install:
   - python -m pip install lit
   - python -c "import lit; lit.main();" --version
   # Download & extract DMD
-  - ps: Start-FileDownload 'http://downloads.dlang.org/releases/2015/dmd.2.069.2.windows.7z' -FileName 'dmd2.7z'
+  - ps: Start-FileDownload 'http://nightlies.dlang.org/dmd-nightly/dmd.master.windows.7z' -FileName 'dmd2.7z'
   - 7z x dmd2.7z > nul
   - set DMD=%CD%\dmd2\windows\bin\dmd.exe
   - dmd2\windows\bin\dmd.exe --version


JohanEngelen:
@rainers What I do to check a druntime/phobos change is: create a phobos branch and push it to LDC's (not your own) repository. Then create an LDC PR that changes the submodule pointer to that branch.

rainers:
@JohanEngelen Yeah, I also considered this, but didn't want to pollute the main repository (even if only temporary, but with a chance to screw up badly deleting the branch ;-) ). Patching appveyor.yml to merge the phobos PR works for my use case, too.

rainers:
I finally figured it out: this is a codegen bug of the 32-bit build of LDC caused by a codegen bug in dmd used to compile it. Here is a reduced test:
```
bool _testr(real x) { return x == 2147483646.5; }

unittest
{
    assert(_testr(int.max - 0.5));
}
```
Fortunately it is already fixed:https://github.com/dlang/dmd/pull/5650

I can imagine 3 solutions:
- use a newer dmd: as the fix is not released yet that means a nightly build
- build LDC with itself again before running the tests
- ditch the 32-bit build and use multilib building. To do this we have to patch the cmake generated files to use the proper compiler. A larger downside will probably be that the tests will run too long and cause timeouts if tests for both architectures are executed. Having multiple configurations as in 
https://github.com/ldc-developers/ldc/pull/1469 would be useful.

JohanEngelen:
Great work figuring this out!

JohanEngelen:
@rainers Green on AppVeyor! Ready to merge? (I think for now this is the best option out of the three)

rainers:
> Green on AppVeyor! Ready to merge? (I think for now this is the best option out of the three)

Yes, readyto merge. I also think using a newer dmd is the simplest option, but nightlies might be unstable. We can later switch to the announced 2.071.1, the changelog also shows the bugfix included.

JohanEngelen:
> We can later switch to the announced 2.071.1, the changelog also shows the bugfix included.

Excellent.

Again: great work! Awesome that all CI testers are green now.

rainers:
> Awesome that all CI testers are green now.

Thanks. Now we have to make sure the build timeouts don't cause random failures ;-)

PR #1497, opened 2016-05-18
[Travis] Switch to LLVM 3.8 for OS X.

JohanEngelen:
Switch to LLVM 3.8 because 3.8.0 was used for the beta release.


JohanEngelen:
This also seems to reduce the build time of the OS X release Travis job by several minutes, so perhaps we won't be hitting the timeout so often any more.

PR #1498, opened 2016-05-18
Add new attribute visualcppmangle.

redstar:
This attribute allows to influence the C++ mangling for VS. This is
necessary because VS has different manglings for structs and classes.
gcc uses the same mangling instead.

### DIFF
--- a/ddmd/aggregate.h
+++ b/ddmd/aggregate.h
@@ -178,6 +178,9 @@ public:
     // (e.g. TypeidExp, NewExp, ArrayLiteralExp, etc) request its TypeInfo.
     // For those, today TypeInfo_Struct is generated in COMDAT.
     bool requestTypeInfo;
+#if IN_LLVM
+    bool cppmangleAsClass;      // true if this struct should be mangled as class (VS only)
+#endif
 
     StructDeclaration(Loc loc, Identifier *id);
     Dsymbol *syntaxCopy(Dsymbol *s);
@@ -273,6 +276,9 @@ public:
     bool cpp;                           // true if this is a C++ interface
     bool isscope;                       // true if this is a scope class
     bool isabstract;                    // true if abstract class
+#if IN_LLVM
+    bool cppmangleAsStruct;             // true if this class should be mangled as struct (VS only)
+#endif
     int inuse;                          // to prevent recursive attempts
     Baseok baseok;                      // set the progress of base classes resolving
     Objc_ClassDeclaration objc;
--- a/ddmd/attrib.d
+++ b/ddmd/attrib.d
@@ -1529,6 +1529,24 @@ extern (C++) static uint setMangleOverride(Dsymbol s, char* sym)
         s.isDeclaration().mangleOverride = sym;
         return 1;
     }
+//version(IN_LLVM) {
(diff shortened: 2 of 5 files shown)


gen/uda.cpp
```diff
@@ -268,3 +269,63 @@ bool hasWeakUDA(Dsymbol *sym) {
...
+  Expressions *attrs = decl->userAttribDecl->getAttributes();
+  expandTuples(attrs);
+  for (auto &attr : *attrs) {
+    auto sle = getLdcAttributesStruct(attr);
+    if (!sle)
+      continue;
+
+    auto name = sle->sd->ident->string;
+    if (name == attr::visualcppmangle) {
+      checkStructElems(sle, { Type::tstring });
```
JohanEngelen: Could you put lines 316-323 into its own function? (like e.g. `applyAttrSection(sle, func);`)

gen/uda.h
```diff
@@ -18,13 +18,17 @@
...
+class StructDeclaration;
+class ClassDeclaration;
 namespace llvm {
 class Function;
 class GlobalVariable;
 }
 
 void applyFuncDeclUDAs(FuncDeclaration *decl, llvm::Function *func);
 void applyVarDeclUDAs(VarDeclaration *decl, llvm::GlobalVariable *gvar);
+void applyStructUDA(StructDeclaration *sd);
```
JohanEngelen: Perhaps `applyStructUDA` --> `applyStructUDAs` with the extra `s` like the 2 functions above it.

JohanEngelen:
Most important: add tests for this! ;-)

JohanEngelen:
How about `@mangleAsCpp("class")` ?  Although afaik this is MSVC specific, I think the problem is that C++ source reads `class` and in D we want to write `struct`, right?

yebblies:
Tests?  Can we use `pragma(mangle)` for this?

redstar:
Does anybody have a good syntax proposal for `pragma(mangle)`? Using a `string` as second argument is not possible. I can think of `pragma(mangle, asclass)` or `pragma(mangletype, "class")`.

Test follow as soon as the VS build is working again.

yebblies:
I was thinking of `pragma(mangle, cppClassMangling("MyClass"))` - the magic is taken care of in the library function.

redstar:
I'm not sure if I understand your idea. Calling a library function (does `cppClassMangling()` exist in druntime?) requires to pass the complete signature, including for which environment the mangling should be done. I only want to influence the existing mangling a bit.

JohanEngelen:
Some notes:
pragma(mangle) currently does nothing for class/struct type definitions, so it could be overloaded for this use case  (e.g. `pragma(mangle,"!cppclass")`). Although I think it is much better to introduce a new pragma with one very clear purpose of telling the compiler that an `extern(C++) class` (or `struct) is actually a C++ `struct` (or `class).

This has to work for templated types too. I think it is easier and cleaner to do with Kai's flag system than a druntime library solution.

redstar:
See https://github.com/dlang/dmd/pull/5809, including template mangling.

redstar:
Yes, fur sure!

redstar:
Next try: https://github.com/dlang/dmd/pull/5875

PR #1503, opened 2016-05-19
Update phobos submodule.

JohanEngelen:
@joakim-noah I saw that you pushed to the `ldc` branch of phobos but didn't update LDC to it. Made a PR for that.


joakim-noah:
Yeah, it's not too important, as it's not officially supported yet.  I figured somebody else would commit something and pull these in too.

PR #1504, opened 2016-05-19
LLVM: Use /LIBPATH for MSVC

redstar:
For MSVC each LLVM library was prefixed with the absolute path of the
library location. This leads to problems with the command line length.

This PR uses "/LIBPATH:..." as link flag for the LLVM libraries and
only appends the suffix .lib to the library names.


redstar:
Might be a solution. I am slowly improving the CMake description for VS2015. I put this on my todo list.

redstar:
Invoking ldc to compile the ddmd code and link all together.

PR #1507, opened 2016-05-20
LLVM/MSVC: Use library path to search for LLVM libraries.

redstar:
For MSVC each LLVM library was prefixed with the absolute path of the
library location. This leads to problems with the command line length.

This PR uses "-L..." as link flag for the LLVM libraries and
only appends the suffix .lib to the library names.

It is important to handle the command line options in the right way.
Otherwise spaces in paths may be lost.


JohanEngelen:
Thanks for the persistence! :)

redstar:
I use VS2015 for development. I _need_ a working Windows build.

JohanEngelen:
So you are not a ninja? ;-)

Sorry I broke the VS builds :(

redstar:
I use ninja only for the release build (much faster).

I guess there are plenty of problems with spaces in path. Only my LLVM installation has spaces in the path. All other parts are currently unquoted.

PR #1510, opened 2016-05-22
Document what -O4/-O5 are actually doing (-O3)

klickverbot:


JohanEngelen:
I think it's not a good idea to update the APT repos so often. Updating takes quite a while, and chances of a Travis build being midway in an APT update...

JohanEngelen:
I'll write something about it on the LLVMdev list.  I think the update frequency change is very recent.

kinke:
Sure.

PR #1511, opened 2016-05-22
AppVeyor: Use a slightly older LLVM build

kinke:
3 days old and free of the MS (CodeView) debug info regression shown by `runnable\opover2.d` on Win64 when using `-enable-inlining`.


kinke:
WTF? dmd-testsuite-debug asserts in `runnable\closure.d:720` for Win32, but AppVeyor somehow still passes...

PR #1506, opened 2016-05-20
[AppVeyor] Update LLVM package

JohanEngelen:
My 1st attempt to create LLVM packages for AppVeyor.

PR #1513, opened 2016-05-24
[AppVeyor] Update LLVM to r271875

JohanEngelen:
See https://github.com/ldc-developers/ldc/pull/1506 for discussion about a Windows PDB issue with recent LLVM versions. Should be fixed at this LLVM rev.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -102,9 +102,9 @@ install:
   # Download & extract a pre-built LLVM (CMAKE_BUILD_TYPE=Release, LLVM_ENABLE_ASSERTIONS=ON)
   - ps: |
         If ($Env:APPVEYOR_JOB_ARCH -eq 'x64') {
-            Start-FileDownload 'https://dl.dropboxusercontent.com/s/4bzfyd0npnya4y1/LLVM-d06ea8a-x64.7z?dl=0' -FileName 'llvm-x64.7z'
+            Start-FileDownload 'https://dl.dropboxusercontent.com/s/972u4dhwif3xh6x/LLVM-r271875-x64.7z?dl=0' -FileName 'llvm-x64.7z'
         } Else {
-            Start-FileDownload 'https://dl.dropboxusercontent.com/s/m94gi730rvsmvrl/LLVM-d06ea8a-x86.7z?dl=0' -FileName 'llvm-x86.7z'
+            Start-FileDownload 'https://dl.dropboxusercontent.com/s/k0h72r6zzhc6t6b/LLVM-r271875-x86.7z?dl=0' -FileName 'llvm-x86.7z'
         }
   - md llvm-%APPVEYOR_JOB_ARCH%
   - cd llvm-%APPVEYOR_JOB_ARCH%


kinke:
Seems to work, cool - thx for updating. :+1:

kinke:
Btw any idea why the failing dmd-testsuite-debug for the x86 debug job is ignored? I don't get it. We don't do anything different from the x64 Debug job, which we know goes red when dmd-testsuite-debug fails (e.g., due to the the recent LLVM regression):
```batch
if "%APPVEYOR_JOB_CONFIG%"=="Debug" ( ctest --verbose -R dmd-testsuite-debug )
```

kinke:
Damn, this LLVM version [shows](https://ci.appveyor.com/project/kinke/ldc/build/1.0.1620/job/osd7hnhmrrftqcv7#L1515) another regression...

JohanEngelen:
:(

JohanEngelen:
Trying with a newer LLVM version.

JohanEngelen:
@kinke Do you think it is an LLVM regression (loop optimizer crash) or an LDC issue? The Travis test doesn't seem to have a problem with it and is also using a very recent LLVM.

kinke:
I'm pretty sure it's LLVM, as we don't have lots of special cases for MSVC (except for exception handling, ABI and 64-bit reals of course). I'd wait for a few more days and see if it's fixed by then; I had a few trunk snapshots which wouldn't even compile with VS2015. If it isn't fixed, we'll have to contact our reliable LLVM guy again.

JohanEngelen:
@kinke Trying with rev 271178 now.

rainers:
> @kinke Trying with rev 271178 now.

I tried it this morning and it seemed to be pretty broken. rev 271118 caused assertions, but even with these "fixed", it still crashes in WinCOFFObjectWriter::recordRelocation.

What is the latest version known to work?

kinke:
> What is the latest version known to work?

AppVeyor uses git d06ea8a (svn 270101), which is 10 days old. The current 3.9 snapshots used by Travis and CircleCI don't work either.

JohanEngelen:
Still not working :(

rainers:
I tried latest LLVM master yesterday and it seems to work fine on Windows (after a few changes to adapt LDC). CircleCI also seems 2000+ commits behind master. Should we update these?

LLVM seems to have working CodeView debug information for variable types in clang now (everything was ìnt` before), but unfortunately, that doesn't automatically makes them work with D...

JohanEngelen:
OK, I'll start builds for new Windows packages today.

Don't mind about CircleCI too much. It is using a temp solution to get LLVM 3.9 packages, I didn't find another way to get LLVM trunk binaries. So if it is behind, tough luck, AppVeyor testing is more important in my opinion, because for Windows LLVM 3.9 is the only option atm.

rainers:
Thanks. I'll create a PR with the changes necessary to adapt LDC to current LLVM.

rainers:
> I'll create a PR with the changes necessary to adapt LDC to current LLVM.

https://github.com/ldc-developers/ldc/pull/1565

Issue #1356, opened 2016-03-13
MSVC: Alignment issues for LL byval/sret params


kinke:
User code may crash when decorating certain function parameters with an LLVM `align` attribute for MSVC targets. It may lead to LLVM assertions in the inliner too, I don't remember exactly.

For 64-bit MSVC, the issue occurs for the special `sret` parameter. We currently work around it by not specifying the alignment for the pointer parameter, but still use the alignment when `alloca`ting the result.
For 32-bit MSVC, `byval` pointer parameters are also affected (we don't use `byval` at all for the Win64 ABI). The resulting user code crashes seem to indicate that LLVM isn't aligning the function parameters stack correctly.

Pinging @majnemer. :)

kinke:
Today's LLVM is able to happily compile all modules (incl. unittests) with all alignment attributes enabled, for both 32-bit and 64-bit MSVC.
All release unittests even still pass on x64.
But a few unittests crash on x86...

majnemer:
I don't think the MS ABI really works with aligned byval arguments.
For:
```
struct __declspec(align(8)) S {};
void f(S);
void g() {
  f(S());
}
```

Clang generates:
```
target datalayout = "e-m:x-p:32:32-i64:64-f80:32-n8:16:32-a:0:32-S32"
target triple = "i686-pc-windows-msvc18.0.0"

%struct.S = type { [8 x i8] }

define void @"\01?g@@YAXXZ"() #0 {
entry:
  %agg.tmp = alloca %struct.S, align 8
  call void @"\01?f@@YAXUS@@@Z"(%struct.S* byval align 4 %agg.tmp)
  ret void
}

declare void @"\01?f@@YAXUS@@@Z"(%struct.S* byval align 4)
```

majnemer:
MSVC won't compile my example.  Instead, it will issue error C2719: 

> 'unnamed-parameter': formal parameter with requested alignment of 8 won't be aligned

https://msdn.microsoft.com/en-us/library/373ak2y1.aspx

kinke:
I was afraid it was that kind of limitation. We could work around that by copying the original argument at the caller's site and passing a pointer to that aligned copy (that's what we do when passing value types > 64bit by value on Win64), at least for `extern(D)`, but then we'd have our own home-brewed Win32 ABI, incompatible with DMD...

majnemer:
[...]

PR #1515, opened 2016-05-24
Lit tests on Windows x86 now PASS

JohanEngelen:
Fixes the failing symbol hashing test.
Marks align.d test as an expected failure on Windows x86, see #1356


kinke:
Nice, thx!

Issue #1519, opened 2016-05-25
mangleof extern(C++) function contains leading "\01"


JohanEngelen:
```D
import std.stdio;
bool foo(bool a, bool b) { return a ? a : b; }
extern(C++) bool fooCpp(bool a, bool b) { return a; }

void getData(alias F)() {}

void main() {
    writeln(foo.mangleof);
    writeln(fooCpp.mangleof);
}
```

At least on Windows, both `foo` and `fooCpp` will have a mangled name with a leading "\01". However, in the extern(D) case this "\01" is not part of the string returned by `mangleof`, but for the extern(C++) function it is part of `mangleof`.
A related issue is that if `fooCpp` is passed as a template alias parameter, `getData!fooCpp()`, that the instatiated template function will have "\01" in the middle of its name, which causes trouble with some LLVM functions.

klickverbot:
The \1 will be stripped off by LLVM and only disables it's extra name decoration steps. It should never become part of mangleof - if it does, that's the issue we need to fix.

JohanEngelen:
Fixed by #1520

PR #1520, opened 2016-05-25
Windows: Move the prepending of 0x1 to C++ mangled functions from frontend to the backend.

JohanEngelen:
This removes the 0x1 byte from `.mangleof` accessible from user code.
Resolves issue #1519

### DIFF
--- a/gen/abi-x86.cpp
+++ b/gen/abi-x86.cpp
@@ -54,21 +54,47 @@ struct X86TargetABI : TargetABI {
     llvm_unreachable("Unhandled D linkage type.");
   }
 
-  std::string mangleForLLVM(llvm::StringRef name, LINK l) override {
+  std::string mangleFunctionForLLVM(std::string name, LINK l) override {
     switch (l) {
     case LINKc:
-    case LINKcpp:
     case LINKobjc:
     case LINKpascal:
     case LINKwindows:
+      return name;
+    case LINKcpp:
+      if (global.params.targetTriple->isOSWindows()) {
+        // Prepend a 0x1 byte to prevent LLVM from prepending an underscore.
+        return name.insert(0, "\1");
+      }
       return name;
     case LINKd:
     case LINKdefault:
       if (global.params.targetTriple->isOSWindows()) {
         // Prepend a 0x1 byte to keep LLVM from adding the usual
-        // "@<paramsize>" stdcall suffix.
-        return ("\1_" + name).str();
+        // "@<paramsize>" stdcall suffix. Also prepend the underscore that
+        // would otherwise be added by LLVM.
+        return name.insert(0, "\1_");
       }
+      return name;
+    }
+    llvm_unreachable("Unhandled D linkage type.");
+  }
+
+  std::string mangleVariableForLLVM(std::string name, LINK l) override {
+    switch (l) {
+    case LINKc:
+    case LINKobjc:
+    case LINKpascal:
+    case LINKwindows:
+      return name;
+    case LINKcpp:
+      if (global.params.targetTriple->isOSWindows()) {
+        // Prepend a 0x1 byte to prevent LLVM from prepending an underscore.
+        return name.insert(0, "\1");
+      }
+      return name;
+    case LINKd:
+    case LINKdefault:
       return name;
     }
     llvm_unreachable("Unhandled D linkage type.");
--- a/gen/mangling.cpp
+++ b/gen/mangling.cpp
@@ -108,7 +108,17 @@ std::string getMangledName(FuncDeclaration *fdecl, LINK link) {
     mangledName = "_D" + hashedName + "Z";
   }
 
-  return gABI->mangleForLLVM(mangledName, link);
+  // TODO: Cache the result?
+
+  return gABI->mangleFunctionForLLVM(std::move(mangledName), link);
+}
+
+std::string getMangledName(VarDeclaration *vd) {
+  // TODO: is hashing of variable names necessary?
+
+  // TODO: Cache the result?
+
+  return gABI->mangleVariableForLLVM(mangle(vd), vd->linkage);
 }
 
 std::string getMangledInitSymbolName(AggregateDeclaration *aggrdecl) {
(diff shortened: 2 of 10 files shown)


JohanEngelen:
Updated with test.

JohanEngelen:
(note that the prepending of "\1" was only done in the MSVC mangler

gen/abi-win64.cpp
```diff
@@ -54,6 +54,22 @@ struct Win64TargetABI : TargetABI {
   Win64TargetABI()
       : isMSVC(global.params.targetTriple->isWindowsMSVCEnvironment()) {}
 
+  std::string mangleForLLVM(llvm::StringRef name, LINK l) override {
+    switch (l) {
+    case LINKcpp:
+      // Prepend a 0x1 byte to prevent LLVM modifying the name.
+      return ("\1" + name).str();
```
kinke: I think this is not needed for Win64, as we use the LLVM C calling convention all the way.

JohanEngelen: Aha, so further simplification is possible, thanks.

gen/abi-x86.cpp
```diff
@@ -54,20 +54,25 @@ struct X86TargetABI : TargetABI {
...
     switch (l) {
     case LINKc:
-    case LINKcpp:
     case LINKobjc:
     case LINKpascal:
     case LINKwindows:
       return name;
+    case LINKcpp:
+      if (global.params.targetTriple->isOSWindows()) {
+        // Prepend a 0x1 byte to keep LLVM from modifying the name.
```
klickverbot: I guess this is required for X86_ThisCall?

gen/mangling.cpp
```diff
@@ -108,7 +108,17 @@ std::string getMangledName(FuncDeclaration *fdecl, LINK link) {
...
+
+  return gABI->mangleForLLVM(std::move(mangledName), link);
+}
+
+std::string getMangledName(VarDeclaration *vd) {
+  // TODO: is hashing of variable names necessary?
+
+  // TODO: Cache the result?
+
+  return gABI->mangleForLLVM(mangle(vd), vd->linkage);
```
klickverbot: I realise that the FuncDecl version already had the same interface before, but it seems a bit weird/error-prone that some of the functions in this module already apply `mangleForLLVM` and some of them don't.

klickverbot: Also, off the top of my head I'm not sure whether using `linkage` for variables makes sense, since e.g. the `\1` override to get rid of the suffix for x86_stdcall should be specific to functions.

gen/abi-x86.cpp
JohanEngelen: I don't know why the 0x1 is there, it was undocumented before.

JohanEngelen: ~~undocumented~~  not further documented

gen/mangling.cpp
JohanEngelen: This is a little strange to me too.
I added it in an attempt to fix a dmd-testsuite error with accessing these variables:
```D
extern(C++) struct Test20
{
    __gshared:
    private extern int test20;
    protected extern int test21;
    public extern int test22;
};
```
Will need to look closer at why it failed, on a Windows machine.

gen/abi-x86.cpp
klickverbot: I realise that the old diff wasn't documented either, but hoped you had discovered why it is needed in testing. ;)

gen/mangling.cpp
JohanEngelen: I just found out that the `\1` is needed for C++-variables because otherwise LLVM will add an underscore to global variable names: `_?test20@Test20@@0HA` without `\1`, but C++ symbols should not have an underscore prepended (also not on Win x86).

JohanEngelen: But now, the added call to `mangleForLLVM` for variables also adds "\1_" to all extern(D) variables. It works, but results in uglier IR...

JohanEngelen: @klickverbot In the new commit, I only call `mangleForLLVM ` for `extern(C++)` variables, to see if it now passes all tests. It's an ugly hack though. But I also don't want to add a `mangleVariableForLLVM` function... Perhaps add a `bool isVariable = false` parameter to `mangleForLLVM`?  Or we can just live with the uglier IR, and adjust IR test cases for it. :(
What do you think?

gen/mangling.cpp
```diff
@@ -108,7 +108,20 @@ std::string getMangledName(FuncDeclaration *fdecl, LINK link) {
...
+  return gABI->mangleForLLVM(std::move(mangledName), link);
+}
+
+std::string getMangledName(VarDeclaration *vd) {
+  // TODO: is hashing of variable names necessary?
+
+  // TODO: Cache the result?
+
+  if (vd->linkage == LINKcpp)
+    return gABI->mangleForLLVM(mangle(vd), vd->linkage);
```
klickverbot: This seems wrong, since it depends on knowledge of the internals of `mangleForLLVM` for all the target platforms. Unless we are adding the extra underscore ourselves somewhere where we can avoid it (I only ever looked at extern(C) and extern(D) on Windows, not C++), we might need to split up `mangleForLLVM` into a version for functions and one for variables.

JohanEngelen: @klickverbot See my notes above for the outdated commit you commented on.

klickverbot: Ah, sorry, missed that one. I think we should go the proper way of adding an LLVM-layer ABI-dependent mangling override for variables, now that I also remember the fact that that the MSVC C++ ABI uses names starting with a question mark instead of an underscore.

I'd personally prefer a second function to a boolean parameter, since they are conceptually separate and the variable case will be a no-op more often than not. It could just do nothing by default. But feel free to go with whatever you like better.

JohanEngelen:
@klickverbot Done.
I also added the missing `mangleVariableForLLVM` to the other functions in mangle.cpp.

JohanEngelen:
Thanks.
This is the last remaining stumbling block for PGO on Windows (besides updating LLVM).

PR #1521, opened 2016-05-25
Remove gratuitous extra leading underscore for x86 abi mangling of D symbols

JohanEngelen:
If this underscore is not needed, I'd rather remove it. Because of this extra underscore, some work with mangled names requires extra special case for Win32. (I ran into this during PGO work)

### DIFF
--- a/gen/abi-x86.cpp
+++ b/gen/abi-x86.cpp
@@ -67,7 +67,7 @@ struct X86TargetABI : TargetABI {
       if (global.params.targetTriple->isOSWindows()) {
         // Prepend a 0x1 byte to keep LLVM from adding the usual
         // "@<paramsize>" stdcall suffix.
-        return ("\1_" + name).str();
+        return ("\1" + name).str();
       }
       return name;
     }


JohanEngelen:
> IIRC this will break naked inline assembly on Win32/MinGW

Do you remember how?
(seems strange to me that a mangling change would do that. I was expecting breaking linking with some external party.)

rainers:
The extra '_' is also prepended for C symbols for x86 on windows, but not for x64. As dmd does this for D symbols, too, I'm not sure it is a good idea to deviate from it, as some tools might break.

If you want to fix the build failure, you'll probably have to change some more places, e.g.:

https://github.com/ldc-developers/ldc/blob/master/gen/asm-x86.h#L2318
https://github.com/ldc-developers/ldc/blob/master/gen/naked.cpp#L181

That will have to distinguish C from D symbols somehow, but these place are probably already wrong for extern(C++) symbols.

JohanEngelen:
> The extra '_' is also prepended for C symbols for x86 on windows, but not for x64. As dmd does this for D symbols, too, I'm not sure it is a good idea to deviate from it, as some tools might break.

`ddemangle` already doesn't work with LDC on OS X, because LDC prepends a second underscore. `c++filt` has the options `--strip-underscore` and `--no-strip-underscore` for that.

Thanks for the pointers to other code fragments (not so good that mangling happens in many different places I feel).

> That will have to distinguish C from D symbols somehow, but these place are probably already wrong for extern(C++) symbols.

What do you mean?  (D symbols are easily recognized because they start with "_D")

rainers:
> What do you mean? (D symbols are easily recognized because they start with "_D")

Unconditionally adding or dropping the underscore might cause link failures with object files compiled from C/C++. These locations don't check the calling convention.

BTW: Probably, pragma(mangle) needs to be checked, too.

JohanEngelen:
> ddemangle already doesn't work with LDC on OS X, because LDC prepends a second underscore.

> That's not quite what is happening there.

Regardless of why there is a second underscore, ddemangle does not work because of it.

JohanEngelen:
I don't care much about the underscore, but it does mean special-casing code for Win32 stuff.

JohanEngelen:
@klickverbot Is your explanation also valid for Win32?  (the code in this PR is only for Win32, not for other platforms)

JohanEngelen:
(It is very rude to close PR that someone is actively looking at. Even more so when you have to edit your response)

smolt:
Just a note: if any platform C compiler prefixes symbols with an underscore, the D symbols should get two underscores otherwise D symbols fall in the unreserved C symbol space.  DMD gets this wrong on OS X.  GDC and LDC get this right.  Also notice that C++ on OS X generates "_ _ Z".

PR #1522, opened 2016-05-26
[AppVeyor] Use LDC for Debug x86 build.

JohanEngelen:
This is to check whether the appveyor debug x86 failure in closure.d is fixed by using LDC instead of DMD.


JohanEngelen:
Also fails in closure.d with LDC.

PR #1523, opened 2016-05-26
AppVeyor: Publish a multilib CI build

kinke:
`get32BitArchVariant()` for x86_64 seems to return i386 which is inconsistent with i686 (Pentium Pro+) for a native 32-bit build.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -155,6 +155,41 @@ after_build:
             cd ..
             Push-AppveyorArtifact $artifactFilename
         }
+  # Debug x86 job: Create a 32/64-bit multilib artifact if the commit is on the master branch.
+  # It consists of the x64 artifact (published by a successful Debug x64 job) downloaded from
+  # GitHub, augmented by the 32-bit libs and an extra section in the ldc2.conf file for `-m32`
+  # (using the lib32 directory).
+  - ps: |
+        If (($Env:APPVEYOR_REPO_BRANCH -eq 'master') -And !(Test-Path Env:APPVEYOR_PULL_REQUEST_NUMBER) -And ($Env:APPVEYOR_JOB_CONFIG -eq 'Debug') -And ($Env:APPVEYOR_JOB_ARCH -eq 'x86')) {
+            echo 'Preparing 32/64-bit multilib artifact...'
+            cd c:\projects
+            $artifact64Downloaded = $True
+            Try {
+                Start-FileDownload "https://github.com/ldc-developers/ldc/releases/download/LDC-Win64-master/LDC-master-$Env:APPVEYOR_BUILD_NUMBER-x64.7z" -FileName 'LDC-master-x64.7z'
+            }
+            Catch {
+                echo 'Failed to download the 64-bit artifact from GitHub (the Debug x64 job probably failed).'
+                echo 'Skipping the 32/64-bit multilib artifact.'
+                $artifact64Downloaded = $False
+            }
+            If ($artifact64Downloaded) {
+                md ldc-multilib
+                cd ldc-multilib
+                7z x ..\LDC-master-x64.7z > $null
+                ren lib lib64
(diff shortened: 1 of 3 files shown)


kinke:
Some PowerShell hacking for the AppVeyor multlilib build:
```PowerShell
If ($Env:APPVEYOR_JOB_CONFIG -eq 'Debug' -And $Env:APPVEYOR_JOB_ARCH -eq 'x86') {
    echo 'Preparing 32/64-bit multilib artifact...'
    cd c:\projects
    Start-FileDownload "https://github.com/ldc-developers/ldc/releases/download/LDC-Win64-master/LDC-master-$Env:APPVEYOR_BUILD_NUMBER-x64.7z" -FileName 'LDC-master-x64.7z'
    md ldc-multilib
    cd ldc-multilib
    7z x ..\LDC-master-x64.7z > $null
    ren lib lib64
    copy ..\ldc-x86\lib > $null
    ren lib lib32
    (gc etc\ldc2.conf).replace('%%ldcbinarypath%%/../lib', '%%ldcbinarypath%%/../lib64') | sc etc\ldc2.conf
    $conf32 = gc ..\ldc-x86\etc\ldc2.conf -Raw
    $conf32 = "`r`ni686-pc-windows-msvc:" + $conf32.Substring($conf32.IndexOf("`r`ndefault:") + 10)
    $conf32 = $conf32.Replace('%%ldcbinarypath%%/../lib', '%%ldcbinarypath%%/../lib32')
    ac etc\ldc2.conf $conf32
    $artifactFilename = "LDC-master-$Env:APPVEYOR_BUILD_NUMBER-multilib.7z"
    7z a "..\$artifactFilename" * > $null
    cd ..
    Push-AppveyorArtifact $artifactFilename
}
```

What needs to be sorted out is how to get the x64 artifact. The above would/should ;) only work if the commit is on the master branch and the Debug x64 job was successful (so that its artifact has already been uploaded to GitHub).

kinke:
The 2nd commit can only be tested by merging this PR. ;)

JohanEngelen:
@kinke Perhaps you can discuss with @redstar whether to add something like this to https://github.com/ldc-developers/ldc-scripts to help with releases?

JohanEngelen:
> The above would/should ;) only work if the commit is on the master branch 

Does that mean that a PR would always fail testing?

kinke:
> > The above would/should ;) only work if the commit is on the master branch 
> Does that mean that a PR would always fail testing?

In the preliminary version I posted here as comment, yes; but not in the 2nd commit, otherwise this PR would already fail because of it.

JohanEngelen:
Ah, yes of course. 

LGTM.

JohanEngelen:
( it does seem to make the AppVeyor build quite a bit slower (~15 mins?) )

kinke:
Nope, the build is untouched, the multilib package is only generated when merging this PR. It's the dynamic AppVeyor VMs which show this inconsistent performance (sometimes, a VM doesn't even get a full physical core on its own...).

JohanEngelen:
Great, so let's merge this.

kinke:
Alright, I'm quite eager to see if it's working too. :)

kinke:
It needed a lil' fix (https://github.com/ldc-developers/ldc/commit/bd3e1b210f4a3c6cfff6d5000c653d57eb2ef149) but should work now.

kinke:
Okay so here's the first one: https://ci.appveyor.com/api/buildjobs/1ytgmu2d0in2qww1/artifacts/LDC-master-1641-multilib.7z
Works fine for me inside a naked cmd.exe instance (no VS environment variables set). It won't work inside a 'VS command prompt' as LDC doesn't attempt to find the VS installation and paths for the target architecture if the environment variables are already set.

rainers:
Cool, works with Visual D, too! (I've tried my soon to be released local version only, though.)

Issue #1518, opened 2016-05-25
ldc.simd: storeUnaligned is missing


9il:

MartinNowak:
You can use `__builtin_ia32_storeups`, `__builtin_ia32_storeupd`, and `__builtin_ia32_storedqu`.

9il:
> You can use __builtin_ia32_storeups, __builtin_ia32_storeupd, and __builtin_ia32_storedqu.

Yes, but they are no crossplatform

JohanEngelen:
Closing because of #1526.

PR #1526, opened 2016-05-27
SIMD: Fix loadUnaligned for newer LLVM types and add storeUnaligned.

JohanEngelen:
Resolves issue #1518. Combines druntime patches from @MartinNovak and @9il.

Druntime changes: https://github.com/ldc-developers/druntime/commit/9189ca33b473d493aea71063ec3003ed9641b283


JohanEngelen:
(I am probably alone in this, but I find it annoying to have large unittests inside the code files, because very quickly the actual function will drown in it, discouraging adding more comprehensive and thorough unittests. Having unittests function as (user) documentation is pretty bad too. But, alas, I think I am the only one with this opinion.)

In this case, I was lucky there is a reason to do IR checking! ;)

Issue #1524, opened 2016-05-26
Incorrect closure


kinke:
The AppVeyor Debug x86 job currently fails in `runnable/closure.d:720`. I've reduced it to this:
```D
import core.stdc.stdio;

void main()
{
    struct S { int a, b, c; }

    static int delegate() makeSum1(S s)
    {
        with (s) return {
            //printf("s = { %d, %d, %d }\n", s.a, s.b, s.c);
            return a + b + c;
        };
    }

    S s = {15, 30, 45};
    auto sum = makeSum1(s);
    int r = sum();
    assert(r == 90);
}
```
It works when accessing the 3 fields in the returned lambda explicitly via `s.a + s.b + s.c` and not via the `with` symbol.

Comparing the IR for `makeSum1()` between x86 and x64 reveals something interesting. x86:
```LL
%nest.makeSum1 = type { %closure.main.S*, %closure.main.S* }
  ...
  %.frame = call noalias i8* @_d_allocmemory(i32 8) ; [#uses = 1]
  %.frame1 = bitcast i8* %.frame to %nest.makeSum1* ; [#uses = 3]
  %__withSym = getelementptr inbounds %nest.makeSum1, %nest.makeSum1* %.frame1, i32 0, i32 0 ; [#uses = 1, type = %closure.main.S**]
  %s = getelementptr inbounds %nest.makeSum1, %nest.makeSum1* %.frame1, i32 0, i32 1 ; [#uses = 1, type = %closure.main.S**]
  %.gc_mem = call noalias i8* @_d_allocmemory(i32 12) ; [#uses = 1]
  %.gc_mem2 = bitcast i8* %.gc_mem to %closure.main.S* ; [#uses = 2]
  %1 = bitcast %closure.main.S* %.gc_mem2 to i8*  ; [#uses = 1]
  %2 = bitcast %closure.main.S* %s_arg to i8*     ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i32(i8* %1, i8* %2, i32 12, i32 1, i1 false)
  store %closure.main.S* %.gc_mem2, %closure.main.S** %s, align 4
  store %closure.main.S* %s_arg, %closure.main.S** %__withSym
  ...
```
x64:
```LL
%nest.makeSum1 = type { %closure.main.S*, %closure.main.S }
  ...
  %1 = alloca %closure.main.S, align 4            ; [#uses = 2, size/byte = 12]
[...]

PR #1527, opened 2016-05-28
Try to boil down special cases in nested.cpp

kinke:


gen/nested.cpp
```diff
@@ -402,31 +384,12 @@ static void DtoCreateNestedContextType(FuncDeclaration *fd) {
...
-        }
-      } else {
-        t = irparam->value->getType();
-      }
-    } else if (isSpecialRefVar(vd)) {
+    if (vd->isRef() || vd->isOut())
       t = DtoType(vd->type->pointerTo());
-    } else {
+    else if (vd->isParameter() && (vd->storage_class & STClazy))
+      t = getIrParameter(vd)->value->getType()->getContainedType(0);
```
kinke: Is there no better way to get the delegate type? `vd->type` seems to be the return value type, so that's one reason we need the IrParameter::value field.

gen/nested.cpp
```diff
@@ -517,9 +480,12 @@ void DtoCreateNestedContext(FuncDeclaration *fd) {
...
         LOG_SCOPE
         IrParameter *parm = getIrParameter(vd);
+        assert(parm->value);
+        assert(parm->value->getType()->isPointerTy());
 
-        if (parm->arg && parm->arg->byref) {
-          storeVariable(vd, gep);
+        if (vd->isRef() || vd->isOut()) {
+          Logger::println("Captured by reference, copying pointer to nested frame");
+          DtoAlignedStore(parm->value, gep);
```
kinke: No need to redirect `parm->value` in the by-reference case, contrary to the by-value case below. That seems to be the key to let the `with` symbol point to the correct variable (the copy inside the frame, not the original parameter). I'm just completely unsure whether this works reliably, because it depends on when this code gets executed in the nested hierarchy. @klickverbot: can you shed some light on this?

klickverbot: Unfortunately not. Your guess is as good as mine.

gen/nested.cpp
```diff
@@ -19,23 +19,6 @@
...
 
-static void storeVariable(VarDeclaration *vd, LLValue *dst) {
-  LLValue *value = getIrLocal(vd)->value;
-  int ty = vd->type->ty;
-  FuncDeclaration *fd = getParentFunc(vd, true);
-  assert(fd && "No parent function for nested variable?");
-  if (fd->needsClosure() && !vd->isRef() && (ty == Tstruct || ty == Tsarray) &&
-      isaPointer(value->getType())) {
-    // Copy structs and static arrays
-    LLValue *mem = DtoGcMalloc(vd->loc, DtoType(vd->type), ".gc_mem");
```
kinke: Afaik, the frame is allocated by the GC, so we shouldn't have to additionally allocate structs and static arrays separately via the GC too - just store them directly inside the frame.

gen/functions.cpp
```diff
@@ -939,11 +931,8 @@ void DtoDefineFunction(FuncDeclaration *fd) {
         ++llArgIdx;
       }
 
-      if (global.params.symdebug &&
-          !(isaArgument(irparam->value) &&
-            isaArgument(irparam->value)->hasByValAttr())) {
```
kinke: Give byval params debug info too.

kinke:
Rather surprisingly, this passes the full debug test suite on my Win64 box. It should also fix #1524.

kinke:
Green - I didn't dare to hope it would work out like this. ;)

gen/functions.cpp
JohanEngelen: Can I poke you to create a few tests that check for debug info in IR?  (perhaps in `tests/debuginfo/`)  ;-)

JohanEngelen:
Great that #1524 is fixed with this. It looks like a nice simplification of code, but I have no expertise in this part of the codebase. Hit the merge button if you are done working on this! :-)

PR #1528, opened 2016-05-28
Optimize ABIRewrite system for lvalues

kinke:
Allow ABIRewrites to return the D parameter's LL lvalue directly.
Most rewrites store to memory anyway, so let the D parameter point directly to that memory instead of a dedicated alloca bitcopy.

### DIFF
--- a/gen/abi.h
+++ b/gen/abi.h
@@ -39,16 +39,16 @@ class FunctionType;
 struct ABIRewrite {
   virtual ~ABIRewrite() = default;
 
-  /// get a rewritten value back to its original form
-  virtual llvm::Value *get(Type *dty, llvm::Value *v) = 0;
-
-  /// get a rewritten value back to its original form and store result in
-  /// provided lvalue
-  /// this one is optional and defaults to calling the one above
-  virtual void getL(Type *dty, llvm::Value *v, llvm::Value *lval);
-
-  /// put out rewritten value
+  /// Transforms the D argument to a suitable LL argument.
   virtual llvm::Value *put(DValue *v) = 0;
+
+  /// Transforms the LL parameter back and returns the address for the D
+  /// parameter.
+  virtual llvm::Value *getLVal(Type *dty, llvm::Value *v) = 0;
+
+  /// Transforms the LL parameter back and returns the value for the D
+  /// parameter.
+  virtual llvm::Value *getRVal(Type *dty, llvm::Value *v);
 
   /// should return the transformed type for this rewrite
   virtual llvm::Type *type(Type *dty, llvm::Type *t) = 0;
--- a/gen/abi.h
+++ b/gen/abi.h
@@ -59,11 +59,6 @@ protected:
   /// Returns the address of a D value, storing it to memory first if need be.
   static llvm::Value *getAddressOf(DValue *v);
 
-  /// Stores a LL value to a specified memory address. The element type of the
-  /// provided pointer doesn't need to match the value type (=> suited for
-  /// bit-casting).
-  static void storeToMemory(llvm::Value *rval, llvm::Value *address);
-
   /// Loads a LL value of a specified type from memory. The element type of the
   /// provided pointer doesn't need to match the value type (=> suited for
   /// bit-casting).
--- a/gen/abi.h
+++ b/gen/abi.h
@@ -35,7 +35,9 @@ class Value;
 class FunctionType;
 }
 
-// return rewrite rule
+/// Transforms function arguments and return values.
+/// This is needed to implement certain ABI aspects which LLVM doesn't get
+/// right by default.
 struct ABIRewrite {
   virtual ~ABIRewrite() = default;
 
@@ -48,10 +50,12 @@ struct ABIRewrite {
 
   /// Transforms the LL parameter back and returns the value for the D
   /// parameter.
+  /// Defaults to loading the lvalue returned by getLVal().
   virtual llvm::Value *getRVal(Type *dty, llvm::Value *v);
 
-  /// should return the transformed type for this rewrite
-  virtual llvm::Type *type(Type *dty, llvm::Type *t) = 0;
+  /// Returns the resulting LL type when transforming an argument of the
+  /// specified D type.
+  virtual llvm::Type *type(Type *t) = 0;
 
 protected:
   /***** Static Helpers *****/
@@ -60,7 +64,7 @@ protected:
   static llvm::Value *getAddressOf(DValue *v);
 
   /// Loads a LL value of a specified type from memory. The element type of the
-  /// provided pointer doesn't need to match the value type (=> suited for
+  /// provided pointer doesn't need to match the value type (=> suitable for
   /// bit-casting).
   static llvm::Value *loadFromMemory(llvm::Value *address, llvm::Type *asType,
                                      const char *name = ".bitcast_result");
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -227,14 +227,14 @@ LLValue *DtoAllocaDump(LLValue *val, Type *asType, const char *name) {
 
 LLValue *DtoAllocaDump(LLValue *val, LLType *asType, int alignment,
                        const char *name) {
-  LLType *valType = i1ToI8(voidToI8(val->getType()));
-  asType = i1ToI8(voidToI8(asType));
+  LLType *memType = i1ToI8(voidToI8(val->getType()));
+  LLType *asMemType = i1ToI8(voidToI8(asType));
   LLType *allocaType =
-      (getTypeStoreSize(valType) <= getTypeAllocSize(asType) ? asType
-                                                             : valType);
+      (getTypeStoreSize(memType) <= getTypeAllocSize(asMemType) ? asMemType
+                                                                : memType);
   LLValue *mem = DtoRawAlloca(allocaType, alignment, name);
-  DtoStoreZextI8(val, DtoBitCast(mem, valType->getPointerTo()));
-  return DtoBitCast(mem, asType->getPointerTo());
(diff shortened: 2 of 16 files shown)


gen/abi.h
```diff
@@ -39,16 +39,17 @@ class FunctionType;
 struct ABIRewrite {
   virtual ~ABIRewrite() = default;
 
-  /// get a rewritten value back to its original form
-  virtual llvm::Value *get(Type *dty, llvm::Value *v) = 0;
+  /// Transforms the D argument to a suited LL argument.
```
klickverbot: "suitable"? I'm not a native speaker either, but "suited" as an adjective conjures weird images of LLVM values wearing suites in my head.

kinke: Heh yes of course ;)

gen/tocall.cpp
```diff
@@ -900,13 +900,11 @@ DValue *DtoCallFunction(Loc &loc, Type *resulttype, DValue *fnval,
     // do ABI specific return value fixups
     if (storeReturnValueOnStack) {
       Logger::println("Storing return value to stack slot");
-      LLValue *mem = DtoAlloca(returntype);
-      irFty.getRet(returntype, retllval, mem);
-      retllval = mem;
+      retllval = irFty.getRetLVal(returntype, retllval);
```
klickverbot: With your new interface, it is still necessary to have both the `getRetLVal` branch and the "manual" `storeReturnValueOnStack` branch below?

kinke:
Passes the full debug test suite on my Win64 box.

gen/tocall.cpp
kinke: That's still a mess to sort out. As is all the repainting going on for return values and parameters...

gen/llvmhelpers.cpp
```diff
@@ -1602,18 +1602,10 @@ DValue *DtoSymbolAddress(Loc &loc, Type *type, Declaration *decl) {
...
       }
-      if (vd->isRef() || vd->isOut() || DtoIsInMemoryOnly(vd->type) ||
-          llvm::isa<llvm::AllocaInst>(getIrValue(vd))) {
-        assert(!isSpecialRefVar(vd) && "Code not expected to handle special "
-                                       "ref vars, although it can easily be "
-                                       "made to.");
-        return new DVarValue(type, getIrValue(vd));
-      }
-      if (llvm::isa<llvm::Argument>(getIrValue(vd))) {
-        return new DImValue(type, getIrValue(vd));
```
kinke: This took me quite some time to get my attention. A delegate rewritten by ExplicitByvalRewrite (an lvalue, a LL pointer) would be bound to the parameter as `DImValue`, and it wouldn't get dereferenced properly later on. That was because the pointer is now the actual argument itself instead of a private alloca (so `llvm::isa<llvm::Argument>()` returns true).

klickverbot: Yeah, as I recently discovered too (cf. all the return value instruction domination issues), there are a lot of places like this that confuse IR-level semantics with whatever confused concepts of lvalues/rvalues we have on the DValue layer.

gen/llvmhelpers.cpp
```diff
@@ -1602,18 +1602,10 @@ DValue *DtoSymbolAddress(Loc &loc, Type *type, Declaration *decl) {
...
-      }
-      llvm_unreachable("Unexpected parameter value.");
-
+      assert(!isSpecialRefVar(vd) && "Code not expected to handle special "
+                                     "ref vars, although it can easily be "
+                                     "made to.");
+      return new DVarValue(type, getIrValue(vd));
```
klickverbot: The special case for `STClazy` just above this can be removed now, right?

gen/abi.h
```diff
@@ -39,16 +39,16 @@ class FunctionType;
...
-  virtual llvm::Value *get(Type *dty, llvm::Value *v) = 0;
+  /// Transforms the D argument to a suitable LL argument.
+  virtual llvm::Value *put(DValue *v) = 0;
 
-  /// get a rewritten value back to its original form and store result in
-  /// provided lvalue
-  /// this one is optional and defaults to calling the one above
-  virtual void getL(Type *dty, llvm::Value *v, llvm::Value *lval);
+  /// Transforms the LL parameter back and returns the address for the D
+  /// parameter.
```
klickverbot: Could you add a short note here that implementation will dump the parameter into a new alloca to make it an lvalue if necessary, and that this makes sense because value parameters are l-values in D? (The last half-sentence probably needs better wording – what I'm trying to get at is that this is where the temporary is created that makes `void foo(int a) { a = 42; }` work).

kinke:
It would be nice if someone could verify that the ARM ABIs still work (@smolt, @redstar, @joakim-noah).

smolt:
I am currently useless.  All my stuff is in boxes as I am moving.

joakim-noah:
Applied this PR to master and cross-compiled for Android/ARM, all the same standard library tests pass.

gen/abi.h
kinke: I've added the comment in `gen/functions.cpp:666`, where all explicit params are defined (incl. non-rewritten ones).

gen/llvmhelpers.cpp
kinke: Done, but I've kept the log msg and assertion.

gen/structs.cpp
```diff
@@ -168,7 +168,7 @@ void DtoPaddedStruct(Type *dty, LLValue *v, LLValue *lval) {
       // Nested structs are the only members that can contain padding
       DtoPaddedStruct(fields[i]->type, fieldval, fieldptr);
     } else {
-      DtoStore(fieldval, fieldptr);
+      DtoStoreZextI8(fieldval, fieldptr);
```
kinke: This bug (unable to store `i1` bits) appears to have been the reason why the return value was never rewritten by the ABI (intrinsics ABI in this case) for intrinsics.
The return value of intrinsics returning a LL struct (e.g., `llvm.sadd.with.overflow`) therefore didn't undergo the usual ABI transformation process; in this case, the LL struct was simply dumped to memory instead of going through `DtoPaddedStruct()`.

kinke:
Thanks @joakim-noah for the ARM tests.

kinke:
This is the effect on the unoptimized IR for Win64 (`ExplicitByvalRewrite`):
```D
struct Struct { long a, b; }

Struct foo(Struct s) { return s; }

void main()
{
    auto s = Struct(1, 2);
    foo(s);
}
```
```LL
// old:
define void @_D5byval3fooFS5byval6StructZS5byval6Struct(%byval.Struct* noalias sret align 8 %.sret_arg, %byval.Struct* noalias nocapture align 16 %s_arg) #0 comdat {
  %s = alloca %byval.Struct, align 8              ; [#uses = 2, size/byte = 16]
  %1 = bitcast %byval.Struct* %s to i8*           ; [#uses = 1]
  %2 = bitcast %byval.Struct* %s_arg to i8*       ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %1, i8* %2, i64 16, i32 1, i1 false)
  %3 = bitcast %byval.Struct* %.sret_arg to i8*   ; [#uses = 1]
  %4 = bitcast %byval.Struct* %s to i8*           ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %3, i8* %4, i64 16, i32 1, i1 false)
  ret void
}

// new:
define void @_D5byval3fooFS5byval6StructZS5byval6Struct(%byval.Struct* noalias sret align 8 %.sret_arg, %byval.Struct* noalias nocapture align 16 %s) #0 comdat {
  %1 = bitcast %byval.Struct* %.sret_arg to i8*   ; [#uses = 1]
  %2 = bitcast %byval.Struct* %s to i8*           ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %1, i8* %2, i64 16, i32 1, i1 false)
  ret void
}

// common:
define i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #0 comdat {
  %s = alloca %byval.Struct, align 8              ; [#uses = 2, size/byte = 16]
  %.structliteral = alloca %byval.Struct, align 8 ; [#uses = 3, size/byte = 16]
  %.rettmp = alloca %byval.Struct, align 8        ; [#uses = 1, size/byte = 16]
  %.ExplicitByvalRewrite_dump = alloca %byval.Struct, align 16 ; [#uses = 2, size/byte = 16]
  // initialize .structliteral with (1,2)
  // memcpy .structliteral to .s
  // memcpy .s to .ExplicitByvalRewrite_dump
  call void @_D5byval3fooFS5byval6StructZS5byval6Struct(%byval.Struct* noalias sret align 8 %.rettmp, %byval.Struct* noalias nocapture align 16 %.ExplicitByvalRewrite_dump) #0
  ret i32 0
}
```

This useless additional copy has been bugging me for quite a while.

smolt:
For ARM, test-suite is where ABI better is tested. @joakim-noah did you run test-suite also?

joakim-noah:
No, I'll try that next.

joakim-noah:
Ran dmd testsuite natively on Android/ARM with this PR, identical results.

smolt:
One more plus for ARM.  Raspberry Pi 3 and master (with this merged PR) passes all tests, except lit-tests don't run correctly due to my setup.  I have a space in a USB HD label and that dorks up the lit command lines.

JohanEngelen:
> except lit-tests don't run correctly due to my setup. I have a space in a USB HD label and that dorks up the lit command lines.

Can you share some details? Sounds like a bug in the way I set up lit.

smolt:
Sure, I'll paste in the error when I get home today.

smolt:
@JohanEngelen lit-test issue ldc-developers/ldc#1536 (which I am sure github emailed you about too)

PR #1529, opened 2016-05-29
[CMake] Fix typo in Debug build flag setting.

JohanEngelen:
This looks like a typo.


JohanEngelen:
Since: 472b12f1b005a9075cc68c651b1f13ea44d95917

JohanEngelen:
@kinke Do you remember why you added `-link-debuglib`?

JohanEngelen:
>  Interestingly, it seems like we have no CI build to cover that, though.

Hmmm... Maybe the AppVeyor x86 build would have triggered? (Debug + DMD compiler)

kinke:
I think I added it to mimic legacy behavior, where we would link to the debug stdlibs when specifying `-g` by default, and to enable full debugging support for a debug build.
The missing 'D' was/is a typo. And yeah we don't use any debug builds for our CI tests, but there's a good reason for that... ;)

kinke:
A CircleCI job may be doable, that's the quickest one atm. But honestly, noone took notice of missing debug infos for the front-end of debug LDC builds until now, so I wonder how many of us use debug builds (I'm quite fine with RelWithDebInfo myself...). And then there's also the Release and RelMinSize (or so) configs...

JohanEngelen:
What's funny is that I _do_ often use "Debug" builds, for a much quicker compile time (for a lot of stuff I don't need a rebuilt phobos/druntime). 
Shall I put the `-link-debuglib` in a conditional checking for LDMD ?

kinke:
Oh now I see why it's a problem, I completely forgot about dmd not supporting this switch - and I actually use DMD to build the front-end. ;)
Yeah then please just remove it.

kinke:
> thought it was just a case of D compilers producing crappy debug info. ;)

Don't you lose your faith! ;)

JohanEngelen:
> and I actually use DMD to build the front-end. ;)

You _what_ ?!?!  ;-) :P
I do too for fast turn-arounds.

JohanEngelen:
I kept the flag for LDMD use.

PR #1531, opened 2016-05-30
Fix LDC ICE for https://issues.dlang.org/show_bug.cgi?id=16022.

JohanEngelen:
Consider 
```D
type == EnumerationOfStructs.Colon, type == EnumerationOfStructs.Comma;
```
This expression currently (front-end v2.071) results in a VarExp for the enum member in the LHS.

The problem appears to be that in a comma expression, an enum member appearing in the LHS is not constant folded and a VarExp remains in the AST. The AST for the LHS of that expression is an ExpStatement with AndAndExp's of EqualExp's for every struct field. 
Because of the missing constant folding, the generated code is verbose: a new struct temporary is created for every struct field comparison, because the EqualExp will still have a DotVarExp into a VarExp (the enumeration member). And we create a new enumeration member temporary for each VarExp (we don't emit enumeration variables as global variables).
With -O3 it all disappears.

Another fix would be to emit enumeration variables as globals (with internal linkage), such that we can take the address of it and VarExp works like the standard case.


gen/toir.cpp
```diff
@@ -357,6 +357,14 @@ class ToElemVisitor : public Visitor {
       genFuncLiteral(fd, nullptr);
     }
 
+  if (auto em = e->var->isEnumMember()) {
```
klickverbot: The indentation seems to be wonky here.

JohanEngelen:
I'm not sure whether this is a real front-end bug (instead of missed const fold opportunity). An EnumMember is subclassed from a VarDeclaration, and a VarDeclaration does not need to have an address (`bool VarDeclaration.canTakeAddressOf()`). When we visit the `VarExp` we assume we can take the address of it, but I think that is not always the case (as in this rare enum case).
So now I think the check that I added for `e->var->isEnumMember()` should be changed to `canTakeAddressOf`.

Issue #1471, opened 2016-05-02
ICE when accessing classinfo (breaks DustMite)


rainers:
Trying to build Visual D with LDC, I got ICEs with code dustmited to this

```
class Node
{
	final void _clone()
	{
		this.classinfo.create();
	}
}
```
Happens for 1.0.0beta1 and merge-2.071 with LLVM master. `typeid(this).create` seems to work.

Top of the stack is:
```
llvm::PointerType::get() + 0x17 bytes(s), c:\s\d\ldc\llvm\lib\ir\type.cpp, line 686
llvm::GetElementPtrInst::getGEPReturnType() + 0x4D bytes(s), c:\s\d\ldc\llvm\include\llvm\ir\instructions.h, line 990 + 0x27 byte(s)
`anonymous namespace'::DtoGEP() + 0xDC bytes(s), c:\s\d\ldc\ldc\gen\tollvm.cpp, line 289 + 0xA7 byte(s)
DtoGEPi() + 0x6E bytes(s), c:\s\d\ldc\ldc\gen\tollvm.cpp, line 319
DtoVirtualFunctionPointer() + 0x12C bytes(s), c:\s\d\ldc\ldc\gen\classes.cpp, line 422
ToElemVisitor::visit() + 0x209 bytes(s), c:\s\d\ldc\ldc\gen\toir.cpp, line 1200 + 0x16 byte(s)
ddmd.expression.DotVarExp.accept() + 0x1D bytes(s), C:\s\d\ldc\ldc\ddmd\expression.d, line 8461 + 0x1 byte(s)
toElem() + 0x74 bytes(s), c:\s\d\ldc\ldc\gen\toir.cpp, line 2918
ToElemVisitor::visit() + 0x20A bytes(s), c:\s\d\ldc\ldc\gen\toir.cpp, line 929 + 0x9 byte(s)
ddmd.expression.CallExp.accept() + 0x1D bytes(s), C:\s\d\ldc\ldc\ddmd\expression.d, line 9747 + 0x1 byte(s)
```

kinke:
Some `-vv` diff between `typeid(this)` and `this.classinfo`:

`typeid(this)`:
```
CallExp::toElem: typeid(this).create() @ object.Object
* DotVarExp::toElem: typeid(this).create @ const Object()
* * ThisExp::toElem: this @ pow.Node
* * * normal this exp
* * isMember = this is: object.TypeInfo_Class
* * DtoFunctionType(const Object())
[...]

PR #1532, opened 2016-05-30
object.classinfo: Return a TypeInfo_Class instead of a TypeInfo_Interface

kinke:
This may be enough to fix https://github.com/ldc-developers/ldc/issues/1471 (at least there's no crash anymore for @rainers' minimal testcase).

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -1143,7 +1143,7 @@ public:
         // member, so we have to add an extra layer of indirection.
         resultType = Type::typeinfointerface->type->pointerTo();
       } else {
-        resultType = Type::typeinfointerface->type;
+        resultType = Type::typeinfoclass->type;
       }
 
       V = DtoBitCast(V, DtoType(resultType->pointerTo()->pointerTo()));


JohanEngelen:
Could you also add a testcase for this?

Perhaps something like:
```D

class Node
{
    final void _clone()
    {
// CHECK-NOT: load %object.TypeInfo_Interface*
// CHECK: load %object.TypeInfo_Class*
        this.classinfo.create();
    }
}
```

kinke:
Thanks for the ready-to-paste test, but I'm not convinced this deserves a test on its own; it rather seems like a bug due to lack of focus or a typo to me, with an extremely low probability of being reintroduced. ;)
Let's see if it passes the tests and whether it also fixes Rainer's problems first.

JohanEngelen:
OK. Pretty much the first thing I think of now is: "does it have tests?" More tests, more better ;-)
 I am not sure the IR has to be checked, but a simple "does it compile" check would be nice, no? It would check for many other things besides your typo ;)

kinke:
I'm pretty sure this typo isn't mine. ;)
Yeah that's what I meant (not checking the IR, just checking whether it compiles and works), but I'd rather see such tests go upstream so that all compilers benefit (I don't know whether GDC uses dmd-testsuite too, but I guess and hope so) from improved test coverage. DMD's test coverage is pretty bad, especially wrt. basic compiler and language stuff - https://github.com/ldc-developers/ldc/issues/1327 wouldn't exist if there was a good test coverage for the evaluation order of expressions.

JohanEngelen:
Opps, sorry misread :(
Upstreaming is good.

rainers:
LGTM, but I haven't tried it yet (I replaced all occurences with typeid). I wonder why this isn't lowered to TypeidExp in the front end, though. 
A test should verify runtime behaviour, too.

kinke:
This may not be correct. `TypeidExp` returns a `TypeInfo_Interface` for interfaces, not a `TypeInfo_Class`. Don't know if that is supposed to happen for `.classinfo` too.

rainers:
According to the doc and Kenji, `typeid()` and `.classinfo` are supposed to do the same. There have been some discussions around what should be returned and that strange things happen with dmd, too: https://github.com/dlang/dmd/pull/4711, https://issues.dlang.org/show_bug.cgi?id=14612 and a number of linked reports.

kinke:
Updated so that .classinfo for interfaces returns a TypeInfo_Interface, as before and compatible to `TypeidExp` codegen.

rainers:
LGTM. It's compatible with dmd with respect to not returning the actual class for interfaces.

PR #1533, opened 2016-05-30
Introduce DtoRVal/DtoLVal convenience functions

kinke:
No functional changes intended.


kinke:
The next step would be to ditch `DValue::getL/RVal()` and simply access their `value` field directly, using `isLVal()` to check whether it's an lvalue or rvalue and handle it accordingly. But enough for tonight. ;)

gen/dvalue.h
```diff
@@ -79,18 +75,13 @@ class DValue {
   virtual DFuncValue *isFunc() { return nullptr; }
 
 protected:
-  DValue() = default;
-  DValue(const DValue &) {}
-  DValue &operator=(const DValue &other) {
-    type = other.type;
-    return *this;
-  }
+  DValue(Type *t, llvm::Value *v) : type(t), val(v) {}
```
JohanEngelen: I think the idea is that DValue cannot be default constructed and DValue.type is always !nullptr, right? (the removal of the assert with getType)  Could you add the assert here to check that `t` is never `nullptr`?

kinke:
> Why do you want to make val directly accessible, though?

I'd prefer a DValue to be an immutable LL value associated with a D type and the info whether it's a LL lvalue or rvalue (there's a difference between LL lvalues and D lvalues - structs and static arrays are always LL lvalues while possibly representing D rvalues), nothing more.
Getting LL rvalues via `getRVal()` implies a load for lvalues, so the resulting immediate depends on when the load happens (our evaluation order issue).
I'd prefer doing that from the outside via a `DtoRVal(DValue*)` overload - pass an Expression for the trivial case where one doesn't need the DValue later on (current status), and otherwise the codegen'd DValue.

gen/dvalue.h
kinke: Thanks, I missed that one.

kinke:
I made the `val` field protected for now.
The 5th commit took its time but is worth it imo. It should fix the previously fuzzy lval/rval semantics. Getting it green was a tedious and error-prone trial-and-error-debugging process, so I may not have gotten all corner cases right yet.

kinke:
That's probably it from the refactoring perspective. The evaluation order issues still need to be tracked down. ;)

JohanEngelen:
I just found out that CircleCI does not rebase a PR onto current master before starting a testbuild  :(
So for this PR it is using an outdated circle.yml file. Could you rebase @kinke ?

kinke:
Thanks @JohanEngelen for fixing Travis and CircleCI. Rebased.

JohanEngelen:
@kinke This looks merge-ready, right?

kinke:
Yep, the evaluation order fixes can be put into a separate PR.

JohanEngelen:
Excellent!

joakim-noah:
I just updated to latest HEAD on Android/ARM and this PR seems to have broken compilation of one test in the dmd testsuite, `runnable/variadic.d`.  I get the following error when trying to compile:
```
Load operand must be a pointer.
  %7 = load i8*, %ldc.internal.vararg.std.__va_list %6
Stored value type does not match pointer operand type!
  store i8* %7, %ldc.internal.vararg.std.__va_list* %ap2
 %ldc.internal.vararg.std.__va_list = type { i8* }LLVM ERROR: Broken function found, compilation aborted!
```
I can't seem to dump the IR to check it, as ldc craps out in llvm's module verifier, which is run at the lowest optimization level also.  I'll look into turning even that pass off to get the IR, in the meantime, just letting you know.

joakim-noah:
It might, I didn't see it in the listed options for ldc.  I'm rebuilding now, will try it and let you know.

joakim-noah:
It turns out [the function pass manager also verifies the IR for each function](https://github.com/ldc-developers/ldc/blob/master/gen/optimizer.cpp#L235), so `-disable-verify` is not enough.  I commented out that line _and_ used your flag to finally get the IR, which is attached to the linked issue.

PR #1539, opened 2016-06-03
Accept LLVM bitcode files (*.bc) on the commandline

JohanEngelen:
Accept LLVM bitcode files (*.bc) on the commandline and add the code to the first source file on the cmdline (the first emitted module).

(succesfully tested on Weka's codebase)

### DIFF
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -13,6 +13,7 @@
 #include "mars.h"
 #include "module.h"
 #include "scope.h"
+#include "driver/linker.h"
 #include "driver/toobj.h"
 #include "gen/logger.h"
 #include "gen/runtime.h"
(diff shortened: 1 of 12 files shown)


driver/codegenerator.cpp
```diff
@@ -99,9 +112,11 @@ void CodeGenerator::writeAndFreeLLModule(const char *filename) {
   ir_->DBuilder.Finalize();
 
   // Add the linker options metadata flag.
-  ir_->module.addModuleFlag(
-      llvm::Module::AppendUnique, "Linker Options",
-      llvm::MDNode::get(ir_->context(), ir_->LinkerMetadataArgs));
+  if (!ir_->module.getModuleFlag("Linker Options")) {
```
klickverbot: Shouldn't the linker options be merged together, or at least set from the main executable?

JohanEngelen: Definitely should be merged, indeed.

driver/linker.cpp
```diff
@@ -103,6 +106,36 @@ static std::string getOutputName(bool const sharedLib) {
...
+/// Insert an LLVM bitcode file into the module
+void insertBitcodeIntoModule(const char *bcFile, llvm::Module &M,
+                             llvm::LLVMContext &Context) {
+  Logger::println("*** Linking-in bitcode file %s ***", bcFile);
+
+  llvm::SMDiagnostic Err;
+  std::unique_ptr<llvm::Module> loadedModule =
+      getLazyIRFileModule(bcFile, Err, Context);
+  if (!loadedModule) {
+    // Err.print(argv0, errs());
```
klickverbot: Huh?

driver/main.cpp
```diff
@@ -1157,17 +1158,27 @@ int cppmain(int argc, char **argv) {
...
-          Port::stricmp(ext, global.bc_ext) == 0)
+          Port::stricmp(ext, global.obj_ext_alt) == 0)
 #endif
       {
         global.params.objfiles->push(static_cast<const char *>(files.data[i]));
         continue;
       }
 
+      // Detect LLVM bitcode files on commandline
+#if LDC_POSIX
```
klickverbot: Now that our main is definitely not source-compatible with DMD's one (C++ vs. D), we should definitely go ahead and clean things up (after bringing them in sync with the upstream changes that happened in the meantime). This can wait until another day, though.

JohanEngelen:
Merging was a pain, dealing the LLVM metadata API could definitely by improved.
Anyway, it's done now. It appears it is only relevant for Windows though. Added test for it too.

PR #1555, opened 2016-06-11
try to get travis back up

JohanEngelen:


.travis.yml
```diff
@@ -1,13 +1,42 @@
 language: d
-sudo: false
+sudo: required
```
klickverbot: Do we really need sudo for using the LLVM release binaries? Last time I checked, build latencies were supposed to be much higher than on the container-based infrastructure, but that might have changed in the meantime, of course.

JohanEngelen: We don't need sudo. But we need something more recent than Ubuntu 12.

JohanEngelen: See : https://docs.travis-ci.com/user/ci-environment/

klickverbot: Ah, sorry – I had vaguely followed the announcements regarding 14.04, but didn't remember that you need to explicitly opt out of the container infrastructure for it. Let's hope for the best, then.

JohanEngelen:
My plan:
- use Travis for LLVM 3.5 - 3.8 testing  (download binaries from llvm.org).
- use CircleCI for LLVM3.9 testing
Then, we can take CircleCI failures a bit less serious, but require Travis to be green for PRs.
How's that sound?

JohanEngelen:
Looking at previous builds, the speed is about the same for the LLVM 3.8 (~38-40 mins). Travis gives us 5 concurrent builds; CircleCI 4. We could move the 6th Travis build to CircleCI to spread it better. Hope that someone else feels up to that task though.

PR #1560, opened 2016-06-14
Fix ARM soft-float ABI when passing structs with floating-point numbers

joakim-noah:
Some tests from the dmd testsuite fail because of this, specifically `ctest10` from `cabi1` and a handful from `ldc_cabi1`.  Once I put this fix in, the entire dmd testsuite passes natively on Android/ARM.  Presumably Dan put this in for `armhf` but iOS is `armel` also, isn't it?  Once @smolt and I hash out how to make this work across all tested ARM systems, I'll modify this patch and remove the WIP tag, then someone can merge.

### DIFF
--- a/gen/abi-arm.cpp
+++ b/gen/abi-arm.cpp
@@ -16,6 +16,7 @@
 #include "gen/abi.h"
 #include "gen/abi-generic.h"
 #include "gen/abi-arm.h"
+#include "llvm/Target/TargetMachine.h"
 
 struct ArmTargetABI : TargetABI {
   HFAToArray hfaToArray;
@@ -38,7 +39,9 @@ struct ArmTargetABI : TargetABI {
       return rt->ty == Tsarray || rt->ty == Tstruct;
 
     return rt->ty == Tsarray ||
-           (rt->ty == Tstruct && rt->size() > 4 && !isHFA((TypeStruct *)rt));
+           (rt->ty == Tstruct && rt->size() > 4 &&
+             (gTargetMachine->Options.FloatABIType == llvm::FloatABI::Soft ||
+             !isHFA((TypeStruct *)rt)));
   }
 
   bool passByVal(Type *t) override {


joakim-noah:
@redstar, can you let me know how this will affect your ARM board too?

smolt:
iOS is a don't care for this situation, it is a variant of older APCS (armel) and doesn't use arm-abi.cpp.

Android must be using AAPCS to work so well with arm-abi.cpp and this one change.  Android is softfp for passing and returning function values (uses general registers), right?  That would explain it.

In targetmachine.cpp `targetOptions.FloatABIType = llvm::FloatABI::Soft`.  Perhaps you can use `FloatABIType` to decide to use `isHFA()`.

For AAPCS reference http://infocenter.arm.com/help/topic/com.arm.doc.ihi0042f/IHI0042F_aapcs.pdf, softfp only wants "Result Return" rules in Section 5.4, and not Section 6.1.2.2 which specifies additional rules for returning result in floating point registers (`llvm::FloatABI::Hard`).

joakim-noah:
>Android must be using AAPCS to work so well with arm-abi.cpp and this one change. Android is softfp for passing and returning function values (uses general registers), right? That would explain it.

Yes, Android is AAPCS and softfp by default, though [they also supported hard-float for some time](https://android.googlesource.com/platform/ndk/+/master/docs/HardFloatAbi.md).

>In targetmachine.cpp targetOptions.FloatABIType = llvm::FloatABI::Soft. Perhaps you can use FloatABIType to decide to use isHFA().

I figured a check like that could be made, but I first wanted to make sure you actually needed that invocation of `isHFA`.

>For AAPCS reference http://infocenter.arm.com/help/topic/com.arm.doc.ihi0042f/IHI0042F_aapcs.pdf, softfp only wants "Result Return" rules in Section 5.4, and not Section 6.1.2.2 which specifies additional rules for returning result in floating point registers (llvm::FloatABI::Hard).

Does `armhf` actually do that?  If so, I'll add some logic to call `isHFA` only in that case.

smolt:
> Does armhf actually do that?

Yes, returns float-only structs in up-to 4 VFP registers.   Very efficient when calling math functions.  runnable/ldc_cabi1 was created to test these cases.

> If so, I'll add some logic to call isHFA only in that case.

I think that will work well.  I can test PR branch on `armhf` (Raspberry Pi) when its updated.

joakim-noah:
Updated to put in the logic to disable `isHFA` for a soft ABI, Dan, let me know if this works on `armhf` too.

I tested by compiling and running the `cabi1` test module both with and without `hardfloat` and running the druntime/phobos tests too.

smolt:
kicked off armhf test...

smolt:
ldc_cabi1 still passes on armhf.  Looks good to me, and I'll run full test suite to be safe.

smolt:
> I'll run full test suite to be safe.

All good.

smolt:
@joakim-noah ok with merging this?

joakim-noah:
Yep, just didn't want to do it myself.

joakim-noah:
Thanks for the review.

Issue #1433, opened 2016-04-14
readln(buf) + slicing reads wrong data / segmentation fault


zoodyy:
Dear LDC-Team,
I've just stumbled across a nasty issue..
readln(buf) in combination with direct slicing produces wrong data on Windows and segfaults on Linux (tested with 0.17.1 and 1.0 alpha).
`char[] line = buf[0..readln(buf)].chomp();`
Attached is a minimal example as well as some input data.
The example works totally fine on DMD and GDC (in debug and release).

[input.txt](https://github.com/ldc-developers/ldc/files/219737/input.txt)
[readlnError.d.txt](https://github.com/ldc-developers/ldc/files/219738/readlnError.d.txt)

LDC works if you change above line to something like:
`ptrdiff_t length = readln(buf); char[] line = buf[0..length].chomp();`

Best regards!

kinke:
Thanks for the report. I guess this is related to https://github.com/ldc-developers/ldc/issues/1327 in the sense that the rvalue `readln(buf)` is probably evaluated **after** evaluating `buf` for slicing.

kinke:
Fixed by #1566.

PR #1562, opened 2016-06-15
Continue with DValue refactoring and fix issue #1327

kinke:
* Introduce a DRValue base class to be able to discriminate between DLValues and DRValues (e.g., function parameters).
* Let DValue::getRVal() return each DValue's value as DRValue. This allows to convert a DLValue to a DRValue while retaining the D type, something we've previously lost when returning the low-level rvalue directly.
* Let the DtoR/LVal() helpers be the only way to convert a DValue to a low-level value.

### DIFF
--- a/gen/dvalue.h
+++ b/gen/dvalue.h
@@ -40,7 +40,7 @@ class DSpecialRefValue;
 class DSliceValue;
 class DFuncValue;
 
-/// Associates an immutable low-level value with an immutable D type.
+/// Represents an immutable pair of LLVM value and associated D type.
 class DValue {
 public:
   Type *const type;
@@ -87,7 +87,10 @@ protected:
   DRValue(Type *t, llvm::Value *v);
 };
 
-/// Represents an immediate D value via a low-level rvalue.
+/// Represents an immediate D value (simple rvalue with no special properties
+/// like being a compile-time constant) via a low-level rvalue.
+/// Restricted to primitive types such as pointers (incl. class references),
+/// integral and floating-point types.
 class DImValue : public DRValue {
 public:
   DImValue(Type *t, llvm::Value *v) : DRValue(t, v) {}
@@ -95,7 +98,7 @@ public:
   DImValue *isIm() override { return this; }
 };
 
-/// Represents a constant D value via a low-level constant.
+/// Represents a D compile-time constant via a low-level constant.
 class DConstValue : public DRValue {
 public:
   DConstValue(Type *t, llvm::Constant *con);
@@ -105,7 +108,7 @@ public:
   DConstValue *isConst() override { return this; }
 };
 
-/// Represents a D null constant.
+/// Represents a D compile-time null constant.
 class DNullValue : public DConstValue {
 public:
   DNullValue(Type *t, llvm::Constant *con) : DConstValue(t, con) {}
@@ -139,7 +142,7 @@ public:
   DFuncValue *isFunc() override { return this; }
 };
 
-/// Represents a D value in memory via a low-level lvalue.
+/// Represents a D value in memory via a low-level lvalue (pointer).
 /// This doesn't imply that the D value is an lvalue too - e.g., we always
 /// keep structs and static arrays in memory.
 class DLValue : public DValue {
--- a/gen/dvalue.h
+++ b/gen/dvalue.h
@@ -18,7 +18,6 @@
 #define LDC_GEN_DVALUE_H
 
 #include "root.h"
-#include <cassert>
 
 class Type;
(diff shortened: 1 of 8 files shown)


kinke:
The 2nd commit demonstrates what I was aiming for with that DValue refactoring, but it's only a beginning, there are surely more issues of this kind. #1327 is fixed with this at least.

kinke:
The 3rd commit fixes #1433.

JohanEngelen:
Could you add some tests for these fixes?

kinke:
Yeah I'll add some tests. They should probably go upstream.

kinke:
I've moved the #1433 fix to its own PR https://github.com/ldc-developers/ldc/pull/1566, as it is yet unclear whether that's a bug or not.

kinke:
I'm not perfectly happy with DLValue::getRVal() creating a new DImValue instance on the heap, but I surely don't want to change everything to a `std::shared_ptr<DValue>` or something like this.
I do really like the ability to convert a DLValue to a DRValue representing its current state though, as it allows us to work even more with DValues (instead of LLValues) in the future. Keeping the D type around and the info whether it's an l- or rvalue is nice and eliminates assumptions about a stand-alone LLValue.

kinke:
Any concerns or has noone found the time to review this yet?

JohanEngelen:
I haven't reviewed this in detail, but I have no concerns :)

gen/dvalue.h
```diff
@@ -31,35 +30,38 @@ class Type;
...
 class DConstValue;
 class DNullValue;
 class DLValue;
 class DSpecialRefValue;
-class DFuncValue;
 class DSliceValue;
+class DFuncValue;
 
-// base class for d-values
+/// Associates an immutable low-level value with an immutable D type.
```
klickverbot: Nitpick: "immutable" is a bit misleading imho (in any other context, "immutable D type" would refer to the `immutable` keyword – maybe `An immutable pair of LLVM value and associated D type.`?

gen/dvalue.h
```diff
@@ -69,22 +71,32 @@ class DValue {
...
+
+  DRValue *isRVal() override { return this; }
+
+protected:
+  DRValue(Type *t, llvm::Value *v);
 };
 
-// immediate d-value
-class DImValue : public DValue {
+/// Represents an immediate D value via a low-level rvalue.
```
klickverbot: Would be nice to document what an "immediate" D value is – I know I found this confusing when first starting to work with/on DValue. (i.e. an rvalue without any other special traits)

gen/dvalue.h
kinke: Very well, done.

gen/dvalue.h
```diff
@@ -69,22 +71,34 @@ class DValue {
...
+/// Suited for primitive low-level types like pointers (incl. class references),
+/// integral and floating-point types.
```
kinke: I tried to elaborate a bit here, but my understanding of 'immediate' is pretty vague too. ;)

klickverbot: Thanks – I think pretty much all there even is to understand is that it is a D rvalue represented by a straight LLVM value, and without any special further properties (e.g. being null, etc).

gen/dvalue.h
```diff
@@ -69,22 +71,35 @@ class DValue {
...
+protected:
+  DRValue(Type *t, llvm::Value *v);
 };
 
-// immediate d-value
-class DImValue : public DValue {
+/// Represents an immediate D value (simple rvalue with no special properties
+/// like being a compile-time constant) via a low-level rvalue.
+/// Restricted to primitive types such as pointers (incl. class references),
+/// integral and floating-point types.
```
kinke: I've incorporated your latest comment here.

kinke:
The tests were green and I've only touched some comments. Merging eagerly. dmd-testsuite test will follow soon.

kinke:
Test: https://github.com/ldc-developers/dmd-testsuite/commit/030de81cdd4a67000a8bcc14c53408cfba20b118

JohanEngelen:
git bisect tells me that this PR (specifically commit a3adb403de1554cab73cfa9df88dc5a3dfc7940f) is causing a unittest failure in Weka's codebase.
I'll look into whether this is because of a bug in Weka's code, or a new codegen bug. The code is involving threads/fibers, if that is a good hint for anyone.

JohanEngelen:
I found the bugged code. It looks like a bug in Weka's code to me. Need to discuss with the author.

JohanEngelen:
Found the problem: #1617

PR #1566, opened 2016-06-18
Fix issue #1433

kinke:
Slice expressions: load the base's pointer after the bounds expressions have been evaluated, since they may affect the pointer if the base is an lvalue, see issue #1433.
Not sure if this is correct/desirable, see http://forum.dlang.org/thread/blkpasccoceipqvbkyla@forum.dlang.org.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -1072,19 +1072,23 @@ public:
     PGO.setCurrentStmt(e);
 
     // value being sliced
+    Type *const etype = e->e1->type->toBasetype();
+    LLValue *eptr = nullptr;
     LLValue *elen = nullptr;
-    LLValue *eptr;
-    DValue *v = toElem(e->e1);
 
-    Type *etype = e->e1->type->toBasetype();
-    if (etype->ty == Tpointer) {
-      // pointer slicing
-      assert(e->lwr);
-      eptr = DtoRVal(v);
-    } else {
-      // array slice
-      eptr = DtoArrayPtr(v);
-    }
+    // evaluate the base expression but delay getting its pointer until the
+    // potential bounds have been evaluated
+    DValue *v = toElem(e->e1);
+    auto getBasePointer = [e, v, etype]() {
+      if (etype->ty == Tpointer) {
+        // pointer slicing
+        assert(e->lwr);
+        return DtoRVal(v);
+      } else {
+        // array slice
+        return DtoArrayPtr(v);
+      }
+    };
 
     // has lower bound, pointer needs adjustment
     if (e->lwr) {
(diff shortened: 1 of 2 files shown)


kinke:
This is most likely NOT the way to go, but as usual, the NG discussion died very quickly. We'll probably have to fix $ and treat it as rvalue, not as lvalue anymore.

kinke:
Hmm nope, I tend towards this solution again. This works with this PR (without it, the returned slices are all based on the old buffer) and shows 2 related bugs (IMO at least) of DMD 2.074:

* slicing a) a dynamic array variable and b) a reference to that same dynamic array isn't the same thing for DMD
* loading $ once at the beginning of the slice-expression (DMD) instead of loading the current `<base>.length` value each time it is accessed (LDC) may result in wrong bounds checks and thus easily allows producing invalid slices in safe code with enabled bounds checks

```D
@safe:

int step;
int[] globalArray;

void reset(int initialStep)
{
    step = initialStep;
    globalArray = [ -1, -2, -3, -4 ];
}

int[] getBaseSlice()
{
    assert(step++ == 0);
    return globalArray;
}

ref int[] getBaseSliceRef()
{
    assert(step++ == 0);
    return globalArray;
}

int getLowerBound(size_t dollar)
{
    assert(step++ == 1);
    assert(dollar == 4);
    globalArray = null;
    return 1;
}

int getUpperBound(size_t dollar, size_t expectedDollar)
{
    assert(step++ == 2);
    assert(dollar == expectedDollar);
    globalArray = [ 1, 2, 3 ];
    return 3;
}

// LDC issue #1433
void main()
{
    reset(0);
    auto r = getBaseSlice()[getLowerBound($) .. getUpperBound($, 4)];
    assert(r == [ -2, -3 ]); // old buffer

    version(LDC) // LDC treats $ as lvalue and loads <base>.length each time it is accessed
        enum expectedDollar = 0;
    else // DMD apparently treats it as rvalue and loads it once at the beginning (=> wrong bounds check)
        enum expectedDollar = 4;

    reset(1);
    r = globalArray[getLowerBound($) .. getUpperBound($, expectedDollar)];
    assert(r == [ 2, 3 ]); // new buffer

    reset(0);
    r = getBaseSliceRef()[getLowerBound($) .. getUpperBound($, expectedDollar)];
    version(LDC)
        assert(r == [ 2, 3 ]); // new buffer
    else
        assert(r == [ -2, -3 ]); // old buffer

    testBoundsCheck();
}

void testBoundsCheck() @trusted // @trusted needed for catching Errors, otherwise @safe
{
    import core.exception : RangeError;
    reset(1);
    try
    {
        auto r = globalArray[getLowerBound($) .. 2]; // null[1 .. 2]
        assert(0); // fails for DMD
    }
    catch (RangeError) {}
}
```

@ibuclaw: Would be interesting to see how GDC handles this code.

kinke:
Thanks Iain, so you're already set. I'll file 2 DMD issues regarding this [sometime soon ;)].

kinke:
Filed https://issues.dlang.org/show_bug.cgi?id=17363 and https://issues.dlang.org/show_bug.cgi?id=17364.

kinke:
Test case amended and ready to be merged.

kinke:
Will merge in the next few days if there are no (diverging) opinions.

PR #1569, opened 2016-06-18
Add IR test for optimized array (in)equality with null

kinke:
As requested in https://github.com/ldc-developers/ldc/pull/1393. Let's close https://github.com/ldc-developers/ldc/issues/894 with this.


JohanEngelen:
nice!

Issue #958, opened 2015-05-31
ICE when providing mismatched _d_assert_msg() declaration


timotheecour:
ldc2 -c -oq -od=/tmp/d08 fun.d

fun.d:

``` d
import std.string;
void fun(){
  format("");
}
extern(C)void _d_assert_msg();
```

ldc2 --version
LDC - the LLVM D compiler (cf0589):
  based on DMD v2.066.1 and LLVM 3.6.1
  Default target: x86_64-apple-darwin14.3.0
  Host CPU: core-avx2

0  ldc2                     0x0000000104878ad1 llvm::sys::PrintStackTrace(__sFILE_) + 57
1  ldc2                     0x0000000104879027 SignalHandler(int) + 217
2  libsystem_platform.dylib 0x00007fff9a658f1a _sigtramp + 26
3  ldc2                     0x0000000104c0ec36 nuls + 69742
4  ldc2                     0x0000000104878eb6 abort + 22
5  ldc2                     0x0000000104878ea0 abort + 0
6  ldc2                     0x000000010480c0d8 llvm::CallInst::init(llvm::Value_, llvm::Twine const&) + 0
7  ldc2                     0x0000000103f38f99 llvm::IRBuilder<true, llvm::ConstantFolder, llvm::IRBuilderDefaultInserter<true> >::CreateCall(llvm::Value_, llvm::ArrayRef<llvm::Value_>, llvm::Twine const&) + 201
8  ldc2                     0x0000000103f3fbde llvm::CallSite IRState::CreateCallOrInvoke<llvm::SmallVector<llvm::Value*, 3u> >(llvm::Value_, llvm::SmallVector<llvm::Value_, 3u> const&, char const_) + 686
9  ldc2                     0x0000000103f39af8 DtoAssert(Module_, Loc&, DValue_) + 312
10 ldc2                     0x0000000103f7427b ToElemVisitor::visit(AssertExp_) + 763
11 ldc2                     0x0000000103f70236 toElemDtor(Expression_) + 614
12 ldc2                     0x0000000103f5f502 ToIRVisitor::visit(ExpStatement_) + 130
13 ldc2                     0x0000000103f5f594 ToIRVisitor::visit(CompoundStatement_) + 116
14 ldc2                     0x0000000103f5f3ee codegenFunction(Statement_, IRState_) + 190
15 ldc2                     0x0000000103f35eee DtoDefineFunction(FuncDeclaration_) + 3982
[...]

PR #1570, opened 2016-06-19
Check existing runtime function declarations for matching type

kinke:
This fixes issue #958 by checking the function type of an existing function declaration with the same runtime function name inside the codegen'd module.
This should normally be an existing forward declaration from a previous runtime call, but the user is free to declare (and even implement) it too.

The LOC for the error is empty as the provided `loc` may likely be from a template instantiation from another module.

### DIFF
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -126,22 +126,24 @@ llvm::Function *getRuntimeFunction(const Loc &loc, llvm::Module &target,
                                    const char *name) {
   checkForImplicitGCCall(loc, name);
 
-  if (!M) {
+  if (!M)
     initRuntime();
-  }
 
-  LLFunction *fn = target.getFunction(name);
-  if (fn) {
-    return fn;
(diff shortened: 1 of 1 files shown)


JohanEngelen:
@kinke Can you add a testcase for it too?

kinke:
I can, it'll be a `fail_compilation` test in dmd-testsuite checking the error msg. Not now though. ;)

kinke:
Any concerns? This always requires the 2nd lookup in the special runtime module to check the type, but that slight overhead should be worth it.

Test will follow once merged.

JohanEngelen:
I would have expected a typecheck of runtime functions during semantic analysis, such that the error message tells the user where the conflicting definition happens. But DMD 2.071 does not do any checking: `extern (C) void _d_assert_msg() {}` does not result in a compile error or anything... 

The special runtime module is small so the extra lookup is fast I think.

PR #1571, opened 2016-06-20
[Travis] Do the LLVM binaries work on Ubuntu 12?

JohanEngelen:
(seem to for Boost.Hana)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: d
-sudo: required
-dist: trusty
+sudo: false
 
 matrix:
   include:


kinke:
Oh yeah baby, looking good. It also fixes the sharedlib issue. WTF?!

kinke:
Damn, multilib still times out. ~1h 20 min?! Are the LLVM binaries slower than the packages we've used previously?

JohanEngelen:
> or the host machines we get this way are just slower.

With this PR we should be back on the same systems as before instead of switching to the Ubuntu 14 machines.

JohanEngelen:
merge?

kinke:
Yes, the longer time-out alone is worth it.

PR #1572, opened 2016-06-20
Add IR-to-Objectcode caching with `-ir2obj-cache=<cache dir>`.

JohanEngelen:
This adds a caching mechanism such that compiling the same source twice should be faster as it will use the previously generated object file.
The caching granularity is per LLVM module.

With  `-ir2obj-cache=<cache dir>`, LDC does:
1. Parse/analyse/etc the code
2. Generate LLVM IR
**3. Create hash of LLVM IR  (NEW)**
**4. Do hash lookup in cachedir.  (NEW)**
   **If hash found: output symlink to cached object file and exit.  (NEW)**
5. Optimization
6. Output object file
**7. Store hash-objectfile in cache dir.  (NEW)**

Machine codegen takes a lot of time, so this caching results in a significant speed up (~65sec -> ~39sec on testcase). Generating the hash is done by hashing the raw bitcode output from LLVM. This is not for free and takes quite some time: it takes time to generate a 200MB bitcode file, I think the hash time itself is not noticable. Future work could try to get a (unique) hash quicker through some other means.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -187,6 +187,10 @@ static cl::opt<bool, true> unittest("unittest",
                                     cl::desc("Compile in unit tests"),
                                     cl::location(global.params.useUnitTests));
 
+cl::opt<std::string>
+    ir2objCacheDir("ir2obj-cache", cl::desc("Use <cache dir> to cache object files for whole IR modules (experimental)"),
+            cl::value_desc("cache dir"), cl::Prefix);
+
 static StringsAdapter strImpPathStore("J", global.params.fileImppath);
 static cl::list<std::string, StringsAdapter>
     stringImportPaths("J", cl::desc("Where to look for string imports"),
--- a/driver/ir2obj_cache.cpp
+++ b/driver/ir2obj_cache.cpp
@@ -1,0 +1,172 @@
+//===-- driver/ir2obj_cache.cpp -------------------------------------------===//
+//
+//                         LDC – the LLVM D compiler
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+//
+// Contains LLVM IR to object code cache functionality.
+//
+// After LLVM IR codegen, the LLVM IR module is hashed for lookup in the cache
+// directory. If the cache directory contains the object file <hash>.o,
+// that file is used and machine code gen is skipped entirely. If the cache
+// doesn't contain that file, machine codegen happens as normal and the object
+// code is added to the cache.
+// The goal is to speed up successive builds of large codebases after minor
+// changes that trigger recompilation of many files but with little effective
+// changes (in the extreme case, adding a comment in a "globals.d").
+//
+// The current limitation is that hashing and cache look-up are done with
+// whole-module granularity. Future work could attempt to do this for module
+// "fragments".
+//
+// The hash depends on the IR code (obviously), but also on the compiler+LLVM
+// versions and several compile flags (e.g. -O*, -mcpu, and -mattr).
+//
+//===----------------------------------------------------------------------===//
+
+#include "driver/ir2obj_cache.h"
+
+#include "ddmd/errors.h"
+#include "driver/cl_options.h"
+#include "driver/ldc-version.h"
+#include "gen/logger.h"
+#include "gen/optimizer.h"
+
+#include "llvm/Bitcode/ReaderWriter.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/MD5.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace {
+
+/// A raw_ostream that creates a hash of what is written to it.
+/// This class does not encounter output errors.
+/// There is no buffering and the hasher can be used at any time.
+class raw_hash_ostream : public llvm::raw_ostream {
+  llvm::MD5 hasher;
+
+  /// See raw_ostream::write_impl.
+  void write_impl(const char *ptr, size_t size) override {
+    hasher.update(
+        llvm::ArrayRef<uint8_t>(reinterpret_cast<const uint8_t *>(ptr), size));
+  }
+
+  uint64_t current_pos() const override { return 0; }
+
+public:
+  raw_hash_ostream() { SetUnbuffered(); }
+  ~raw_hash_ostream() override {}
+
+  void flush() = delete;
+
+  void finalResult(llvm::MD5::MD5Result &result) { hasher.final(result); }
+  void resultAsString(llvm::SmallString<32> &str) {
+    llvm::MD5::MD5Result result;
+    hasher.final(result);
+    llvm::MD5::stringifyResult(result, str);
+  }
+};
+
+const char *cacheObjectExtension() {
+  return global.params.targetTriple->isOSWindows() ? global.obj_ext_alt
+                                                   : global.obj_ext;
+}
+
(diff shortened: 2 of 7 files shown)


JohanEngelen:
now compiles with LLVM != 3.8

driver/cl_options.cpp
```diff
@@ -187,6 +187,10 @@ static cl::opt<bool, true> unittest("unittest",
                                     cl::desc("Compile in unit tests"),
                                     cl::location(global.params.useUnitTests));
 
+cl::opt<std::string>
+    ir2objCacheDir("ir2obj-cache", cl::desc("Write&read LLVM IR to object code cache in <cache dir> directory"),
```
klickverbot: Maybe something like "Use <cache dir> to cache object files for specific LLVM IR modules"? The fact `ir2obj-cache` has something to do with an IR to object code cache is fairly obvious.

driver/ir2obj_cache.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+#include "llvm/Bitcode/ReaderWriter.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/MD5.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace {
+
+/// A raw_ostream that creates a hash of what is written to it.
```
klickverbot: Is this from LLVM? Should probably point that out because of the slightly different license.

JohanEngelen: Hm, actually pretty much only the comment is from LLVM.

driver/cl_options.cpp
JohanEngelen: Slight improvement "Use <cache dir> to cache object files for specific LLVM IR modules"?

How about the flag itself?

klickverbot: _lol_ indeed – that must be a HTML tag escaper gone crazy or something along these lines. I edited the old comment to work around it.

As for the flag itself, if you advertise the caching functionality as "IR to object file cache", then I suppose it is about as clear as it gets. I've been trying to think about alternative names, but everything involving "codegen" somehow seems to be too unspecific.

driver/ir2obj_cache.cpp
klickverbot: Ah, okay – I just vaguely remembered reading the comment in the upstream ThinLTOCodeGenerator stuff.

driver/ir2obj_cache.cpp
```diff
@@ -0,0 +1,163 @@
...
+void recoverObjectFile(llvm::StringRef cacheObjectHash,
+                       llvm::StringRef objectFile) {
+  llvm::SmallString<128> cacheFile;
+  storeCacheFileName(cacheObjectHash, cacheFile);
+
+  llvm::sys::fs::remove(objectFile);
+
+  IF_LOG Logger::println("SymLink output to cached object file: %s -> %s",
+                         objectFile.str().c_str(), cacheFile.c_str());
+  if (llvm::sys::fs::create_link(cacheFile.c_str(), objectFile)) {
```
klickverbot: What does this actually do on Windows?

tests/linking/ir2obj_caching.d
```diff
@@ -0,0 +1,16 @@
+// Test recognition of -ir2obj-cache commandline flag
+
+// RUN: %ldc -ir2obj-cache=%T/cachedirectory %s -vv | FileCheck --check-prefix=FIRST %s \
+// RUN: && %ldc -ir2obj-cache=%T/cachedirectory %s -vv | FileCheck --check-prefix=SECOND %s
+
+
+// FIRST: Use IR-to-Object cache in {{.*}}cachedirectory
+// Don't check whether the object is in the cache on the first run, because if this test is ran twice the cache will already be there.
```
klickverbot: Can't we just nuke the cache directory first?

driver/toobj.cpp
```diff
@@ -332,30 +343,62 @@ class AssemblyAnnotator : public AssemblyAnnotationWriter {
...
-
   // There is no integrated assembler on AIX because XCOFF is not supported.
   // Starting with LLVM 3.5 the integrated assembler can be used with MinGW.
   bool const assembleExternally =
       global.params.output_o &&
       (NoIntegratedAssembler ||
        global.params.targetTriple->getOS() == llvm::Triple::AIX);
 
+  // Use cached object code if possible
+  bool useIRToObjCache = !opts::ir2objCacheDir.empty();
```
klickverbot: Very, very minor nitpick: I'd go for consistent rendition as "useIR2ObjCache", if your are using it for the user-facing parts (there is, of course, some precedent in LLVM – mem2reg, etc.).

driver/ir2obj_cache.h
```diff
@@ -0,0 +1,29 @@
+//===-- driver/ir2obj-cache.h -----------------------------------*- C++ -*-===//
```
klickverbot: `-` vs. `_`. ;)

tests/linking/ir2obj_caching.d
JohanEngelen: I don't know how to remove a directory with a cross-platform command and without errorring if the directory does not exist.

driver/ir2obj_cache.cpp
JohanEngelen: Windows also has symlinks. According to LLVM's documentation:
"The link may be a soft or a hard link, depending on the platform. The caller may not assume which one. Currently on windows it creates a hard link since soft links require extra privileges. On unix, it creates a soft link since hard links don't work on SMB file systems."

klickverbot: I suppose that assumes that nobody is trying to build on FAT32 or weird network shares? As long as the error message is clear enough, this isn't going to be an issue, though.

driver/ir2obj_cache.cpp
```diff
@@ -0,0 +1,163 @@
...
+//===----------------------------------------------------------------------===//
+//
+// Contains LLVM IR to object code cache functionality.
+//
+// After LLVM IR codegen, the LLVM IR module is hashed for lookup in the cache
+// directory. If the cache directory contains the object file <hash>.o,
+// that file is used and machine code gen is skipped entirely. If the cache
+// doesn't contain that file, machine codegen happens as normal and the object
+// code is added to the cache.
+// The goal is to speed up successive builds with only minor changes.
```
klickverbot: Maybe "successive builds of a big codebase with several object files after minor changes" or something like that, to clarify that this doesn't apply to `-singleobj` (or equivalent) builds. The fact that there is a comment in the first place is of course quite a lot for LDC already, but I think it would make sense to be clear about this "limitation" (i.e. us not caching module fragments) in the user documentation and initial announcements.

driver/ir2obj_cache.cpp
JohanEngelen: Hmm... perhaps add the word "symbolic link" or something to the error message?
Before, I used `llvm::sys::fs::copy`, but then I "measured" that for Weka's huge object files that actually takes more than a second on my PC and was happy to find LLVM's (semi-?)cross-platform `create_link`. :)

JohanEngelen:
@klickverbot I think I've addressed all your comments.

JohanEngelen:
ThinLTO is definitely something interesting to look into for speeding up builds too.
For future reference: http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html

I'll add "(experimental)" and merge. Perhaps someone can work-out a faster way of obtaining a unique ID for the IR, which would then break existing hashes. If people like to store them for a long time and have them be reusable, we can tell them: "hey, it says 'experimental'!" ;-).

PR #1577, opened 2016-06-22
Implement cross-module inlining

JohanEngelen:
An "inlined" function will be defined with the LLVM attribute `available_externally`, so that LLVM can decide to actually inline the function or not.

`pragma(inline, false)` --> the function receives the `noinline` attribute.
`pragma(inline, true)`  --> the function receives the `alwaysinline` attribute and cross-module inlining for this function is performed even at `-O0`.

The current heuristic for (defining available_externally functions to enable) inlining functions is: `true`  :-)
Very aggressive indeed; good for testing :). Perhaps a simple statement count heuristic can be used . 

Functions that have `__FILE__` as default _template_ parameter need to always be inlined `pragma(inline, true)`, otherwise it results in link errors. See the forum discussion:
https://forum.dlang.org/thread/induraqegvpyomvzxwmf@forum.dlang.org

To inline a function, semantic analysis needs to be completed for that function (DMD does the same). Doing that however, may add new symbols to the module member list for codegenning. This means that the order of how modules are codegenned needs to be reversed (or, members[0] needs to be codegenned last, as that one receives all new symbols). This showed up while testing Weka's codebase and cost me a lot of time. Unfortunately, I have not managed to recreate this problem in a testcase.

Tested succesfully on Weka's codebase.


gen/functions.cpp
```diff
@@ -527,8 +527,15 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
     gIR->mainFunc = func;
   }
 
+  // Set inlining attribute
   if (fdecl->neverInline) {
```
klickverbot: This is of course completely orthogonal to your PR, but we should probably make LDC_never_inline just set ::inlining appropriately instead of having ::neverInline, now that the former exists.

gen/functions.cpp
```diff
@@ -42,6 +42,123 @@
...
+bool isInlineCandidate(FuncDeclaration &fdecl) {
+  if (fdecl.inlining == PINLINEalways)
+    return true;
+
+  return true;
+}
+
+/// Check whether the frontend knows that the function is already defined
+/// in some other module (see DMD's FuncDeclaration::toObjFile).
+bool alreadyOrWillBeDefined(FuncDeclaration &fdecl) {
```
klickverbot: Instead of recomputing the value here, would it make sense to just check whether the function exists already in the current llvm::Module (either by literally checking the llvm::Module, or just the codegen status)? If it is already defined, this should be much quicker than iterating over the AST. If not, just go ahead potentially emitting it. If do end up actually defining the regular way, just modify the linkage appropriately.

gen/functions.cpp
```diff
@@ -42,6 +42,123 @@
...
+    assert(fdecl.semanticRun >= PASSsemantic3done);
+  }
+
+  // For naked functions (inline assembly), we emit the assembly directly as
+  // globals in the text section. Emitting them during this inline pass will
+  // therefore result in multiple definitions. Solution: don't try to inline
+  // them.
+  // These naked functions don't appear to be inlined anyway, so it is pointless
+  // at this moment to try.
+  // FuncDeclaration::naked is set by the AsmParser during semantic analysis.
```
klickverbot: That's a bit of an odd comment to find here, given that `naked` is an upstream DMD member – when else would it be set if not during semantic analysis?

gen/functions.cpp
```diff
@@ -42,6 +42,123 @@
...
+      return false;
+    }
+    assert(fdecl.semanticRun >= PASSsemantic3done);
+  }
+
+  // For naked functions (inline assembly), we emit the assembly directly as
+  // globals in the text section. Emitting them during this inline pass will
+  // therefore result in multiple definitions. Solution: don't try to inline
+  // them.
+  // These naked functions don't appear to be inlined anyway, so it is pointless
```
klickverbot: They are not, and will never be. Not only would LLVM need to add a machine code-level inliner – even if they decided to do that, inlining arbitrary assembly would break semantics. So, no need to be vague.

gen/functions.cpp
JohanEngelen: I'll clarify it :-)
The comment should say that the FuncDeclaration::naked check cannot be done before semantic3 (important: "3").

gen/functions.cpp
JohanEngelen: Ah, thanks. I thought DMD perhaps had a way of inlining naked asm functions.

gen/functions.cpp
JohanEngelen: If it is already defined in any module during this compilation run, we wouldn't reach here (because the call-chains that call this function already do a check on codegen status). If my understanding is correct, the "already" means something like "should be defined in some other module that we have to link to".
I tried defining aggressively, not doing this check basically, but then codegen runs afoul with some complex ASTs (the call stack also becomes very long, with annoyingly nested `-vv` output).
For the case that we do need codegen the regular way, indeed there is code at the start of `DtoDefineFunction` to fixup the linkage.

gen/declarations.cpp
```diff
@@ -432,8 +435,17 @@ class CodegenVisitor : public Visitor {
 
     // FIXME: This is #673 all over again.
     if (!decl->needsCodegen()) {
-      Logger::println("Does not need codegen, skipping.");
-      return;
+      // Force codegen if this is a templated function with pragma(inline, true).
+      if ((decl->members->dim == 1) &&
```
klickverbot: I'd probably move this in front of the (potentially fairly expensive) `needsCodegen` check. The overhead is probably insignificant, though, unless there is a huge amount of pragma(inline, true) templates.

Also, wouldn't it make more sense to check `decl->onemember`? My DMD AST working knowledge isn't great, but I would suppose that this would also allow you to handle an always-inline `template foo() { enum randomStuff = false; void foo() {} }`.

gen/functions.cpp
klickverbot: > but then codegen runs afoul with some complex ASTs

What would be an example of the symptoms?

gen/functions.cpp
```diff
@@ -42,6 +42,142 @@
...
+    if (f->isNested()) {
+      f = f->toParent2()->isFuncDeclaration();
+    } else {
+      break;
+    }
+  }
+  return true;
+}
+
+/// If true: define this function with externally_available linkage, for
```
klickverbot: Apparently, the only thing I can contribute here is to complain about comments, but the two "if true"s with different meanings are a bit confusing for me. ;)

Maybe:

```
/// Returns whether `fdecl` should be emitted with externally_available
/// linkage to make it available for inlining.
///
/// If true, `semantic3` will have been run on the function already.
```

gen/functions.cpp
```diff
@@ -42,6 +42,142 @@
...
+  if (fdecl.isFuncAliasDeclaration()) {
+    IF_LOG Logger::println("isFuncAliasDeclaration() == true");
+    return false;
+  }
+  if (!fdecl.fbody) {
+    IF_LOG Logger::println("No function body available for inlining");
+    return false;
+  }
+
+  // Disable inlining functions from object.d because of TypeInfo related
```
klickverbot: What "TypeInfo related issue" is this?

gen/functions.cpp
```diff
@@ -42,6 +42,142 @@
...
+// TODO: move to its own file?
+namespace {
+
+// Use a heuristic to determine if it could make sense to inline this fdecl.
+// Note: isInlineCandidate is called _before_ full semantic analysis of fdecl.
+bool isInlineCandidate(FuncDeclaration &fdecl) {
+  if (fdecl.inlining == PINLINEalways)
+    return true;
+
+  return true;
```
klickverbot: I presume you are going to augment this before merging? I would at least change it to default to `false`, along with a comment that we need some heuristic based on size.

gen/functions.cpp
```diff
@@ -42,6 +42,142 @@
 #include "llvm/IR/CFG.h"
 #include <iostream>
 
+////////////////////////////////////////////////////////////////////////////////
+// Function inlining
+// TODO: move to its own file?
+namespace {
+
+// Use a heuristic to determine if it could make sense to inline this fdecl.
+// Note: isInlineCandidate is called _before_ full semantic analysis of fdecl.
```
klickverbot: I'd find it much more helpful to specify which phase is guaranteed to have been run, instead of noting that "full" analysis (semantic3?) has not necessarily been done yet.

gen/functions.cpp
```diff
@@ -42,6 +42,142 @@
...
+    IF_LOG Logger::println("@weak functions cannot be inlined.");
+    return false;
+  }
+
+  if (!isInlineCandidate(fdecl))
+    return false;
+
+  IF_LOG Logger::println("Potential inlining candidate");
+
+  if (fdecl.semanticRun >= PASSsemantic3) {
```
klickverbot: Isn't this a much cheaper check than `alreadyOrWillBeDefined`/`hasWeakUDA`/`isInlineCandidate` (the latter once a proper heuristic has been added)?

gen/functions.cpp
```diff
@@ -457,6 +593,15 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
     fatal();
   }
 
+  // Check if fdecl should be defined too for cross-module inlining.
+  // If true, semantic is fully done for fdecl which is needed for some code
+  // below (e.g. code that uses fdecl->vthis).
+  bool defineAtEnd = defineAsExternallyAvailable(*fdecl);
```
klickverbot: `const`

gen/functions.cpp
klickverbot: Also, DMD style is to pass around pointers to AST nodes, so I'd probably stick to that.

gen/functions.cpp
```diff
@@ -712,12 +871,22 @@ void defineParameters(IrFuncTy &irFty, VarDeclarations &parameters) {
 
 } // anonymous namespace
 
-void DtoDefineFunction(FuncDeclaration *fd) {
+void DtoDefineFunction(FuncDeclaration *fd, bool availableExternally) {
```
klickverbot: Another tiny nitpick: Could you name this `makeAvailableExternally`, `asAvailableExternally` or something like that? Ever since going through (and removing) the old incomplete implementation, I always found the LLVM linkage name confusing in code.

gen/functions.cpp
```diff
@@ -760,21 +930,12 @@ void DtoDefineFunction(FuncDeclaration *fd) {
     return;
   }
 
-  // Check whether the frontend knows that the function is already defined
-  // in some other module (see DMD's FuncDeclaration::toObjFile).
-  for (FuncDeclaration *f = fd; f;) {
-    if (!f->isInstantiated() && f->inNonRoot()) {
+  if (!availableExternally) {
```
klickverbot: Nitpick: Fold `if`s?

gen/functions.cpp
JohanEngelen: Ah. I like to use references because then it is clear that nullptr is not a valid argument value.

klickverbot: I prefer to use references as much as possible too – it just seems weird to have pretty much a single instance of an AST node reference in the entire code base. ;) But as I said, that's a completely inconsequential thing.

JohanEngelen:
Thanks for your comments David. Trying to incorporate them all.

Indeed, a specific switch to turn this off is good to have (also for testing purposes).

JohanEngelen:
Note for future work. Some of the functions in the testcase are not inlined yet with this:
https://issues.dlang.org/show_bug.cgi?id=9646
functions `_D3std5range10primitives12__T5emptyTaZ5emptyFNaNbNdNiNfxAaZb` and `_D3std9algorithm9searching34__T4findVAyaa6_61203d3d2062TAhTAhZ4findFNaNbNiNfAhAhZAh`.

JohanEngelen:
The AppVeyor -O3 bug appears to be an optimization bug with LLVM3.9 (can reproduce on Mac, but not with LLVM3.8).

JohanEngelen:
For future work: nested functions are not chain-inlined.
```d
pragma(inline, true) bool addAndCheckInsideNestedFunc(int checkbefore, int increment)
{
    pragma(inline, true)
    bool addCheckNested(int checkbefore, int increment)
    {
        return true;
    }

    return addCheckNested(checkbefore, increment);
}
```

JohanEngelen:
The only remaining problem is a linking problem: for LLVM < 3.7 we don't emit symbols in COMDAT any groups. So for some symbols there are multiple definitions link errors (e.g. initializers of structs nested inside an inlined function). Now I can either complicate the LDC logic. Or simple disable cross-module inlining for LLVM < 3.7. Strongly levitating toward the latter...

ddmd/globals.d
```diff
@@ -246,6 +246,8 @@ struct Global
         const(char)* s_ext;
         const(char)* ldc_version;
         const(char)* llvm_version;
+
+        bool gaggedForInlining; // Set for functionSemantic3 for external inlining canditates
```
kinke: Just caught my eye, no time for a proper review: candi**d**ates

JohanEngelen:
With disabling for LLVM < 3.7, it is all green now. 
Inlining heuristic + cleanup and done!

JohanEngelen:
The compile slowdown may not be too terrible. LLVM will do its optimizations first, and non-inlined functions are discarded before machine codegen. So I think we can live with a superdumb inlining heuristic, and leave the good stuff up to LLVM? I've implemented a simple statement counting heuristic now.

ddmd/globals.d
JohanEngelen: Thanks, fixed :)

JohanEngelen:
no more "WIP" ! :)

tests/codegen/inlining_staticvar.d
```diff
@@ -0,0 +1,86 @@
...
+    checkModuleScope_1();
+    checkModuleScope_2();
+    checkInsideFunc_1();
+    checkInsideFunc_2();
+    checkInsideNestedFunc_1();
+    checkInsideNestedFunc_2();
+    checkNestedStruct_1();
+    checkNestedStruct_2();
+}
+
```
kinke: Too many trailing newlines...

JohanEngelen: :/  fixed!

gen/declarations.cpp
```diff
@@ -430,10 +433,19 @@ class CodegenVisitor : public Visitor {
       return;
     }
 
-    // FIXME: This is #673 all over again.
-    if (!decl->needsCodegen()) {
-      Logger::println("Does not need codegen, skipping.");
-      return;
+    // Force codegen if this is a templated function with pragma(inline, true).
+    if ((decl->members->dim == 1) &&
```
klickverbot: Did you ever get around to looking into using `onemember`?

JohanEngelen: No I haven't yet. Thanks for the reminder.

gen/declarations.cpp
```diff
@@ -430,10 +433,19 @@ class CodegenVisitor : public Visitor {
...
+        ((*decl->members)[0]->isFuncDeclaration()) &&
+        ((*decl->members)[0]->isFuncDeclaration()->inlining == PINLINEalways)) {
+      Logger::println("needsCodegen() == false, but function is marked with "
+                      "pragma(inline, true), so it really does need "
+                      "codegen.");
+    } else {
+      // FIXME: This is #673 all over again.
```
klickverbot: Your work should actually fix #673/#674 for good. ;)

gen/function-inlining.cpp
```diff
@@ -0,0 +1,196 @@
...
+  if (fdecl.isFuncAliasDeclaration()) {
+    IF_LOG Logger::println("isFuncAliasDeclaration() == true");
+    return false;
+  }
+  if (!fdecl.fbody) {
+    IF_LOG Logger::println("No function body available for inlining");
+    return false;
+  }
+
+  // TODO: Fix inlining functions from object.d. Currently errors because of
```
klickverbot: Could you add an issue for this on merging, please?

gen/function-inlining.cpp
```diff
@@ -0,0 +1,196 @@
...
+  if (fdecl.getModule()->ident == Id::object) {
+    IF_LOG Logger::println("Inlining of object.d functions is disabled");
+    return false;
+  }
+
+  if (fdecl.semanticRun >= PASSsemantic3) {
+    // If semantic analysis has come this far, the function will be defined
+    // elsewhere and should not get the available_externally attribute from
+    // here.
+    // TODO: This check prevents inlining of nested functions.
```
klickverbot: Why do we need to be cautious here then in the first place? Shouldn't the caller from "elsewhere" take the path that resets the linkage to the proper one anyway, making it immaterial where the code has been emitted?

ddmd/func.d
```diff
@@ -2484,8 +2484,16 @@ else
             TemplateInstance spec = isSpeculative();
             uint olderrs = global.errors;
             uint oldgag = global.gag;
+version (IN_LLVM)
+{
+            if (global.gag && !spec && !global.gaggedForInlining)
```
klickverbot: I think it would be beneficial to discuss whether this is the "proper" way of doing it with upstream, although I understand if you don't want to spend the energy on it.

gen/function-inlining.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include "gen/optimizer.h"
+#include "gen/recursivevisitor.h"
+#include "gen/uda.h"
+
+namespace {
+
+/// An ASTVisitor that checks whether the number of statements is larger than a
+/// certain number.
+struct MoreThanXStatements : public StoppableVisitor {
+  /// Are there more or fewer statements than `threshold`?.
```
klickverbot: This looks like it should be the comment for a boolean result, not `threshold`. ;)

gen/function-inlining.cpp
JohanEngelen: I'll make an issue about it upon merging. I had some compile errors without this check before. But now some stuff has changed so I'm not sure it is needed any more.

JohanEngelen:
> Random side note: It might be interesting to add diagnostic counters

I've been thinking the same. Clang/LLVM's code seems to use an interface for reporting internal compiler diagnostics. Would be great to have something similar in LDC (also for e.g. PGO things).

gen/function-inlining.cpp
JohanEngelen: I'll add several issues regarding inlining. The forward-defined druntime functions are also not available for inlining.

JohanEngelen:
Great!

kinke:
Nice work, thanks! I fast-forwarded the `ldc` Phobos branch accordingly.

> AppVeyor actually ran is green. A unique opportunity.

;) AppVeyor is currently the only fully working CI system, the only green spark in a sea of red. ;) It does take a painful lot of time though.

gen/declarations.cpp
JohanEngelen: Having looked more closely at it now, it is not completely clear to me how to work with the `onemember` functions and the AST :S. It'd be good to have a testcase where something that should be inlined isn't.
In some parts of DMD I see that instead `decl.aliasdecl` is checked to see if it is a template with eponymous function... I'm confused. Testcase needed! ;)

JohanEngelen:
Checked just now: llvm.org/apt shows that the APT repo is now up-to-date with trunk head again, so CircleCI should be green soon (because it is at roughly the same version as AppVeyor).
At the moment it fails because a slightly more recent LLVM than AppVeyor has an API change. I already mailed the person working on it to restore the old default behavior (i.e., no need for us to make code modifications). Hopefully done in a few days, then CircleCI should be green too. Fingers crossed!

gen/declarations.cpp
klickverbot: I might have misremembered how `onemember` works myself, then.

9il:
Thanks!

PR #1578, opened 2016-06-22
[profile-rt] Add check for defining COMPILER_RT_HAS_FCNTL_LCK in CMake script.

JohanEngelen:
I just discovered that PGO in LLVM3.9 has yet again become more awesome: automatic merging of profile data across multiple runs of the same executable.

The define checked for in this PR is for file-locking during the awesome on-the-fly profile merging feature using `%m` in `-fprofile-instr-generate=abc_%m_xyz` (for example).

### DIFF
--- a/runtime/profile-rt/DefineBuildProfileRT.cmake
+++ b/runtime/profile-rt/DefineBuildProfileRT.cmake
@@ -44,6 +44,24 @@ if (LDC_WITH_PGO)
      set(PROFRT_EXTRA_FLAGS "${PROFRT_EXTRA_FLAGS} -DCOMPILER_RT_HAS_ATOMICS=1")
     endif()
 
+    CHECK_CXX_SOURCE_COMPILES("
+    #if defined(__linux__)
+    #include <unistd.h>
+    #endif
+    #include <fcntl.h>
+    int fd;
+    int main() {
+     struct flock s_flock;
+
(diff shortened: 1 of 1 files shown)


JohanEngelen:
(will add a small testcase that tests profile merging)

JohanEngelen:
The test below should work, but because ldc-profdata does not accept wildcard in the input files without shell support, it doesn't work. It's more of a profile-rt test anyway, so skipping adding the test. I've put it here for reference.

```D
// Tests automatic profile merging through the %m specifier in the profile file

// REQUIRES: atleast_llvm309

// RUN: %ldc -fprofile-instr-generate=%t.%m.profraw %s -of=%t%exe  \
// RUN:   &&  %t%exe \
// RUN:   &&  %t%exe \
// RUN:   &&  %t%exe \
// RUN:   &&  %profdata merge -o %t.profdata ""%t.*.profraw"" \
// RUN:   &&  %ldc -c -output-ll -of=%t2.ll -fprofile-instr-use=%t.profdata %s \
// RUN:   &&  FileCheck %s < %t2.ll

// CHECK: !{{[0-9]+}} = {{.*}}!"MaxFunctionCount", i{{(32|64)}} 6}

void foo() {}
void bar() {}

void main() {
  foo();
  bar();
  bar();
}
```

Issue #1581, opened 2016-06-24
Enums with referenced struct members result in floating point error


Philpax:
Minimised test case:
```D
struct OperandFormatDescriptor
{
    string name;
}

enum OperandFormat
{
    DstSrc = OperandFormatDescriptor("DstSrc"),
}

struct OpcodeDescriptor
{
    OperandFormat operandFormat;
}

enum Opcodes
{
    // Memory
    Load = OpcodeDescriptor(OperandFormat.DstSrc),
}

void main() {}
```

results in
`test.d(8): Error: floating point constant expression expected instead of OperandFormatDescriptor("DstSrc")`

when compiled with LDC 1.0.0. This works fine with DMD 2.069 and 2.071 (the two versions of DMD I tested).

kinke:
Thanks for the report!

PR #1585, opened 2016-06-25
Fix issue #1581

kinke:
There was an ICE when building an enum's initializer for its TypeInfo if the enum's base type was a struct.


kinke:
I'm not really sure about this, but it seems to work. ;)

kinke:
Seems so, dmd-testsuite-debug passes on my box. Now using `toConstElem()` for all non-zero initializers.

Issue #1548, opened 2016-06-09
Static array initialization with single element misdetected as NRVO case


puffi:
The following code results in "Call parameter type does not match function signature!" error when building with ldc2 on Arch Linux.
```
LDC - the LLVM D compiler (1.0.0):
  based on DMD v2.070.2 and LLVM 3.8.0
  built with DMD64 D Compiler v2.071.0
```
No flags passed to ldc2, only the source file.

```
void main()
{
	mat4[1024] transforms = mat4.identity;
}

struct mat4
{
	float[16] data;
	
	static @property auto identity()
	{
		return mat4(
			[1, 0, 0, 0,
			 0, 1, 0, 0,
			 0, 0, 1, 0,
			 0, 0, 0, 1]
		);
	}
}
```

The whole error message:
```
Call parameter type does not match function signature!
  %transforms = alloca [1024 x %app.mat4], align 4
 %app.mat4*  call void @_D3app4mat48identityFNaNbNdNiNfZS3app4mat4([1024 x %app.mat4]* noalias sret align 4 %transforms)
LLVM ERROR: Broken function found, compilation aborted!
```

redstar:
Confirmed. Thanks for the nice test case.

klickverbot:
The fix should probably go into a point release, if we are doing one (which I think would be a good idea for all the build system issues).

klickverbot:
Fixed by #1586, pending addition of test case.

kinke:
Original testcase added by https://github.com/ldc-developers/dmd-testsuite/commit/30a2ffab455dee013b9076ddb7b5f4ba7f5e821d.

PR #1586, opened 2016-06-25
Tighten NRVO detection for VarDeclarations

kinke:
The D types of the variable and the right-hand-side of the init `AssignExp` must match; this fixes issue #1548.
Don't restrict it to memory-only types anymore though; the sret-check performed by `DtoIsReturnInArg()` should suffice and makes more sense to me.

A test will follow once merged.


kinke:
Thinking about it, this has actually nothing to do with NRVO at all. It's just a direct construction of a variable into the hidden `sret` argument of the rhs call. This should probably all be handled in `ToElemVisitor::visit(AssignExp*)`.

PR #1588, opened 2016-06-26
AssignExp: Refactor direct construction by rhs call via sret

kinke:
A follow-up for https://github.com/ldc-developers/ldc/pull/1586.


kinke:
Yep, one more biggie to be split up when there's time... ;)

PR #1589, opened 2016-06-27
Remove explicit `noinline` call attribute inside SEH catch/cleanup pads.

JohanEngelen:
See https://llvm.org/bugs/show_bug.cgi?id=25162
However the LLVM inliner now knows not to inline functions inside cleanup pads and so this is no longer needed. I have tested with clang (trunk) and clang also no longer marks function calls explicitly with noinline: the inliner no longer tries to inline `alwaysinline` functions.
http://llvm.org/viewvc/llvm-project/cfe/trunk/test/CodeGenCXX/microsoft-abi-eh-cleanups.cpp?r1=251576&r2=255423&diff_format=h

The change is necessary because setting `noinline` conflicts with functions that have `alwaysinline` attribute added (due to `pragma(inline, true)`).

@rainers Can you confirm this is correct? Thanks.

(this PR is needed for https://github.com/ldc-developers/ldc/pull/1577)

### DIFF
--- a/gen/ms-cxx-helper.cpp
+++ b/gen/ms-cxx-helper.cpp
@@ -91,7 +91,6 @@ void remapBlocksValue(std::vector<llvm::BasicBlock *> &blocks,
 // - redirect srcTarget to continueWith
 // - set "funclet" attribute inside catch/cleanup pads
 // - inside funclets, replace "unreachable" with "branch cleanupret"
-// - disable inlining inside a funclet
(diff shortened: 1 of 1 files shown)


rainers:
Appveyor suggests there isn't a problem anymore, so I'm fine with this, too. You could also remove the intermediate `invoke` variables. Feel free to merge with or without that change.

JohanEngelen:
Thanks Rainer. I'm not sure exactly what you mean about the intermediate variables, could you... ah, now I think I do know what you mean... sorry.

PR #1591, opened 2016-06-27
optimizer: Skip adding verifier function pass if -disable-verify is given

klickverbot:


joakim-noah:
LGTM

kinke:
Great, I didn't know there was a way to get hold of faulty IR. ;)

JohanEngelen:
`-print-before-all` is good too.

Issue #1590, opened 2016-06-27
Regression: recent PR in master breaks variadics on Android/ARM


joakim-noah:
If I revert #1533, `runnable/variadic.d` from the dmd-testsuite works fine again.  I can't reproduce on linux/x64, so this is related to the ARM ABI and how variadics are done there.  LLVM's IR verifier flags some instructions in the `copya` function from `runnable/variadic.d`, with the following error output:
```
Error: Load operand must be a pointer.
%7 = load i8*, %ldc.internal.vararg.std.__va_list %6
Stored value type does not match pointer operand type!
store i8* %7, %ldc.internal.vararg.std.__va_list* %ap2
```
I had to turn off basic IR verification [to dump this faulty IR output](https://github.com/ldc-developers/ldc/files/334733/var.ll.txt), see lines 3962 and 3963.

kinke:
Please try to produce a minimal testcase - I guess something like

```D
void foo(...) { auto ap = _argptr; }
void main() {}
```

should do the trick.

JohanEngelen:
@joakim-noah What is the ARM triple that you are using? (it's nice to know for testing)

joakim-noah:
It's listed at the top of the IR file I linked, `armv7-none-linux-android`.  I can provide a build of ldc master for that target, for poking at this bug: would that be helpful?

JohanEngelen:
No special build needed for me: `ldc2 -mtriple=armv7-none-linux-android -c -output-ll ../tests/d2/dmd-testsuite/runnable/variadic.d` reproduces it.

What I was thinking: should we set up an automated tester that cross-compiles druntime/phobos unittests (without running them) to catch this kind of problem? Or is this a rare kind of bug?

joakim-noah:
I think it's uncommon.  I need to look at using the Android/ARM emulator on Travis to run all the tests, that's a better route.

kinke:
This crashes too, this time due to getRVal() for a memory-only type (apparently ARM uses a struct with special mangling containing a pointer):

```
import core.vararg;

void foo(...)
{
    va_list ap;
    va_copy(ap, _argptr);
}
```

I'll have a look and may need to switch to DValue for `TargetABI::vaCopy()` and friends.

kinke:
Looks like the ARM TargetABI relied on the `va_list` DValues (structs, otherwise mostly naked pointers for other platforms) returning their address in (old) `DValue::getRVal()`.

kinke:
[...]

PR #1592, opened 2016-06-27
Refactor and fix intrinsics for variadic functions

kinke:
I.e., `va_start()`, `va_copy()` and `va_arg()`.

### DIFF
--- a/gen/abi-arm.cpp
+++ b/gen/abi-arm.cpp
@@ -119,12 +119,6 @@ struct ArmTargetABI : TargetABI {
     }
   }
 
-  void vaCopy(LLValue *pDest, LLValue *src) override {
-    // simply bitcopy src over dest.  src is __va_list*, so need load
-    auto srcval = DtoLoad(src);
-    DtoStore(srcval, pDest);
-  }
-
   Type *vaListType() override {
     // We need to pass the actual va_list type for correct mangling. Simply
     // using TypeIdentifier here is a bit wonky but works, as long as the name
--- a/gen/tocall.cpp
+++ b/gen/tocall.cpp
@@ -264,17 +264,16 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
       e->error("va_start instruction expects 1 (or 2) arguments");
       fatal();
     }
-    LLValue *pAp = DtoLVal((*e->arguments)[0]); // va_list*
+    DLValue *ap = toElem((*e->arguments)[0])->isLVal(); // va_list
+    assert(ap);
     // variadic extern(D) function with implicit _argptr?
-    if (LLValue *pArgptr = p->func()->_argptr) {
-      DtoMemCpy(pAp, pArgptr); // ap = _argptr
-      result = new DImValue(e->type, pAp);
+    if (LLValue *argptrMem = p->func()->_argptr) {
+      DtoMemCpy(DtoLVal(ap), argptrMem); // ap = _argptr
     } else {
-      LLValue *vaStartArg = gABI->prepareVaStart(pAp);
-      result =
-          new DImValue(e->type, p->ir->CreateCall(GET_INTRINSIC_DECL(vastart),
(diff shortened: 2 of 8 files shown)


kinke:
This lil' druntime fix goes with it: https://github.com/ldc-developers/druntime/commit/de7ca661326cf927075651ab107a55de9c63a2ab

gen/abi-arm.cpp
```diff
@@ -119,12 +119,6 @@ struct ArmTargetABI : TargetABI {
     }
   }
 
-  void vaCopy(LLValue *pDest, LLValue *src) override {
-    // simply bitcopy src over dest.  src is __va_list*, so need load
-    auto srcval = DtoLoad(src);
-    DtoStore(srcval, pDest);
```
kinke: `src` was `__va_list*` because the old `DValue::getRVal()` would return the lvalue for structs. New `DtoRVal()` refuses to load structs from memory and bails out.
The new default implementation performs a memcpy if `src` is an lvalue and a store if it's an rvalue.

gen/tocall.cpp
```diff
@@ -264,17 +264,16 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
...
+      DtoMemCpy(DtoLVal(ap), argptrMem); // ap = _argptr
     } else {
-      LLValue *vaStartArg = gABI->prepareVaStart(pAp);
-      result =
-          new DImValue(e->type, p->ir->CreateCall(GET_INTRINSIC_DECL(vastart),
-                                                  vaStartArg, ""));
+      LLValue *llAp = gABI->prepareVaStart(ap);
+      p->ir->CreateCall(GET_INTRINSIC_DECL(vastart), llAp, "");
     }
+    result = new DVoidValue();
```
kinke: `va_start` and `va_copy` return `void`. Our previous resulting DValues were... well, at least pointless.

gen/tocall.cpp
```diff
@@ -297,13 +297,15 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
...
     }
-    LLValue *pAp = DtoLVal((*e->arguments)[0]); // va_list*
-    LLValue *vaArgArg = gABI->prepareVaArg(pAp);
-    LLType *llType = DtoType(e->type);
     if (DtoIsInMemoryOnly(e->type)) {
-      llType = getPtrToType(llType);
+      e->error("va_arg instruction does not support structs and static arrays");
+      fatal();
     }
-    result = new DImValue(e->type, p->ir->CreateVAArg(vaArgArg, llType));
```
kinke: This would have never worked for memory-only types anyway (cannot be handled by any DRValues), so bail out instead. Last time I checked (~2 years ago or so) the LLVM stance was that the va_arg intrinsic won't support complex types. But we have runtime functions and afaik don't use this LDC intrinsic at all.

joakim-noah:
This PR and the linked druntime commit fix #1590: the entire test suite passes natively on Android/ARM with this patch applied.  I don't know enough about ldc and variadics to review it, however, and frankly am not interested in digging into this stuff.  Thanks for fixing the issue, the PR works for me.

@smolt, got an opinion?

smolt:
I don't have cycles to review, but if it works for @joakim-noah, then ARM is probably fine.  AArch64 change looks more extensive, @redstar might want to look at it.  I don't know how much of AArch64 he got working.

kinke:
The AArch64 variadic stuff is just a copy of the x86_64 implementation, so I wouldn't expect any surprises there (but IIRC `core.stdc.stdarg.va_arg()` isn't implemented in druntime). Thanks @joakim-noah for testing.

kinke:
> The DVoidValues seem rather out of place, though – if the intrinsics return void, why do we need a value at all?

That's exactly why I waited for your feedback. :) I wasn't sure `toElem()` is allowed to return null.

kinke:
`DVoidValue` is no more. Awaiting some green.

PR #1593, opened 2016-06-28
use *W-functions if UNICODE #defined on windows

rainers:
Recent LLVM master switched to enable UNICODE on Windows, breaking some code in LDC.

### DIFF
--- a/driver/linker.cpp
+++ b/driver/linker.cpp
@@ -28,6 +28,7 @@
 #include "llvm/Support/SourceMgr.h"
 #if _WIN32
 #include "llvm/Support/SystemUtils.h"
+#include "llvm/Support/ConvertUTF.h"
 #include <Windows.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


driver/linker.cpp
```diff
@@ -426,9 +427,17 @@ int executeAndWait(const char *commandLine) {
 
   DWORD exitCode;
 
+#if UNICODE
+  std::wstring wcommandLine;
+  if (!llvm::ConvertUTF8toWide(commandLine, wcommandLine))
+    return -3;
```
kinke: I assume `-3` is purely deliberate?

kinke:
Lgtm. I'm sure you've also tested it. Could you please post the git/svn commit you used (if it works)?

rainers:
Yeah, I've tried it with LLVM svn@273940.

driver/linker.cpp
rainers: Yes, just as much as the other return codes -2 and -1.

rainers:
BTW: It could be possible now to allow unicode names in installation paths and filenames passed to LDC. One impediment is libconfig which seems to just accept filenames in the current codepage. Should we try to replace it?

JohanEngelen:
I am definitely in favor of replacing libconfig with something in tree (preferably in D of course! ;)), because it would remove an external dependency.

Issue #1544, opened 2016-06-06
Can't build 1.0.0 with cmake


Dicebot:
Trying to use my usual build script which looks like this:

```
    cd $srcdir/ldc

    git submodule update --init --recursive

    mkdir -p build && cd build
    cmake \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_SKIP_RPATH=ON \
    -DINCLUDE_INSTALL_DIR=/usr/include/dlang/ldc \
    -DBUILD_SHARED_LIBS=ON \
    ..
    make -j 4
```

LDC does build bunch of files just fine but whenever it gets to linking, this error shows up:

```
/usr/bin/ld: unrecognized option '-Wl,-O1,--sort-common,--as-needed,-z,relro'
```

Arch Linux x86_64 (ld 2.26.0 , cmake 3.5.2)

yannick:
hmm, for me it works, at least for the static PKGBUILD i have and also for ldc-git

JohanEngelen:
> https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ldc-git

Perhaps with `-DCMAKE_EXE_LINKER_FLAGS='-static-libstdc++ -Wl,-rpath,\$ORIGIN' \` you override the default linker flags that contain stuff that our build script / the D compiler doesn't like?

Kozzi11:
Problem is with Archlinux /etc/makepkg.conf it defines:

    LDFLAGS="-Wl,O1,--sort-common,--as-needed,-z,relro"

which does not work with ldmd.

One workaround is to use this in PKGBUILD before cmake

    export LDFLAGS="-L-O1 -L--sort-common -L--as-needed -L-z -Lrelro"

Kozzi11:
But you still need to uninstall ldc first and use dmd to build ldc 1.0.0. Because there is a some ABI issue with ldc 0.17.1

wilzbach:
Related to #1530

klickverbot:
We could of course add another special case to the CMake script that properly splits up such flags, but what we should really do is to add something like a `-linker-driver-arg=…` command line switch similar to `-L` that does not prepend `-Xlinker`. We can just forward the flags from llvm-config, pkg-config, CMAKE_EXE_LINKER_FLAGS (which is the issue here) and so on then. Everything else seems very brittle.

Dicebot:
[...]

PR #1594, opened 2016-06-28
cmake: Directly use (g)cc for linking on Unix-like systems

### LINKED ISSUE #1494: unrecognized option '-Wl,--as-needed'
On openSUSE Leap 42.1, building `ldc2` and `ldmd2` now fails with:
`/usr/bin/ld: unrecognized option '-Wl,--as-needed'`

The most likely culprit is https://github.com/ldc-developers/ldc/pull/1489, though I haven't verified that.

klickverbot:
Currently, we are translating the linker flags for LLVM/libconfig and from `CMAKE_EXE_LINKER_FLAGS` into `-L` options. Since DMD/LDC prefix `-L` options by `-Xlinker` before passing them to the GCC driver (causing the latter to forward it directly to the underlying `ld`), this does not work in the general case, causing issues such as #1494.

In https://github.com/ldc-developers/ldc/pull/1468, a few driver-only flags were special-cased in LDC so that `-Xlinker` is not applied to them. However, I don't think this is a viable strategy in the long run. First of all, there are quite a few more options which are supposed to be passed to the linker driver instead of the linker. But rather more importantly, this only works with an LDC version that contains this fix, and not DMD.

Thus, we need a solution that works on the build-system level. One possibility would be to white-list flags we know how to handle in the CMake scripts (e.g. all of them starting with `-Xlinker` or `-Wl,…`, where we can just strip off the prefix and pass them to the D compiler using `-L`), and just ignore all the others. However, for some common use cases, we do want to pass flags to the linker driver, such as to link against a static C++ standard library when building binary releases.

Since there is no DMD flag for passing flags to the linker driver (we could of course add one in the next LDC version, but this wouldn't help with DMD or older versions), this PR changes the build system to again use the system linker directly from CMake instead of going through the D compiler for linking. Compared to @JohanEngelen's earlier version (sorry!), this now parses the `-v` output to determine the default linker flags used internally by the D compiler instead of trying to guess them.

Fixes #1494, #1544.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -521,38 +521,76 @@ set(LDC_EXE_FULL ${PROJECT_BINARY_DIR}/bin/${LDC_EXE_NAME}${CMAKE_EXECUTABLE_SUF
 set(LDMD_EXE_FULL ${PROJECT_BINARY_DIR}/bin/${LDMD_EXE_NAME}${CMAKE_EXECUTABLE_SUFFIX})
 add_custom_target(${LDC_EXE} ALL DEPENDS ${LDC_EXE_FULL})
 add_custom_target(${LDMD_EXE} ALL DEPENDS ${LDMD_EXE_FULL})
+
+# Figure out how to link the main LDC executable, for which we need to take the
+# libconfig/LLVM flags into account.
 set(LDC_LINKERFLAG_LIST "${SANITIZE_LDFLAGS};${LIBCONFIG_LIBRARY};${LLVM_LIBRARIES};${LLVM_LDFLAGS}")
-set(tempVar "")
-foreach(f ${LDC_LINKERFLAG_LIST})
-    string (REPLACE "-Wl," "" f ${f})
-    string (REPLACE "-LIBPATH:" "/LIBPATH:" f ${f})
-    append("-L\"${f}\"" tempVar)
-endforeach(f)
-if(MSVC)
-    # Issue 1297
-    # The default system-allocated stack size is 8MB on Linux and Mac, but only 1MB on Windows
-    # Set LDC's stack to 8MB also on Windows:
-    append("-L/STACK:8388608" tempVar)
+
+set(LDC_LINK_MANUALLY OFF)
+if(UNIX AND (CMAKE_COMPILER_IS_GNUCXX OR (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")))
+    # On Unix-like systems, DMD and LDC will use the C compiler for linking, but
+    # will pass on -L options prefixed by -Xlinker to directly forward them to
+    # the underlying ld. Since there are some flags the GCC driver handles itself
+    # rather than passing them to ld, we cannot just directly translate
+    # LDC_LINKERFLAG_LIST to -L options. To be able to handle general linker flags,
+    # we manually invoke the linker instead of using the D compiler to do so.
+    set(LDC_LINK_MANUALLY ON)
+
+    include(ExtractDMDSystemLinker)
+    list(APPEND LDC_LINKERFLAG_LIST ${D_LINKER_ARGS})
+
+    if(NOT "${CMAKE_EXE_LINKER_FLAGS}" STREQUAL "")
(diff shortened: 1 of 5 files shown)


JohanEngelen:
This could actually speed-up re-builds quite a bit when the D-source hasn't changed! :-)

kinke:
> a bit of a pain

That's an understatement. :P

kinke:
Wrt. the AppVeyor failure:

```
c:\projects\ldc2-1.0.0-win64-msvc\bin\ldmd2.exe "-wi -g -O -inline -release -JC:/projects/ldc/ddmd -IC:/projects/ldc -IC:/projects/ninja-ldc -version=IN_LLVM_MSVC -version=IN_LLVM" "-L/NODEFAULTLIB:libcmt -L/NODEFAULTLIB:libvcruntime" ...
```

I guess the problem are the quotes surrounding `-wi ...`.

kinke:
@klickverbot: This works for me:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 711e56c..54f3ed5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -581,14 +581,21 @@ function(build_d_executable output_exe compiler_args linker_args dependencies)
             DEPENDS ${object_file}
         )
     else()
+        set(dflags "${D_COMPILER_FLAGS} ${DDMD_DFLAGS} ${DDMD_LFLAGS}")
         set(lflags ${LDC_TRANSLATED_LINKER_FLAGS})
-        separate_arguments(lflags)
+        if(UNIX)
+          separate_arguments(dflags UNIX_COMMAND "${dflags}")
+          separate_arguments(lflags UNIX_COMMAND "${lflags}")
+        else()
+          separate_arguments(dflags WINDOWS_COMMAND "${dflags}")
+          separate_arguments(lflags WINDOWS_COMMAND "${lflags}")
+        endif()
         foreach(f ${linker_args})
             append("-L\"${f}\"" lflags)
         endforeach()
         add_custom_command(
             OUTPUT ${output_exe}
-            COMMAND ${D_COMPILER} ${D_COMPILER_FLAGS} ${DDMD_DFLAGS} ${DDMD_LFLAGS} ${lflags} -I${PROJECT_SOURCE_DIR}/$+            COMMAND ${D_COMPILER} ${dflags} ${lflags} -I${PROJECT_SOURCE_DIR}/${DDMDFE_PATH} -of${output_exe} ${compile             WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
             DEPENDS ${dependencies}
         )
```

kinke:
We need this in 1.1 too. ;)

kinke:
Interesting, LDC 1.1 as front-end compiler now doesn't work anymore for Travis due to multiple definitions of `pragma(LDC_extern_weak) void* _d_execBssBegAddr`...

JohanEngelen:
@kinke Is it a built with `-O3`? Cross-module inlining gone bad can result in multiple definition problems maybe. What happens when you build with `-disable-cross-module-inlining` ?

kinke:
@JohanEngelen: I was just trying to summarize the Travis logs. The only difference wrt. the previous successful Travis runs (before the latest commit) is/should be that the arguments in D_COMPILER_FLAGS, DDMD_DFLAGS and DDMD_LFLAGS are now separated too (I thought that was unnecessary for Unixoids and only relevant for Windows...). That could be easily checked by diffing the generated makefiles.

I don't get why there's a `ldmd2.o` and a `ldc2.o`. There are no corresponding .cpp modules, so they seem to be dummy object files for the 2 executables (with that name)...

kinke:
Thanks, so how did you just fix it? :)

JohanEngelen:
Looks good to me.

Things for future work:
- source dependencies are a little off somehow: if I change a .cpp file, the D source is still recompiled too. 
- compile the D source per file, instead of all at once (with cross-module inlining in 1.1.0, the built compiler performance should be the same; we must not forget to check!). That should help the ARM buildbots that timeout on the current long compile time.

JohanEngelen:
(it works without issue on my machine too btw)

Issue #1450, opened 2016-04-19
super doesn't work


Temtaime:
Hi !
I'm able to build ldc on windows with dmd and it's able to build itself then.

Using vs2015 and master ldc with trunk llvm:

```
module b;

import
		std.path,
		std.file,
		std.mmfile,
		std.typecons;

class FileSystem {
	FsData dataOf(string name) {
		FsData res;
		return res;
	}
}

struct FsData {
	MmFile file;
	const(void)[] data;
}

```

```
module a;
import b;

class RoFileSystem : FileSystem {

	override FsData dataOf(string name) {
		auto r = super.dataOf(name);
		return r;
	}
}

void main() {
	(new RoFileSystem).dataOf("");
}
```

It craches because super.dataOf calls RoFileSystem.dataOf, not FileSystem.dataOf

JohanEngelen:
This works with LDC master and DMD v2.070.2 :

```D
char[] a = [];

class A {
    void foo() {
        a ~= 'A';
    }
}

class B : A {
    override void foo() {
        a ~= 'B';
        super.foo();
    }
}

void main() {
    (new B).foo();
    assert(a == ['B', 'A']);
}
```

Temtaime:
This works with me too.
And my example works too if i change 

```
  auto r = super.dataOf(name);
  return r;
```

to

`return super.dataOf(name);`

Seems to be heisenbug

JohanEngelen:
Reduced testcase:
```D
struct AA{}
version(bad)
{
    alias A = AA;
}
else
{
    alias A = int;
}

class B
{
    A foo()
    {
        A res;
        return res;
    }
}

class C : B
{
    override A foo()
    {
        auto r = super.foo();
        return r;
    }
}

void main()
{
    (new C).foo();
}
```

Temtaime:
It's strange but reduced test case doesn't crash on my setup.

JohanEngelen:
(you need to specify `-d-version=bad` to make it crash)

Thanks for finding out that the return statement change "fixes" things. I think that is a good lead to where to fix the bug!
```D
struct A {}

class B {
    A foo() {
        A res;
        return res;
    }
}

class C : B {
    override A foo() {
        version (bad)
        {
            auto r = super.foo();
            return r;
        }
        else
        {
            return super.foo();
        }
    }
}

void main() {
    (new C).foo();
}
```

Temtaime:
Yes i know and tried both version bad and without it.
De rien.

[...]

PR #1596, opened 2016-06-29
Resolve virtual base functions (super.foo()) correctly

kinke:
Don't perform a vtable lookup for those, resolve them directly instead.
Fixes issue #1450, where calling the base method in the overriding method resulted in an infinite loop and corresponding stack overflow.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -1276,7 +1276,7 @@ public:
 
       // Get the actual function value to call.
       LLValue *funcval = nullptr;
-      if (nonFinal) {
+      if (e->e1->op != TOKsuper && nonFinal) {
         funcval = DtoVirtualFunctionPointer(l, fdecl, e->toChars());
       } else {
         funcval = getIrFunc(fdecl)->func;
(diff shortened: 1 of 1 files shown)


kinke:
We have the [same check for DelegateExp](https://github.com/ldc-developers/ldc/blob/master/gen/toir.cpp#L2248), the only other place where we call `DtoVirtualFunctionPointer()`.

kinke:
Updated as I figured the actual reason. not just the symptom. This was just begging for more issues.

kinke:
I'm planning to extend the direct-sret-construction for return statements too. Currently, presumably front-end NRVO produces ideal IR while returning a return value directly (e.g., `return super.foo()`) performs an obsolete copy (possibly even a postblit call).

JohanEngelen:
> Test will follow once merged. ;)

For dmd-testsuite additions, I understand. But for IR tests, could you add them to the PR? (or perhaps it doesn't involve IR tests yet)

kinke:
There's no need to check the IR for this. It's actually quite hard to test this (properly visiting the CallExp); the testcase in #1450 can test one (current) symptom if it isn't visited.

kinke:
Test: https://github.com/ldc-developers/dmd-testsuite/commit/2fb4d566580e6d096a58ca00e37293ccb9e59205

PR #1597, opened 2016-06-30
(release-1.0.1) Debug info generation fixes for LLVM >= 3.8

Syniurge:
Debug info generation is broken when LDC is built against LLVM 3.8:

```
(gdb) break test.d:20  <= works
(gdb) run
(gdb) bt  <= no location info
#0  0x0000000000402f49 in D main ()
(gdb) info locals
No symbol table info available.
```

A `llvm::Function::setSubprogram()` call to assign the descriptor to the function was missing. Restoring it resulted in new compilation errors while building druntime and phobos, which were due to function parameters accessed from a nested function being declared as function parameters (they're not, they're fields of `nest_arg`) instead of "auto variables", and setting two variables for the same parameter index was triggering an assert:

```
ldc2: DwarfDebug.h:125: void llvm::DbgVariable::addMMIEntry(const llvm::DbgVariable&): Assertion `V.Var == Var && "conflicting variable"' failed.
```

There's also a fix not specific to LLVM 3.8 for closure variables not getting declared correctly in the parent function.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -900,7 +900,7 @@ void ldc::DIBuilder::EmitValue(llvm::Value *val, VarDeclaration *vd) {
 }
 
 void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
-                                       Type *type, bool isThisPtr,
+                                       Type *type, bool isThisPtr, bool fromNested,
 #if LDC_LLVM_VER >= 306
                                        llvm::ArrayRef<int64_t> addr
 #else
@@ -935,7 +935,7 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
 
 #if LDC_LLVM_VER < 308
   unsigned tag;
-  if (vd->isParameter()) {
+  if (!fromNested && vd->isParameter()) {
     tag = llvm::dwarf::DW_TAG_arg_variable;
   } else {
     tag = llvm::dwarf::DW_TAG_auto_variable;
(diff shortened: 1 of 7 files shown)


redstar:
Thanks! That's a very important PR.

JohanEngelen:
Thanks!

Could you add some IR tests for debug information Syniurg? (I can help you with it if you want)

JohanEngelen:
@Syniurge I don't know if this is possible, but could you retarget this PR to master instead?

Syniurge:
@JohanEngelen does the IR test look ok? not too verbose?

Github doesn't let me change the target branch, should I make another PR?

JohanEngelen:
Great that you made an IR testcase! Perhaps you can put it in a new "debuginfo" directory (or some name like that)? So that we don't put everything in the codegen dir ;)
About the failures on Travis... I was afraid that it would be very dependent on LLVM version :(
You can use e.g. `// REQUIRES: atleast_llvm308` to filter IR testing. I don't know much about the debug info in IR, so can't judge. It looks like you tried to make it flexible in pattern matching, so that's great.

About retargeting to master: it'd be nice to have AppVeyor testing, which is apparently broken on the 1.0.1 branch at the moment. I think I can fix that soon. I don't know how @redstar wants it for release management, but I usually target master for PRs and then release branches can cherry-pick it if need be.

Syniurge:
Here's the IR from LLVM 3.5 to 3.8:
    https://gist.github.com/Syniurge/66f14ec673ffc82c3c3d1e2b55473c4b
    https://gist.github.com/Syniurge/a3629483aa2d390ef985b4358a7fec82
    https://gist.github.com/Syniurge/7fca844f59fb0ced480d5644bef81bde
    https://gist.github.com/Syniurge/a54a7d2a6723dda83e9e7cb05ddd9281

So yep it needs at least 2 more test cases for LLVM 3.5 to 3.7, and let's hope this doesn't fluctuate too much with the next releases :)

JohanEngelen:
@Syniurge Excellent!
(sorry, I meant `tests/debuginfo` instead of `tests/codegen/debuginfo` ... :S )

Can you make a PR for master? I'll pull it, so it can be cherry-picked for release-1.0.1 if we want.
Thanks.

Syniurge:
Replaced by PR #1598.

PR #1598, opened 2016-07-01
Debug info generation fixes for LLVM >= 3.8

Syniurge:
This is PR #1597 recreated for `master`.
*** * * * * * * ***

Debug info generation is broken when LDC is built against LLVM 3.8:

```
(gdb) break test.d:20  <= works
(gdb) run
(gdb) bt  <= no location info
#0  0x0000000000402f49 in D main ()
(gdb) info locals
No symbol table info available.
```

A `llvm::Function::setSubprogram()` call to assign the descriptor to the function was missing. Restoring it resulted in new compilation errors while building druntime and phobos, which were due to function parameters accessed from a nested function being declared as function parameters (they're not, they're fields of `nest_arg`) instead of "auto variables", and setting two variables for the same parameter index was triggering an assert:

```
ldc2: DwarfDebug.h:125: void llvm::DbgVariable::addMMIEntry(const llvm::DbgVariable&): Assertion `V.Var == Var && "conflicting variable"' failed.
```

There's also a fix not specific to LLVM 3.8 for closure variables not getting declared correctly in the parent function, and a small IR test case checking that debug info gets emitted properly for functions and nested variables.

### DIFF
--- a/gen/nested.cpp
+++ b/gen/nested.cpp
@@ -154,7 +154,7 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
   }
 
   if (dwarfValue && global.params.symdebug) {
-    gIR->DBuilder.EmitLocalVariable(dwarfValue, vd, nullptr, false, dwarfAddr);
+    gIR->DBuilder.EmitLocalVariable(dwarfValue, vd, nullptr, false, /*fromNested=*/ true, dwarfAddr);
   }
 
   return makeVarDValue(astype, vd, val);
@@ -508,7 +508,7 @@ void DtoCreateNestedContext(FuncDeclaration *fd) {
         LLSmallVector<LLValue *, 2> addr;
 #endif
         gIR->DBuilder.OpOffset(addr, frameType, irLocal->nestedIndex);
-        gIR->DBuilder.EmitLocalVariable(gep, vd, nullptr, false, addr);
+        gIR->DBuilder.EmitLocalVariable(frame, vd, nullptr, false, false, addr);
       }
     }
   }
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -900,7 +900,7 @@ void ldc::DIBuilder::EmitValue(llvm::Value *val, VarDeclaration *vd) {
 }
 
 void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
-                                       Type *type, bool isThisPtr,
+                                       Type *type, bool isThisPtr, bool fromNested,
 #if LDC_LLVM_VER >= 306
                                        llvm::ArrayRef<int64_t> addr
 #else
@@ -935,7 +935,7 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
 
 #if LDC_LLVM_VER < 308
   unsigned tag;
-  if (vd->isParameter()) {
+  if (!fromNested && vd->isParameter()) {
     tag = llvm::dwarf::DW_TAG_arg_variable;
   } else {
     tag = llvm::dwarf::DW_TAG_auto_variable;
@@ -979,7 +979,7 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
                                                Flags                // flags
                                                );
 #else
-  if (vd->isParameter()) {
+  if (!fromNested && vd->isParameter()) {
     FuncDeclaration *fd = vd->parent->isFuncDeclaration();
     assert(fd);
     size_t argNo = 0;
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -650,7 +650,7 @@ ldc::DISubprogram ldc::DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
     CreateFunctionType(static_cast<TypeFunction *>(fd->type));
 
   // FIXME: duplicates?
-  return DBuilder.createFunction(
+  auto SP = DBuilder.createFunction(
       CU,                                 // context
       fd->toPrettyChars(),                // name
       getIrFunc(fd)->func->getName(),     // linkage name
@@ -667,6 +667,12 @@ ldc::DISubprogram ldc::DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
       getIrFunc(fd)->func
 #endif
       );
+#if LDC_LLVM_VER >= 308
+  if (fd->fbody) {
+    getIrFunc(fd)->func->setSubprogram(SP);
+  }
+#endif
+  return SP;
 }
 
 ldc::DISubprogram ldc::DIBuilder::EmitThunk(llvm::Function *Thunk,
@@ -695,7 +701,7 @@ ldc::DISubprogram ldc::DIBuilder::EmitThunk(llvm::Function *Thunk,
   name.append(".__thunk");
 
   // FIXME: duplicates?
-  return DBuilder.createFunction(
+  auto SP = DBuilder.createFunction(
       CU,                                 // context
       name,                               // name
       Thunk->getName(),                   // linkage name
@@ -712,6 +718,12 @@ ldc::DISubprogram ldc::DIBuilder::EmitThunk(llvm::Function *Thunk,
       getIrFunc(fd)->func
 #endif
       );
+#if LDC_LLVM_VER >= 308
+  if (fd->fbody) {
+    getIrFunc(fd)->func->setSubprogram(SP);
+  }
+#endif
+  return SP;
 }
 
 ldc::DISubprogram ldc::DIBuilder::EmitModuleCTor(llvm::Function *Fn,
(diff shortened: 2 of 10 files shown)


ir/irfunction.h
```diff
@@ -421,15 +421,26 @@ llvm::CallSite ScopeStack::callOrInvoke(llvm::Value *callee, const T &args,
     BundleList.push_back(llvm::OperandBundleDef("funclet", funclet));
 #endif
 
+  llvm::AttributeSet attrs;
+  if (calleeFn) {
+    attrs = calleeFn->getAttributes();
+    if (global.params.symdebug && !gIR->ir->getCurrentDebugLocation()) {
+      // LLVM >=3.9 requires that "inlinable function call in a function with
+      // debug info must have a !dbg location"
```
JohanEngelen: how come there is no debug info at the callsite?

Syniurge: @klickverbot You're right, I saw llvm-mirror/llvm@c5cc3f2a20591c1978136310e453d8f801563900 and thought this was new to LLVM 3.9, but the assert mentioned in the commit message has been around for a long time.

@JohanEngelen I don't really know, and actually `gIR->ir->getCurrentDebugLocation()` never returns false so the noinline attribute doesn't get added. The last commit was a blind attempt at fixing the AppVeyor build, I'll look more into MSVC + LLVM 3.9 today and check how Clang does it, whether its runtime calls always have !dbg info attached or not.

BTW Clang is doing something similar for SEH blocks:
```cpp
  // Disable inlining inside SEH __try blocks.
  if (isSEHTryScope())
    Attrs =
        Attrs.addAttribute(getLLVMContext(), llvm::AttributeSet::FunctionIndex,
                           llvm::Attribute::NoInline);
```

Syniurge:
Oops, sorry for spamming the build queue (can people with write access cancel jobs?)

If the AppVeyor build is green, then the PR's ready.

gen/nested.cpp
```diff
@@ -508,7 +508,7 @@ void DtoCreateNestedContext(FuncDeclaration *fd) {
         LLSmallVector<LLValue *, 2> addr;
 #endif
         gIR->DBuilder.OpOffset(addr, frameType, irLocal->nestedIndex);
-        gIR->DBuilder.EmitLocalVariable(gep, vd, nullptr, false, addr);
+        gIR->DBuilder.EmitLocalVariable(frame, vd, nullptr, false, false, addr);
```
kinke: So you use the frame's base address as (debug info) storage for each captured variable?! How does that work? `gep` seems perfect as it's the address of the corresponding field in the frame.

Syniurge: If `gep` is used with `addr` (which doubles the offset) gdb reports `DWARF-2 expression error: DW_OP_reg operations must be used either alone or (...)`
If `gep` is used alone then I get:
```
(gdb) info locals
normalVar = 1234
closureVar1 = <optimized out>
// closureVar2 doesn't even show up
(gdb) print closureVar2
$1 = <optimized out>
```

Passing `gep` seems to tell gdb to look into a register. I haven't looked deeper than that, so I don't know why passing the alloca instruction with an "offset expression" seems to be the expected way to do things:

http://llvm.org/docs/SourceLevelDebugging.html#llvm-dbg-declare

> The first argument is metadata holding the alloca for the variable.

JohanEngelen:
I have restarted the CircleCI build.
This looks ready for merging?

JohanEngelen:
The CircleCI failure in bitmanip.d looks unrelated to this PR. Looks like a funny artifact of how CircleCI tests PRs (it is not merging with ldc master but with a prior point), and I think the bitmanip bug was fixed already by the inlining+optimization PR.

kinke:
It's definitely mergeable and a nice improvement. I'd just like to know why this offset is required and the GEP alone doesn't work. But I'm currently too lazy to start experimenting in a Linux VM; the .pdb on Windows is still not really functional with a very recent LLVM (well, int locals show up correctly in the VS debugger, but no structs etc.).

JohanEngelen:
It'd be nice to get this into 1.1.0.

JohanEngelen:
@Syniurge Thanks a lot :)

JohanEngelen:
@Syniurge Commit 7ac12162920dd6e5d8322e31b22b4f6a97f9b0e3 introduced a bug unfortunately.

Compilation of the following code fails with `-g -O3`
(`ldc2 -g -O3 -c debuginfobug.d` fails):

```d
void foo(string fmt)
{
    size_t fmtIdx;
    void nested()
    {
        auto a = fmt[fmtIdx .. $];
    }
}
```

```
❯ ldc2 -g -O3 -c debuginfobug.d
Error: piece is larger than or outside of variable
  tail call void @llvm.dbg.declare(metadata { i64, i8* }* undef, metadata !17, metadata !18), !dbg !19
!17 = !DILocalVariable(name: "fmtIdx", scope: !6, file: !4, line: 3, type: !13)
!18 = !DIExpression(DW_OP_bit_piece, 0, 128)
piece is larger than or outside of variable
  tail call void @llvm.dbg.value(metadata i64 0, i64 0, metadata !17, metadata !23), !dbg !19
!17 = !DILocalVariable(name: "fmtIdx", scope: !6, file: !4, line: 3, type: !13)
!23 = !DIExpression(DW_OP_bit_piece, 128, 64)
```

JohanEngelen:
If I compare our output for captured variables with Clang's:
- clang's nested-frame function parameter is properly typed as a struct, instead of our `i8*`.
- clang attaches debug information about the captured variables to that struct, instead of defining separate debug info for each variable like we do (all pointing to the same alloca with offsets)

JohanEngelen:
I have been working on a fix, but keep stumbling into other troubles elsewhere (I can fix the example, but then something else breaks). I think the root-cause is that we are doing something that LLVM doesn't support well (multiple dbg.declare with same alloca).

kinke:
Okay then the clang way seems to be the way to go, thanks for checking. Adding proper debuginfos directly to the struct is probably best done as extension of a refactored `gen/nested.cpp`. A nested context definitely deserves its own C++ class in LDC, there's a lot of complexity to be encapsulated in a better way.

JohanEngelen:
Yeah a refactoring is definitely needed. It's not trivial though, so it will take time.

Shall I put in a PR with a temporary fix, such that master can at least compile the example again?

kinke:
Sure, why not. While at it, you could add your switch-fix too. As David would probably put it, it's definitely better than what we currently have. :)

JohanEngelen:
(working on debug fix too, not a refactor, but something that "should" work)

JohanEngelen:
https://github.com/ldc-developers/ldc/pull/1644

JohanEngelen:
@kinke @klickverbot Digging further into LLVM's source code, I hit upon something and now I have a much simpler fix, which preserves the presentation of the variables as local variables (not in a frame) in the enclosing function. Inside the nested function, the variables also show up as local variables, but there i think it'd be better to present them inside a "this" struct, but that's for another time.
Preparing a second PR.

kinke:
> preserves the presentation of the variables as local variables (not in a frame) in the enclosing function

Perfect.

> Inside the nested function, the variables also show up as local variables, but there i think it'd be better to present them inside a "this" struct, but that's for another time.

Locals seem okay to me. If we want to use a debug-struct, it's probably quite complex due to the pointers to the parent frames inside the innermost frame (additional indirection). And a imho more appropriate name for that debug-struct would be 'closure' instead of 'this'.

> Preparing a second PR.

Thanks! I just hope it doesn't take that much time until I can actually see some debuginfos in VS...

JohanEngelen:
I am more confused about it now :(
I'm reading dwarf spec and llvm code, trying to figure out what is going on and what is supposed to be going on. Is it an LLVM bug, LDC bug, ??!?
By now, I think what we are doing is correct, and that we hit upon an LLVM bug (clang perhaps never does the thing we are doing). I need to sleep over it, and probably need someone from LLVM who knows more about debuginfo (documentation is scarce).

The workaround that disables the LLVM checks that bug us: https://github.com/ldc-developers/ldc/compare/master...JohanEngelen:alternativedebugfix
It makes use of code for which the devs commented: ` // FIXME: Remove this check as soon as clang stops doing this; it hides bugs.` ("this check" refers to `if(shouldWeStopVerifying)`) ....

JohanEngelen:
@kinke @klickverbot @Syniurge By now, I believe we have found an LLVM bug. And we have to find a workaround. I think I've found the best workaround. The clue was in an earlier discussion: https://github.com/ldc-developers/ldc/pull/1598#discussion-diff-69776163 . The original code was wrong (GEP+offset), it was changed to alloca+offset but hits an LLVM bug, and now I've found something that does not bug:  GEP + _deref_. In LLVM source I found that an implicit deref is added on AllocA instructions passed to llvm.dbg.declare; I can't find where anymore... I don't know if a DW_OP_offset is applied before or after that implicit deref. With a GEP + deref, compilation no longer fails, and things look good inside LLDB.
We now have 3 "fixes" for the bug:
1. https://github.com/ldc-developers/ldc/pull/1644
2. https://github.com/ldc-developers/ldc/compare/master...JohanEngelen:alternativedebugfix
3. https://github.com/ldc-developers/ldc/pull/1646

__Number 3 is the correct fix (I think).__

kinke:
> Number 3 is the correct fix (I think).

I think too. It's also just a tiny diff, and I like the simplicity via the direct GEP. Thanks for browsing through the LLVM source to figure out we need to dereference the address manually. I guess that will come in handy in more places in the future...

JohanEngelen:
> I guess that will come in handy in more places in the future...

If my reasoning is correct, the other `EmitLocalVariable` in `nested.cpp` is wrong too; but I don't dare touch it: it works! (no crash and somehow it works in lldb).

Issue #1335, opened 2016-03-03
out contract breaks returning non-copyable structs from functions.


epi:
The following code compiles with DMD 2.069.2 and works, but fails to compile with ldc2-1.0.0-alpha1-linux-x86_64:
```
Error: struct test.Bar is not copyable because it is annotated with @disable
```
It compiles when the out contract is removed.

```
struct Bar
{
	@disable this(this);
}

auto foo()
out {}
body
{
	return Bar();
}

void main()
{
	auto x = foo();
}
```

kinke:
Confirmed with current master.
We're apparently calling the postblit ctor on the return value because the return expression's type (op) is `TOKvar` if there's an out contract and `TOKstructliteral` if there's no contract.

PR #1599, opened 2016-07-02
Some more direct in-place construction

kinke:
The plan is to in-place construct return expressions too if the function uses sret.
And I'd also like to get rid of all the temporary struct literals (possibly array literals too) which get memcopied too instead of being directly constructed into the variable. This should make life easier for the LLVM optimizer and speed-up debug builds as well as render the IR more compact and easier to read.

### DIFF
--- a/gen/statements.cpp
+++ b/gen/statements.cpp
@@ -145,12 +145,12 @@ public:
           LLType::getVoidTy(irs->context())) {
         // sanity check
         IrFunction *f = irs->func();
-        assert(getIrFunc(f->decl)->retArg);
+        assert(getIrFunc(f->decl)->sretArg);
 
         // FIXME: is there ever a case where a sret return needs to be rewritten
         // for the ABI?
 
-        LLValue *sretPointer = getIrFunc(f->decl)->retArg;
+        LLValue *sretPointer = getIrFunc(f->decl)->sretArg;
         DValue *e = toElemDtor(stmt->exp);
         // store return value
         if (!e->isLVal() || DtoLVal(e) != sretPointer) {
--- a/gen/statements.cpp
+++ b/gen/statements.cpp
@@ -150,17 +150,18 @@ public:
         // FIXME: is there ever a case where a sret return needs to be rewritten
         // for the ABI?
 
-        // get return pointer
-        DValue *rvar = new DLValue(f->type->next, getIrFunc(f->decl)->retArg);
+        LLValue *sretPointer = getIrFunc(f->decl)->retArg;
         DValue *e = toElemDtor(stmt->exp);
         // store return value
-        if (!e->isLVal() || DtoLVal(e) != DtoLVal(rvar))
-          DtoAssign(stmt->loc, rvar, e, TOKblit);
+        if (!e->isLVal() || DtoLVal(e) != sretPointer) {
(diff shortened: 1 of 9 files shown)


kinke:
The last commit doesn't work, producing runtime crashes for a few Phobos unittests and the `d_do_test` tool. I'd be grateful for any hints as to what could be wrong or what needs to be taken into account for return statements... ;)

gen/statements.cpp
```diff
@@ -151,17 +150,22 @@ class ToIRVisitor : public Visitor {
...
-        // store return value
-        if (!e->isLVal() || DtoLVal(e) != sretPointer) {
-          DLValue rvar(f->type->next, sretPointer);
-          DtoAssign(stmt->loc, &rvar, e, TOKblit);
-        }
+        DLValue returnValue(f->type->next, sretPointer);
+
+        // construct it directly if the expression is a suited sret call (just
+        // forward the sret pointer to the callee)
+        bool constructed = toDirectSretConstruction(&returnValue, stmt->exp);
```
kinke: Subtle potential bug: `toDirectSretConstruction()` may not destruct temporaries (unlike `toElemDtor()` below). Is that done after each statement anyway or do I need to take care of this?

kinke:
I've added in-place construction for struct literals.

gen/statements.cpp
```diff
@@ -135,32 +135,37 @@ class ToIRVisitor : public Visitor {
...
+
+        LLValue *sretPointer = getIrFunc(fd)->sretArg;
+        assert(sretPointer);
+
+        assert(!f->irFty.arg_sret->rewrite &&
+               "ABI shouldn't have to rewrite sret returns");
+        DLValue returnValue(f->type->next, sretPointer);
+
+        // try to construct the return value in-place
+        bool constructed = toInPlaceConstruction(&returnValue, stmt->exp);
```
kinke: The last commit works on my box (dmd-testsuite-debug passes at least) if I disable the in-place construction here for return values.

kinke:
I just noticed that AppVeyor always merges the latest head of the PR branch instead of the triggering commit. :/

kinke:
I've dropped the in-place construction for return values for now.

kinke:
Now also includes in-place construction for static array literals, getting rid of another obsolete memcpy.

kinke:
I've added 2 more commits wrt. the special `__result` variable inserted for functions with out contracts, fixing issue #1335 and eliding another useless alloca + memcpy for sret functions with out contracts.

kinke:
@klickverbot: You may mean the issue I've created (https://github.com/ldc-developers/ldc/issues/1601). So you're saying that's not an issue? Could well be, I haven't thought a lot about it. ;)

kinke:
Getting these IR tests to work is quite a pain, due to different LLVM version differences and platform differences. This commit should work for x86 too now (different memset intrinsic...).

JohanEngelen:
@kinke The work is much appreciated!  (needless to say I am very happy with the IR tests too ;)

kinke:
Test for issue #1335: https://github.com/ldc-developers/dmd-testsuite/commit/aeeee0f3e89020759467e6fe1b8fc0be09e72f81

PR #1602, opened 2016-07-04
Use direct sret construction for return statements too

kinke:
Okay so this sadly isn't working.
Some Phobos unittests crash (and so does dmd-testsuite's `d_do_test` tool), apparently due to a lifetime issue exposed by regex stuff. Specifically, the `std.regex.internal.parser` tests crash in a `std.uni.CowArray` dtor (the single field, a `uint[]` slice, seems corrupt when setting the `refCount` property). This is triggered by `std.regex.internal.ir.wordCharacter()` which uses lazy caching via `memoizeExpr()`.
So it seems like there's a missing postblit call. But iirc, the IR for `std.regex.internal.ir` was pretty similar (except for a lot of omitted `.sret_tmp` allocas + memcopies), I didn't immediately spot the missing postblit call. Various experiments with `doPostblit` all failed (e.g., using the current logic for in-place construction too), and so did my attempts at generating a reduced test case.

So any hints are welcome. This isn't crucial, but the IR would be much nicer. ;)

PS: It has nothing to do with directly constructing struct/array literals. It works if I disable the forwarding of the sret pointer to a function producing the return value (`S foo() { return bar(); }`).

### DIFF
--- a/gen/statements.cpp
+++ b/gen/statements.cpp
@@ -135,52 +135,69 @@ public:
     emitCoverageLinecountInc(stmt->loc);
 
     // The LLVM value to return, or null for void returns.
-    llvm::Value *returnValue = nullptr;
+    LLValue *returnValue = nullptr;
+
+    IrFunction *const f = irs->func();
+    FuncDeclaration *const fd = f->decl;
+    LLFunction *const llFunc = f->func;
 
     // is there a return value expression?
-    if (stmt->exp || (!stmt->exp && (irs->topfunc() == irs->mainFunc))) {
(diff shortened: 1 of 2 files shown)


kinke:
Updated - should be working now. :)

The cleanups need to be run directly after in-place construction (mimicking `toElemDtor()`). I though the cleanup code a bit further down in the same function (`visit(ReturnStatement*)`) would take care of this, but apparently it doesn't.

kinke:
Noone willing to review this little thing? ;)

Issue #1608, opened 2016-07-06
Missing post-blit calls when copying a static array from a slice.


EyalIO:
This program exhibits buggy behavior:
```D
import std.stdio:writeln;

struct Elem {
    int x = -1;
    this(int x) { this.x = x; writeln("CTOR ", x, " (", cast(void*)&this, ")"); }
    this(this) { writeln("POSTBLIT ", x, " (", cast(void*)&this, ")"); }
    ~this()    { if (x!=-1) writeln("DTOR "    , x, " (", cast(void*)&this, ")"); }
}

struct Ctr {
    Elem[3] arr;
    Elem[] slice() { return arr; }
    Elem[3] arrVal() { return arr; }
}

void main() {
    writeln("construct Ctr {");
    auto p = Ctr();
    writeln("}");
    writeln("assign arr {");
    p.arr = [Elem(1), Elem(2), Elem(3)];
    writeln("}");
    {
        writeln("slice rval -> arr {");
        Elem[3] _arr = p.slice;
        writeln("}");
    }
    {
        writeln("arr rval -> arr {");
        Elem[3] _arr = p.arrVal();
        writeln("}");
    }
}
```

This program fails to call POSTBLIT when copying from a slice-type, thus causing the number of CTOR/POSTBLITS to mismatch the number of DTORs.

Program's output:
```
construct Ctr {
}
assign arr {
CTOR 1 (7FFFCA8A2468)
CTOR 2 (7FFFCA8A2458)
CTOR 3 (7FFFCA8A2448)
}
slice rval -> arr {
}
DTOR 3 (7FFFCA8A2410)
DTOR 2 (7FFFCA8A240C)
DTOR 1 (7FFFCA8A2408)
arr rval -> arr {
POSTBLIT 1 (7FFFCA8A2370)
POSTBLIT 2 (7FFFCA8A2374)
POSTBLIT 3 (7FFFCA8A2378)
}
DTOR 3 (7FFFCA8A2400)
DTOR 2 (7FFFCA8A23FC)
DTOR 1 (7FFFCA8A23F8)
DTOR 3 (7FFFCA8A24B0)
DTOR 2 (7FFFCA8A24AC)
DTOR 1 (7FFFCA8A24A8)
```
Note the missing `POSTBLIT` printouts under `slice rval -> arr {`...

EyalIO:
Similar, but different bug in dmd: https://issues.dlang.org/show_bug.cgi?id=16197

[...]

PR #1609, opened 2016-07-06
Fix issue #1608

kinke:
Not happy at all with this change, that whole `canSkipPostblit` logic seems quite fragile. It's determined in `visit(AssignExp*)`, passed to `DtoAssign()`, and exclusively used in `DtoArrayAssign()` (i.e., when assigning to a static/dynamic array).

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -106,7 +106,7 @@ void DtoSetArrayToNull(LLValue *v) {
 ////////////////////////////////////////////////////////////////////////////////
 
 static void DtoArrayInit(Loc &loc, LLValue *ptr, LLValue *length,
-                         DValue *dvalue, int op) {
+                         DValue *dvalue) {
   IF_LOG Logger::println("DtoArrayInit");
   LOG_SCOPE;
 
@@ -165,7 +165,7 @@ static void DtoArrayInit(Loc &loc, LLValue *ptr, LLValue *length,
   // assign array element value
   DLValue arrayelem(dvalue->type->toBasetype(),
                     DtoGEP1(ptr, itr_val, true, "arrayinit.arrayelem"));
-  DtoAssign(loc, &arrayelem, dvalue, op);
+  DtoAssign(loc, &arrayelem, dvalue, TOKblit);
 
   // increment iterator
   DtoStore(gIR->ir->CreateAdd(itr_val, DtoConstSize_t(1), "arrayinit.new_itr"),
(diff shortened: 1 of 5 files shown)


kinke:
Slices (dynamic arrays) always represent ranges of lvalues since they point to unowned memory. Is there ever a case where copying from a slice can be optimized to a blit, without postblit for the copied elements and without destroying the assigned-from right-hand-side elements? Figuring out whether it's legal to do so should be quite complex, so I doubt it's feasible and required.

ZombineDev:
> Is there ever a case where copying from a slice can be optimized to a blit, without postblit for the copied elements

That should be when the array element type is a POD (Plain Old Data) type. Perhaps you can usr [this](https://github.com/dlang/dmd/blob/v2.071.1/src/aggregate.h#L171) to determine that.

> and without destroying the assigned-from right-hand-side elements? 

That would mean that the array elements were moved instead of copied. I think only `std.algorithm.mutation : move` does that. I am not aware of a case where the frontend would generate moves. Furthermore only owners should be allowed to move elements and since slices conceptually borrow elements I don't think computer would generate implicit moves. OTOH, dmd is known to move objects placed on thr stack, so if were talking for static arrays, I may be wrong

kinke:
> Furthermore only owners should be allowed to move elements and since slices conceptually borrow elements I don't think the compiler would generate any moves.

Thanks, that's what I thought.
[We (LDC) obviously already have the POD optimization, I meant non-POD types only.]

kinke:
Wrt. moving from a static array, here'd be an example for code that could possibly be moved, but shows the difficulties of analyzing this:

```D
int numPostblit = 0, numDtor = 0;

struct S
{
    int v;
    this(this) { ++numPostblit; }
    ~this() { ++numDtor; } }

void foo()
{
    S[4] sa = [ S(1), S(2), S(3), S(4) ];

    // helper to generate a slice rvalue
    static S[] toSlice(ref S[4] sa) { return sa[1..$]; }

    S[3] r = toSlice(sa); // elements 2,3,4 could possibly be moved as they'd be destructed
                          // immediately anyway as the owner `sa` goes out of scope
                          // element 1 would still need to be destructed as non-moved part of `sa`
}

void main()
{
    foo();
    assert(numPostblit == 3);
    assert(numDtor == 7);
}
```

The above passes with this PR and DMD 2.071; fails due to missing postblit for LDC head.

kinke:
I tried to make the `DtoArrayAssign()` postblit behavior consistent with our manual `callPostblit()` calls for structs (e.g., by passing `canSkipPostblit = false` when we do a postblit for structs manually afterwards), but apparently there are cases where we must not do postblits for array elements but where they are required for structs. So this still feels fragile to me.

What remains is a fix for issue #1608 and a visit of all `DtoAssign()` call sites, where I replaced the TOKconstruct + canSkipPostblit combo by a semantically equivalent TOKblit. I also replaced assignments from DSliceValues to simple and intuitive stores (of the length+ptr pair).

kinke:
Looking forward to see this fix in 1.1.

kinke:
Löl that was quick. ;)

Issue #1452, opened 2016-04-21
Add support for new `allocsize` attribute in LLVM


JohanEngelen:
(a todo item, copied from rust's https://github.com/rust-lang/rust/issues/33075 ;)

See http://reviews.llvm.org/rL266032

PR #1610, opened 2016-07-06
Implement the @allocSize magic UDA.

JohanEngelen:
Resolves https://github.com/ldc-developers/ldc/issues/1452

I implemented it "because I can". I think this leads to optimization potential for LLVM (InstCombine?) but I have not tested what kind of code becomes better.

In subsequent work, the attribute should be applied to druntime functions. In the end, I am not sure how much this brings. When we can inline e.g. `_d_allocmemory`, it will be inlined to a `malloc` call (if I'm correct) for which LLVM automatically assumes the `allocsize` attribute...

### DIFF
--- a/gen/uda.cpp
+++ b/gen/uda.cpp
@@ -15,6 +15,7 @@ namespace {
 
 /// Names of the attribute structs we recognize.
 namespace attr {
+const std::string allocSize = "allocSize";
 const std::string llvmAttr = "llvmAttr";
 const std::string llvmFastMathFlag = "llvmFastMathFlag";
 const std::string optStrategy = "optStrategy";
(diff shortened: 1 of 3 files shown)


JohanEngelen:
Pursuing this again, because I found out it is important in removing allocations that are not used.
For example, code like this:
```d
void foo()
{
    auto p = myAlloc(100);
}
```
When the user marks `myAlloc` with `allocsize`, it tells the compiler that the call can be removed when the allocated memory is not used meaningfully. So, importantly, code like this will be completely removed:
```d
void foo()
{
    auto p = myAlloc(100);
    p[0] = 1;
}
```

JohanEngelen:
jay, green

thewilsonator:
have you had a chance to test this with sample code using std(.experimental).allocator?

JohanEngelen:
@thewilsonator I have only tested this with the simple example above (https://github.com/ldc-developers/ldc/pull/1610#issuecomment-261229294). But indeed, it is definitely worthwhile adding this to relevant functions in druntime. The attribute does assume that there are no other (important) sideeffects besides allocation though, so that it is valid to remove the calls in the two examples.
I'll add that to the description of the attribute and merge.

thewilsonator:
From what I understand OpenBSD's reallocarray (`void *
reallocarray(void *ptr, size_t nmemb, size_t size);`) would be `@allocSize(1,2)` ?

Maybe give that as the source example in `ldc.attributes`

JohanEngelen:
I've added the `reallocarray` example.

Merging when green.

JohanEngelen:
http://lists.llvm.org/pipermail/llvm-dev/2017-April/112525.html

PR #1612, opened 2016-07-07
Fix issue 1611

JohanEngelen:


kinke:
Skipping empty fields in `write_struct_literal()` seems to work:
```C++
if (vd->size(loc) == 0)
  continue;
```
right after setting `vd`.

kinke:
Side note: 0-length static array arguments seem to be passed to the callee and taken as parameters by the callee. On Win64, `ExplicitByvalRewrite` is applied, i.e., the caller `alloca`tes 0 bytes for the argument copy, performs a 0-length memcpy, and then passes the pointer to the callee. This seems kind of obsolete... ;)

JohanEngelen:
@kinke Thanks a lot for the fix. 

(I fixed my missing asserts problem. Soo much better!)

tests/codegen/zerolengtharray_gh1611.d
```diff
@@ -0,0 +1,39 @@
...
+//         uint a = 111;
+//         uint c = 333;
+//     }
+
+// CHECK-LABEL: define{{.*}}fooA0{{.*}} {
+auto fooA0()
+{
+    return A0();
+    // Intentionally a regexp to not match "sret"
+    // CHECK: {{ret }}
```
kinke: How does that not match s**ret**?

tests/codegen/zerolengtharray_gh1611.d
```diff
@@ -0,0 +1,39 @@
+// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+struct A0
+{
+    ubyte[0] zerolen;
+}
+// CHECK-DAG: %{{.*}}.A0 = type { [1 x i8] }
+
+struct uint_0_bool
```
kinke: No `bool` in there.

gen/toir.cpp
```diff
@@ -98,6 +98,11 @@ static void write_struct_literal(Loc loc, LLValue *mem, StructDeclaration *sd,
   for (size_t index = 0; index < nfields; ++index) {
     VarDeclaration *vd = sd->fields[index];
 
+    // Skip fields with zero size. An example of a zero-size field is
+    // zero-length static array: `ubyte[0] data`.
```
kinke: I'd shorten that to `skip zero-sized fields, such as zero-length static arrays`.

tests/codegen/zerolengtharray_gh1611.d
JohanEngelen: It was meant to not match sret_... but yeah, needs fixing. Thanks.

gen/toir.cpp
JohanEngelen: changed.

tests/codegen/zerolengtharray_gh1611.d
JohanEngelen: fixed.

tests/codegen/zerolengtharray_gh1611.d
JohanEngelen: Before, I had tests that counted the number of stores, but I thought it is too flaky. Codegen could be optimized to just completely ignore the ubyte[0] and do only one store (like it does for when ubyte[0] field is removed). I think this is what you meant with your earlier post, right?
Anyway, don't really know how to test the currently generated IR well, so decided to not test it at all for now.

kinke: If you want to make sure the empty struct (padded to 1 byte) isn't returned via sret, you should simply make sure the function has no parameters `CHECK-LABEL: define {{.*}}fooA0{{.*}}()`. That doesn't have anything to do with zero-length static arrays though; it's just how empty structs with their `[1 x i8]` padding are handled generally in LDC.
Imho, your checks wrt. to the IR type definitions are sufficient here (no zero-length static array fields); outputting the IR also makes sure the ICE is gone.

tests/codegen/zerolengtharray_gh1611.d
kinke: It's still in the name of the 2nd function. ;)

kinke:
I think we should try to ignore empty arguments and parameters on the IR side, at least for `extern(D)` (where the function's D signature is encoded via mangling but the LDC ABI may omit some useless parameters). I.e.,

1. omitting by-value passing of zero-length-static array arguments
2. skipping those parameters when generating the LL function type
3. setting the parameter variables when defining the function (if not at all omittable)

For that to work, we need a more flexible D-argument-to-LL-parameter mapping. And at some point we could also think about doing the same for empty structs (e.g., by zero-initializing empty struct parameters with 0 and treating all empty structs as single 0-byte), depending on the `TargetABI` to decide whether it can be omitted or not (e.g., depending on linkage - IIRC, empty structs aren't passed for `extern(C)` on OSX).

kinke:
So you merged without considering my latest comments? I'd prefer you wait for others to properly review and merge your PRs. I know I merge mine too myself sometimes, but I try to restrict that to tiny fixes or to cases where I can't wait any longer. ;)

JohanEngelen:
Which comment do you refer to?

I saw the "sret" comment. I wasn't checking whether the struct is returned by sret or not. The regexp thing was there to be able to properly test absence/existence of instructions in the function body.

Ah, just now I see your other comment about the testcase. dammit. sry.

kinke:
Yep, comments https://github.com/ldc-developers/ldc/pull/1612#discussion_r70177990 and https://github.com/ldc-developers/ldc/pull/1612#discussion_r70177996.

JohanEngelen:
Fixed the testcase naming in master.

kinke:
And what about the IMO useless ' ret '-checks? I still don't get what you want to test there. And what about the comment wrt. `uint_0_uint` codegen differing from `uint_uint`? I thought there'd be no diff with this change.

Here's how I imagine the IR test:

```D
// RUN: %ldc -run %s

// 1) Make sure zero-length static arrays are omitted as struct fields.

struct A0
{
    ubyte[0] zerolen;
}
// make sure it's a regular empty struct (and as such padded to 1 byte)
// CHECK-DAG: %{{.*}}.A0 = type { [1 x i8] }

struct uint_0_uint
{
    uint a = 111;
    ubyte[0] zerolen;
    uint c = 333;
}
// CHECK-DAG: %{{.*}}.uint_0_uint = type { i32, i32 }

// 2) Make sure literals can be constructed.

void main()
{
    auto a = A0();
    assert(a == A0.init);

    auto b = uint_0_uint();
    assert(b == uint_0_uint.init);
}
```

JohanEngelen:
There is no codegen change with this PR. But currently the IR is different for
```d
struct uint_0_uint
{
    uint a = 111;
    ubyte[0] zerolen;
    uint c = 333;
}
```
vs.
```d
struct uint_uint
{
    uint a = 111;
    uint c = 333;
}
```

`uint_uint` results in an i64 return instead of sret (at least for Mac OS triple).

Before the fix, this would not result in an ICE:
```d
A0 foo()
{
    auto a = A0();
    return a;
}
```
whereas this would:
```d
A0 foo()
{
    return A0();
}
```
So the test explicitly tests compilation of the second crashing case.

Yes the " ret " checks are useless. But they don't harm either, and may help when someone wants to elaborate the tests. The checks could just aswell be a test for the closing "}", it is there to detect function end.

kinke:
> So the test explicitly tests compilation of the second crashing case.

Okay good. That may be affected by #1602 though.

> The checks could just aswell be a test for the closing "}", it is there to detect function end.

Then I find these checks even more confusing - you check that the 2 functions are there and that they end? ;) Not confusing enough to justify another commit though.

> uint_uint results in an i64 return instead of sret (at least for Mac OS triple).

On Win64, `uint_0_uint` is also returned as int64, fitting nicely into a 64-bit register. So just for future reference: this must be a bug in DMD's `toArgTypes()` which we use for the System V (non-Windows x86_64) ABI. We already know that it's a pretty lousy implementation, and here it proves it again.

JohanEngelen:
:( :(  Open an issue about improving zero-length array stuff?

PR #1619, opened 2016-07-12
Non-Windows x86_64: Pass more structs in registers

kinke:
* Treat empty structs as single byte instead of forcing them to be passed
  on the stack and returned via sret.
  This also enables other structs containing empty struct fields to be
  passed/returned in registers.
* Ignore zero-sized fields of a struct, such as zero-length static arrays.
  These fields previously forced the struct to be passed/returned on the
  stack.

I'll try to get that into upstream.

### DIFF
--- a/ddmd/argtypes.d
+++ b/ddmd/argtypes.d
@@ -276,7 +276,7 @@ extern (C++) TypeTuple toArgTypes(Type t)
         override void visit(TypeStruct t)
         {
             //printf("TypeStruct.toArgTypes() %s\n", t.toChars());
-            if (!t.sym.isPOD() || t.sym.fields.dim == 0)
+            if (!t.sym.isPOD())
             {
             Lmemory:
                 //printf("\ttoArgTypes() %s => [ ]\n", t->toChars());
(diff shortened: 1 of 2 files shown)


kinke:
C(++) requires empty structs to be passed on the stack. We'd have to special-case in `toArgTypes()`, based on function linkage, but that's tedious. I'll probably just reduce the change to skipping zero-sized struct fields.

thewilsonator:
For completely empty structs this should be fine. However from the spec 

    A static array with a dimension of 0 is allowed, but no space is allocated for it. _It's useful as the last member of a variable length struct_, or as the degenerate case of a template expansion.

 So I don't think its a good idea, mind you variable length structs should only ever be accessed by a pointer anyway.

kinke:
It's the other way around. There is no such thing as a completely empty struct in D, they always occupy one byte for some bizarre reason (need to have an identity and so a unique storage, or something along those lines).
That hack for 'variable length structs' is only applicable if you know what you are doing. Copying such structs (as done when passed as by-value parameter) will obviously result in mayhem, so there's no need to treat 0-length static array fields as truly empty types in any special way. Currently though, such fields prevent the containing struct from being passed in registers.

PR #1620, opened 2016-07-13
Fix evaluation order issues

### LINKED ISSUE #1617: opOp evaluation order difference with DMD (e.g. `+=`)
Consider the following code:
```d
int sum;

int return1_add9tosum()
{
    sum += 9;
    return 1;
}

void main()
{
    sum = 0;
    sum += return1_add9tosum();

    import std.stdio;
    writeln(sum);
}
```

DMD 2.071 writes 10. LDC writes 1. It's an evaluation order difference. The spec says that `sum` should only be evaluated once; left-to-right evaluation is specced, but then the spec also says that DMD does not comply with left-to-right evaluation of AssignExpressions.
Commit a3adb403de1554cab73cfa9df88dc5a3dfc7940f of #1562 changed LDC's behavior (before it behaved like DMD).

With fibers in play, the evaluation order is important even if `foo()` does not modify `sum` in `sum += foo()`. When `foo()` yields, another fiber might modify `sum`, and thus an older version of `sum` will be used when the fiber is resumed. I'm pretty sure this is what happened in Weka's unittest failure. In light of that, I think DMD's evaluation order makes more sense (it would make "+=" do the intuitive thing when using fibers).

JohanEngelen:
Resolves #1617 .

There is a lot of refactoring potential here (duplication of code), but it is additional work and we first need to get this right.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -626,57 +626,6 @@ public:
 
   //////////////////////////////////////////////////////////////////////////////
 
-  template <typename BinExp, bool useLvalForBinExpLhs>
-  static DValue *binAssign(BinAssignExp *e) {
-    Loc loc = e->loc;
-
-    // find the lhs' lvalue expression
-    Expression *lvalExp = findLvalueExp(e->e1);
-    if (!lvalExp) {
-      e->error("expression %s does not mask any l-value", e->e1->toChars());
-      fatal();
-    }
-
-    // pre-evaluate and cache the lvalue subexpression
-    DValue *lval = nullptr;
-    {
-      IF_LOG Logger::println("Caching l-value of %s => %s", e->toChars(),
-                             lvalExp->toChars());
-
-      LOG_SCOPE;
-      lval = toElemAndCacheLvalue(lvalExp);
-    }
-
-    // evaluate the underlying binary expression
-    Expression *lhsForBinExp = (useLvalForBinExpLhs ? lvalExp : e->e1);
-    BinExp *binExp = bindD<BinExp>::create(loc, lhsForBinExp, e->e2);
-    binExp->type = lhsForBinExp->type;
-    DValue *result = toElem(binExp);
-
-    lvalExp->cachedLvalue = nullptr;
-
-    // assign the (casted) result to lval
-    DValue *assignedResult = DtoCast(loc, result, lval->type);
-    DtoAssign(loc, lval, assignedResult);
-
-    // return the (casted) result
-    return e->type == lval->type ? lval : DtoCast(loc, lval, e->type);
-  }
-
-#define BIN_ASSIGN(Op, useLvalForBinExpLhs)                                    \
-  void visit(Op##AssignExp *e) override {                                      \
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Let's see what the CI testers say... meanwhile, I am going to try and compile Weka's codebase with it.

JohanEngelen:
There is some funny business happening in the code that is wholly unnecessary I think. Let's see what the CI testers think... First I'll commit the (I hope) fully fixed version, then a second commit with the funny business cut out.

JohanEngelen:
Cancelled all CI tests, amend+pushforce to start all of them again.
Hopefully mergable when all is green.

gen/toir.cpp
```diff
@@ -2152,22 +2249,97 @@ class ToElemVisitor : public Visitor {
...
+    // find the lhs' lvalue expression
+    Expression *lvalExp = findLvalueExp(e->e1);
+    if (!lvalExp) {
+      e->error("expression %s does not mask any l-value", e->e1->toChars());
+      fatal();
+    }
+
+    // FIXME: I don't know why this is there. Is it needed?
+    bool useFindLval =
+        (binOp == llvm::Instruction::Shl) || (binOp == llvm::Instruction::LShr);
```
kinke: Yes I think it's needed. Test:

````D
int x = 6;
int inc() { ++x; return 4; }
(x++) <<= inc();
assert(x == 8 << 4);
```

[Note that I haven't actually tested this code. ;)]
Using the `lhs_val` for the lhs makes sure we use the lvalue subexpression `x` which should be 8 by the time we load the lhs for the binop. `lhs_fullEvaluation`, i.e., `(x++)`, would be 6.

But I now think that this should probably be the case for all other binassign expressions, not just for shifts...

kinke:
> There is some funny business happening in the code that is wholly unnecessary I think.

Hmm? The pre-evaluation of the lvalue subexpression or what do you mean?

> Hopefully mergable when all is green.

No rush mate ;) - let's refactor this a bit, 170 additional lines shouldn't be necessary. That doesn't have to be you, I could do that as well, but that might take some time.

JohanEngelen:
Indeed I thought pre-evaluation of `findLvalueExp` was not needed (because right after, we evaluate `e->e1`). But perhaps that's because I don't fully understand what is going on. With your explanation for the <<= case, I think I understand better now why it is needed.

`lhs_fullEvaluation` is not a good name, do you have a better idea for a name?

(keep pasting testcase ideas too!)

kinke:
Hmm this may all be a tedious fight with a symptom. Firstly, the lhs `x++` is refused by the front-end as it yields no lvalue, makes sense. So I think this whole pre-evaluation thing may only be needed because we don't/used not to always return a `DLValue` for `toElem(e->e1)`...

I'll have a go at some refactoring and will then try to get rid of the pre-evaluation.

kinke:
Continued here: https://github.com/ldc-developers/ldc/pull/1623

JohanEngelen:
@kinke shall I close this one?

PR #1622, opened 2016-07-16
Tell Travis to use LLVM 3.8.1

JohanEngelen:
(we are again part of the release notes :-) 👍 )

http://llvm.org/releases/3.8.1/docs/ReleaseNotes.html


JohanEngelen:
http://lists.llvm.org/pipermail/llvm-dev/2016-July/102435.html

PR #1623, opened 2016-07-16
Fix evaluation order issues (continuation of #1620)

kinke:
Refactorings only for now.


JohanEngelen:
Nice work!
Haven't look at in detail, but looks great already.

kinke:
Using the lhs lvalue for shift-assignments seems to have to do with the precision in which the shift operation is executed. We come across stuff rewritten by the front-end like

```D
ushort bar = ...;
cast(ulong)cast(int)bar <<= 8;
```

The front-end seems to like to cast the lhs to specify the precision of the intermediate binop result, not just for shift-assignments.

Anyway, the last commit with the enforced DValue type consistency was a trip down the rabbit hole (I thought this may be related to the shift-assign strangeness). But I'm sure it's worth it, possibly rendering casts in other places useless. I only fixed all assertions popping up when executing the full test suite - that took long enough. ;)

gen/tocall.cpp
```diff
@@ -596,7 +596,7 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
 
     Expression *exp1 = (*e->arguments)[0];
     LLValue *ptr = DtoRVal(exp1);
-    result = new DImValue(exp1->type, DtoVolatileLoad(ptr));
+    result = new DImValue(e->type, DtoVolatileLoad(ptr));
```
kinke: The previous D return type for the volatile-load-intrinsic was `T*` (as the pointer arg to be loaded from) instead of `T`.

gen/toir.cpp
```diff
@@ -1747,13 +1729,8 @@ class ToElemVisitor : public Visitor {
     auto &PGO = gIR->func()->pgo;
     PGO.setCurrentStmt(e);
 
-    // DMD allows syntax like this:
-    // f() == 0 || assert(false)
-    result = new DImValue(e->type, DtoConstBool(false));
```
kinke: Asserts don't yield a value anymore (their type is void).

gen/toir.cpp
```diff
@@ -1885,6 +1862,11 @@ class ToElemVisitor : public Visitor {
     llvm::BranchInst::Create(andandend, p->scopebb());
     p->scope() = IRScope(andandend);
 
+    if (e->type->toBasetype()->ty == Tvoid) {
+      result = nullptr;
+      return;
+    }
+
```
kinke: `&&` and `||` expressions may have a void type if a side is an assert expression - `condition() || assert(0)`.

gen/toir.cpp
```diff
@@ -2181,11 +2168,8 @@ class ToElemVisitor : public Visitor {
     llvm::BranchInst::Create(condend, p->scopebb());
 
     p->scope() = IRScope(condend);
-    if (retPtr) {
+    if (retPtr)
       result = new DSpecialRefValue(e->type, retPtr);
-    } else {
-      result = new DConstValue(e->type, getNullValue(DtoMemType(dtype)));
-    }
```
kinke: No more dummy value if `CondExp::type` is void.

gen/dvalue.cpp
```diff
@@ -86,31 +98,8 @@ bool DFuncValue::definedInFuncEntryBB() {
...
-  }
-
-  // bools should not be stored as i1 any longer.
-  if (pt->getElementType() == llvm::Type::getInt1Ty(gIR->context())) {
-    return false;
-  }
-
-  return true;
-}
-
```
kinke: The new asserts in the `DLValue` and `DSpecialRefValue` ctors are tighter.

kinke:
Alright, this should finally fix all binop/binassign eval/load order issues. Good to go as far as I'm concerned.

kinke:
Ah wait, there are still a few bugs.

kinke:
Okay, this should really be it now. I finally understood the shiftassign strangeness and fixed a related bug for all non-shift binassigns, see latest commit.

kinke:
The last commit is nice. :)

gen/toir.cpp
klickverbot: Might be non-obvious enough to justify a comment in the source.

gen/binops.cpp
```diff
@@ -7,102 +7,298 @@
...
   }
 
-  return new DImValue(targettype, res);
+  return new DImValue(type, res);
+}
+
+//////////////////////////////////////////////////////////////////////////////
+
+namespace {
+template <llvm::Instruction::BinaryOps binOp>
```
klickverbot: Does it really buy us anything to make this a template? Inlining should be sufficient?

gen/binops.h
```diff
@@ -0,0 +1,61 @@
...
+struct Loc;
+
+namespace llvm {
+class Value;
+}
+
+class DValue;
+
+// lhs + rhs
+DValue *binAdd(Loc &loc, Type *type, DValue *lhs, Expression *rhs,
```
klickverbot: Where are these used from?

gen/binops.cpp
kinke: Yeah no need for a template here anymore.

gen/binops.h
kinke: `toir.cpp` - each AssignExp and BinAssignExp uses the corresponding function from `binops.cpp`. It's not that obvious as we use macros in `toir.cpp` to reduce boilerplate.

kinke:
@klickverbot: Thanks for the suggestions.

JohanEngelen:
I've been trying to test this on Weka's codebase, but haven't been successful so far. Could be for unrelated reasons though.

Let's merge?

kinke:
Well I hope the Weka issues are unrelated, as I'm pretty confident this should work now. Should also definitely be in 1.1.

JohanEngelen:
The Weka issue is solved, was unrelated apparently.

JohanEngelen:
Definitely should be in 1.1!

PR #1630, opened 2016-07-20
More in-place construction

kinke:
Also for struct literal fields, array literal elements and associative array literal values as well as basic types allocated on the heap.

Touched files were re-clang-formatted, so there are pure formatting fixes too. :/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,7 @@ addons:
     - g++-multilib
     - libedit2
     - libedit-dev
+    - libcurl3:i386
 before_install:
   -
     if [ "${TRAVIS_OS_NAME}" = "linux" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,14 +76,17 @@ script:
   # Build Phobos & druntime unittest modules.
   -
     if [ "${TEST_CONFIG}" = "Debug" ]; then
-      MAKEOPTS=-j2 ctest --verbose -R "build-phobos2-ldc-unittest-debug";
-      MAKEOPTS=-j3 ctest --verbose -R "build-druntime-ldc-unittest-debug";
+      make -j2 phobos2-ldc-unittest-debug;
+      make -j3 druntime-ldc-unittest-debug;
     elif [ "${TEST_CONFIG}" = "Release" ]; then
-      MAKEOPTS=-j2 ctest --verbose -R "build-phobos2-ldc-unittest" -E "-debug";
-      MAKEOPTS=-j3 ctest --verbose -R "build-druntime-ldc-unittest" -E "-debug";
+      make -j2 phobos2-ldc-unittest;
+      make -j3 druntime-ldc-unittest;
+    elif [ "${OPTS}" = "-DMULTILIB=ON" ]; then
+      make -j2 phobos2-ldc-unittest-debug phobos2-ldc-unittest phobos2-ldc-unittest-debug-32 phobos2-ldc-unittest-32;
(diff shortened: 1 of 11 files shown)


kinke:
Consider an array-of-structs declaration like:
```D
struct S { int x; }
S[3] a = [ S(1), S(2), S(3) ];
```

Before all these in-place constructions, we would create 3 struct literals, memcpy them into an array literal, and finally memcpy that one into `a`. With this PR, the 3 struct literals are constructed directly into `a`, there's no memcpy anymore.

kinke:
@klickverbot: Wrong browser tab? ;)

kinke:
I've added a corresponding IR test. I also noticed that the AppVeyor performance has seen a big improvement, so I decided to merge the Debug+Release jobs, see last commit.

.travis.yml
```diff
@@ -76,14 +76,14 @@ script:
...
     elif [ "${TEST_CONFIG}" = "Release" ]; then
-      MAKEOPTS=-j2 ctest --verbose -R "build-phobos2-ldc-unittest" -E "-debug";
-      MAKEOPTS=-j3 ctest --verbose -R "build-druntime-ldc-unittest" -E "-debug";
+      make -j2 phobos2-ldc-unittest;
+      make -j3 druntime-ldc-unittest;
     else
-      MAKEOPTS=-j2 ctest --verbose -R "build-phobos2-ldc-unittest";
-      MAKEOPTS=-j3 ctest --verbose -R "build-druntime-ldc-unittest";
+      make -j2 phobos2-ldc-unittest-debug phobos2-ldc-unittest;
+      make -j3 druntime-ldc-unittest-debug druntime-ldc-unittest;
```
kinke: Hmm the 32-bit variants for the multilib job are missing here. Based on the [Travis results](https://travis-ci.org/ldc-developers/ldc/builds/146254327) (33 mins for OSX release, 24-28 mins for Linux x64), I'd say that `MAKEOPTS` didn't work as expected, leading to the unittests being compiled serially.

gen/toir.cpp
```diff
@@ -1459,9 +1459,10 @@ class ToElemVisitor : public Visitor {
       return;
     }
 
-    if (!elen) {
+    if (!elen)
       elen = DtoArrayLen(v);
-    }
+    eptr = DtoBitCast(eptr, DtoPtrToType(e->type->nextOf()));
```
JohanEngelen: Why is this cast needed now?

kinke: The new assert in the DSliceValue ctor - the LL pointer sometimes didn't match the D type.

kinke:
Alright, so there's a couple of intesting things here wrt. CI:

* AppVeyor timings for a full run have been halved compared to a week ago. Merging the 2 jobs for an architecture has contributed by about 25 minutes (1h 50min => 1h 25min).
* The AppVeyor x64 job is consistently 10% faster than the x86 job. That's surprising, as the x86 job used to be slightly faster. The only logical explanation is the usage of LDC (x64) vs. DMD (x86) to compile the front-end. I still find an overall reduction for the full build+test job by 10% rather surprising; I thought the time spent in the front-end was rather negligible compared to the LLVM backend.
* Travis would have been green for the first time in ages again - if it wasn't for two hanging 32-bit curl unittests. So that's the actual multilib Travis issue, most likely a corresponding package is missing.
* Replacing the `ctest` invokations to build the unittests by corresponding `make` commands seems to have a positive effect on the runtime as well. The Travis timings vary quite a bit, but the [results for this PR](https://travis-ci.org/ldc-developers/ldc/builds/146448735) are significantly faster than the [last master results](https://travis-ci.org/ldc-developers/ldc/builds/146371226).

kinke:
Yep baby, here we go. All green again.

PR #1633, opened 2016-07-22
Misc. debuginfo fixes + new LLVM 4.0 for AppVeyor

kinke:


gen/dibuilder.cpp
```diff
@@ -507,30 +506,38 @@ ldc::DISubroutineType ldc::DIBuilder::CreateFunctionType(Type *type) {
...
+  DIType voidPtrType = CreatePointerType(Type::tvoidptr);
 #if LDC_LLVM_VER >= 306
-  auto EltTypeArray = DBuilder.getOrCreateTypeArray(Elts);
+  llvm::Metadata *elems[] =
 #else
-  auto EltTypeArray = DBuilder.getOrCreateArray(Elts);
+  llvm::Value *elems[] =
 #endif
+      {voidPtrType, voidPtrType};
+  llvm::DINodeArray elemsArray = DBuilder.getOrCreateArray(elems);
```
kinke: Just represent a delegate as pair of void pointers, at least for now.

gen/dibuilder.cpp
```diff
@@ -544,49 +551,40 @@ bool isOpaqueEnumType(Type *type) {
...
     type = type->pointerTo();
     t = type->toBasetype();
   }
 
-  if (t->ty == Tvoid || t->ty == Tnull) {
+  if (t->ty == Tnull)
     return DBuilder.createUnspecifiedType(t->toChars());
-  }
+  if (t->ty == Tvoid)
+    return CreateBasicType(Type::tuns8); // byte
```
kinke: If `void` makes it here, it's most likely an array element type and needs to be treated as byte.

gen/dibuilder.cpp
```diff
@@ -478,8 +478,7 @@ ldc::DIType ldc::DIBuilder::CreateSArrayType(Type *type) {
 }
 
 ldc::DIType ldc::DIBuilder::CreateAArrayType(Type *type) {
-  // FIXME: Implement
-  return DBuilder.createUnspecifiedType(type->toChars());
+  return CreatePointerType(Type::tvoidptr);
```
kinke: Just a void pointer. Isn't that what it actually is?

klickverbot: On the ABI level, yes, but not what the programmer expects reading the source code. I don't have enough LLVM debug info working knowledge to judge how much of an improvement translating it to void* is.

gen/dibuilder.cpp
klickverbot: "Most likely"? ;)

A similar comment should probably be in the source.

kinke:
The AppVeyor x64 issue (LLVM segfault) is due to the new LLVM. All x64 **debug** tests pass.

@majnemer: If you have some time, it'd be great if you could look at [this IR](https://dl.dropboxusercontent.com/s/w4t7xrs3q89blqz/concurrency.zip?dl=0). If it is optimized, current `release_39` head segfaults. Stack trace:
```
llvm::LoopBase<llvm::BasicBlock,llvm::Loop>::getExitBlocks(llvm::SmallVectorImpl<llvm::BasicBlock *> & ExitBlocks) Line 66
llvm::formLCSSAForInstructions(llvm::SmallVectorImpl<llvm::Instruction *> & Worklist, llvm::DominatorTree & DT, llvm::LoopInfo & LI) Line 79
llvm::formLCSSA(llvm::Loop & L, llvm::DominatorTree & DT, llvm::LoopInfo * LI, llvm::ScalarEvolution * SE) Line 254
llvm::formLCSSARecursively(llvm::Loop & L, llvm::DominatorTree & DT, llvm::LoopInfo * LI, llvm::ScalarEvolution * SE) Line 276
llvm::formLCSSARecursively(llvm::Loop & L, llvm::DominatorTree & DT, llvm::LoopInfo * LI, llvm::ScalarEvolution * SE) Line 274
formLCSSAOnAllLoops(llvm::LoopInfo * LI, llvm::DominatorTree & DT, llvm::ScalarEvolution * SE) Line 285
llvm::FPPassManager::runOnFunction(llvm::Function & F) Line 1526
`anonymous namespace'::CGPassManager::RunPassOnSCC(llvm::Pass * P, llvm::CallGraphSCC & CurSCC, llvm::CallGraph & CG, bool & CallGraphUpToDate, bool & DevirtualizedCall) Line 151
`anonymous namespace'::CGPassManager::RunAllPassesOnSCC(llvm::CallGraphSCC & CurSCC, llvm::CallGraph & CG, bool & DevirtualizedCall) Line 419
`anonymous namespace'::CGPassManager::runOnModule(llvm::Module & M) Line 475
`anonymous namespace'::MPPassManager::runOnModule(llvm::Module & M) Line 1603
llvm::legacy::PassManagerImpl::run(llvm::Module & M) Line 1707
```

`L` in https://github.com/llvm-mirror/llvm/blob/release_39/lib/Transforms/Utils/LCSSA.cpp#L77 is null.

gen/dibuilder.cpp
kinke: It has to be something different than the previous 'unspecified' type, as that one seems to have no size, and leads to assertions when declaring a static array of associative arrays.

majnemer:
@kinke I've identified the revision which fixes this bug and proposed it to be merged into the 3.9 release branch: http://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20160718/375427.html

kinke:
@majnemer: Superb, thanks a lot. We're obviously very keen on being fully compatible with the upcoming release as it's the key for our MSVC EH and thus crucial for LDC on Windows.

kinke:
Next try with LLVM master which works for me.

kinke:
@JohanEngelen: Any idea wrt. the `LLVM Profile Note: Set profile file path to "default.profraw" via default setting.` hint for the failing PGO IR tests?

JohanEngelen:
Somehow the PGO output file is not set correctly. Perhaps the profile runtime is out-of-sync with LLVM 3.9, or very recently they changed how to set it up with LLVM 3.9. Investigating.

kinke:
Just note that it's LLVM 4.0 now (yesterday's master).

JohanEngelen:
Ah, wow, thanks for the notice :) Makes more sense, I see something on the maillist about changing behavior.

JohanEngelen:
@kinke profile-rt has changed quite a bit recently for LLVM 4.0. Updated it and restarted the AppVeyor build.

kinke:
Thanks - I hope AppVeyor actually uses the updated master...

kinke:
@JohanEngelen: Thanks, works on my box!

kinke:
So they replaced the function call `__llvm_profile_override_default_filename()` with a global variable holding the output file name, right? Then I guess my workaround https://github.com/ldc-developers/ldc/pull/1633/commits/9e0b2254257bad9d806628932fecbaef16571511 is fine, isn't it?

JohanEngelen:
Yeah that's fine, thanks. No need to test it in IR (if it's not working, we clearly see it in tests errorring).

kinke:
I'd like to see this in v1.1 too, as it allows to build against latest LLVM master.

JohanEngelen:
(it's kind of a huge change list with many irrelevant changes ;)

kinke:
irrelevant => -70 lines due to less duplicated code + improved readability ;)

JohanEngelen:
I didn't mean the changes are not welcome, it's just very annoying to have to wade through all the NFC changes among the actual functional changes. Because commit 2 actually functionally changes something from commit 1, you can also not look at the commits in isolation.

kinke:
Yeah I know, but that's how I do the job, I believe in merciless refactoring. ;) Commit 1 was still buggy, so the subtle change wrt. the delegate debug info (member types now instead of simple types, functions promoted to function pointers) is an important fix caught in the middle of a refactoring orgy. Thanks for wading through the tedious diff. :+1

JohanEngelen:
>  I believe in merciless refactoring.

+1 !

JohanEngelen:
@kinke Is the plan to move back to LLVM 3.9 on AppVeyor once LLVM's release_39 branch is fixed?

majnemer:
I believe the release branch for 3.9 has the fix.

kinke:
If it is fixed, I'd say let's switch back to 3.9, so that we don't have to adapt to LLVM API changes and rebuild the binaries for AppVeyor all the time to keep our CI systems green.

Issue #1621, opened 2016-07-15
LDC+DUB on Windows: folder separator is ignored


redstar:
Executing DUB with LDC2 1.1.0-alpha1 on Windows results in:

    C:\OpenSource\cpuid>dub
    Performing "debug" build using ldc2 for x86_64.
    cpuid 0.2.4+commit.4.g36ec8b4: building configuration "library"...
    Error: module sourcecpuidamd is in file 'sourcecpuidamd.d' which cannot be read
    import path[0] = C:/Temp/ldc2-1.1.0-alpha1-win64-msvc/bin/../include/d/ldc
    import path[1] = C:/Temp/ldc2-1.1.0-alpha1-win64-msvc/bin/../include/d
    import path[2] = source
    ldc2 failed with exit code 1.

Obviously the path is `source\cpuid\amd.d`. DUB display the right path with `-v`:

    Generate target cpuid (staticLibrary C:\OpenSource\cpuid cpuid)
    Performing "debug" build using ldc2 for x86_64.
    Target 'C:\OpenSource\cpuid\.dub\build\library-debug-windows-x86_64-ldc_0-B933F7C743324C2F06A81BFB0EDA156F\cpuid.lib' doesn't exist, need rebuild.
    cpuid 0.2.4+commit.4.g36ec8b4: building configuration "library"...
    ldc2 -lib -of.dub\build\library-debug-windows-x86_64-ldc_0-B933F7C743324C2F06A81BFB0EDA156F\cpuid.lib -d-debug -g -w -oq -od=.dub/obj -d-version=Have_cpuid -Isource source\cpuid\amd.d source\cpuid\common.d source\cpuid\intel.d source\cpuid\report.d source\cpuid\unified.d source\cpuid\x86_any.d
    Error: module sourcecpuidamd is in file 'sourcecpuidamd.d' which cannot be read
    import path[0] = C:/Temp/ldc2-1.1.0-alpha1-win64-msvc/bin/../include/d/ldc
    import path[1] = C:/Temp/ldc2-1.1.0-alpha1-win64-msvc/bin/../include/d
    import path[2] = source
    FAIL .dub\build\library-debug-windows-x86_64-ldc_0-B933F7C743324C2F06A81BFB0EDA156F\ cpuid staticLibrary
    ldc2 failed with exit code 1.

It is yet unclear if this is an DUB or an LDC problem.

redstar:
[...]

PR #1635, opened 2016-07-23
Fix issue #1621

kinke:
LLVM's ParseCommandLineOptions() expands response files automatically, but always uses the TokenizeGNUCommandLine tokenizer, which apparently eats backslashes.

The solution is to pre-expand the response files into the arguments vector on the LDC side, and using TokenizeWindowsCommandLine if LDC is built for Windows.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -44,6 +44,9 @@
 #include "llvm/Support/Host.h"
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/Path.h"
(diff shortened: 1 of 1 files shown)


kinke:
The actual `StringSaver` type differs in LLVM versions < 3.8, so I restricted the fix to LLVM 3.8+. This shouldn't matter as we require at least 3.8 (not sure if 3.8 was just buggy or whether we even need 3.9 for successful compilation) for MSVC anyway.

PR #1637, opened 2016-07-23
Implement @optStrategy UDA.

### LINKED ISSUE #1549: Add UDA for per-function optimization settings
Implement support for `optnone`, `optsize`, `minsize` (which I think are all related so perhaps one UDA can be designed for them?).  `optnone` is the most interesting I think. In a PS4 presentation [1] it is mentioned as a great debugging tool when -O0 builds are not an option; makes total sense!

[1] http://llvm.org/devmtg/2013-11/slides/Robinson-PS4Toolchain.pdf

JohanEngelen:
I need this for debugging so implemented it. Any alternative ideas for the name "optStrat"?

Resolves #1549.

druntime change: https://github.com/ldc-developers/druntime/commit/1a718598bc8bf6ffbc68492544df33fe8558b8a4

### DIFF
--- a/gen/uda.cpp
+++ b/gen/uda.cpp
@@ -17,6 +17,7 @@ namespace {
 namespace attr {
 const std::string llvmAttr = "llvmAttr";
 const std::string llvmFastMathFlag = "llvmFastMathFlag";
+const std::string optStrategy = "optStrategy";
 const std::string section = "section";
 const std::string target = "target";
 const std::string weak = "_weak";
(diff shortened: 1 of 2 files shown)


kinke:
Thanks, makes a lot of sense. I'd prefer `optStrategy` to make clear we're not talking about optimization **strat**osphere or something. ;)

JohanEngelen:
I changed the name to `optStrategy`.

JohanEngelen:
Fixed merge conflict (because of druntime changes).

ready to merge?

gen/uda.cpp
```diff
@@ -147,6 +148,30 @@ void applyAttrLLVMFastMathFlag(StructLiteralExp *sle, IrFunction *irFunc) {
   }
 }
 
+void applyAttrOptStrat(StructLiteralExp *sle, IrFunction *irFunc) {
```
kinke: OptStrategy here as well?

JohanEngelen: done

JohanEngelen:
thanks!
(now on to the real bug then......)

Issue #1634, opened 2016-07-22
Missing codegen for lambda passed as default parameter value


lodo1995:
I have a project that compiles fine on DMD 1.071, but on ldc-1.1.0-alpha1 the linker complains about some undefined references. The missing code is for lambdas that only exist as default parameters to a template function:

```
struct S(T, D)
{
  T t;
  D d;
}

auto makeS(T, D)(T t, D d = (string s) {})  // looks like code for this lambda is never generated
{
  return S!(T, D)(t, d);
}

void main()
{
  auto s = makeS(1);
  s.d("foo");
}
```

I don't have an LDC compiler available right now. I encountered the error on Travis with a very big project. I wrote this small example and tried it on http://ldc.acomirei.ru/ and apparently no code is generated for symbol `pure nothrow @nogc @safe void example.makeS!(int, pure nothrow @nogc @safe void function(immutable(char)[])*).__lambda2(immutable(char)[])`.

lodo1995:
Downloaded and tested locally. I can now 100% confirm that the code above does not link: undefined reference to the lambda function.
I think this is a *MAJOR* issue.

lodo1995:
Ping @kinke @klickverbot 
Sorry if I bother you, but I think this issue deserves some attention (unless I'm missing something; if that's the case, please tell me, and sorry for the noise).

JohanEngelen:
Also occurs for 0.17.1 and 1.0.0.

JohanEngelen:
Simplified test case:
```d
void makeS(void function(string s) d = function(string s) {})  // looks like code for this lambda is never generated
{
}

void main()
{
  makeS();
}
```

We apparently don't codegen default parameter values (lambdas).

kinke:
I had a quick look yesterday. We should probably make sure that lambdas are defined in the current module in `DtoSymbolAddress()`, right after https://github.com/ldc-developers/ldc/blob/master/gen/llvmhelpers.cpp#L1566 (`DtoDefineFunction()`). I firstly don't know how to identify lambdas though and secondly nothing about that `availableExternally` thing, so I don't know how to best tackle this.

JohanEngelen:
I think upon codegen of a FuncDecl, we should check whether there are any default parameter values that need codegen (so no codegen on the callsite when building the argument list, but on the function definition site). Working on it.

kinke:
What about templated types with defaulted template params?

JohanEngelen:
Code for default template arg:
[...]

PR #1641, opened 2016-07-24
Implement codegen of function literals passed as default argument.

JohanEngelen:
Fixes issue #1634

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -1551,6 +1551,17 @@ DValue *DtoSymbolAddress(Loc &loc, Type *type, Declaration *decl) {
     }
   }
 
+  if (FuncLiteralDeclaration *flitdecl = decl->isFuncLiteralDeclaration()) {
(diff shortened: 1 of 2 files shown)


gen/llvmhelpers.cpp
```diff
@@ -1551,6 +1551,17 @@ DValue *DtoSymbolAddress(Loc &loc, Type *type, Declaration *decl) {
     }
   }
 
+  if (FuncLiteralDeclaration *flitdecl = decl->isFuncLiteralDeclaration()) {
+    Logger::println("FuncLiteralDeclaration");
+
+    // We need to codegen the function here, because literals are not added
+    // to the module member list.
+    DtoDefineFunction(flitdecl);
+    assert(getIrFunc(flitdecl)->func);
```
kinke: The DFuncValue ctor makes sure the function isn't null, but the explicit check here doesn't hurt.

klickverbot: I'd actually be in favour of removing it then, but no reason to delay the fix over that.

PR #1644, opened 2016-07-26
[don't pull] Fix debuginfo generation for the local variable frame for a function with nested functions.

JohanEngelen:
Instead of defining debug information for each variable separately, the frame that contains them should get debug info (it is a struct with named variables inside).

Resolves an ICE introduced by PR #1598. See #1598 for more context.

Debug information is still "broken" inside the _nested_ function. That will require more work:
- the frame argument type is `i8*` instead of the proper struct type e.g. `%nest.encloser = type { i32, i32, i32 }`.
- similarly to this PR, debug information should be emitted about the frame _struct_  (perhaps exactly the same info as in the enclosing function), instead of for each variable individually.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -1002,6 +1002,62 @@ ldc::DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *ll,
       );
 }
 
+void ldc::DIBuilder::EmitNestedContextDebugInfo(llvm::Value *frame,
+                                                FuncDeclaration *fd) {
+
+  IrFunction *irFunc = getIrFunc(fd);
+  assert(irFunc);
+  auto frameType = irFunc->frameType;
+
+  ldc::DICompileUnit CU(GetCU());
+  assert(CU && "Compilation unit missing or corrupted");
+  ldc::DIFile file = CreateFile(fd->loc);
+
+  llvm::SmallVector<LLMetadata *, 16> elems;
+  elems.reserve(fd->closureVars.dim);
+  for (auto vd : fd->closureVars) {
+    IrLocal *irLocal = getIrLocal(vd);
+    assert(irLocal);
+    uint64_t offset = gDataLayout->getStructLayout(frameType)->getElementOffset(
+        irLocal->nestedIndex);
+
+    auto dt = CreateMemberType(vd->loc.linnum, vd->type, file, vd->toChars(),
+                               offset, vd->prot().kind);
+    elems.push_back(dt);
+  }
+  auto elemsArray = DBuilder.getOrCreateArray(elems);
+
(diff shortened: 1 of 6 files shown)


gen/dibuilder.cpp
```diff
@@ -1002,6 +1002,62 @@ ldc::DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *ll,
...
+    elems.push_back(dt);
+  }
+  auto elemsArray = DBuilder.getOrCreateArray(elems);
+
+  unsigned alignment =
+      std::max(getABITypeAlign(frameType), irFunc->frameTypeAlignment);
+
+  auto TD =
+      DBuilder.createStructType(CU,             // compile unit where defined
+                                "",             // name
```
JohanEngelen: any ideas for a name here? right now, the struct type shows up as "anonymous struct" in lldb.

gen/dibuilder.cpp
```diff
@@ -1002,6 +1002,62 @@ ldc::DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *ll,
...
+                                fd->loc.linnum, // line number where defined
+                                getTypeAllocSize(frameType) * 8, // size in bits
+                                alignment * 8,           // alignment in bits
+                                DIFlags::FlagArtificial, // flags
+                                getNullDIType(),         // DerivedFrom
+                                elemsArray);
+
+  auto debugVariable =
+      DBuilder.createAutoVariable(GetCurrentScope(),      // scope
+                                  ".frame",               // name
```
JohanEngelen: better ideas for a name here? right now, lldb shows:
`(anonymous struct) .frame = (arg0 = 0, arg1 = 1, enc_n = 0)`

JohanEngelen:
bug: nested nested function's frame debuginfo is not correctly emitted.

gen/dibuilder.cpp
klickverbot: It doesn't seem obvious that exposing the frame struct is the right thing to do in the first place (from an user's perspective).

klickverbot: I had missed the rationale in #1598. If we have to do this due to LLVM inadequacies, can we "alias" the individual members into the list of locals? Is this being done already somehow and I just missed it? I wouldn't worry too much about the frame struct type/name if the user never has to interact with it.

kinke:
The parent frames are referenced via a pointer in the nested frame, right? Nested context hierarchies are one point why I suggested doing the debuginfos properly after a refactoring, as it should be a lot easier afterwards.

JohanEngelen:
This PR just adds complication that we don't want.
https://github.com/ldc-developers/ldc/pull/1646 is a much better fix.

PR #1646, opened 2016-07-27
Fix debuginfo generation for the local variable frame for a function with nested functions

JohanEngelen:
__See #1598 for context.__

Declare debuginfo directly on the GEP instead of on the frame alloca+offset, . An explicit DW_OP_deref is necessary (tested in LLDB).

Resolves an ICE introduced by PR #1598.

### DIFF
--- a/gen/nested.cpp
+++ b/gen/nested.cpp
@@ -507,8 +507,10 @@ void DtoCreateNestedContext(FuncDeclaration *fd) {
 #else
         LLSmallVector<LLValue *, 2> addr;
 #endif
-        gIR->DBuilder.OpOffset(addr, frameType, irLocal->nestedIndex);
-        gIR->DBuilder.EmitLocalVariable(frame, vd, nullptr, false, false, addr);
+        // Because we are passing a GEP instead of an alloca to
+        // llvm.dbg.declare, we have to make the address dereference explicit.
+        gIR->DBuilder.OpDeref(addr);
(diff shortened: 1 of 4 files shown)


Syniurge:
Does it work in GDB too? I'll test tonight if I get the chance.

JohanEngelen:
close+reopen to retriggers the CI system

JohanEngelen:
(i may have rapidly amended the commit and push-forced it, leaving travis/github confused?)

JohanEngelen:
The plot thickens... 
LLVM 3.9: with Deref things work well in LLDB
LLVM 3.5: does __not__ work well in LLDB

JohanEngelen:
So, just for kicks :'( :'( :'''''(, I tested this stuff with LLVM 3.5-3.9.

OK, the plot did not thicken that much. With LLVM 3.5, with or without deref it shows wrong variable contents. LLVM 3.6, 3.7, 3.8, 3.9 are all fine.

kinke:
Thanks for checking. LLVM 3.6+ seems good enough to me.

JohanEngelen:
@Syniurge Did you have time already to test with gdb?

Syniurge:
@JohanEngelen Sorry I took so long, everything's fine in GDB too.

(I feared that with "DW_OP_reg" the addresses of nested variables may not be stable but they are, it's all good.)

PR #1647, opened 2016-07-27
Fix ICE upon second codegen pass of the same SwitchStatement.

JohanEngelen:
Note: this fixes the bug but leaves proper refactoring for later.
We should not store codegen state into the AST and the variables touched in this commit should be moved out of the AST. Added a TODO item for it.

(I have tested several times that the testcase fails without the fix.)

### DIFF
--- a/gen/statements.cpp
+++ b/gen/statements.cpp
@@ -1125,6 +1125,16 @@ public:
 
     llvm::BasicBlock *oldbb = irs->scopebb();
 
+    // Codegen state variables stored in the AST must be reset (see end of
+    // function)
+    for (CaseStatement *cs : *stmt->cases) {
+      assert(cs->bodyBB == nullptr);
+      assert(cs->llvmIdx == nullptr);
+    }
+    if (stmt->sdefault) {
+      assert(stmt->sdefault->bodyBB == nullptr);
+    }
+
     // If one of the case expressions is non-constant, we can't use
     // 'switch' instruction (that can happen because D2 allows to
     // initialize a global variable in a static constructor).
@@ -1357,6 +1367,17 @@ public:
     }
 
     irs->scope() = IRScope(endbb);
+
+    // Reset backend variables to original state (to allow multiple codegen
+    // passes of same ast nodes)
+    // TODO: move the codegen state variables out of the AST.
+    for (CaseStatement *cs : *stmt->cases) {
+      cs->bodyBB = nullptr;
+      cs->llvmIdx = nullptr;
+    }
+    if (stmt->sdefault) {
+      stmt->sdefault->bodyBB = nullptr;
+    }
   }
 
   //////////////////////////////////////////////////////////////////////////
--- a/tests/codegen/inputs/switch_ICE_gh1638_bar.d
+++ b/tests/codegen/inputs/switch_ICE_gh1638_bar.d
@@ -1,17 +1,26 @@
-// Don't make any changes to this file without consulting Github issue 1638 first.
+// Don't make any changes/additions to this file without consulting Github issue 1638 first.
 
 module switch_ICE_gh1638_bar;
 
-import switch_ICE_gh1638;
-import std.conv;
-
-static class F
+struct S(T)
 {
-    static Q[] X;
+    auto fun = (T a) {
+        T r;
+        switch (a)
+        {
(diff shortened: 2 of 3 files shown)


tests/codegen/inputs/switch_ICE_gh1638_bar.d
```diff
@@ -0,0 +1,17 @@
+// Don't make any changes to this file without consulting Github issue 1638 first.
```
klickverbot: Huh?

JohanEngelen: This testcase is very particular and one year from now it may be easy to miss that it is so. So someone may add something to the file, seemingly keeping the test intact. No harm done in making sure it is kept as is. (of course FE code will change and it may no longer trigger the same codegen path...)

kinke:
I'd have a simpler test case:

Module 1:
```D
struct S(T)
{
    auto fun = (T a) {
        T r;
        switch (a)
        {
        case 1:
            r = 1;
            break;
        default:
            return 0;
        }
        return r * 2;
    };
}

void f(int a)
{
    alias T = S!int;
    int r = T().fun(a);
}
```

Module 2:
```D
import <module1>;

int main()
{
    alias T = S!int;
    return T().fun(123);
}
```

Fails with the same `While deleting: label %case - Use still stuck around after Def is destroyed:  br label %case` ICEs on my box with current master.

JohanEngelen:
@kinke OK, I've used your testcase now.

gen/statements.cpp
```diff
@@ -1125,6 +1125,16 @@ class ToIRVisitor : public Visitor {
...
 
+    // Codegen state variables stored in the AST must be reset (see end of
+    // function)
+    for (CaseStatement *cs : *stmt->cases) {
+      assert(cs->bodyBB == nullptr);
+      assert(cs->llvmIdx == nullptr);
+    }
+    if (stmt->sdefault) {
+      assert(stmt->sdefault->bodyBB == nullptr);
+    }
```
kinke: Could you guard this with `#ifndef NDEBUG`?

gen/statements.cpp
```diff
@@ -1357,6 +1367,17 @@ class ToIRVisitor : public Visitor {
     }
 
     irs->scope() = IRScope(endbb);
+
+    // Reset backend variables to original state (to allow multiple codegen
+    // passes of same ast nodes)
+    // TODO: move the codegen state variables out of the AST.
```
kinke: You already know I can be pretty pedantic. ;) How about
```D
// Reset codegen state in AST nodes (to allow for multiple codegen passes).
// TODO: move state to module `ir/irfunction`.
```

tests/codegen/inputs/switch_ICE_gh1638_bar.d
```diff
@@ -0,0 +1,26 @@
+// Don't make any changes/additions to this file without consulting Github issue 1638 first.
```
kinke: I agree with David, I find these notes superfluous and would prefer them gone. IMHO, the new test case doesn't provoke the immediate urge to refactor/simplify the code anymore. :D

gen/statements.cpp
klickverbot: Why not just move the state there right now instead? ;)

gen/statements.cpp
JohanEngelen: Won't it only make the code more verbose?

gen/statements.cpp
JohanEngelen: go ahead and do the removal!

JohanEngelen:
Guys, it is annoying to have to deal with subjective comments about comments. If you don't like them as much, fine, I don't like other comments as much either. Let's not discuss bikesheds.

kinke:
I consider comments just as important as code. Reading the 2 superfluous notes in the test files would leave at least a few guys wondering why it's explicitly there, and reading takes some time (multiply that by the number of times it's going to be read in the future, globally). And my suggestion wrt. the state reset contains a hint to the place suited for such codegen state stacks (and I really dislike `ast` in lowercase, pedantic that I am). Some months ago I didn't know it existed, and first came up with a `static` state hack in the relevant module...

PR #1649, opened 2016-07-27
Fix a few clang warnings [NFC]

JohanEngelen:


ir/irclass.cpp
```diff
@@ -371,8 +371,9 @@ llvm::GlobalVariable *IrAggr::getInterfaceVtbl(BaseClass *b, bool new_instance,
 
       // Create a dummy FuncDeclaration with enough information to satisfy the
       // DIBuilder
-      FuncDeclaration *thunkFd = reinterpret_cast<FuncDeclaration *>(memcpy(
-          new char[sizeof(FuncDeclaration)], fd, sizeof(FuncDeclaration)));
+      FuncDeclaration *thunkFd = reinterpret_cast<FuncDeclaration *>(
+          memcpy(new char[sizeof(FuncDeclaration)], (void *)fd,
+                 sizeof(FuncDeclaration)));
```
kinke: clang really warns about implicitly casting a `FuncDeclaration*` to a `void*`?! That's pedantic.

JohanEngelen: The cast is there to tell Clang that we really want to do this. Where "this" is:
```
../ir/irclass.cpp:375:46: warning: source of this 'memcpy' call is a pointer to dynamic class 'FuncDeclaration'; vtable pointer will be copied [-Wdynamic-class-memaccess]
          new char[sizeof(FuncDeclaration)], fd, sizeof(FuncDeclaration)));
                                             ^
../ir/irclass.cpp:375:46: note: explicitly cast the pointer to silence this warning
          new char[sizeof(FuncDeclaration)], fd, sizeof(FuncDeclaration)));
                                             ^
                                             (void*)
```

kinke: Ah thanks for the heads-up.

PR #1650, opened 2016-07-27
Refactor the call to `llvm::sys::fs::getMainExecutable`.

JohanEngelen:
Taking the address of `main` is UB in C++. Reading LLVM's source, any function address can be used (that's what clang does too).

See Clang's examples/clang-interpreter/main.cpp - Clang C Interpreter Example.

### DIFF
--- a/driver/exe_path.cpp
+++ b/driver/exe_path.cpp
@@ -19,9 +19,15 @@ namespace {
 string exePath;
 }
 
-void exe_path::initialize(const char *arg0, void *mainAddress) {
+void exe_path::initialize(const char *arg0) {
   assert(exePath.empty());
-  exePath = llvm::sys::fs::getMainExecutable(arg0, mainAddress);
+
+  // Some platforms can't implement LLVM's getMainExecutable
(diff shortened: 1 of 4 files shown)


driver/exe_path.cpp
```diff
@@ -19,9 +19,18 @@ namespace {
...
+void exe_path::initialize(const char *arg0) {
   assert(exePath.empty());
-  exePath = llvm::sys::fs::getMainExecutable(arg0, mainAddress);
+
+  // Some platforms can't implement LLVM's getMainExecutable
+  // without being given the address of a function in the main executable.
+  // Thus getMainExecutable needs the address of a function;
+  // any function address in the main executable will do.
+  exePath = llvm::sys::fs::getMainExecutable(
+      arg0, reinterpret_cast<void *>(&functionWithExternalLinkage));
```
kinke: Can't you use `&exe_path::initialize` itself then?

JohanEngelen: lol yes :)

JohanEngelen: fixed.

PR #1653, opened 2016-07-29
[PGO] Add Indirect Call instrumentation and promotion.

JohanEngelen:
Now that LLVM 3.9 is being prepared for release, I am adding my indirect call PGO work.

_Only applies to LLVM >= 3.9. No functional changes intended for LLVM < 3.9._

This PR adds instrumentation code and metadata tagging code for PGO of indirect calls (a call through a function pointer that is unknown at compile-time).
The new PGO functionality is enabled by default but can be disabled with `-disable-pgo-indirect-calls`.
AFAIK this puts us on par with all front-end PGO capabilities of LLVM/Clang 3.9.

Because it is a relatively simple addition to the code (and for the unreleased LLVM 3.9 only), I hope this can still make it into the 1.1.0 release.

Blog article: https://johanengelen.github.io/ldc/2016/04/13/PGO-in-LDC-virtual-calls.html

### DIFF
--- a/gen/pgo.h
+++ b/gen/pgo.h
@@ -82,8 +82,10 @@ public:
     return I;
   }
 
+  void emitIndirectCallPGO(llvm::Instruction *callSite, llvm::Value *funcPtr) {}
+
   void valueProfile(uint32_t valueKind, llvm::Instruction *valueSite,
-                    llvm::Value *valuePtr) {}
+                    llvm::Value *value, bool ptrCastNeeded) {}
 };
 
 #else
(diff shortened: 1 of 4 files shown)


kinke:
Nice article! :+1:

gen/pgo.h
```diff
@@ -158,6 +170,12 @@ class CodeGenPGO {
     return I;
   }
 
+  /// This method either inserts a call to the profile run-time during
+  /// instrumentation or puts profile data into metadata for PGO use.
+  /// Does nothing for LLVM < 3.9.
+  void valueProfile(uint32_t valueKind, llvm::Instruction *valueSite,
```
klickverbot: A more descriptive function name (and maybe even some minimal docs for the parameters * gasp *) would be nice – all this currently tells somebody not familiar with the contents is that the call has something to do with value profiling.

JohanEngelen: I'll add a bit more but, it really is that simple:
`valueProfile` adds profiling instrumentation/annotation. The profiling value is of kind `valueKind`, should be added to IR code site `valueSite`, and the to be profiled value is given by `valuePtr`.

klickverbot: (I inferred as much from the comment. `valueProfile` just sounds very much like a noun to me, although of course both halves could also be verbs.)

JohanEngelen:
>If I understand correctly, value profiling is currently only used for speculative indirect calls? Do you anticipate further uses of the current code for other situations? The current comments and nomenclature seem like a mix between general and specific (e.g. the valueProfile implementation talks about function calls).

LLVM currently only has a valueKind for indirect-call targets. But the code is reusable for other values. The comments should be generalized. Thanks for noticing.

JohanEngelen:
> Do we actually need to disable LDC_WITH_PGO for anything? Could you remove it a follow-up PR otherwise?

Removing will mean: replace with `LLVM_LDC_VER >= 307`. I have not much preference, but for me it's clearer to keep `LDC_WITH_PGO` (also for CMake/lit-testing business it is useful)

gen/pgo.h
JohanEngelen: I guess I should clarify that I intentionally keep the code very similar to Clang's. This makes it much easier to deal with changes to LLVM's way of dealing with things.
`valueProfile` is a Clang name, and could definitely be better.

klickverbot: Okay, makes sense then – I never looked at Clang's implementation of it.

JohanEngelen:
I generalized the value profiling code (non-trivial) and added extra documentation.

Issue #1104, opened 2015-09-21
Fix up basic block order


klickverbot:
When I rewrote the whole cleanup handling code and removed the `end` block from IRScope, I didn't pay much attention to the order the basic blocks would end up in. However, at least for unoptimized builds, this is more or less 1:1 the order the blocks end up in the assembly. We should make sure that e.g. the normal case path for invokes/asserts is contiguous for asserts/invokes such that we don't end up with quite as many jumps.

choikwa:
this is for no-opt build correct? I assume llvm has some sort of basic block reordering.

klickverbot:
It's for unoptimized builds, yes, and also helpful for looking at the compiler-produced IR.

kinke:
Fixed by #1656.

PR #1656, opened 2016-07-29
Revive #1202 (propagate "nothrow" to LLVM IR)

kinke:
This small change has a huge impact, especially when decorating destructors with nothrow.

nothrow functions will always be called (not invoked) from modules compiled with `-release`. If the nothrow callee throws an Error, the Error cannot be caught by the caller anymore, and it won't run any clean-ups during unwind.

Inside modules with enabled assertions, nothrow callees are still invoked, so the caller may catch the Error (needed for stuff like `assertThrown!AssertError()`) and cleans up properly.

nothrow functions still throw any errors the regular way via _d_throw_exception(), regardless of compiling the containing module via `-release` or not. Mixing debug and release modules for linking should be
possible - but Errors thrown by nothrow callees will only be catchable in debug modules, and during unwind, only frames of functions defined in debug modules will run their clean-ups. The latter could be problematic under certain circumstances, but it would at least only matter after an Error after all.


rainers:
See also https://github.com/ldc-developers/ldc/pull/1202.

I don't think nothrow semantics should change between debug or release.

kinke:
> I don't think nothrow semantics should change between debug or release.

I'd have absolutely no problem with that in this regard. I tried to implement what @klickverbot and @deadalnix suggested in #1202.

But the 'huge impact' I was expecting doesn't seem that huge actually.

```D
struct S
{
    ~this() nothrow {}
    void foo() nothrow { throw new Error("foo"); }
}

void main()
{
    S a;
    a.foo();

    {
        S b;
        b.foo();

        S().foo();
    }
}
```

With `-release`:
```LL
define i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #4 comdat {
  %a = alloca %nothrow.S, align 1                 ; [#uses = 3, size/byte = 1]
  %b = alloca %nothrow.S, align 1                 ; [#uses = 3, size/byte = 1]
  %__slS53 = alloca %nothrow.S, align 1           ; [#uses = 3, size/byte = 1]
  %1 = bitcast %nothrow.S* %a to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %1, i8 0, i64 1, i32 1, i1 false)
  call void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %a) #0
  %2 = bitcast %nothrow.S* %b to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %2, i8 0, i64 1, i32 1, i1 false)
  call void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %b) #0
  %3 = bitcast %nothrow.S* %__slS53 to i8*        ; [#uses = 1]
  %.padding = getelementptr inbounds i8, i8* %3, i32 0 ; [#uses = 1, type = i8*]
  call void @llvm.memset.p0i8.i64(i8* %.padding, i8 0, i64 1, i32 1, i1 false)
  call void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %__slS53) #0
  br label %dtor.__slS53

dtor.__slS53:                                     ; preds = %0
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %__slS53) #0
  br label %toElem.success

toElem.success:                                   ; preds = %dtor.__slS53
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %b) #0
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %a) #0
  ret i32 0
}
```

Without `-release`:
```LL
define i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #0 comdat personality i32 (i8*, i8*, i8*, i8*)* @__CxxFrameHandler3 {
  %a = alloca %nothrow.S, align 1                 ; [#uses = 3, size/byte = 1]
  %b = alloca %nothrow.S, align 1                 ; [#uses = 3, size/byte = 1]
  %__slS53 = alloca %nothrow.S, align 1           ; [#uses = 4, size/byte = 1]
  %1 = bitcast %nothrow.S* %a to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %1, i8 0, i64 1, i32 1, i1 false)
  call void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %a) #0
  %2 = bitcast %nothrow.S* %b to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %2, i8 0, i64 1, i32 1, i1 false)
  call void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %b) #0
  %3 = bitcast %nothrow.S* %__slS53 to i8*        ; [#uses = 1]
  %.padding = getelementptr inbounds i8, i8* %3, i32 0 ; [#uses = 1, type = i8*]
  call void @llvm.memset.p0i8.i64(i8* %.padding, i8 0, i64 1, i32 1, i1 false)
  invoke void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %__slS53) #0
          to label %postinvoke unwind label %cleanuppad

dtor.__slS53:                                     ; preds = %postinvoke
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %__slS53) #0
  br label %toElem.success

postinvoke:                                       ; preds = %0
  br label %dtor.__slS53

cleanuppad:                                       ; preds = %0
  %4 = cleanuppad within none []                  ; [#uses = 4]
  %5 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %4) ] ; [#uses = 1]
  br i1 %5, label %dtor.__slS531, label %cleanupret

dtor.__slS531:                                    ; preds = %cleanuppad
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %__slS53) #0 [ "funclet"(token %4) ]
  br label %cleanupret

cleanupret:                                       ; preds = %cleanuppad, %dtor.__slS531
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %4) ]
  cleanupret from %4 unwind to caller

toElem.success:                                   ; preds = %dtor.__slS53
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %b) #0
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %a) #0
  ret i32 0
}
```

So there's a single `invoke` in the debug version. I was expecting a lot more. So clean-ups aren't always run in debug modules when a nothrow callee throws an Error. No idea what's really going on here.

kinke:
I was expecting this ugly IR with enabled assertions (generated by omitting `nothrow` in the D source):
```LL
define i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #0 comdat personality i32 (i8*, i8*, i8*, i8*)* @__CxxFrameHandler3 {
  %a = alloca %nothrow.S, align 1                 ; [#uses = 4, size/byte = 1]
  %b = alloca %nothrow.S, align 1                 ; [#uses = 4, size/byte = 1]
  %__slS53 = alloca %nothrow.S, align 1           ; [#uses = 4, size/byte = 1]
  %1 = bitcast %nothrow.S* %a to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %1, i8 0, i64 1, i32 1, i1 false)
  invoke void @_D7nothrow1S3fooMFZv(%nothrow.S* nonnull %a) #0
          to label %postinvoke unwind label %cleanuppad

finally:                                          ; preds = %try.success
  call void @_D7nothrow1S6__dtorMFZv(%nothrow.S* nonnull %a) #0
  br label %try.success15

postinvoke:                                       ; preds = %0
  %2 = bitcast %nothrow.S* %b to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %2, i8 0, i64 1, i32 1, i1 false)
  invoke void @_D7nothrow1S3fooMFZv(%nothrow.S* nonnull %b) #0
          to label %postinvoke8 unwind label %cleanuppad4

postinvoke3:                                      ; preds = %finally2
  br label %try.success

cleanuppad:                                       ; preds = %finally26, %cleanupret5, %finally2, %0
  %3 = cleanuppad within none []                  ; [#uses = 4]
  %4 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %3) ] ; [#uses = 1]
  br i1 %4, label %finally1, label %cleanupret

finally1:                                         ; preds = %cleanuppad
  call void @_D7nothrow1S6__dtorMFZv(%nothrow.S* nonnull %a) #0 [ "funclet"(token %3) ]
  br label %cleanupret

cleanupret:                                       ; preds = %cleanuppad, %finally1
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %3) ]
  cleanupret from %3 unwind to caller

finally2:                                         ; preds = %toElem.success
  invoke void @_D7nothrow1S6__dtorMFZv(%nothrow.S* nonnull %b) #0
          to label %postinvoke3 unwind label %cleanuppad

postinvoke8:                                      ; preds = %postinvoke
  %5 = bitcast %nothrow.S* %__slS53 to i8*        ; [#uses = 1]
  %.padding = getelementptr inbounds i8, i8* %5, i32 0 ; [#uses = 1, type = i8*]
  call void @llvm.memset.p0i8.i64(i8* %.padding, i8 0, i64 1, i32 1, i1 false)
  invoke void @_D7nothrow1S3fooMFZv(%nothrow.S* nonnull %__slS53) #0
          to label %postinvoke14 unwind label %cleanuppad10

postinvoke9:                                      ; preds = %dtor.__slS53
  br label %toElem.success

cleanuppad4:                                      ; preds = %dtor.__slS5312, %cleanupret11, %dtor.__slS53, %postinvoke
  %6 = cleanuppad within none []                  ; [#uses = 4]
  %7 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %6) ] ; [#uses = 1]
  br i1 %7, label %finally26, label %cleanupret5

finally26:                                        ; preds = %cleanuppad4
  invoke void @_D7nothrow1S6__dtorMFZv(%nothrow.S* nonnull %b) #0 [ "funclet"(token %6) ]
          to label %postinvoke37 unwind label %cleanuppad

postinvoke37:                                     ; preds = %finally26
  br label %cleanupret5

cleanupret5:                                      ; preds = %cleanuppad4, %postinvoke37
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %6) ]
  cleanupret from %6 unwind label %cleanuppad

dtor.__slS53:                                     ; preds = %postinvoke14
  invoke void @_D7nothrow1S6__dtorMFZv(%nothrow.S* nonnull %__slS53) #0
          to label %postinvoke9 unwind label %cleanuppad4

postinvoke14:                                     ; preds = %postinvoke8
  br label %dtor.__slS53

cleanuppad10:                                     ; preds = %postinvoke8
  %8 = cleanuppad within none []                  ; [#uses = 4]
  %9 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %8) ] ; [#uses = 1]
  br i1 %9, label %dtor.__slS5312, label %cleanupret11

dtor.__slS5312:                                   ; preds = %cleanuppad10
  invoke void @_D7nothrow1S6__dtorMFZv(%nothrow.S* nonnull %__slS53) #0 [ "funclet"(token %8) ]
          to label %postinvoke913 unwind label %cleanuppad4

postinvoke913:                                    ; preds = %dtor.__slS5312
  br label %cleanupret11

cleanupret11:                                     ; preds = %cleanuppad10, %postinvoke913
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %8) ]
  cleanupret from %8 unwind label %cleanuppad4

toElem.success:                                   ; preds = %postinvoke9
  br label %finally2

try.success:                                      ; preds = %postinvoke3
  br label %finally

try.success15:                                    ; preds = %finally
  ret i32 0
}
```

kinke:
This front-end optimization apparently quickly isn't applied anymore:
```D
struct S
{
    ~this() nothrow {}
    void foo() nothrow { throw new Error("foo"); }
}

struct Throwing
{
    ~this() {}
    void bar() { throw new Exception("bar"); }
}

void main()
{
    Throwing t;

    S a;
    a.foo();
    t.bar();

    {
        S b;
        t.bar();
        b.foo();

        S().foo();
    }
}
```

Release:
```LL
define i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #4 comdat personality i32 (i8*, i8*, i8*, i8*)* @__CxxFrameHandler3 {
  %t = alloca %nothrow.Throwing, align 1          ; [#uses = 5, size/byte = 1]
  %a = alloca %nothrow.S, align 1                 ; [#uses = 4, size/byte = 1]
  %b = alloca %nothrow.S, align 1                 ; [#uses = 4, size/byte = 1]
  %__slS54 = alloca %nothrow.S, align 1           ; [#uses = 3, size/byte = 1]
  %1 = bitcast %nothrow.Throwing* %t to i8*       ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %1, i8 0, i64 1, i32 1, i1 false)
  %2 = bitcast %nothrow.S* %a to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %2, i8 0, i64 1, i32 1, i1 false)
  call void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %a) #0
  invoke void @_D7nothrow8Throwing3barMFZv(%nothrow.Throwing* nonnull %t) #4
          to label %postinvoke unwind label %cleanuppad3

finally:                                          ; preds = %try.success11
  call void @_D7nothrow8Throwing6__dtorMFZv(%nothrow.Throwing* nonnull %t) #4
  br label %try.success12

finally1:                                         ; preds = %try.success
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %a) #0
  br label %try.success11

cleanuppad:                                       ; preds = %cleanupret4
  %3 = cleanuppad within none []                  ; [#uses = 4]
  %4 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %3) ] ; [#uses = 1]
  br i1 %4, label %finally2, label %cleanupret

finally2:                                         ; preds = %cleanuppad
  call void @_D7nothrow8Throwing6__dtorMFZv(%nothrow.Throwing* nonnull %t) #4 [ "funclet"(token %3) ]
  br label %cleanupret

cleanupret:                                       ; preds = %cleanuppad, %finally2
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %3) ]
  cleanupret from %3 unwind to caller

postinvoke:                                       ; preds = %0
  %5 = bitcast %nothrow.S* %b to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %5, i8 0, i64 1, i32 1, i1 false)
  invoke void @_D7nothrow8Throwing3barMFZv(%nothrow.Throwing* nonnull %t) #4
          to label %postinvoke10 unwind label %cleanuppad7

cleanuppad3:                                      ; preds = %cleanupret8, %0
  %6 = cleanuppad within none []                  ; [#uses = 4]
  %7 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %6) ] ; [#uses = 1]
  br i1 %7, label %finally15, label %cleanupret4

finally15:                                        ; preds = %cleanuppad3
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %a) #0 [ "funclet"(token %6) ]
  br label %cleanupret4

cleanupret4:                                      ; preds = %cleanuppad3, %finally15
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %6) ]
  cleanupret from %6 unwind label %cleanuppad

finally6:                                         ; preds = %toElem.success
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %b) #0
  br label %try.success

postinvoke10:                                     ; preds = %postinvoke
  call void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %b) #0
  %8 = bitcast %nothrow.S* %__slS54 to i8*        ; [#uses = 1]
  %.padding = getelementptr inbounds i8, i8* %8, i32 0 ; [#uses = 1, type = i8*]
  call void @llvm.memset.p0i8.i64(i8* %.padding, i8 0, i64 1, i32 1, i1 false)
  call void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %__slS54) #0
  br label %dtor.__slS54

cleanuppad7:                                      ; preds = %postinvoke
  %9 = cleanuppad within none []                  ; [#uses = 4]
  %10 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %9) ] ; [#uses = 1]
  br i1 %10, label %finally69, label %cleanupret8

finally69:                                        ; preds = %cleanuppad7
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %b) #0 [ "funclet"(token %9) ]
  br label %cleanupret8

cleanupret8:                                      ; preds = %cleanuppad7, %finally69
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %9) ]
  cleanupret from %9 unwind label %cleanuppad3

dtor.__slS54:                                     ; preds = %postinvoke10
  call void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %__slS54) #0
  br label %toElem.success

toElem.success:                                   ; preds = %dtor.__slS54
  br label %finally6

try.success:                                      ; preds = %finally6
  br label %finally1

try.success11:                                    ; preds = %finally1
  br label %finally

try.success12:                                    ; preds = %finally
  ret i32 0
}
```

Debug:
```LL
define i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #0 comdat personality i32 (i8*, i8*, i8*, i8*)* @__CxxFrameHandler3 {
  %t = alloca %nothrow.Throwing, align 1          ; [#uses = 5, size/byte = 1]
  %a = alloca %nothrow.S, align 1                 ; [#uses = 4, size/byte = 1]
  %b = alloca %nothrow.S, align 1                 ; [#uses = 4, size/byte = 1]
  %__slS54 = alloca %nothrow.S, align 1           ; [#uses = 4, size/byte = 1]
  %1 = bitcast %nothrow.Throwing* %t to i8*       ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %1, i8 0, i64 1, i32 1, i1 false)
  %2 = bitcast %nothrow.S* %a to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %2, i8 0, i64 1, i32 1, i1 false)
  invoke void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %a) #0
          to label %postinvoke7 unwind label %cleanuppad3

finally:                                          ; preds = %try.success23
  call void @_D7nothrow8Throwing6__dtorMFZv(%nothrow.Throwing* nonnull %t) #0
  br label %try.success24

finally1:                                         ; preds = %try.success
  invoke void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %a) #0
          to label %postinvoke unwind label %cleanuppad

postinvoke:                                       ; preds = %finally1
  br label %try.success23

cleanuppad:                                       ; preds = %finally15, %cleanupret4, %finally1
  %3 = cleanuppad within none []                  ; [#uses = 4]
  %4 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %3) ] ; [#uses = 1]
  br i1 %4, label %finally2, label %cleanupret

finally2:                                         ; preds = %cleanuppad
  call void @_D7nothrow8Throwing6__dtorMFZv(%nothrow.Throwing* nonnull %t) #0 [ "funclet"(token %3) ]
  br label %cleanupret

cleanupret:                                       ; preds = %cleanuppad, %finally2
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %3) ]
  cleanupret from %3 unwind to caller

postinvoke7:                                      ; preds = %0
  invoke void @_D7nothrow8Throwing3barMFZv(%nothrow.Throwing* nonnull %t) #0
          to label %postinvoke8 unwind label %cleanuppad3

postinvoke8:                                      ; preds = %postinvoke7
  %5 = bitcast %nothrow.S* %b to i8*              ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %5, i8 0, i64 1, i32 1, i1 false)
  invoke void @_D7nothrow8Throwing3barMFZv(%nothrow.Throwing* nonnull %t) #0
          to label %postinvoke15 unwind label %cleanuppad11

postinvoke10:                                     ; preds = %finally9
  br label %try.success

cleanuppad3:                                      ; preds = %finally913, %cleanupret12, %finally9, %postinvoke7, %0
  %6 = cleanuppad within none []                  ; [#uses = 4]
  %7 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %6) ] ; [#uses = 1]
  br i1 %7, label %finally15, label %cleanupret4

finally15:                                        ; preds = %cleanuppad3
  invoke void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %a) #0 [ "funclet"(token %6) ]
          to label %postinvoke6 unwind label %cleanuppad

postinvoke6:                                      ; preds = %finally15
  br label %cleanupret4

cleanupret4:                                      ; preds = %cleanuppad3, %postinvoke6
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %6) ]
  cleanupret from %6 unwind label %cleanuppad

finally9:                                         ; preds = %toElem.success
  invoke void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %b) #0
          to label %postinvoke10 unwind label %cleanuppad3

postinvoke15:                                     ; preds = %postinvoke8
  invoke void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %b) #0
          to label %postinvoke16 unwind label %cleanuppad11

postinvoke16:                                     ; preds = %postinvoke15
  %8 = bitcast %nothrow.S* %__slS54 to i8*        ; [#uses = 1]
  %.padding = getelementptr inbounds i8, i8* %8, i32 0 ; [#uses = 1, type = i8*]
  call void @llvm.memset.p0i8.i64(i8* %.padding, i8 0, i64 1, i32 1, i1 false)
  invoke void @_D7nothrow1S3fooMFNbZv(%nothrow.S* nonnull %__slS54) #0
          to label %postinvoke22 unwind label %cleanuppad18

postinvoke17:                                     ; preds = %dtor.__slS54
  br label %toElem.success

cleanuppad11:                                     ; preds = %dtor.__slS5420, %cleanupret19, %dtor.__slS54, %postinvoke15, %postinvoke8
  %9 = cleanuppad within none []                  ; [#uses = 4]
  %10 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %9) ] ; [#uses = 1]
  br i1 %10, label %finally913, label %cleanupret12

finally913:                                       ; preds = %cleanuppad11
  invoke void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %b) #0 [ "funclet"(token %9) ]
          to label %postinvoke1014 unwind label %cleanuppad3

postinvoke1014:                                   ; preds = %finally913
  br label %cleanupret12

cleanupret12:                                     ; preds = %cleanuppad11, %postinvoke1014
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %9) ]
  cleanupret from %9 unwind label %cleanuppad3

dtor.__slS54:                                     ; preds = %postinvoke22
  invoke void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %__slS54) #0
          to label %postinvoke17 unwind label %cleanuppad11

postinvoke22:                                     ; preds = %postinvoke16
  br label %dtor.__slS54

cleanuppad18:                                     ; preds = %postinvoke16
  %11 = cleanuppad within none []                 ; [#uses = 4]
  %12 = call i1 @_d_enter_cleanup(i8* null) [ "funclet"(token %11) ] ; [#uses = 1]
  br i1 %12, label %dtor.__slS5420, label %cleanupret19

dtor.__slS5420:                                   ; preds = %cleanuppad18
  invoke void @_D7nothrow1S6__dtorMFNbZv(%nothrow.S* nonnull %__slS54) #0 [ "funclet"(token %11) ]
          to label %postinvoke1721 unwind label %cleanuppad11

postinvoke1721:                                   ; preds = %dtor.__slS5420
  br label %cleanupret19

cleanupret19:                                     ; preds = %cleanuppad18, %postinvoke1721
  call void @_d_leave_cleanup(i8* null) [ "funclet"(token %11) ]
  cleanupret from %11 unwind label %cleanuppad11

toElem.success:                                   ; preds = %postinvoke17
  br label %finally9

try.success:                                      ; preds = %postinvoke10
  br label %finally1

try.success23:                                    ; preds = %postinvoke
  br label %finally

try.success24:                                    ; preds = %finally
  ret i32 0
}
```

rainers:
> I tried to implement what @klickverbot and @deadalnix suggested in #1202.

Ok, but I don't think this is the right direction. To allow catching Errors of nothrow functions, my PR adds the "NoUnwind" attribute to the call sites (not the function), unless the call is inside a catch statement.

kinke:
I know, but I think this is the right direction. I don't see why catching Errors would make sense and justify the overhead for release builds. Restricting nothrow-invokes for debug builds to try/catch blocks may be worth thinking about IMO.

deadalnix:
To relax what @rainers said, maybe keeping the invoke in try/catch block that DO catch some form of error is a good thing. For others, just ignore an Error may be thrown.

This would keep thing like unittests being able to check for assertions without compromising 99% of the code. Alternatively, throwing Error can be done via some sort of custom SJLJ in release mode.

deadalnix:
Also, these windows exception handling IR are so aweful :/

JohanEngelen:
> keeping the invoke in try/catch block that DO catch some form of error is a good thing. For others, just ignore an Error may be thrown.

Is an `Error` supposed to be recoverable? If so, the proposal won't work because dtors will not be run correctly in functions called by functions called by try/catch(Error), right?
Perhaps LLVM already does a nothrow analysis, so we should check if the IR is already optimized when function implementations are available in the IR (at -O3, we emit all (small) functions called from other modules to do cross-module inlining).

kinke:
> Is an Error supposed to be recoverable?

Most likely not really, as the clean-ups aren't always run, as the IR in https://github.com/ldc-developers/ldc/pull/1656#issuecomment-236356853 clearly shows.

kinke:
I've added an IR test.

It currently fails on Windows because MSVC EH doesn't use the `ScopeStack::catchScopes` stack; it pushes clean-ups instead. @rainers knows all about this. :)

kinke:
Squashed & tests slightly extended. The new commit msg should be clear enough. Looking forward to seeing this in LDC 1.1 final.

kinke:
Given the front-end's existing nothrow optimizations, I wouldn't expect any big surprises due to more skipped clean-ups after an Error. Relying on proper clean-up must have been broken at least since this existing optimization.

rainers:
> It currently fails on Windows because MSVC EH doesn't use the ScopeStack::catchScopes stack; it pushes clean-ups instead. @rainers knows all about this. :)

Yeah, maybe not the cleanest part of the MSVC implementation. PR #1202 predates work on the exception handling, though.

rainers:
> maybe keeping the invoke in try/catch block that DO catch some form of error is a good thing.

As mentioned in the other PR, I think this should be a valid optimization. Unfortunately, some lowerings use Throwables instead of Exceptions, e.g. scope(...).

kinke:
Further optimization for try blocks only catching Exceptions added.

ir/irfunction.h
```diff
@@ -388,6 +399,9 @@ class ScopeStack {
   ///
   std::vector<CatchScope> catchScopes;
 
+  ///
+  std::vector<bool> catchingNonExceptions;
```
klickverbot: I'd probably move this out of ScopeStack. The latter is one of the newest pieces of LDC, yet is very close to becoming yet another god object already. I was unhappy with the size of the abstraction already when I originally wrote it, but it seems to keep accumulating stuff things. Some of them are unavoidable due to the new IR representation of (MSVC) EH, but e.g. `catchingNonExceptions` is more or less not coupled to the core functionality (`cleanupScopes`, …) at all. I'd rather keep it outside for now and make the callers that care about `isNothrow` explicitly pass it in.

klickverbot: (Yes, after just looking at the header file I realise that the doc comment for ScopeStack makes it sound like it was designed to be a kitchen sink. That was just laziness in writing the comment, though.)

kinke: If it wasn't for MSVC EH, I would have extended each `CatchScope` by a bool indicating whether it catches non-Exceptions (and `isCatchingNonExceptions()` would simply check if there's such a `CatchScope`). But as MSVC doesn't use `catchScopes` at all (IIRC; it may store a dummy one though, not sure), that wasn't an option, so I went with an explicit stack of try-blocks, currently only represented by this bool.

kinke:
Adapted to new `FuncGenState` and extracted a `TryCatchScope(s)` abstraction out of it, encapsulating catch bodies emission and landing pad emission. It's tightly coupled to the scope/cleanups stack (especially for MSVC EH), so there's a friend relationship between `TryCatchScopes` and `ScopeStack`.

kinke:
Oh there is anoter change: the catch bodies will now be emitted in normal order (used to be reversed, last catch block first). I guess that's why `tests/PGO/exceptions.d` fails. The PGO `matchWeights` will also differ as `uncaughtCount` is now also iteratively increased in normal order (pinging @JohanEngelen).

gen/trycatch.h
```diff
@@ -0,0 +1,88 @@
...
+  void emitCatchBodiesMSVC(IRState &irs);
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+class TryCatchScopes {
+public:
+  TryCatchScopes(IRState &irs) : irs(irs) {}
+
+  void push_back(TryCatchStatement *stmt, llvm::BasicBlock *endbb);
```
klickverbot: Naming is of course a subjective thing, but I really don't like using `push_back`/`pop_back` here. It's a stack anyway, so no need for the `back`, but what I find more confusing reading the code is that it doesn't just `push_back` something onto a STL container, but actually emits code.

kinke: I'll switch back to push/pop once the PGO test is green. Wrt. catch bodies emission, the idea was to defer their emission until the first landing pad is created, to avoid unnecessary emissions. This isn't required for this trivial snippet here though:
```D
void foo() nothrow {}

int main()
{
    try
    {
        foo();
    }
    catch (Exception)
    {
        return 1;
    }
    return 0;
}
```
Apparently there's no `TryCatchStatement` in the AST, so I'm not sure deferring would pay off.

kinke:
The last commit is a total revision of all places where we generate new basic blocks, trying to give them a more human-readable (and potentially cache-friendly) order. It includes a few significant yet subtle changes which make for a very much improved IR (based on `tests/PGO/exceptions.d`).
One significant change is that the catch handlers will be emitted right before the try-catch continuation block ('try.success.or.caught'), i.e., usually a lot further down. And generally, the normal control flow in a D code block (successful invokes/assertions/bound checks) is now also reflected accordingly in the IR, and sequences of related basic blocks aren't torn apart anymore.

kinke:
It should fix https://github.com/ldc-developers/ldc/issues/1104.

kinke:
The PGO branch_weights don't match up anymore for `tests/PGO/exception.d`. I'm pretty sure it's because the code emission (and hence PGO counter increment) order doesn't correspond with the final basic block order anymore, so the variable numbering `![[TC<N>]]` is probably different. @JohanEngelen: Any idea how to get around this?

kinke:
> Could you quickly summarise what the relationship between TryCatchScopes and ScopeStack is

`TryCatchScopes` is the `TryCatchStatement`-only stack, replacing the previous flat list of `ScopeStack::catchBlocks`. As its own entity, it also handles emission of the catch bodies, which was previously in `visit(TryCatchStatement*)`. The landing pad emission was in `ScopeStack`, so the MSVC EH special cases were in 2 modules. That's why I moved landing pad emission to `TryCatchScopes`. The price for that is that it requires direct access to the cleanups stack, for both landing pad emission and even more so for MSVC EH.
The try/catch and cleanup stacks are tightly coupled, so having both in a central place was/is a good idea. But `ScopeStack` is already too big IMO. It may pay off to extend `TryCatchScopes` to `TryCatchFinallyScopes`, managing the cleanups too (and the landing pads stack).

kinke:
I see now. ;) Well, I'll probably end up putting `TryCatchScopes` back in, but give `ScopeStack` the dedicated module it deserves. ;)

kinke:
I'll see whether there's some 'core' clean-up logic to be moved to `trycatch.h`, i.e., extending `TryCatchScopes` to `TryCatchFinallyScopes` by just a little more than a cleanups stack and landing pad caching. Primary goal would be isolating all (MSVC) EH special cases to a single module; later we could take this further and use polymorphism for MSVC.
`ScopeStack` would then build on this `TryCatchFinallyScopes` and thus loose considerable weight.

JohanEngelen:
> The PGO branch_weights don't match up anymore for tests/PGO/exception.d. I'm pretty sure it's because the code emission (and hence PGO counter increment) order doesn't correspond to the final basic block order anymore, so the variable numbering ![[TC<N>]] is probably different. @JohanEngelen: Any idea how to get around this?

The order (in time, not in IR) in which the code is emitted does not matter for the PGO code. The counter indices are assigned based on AST ordering, and are stored in a map that maps AST statement nodes to counter index. If you didn't change any control flow logic in the IR (e.g. changing an "if true" into an "if false"), then the branch weights should still be the same as they were, but in a different order. So then you have to find out which branches in the new IR layout correspond to which branches in the old IR, and adjust the test. It looks like you already did that though. Perhaps a small mistake was made?
(the order of `-DAG` checks does not matter btw)

JohanEngelen:
(the branch_weights metadata has an offset of 1, so `!{!"branch_weights", i32 4, i32 1}` means the branch was taken 3 times, and not taken 0 times)

kinke:
I've reworked it and came up with `ScopeStack -> JumpTargets` and `TryCatchScopes -> TryCatchFinallyScopes`, both members of `FuncGenState`. I'm quite happy with this design. Of course there's still much room for improvement, especially wrt. MSVC, but this separation should be seen as a first step to figure out where MSVC diverges and how to model this appropriately, which is out of scope of this PR.
The new BB order is much more important for LDC users. If we don't want to introduce the optimization wrt. additional skipped cleanups when a nothrow function throws an Error in v1.1 yet, I can temporarily disable it.

kinke:
@JohanEngelen: Thanks, it was the order of the PROFUSE clauses setting the `TC<N>` variables which needed to be adjusted.

> the branch_weights metadata has an offset of 1

Yes, I figured. ;) Is that deliberate?

gen/trycatchfinally.h
```diff
@@ -0,0 +1,282 @@
...
+  void popTryCatch();
+
+  /// Indicates whether there are any active catch blocks that handle
+  /// non-Exception Throwables.
+  bool isCatchingNonExceptions() const;
+
+  /// Registers a piece of cleanup code to be run.
+  ///
+  /// The end block is expected not to contain a terminator yet. It will be
+  /// added by ScopeStack as needed, based on what follow-up blocks code from
```
klickverbot: Still says ScopeStack.

gen/trycatchfinally.h
```diff
@@ -0,0 +1,282 @@
...
+    llvm::MDNode *branchWeights;
+  };
+
+  TryCatchScope(TryCatchStatement *stmt, llvm::BasicBlock *endbb,
+                CleanupCursor cleanupScope);
+
+  CleanupCursor getCleanupScope() const { return cleanupScope; }
+  bool isCatchingNonExceptions() const { return catchesNonExceptions; }
+
+  /// Emits the catch bodies. Must be called only once, and this scope must not
```
klickverbot: Why not just call this from the constructor or a factory function then?

gen/funcgenstate.h
```diff
@@ -65,218 +59,10 @@ struct JumpTarget {
...
-/// This includes finally blocks (which are also generated by the frontend for
-/// running the destructors of non-temporary variables) and the destructors of
-/// temporaries (which are unfortunately not lowered by the frontend).
-///
-/// Our goal is to only emit each cleanup once such as to avoid generating an
-/// exponential number of basic blocks/landing pads for handling all the
-/// different ways of exiting a deeply nested scope (consider e.g. ten
-/// local variables with destructors, each of which might throw itself).
-class CleanupScope {
+/// Keeps track of labels for goto/break/continue.
```
klickverbot: Not only labels, but also implicit targets created by loop statements.

gen/funcgenstate.h
```diff
@@ -507,7 +196,9 @@ class FuncGenState {
   IrFunction &irFunc;
 
   /// The stack of scopes inside the function.
```
klickverbot: That comment is now even less useful.

gen/trycatchfinally.h
```diff
@@ -0,0 +1,282 @@
...
+
+  /// Returns a cursor that identifies the current cleanup scope, to be later
+  /// used with #runCleanups() et al.
+  ///
+  /// Note that this cursor is only valid as long as the current scope is not
+  /// popped.
+  CleanupCursor currentCleanupScope() const { return cleanupScopes.size(); }
+
+  /// Gets the unresolved gotos for the current cleanup scope.
+  std::vector<GotoJump> &currentUnresolvedGotos();
```
klickverbot: Exposing the whole vector here seems a bit iffy in terms of encapsulation – I'd go for two methods, one for pushing new unresolved jumps, and another one called with identifier and target block to resolve any potentially in-flight jumps. That is, just move the implementations from JumpTargets here. Otherwise, GotoJump handling seems weirdly split between the two.

JohanEngelen:
>> the branch_weights metadata has an offset of 1

> Yes, I figured. ;) Is that deliberate?

Yes. That's what Clang/LLVM used to do. IIRC from LLVM 3.9, the offset is no longer necessary (LLVM can handle the "0" probability not really being zero).

kinke:
Thanks David, good observations; I tried to incorporate all suggestions.

gen/trycatchfinally.h
```diff
@@ -0,0 +1,301 @@
...
+
+  /// Returns the stack slot that contains the exception object pointer while a
+  /// landing pad is active, lazily creating it as needed.
+  ///
+  /// This value must dominate all uses; first storing it, and then loading it
+  /// when calling _d_eh_resume_unwind. If we take a select at the end of any
+  /// cleanups on the way to the latter, the value must also dominate all other
+  /// predecessors of the cleanup. Thus, we just use a single alloca in the
+  /// entry BB of the function.
+  llvm::AllocaInst *getOrCreateEhPtrSlot();
```
klickverbot: This could be made private by passing the slot into TryCatchScope.

gen/trycatchfinally.h
```diff
@@ -0,0 +1,301 @@
...
+  void popCleanups(CleanupCursor targetScope);
+
+  /// Returns a cursor that identifies the current cleanup scope, to be later
+  /// used with #runCleanups() et al.
+  ///
+  /// Note that this cursor is only valid as long as the current scope is not
+  /// popped.
+  CleanupCursor currentCleanupScope() const { return cleanupScopes.size(); }
+
+  void pushUnresolvedGoto(Loc loc, Identifier *labelName);
```
klickverbot: Some final nitpicks (I promise I'm running out of them^^):
 - Documentation.
 - `push` is used here with a different meaning than in the other `push…()` members, which might be a bit confusing (especially without further documentation^^). Maybe `runUnresolvedGoto()` as in `runCleanups()`, since both terminate the current BB? (Feel free to rename both, `run` is not ideal.)

kinke:
Final nitpixx (good ones, I didn't really like `push` there either) addressed. Refactoring journey suspended. ^^

kinke:
No worries, thx for merging. You'll soon all enjoy the reordered IR. :)

Issue #1190, opened 2015-11-03
master: GDB test failures


klickverbot:
If you look at a Travis log for the currently red master, you'll see that there are some failing tests that use a GDB script to check for debug info generation. Some of these fail with "unknown symbol" error messages. However, I can't reproduce those on an Ubuntu 14.04 box with
```
gdb --version
GNU gdb (Ubuntu 7.7.1-0ubuntu5~14.04.2) 7.7.1
```

Does anybody else have more luck?

kinke:
Same gdb as you:

`dmd-testsuite-debug`:
* `runnable/gdb14313.d`: expected regex `.*4000065002`, actual `$1 = -294902294`

`dmd-testsuite`:
* `runnable/gdb14276.d`: expected `.*1234`, actual `Error in sourced command file: value has been optimized out` ;)
* `runnable/gdb14330.d`: expected `.*something`, actual `$1 = <error reading variable>`
* `runnable/gdb4149.d`: expected `.*33`, actual `$1 = 32`

klickverbot:
The following tests are now temporarily disabled on master (some already have been for quite some time):

```
DISABLED_TESTS += gdb1
DISABLED_TESTS += gdb4149
DISABLED_TESTS += gdb4181
DISABLED_TESTS += gdb10311
DISABLED_TESTS += gdb14225
DISABLED_TESTS += gdb14276
DISABLED_TESTS += gdb14313
DISABLED_TESTS += gdb14330
```

Somebody needs to sit down and spend some time on debug info soon. Now that we only support 3.5+ this should be quite a bit easier.

kinke:
With gdb 7.11.1, LDC master, LLVM 3.9 and Linux x64, all gdb tests pass when performed manually EXCEPT for gdb10311 (captured local showing up as `<optimized out>`).

kinke:
Inside dmd-testsuite-debug, they all fail though, due to:
```No line <N> in the current file.
Make breakpoint pending on future shared library load? (y or [n]) [answered N; input not from terminal]```

If I execute the two commands in the `tests/d2/dmd-testsuite` directory (compile + gdb batch run), it all works:
```
Breakpoint 1 at 0x401e44: file gdb10311.d, line 19.
```

klickverbot:
Maybe an issue with absolute/relative source paths?

rainers:
> No line <N> in the current file.

[...]

PR #1658, opened 2016-07-30
dmd-testsuite: Enable GDB tests to see which ones still fail

kinke:
See https://github.com/ldc-developers/ldc/issues/1190.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,6 +90,7 @@ script:
       make -j3 druntime-ldc-unittest-debug druntime-ldc-unittest;
     fi
   # Run dmd-testsuite.
+  - gdb --version
   -
     if [ "${TEST_CONFIG}" = "Debug" ]; then
       CC="" MAKEOPTS=-j4 ctest --verbose -R "dmd-testsuite-debug";


kinke:
I don't think so. All but one already worked (with gdb 7.11.1) if run manually outside dmd-testsuite, so we'd need to figure out why they fail as part of dmd-testsuite.

rainers:
> Any change to this with your recent improvements?

I would not expect much. 

Trying to compile LDC on Ubuntu 16.04 with precompiled LLVM by just following the the steps in circleci.yml, I keep getting link errors with symbols containing std::string that have an additional abi tag cxx11 in the LLVM library. I'm probably hitting problems like http://stackoverflow.com/questions/36159238/linking-problems-due-to-symbols-with-abicxx11. Is there some way around that?

rainers:
> link errors with symbols containing std::string that have an additional abi tag cxx11 in the LLVM library

It's the other way around.

> Is there some way around that?

Disabling the new abi works for me:
```
CC=clang CXX=clang++ cmake -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0" .
```

kinke:
Closing in favor of #1784.

Issue #1618, opened 2016-07-12
Ubuntu 16.10 linker failures due to PIE by default (relocation R_X86_64_32S … can not be used)


ximion:
Hi!
I am trying to make [appstream-generator](https://github.com/ximion/appstream-generator) compile with LDC, and also make it your new default D compiler on Debian.
Unfortunately, compiling fails with
```
collect2: fatal error: ld terminated with signal 6 [Abgebrochen], core dumped
compilation terminated.
/usr/bin/ld.bfd.real: ld.bfd.real: wcsrtombs.c:94: __wcsrtombs: Zusicherung »data.__outbuf[-1] == '\0'« nicht erfüllt.
Error: /usr/bin/gcc failed with status: 1
```
(assert failure)

When running with LANG=C, I get a different error:
```
/usr/bin/ld.bfd.real: .dub/obj/app.o: relocation R_X86_64_32S against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
.dub/obj/app.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
Error: /usr/bin/gcc failed with status: 1
/tmp/ldc2-1.0.0-linux-x86_64/bin/ldc2 failed with exit code 1.
```

A similar error also appears when trying to compile [Terminix](https://github.com/gnunn1/terminix)
```
terminix ~master: building configuration "dynamic"...
/usr/bin/ld.bfd.real: .dub/obj/app.o: Die Umlagerung von 
.dub/obj/app.o: error adding symbols: Ungültiger Wert
collect2: error: ld returned 1 exit status
Error: /usr/bin/gcc failed with status: 1
/tmp/ldc2-1.0.0-linux-x86_64/bin/ldc2 failed with exit code 1.
```
```
terminix ~master: building configuration "dynamic"...
/usr/bin/ld.bfd.real: .dub/obj/app.o: relocation R_X86_64_32 against `_D3app4mainFAAyaZ12__dgliteral2MFNaNbNfZAya' can not be used when making a shared object; recompile with -fPIC
.dub/obj/app.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
Error: /usr/bin/gcc failed with status: 1
```

I am running on an up-to-date Debian, compiling is done with dub.
GCC version is 5.4, the issue appears when using the precompiled sources as well as when using manually compiled LDC binaries.
[...]

PR #1664, opened 2016-08-02
Default to PIC code on Linux

ximion:
Modern Linux distributions have their toolchain generate PIC code for
additional security features (like ASLR).
Since there is no (sane) way to detect whether the toolchain defaults to
PIC code, we simply default to PIC code on all Linux
distributions to avoid linking issues on these OSes.

The relocation model can be switched back to non-PIC code manually at
any time.

This change resolves issue #1618

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -528,7 +528,10 @@ llvm::TargetMachine *createTargetMachine(
       // Darwin defaults to PIC (and as of 10.7.5/LLVM 3.1-3.3, TLS use leads
       // to crashes for non-PIC code). LLVM doesn't handle this.
       relocModel = llvm::Reloc::PIC_;
-    } else if (triple.getEnvironment() == llvm::Triple::Android) {
+    } else if (triple.isOSLinux()) {
+      // Modern Linux distributions have their toolchain generate PIC code for additional security
+      // features (like ASLR). We default to PIC code to avoid linking issues on these OSes.
+      // On Android, PIC is default as well.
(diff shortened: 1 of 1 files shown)


driver/targetmachine.cpp
```diff
@@ -530,6 +530,10 @@ llvm::TargetMachine *createTargetMachine(
       relocModel = llvm::Reloc::PIC_;
     } else if (triple.getEnvironment() == llvm::Triple::Android) {
       relocModel = llvm::Reloc::PIC_;
+    } else if (triple.getOS() == llvm::Triple::Linux) {
```
JohanEngelen: could you change it to use `isOSLinux()` like the Darwin check above?

ximion: Will `isOSLinux` be true for Android too? (because in theory, the triplet should be something like `<arch>-linux-android` for Android (and `<arch>-linux-gnu` for traditional distros).
I don't know how LLVM handles this.

ximion: The docs answer this: It's exactly as I expected and we can assume a Linux OSType for Android as well (which makes the code a bit simpler).
PR is updated, thanks!

ximion:
Right, what I was actually wanting to write there was "have their toolchain output PIC code" :)

JohanEngelen:
@redstar This is important to mention in the release notes. At least one project fails to build after this change, so it's good to know what to look for.

JohanEngelen:
@redstar I see you already added something about it to 1.1.0-beta3 release notes. I will add a little more.

PR #1662, opened 2016-08-01
gen: Do not store CaseStatement index values in AST nodes

klickverbot:
Another PR to remove bodyBB is upcoming, but the latter will include more refactorings because it makes sense to address #1661 first.

PR #1665, opened 2016-08-02
gen/ir: Move function-local codegen state into separate class, also store case/default body bbs there

### LINKED ISSUE #1661: IrFunction: Separate transitory function body codegen state from permanent parts
Currently, `IrFunction` contains transitory state needed during codegen of the function body (ScopeStack, allocaPoint, etc.) as well as information used from other parts of gen/ir (e.g. for generating a function call, and other instances after a call to DtoDeclareFunction).

klickverbot:
Previously, the transitory state only needed and valid during
generation of the LLVM IR for the function body was conflated
with the general codegen metadata for the function declaration
in IrFunction.

There is further potential for cleanup regarding the use of
gIR->func() and so on all over the code base, but this is out
of scope of this commit, which is only concerned with those
IrFunction members moved to FuncGenState.

GitHub: Fixes #1661.

Includes/depends on #1662.


JohanEngelen:
Ah shoot, I didn't see this PR includes #1662 .

JohanEngelen:
Yes. Several PGO tests depend on the ordering in IR (and our order is not always very logical, iirc the PGO merge also included some IR reordering to make things a little easier to understand)

JohanEngelen:
But then someone is going to write
```d
switch() {
  default: ...
 case 1:...
}
```
and complain ;-)

JohanEngelen:
Yeah, I think the new order is nicer too.
I think for PGO I only reordered for-loop BBs.

kinke:
Oh green now. Thanks for the work, David. I only took a very quick glance at it, I like it. I may take this a bit further at some point by introducing a TryCatchScope class.

PR #1672, opened 2016-08-05
Add -cov-merge commandline flag.

JohanEngelen:
A simple addition such that this eye-sore is not needed: https://github.com/dlang/druntime/blob/2db9857ce23f93767856e24c4afc63842a548181/src/test_runner.d#L88-L92

### DIFF
--- a/gen/module.cpp
+++ b/gen/module.cpp
@@ -65,6 +65,12 @@ static llvm::cl::opt<bool>
              llvm::cl::desc("Write object files with fully qualified names"),
              llvm::cl::ZeroOrMore);
 
+static llvm::cl::opt<bool> coverageDMDMerge(
+    "cov-merge",
+    llvm::cl::desc(
+        "Enable merging of DMD-style coverage reports with existing data."),
+    llvm::cl::ZeroOrMore);
+
 static void check_and_add_output_file(Module *NewMod, const std::string &str) {
   static std::map<std::string, Module *> files;
 
@@ -636,8 +642,9 @@ static void addCoverageAnalysis(Module *m) {
                           m->d_cover_data, idxs, true));
   }
 
-  // Create "static constructor" that calls _d_cover_register2(string filename,
-  // size_t[] valid, uint[] data, ubyte minPercent)
+  // Create "static constructor" that calls _d_cover_register2. Also call
+  // dmd_coverSetMerge if needed. This way each module will call
+  // dmd_coverSetMerge (so more often than needed), but this does no harm.
   // Build ctor name
   LLFunction *ctor = nullptr;
   std::string ctorname = "_D";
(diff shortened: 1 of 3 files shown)


JohanEngelen:
(once merged, I'll move all this coverage code into coverage.cpp and rename it to something like coverage_dmdstyle.cpp to separate it from future llvm-cov support)

JohanEngelen:
> Are there plans to add a similar switch to DMD?

Definitely not by me.

I thought it is a simple addition, for more user convenience (with motto: let's be better than DMD, instead of on-par). 
Btw, I haven't spent much time on LLVM-style coverage yet, but I fear that it may be impossible with some front-end AST modifications/lowerings. It may take some time before we get it working well.

JohanEngelen:
> perfectly good interface already exists

I think the current interface is really terrible.

wilzbach:
> I think the current interface is really terrible.

Yep a CLI switch would be more intuitive for a user, but I agree with @klickverbot that this should be upstreamed.

Speaking of __dmd__  - the function name is actually pretty funny [dmd_coverSetMerge](https://dlang.org/phobos/core_runtime.html#.dmd_coverSetMerge) and the docs even explicitly say

> This is a dmd specific setting. 

If it's supported by LDC, then you should propose a rename in `coverSetMerge` or better something with `coverage` in its name like `coverageMerge` (no need to save the three chars of "age" and makes searching for it easier).

JohanEngelen:
> but I agree with @klickverbot that this should be upstreamed.

This cannot be upstreamed. It is not a "front-end" thing. Yes, the DMD guys can implement it themselves too. Would be great.

wilzbach:
> This cannot be upstreamed. It is not a "front-end" thing.

I know, but @klickverbot/I were saying that the CLI/ways to do things should be similar.
(This doesn't mean I dislike LDC having a better interface).

> Yes, the DMD guys can implement it themselves too. Would be great.

:+1:

JohanEngelen:
> the CLI/ways to do things should be similar.

Yes.

All I read is "we don't add something until DMD adds it"...

wilzbach:
FYI druntime will support setting these flags with 2.078.0: https://dlang.org/changelog/2.078.0.html#covopts

Is this still needed then?

JohanEngelen:
@wilzbach Thanks for the info. I guess that's similar enough to cover a large part of this. For some use cases where it is not possible to add commandline flags to the executable (e.g. in scripts that one can't/shouldn't change), it'd be nice to be able to set the configuration through other means, e.g. built-in like in this PR, or probably better: through an ENV variable. ASan uses an ENV var, which is quite nice. So perhaps an idea to be able to set DRT options through an ENV variable aswell.

wilzbach:
I agree. I opened an issue, s.t. it doesn't get forgotten:

https://issues.dlang.org/show_bug.cgi?id=18120

Now that there's a unified handling of CLI parameters at druntime, it shouldn't be too difficult to add...

PR #1675, opened 2016-08-08
[cmake] Fix CMake for when multiple D_COMPILER_FLAGS are passed.

JohanEngelen:
CMake generation is broken when multiple D compiler flags are passed to the cmake script, e.g. for `export DMD="ldmd2 -fprofile-instr-generate -vp-counters-per-site=20"`.

I tried doing the `separate_arguments` inside FindDCompiler, but then other parts of the generation start failing.

(I think we do have to pass the compiler flags to the compiler when extracting the system linker, because some flags can influence the outcome)


JohanEngelen:
(hm, did something wrong with git apparently, now Kai shows up as author :))

JohanEngelen:
> IIRC I even mentioned in a commit message that I couldn't get D_COMPILER_FLAGS to work when testing the flag extraction thing.

Yeah I remember seeing that too. Fixed now too.

Relative path to host compiler in DMD env variable is fixed now too.

Issue #1652, opened 2016-07-29
PPC64: Assertion failure in toir.cpp


redstar:
On PPC64 there is an assertion failure while compiling `std.math`:

    ldc2: /opt/llvm/include/llvm/IR/Instructions.h:1113: void llvm::ICmpInst::AssertOK(): Assertion `(getOperand(0)->getType()->isIntOrIntVectorTy() || getOperand(0)->getType()->isPtrOrPtrVectorTy()) && "Invalid operand types for ICmp instruction"' failed.

In `toir.cpp`, line 1195, an `ICmp` instruction is generated. The assertion is triggered because both operands have floating point type.

redstar:
It turns out that there is still an ABI problem with ppc/ppc64. I fixed most of it but not yet everything.

redstar:
The ABI fixes are now good enough that LDC can compile itself. There's still an error left...

JohanEngelen:
@redstar What's the status of this?

PR #1680, opened 2016-08-13
[WIP] PPC/PPC64: Fix ABI errors.

redstar:
Issue #1652 seems to be caused by ABI errors in the bootstrap compiler.
This PR updates the PPC/PPC64 ABI to conform to clang/gcc output.


redstar:
This is not yet perfect. It fixes some errors but not all. Especially it is not possible to compile LDC 1.1.0 with this version.

kinke:
Is the omission of `IntegerRewrite::isObsoleteFor()` really necessary?

redstar:
This is what clang seems to do. But I can check it.

kinke:
IIRC, clang also does obsolete integer rewrites for System V. It's just that a LL struct with a single int/pointer is normally passed/returned identically as a rewritten integer. Omitting the rewrite in these cases just leads to a nicer and shorter IR and a shorter log file.

redstar:
It seems to be different on PPC. From our test suite, using clang 3.7 on Linux/PPC64:

`ldc_cabi2.cpp`:

    declare void @dretb1(%struct.B1* sret, i8) #0

With the obsolete integer rewrites in place I get for `ldc_cabi1.d`:

    define void @dretb1(%ldc_cabi1.B1* noalias sret align 1 %.sret_arg, %ldc_cabi1.B1 %x_arg) comdat {

And without it:

    define void @dretb1(%ldc_cabi1.B1* noalias sret align 1 %.sret_arg, i8 %x_arg) comdat {

I prefer to match clang as close as possible here.

kinke:
As I said, as long as `ldc_cabi1.B1` contains a single `i8`, there should be absolutely no difference except for uglier IR.

> I prefer to match clang as close as possible here.

Then we should do it consistently, either removing the obsoleteness check for all ABIs or doing it everywhere (as currently done for at least all 3 major ABIs I'm familiar with - x86 and x86_64). Needless to say, I prefer keeping it.

PR #1681, opened 2016-08-13
Disable cross-module inlining of functions that call __invariant*.

### LINKED ISSUE #1678: [ldc2-1.1.0-beta2] Missing symbol with inlining enabled
Compiling the following module `dt.d`

    import std.datetime;

    void main()
    {
        auto currentTime = Clock.currTime();
        auto timeString = currentTime.toISOExtString();
        auto restoredTime = SysTime.fromISOExtString(timeString);
    }

with `ldc2 --enable-inlining dt.d` results in the linker error

    dt.obj : error LNK2019: unresolved external symbol _D3std8datetime4Date14__invariant234MxFNaNfZv referenced in function _D3std8datetime4Date26__T16fromISOExtStringTAywZ16fromISOExtStringFNaNfxAywZS3std8datetime4Date
    dt.obj : error LNK2019: unresolved external symbol _D3std8datetime9TimeOfDay14__invariant265MxFNaNfZv referenced in function _D3std8datetime9TimeOfDay26__T16fromISOExtStringTAywZ16fromISOExtStringFNaNfxAywZS3std8datetime9TimeOfDay
    dt.exe : fatal error LNK1120: 2 unresolved externals

The error was reported in the forum: http://forum.dlang.org/post/tsziqjllwzthbnxmpmna@forum.dlang.org

JohanEngelen:
Resolves #1678.

If this fixes things, I'll try to make the testcase independent of phobos. That should also make it compile much faster.

### DIFF
--- a/gen/function-inlining.cpp
+++ b/gen/function-inlining.cpp
@@ -123,6 +123,17 @@ bool defineAsExternallyAvailable(FuncDeclaration &fdecl) {
     return false;
   }
 
+  // Because the frontend names `__invariant*` functions differently depending
+  // on the compilation order, we cannot emit the `__invariant` wrapper that
+  // calls the `__invariant*` functions.
+  // This is a workaround, the frontend needs to be changed such that the
+  // __invariant* names no longer depend on semantic analysis order.
+  // See https://github.com/ldc-developers/ldc/issues/1678
+  if (fdecl.isInvariantDeclaration()) {
+    IF_LOG Logger::println("__invariant cannot be emitted.");
+    return false;
+  }
+
   // TODO: Fix inlining functions from object.d. Currently errors because of
   // TypeInfo type-mismatch issue (TypeInfo classes get special treatment by the
   // compiler). To start working on it: comment-out this check and druntime will
--- a/tests/codegen/inlining_invariants_gh1678.d
+++ b/tests/codegen/inlining_invariants_gh1678.d
@@ -1,0 +1,21 @@
+// RUN: %ldc --enable-inlining -of=%t%exe %s
+
+// https://github.com/ldc-developers/ldc/issues/1678
+
+import std.datetime;
+
+// Extra test that fail when a simple frontend change is tried that names __invariant using the line and column number.
+class A {
+    mixin(genInv("666")); mixin(genInv("777"));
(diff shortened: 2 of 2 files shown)


JohanEngelen:
The performance impact of this cross-module emission disabling is not so bad, invariant checks are only generated for non-release builds.

JohanEngelen:
After more thought, this may be overly defensive. I'll try with not emitting the "master" __invariant that calls the other __invariant*.

JohanEngelen:
Made it much simpler: we only need to not emit the main __invariant that calls the sub-invariant functions whose names are unpredictable.

gen/function-inlining.cpp
```diff
@@ -123,6 +123,17 @@ bool defineAsExternallyAvailable(FuncDeclaration &fdecl) {
     return false;
   }
 
+  // Because the frontend names __invariant* functions differently depending on
+  // the compilation order, we cannot emit functions that call __invariant*
+  // because it results in missing symbols during linking.
+  // This is a workaround, the frontend needs to be changed such that the
+  // __invariant* names no longer depend on semantic analysis order.
+  // See https://github.com/ldc-developers/ldc/issues/1678
+  if (fdecl.isInvariantDeclaration()) {
```
klickverbot: If I understand correctly, this works because the problematic references are only within the "top-level" __invarant functions, which match `isInvariantDeclaration`?

Then maybe this is clearer:

Because the frontend names __invariant* functions differently depending on the compilation order, we cannot emit <del>functions that call __invariant*</del> the __invariant wrappers forwarding to them <del>because it results in missing symbols during linking</del>.

klickverbot: If I understand correctly, this works because the problematic references are only within the "top-level" __invarant functions, which match `isInvariantDeclaration`?

Then maybe this is clearer?

Because the frontend names __invariant* functions differently depending on the compilation order, we cannot emit <del>functions that call __invariant*</del> the __invariant wrappers forwarding to them <del>because it results in missing symbols during linking</del>.

JohanEngelen: ok, modified it.

redstar:
Yes.

PR #1683, opened 2016-08-14
PPC/PPC64: Fix ABI errors.

redstar:
Issue #1652 seems to be caused by ABI errors in the bootstrap compiler.
This PR updates the PPC/PPC64 ABI to conform to clang/gcc output.


gen/abi-ppc.cpp
```diff
@@ -39,20 +41,49 @@ struct PPCTargetABI : TargetABI {
...
+    // On ppc64, they are always passed by value. However, clang
+    // used byval for type > 64 bytes.
+    t = t->toBasetype();
+    return (t->ty == Tsarray || t->ty == Tstruct) && (!Is64Bit || t->size() > 64);
   }
 
   void rewriteFunctionType(TypeFunction *tf, IrFuncTy &fty) override {
+    // RETURN VALUE
+    Type *retTy = fty.ret->type->toBasetype();
+    if (!fty.ret->byref) {
```
kinke: Since what's happening in the `true` branch is identical with the body of `rewriteArgument()`, why not simply call `rewriteArgument(fty, *fty.ret)`?

gen/abi-ppc.cpp
```diff
@@ -29,6 +29,8 @@
 
 struct PPCTargetABI : TargetABI {
   ExplicitByvalRewrite byvalRewrite;
```
kinke: Not needed anymore.

gen/abi-ppc.cpp
```diff
@@ -39,20 +41,49 @@ struct PPCTargetABI : TargetABI {
...
     }
 
-    // Return structs and static arrays on the stack. The latter is needed
-    // because otherwise LLVM tries to actually return the array in a number
-    // of physical registers, which leads, depending on the target, to
-    // either horrendous codegen or backend crashes.
+    // FIXME
     Type *rt = tf->next->toBasetype();
-    return (rt->ty == Tstruct || rt->ty == Tsarray);
+    if (tf->linkage == LINKd)
```
kinke: It'd probably make sense not to have a special case for `extern(D)`. The D ABI should normally match the platform's C ABI (except for reversed arguments order and excluding Win32) according to the docs.

gen/abi-ppc.cpp
redstar: Thanks!

gen/abi-ppc.cpp
redstar: See the comment in `abi-ppc64le.cpp` and `abi-arm.cpp`. If we match the C ABI  then NVRO does not work (failures in CTFE test cases!)

gen/abi-ppc.cpp
redstar: I am not yet sure that this is the final version. When I am sure I do.

gen/abi-ppc.cpp
kinke: Hmm, if I understand this correctly, the issue is that the postblit ctor is invoked when assigning the result value to a variable if the struct isn't returned via sret. So the correct fix seems to be either eliding the postblit (=> move the result value into the variable by a simple blit and don't destruct the return value) or forcing sret for all structs with postblit and/or dtor.

kinke: The latter (forcing sret) is actually done for Win64: https://github.com/ldc-developers/ldc/blob/master/gen/abi-win64.cpp#L71

kinke: And also for System V: https://github.com/ldc-developers/ldc/blob/master/ddmd/argtypes.d#L279

kinke: And x86 as well: https://github.com/ldc-developers/ldc/blob/master/gen/abi-x86.cpp#L131

klickverbot: Regardless, this currently says `if (x) return a; return a;`, which – even with the fixme comment there – doesn't make a lot of sense.

kinke:
Merged as part of #1905.

Issue #966, opened 2015-06-06
Class instances are initialized twice


kinke:
Runtime function `_d_newclass()` initializes the instance with `T.init`, and LDC then re-initializes it again in `DtoNewClass()`. This certainly doesn't help performance.
https://github.com/ldc-developers/ldc/pull/916 was an attempt to fix this, but got reverted because the 2nd initialization is currently needed for the `GarbageCollect2Stack` optimization pass (run as part of release builds), as the call to `_d_newclass()` is omitted when replacing a heap allocation by a stack allocation.
That GC2Stack pass operates on the IR layer, so I'm not sure we can easily get ahold of the D type's `ClassInfo` and its initializer to fix this properly in that pass.

/edit: Well, the ClassInfo is the single argument to `_d_newclass()`, so it can be inferred from there.

kinke:
Fixed by https://github.com/ldc-developers/ldc/pull/1684.

PR #1684, opened 2016-08-14
Fix issue #966 (initializing class instances twice)

kinke:
See https://github.com/ldc-developers/druntime/pull/81 and #966. Fixing it in druntime is way simpler. :)

### DIFF
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -84,8 +84,8 @@ DValue *DtoNewClass(Loc &loc, TypeClass *tc, NewExp *newexp) {
   // allocate
   LLValue *mem;
   if (newexp->onstack) {
-    // FIXME align scope class to its largest member
-    mem = DtoRawAlloca(DtoType(tc)->getContainedType(0), 0, ".newclass_alloca");
+    mem = DtoRawAlloca(DtoType(tc)->getContainedType(0), DtoAlignment(tc),
+                       ".newclass_alloca");
   }
   // custom allocator
   else if (newexp->allocator) {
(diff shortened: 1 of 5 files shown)


JohanEngelen:
`_d_allocclass` also needs to be added to the `static if (OVERRIDE_MEMALLOC)` block in `ddmd/root/rmem.d`.

kinke:
Thanks Johan; that's what happens if the .d files aren't part of the Visual Studio solution...

gen/classes.cpp
```diff
@@ -84,8 +84,8 @@ DValue *DtoNewClass(Loc &loc, TypeClass *tc, NewExp *newexp) {
   // allocate
   LLValue *mem;
   if (newexp->onstack) {
-    // FIXME align scope class to its largest member
-    mem = DtoRawAlloca(DtoType(tc)->getContainedType(0), 0, ".newclass_alloca");
+    mem = DtoRawAlloca(DtoType(tc)->getContainedType(0), DtoAlignment(tc),
```
klickverbot: Did you verify this? I'm not sure off the top of my head whether alignment() for TypeClass returns the (pointer-size) alignment for the reference or for the contents.

Here, an IR-level test would actually make sense. ;)

kinke: Nope, I was too lazy. I gave it a very quick look, and IIRC, it uses the aggregate's `alignsize()` etc., so I'd expect the content alignment. So I'm rather worried that we use the content alignment for the references too. ;)
Needs to be investigated.

klickverbot: Could you just add a quick issue then, please? (We should probably do that for all leftover FIXMEs anyway.)

gen/passes/SimplifyDRuntimeCalls.cpp
```diff
@@ -372,6 +372,7 @@ void SimplifyDRuntimeCalls::InitOptimizations() {
   Optimizations["_d_newarraymiT"] = &Allocation;
   Optimizations["_d_newarraymvT"] = &Allocation;
   Optimizations["_d_newclass"] = &Allocation;
```
klickverbot: Isn't this actually an invalid optimisation if _d_newclass invokes the constructor?

kinke:
Upstream PR: https://github.com/dlang/dmd/pull/6067

gen/classes.cpp
kinke: Issue #1692 created.

PR #1685, opened 2016-08-14
cmake: Split D module compile and link dependencies

klickverbot:
This way, we don't rebuild all the D modules on changes to the
C++ bits anymore.


JohanEngelen:
Fantastic :)

PR #1687, opened 2016-08-14
AppVeyor: Try to figure out what's happening in core.thread unittests

kinke:


rainers:
For the record: tests also pass for me with "-O3" or "-gc -link-debuglib", but I have the same setup as you do.

Maybe provide the test_results directory with the failing executables as an artifact, so we can check whether it's a build issue or a runtime problem with the environment.

kinke:
I downloaded the published x64 compiler artifact (from AppVeyor directly) and used that one to compile `core.thread` and `runnable/eh.d` locally on my box, and both worked. So it has to be a runtime problem with the AppVeyor environment.

kinke:
Good idea, @rainers. It really is a build issue; both executables crash on my system too. As I had no issue with the identical compiler, it's most likely a lib in the AppVeyor VM.

kinke:
May be caused somehow by exception chaining. I hit a `terminate()` in `__FrameUnwindFilter`...

kinke:
Oh I've just spotted this `__processing_throw()` which rings some bells. Don't you read some machine code bytes trying to determine whether it's the debug/release dynamic/static MSVC runtime or something like that?

kinke:
AppVeyor seems to use the WinSDK v10.0.14393, while I'm still at v10.0.240 (INCLUDE/LIB environment variables)...

kinke:
Yep, as soon as I install the newer WinSDK, I get the crashes too. I guess [this beauty](https://github.com/ldc-developers/druntime/blob/ldc/src/ldc/eh/msvc.d#L317) is to be adapted.

rainers:
> I guess this beauty is to be adapted.

Yeah, looks like a good candidate. I can have a look tonight.

rainers:
Thanks for pinpointing the problem, should be fixed now.

PR #1688, opened 2016-08-14
CI: Parallelize dmd-testsuite execution

kinke:
Result on my quad-core for dmd-testsuite-debug, using `DMD_TESTSUITE_MAKE_ARGS=-j4`: 490 secs => 175 secs

I've also used the opportunity to streamline the Travis/CircleCI/AppVeyor scripts.
The tests order has changed for AppVeyor (dmd-testsuite => Lit tests => unit tests) and the unit tests execution has been parallelized.
For CircleCI, the unittest compilation should now be properly parallelized, which should give another nice speed-up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,22 +92,22 @@ script:
   # Run dmd-testsuite.
   -
     if [ "${TEST_CONFIG}" = "Debug" ]; then
-      CC="" MAKEOPTS=-j4 ctest --verbose -R "dmd-testsuite-debug";
+      CC="" DMD_TESTSUITE_MAKE_ARGS=-j3 ctest --verbose -R "dmd-testsuite-debug";
     elif [ "${TEST_CONFIG}" = "Release" ]; then
-      CC="" MAKEOPTS=-j4 ctest --verbose -R "dmd-testsuite" -E "-debug";
+      CC="" DMD_TESTSUITE_MAKE_ARGS=-j3 ctest --verbose -R "dmd-testsuite" -E "-debug";
     else
-      CC="" MAKEOPTS=-j4 ctest --verbose -R "dmd-testsuite";
+      CC="" DMD_TESTSUITE_MAKE_ARGS=-j3 ctest --verbose -R "dmd-testsuite";
     fi
   # Run LLVM IR testsuite.
   - ctest --output-on-failure -V -R "lit-tests"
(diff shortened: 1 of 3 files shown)


wilzbach:
Nice job :tada: 

Just a question out of interest:

> DMD_TESTSUITE_MAKE_ARGS=-j3 

How did you figure out that 3 if the best parallel factor for the CI? Did you by chance test that?
(the upstream testsuite uses 2 as parallelism factor)

kinke:
I mostly took what was previously intended (with that apparently useless `MAKEOPTS` variable), except that I don't think `-j4` will ever pay off, so I reduced that to `-j3`. For Travis, we're supposed to have 2 cores on Linux and only 1 on OSX. AppVeyor provides 2 as well. I haven't found out about CircleCI, but I guess it's 2 there too.
My best guess is that the original author of the `-j<N>` flags did some tests and figured that `-j3/4` may be faster than `-j2` while the additional memory consumption isn't an issue. I guess this could be due to SMT (HyperThreading etc.; not sure how the CPU core numbers are to be interpreted) or stalling I/O.

kinke:
Btw, the Travis results vary quite wildly unfortunately, so exact comparisons can't be made.

kinke:
But the Travis results so far clearly indicate a speed-up. E.g., the OSX Debug job didn't even take 16 mins, and used to take around 20 mins. And that only due to `-j3` for the DMD testsuite, although Travis states the OSX VM only has a single CPU core...

kinke:
CircleCI has only improved by about 3.5 minutes (~10%).

kinke:
AppVeyor is faster by about 7 minutes per job (~20%).

kinke:
Travis OSX: The merged debug+release job finishes in about the same time as the previous release job (which used DMD to compile the front-end, to be fair) - 35 mins. This means getting rid of the 20 mins debug job for free incl. Travis script simplifications. :)
The Travis Linux jobs also hit new records, the 3.5 coverage job finishing in about 20 mins, the 3.8.1 job in 23.5 mins and the multilib job twice in a row in under 40 mins (!).

kinke:
> Where is DMD_TESTSUITE_MAKE_ARGS actually read?

https://github.com/ldc-developers/dmd-testsuite/commit/999d062961c1366732f78fe936bbc3a84155c0c2

PR #1694, opened 2016-08-15
Fix unused private field warning (`SwitchCaseTargets::llFunc`) and remove use of global gIR.

JohanEngelen:
We got two new warnings:
```
In file included from ../gen/funcgenstate.cpp:10:
.././gen/funcgenstate.h:129:12: warning: private field 'irs' is not used [-Wunused-private-field]
  IRState &irs;
           ^
.././gen/funcgenstate.h:163:25: warning: private field 'llFunc' is not used [-Wunused-private-field]
  llvm::Function *const llFunc;
                        ^
```

The first warning appears to be a clang bug :(, the second one is fixed by this PR.
Also removed use of `gIR` (originally `llFunc` was used instead of `gIR` for creating BBs).

### DIFF
--- a/gen/funcgenstate.cpp
+++ b/gen/funcgenstate.cpp
@@ -20,8 +20,7 @@ JumpTarget::JumpTarget(llvm::BasicBlock *targetBlock,
     : targetBlock(targetBlock), cleanupScope(cleanupScope),
       targetStatement(targetStatement) {}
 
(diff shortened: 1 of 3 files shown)


JohanEngelen:
> The first warning appears to be a clang bug :(

My god... I have been trying to reproduce the "bug" and only now I find out I have been staring at the wrong `irs` the whole time............. Will fix in another PR.

PR #1695, opened 2016-08-15
Fix too restrictive assertion for binassigns

kinke:
This was hit when compiling `ddmd/lexer.d:1189`, where a `const(T)*` is implicitly cast to a `const(T*)`.
The `DLValue` ctor makes sure the LL pointer and the D type are compatible.


kinke:
Except for this, there's no other LDC/LLVM assertion when compiling master with itself on Win64.

Issue #1690, opened 2016-08-15
Make sure MSVC Build Tools are automatically detected by LDC


klickverbot:
There was a short discussion about the [Visual C++ Build Tools 2015](https://blogs.msdn.microsoft.com/vcblog/2015/11/02/announcing-visual-c-build-tools-2015-standalone-c-tools-for-build-environments/) (just the compiler/linker/build system without the IDE) on the NG before, but I don't think we ever got around to add auto-detection support to it.

This is important as it shapes the experience for new users.

kinke:
Afaik, the build tools and VS 2015 can't co-exist on the same box. Meaning that I can't test how to reliably detect the build tools in the registry (and I'm not willing to create a Win VM for this purpose).

JohanEngelen:
Did https://github.com/ldc-developers/ldc/pull/1699 resolve this?
Or are we waiting for someone to confirm that it works?

klickverbot:
I'd say we reopen it if it persists. ;)

PR #1699, opened 2016-08-16
Windows: Detect Visual C++ Build Tools automatically

kinke:
Based on http://forum.dlang.org/post/xnjxxqypqycxlhzgzsrt@forum.dlang.org (thanks André!). Visual Studio provides the same keys in HKLM, so detecting them this way should work for both Visual Studio and the stand-alone Build Tools.
The lookup outside `WOW6432Node` is for 32-bit Windows.
A user-provided `LDC_VSDIR` environment variable doesn't have to end with a backslash anymore, and may point to a VS folder (backward-compatible) or a VC folder directly.

### DIFF
--- a/vcbuild/msvcEnv.bat
+++ b/vcbuild/msvcEnv.bat
@@ -3,24 +3,31 @@
 :: Environment already set up?
 if not "%VSINSTALLDIR%"=="" goto :eof
 
-:: Clear an existing LDC_VSDIR environment variable if the directory doesn't exist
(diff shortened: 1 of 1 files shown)


kinke:
This should fix #1690.

kinke:
> my limited Windows knowledge

Don't worry, I've just found out about that `%<VAR>:~<startIndex>[,<endIndex>]%` substring notation (incl. negative indices) too. ;)

PR #1704, opened 2016-08-17
Refactor ModuleInfo registry code, optimize code size

klickverbot:
Some general improvements as a prelude to OS X shared library support.

Issue #1071, opened 2015-09-10
OS X: Can't load a D shared library first, then load a C shared library


p0nce:
**OS:** `OSX 10.10.4`
To reproduce this you need a host program:

**Setup**

Here the host program source.

What it does is:
- for each dynlib in command line:
  - load that dynlib
  - call the VSTPluginMain function if it exist, 
  - unload it

File `ldvst.cpp`

``` cpp

#include <dlfcn.h>
#include <cstdio>
#include <cstring>
#include <vector>

typedef __cdecl void* (*VSTPluginMain_t)(void*);

int main(int argc, char**argv)
{
    std::vector<char*> dllPaths;
    if (argc < 2)
    {
        printf("usage: ldvst [-lazy] <thing.vst>\n");
        return 1;
    }

    bool lazy = false;

    for (int i = 1; i < argc; ++i)
    {
        char* arg = argv[i];
        if (strcmp(arg, "-lazy") == 0)
            lazy = true;
        else if (strcmp(arg, "-now") == 0)
            lazy = false;
        else
            dllPaths.push_back(arg);
    }

    for (int i = 0; i < dllPaths.size(); ++i)
    {
        char* dllPath = dllPaths[i];

        printf("dlopen(%s)\n", dllPath);
        void* handle = dlopen(dllPath, lazy ? RTLD_LAZY : RTLD_NOW);
        if (handle == NULL)
        {
            printf("error: dlopen of %s failed\n", dllPath);
            return 2;
        }

        VSTPluginMain_t VSTPluginMain = (VSTPluginMain_t) dlsym(handle, "VSTPluginMain");
        printf("dlsym returned %p\n", (void*)VSTPluginMain);

        if (VSTPluginMain != NULL)
        {
            void* result = VSTPluginMain(NULL);
            printf("VSTPluginMain returned %p\n", result);
        }

        printf("dlclose(%s)\n\n", dllPath);
        dlclose(handle);
    }
    return 0;
}
```

The host is compiled with:

`$ clang++ ldvst.cpp -o ldvst-cpp`

Here the D dynlib source:

File `distort.d`

``` d
extern(C) void* VSTPluginMain(void* hostCallback)
{
    import core.runtime;
    Runtime.initialize();
    import std.stdio;
    import core.stdc.stdio;
    printf("Hello !\n");
    Runtime.terminate();
    return null;
}
```

This dynlib can be built with ldc 0.15.2-b2:

[...]

Issue #129, opened 2012-06-20
More PPC and SPARC code to be removed in upstream druntime


alexrp:
See: https://github.com/D-Programming-Language/druntime/pull/246

If this gets merged, please skip the commit that removes the PPC and SPARC code.

PR #1705, opened 2016-08-17
Shared library support for OS X

klickverbot:
Passes all of the test suite, including the shared library tests, except for module collision checking (which is most easily implemented with a hash table, as symbol resolution works differently on OS X).

Requires/includes #1704.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,9 @@ matrix:
     - os: osx
       d: ldc
       env: LLVM_CONFIG="llvm-config-3.8"
+    - os: osx
+      d: ldc
+      env: LLVM_CONFIG="llvm-config-3.8" OPTS="-DBUILD_SHARED_LIBS=ON"
   allow_failures:
     #- env: LLVM_VERSION=3.9
 
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -95,9 +95,52 @@ void Module::makeObjectFilenameUnique() {
 }
 
 namespace {
+/// Ways the druntime module registry system can be implemented.
+enum class RegistryStyle {
+  /// Modules are inserted into a linked list starting at the _Dmodule_ref
+  /// global.
+  legacyLinkedList,
 
-// build ModuleReference and register function, to register the module info in
-// the global linked list
+  /// Module references are emitted into the .minfo section, bracketed with
+  /// extra sections to access the .minfo range. Global constructors/
+  /// destructors make sure _d_dso_registry is invoked once per ELF object.
+  sectionELF,
+
+  /// Module references are emitted into the .minfo section. Global
+  /// constructors/destructors make sure _d_dso_registry is invoked once per
+  /// shared object. A "TLS anchor" function to identify the TLS range
+  /// corresponding to this image is also passed to druntime.
+  sectionDarwin
+};
+
+/// Returns the module registry style to use for the current target triple.
+RegistryStyle getModuleRegistryStyle() {
+  const auto t = global.params.targetTriple;
+
+  if (t->isMacOSX()) {
+    return RegistryStyle::sectionDarwin;
+  }
+
+  if ((t->isOSLinux() && t->getEnvironment() != llvm::Triple::Android) ||
+      t->isOSFreeBSD() ||
+#if LDC_LLVM_VER > 305
+      t->isOSNetBSD() || t->isOSOpenBSD() || t->isOSDragonFly()
+#else
+      t->getOS() == llvm::Triple::NetBSD ||
+      t->getOS() == llvm::Triple::OpenBSD ||
+      t->getOS() == llvm::Triple::DragonFly
+#endif
+          ) {
+    return RegistryStyle::sectionELF;
+  }
+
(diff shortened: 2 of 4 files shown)


JohanEngelen:
sry @klickverbot , been busy lately. I'll try on my machine.

kinke:
It even only manages to build ~40 Phobos release unittest modules, so it's not even close to making it under 48 mins. I'm not even sure it would manage a debug-only run, considering that it would have to run all tests, incl. dmd-testsuite-debug, in the time it took to build the druntime release unittest modules (quick) and those 40 Phobos ones. I bet the release job would never make it. So I'd say that newer OSX image is sadly not really an option. :/

kinke:
Skipping the unittests and running dmd-testsuite (debug and release) and lit-tests only would be an option. I'm not sure I'd like that in master though; maybe opening a special PR with that Travis change (once this is merged) could be another option, as sort of a manual CI test, with occasional manual triggering.

p0nce:
Shit hit the fans for the workaround. It was working well from 10.7 to 10.11 but the dyld function myst have moved in macOS Sierra so linking to these functions statically doesn't work anymore (#129).

So, I'm still very much interested by OSX shared library support (on 32-bit too...), let me know if you need more incentives.

p0nce:
How can I help you move forward? I can test whatever you send me. My home internet connection is heavily limited though.

p0nce:
Trying to test this.
I'm getting build issues with this branch, and not with the master branch. Perhaps a LLVM version problem? I use LLVM 3.6.

JohanEngelen:
@klickverbot Perhaps a rebase helps: some things were fixed with old LLVM versions in https://github.com/ldc-developers/ldc/pull/1783

p0nce:
It's probably this.

p0nce:
The error is the following:

```
...
[  9%] Built target LDCShared
[  9%] Built target LDMD_CXX_LIB
[  9%] Generating bin/ldc2.o
[  9%] Generating bin/ldc2
  "isCatchSwitchBlock(llvm::BasicBlock*)", referenced from:
      TryCatchFinallyScopes::popTryCatch() in libldc.a(trycatchfinally.cpp.o)
make[2]: *** [bin/ldc2] Error 1
make[1]: *** [CMakeFiles/ldc2.dir/all] Error 2
make: *** [all] Error 2
```

The error went away when I rebased the PR on master.

p0nce:
The problem now is that I can't link with that new LDC, even a hello world.
The error is at link:
```
  "__d_dso_registry", referenced from:
      _ldc.register_dso in hello.o
      _ldc.register_dso in libphobos2-ldc.a(utf.o)
      _ldc.register_dso in libphobos2-ldc.a(stdio.o)
      _ldc.register_dso in libphobos2-ldc.a(exception.o)
      _ldc.register_dso in libdruntime-ldc.a(aApply.o)
      _ldc.register_dso in libdruntime-ldc.a(lifetime.o)
      _ldc.register_dso in libdruntime-ldc.a(exception.o)
      ...
```

Not sure why.

p0nce:
Where can I set this -DBUILD_SHARED_LIBS=ON ?
Is it for building druntime and phobos or just LDC?

p0nce:
Thank you for the hand-holding.

I've tried the repro code of #1071 and this patch fixes it for macOS Sierra. 

Previously with unmodified LDC this would crash
```
dlopen(libdistort.so)
dlsym returned 0x106e00e90
Hello !
VSTPluginMain returned 0x0
dlclose(libdistort.so)

dlopen(/System/Library/Frameworks/Cocoa.framework/Cocoa)
./repro.sh: line 4: 10281 Segmentation fault: 11  ./ldvst-cpp libdistort.so /System/Library/Frameworks/Cocoa.framework/Cocoa
```

With the patch it is solved.

```
dlopen(libdistort.so)
dlsym returned 0x101971d70
Hello !
VSTPluginMain returned 0x0
dlclose(libdistort.so)

dlopen(/System/Library/Frameworks/Cocoa.framework/Cocoa)
dlsym returned 0x0
dlclose(/System/Library/Frameworks/Cocoa.framework/Cocoa)

```

I can attach/threads threads and use runtime features => this is awesome.

The only thing that need explaining is how is that fix tied to making the runtime a shared library itself? Will the executable work without the dynlib? Ideally the runtime would keep being statically linked but well :).

p0nce:
Since libphobos2.dylib and libdruntime.dynlib are so big, bundling them in releases would multiply the file size by about 3 (there are 2 format and 2 archs to support, so more dynlibs). The most likely is that I will still link the runtime statically, but not initialize it and not use runtime features.

p0nce:
@klickverbot Thanks for the possible solution.

Though it isn't fun, I'm almost done with the runtime disabling and going fully @nogc. Which is also the opportunity to reduce the impact of `thread_attachThis` / thread_detachThis` in my products (they happen quite often because of timer callbacks). My only hope at this point is that the compiler doesn't assume an initialized runtime ; fortunately that doesn't seem to be the case.A third reason is that there is tooling to detect manual leaks.

PR #1706, opened 2016-08-17
WinEH: test fix for double exception handling with UCRT lib 10.0.14393.0

rainers:
Don't merge, druntime is referenced from branch fix_ucrt14393, not ldc.


rainers:
Seems to work, manually pushed the updated druntime/ldc link.

kinke:
Thanks Rainer!

PR #1707, opened 2016-08-18
cmake: Avoid deprecated TARGET/LOCATION property

klickverbot:
Avoids the associated deprecation/policy warning.


kinke:
Yep, the Travis job timings vary extremely lately, especially the OSX ones (with the least available memory). The OSX job of Travis build 4675 took 36.5 mins, and your 4676 job timed out after 48.5 mins (a single unittest missing). As the OSX VM is supposed to feature a single CPU core, we could try using `-j2` for OSX and `-j3` for Linux only and see if that helps. I guess the memory is managed dynamically, i.e., we get more if the other VMs on the host require less etc.

Issue #1698, opened 2016-08-16
ICE: function not fully analyzed; previous unreported errors compiling std.variant.VariantN!(16LU, int, string).VariantN.__xopEquals?


ximion:
This is a "spinoff" of issue #1696 

I am split-compiling a project. After fixing a lot of issues in the form of:
```error: module utils from file /path/to/project/source/mod/utils.d conflicts with another module utils from file source/mod/utils.d```

I "fixed" them by manually defining `module` for those projects. This smells like another bug, since there was no actual conflict - the paths are the same, the only exception is that one is absolute and the other one is relative.

After resolving these issues, LDC now runs into an ICE when compiling the project:
```
ldc2  '-Iasgen@exe' '-I..' '-I.' '-Isource/' '-I../source/' '-Isource/bindings/auto' '-I../source/bindings/auto' '-I/usr/include/d/mustache-d/src/' '-I/usr/include/d/mustache-d/src/' '-w' '-g'  -of 'asgen@exe/source_engine.d.o' -c ../source/engine.d
Error: Internal Compiler Error: function not fully analyzed; previous unreported errors compiling std.variant.VariantN!(16LU, int, string).VariantN.__xopEquals?
```

Any idea what could be wrong here? Can I get LDC to show the "unreported errors" somehow?

**Comment by kinke** [comment]
What you're running into is [this](https://github.com/ldc-developers/ldc/blob/master/gen/functions.cpp#L774), so apparently the automatically generated `__xopEquals()` functions is missing some semantic passes. Sounds interesting, and a reduced test case would be nice (dustmite?).

**Comment by ximion** [comment]
First of all, Dustmite is amazing! It's an incredibly helpful tool!
Here is a minimized testcase, reproducing this ICE: [ldc_testcase-ICE_1698.zip](https://github.com/ldc-developers/ldc/files/425957/ldc_testcase-ICE_1698.zip)

**Comment by kinke** [comment]
Hmm this is a strange one. I reduced it to:

M.d:
```D
import N;

void foo(C) {}
```

N.d:
```D
import std.variant;

class E {
    this(void delegate() ) {}
}

class R
{
    Algebraic!int[string][] W;

    auto Y()
    {
        new E(
        {
            foreach (AB; W) AB["path"].get!string;
        });
    }
}

class C
{
    R DB;
}
```

1. `ldc2 -c M.d` results in the ICE.
2. As soon as the return type of method `N.R.Y()` is changed to `void`, it compiles fine.
[...]

PR #1709, opened 2016-08-19
Fix issue #1698 (missing semantic3() for special member functions)

kinke:
I have no idea about all the implications of this, but it seems to fix issue #1698.


kinke:
The 3rd commit makes more sense and may even reduce compilation times a bit while still fixing #1698.

Issue #1712, opened 2016-08-20
Problem compiling simple vibe.d project with dub


tchaloupka:
Just tried to build sample project (`dub init -t vibe.d test`) with `dub build -b release --compiler=ldc2 --combined` and it fails. Its ok with dmd.

It's ok without combined flag and also when built without release optimisations, so I guess something with the inlining.

**DUB** version 1.0.0+28-geeff313, built on Aug 20 2016
It issues these flags to ldc2: `-release -enable-inlining -O -w -oq`

**LDC** the LLVM D compiler (1.1.0-beta2):
  based on DMD v2.071.1 and LLVM 3.8.0
  built with LDC - the LLVM D compiler (1.1.0-beta2)
  Default target: x86_64-unknown-linux-gnu
  Host CPU: broadwell

**Fedora 24**

**Output:**
```
$ dub build --combined --compiler=ldc2 -b release | ddemangle
Performing "release" build using ldc2 for x86_64.
depstest ~master: building configuration "application"...
.dub/obj/vibe.crypto.cryptorand.o: In function `vibe.crypto.cryptorand.SystemRNG vibe.crypto.cryptorand.SystemRNG.__ctor()':
../../../.dub/packages/vibe-d-0.7.30-alpha.4/vibe-d/source/vibe/crypto/cryptorand.d:(.text.vibe.crypto.cryptorand.SystemRNG vibe.crypto.cryptorand.SystemRNG.__ctor()[vibe.crypto.cryptorand.SystemRNG vibe.crypto.cryptorand.SystemRNG.__ctor()]+0x66): undefined reference to `@safe shared(core.stdc.stdio._IO_FILE)* std.exception.__T12errnoEnforceTPOS4core4stdc5stdio8_IO_FILEVAyaa60_2f6f70742f6c6463322d312e312e302d62657461322d6c696e75782d7838365f36342f62696e2f2e2e2f696d706f72742f7374642f737464696f2e64Vmi498Z.errnoEnforce(shared(core.stdc.stdio._IO_FILE)*, lazy immutable(char)[])'
[...]

Issue #1703, opened 2016-08-17
When __FILE__ is template parameter, disable inlining.


JohanEngelen:
A function with __FILE__ as default template parameter must explicitly be marked with `pragma(inline, false)` otherwise linking problems may occur:
```d
void foo(string file = __FILE__)(){ ...}
```

See the forum discussion:
https://forum.dlang.org/post/induraqegvpyomvzxwmf@forum.dlang.org

This issue is to track progress, and as an easy referral.

A possible solution is to "infer" `pragma(inline, false)` when __FILE__ is used as default parameter.

Issue #1603, opened 2016-07-04
Nested functions are not cross-module inlined


JohanEngelen:
E.g. with this code in a module:
```d
pragma(inline, true)
bool outerFunc()
{
    pragma(inline, true)
    bool innerFunc()
    {
        return true;
    }

    return innerFunc();
}
```

`innerFunc` will not be available for inlining (`available_externally`) when `outerFunc` is called from another module.
At least part of the problem is that the nested function has already passed semantic3 during the `defineAsExternallyAvailable` checks, see https://github.com/ldc-developers/ldc/commit/75e1573cb25ba2168c839c957a63ed0f60811e8e#diff-db5a49fcfce323871f6dcb6886c9155eR139

Also see #1577 .

PR #1718, opened 2016-08-25
Fix issue #1712

JohanEngelen:
Cross-module inlining (i.e. emitting some code from another module) is only done when a function is not fully semantic3 analysed.
The problem is that sometimes, a function has already been fully "semantic3'ed", but is not emitted anywhere. This happens for nested functions, and those are currently not cross-module inlined (see #1603).
In issue #1712, it happens for a function with template parameter `__FILE__`, i.e. a function that should always be inlined (#1703) exactly because of the link problem observed in #1712.

This fix for #1712 attempts to solve this for now, by emitting `pragma(inline, true)` functions more aggressively. It fixes the current issue, and I hope it does not create new bugs.

Currently dustmiting to create a testcase; it will take a lot of time.


kinke:
Any progress? Otherwise I'd suggest disabling cross-module inlining by default (or only enabling it for optimization levels > 3) for now, until the issue is fully understood.

JohanEngelen:
Found the problem (it's bad), fixing it.

JohanEngelen:
The problem was that inside `DtoDefineFunction`, a call to `DtoResolveFunction` would end up invoking `DtoDefineFunction` and defining the function with `available_externally`, which would not be "fixed up" later because it all happened in the call trace from the `DtoDefineFunction` that should do the fix up. The fix is to tell `DtoResolveFunction` and `DtoDeclareFunction` that they may _not_ end up defining the function.

JohanEngelen:
(I'm still dustmiting the testcase btw. It takes forever)

JohanEngelen:
Dustmiting is done, and a fairly simple testcase resulted. I've modified it such that it doesn't depend on Phobos and added it to the PR.
Further fixes were needed involving with nested functions and nested structs. The fix for this issue uncovers unimplemented nested function cross-module emitting problems.

kinke:
With every new fix, a new issue pops up. ;/
I'd really recommend not enabling cross-module inlining by default for v1.1 yet, and exposing it via optimization levels > 3 for now.

JohanEngelen:
Yeah, I think you're right :(

gen/declarations.cpp
```diff
@@ -206,10 +206,15 @@ class CodegenVisitor : public Visitor {
...
 
     // Define the __initZ symbol.
-    IrAggr *ir = getIrAggr(decl);
-    llvm::GlobalVariable *initZ = ir->getInitSymbol();
-    initZ->setInitializer(ir->getDefaultInit());
-    setLinkage(decl, initZ);
+    // If we reach here during codegen of an available_externally function,
+    // the struct initializer should stay external and therefore must not
+    // have an initializer.
+    if ((decl->getModule() == gIR->dmodule) || decl->isInstantiated()) {
```
klickverbot: This looks like a good candidate for a helper function that takes a `Dsymbol *` or whatever – feeling compelled to copy-and-paste comments is probably a good indicator for that.

gen/function-inlining.cpp
```diff
@@ -143,18 +143,21 @@ bool defineAsExternallyAvailable(FuncDeclaration &fdecl) {
...
     // TODO: This check prevents inlining of nested functions.
     IF_LOG Logger::println("Semantic analysis already completed");
     return false;
   }
 
-  if (alreadyOrWillBeDefined(fdecl)) {
+  if ((fdecl.inlining != PINLINEalways) && alreadyOrWillBeDefined(fdecl)) {
     // This check is needed because of ICEs happening because of unclear issues
     // upon changing the codegen order without this check.
+    // pragma(inline, true) functions remain defineAsExternallyAvailable
```
klickverbot: So what this is saying then is that `alreadyOrWillBeDefined` is buggy?

JohanEngelen: The problem is non-singleobj compilation, because most (all?) instantiated functions are put into one module, which may be different from the callsite's module. So if alreadyOrWillBeDefined  is true the function will be defined, yes, but for inlining it needs to be defined in every calling module.

gen/declarations.cpp
JohanEngelen: Definitely.

JohanEngelen:
I opened #1737 to "fix" all cross-module inlining issues: disable the thing.

I'd like to enable it at `-O4`, but it's up for debate. I think it's nice to be able to compile with -O4 such that compilation also works with earlier LDC versions (compared with the previously unknown `-enable-cross-module-inlining`) and get the boost of cross-module inlining when compiler with the new LDC. (e.g. in our own CMake file, I imagine using `-O4` when LDC is detected)

Upon 1.1.0 release, I'd like to enable cross-module inlining per default again in master, so it gets more testing exposure.

Issue #1324, opened 2016-02-29
Bug with CTFE and unions


JohanEngelen:
The following code outputs "LEN:0" (wrong) instead of "LEN:3" (correct):

``` D
struct TestArray
{
    this(T...)(T vals)  { append(vals); }

    void append(T...)(T vals)
    {
        foreach(v; vals)
            value ~= TestValue();
    }

    import std.typecons;
    mixin Proxy!value;

    TestValue[] value;
}

struct TestValue
{
    union TestType
    {
        int asInt;
        TestArray asTestArray;
    }
    TestType value;

    this(TestArray val) { assign(val); }
    void assign(TestArray val) { value.asTestArray = val; }
    auto get() { return value.asTestArray; }
}

void main()
{
    import std.stdio;
    enum ctObj = TestValue(TestArray(1,2,3));
    writeln("LEN:", ctObj.get.length);
}
```

Changing `union` to `struct` fixes the problem. It's also fixed by changing `enum` to `auto` (CTFE->runtime).

kinke:
Is this still the case? If so, does it work for DMD? CTFE is front-end territory afaik...

JohanEngelen:
I found the culprit.
https://issues.dlang.org/show_bug.cgi?id=16471

JohanEngelen:
https://forum.dlang.org/thread/jzfalivqsvtrnnaubpdh@forum.dlang.org

kinke:
https://github.com/ldc-developers/ldc/pull/1846 moved to DMD behavior in this regard, giving preference to the latest initializer in field declaration order. This should be it from the LDC side; the exact semantics (i.e., no overlapping initializers) should be clearly defined in the front-end.

thewilsonator:
This prints LEN:3 now

PR #1719, opened 2016-08-25
Optimize static array comparisons to a memcmp call for types for which this is valid.

### LINKED ISSUE #1632: Array equal code
Currently a == b generates ugly amount of code calling _adEq2 from druntime, involving typeinfo classes, vtbl lookups and cetera...

For example current foo, and optimized foo2 function.
http://goo.gl/0yWheD

I think we should define compareMemory in druntime and call it on POD arrays. And for most cases it will be inlined.

JohanEngelen:
Resolves #1632.

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -962,8 +962,9 @@ DSliceValue *DtoAppendDCharToUnicodeString(Loc &loc, DValue *arr,
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+namespace {
 // helper for eq and cmp
-static LLValue *DtoArrayEqCmp_impl(Loc &loc, const char *func, DValue *l,
+LLValue *DtoArrayEqCmp_impl(Loc &loc, const char *func, DValue *l,
                                    DValue *r, bool useti) {
   IF_LOG Logger::println("comparing arrays");
   LLFunction *fn = getRuntimeFunction(loc, gIR->module, func);
@@ -994,14 +995,117 @@ static LLValue *DtoArrayEqCmp_impl(Loc &loc, const char *func, DValue *l,
   return gIR->funcGen().callOrInvoke(fn, args).getInstruction();
 }
 
+/// When `true` is returned, the type can be compared using `memcmp`.
+/// See `validCompareWithMemcmp`.
+bool validCompareWithMemcmpType(Type *t) {
+  switch (t->ty) {
+  case Tsarray: {
+    auto *elemType = t->nextOf()->toBasetype();
+    return validCompareWithMemcmpType(elemType);
+  }
+
+  case Tstruct:
+    // TODO: Implement when structs can be compared with memcmp. Remember that
+    // structs can have a user-defined opEquals, alignment padding bytes (in
+    // arrays), and padding bytes.
+    return false;
+
+  case Tvoid:
+  case Tint8:
+  case Tuns8:
+  case Tint16:
+  case Tuns16:
+  case Tint32:
+  case Tuns32:
+  case Tint64:
+  case Tuns64:
+  case Tint128:
+  case Tuns128:
+  case Tbool:
+  case Tchar:
+  case Twchar:
+  case Tdchar:
+  case Tpointer:
+    return true;
+
+    // TODO: Determine whether this can be "return true" too:
+    // case Tvector:
+  }
+
+  return false;
+}
+
+/// When `true` is returned, `l` and `r` can be compared using `memcmp`.
+///
+/// This function may return `false` even though `memcmp` would be valid.
+/// It may only return `true` if it is 100% certain.
+///
+/// Comparing with memcmp is often not valid, for example due to
+/// - Floating point types
+/// - Padding bytes
+/// - User-defined opEquals
+bool validCompareWithMemcmp(DValue *l, DValue *r) {
+  auto *ltype = l->type->toBasetype();
+
+  // TODO: Remove this check once `DtoArrayEqCmp_memcmp` can handle dynamic
+  // array comparisons.
+  if (ltype->ty != Tsarray)
+    return false;
+
+  auto *rtype = r->type->toBasetype();
+  // Only memcmp equivalent types (memcmp should be used for `const int[3] ==
+  // int[3]`, but not for `int[3] == short[3]`).
+  if (!ltype->equivalent(rtype))
+    return false;
+
+  auto *elemType = ltype->nextOf()->toBasetype();
+  return validCompareWithMemcmpType(elemType);
+}
+
+/// Compare `l` and `r` using memcmp. No checks are done for validity.
+///
+/// This function can currently only deal with comparisons of static arrays
+/// with memcmp.
+/// TODO: Implement dynamic array comparison: length equality check
+/// (and perhaps pointer equality check) before memcmp.
+LLValue *DtoArrayEqCmp_memcmp(Loc &loc, DValue *l, DValue *r, IRState &irs) {
+  IF_LOG Logger::println("Comparing arrays using memcmp");
+  LLFunction *fn = getRuntimeFunction(loc, gIR->module, "memcmp");
+  assert(fn);
+
+  // TODO: Remove this check once dynamic arrays are correctly dealt with.
+  assert(l->type->toBasetype()->ty == Tsarray);
+
+  auto *l_ptr = DtoArrayPtr(l);
+  auto *r_ptr = DtoArrayPtr(r);
+
+  auto *size = DtoArrayLen(l);
+  size_t elementSize = getTypeAllocSize(l_ptr->getType()->getContainedType(0));
+  if (elementSize != 1) {
+    size = irs.ir->CreateMul(size, DtoConstSize_t(elementSize));
+  }
+
+  LLValue *args[] = {DtoBitCast(l_ptr, getVoidPtrType()),
+                     DtoBitCast(r_ptr, getVoidPtrType()), size};
+  return irs.funcGen().callOrInvoke(fn, args).getInstruction();
+}
+} // end anonymous namespace
+
 ////////////////////////////////////////////////////////////////////////////////
 LLValue *DtoArrayEquals(Loc &loc, TOK op, DValue *l, DValue *r) {
   LLValue *res = nullptr;
 
-  // optimize comparisons against null by rewriting to `l.length op 0`
   if (r->isNull()) {
+    // optimize comparisons against null by rewriting to `l.length op 0`
     const auto predicate = eqTokToICmpPred(op);
     res = gIR->ir->CreateICmp(predicate, DtoArrayLen(l), DtoConstSize_t(0));
+  } else if (validCompareWithMemcmp(l, r)) {
+    // Use memcmp directly if possible. This avoids typeinfo lookup, and enables
+    // further optimization because LLVM understands the semantics of C's
+    // `memcmp`.
+    const auto predicate = eqTokToICmpPred(op);
+    const auto memcmp_result = DtoArrayEqCmp_memcmp(loc, l, r, *gIR);
+    res = gIR->ir->CreateICmp(predicate, memcmp_result, DtoConstInt(0));
   } else {
     res = DtoArrayEqCmp_impl(loc, "_adEq2", l, r, true);
     const auto predicate = eqTokToICmpPred(op, /* invert = */ true);
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -323,6 +323,8 @@ static void buildRuntimeModule() {
                                   llvm::Attribute::NoCapture),
       Attr_ReadOnly_NoUnwind_1_NoCapture(Attr_ReadOnly_1_NoCapture, ~0U,
                                          llvm::Attribute::NoUnwind),
+      Attr_ReadOnly_NoUnwind_1_2_NoCapture(Attr_ReadOnly_NoUnwind_1_NoCapture,
+                                           2, llvm::Attribute::NoCapture),
       Attr_ReadNone(NoAttrs, ~0U, llvm::Attribute::ReadNone),
       Attr_1_NoCapture(NoAttrs, 1, llvm::Attribute::NoCapture),
       Attr_NoAlias_1_NoCapture(Attr_1_NoCapture, 0, llvm::Attribute::NoAlias),
@@ -753,6 +755,14 @@ static void buildRuntimeModule() {
       break;
     }
   }
+
+  //////////////////////////////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////////////
+  ////// C standard library functions (a druntime link dependency)
+
+  // int memcmp(const void *s1, const void *s2, size_t n);
+  createFwdDecl(LINKc, intTy, {"memcmp"}, {voidPtrTy, voidPtrTy, sizeTy}, {},
+                Attr_ReadOnly_NoUnwind_1_2_NoCapture);
 }
 
 static void emitInstrumentationFn(const char *name) {
--- a/tests/codegen/array_equals_memcmp.d
+++ b/tests/codegen/array_equals_memcmp.d
@@ -1,0 +1,146 @@
+// Tests that static array (in)equality is optimized to a memcmp call when valid.
+// More importantly: test that memcmp is _not_ used when it is not valid.
+
+// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s --check-prefix=LLVM < %t.ll
+// RUN: %ldc -c -output-s  -of=%t.s  %s && FileCheck %s --check-prefix=ASM  < %t.s
+// RUN: %ldc -O3 -run %s
+
+module mod;
+
+struct ThreeBytes
+{
+    byte a;
+    byte b;
+    byte c;
+}
+
+align(4) struct ThreeBytesAligned
+{
+    byte a;
+    byte b;
+    byte c;
+}
+
+struct Packed
+{
+    byte a;
+    byte b;
+    byte c;
+    byte d;
+}
+
+struct PackedPacked
+{
+    Packed a;
+    Packed b;
+}
+
+struct WithPadding
+{
+    int b;
+    byte a;
+}
+
+// LLVM-LABEL: define{{.*}} @{{.*}}two_uints
+bool two_uints(ref uint[2] a, const ref uint[2] b)
+{
+    // LLVM: call i32 @memcmp({{.*}}, {{.*}}, i{{32|64}} 8)
+    return a == b;
+}
+
+// LLVM-LABEL: define{{.*}} @{{.*}}unequal_two_uints
+bool unequal_two_uints(ref uint[2] a, uint[2] b)
+{
+    // LLVM: call i32 @memcmp({{.*}}, {{.*}}, i{{32|64}} 8)
+    return a != b;
+}
+
+// LLVM-LABEL: define{{.*}} @{{.*}}two_floats
+bool two_floats(float[2] a, float[2] b)
+{
+    // LLVM-NOT: memcmp
+    return a == b;
+}
+
+// LLVM-LABEL: define{{.*}} @{{.*}}four_bools
+// ASM-LABEL: four_bools{{.*}}:
+bool four_bools(bool[4] a, bool[4] b)
+{
+    // LLVM: call i32 @memcmp({{.*}}, {{.*}}, i{{32|64}} 4)
+
+    // Make sure that LLVM recognizes and optimizes-out the call to memcmp for 4 byte arrays:
+    // ASM-NOT: memcmp
+    return a == b;
+}
+
+// LLVM-LABEL: define{{.*}} @{{.*}}array_of_array
+// ASM-LABEL: array_of_array{{.*}}:
+bool array_of_array(byte[3][3] a, const byte[3][3] b)
+{
+    // LLVM: call i32 @memcmp({{.*}}, {{.*}}, i{{32|64}} 9)
+    return a == b;
+}
+
+// LLVM-LABEL: define{{.*}} @{{.*}}int3_short3
+bool int3_short3(int[3] a, short[3] b)
+{
+    // LLVM-NOT: memcmp
+    return a == b;
+    // LLVM-LABEL: ret i1
+}
+
+// LLVM-LABEL: define{{.*}} @{{.*}}pointer3
+bool pointer3(int*[3] a, int*[3] b)
+{
+    // LLVM: call i32 @memcmp({{.*}}, {{.*}}, i{{32|64}} {{12|24}})
+    return a == b;
+}
+
+// LLVM-LABEL: define{{.*}} @{{.*}}enum3
+enum E : char { a, b, c, d, e, f };
+bool enum3(E[3] a, E[3] b)
+{
+    // LLVM: call i32 @memcmp({{.*}}, {{.*}}, i{{32|64}} 3)
+    return a == b;
+}
+
+class K {}
+// LLVM-LABEL: define{{.*}} @{{.*}}klass2
+bool klass2(K[2] a, K[2] b)
+{
+    // LLVM-NOT: memcmp
+    return a == b;
+    // LLVM-LABEL: ret i1
+}
+
+void main()
+{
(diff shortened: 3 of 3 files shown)


gen/arrays.cpp
```diff
@@ -937,14 +938,109 @@ static LLValue *DtoArrayEqCmp_impl(Loc &loc, const char *func, DValue *l,
...
+/// - Padding bytes
+/// - User-defined opEquals
+bool validCompareWithMemcmp(DValue *l, DValue *r) {
+  auto *ltype = l->type->toBasetype();
+
+  // Only static arrays are potentially compared using memcmp.
+  if (ltype->ty != Tsarray)
+    return false;
+
+  auto *elemType = ltype->nextOf()->toBasetype();
```
kinke: There's a function somewhere (I don't recall its name right now) which descends to the first non-static-array element type. By using that, you can get rid of the recursion in the function above.

JohanEngelen: OK, didn't know that.
But the recursion isn't so bad, I think? `validCompareWithMemcmpType` will become recursive again when someone implements the logic for Tstruct, and then the Tsarray logic is also needed there.

gen/arrays.cpp
```diff
@@ -937,14 +938,109 @@ static LLValue *DtoArrayEqCmp_impl(Loc &loc, const char *func, DValue *l,
...
 
-  // optimize comparisons against null by rewriting to `l.length op 0`
   if (r->isNull()) {
+    // optimize comparisons against null by rewriting to `l.length op 0`
     const auto predicate = eqTokToICmpPred(op);
     res = gIR->ir->CreateICmp(predicate, DtoArrayLen(l), DtoConstSize_t(0));
+  } else if (validCompareWithMemcmp(l, r)) {
+    // Use memcmp directly if possible. This avoids typeinfo lookup, and enables
+    // further optimization because LLVM understands the semantics of C's
+    // `memcmp`.
```
kinke: Does it really?

JohanEngelen: LLVM? Yes it does.

JohanEngelen: See `SelectionDAGBuilder::visitMemCmpCall`

gen/arrays.cpp
JohanEngelen: Thanks a lot about the testcase, bad assumption on my part. I'll just check that both lhs and rhs types are exactly the same then?

kinke: Alright, the recursion may really be needed for structs and classes later on.

> I'll just check that both lhs and rhs types are exactly the same then?
I always find this a bit tricky. 'Exactly the same' includes const/immutable modifiers, which don't matter here. There's a stripModifiers() function or so (which already offers recursion, which we need here), but I find that one a bit tedious to use. An idea might be checking the LLVM types: `DtoMemType(l->type) == DtoMemType(r->type)` (e.g., this would work for `char[] == byte[]` and even `char[] == bool[]`, both of these are allowed by the front-end).

gen/arrays.cpp
```diff
@@ -937,14 +938,109 @@ static LLValue *DtoArrayEqCmp_impl(Loc &loc, const char *func, DValue *l,
...
+  case Tchar:
+  case Twchar:
+  case Tdchar:
+    return true;
+
+    // TODO: Determine whether these are "return true" too:
+    // case Tpointer:
+    // case Treference:
+    // case Tenum:
+    // case Tvector:
```
kinke: I'd include `Tpointer` and `Tclass`. For `Tclass`, I guess we could allow memcmp even if the LLVM types of both classes differ (MyClass vs. Object etc.). `Tenum` shouldn't be hit here iirc, as `Type::toBasetype()` lowers it to the underlying integer. `Tvector` would also be nice, as long as the element type is `validCompareWithMemcmpType()` (i.e., not a floating-point vector).

gen/arrays.cpp
JohanEngelen: For when the types are different (`int[3] == short[3]` but also `byte[3] == char[3]`), the front-end lowers the code into a call to (e.g.) `object._ArrayEq!(byte, char)._ArrayEq(byte[], char[])`. So optimizing those will require much more and different work.
I will add a check that the types have to be equivalent (ignoring constness), just in case.

gen/arrays.cpp
JohanEngelen: I've included `Tpointer`. `Tenum` is indeed already covered. For `Tvector`, I'm not sure. We'd need to be sure about the data layout of those things. Leaving it for future work.

JohanEngelen:
`tests\debuginfo\codeview.d` fails on Windows, but I don't see how it is related to this PR (no array comparisons happening in codeview.d afaict). @rainers Do you know what's going on? Thanks.

rainers:
> @rainers Do you know what's going on? Thanks.

Sorry for the trouble. I suspect it is again displaying a different definition of "string" from another DLL or static library. Making the symbol search case sensitive might help. You can switch this by adding a line `// CDB: .symopt-1` before `// CDB: dt string` in codeview.d.

gen/arrays.cpp
```diff
@@ -937,14 +938,113 @@ static LLValue *DtoArrayEqCmp_impl(Loc &loc, const char *func, DValue *l,
...
+/// It may only return `true` if it is 100% certain.
+///
+/// Comparing with memcmp is often not valid, for example due to
+/// - Floating point types
+/// - Padding bytes
+/// - User-defined opEquals
+bool validCompareWithMemcmp(DValue *l, DValue *r) {
+  auto *ltype = l->type->toBasetype();
+
+  // Only static arrays are potentially compared using memcmp.
```
klickverbot: Why not dynamic arrays?

kinke: And maybe a fast-return if the pointers match too, to optimize checks against the same memory.

JohanEngelen: All left for future work. It just adds complexity.

klickverbot: The comment is not very useful, since it exactly replicates the content of the code itself, yet leaves the question as to why unanswered.

rainers:
> You can switch this by adding a line // CDB: .symopt-1 before // CDB: dt string in codeview.d.

Also improved symbol loading in https://github.com/ldc-developers/ldc/pull/1731

kinke:
AppVeyor jobs retriggered.

kinke:
Wow, something bad is happening here. In master, `tests/debuginfo/codeview.d` takes 1:43 mins for [x64](https://ci.appveyor.com/project/kinke/ldc/build/1.0.2261/job/gki9he10xcysyr3e#L6299) and 0:23 for [x86](https://ci.appveyor.com/project/kinke/ldc/build/1.0.2261/job/50lx6o2l4er2o7us#L6282) [when hovering over an AppVeyor log line, a tooltip pops up with a timestamp).

With this PR, the [x86](https://ci.appveyor.com/project/kinke/ldc/build/1.0.2262/job/6li1pkx80m2bw4ud#L6286) test takes a massive 6:21, and the x64 job times out after more than 26 minutes for that single test! Pinging @rainers.

rainers:
> Pinging @rainers.

I suspect cdb is trying to load symbols from the MS symbol servers. Passing `-snul` (disables automatic symbol loading for unqualified names) to the cdb command line might help.

gen/arrays.cpp
JohanEngelen: I've improved the comments.

JohanEngelen:
Now the Windows unittest fails on a uuid.d assert, line 944. I've looked at the IR (Mac and Windows), but for that particular comparison, memcmp is not used (it is a dyn array comparison).
Anyone know what's going on?

kinke:
AppVeyor is very strange today. For your x64 run:
```
1/1 Test #957: lit-tests ........................   Passed  1230.67 sec
```

`tests/codegen/inlining_templates.d` takes 17 minutes. For my LLVM 3.9 final PR, it even took almost 25 minutes. So I'm not sure AppVeyor can be trusted today...

rainers:
> tests/codegen/inlining_templates.d takes 17 minutes

I suspect that it is again the next test that takes so long: codeview.d ;-( cvbasictypes.d also too more than 3 minutes.

rainers:
> It's the debuginfo tests with cdb that take forever (unfortunately, on my box too).

Can you run them without redirecting the output and see what it is doing? The only issue I've seen as a cause for slowdoen is loading symbols from the symbol servers.

rainers:
> But that should be disabled in master now?

Yes, that's what I thought. With cdb from the Windows 10 SDK, I also see a pause (less than a minute) when loading the symbols of the executable explicitely. That doesn't happen for cdb from the Windows 8.1 SDK.

rainers:
> Do you mean, due to network traffic? 30s or whatever it is still seems awfully long for loading symbols for an executable.

Yes. It is looking for codeview.exe`s symbols on the MS symbol servers, too. I'm trying to disable that...

rainers:
I hope this helps: https://github.com/ldc-developers/ldc/pull/1743

kinke:
> Anyone know what's going on?

Nope. I can reproduce it here though, it starts with `-O2`. It works fine if the unittest is isolated in its own module. But a printf in the original source right before the assertion reveals that `str` (the runtime array) is all set to nulls...
```
str = 0x000000C2199FF4E0 (00000000-0000-0000-0000-000000000000)
s   = 0x00007FF7276097B0 (8ab3060e-2cba-4f23-b74c-b52db3bdfb46)
```

kinke:
Well, it's not all nulls, the dashes are part of the string. This is more complete:
```D
// printf("id = 0x%p 0x%p\n", id.ulongs[0], id.ulongs[1]);
id = 0x234FBA2C0E06B38A 0x46FBBDB32DB54CB7
// printf("str = 0x%p (%.36s)\ns   = 0x%p (%.36s)\n", str.ptr, str.ptr, s.ptr, s.ptr);
str = 0x000000C43AAFF710 (00000000-0000-0000-0000-000000000000)
s   = 0x00007FF60AA297B0 (8ab3060e-2cba-4f23-b74c-b52db3bdfb46)
```

So it looks as if the `this` pointer in `id.toString(str[])` (`id` is an `enum struct std.UUID`) could point to something else with all zeros.

kinke:
Well, this could very well be another symptom of https://github.com/ldc-developers/ldc/issues/1324, as we have a compile-time instance of a struct with a union again.

JohanEngelen:
I don't know what's going on here :(
I can reproduce it too on Windows, but with very slight changes, the bug is gone. Definitely something with unions+CTFE. I can fix Phobos so the bug is gone but... :'( :'(
It is also very strange that it depends on the LLVM optimization level.

kinke:
AppVeyor retriggered after merging https://github.com/ldc-developers/ldc/pull/1846 and also verified locally - no change unfortunately, `std.uuid:944` still asserts starting with `-O2` on Win64. :(

JohanEngelen:
:(

JohanEngelen:
@kinke Could you try to dustmite the std.uuid failure issue? Thanks!

kinke:
On it. A good excuse to get started with dustmite. :]

kinke:
;) - and the first experience isn't exactly pleasant:

```D
import std.range;
struct UUID
{
    import std.meta ;

unittest
        {
            import std.conv ;

            foreach(S; AliasSeq!(char[]))
                //Test valid, working cases
                assert(to!S("00000000-0000-0000-0000-000000000000").empty);

        }

}
```

It got rid of the conversion of the char[] slice to a UUID and invoking `empty` on that UUID instance (original: `assert(UUID(to!S("00000000-0000-0000-0000-000000000000")).empty);`). This obviously asserts with all valid D compilers. And it even managed to create a version crashing at runtime - a problem on Windows, where a dialog pops up, and dustmite is suspended, so that one really has to sit in front of the computer and babysit the process.

kinke:
Alright I guess I have to hack in a printf and check for that one instead.

kinke:
Thanks David. After quite a while, here's a reduced testcase:

```D
import std.range;
import std.traits;

struct UUID
{
    import std.meta ;

        alias skipSeq = AliasSeq!(8, 13, 18, 23);
        alias byteSeq = AliasSeq!(0,2,4,6,/++/9,11,/++/14,16,/++/19,21,/++/24,26,28,30,32,34);

        @safe pure nothrow @nogc Char toChar(Char)(size_t i) const
        {
            if(i <= 9)
                return cast(Char)('0' + i);
            else
                return cast(Char)('a' + (i-10));
        }
unittest
        {
UUID([138, 179, 6, 14, 44, 186, 79, 35, 183, 76, 181, 45,
                179, 189, 251, 70]);
        }

        union
        {
            ubyte[16] data;
        }

@nogc this(ubyte[] )
        {
        }

        this(T)(in T[] uuid) if(isSomeChar!(Unqual!T))
        {
            import std.conv : to, parse;
            if(uuid.length != 36)
                throw new Exception("abc");
            enum skipInd = [skipSeq];
            foreach(pos; skipInd)
                assert(uuid[pos] == '-');

            ubyte[16] data2; //ctfe bug
            uint pos = void;

            foreach(i, p; byteSeq)
            {
                enum uint s = 'a'-10-'0';
                uint h = uuid[p];
                uint l = uuid[p+1];
                pos = p;
                if(h < '0') goto Lerr;
                if(l < '0') goto Lerr;
                if (h > '9')
                {
                    h |= 0x20; //poorman's tolower
                    if (h < 'a') goto Lerr;
                    if (h > 'f') goto Lerr;
                    h -= s;
                }
                if (l > '9')
                {
                    l |= 0x20; //poorman's tolower
                    if (l < 'a') goto Lerr;
                    if (l > 'f') goto Lerr;
                    l -= s;
                }
                h -= '0';
                l -= '0';

                data2[i] = cast(ubyte)((h << 4) ^ l);
            }
            this.data = data2;
            return;

        Lerr: throw new Exception("def");
        }
unittest
        {
            import std.exception;
            import std.conv ;

collectException!Exception(
                    UUID(to!(wchar[])("5668122d-9df0-49a4-ad0b-b9b0a57f886")));
        }

        @safe pure nothrow @nogc void toString(Range)(Range result) const
        {
            alias Char = typeof(Range.init[0]);
            foreach(pos; skipSeq)
                result[pos] = cast(Char)('-');
            foreach(i, pos; byteSeq)
            {
                const uint entry = this.data[i];
                const uint hi = entry >> 4;
                result[pos  ] = toChar!Char(hi);
                const uint lo = (entry) & 0x0F;
                result[pos+1] = toChar!Char(lo);
            }
        }

        nothrow @nogc unittest
        {
            import std.encoding: Char = AsciiChar;
            enum  utfstr = "8ab3060e-2cba-4f23-b74c-b52db3bdfb46";
            alias String = immutable(Char)[];
            enum String s = cast(String)utfstr;
            enum id = UUID(utfstr);
            //nogc
            Char[36] str;
            id.toString(str[]);
            assert(str == s);
        }
}
```

Works with DMD, consistently asserts in the bottom-most assert for LDC with this PR. Needs both prior unittests and -O3 (-O2 passes!). May be an LLVM bug.

JohanEngelen:
I've been able to cut it down some more, and found an interesting requirement for it to fail.
The UUID of the first unittest block has to be the same as the UUID in the last block!

```d
// RUN: %ldc -unittest -release -boundscheck=off -main -run %s
// RUN: %ldc -unittest -release -boundscheck=off -main -O3 -run %s

import std.traits;

struct UUID
{
    char toChar(size_t i) const
    {
        return cast(char)('0' + i);
    }

    ubyte[16] data;

    @nogc this(ubyte[])
    {
    }

    this(T)(T[] uuid) if (isSomeChar!(Unqual!T))
    {
        if (uuid.length != 16 * 2)
            throw new Exception("abc");

        ubyte[16] data2 = void;

        foreach (i; 0 .. 16)
        {
            enum uint s = 'a' - 10 - '0';
            uint h = uuid[i * 2];
            uint l = uuid[i * 2 + 1];
            h -= '0';
            l -= '0';

            data2[i] = cast(ubyte)((h << 4) ^ l);
        }
        this.data = data2;
    }

    unittest
    {
        UUID([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]); // THIS HAS TO BE THE SAME AS BELOW
    }

    unittest
    {
        import std.exception;
        import std.conv;

        collectException!Exception(UUID(to!(wchar[])("")));
    }

    void toString(char[] result)
    {
        import std.meta;

        foreach (i, pos; AliasSeq!(0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30))
        {
            const uint entry = this.data[i];
            const uint hi = entry >> 4;
            const uint lo = entry & 0x0F;
            result[i * 2] = toChar(hi);
            result[i * 2 + 1] = toChar(lo);
        }
    }

    unittest
    {
        enum utfstr = "01010101010101010101010101010101"; // THIS HAS TO BE THE SAME AS ABOVE
        enum id = UUID(utfstr);
        char[32] str = void;
        id.toString(str[]);
        assert(str == utfstr);
    }
}
```

kinke:
Interesting. What I also found interesting is the comment `ctfe bug` in the string ctor which is apparently the reason for populating a local `data2` array and then copying that over to the field in the end.

JohanEngelen:
@uplinkcoder Because you know so much about CTFE code:  can you have a look at try to figure out why this testcase fails with `-O3`? https://github.com/ldc-developers/ldc/pull/1719#issuecomment-259936550
Thanks a lot.

UplinkCoder:
@JohanEngelen I'll have a look

UplinkCoder:
@JohanEngelen I know what is going on.
Enums do not have storage.
therefore you are comparing an invalid pointer.
change it to a static immutable and the troubles are gone.

JohanEngelen:
@UplinkCoder So a Phobos bug?

UplinkCoder:
is this phobos code ?
the trouble is here:
`
enum utfstr = "01010101010101010101010101010101"; // THIS HAS TO BE THE SAME AS ABOVE
enum id = UUID(utfstr);

// utfstr does not have an address.
// or rather there is one but it might be invalid.
`

JohanEngelen:
@UplinkCoder https://github.com/dlang/phobos/blob/master/std/uuid.d#L936

UplinkCoder:
@JohanEngelen I see. you should submit a phobos pr.

JohanEngelen:
@UplinkCoder Could you?  I actually don't understand what the bug is :S

kinke:
> So it looks as if the this pointer in `id.toString(str[])` (`id` is an `enum struct std.uuid.UUID`) could point to something else with all zeros.

> Enums do not have storage.
> or rather there is one [address] but it might be invalid.

So if I understand correctly and that can't be fixed, CTFE should be disallowed for all methods, as `this` may point to garbage.

UplinkCoder:
@kinke yes the enum string.

UplinkCoder:
ctfe does not allocate storage for enums.
the pointers that come back are bogus.
https://issues.dlang.org/show_bug.cgi?id=15989

kinke:
Okay, so the bug is that the runtime `this` pointer of struct `enum UUID id` is garbage, this explains the original issue. And it's already supposed to be disallowed upstream since 2.073. But... the unittest is unchanged in Phobos master... %)

UplinkCoder:
@klickverbot You know certainly alot more about ldc internals then I do (I know nothing about them).
@JohanEngelen If this were an issue with ctfe it should not be affected by the optimization level.
Maybe the optimizer assumes some things that are not true ?

kinke:
The unoptimized IR (I've used the merge-2.072 branch, this PR not merged in) looks absolutely fine:

```D
import std.uuid;

void main()
{
    import std.encoding : Char = AsciiChar;
    enum  utfstr = "8ab3060e-2cba-4f23-b74c-b52db3bdfb46";
    alias String = immutable(Char)[];
    enum String s = cast(String)utfstr;
    enum id = UUID(utfstr);
    Char[36] str;
    id.toString(str[]);
    assert(str == s);
}
```

```LL
%std.uuid.UUID = type { [16 x i8] }

; data for `enum UUID id`
@.arrayliteral = internal unnamed_addr constant [16 x i8] c"\8A\B3\06\0E,\BAO#\B7L\B5-\B3\BD\FBF" ; [#uses = 1]
; `enum String s`, used in the comparison
@.str = private unnamed_addr constant [37 x i8] c"8ab3060e-2cba-4f23-b74c-b52db3bdfb46\00" ; [#uses = 1]

  %str = alloca [36 x i8], align 1                ; [#uses = 4, size/byte = 36]
  ; enum UUID id
  %.structliteral = alloca %std.uuid.UUID, align 8 ; [#uses = 2, size/byte = 16]
  ; zero-initialize `str`
  %1 = bitcast [36 x i8]* %str to i8*             ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %1, i8 0, i64 36, i32 1, i1 false)
  ; [...]
  ; initialize single `id` field via memcpy from @.arrayliteral
  %4 = getelementptr inbounds %std.uuid.UUID, %std.uuid.UUID* %.structliteral, i32 0, i32 0 ; [#uses = 1, type = [16 x i8]*]
  %5 = bitcast [16 x i8]* %4 to i8*               ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %5, i8* getelementptr inbounds ([16 x i8], [16 x i8]* @.arrayliteral, i32 0, i32 0), i64 16, i32 1, i1 false)
  ; str[]
  %6 = bitcast [36 x i8]* %str to i8*             ; [#uses = 1]
  %7 = insertvalue { i64, i8* } { i64 36, i8* undef }, i8* %6, 1 ; [#uses = 1]
  ; id.toString(str[])
  call void @_D3std4uuid4UUID39__T8toStringTAE3std8encoding9AsciiCharZ8toStringMxFNaNbNiNfMAE3std8encoding9AsciiCharZv(%std.uuid.UUID* nonnull %.structliteral, { i64, i8* } %7) #0
  ; str[]
  %8 = bitcast [36 x i8]* %str to i8*             ; [#uses = 1]
  %9 = insertvalue { i64, i8* } { i64 36, i8* undef }, i8* %8, 1 ; [#uses = 1]
  ; str == s
  %10 = call i32 @_adEq2({ i64, i8* } %9, { i64, i8* } { i64 36, i8* getelementptr inbounds ([37 x i8], [37 x i8]* @.str, i32 0, i32 0) }, %object.TypeInfo* bitcast (%"typeid(AsciiChar[])"* @_D34TypeInfo_AE3std8encoding9AsciiChar6__initZ to %object.TypeInfo*)) #2 ; [#uses = 1]
```

My results back then via printf clearly indicated that the run-time data of `id` was garbage/zero-initialized. The `this` pointer of `id` (the `%.structliteral` alloca) is fine, and it's initialized. I don't think the LLVM optimizer chokes on something as simple as this. So it's very likely that the problem lies somewhere in the front-end when constructing `enum id = UUID(utfstr);` at compile-time and ending up with the (parsed) 16 bytes of UUID runtime data.

kinke:
For the record: I'm absolutely eager to merge this and am in favor of just using `static immutable id = UUID(utfstr);` for LDC to work around this apparently invalid or at least problematic unittest.

kinke:
This PR doesn't affect the problematic unittest itself IR-wise; two slices are compared, and this PR only supports static arrays so far.

JohanEngelen:
Shall I modify Phobos and merge this?
(and open a new issue with the reduced problematic testcase from this PR, so we can investigate there what is going on?)

JohanEngelen:
Yeah... this is a strange bug. I'm not sure whether it is a miscompile or a unittest bug. My current feeling is that it is a unittest bug, exposed by aggressive optimization enabled by this PR.

JohanEngelen:
green after phobos modification.

kinke:
Stefan suggested a `static immutable` instead of `auto`, did you try that?

JohanEngelen:
With `static immutable` we don't have to change the `nothrow @nogc`. So if that works, great. Updated the PR.

JohanEngelen:
Travis is not retriggering :(

kinke:
Yay, finally. ;)

I don't think it'll make a huge impact in the current form. As soon as slices are supported (should be straight foward), that will definitely change, and I actually expect noticeable performance improvements for client code (incl. synthetic benchmarks, where LDC and D in general could even shine a bit more). And we should work on supporting suited structs soon too.

PR #1722, opened 2016-08-27
Streamline main() with DMD's (ddmd/mars.d)

kinke:
Customizing `ddmd/mars.d` for LDC would probably make sense to avoid getting out of sync in the future.


ddmd/mars.d
```diff
@@ -1211,18 +1248,40 @@ Language changes listed by -transition=id:
         {
             /* Deduce what to do with a file based on its extension
              */
+          version (IN_LLVM)
+          {
+            if (FileName.equals(ext, global.obj_ext) ||
+                (TARGET_WINDOS && FileName.equals(ext, global.obj_ext_alt)))
```
kinke: I've translated the C++ `LDC_POSIX` define to `enum bool TARGET_WINDOS = xversion!'Windows'`.

ddmd/mars.d
```diff
@@ -1243,11 +1303,14 @@ Language changes listed by -transition=id:
...
             }
+          version (IN_LLVM) {} else
+          {
             if (FileName.equals(ext, global.map_ext))
             {
                 global.params.mapfile = files[i];
                 continue;
             }
+          }
             static if (TARGET_WINDOS)
```
kinke: Same here.

kinke:
-250 lines, minimized chance of missing DMD changes to its `tryMain()` and 2 resolved issues - it would be cool if this made it into 1.1-b3. ;)

ddmd/mars.d
klickverbot: This seems wrong – the TARGET_* versions are about the target, not the host platform.

kinke: Yep, and that didn't seem right. If I'm targeting Windows, I'll care about the `.obj` file extension, but not if I'm cross-compiling to Linux on a Windows host.

klickverbot: In any case, there is no place for compile-time TARGET_* values in LDC as we are a cross-compiler by default.

kinke: It was compile time already, guarded by the LDC_POSIX define. So no functional change here. I'm currently evaluating getting rid of `global.obj_ext_alt` altogether.

ddmd/mars.d
```diff
@@ -1142,21 +1155,39 @@ Language changes listed by -transition=id:
...
+extern (C++) int mars_mainBody(ref Strings files, ref Strings libmodules,
+        bool createStaticLib = false, bool createSharedLib = false, bool staticFlag = false)
+{
     // Initialization
     Type._init();
     Id.initialize();
     Module._init();
     Target._init();
     Expression._init();
+  version (IN_LLVM) {} else
```
klickverbot: It seems like should just rename our method to `objc_init` too?

ddmd/mars.d
```diff
@@ -1195,14 +1226,20 @@ Language changes listed by -transition=id:
     // Create Modules
     Modules modules;
     modules.reserve(files.dim);
+  version (IN_LLVM) {} else
+  {
     bool firstmodule = true;
```
klickverbot: I'd rather err on the side of minimal changes and leave this in – just more chances for irrelevant merge conflicts in the future. Same goes for `libmodules`.

ddmd/mars.d
klickverbot: > So no functional change here. 

Sure. But TARGET_* is still the wrong set of compile-time constants to use, even if they might currently be set the same in the DMD source code.

ddmd/mars.d
kinke: Alright. It should be pretty obvious that we can build from here and later streamline output file names generation as well, there are numerous issues about differences compared to DMD wrt. output dirs, object and executable file names etc.

ddmd/mars.d
kinke: I didn't want to check whether our objc_init() relies on being called after `builtin_init()`. ;)

gen/objcgen.cpp
```diff
@@ -122,9 +115,12 @@ bool objc_isSupported(const llvm::Triple &triple) {
   return false;
 }
 
-void objc_init() {
-  initSymbols();
-  ObjcSelector::_init();
```
kinke: The DMD version `ddmd.objc.objc_tryMain_init` is defined as:
```D
objc_initSymbols();
ObjcSelector._init();
```
(and `ObjcSelector` is the same D struct)

Issue #1723, opened 2016-08-28
ICE: toConstElem(CastExp) doesn't support NewExp as cast source


antoniocorbibellot:
Usin ldc-git ( based on DMD v2.071.2-b2 and LLVM 3.8.1  built with DMD64 D Compiler v2.071.1) to compile this code ends up in an ICE:

```
interface IBase {}

class Impl(T): IBase {
    T value;
}
void main() {
    IBase x = new Impl!uint;
    IBase y = new Impl!string;
    IBase z = new Impl!double;

    IBase[string] fldFormats1 = [
        "Key": x,
        "Name": y,
        "Percent": z
    ];

    IBase[string] fldFormats2 = cast(IBase[string])[
        "Key": new Impl!uint,
        "Name": new Impl!string,
        "Percent": new Impl!double
    ];
}
```
It also happens with ldc-1.0.0 (LDC - the LLVM D compiler (1.0.0):  based on DMD v2.070.2 and LLVM 3.8.0)

kinke:
Confirmed. `toConstElem()` doesn't support new-expressions when casting the new `Object` (line 18, instead of `Impl!uint`...) to `IBase`.

kinke:
Correction: `toConstElem()` doesn't support `NewExp`ressions at all, we currently rely on `ClassReferenceExp`ressions apparently.

klickverbot:
CTFE'd class instances should consistently be lowered to ClassReferenceExp. NewExp is for runtime `new` calls (which of course can never be supported in `toConstElem`)

Could this just be an issue where the toConstElem code elsewhere doesn't handle non-constant input correctly? Remember that the AA literal lowering code currently tries to do `toConstElem` with errors gagged, and if it fails, falls back to runtime arrays.

kinke:
I just figured. Fix is on its way.

kinke:
Fixed in master.

PR #1725, opened 2016-08-28
Fix issue #1723

kinke:
Errors are gagged while trying to create a constant initializer for an associative array literal. So don't die with an assertion if the cast source is a NewExp (or something else than a supported ClassReferenceExp),
emit a gaggable error instead.

A dmd-testsuite test will follow once merged, as usual.


kinke:
Yep, a higher-level overview would be very helpful. A Wiki page would be a good start I guess.

Issue #1724, opened 2016-08-28
ldmd2 ignores -od option for libraries


croizier:
When you specify an object folder with the -od option, ldmd2 puts .o files there but not .a files, unlike dmd and what ldmd2 documentation says: "-odobjdir      write object & library files to directory objdir".

kinke:
Resolved by #1726. This eliminated some (ignored) errors for dmd-testsuite bash scripts removing static libs manually.

Issue #1666, opened 2016-08-02
LDC enforces suffix for `-of` option to generate object files


tom-tan:
I reproduce it with LDC 1.0.0 on Linux 64bit systems.
I also see a similar issue when I try to make a cross compiler whose target is Windows system and whose host is Linux system.

The following command generates `foo.o` instead of `foo.obj`.

```
$ ldc2 -c sample.d -offoo.obj
```

Also, the following command generates `foo.i_want_this_suffix.o` instead of `foo.i_want_this_suffix`.

```
$ ldc2 -c sample.d -offoo.i_want_this_suffix
```

kinke:
Fixed by #1726.

tom-tan:
Thanks!

JohanEngelen:
No tests for this though.

Issue #1203, opened 2015-11-09
ldmd2 does not default to same executable name as DMD


ufm:
Simple test:
File "a.d" contained this:

```
module tst;
import std.stdio;
void main() {
     writeln("Hello world");
}
 ```

After compiling this file by "dmd a.d" I get executable with name "a".
After compiling by "ldc2 a.d" - executable with name "tst".

Comment by Kai Nacke on dlang forum :
> ldc2 does not have the goal to be compatible in all aspects with dmd.
> So this behaviour is ok. But: ldmd2 is meant to be compatible with dmd.
> This tool has the same problem which is clearly a bug.

WBR,
    Fyodor.

kinke:
Both ldmd2 and ldc2 now use DMD's name after #1726 and #1771.

Issue #455, opened 2013-08-25
Error: multiple source files, but only one .obj name


timotheecour:
When using -o-, I shouldn't get errors asking for -oq, since I'm not requesting object files

$ldc_X -oq -of=main -I=. main.d foo.d bar/foo.d =>works

$ldc_X -of=main -I=. main.d foo.d bar/foo.d
Error: Output file 'foo.o' for module 'bar.foo' collides with previous module 'foo'. See the -oq option

ldc_X -o- -of=main -I=. main.d foo.d bar/foo.d
Error: multiple source files, but only one .obj name

----files:

cat bar/foo.d
module bar.foo;
void fun(){}

cat foo.d
public import bar.foo;

cat main.d
import foo;
import bar.foo;
void main(){fun;}

kinke:
Fixed by #1771.

PR #1726, opened 2016-08-30
Streamline output filenames with DMD

kinke:


ddmd/dmodule.d
```diff
@@ -1345,45 +1358,10 @@ else
...
-                    memcpy(s + len + 1, ext, extlen + 1);
-                    s[len + 1 + extlen] = 0;
-                    argobj = s;
-                }
-            }
-
-            if (!FileName.absolute(argobj)) {
-                argobj = FileName.combine(path, argobj);
-            }
+        void check_and_add_output_file(File* file);
```
klickverbot: Why no camel caps?

kinke:
The 5th commit is debatable. I personally dislike the current behavior of considering the `module xxx;` declaration of the first source file instead of the first source filename (e.g., when copying a file to make tiny changes, compiling and ending up overwriting the original exe, although I wanted to compare both), and there's at least one open issue about that too.

driver/main.cpp
```diff
@@ -640,16 +640,34 @@ static void parseCommandLine(int argc, char **argv, Strings &sourceFiles,
...
   }
 
-  if (global.params.link && !global.params.dll) {
+  if (global.params.link) {
     global.params.exefile = global.params.objname;
-    if (sourceFiles.dim > 1) {
-      global.params.objname = nullptr;
+    global.params.oneobj = true;
+    if (global.params.objname) {
+      /* Use this to name the one object file with the same
```
klickverbot: Nitpick: The vast majority of our code uses C++-style comments.

kinke: That whole part here is an almost 1:1 copy of DMD's code (just excluding setting `global.params.multiobj` for coverage builds IIRC), just clang-formatted.

kinke:
The 4th commit is also worth pointing to, as it enforces `-singleObj` for cases where DMD does it implicitly too. I seem to recall David suggested enabling it by default a while ago; that patch enforces it when outputting executables/shared libraries or when compiling multiple source files in combination with a single object output file. It also gets rid of issue #455.

kinke:
There are more subtle changes too. E.g., I think https://github.com/ldc-developers/ldc/issues/1666 should be fixed too by this, as well as https://github.com/ldc-developers/ldc/issues/1724. The `-op` switch to preserve object file paths is now implemented (it had no effect previously IIRC).

kinke:
Wrt. the inferred output name, there's issue #1203. We could only fix it on the LDMD side (by specifying `-of=<firstSourceFile.extension>` if not already provided by the user), but I'd rather change LDC itself here. I'd be okay with splitting the last commit out into another PR, but I don't think waiting makes much sense. We 'just' need consensus that taking the root module into account is not very intuitive and that it's better to change it as soon as possible, before LDC gets real traction.

driver/main.cpp
kinke: So I'd rather leave this as-is, exactly to recognize it as code originating from DMD ;) and to simplify diffing.

kinke:
I only wanted to express optimism wrt. a bright future for D and LDC (v1.1 will rock) but that we're still at an early stage though, where I'd consider these breaking changes wrt. command-line semantics to be allowed.
It'd surely make sense to check that dub, VisualD etc. still work (but I doubt they care to parse the first source file to infer LDC's automatically generated output, they probably had to explicitly specify an output file precisely for this reason).

kinke:
To recap, there are 2 major changes to the command-line semantics to be agreed on:

* `-singleObj` is now enforced (overriding the command-line option) when DMD would do it, i.e., when
  * specifying a single output object file but multiple source files [makes perfect sense IMO]
  * building an executable or a shared library (not for static libraries though) [=> Cheap inter-module optimizations across the specified source files - why not do so if the user compiles all files at once? We could allow an explicit `-disable-singleObj` to override the default here though, if people really care about separate object files.]
* The name of the produced object/library/executable, if not given explicitly via `-of`, is now based on the first source file, just like DMD does. This allows for a smoother switch from DMD to LDC but may break existing build scripts. I'm pretty sure proper build systems like dub and VisualD specify `-of` anyway, so I'd think only a few manual makefiles/scripts would be affected if the first source file features a `module <bla>;` declaration AND that module name doesn't correspond to the filename. Mentioning this in the change log/announcements should be enough IMO.

So please share your thoughts @redstar @JohanEngelen @rainers and whoever reads this. :)

kinke:
An example why I prefer DMD's inferred exe filename: my working directory is littered with .d testcases already, so whenever I wanna test something new, I simply reuse a `current.d` file. Now if the pasted code contains a `module` declaration, I need to manually delete that line, in order to compile and then run that thing with `ldc2 current.d && current` (I don't use `-run` as I prefer having the executable, just in case).

JohanEngelen:
I don't have a strong opinion on this, other than that non-singleobj builds are funky anyway (they always bug me e.g. when working on cross-module inlining).

rainers:
IMO using the module name for the executable is pretty unintuitive and inaccessable to a build tool (e.g. it would have to understand all kind of comments). Using a fixed name (as clang's a.exe) is still better than the module name. Obviously Visual D uses -of anyway, even though it knows how to parse the module (and has to figure the module name for other functionality).

I think the first object file is fine, it is also what the Microsoft compiler does.

JohanEngelen:
What was missing is a testcase! The bug would have quickly been found:
instead of using the _first_ D source file to infer the output filename, the _last_ one is used. It's probably because we do codegen in reverse order.

kinke:
Back from vacation; I'll check the source file ordering, close the 5-6 resolved issues and post something in the forum/NG.

kinke:
> It's probably because we do codegen in reverse order.

Yep, that's it. I don't really like that reversed order, as there's other logic depending on regular order (e.g., `CodeGenerator::firstModuleObjfileName_`, and bitcode files are injected into first processed module etc.). Based on your comment, there's no simple alternative, so I'll probably populate the `objfiles` array in a prior pass.

JohanEngelen:
I am rethinking the x-module inlining. It is too hackish the way it's done now (which is the sole reason for the reversed order). Hope to get rid of it in the not-too-distant future.

Issue #1716, opened 2016-08-21
Incomplete CodeView debug information


rainers:
I've compiled my cv2pdb test file with LDC for Win64 and stepped through it in VS2015.
This is a list of problems I noticed:

- [ ] void pointer: error shown
- [ ] ubyte,byte: error shown
- [ ] class/struct: no members shown
- [ ] class/struct: static members not shown
- [ ] UDT: unique name missing (clang writes it)
- [ ] array: no type name
- [ ] const void[] cvarr; const(char)[], const char[]: error shown (immutable ok)
- [ ] lexical scope: variable life times set, "optimized away"
- [ ] closure vars: show nonsense
- [ ] class: base class not shown, base class members merged
- [ ] assoc array: displays as unsigned char*
- [ ] template class: arguments not in type name
- [ ] ifloat/cfloat: error shown

I've made this a list of checkable items as I'll try to work on these and track progress here.

kinke:
Sounds like a very good plan. :+1:
My understanding was that it's only a matter for LLVM to more or less fully support translating the currently emitted (incomplete) debug info to a .pdb, and that LDC itself wouldn't have to be adapted to use a different set of LLVM DebugInfo classes. So I'm keen on seeing how things will work out.

rainers:
Yeah, the debug support inside LLVM looks like a DWARF to CodeView converter. It still needs some sensible input :)

kinke:
Thanks, nice progress!

> x64: struct value arguments implicitly passed by reference show wrong pointer

This one's new, right? And a major one. Does it show the source pointer, i.e., the LL argument before rewriting it via `ExplicitByvalRewrite`, which makes a bitcopy on the caller's stack and then passes a pointer to that copy as real parameter? IIRC, `IrParam::value` should be the correct pointer.

rainers:
> This one's new, right?

Yes. For `main(string[] args)` cdb shows `struct string[] * args = 0x00000000 00000001`. Not sure how this relates to the rewrite.

kinke:
Ah. Delegates and slices currently aren't rewritten, we'd need to rewrite runtime function calls first. So in this case, the slice is passed as 2x64-bit aggregate, i.e., in the first 2 registers. So you see the array length instead of the pointer, which would be in the next register/dumped to memory right after the length field.

kinke:
Oh I think the intent was to display it as a struct with both fields...

rainers:
[...]

PR #1730, opened 2016-09-01
CodeView: add support for imaginary and complex types

rainers:
Maps imaginary types to the floating point types, creates a struct for complex types.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -180,11 +180,22 @@ ldc::DIType ldc::DIBuilder::CreateBasicType(Type *type) {
   case Timaginary32:
   case Timaginary64:
   case Timaginary80:
+    if (global.params.targetTriple->isWindowsMSVCEnvironment()) {
+      // DW_ATE_imaginary_float not supported by the LLVM DWARF->CodeView
+      // conversion
+      Encoding = DW_ATE_float;
+      break;
+    }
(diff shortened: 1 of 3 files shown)


gen/dibuilder.cpp
```diff
@@ -262,6 +273,46 @@ ldc::DIType ldc::DIBuilder::CreateVectorType(Type *type) {
...
+        break;
+    default:
+        llvm_unreachable(
+            "Unexpected type for debug info in DIBuilder::CreateComplexType");
+    }
+    ldc::DIFile file = CreateFile();
+
+    LLMetadata *elems[] = {
+        CreateMemberType(0, elemtype, file, "re", 0, PROTpublic),
+        CreateMemberType(0, elemtype, file, "im", t->alignsize(), PROTpublic)};
```
kinke: I think there's a chance that `t->alignsize() != getTypeAllocSize(T)` for 80-bit `real`. So I'd suggest using `getTypeAllocSize(T)` here, like below.

rainers: Thanks, changed that. alignsize was even called on the wrong type.
Also removed the failing assert left over from copy 'n paste.

kinke:
Travis and CircleCI retriggered.

rainers:
Thanks, another one checked off the list at https://github.com/ldc-developers/ldc/issues/1716 (but a different one added, too).

Issue #1715, opened 2016-08-21
Mark runtime intrinsic shims as pragma(inline, true)


klickverbot:
This is one of the few places where I'd actually argue for using the pragma – emitting a function call for those would actually break user expectations.

kinke:
Which functions do you have in mind here?

klickverbot:
The core.bitop/… etc. function bodies that are just wrappers around the LLVM intrinsics.

PR #1733, opened 2016-09-02
druntime: Resolve issue #1715 (force-inline intrinsic shims)

kinke:


kinke:
;)

tests/codegen/inlining_stdlib.d
```diff
@@ -2,7 +2,6 @@
 
 // REQUIRES: atleast_llvm307
 
-// Test also that the tested functions are indeed not inlined at -O0 (basically verifying that we are testing something real)
```
JohanEngelen: ???

tests/codegen/inlining_stdlib.d
```diff
@@ -12,19 +11,20 @@ extern (C): // simplify mangling for easier matching
 // OPT3-LABEL: define{{.*}} @foo(
 int foo(size_t i)
 {
+    // core.bitop.bsf() is force-inlined
```
JohanEngelen: So we have to find another function that is not inlined, right? Otherwise the test is useless.

tests/codegen/inlining_stdlib.d
JohanEngelen: OK. Changing the intent of the tests is done a bit lightly, but OK.

tests/codegen/inlining_stdlib.d
JohanEngelen: OK, we still have the non-inlined call to fromStringz.

PR #1735, opened 2016-09-03
Automatically generate debug/shared stdlib names

Dicebot:
Replaces https://github.com/ldc-developers/ldc/pull/1305 by
implementing proposal from @kinke


kinke:
Continued in #2044.

PR #1738, opened 2016-09-04
[CircleCI] Use LLVM 4.0 from APT

JohanEngelen:
Now that LLVM 3.9 is released, we can use Travis and AppVeyor for all "serious" testing: they must be green. And we can use CircleCI for LLVM 4.0 (trunk) testing that may break e.g. upon LLVM API changes. This way, we get notified early about relevant LLVM trunk changes, and can react with easy small LDC patches.
Makes sense?

Also see https://forum.dlang.org/thread/nwajbeceaokohknjhlqb@forum.dlang.org

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,7 +9,6 @@ dependencies:
     #- clang+llvm-3.8.0-x86_64-linux-gnu-ubuntu-14.04
   pre:
     # LLVM's official APT repo:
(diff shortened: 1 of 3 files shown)


kinke:
> Makes sense?

Very much so IMO.

JohanEngelen:
LLVM 4.0 is green :)

PR #1740, opened 2016-09-04
dmd-testsuite: Resolve RESULTS_DIR issue for bash scripts on Windows

kinke:


kinke:
I built it myself using the latest sources. They surprisingly ship with a Visual Studio (2003 ;)) solution. I built the Release x86 configuration (which links statically against the MS runtimes). I don't think this detail is worth mentioning somewhere, after all, it's only used by AppVeyor, and we won't need to update `make` that often (= not once in the next decade I'd bet).

PR #1753, opened 2016-09-11
Add ir2obj cache pruning support.

### LINKED ISSUE #1747: Implement ir2obj cache expiration
The cache grew to 60GB for some Weka devs, not good.

Idea: add module name to cache filenames, and start deleting the oldest file for a module when the count reaches 10 or some other value.

JohanEngelen:
This implements ir2obj cache pruning in the compiler itself, and also through a stand-alone tool (they share the cache pruning D implementation).

Cache pruning scheme:
1. Check the prune interval (`-ir2obj-cache-prune-interval` and `ldc-prune-cache --interval`) to see if pruning is already needed again. A timestamp file is written to the cache for this.
2. First remove all "expired" files, i.e. files with an age older than a set limit (`-ir2obj-cache-prune-expiration` and `ldc-prune-cache --expiry`).
3. Then, if the cache size is still too large, remove files (oldest first) until the size is below a set limit. The limit can be specified absolutely (`--ir2obj-cache-prune-maxbytes` and `ldc-prune-cache --max-bytes`) and/or relatively (`--ir2obj-cache-prune-maxpercentage` and `ldc-prune-cache --max-percentage-of-avail`). 

Pruning is done after codegen, but before linking. So the prune parameters must be such that the object files needed for linking will survive (e.g. the cache size limit should be large enough).

Pruning is off by default, but specifying any pruning parameters on the commandline implies enabling it (`--ir2obj-cache-prune`).
The reasoning for having it off per default is that in projects using parallel compilation threads, you'd want the pruning to happen at the end of a build cycle (using the standalone tool). Otherwise, the first finishing compilation thread may start pruning stuff that you might cache-hit in another compile thread. Also, multi-threaded pruning is untested at best.

The pruning code uses the file's "last access time". But because this is not automatically updated upon symlinking or usage by the linker on some platforms (Windows 7, from what I read online), I added an extra "touch" upon a cache hit such that the last access time is refreshed at least during compilation.

Resolves #1747


driver/ir2obj_cache_pruning.h
```diff
@@ -0,0 +1,17 @@
...
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LDC_DRIVER_IR2OBJ_CACHE_PRUNING_H
+#define LDC_DRIVER_IR2OBJ_CACHE_PRUNING_H
+
+void DpruneCache(const char *cacheDirectoryPtr, size_t cacheDirectoryLen,
```
klickverbot: Sorry for nit-picking, but `DpruneCache` is not a great name ("D"?), especially since it is hardcoded to `"ircache_"`.

JohanEngelen: wip !
(it was to set it apart from the previous C++ implementation)

JohanEngelen:
All green :)
Don't think I can make it better than this.

driver/ir2obj_cache_pruning.h
```diff
@@ -0,0 +1,23 @@
...
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LDC_DRIVER_IR2OBJ_CACHE_PRUNING_H
+#define LDC_DRIVER_IR2OBJ_CACHE_PRUNING_H
+
+#if __LP64__
+using d_ulong = unsigned long;
```
klickverbot: `#include <cstdint>` and use `uint64_t`.

driver/main.cpp
```diff
@@ -1171,6 +1172,11 @@ void codegenModules(Modules &modules) {
     }
   }
 
+  // Prune the ir2obj cache if needed.
+  if (!opts::ir2objCacheDir.empty()) {
```
klickverbot: Minor comment: I found this a bit misleading to read – especially in conjunction with the comment –, since `pruneCache` doesn't actually do any pruning unless it is enabled.

tests/linking/ir2obj_cache_pruning2.d
```diff
@@ -0,0 +1,38 @@
...
+    version(NEW_OBJ_FILE)
+    {
+        auto a = __TIME__;
+    }
+
+    version(SLEEP)
+    {
+        // Sleep for 2 seconds, so we are sure that the cache object file timestamps are "aging".
+        import core.thread;
+        Thread.sleep( dur!("seconds")( 2 ) );
```
klickverbot: D style comment: `dur!"seconds"(2)` would be more in line with upstream/Phobos.

tests/linking/ir2obj_cache_pruning2.d
```diff
@@ -0,0 +1,38 @@
...
+// RUN: && %ldc %s -ir2obj-cache=%T/prunecache2 -ir2obj-cache-prune-interval=1 -ir2obj-cache-prune-maxbytes=200000 -d-version=NEW_OBJ_FILE \
+// RUN: && %ldc %s -ir2obj-cache=%T/prunecache2 -ir2obj-cache-prune -ir2obj-cache-prune-interval=0 -vv | FileCheck --check-prefix=NO_HIT %s
+
+// MUST_HIT: Cache object found!
+// NO_HIT-NOT: Cache object found!
+
+void main()
+{
+    // Add non-zero static data to guarantee a binary size larger than 200000/2.
+    static byte[120000] dummy = 1;
```
klickverbot: D style tip: Underscores (`120_000`) are great.

driver/ir2obj_cache.cpp
```diff
@@ -41,8 +42,57 @@
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
 
+// Include close() declaration.
```
klickverbot: Pretty sure LLVM guarantees availability of that.

driver/ir2obj_cache_pruning.h
JohanEngelen: I had tried it, and it doesn't work :(

driver/ir2obj_cache.cpp
JohanEngelen: Nope.

JohanEngelen:
> I suppose you'll move ldc-profdata over in a separate PR?

Yep!

driver/ir2obj_cache.cpp
JohanEngelen: Platform support was painful for this PR :(   On my Mac, things all worked fined, but to get it to compile and pass tests on all CI systems took a lot of trial and error.

driver/main.cpp
JohanEngelen: I've moved the check to ir2obj::pruneCache(), and removed the comment.

tests/linking/ir2obj_cache_pruning2.d
JohanEngelen: ok, fixed

tests/linking/ir2obj_cache_pruning2.d
JohanEngelen: great, fixed

driver/ir2obj_cache_pruning.h
klickverbot: Why?

JohanEngelen: Mac's compiler produces:
`pruneCache(char const*, unsigned long, unsigned int, unsigned int, unsigned long long, unsigned int)`
DMD and LDC produce: 
`pruneCache(char const*, unsigned long, unsigned int, unsigned int, unsigned long, unsigned int)`

Even the table here https://dlang.org/spec/interfaceToC.html  is ambiguous for `ulong`...

JohanEngelen:
(CircleCI was green. Currently broken because LLVM 4.0 API change)

driver/ir2obj_cache_pruning.h
klickverbot: Sorry, I didn't notice this was actually an extern(C++) interface.

PR #1759, opened 2016-09-14
LLVMContext::setDiscardValueNames(true) when not compiling to textual IR

### LINKED ISSUE #1749: llvm::Context::setDiscardValueNames() when not compiling to textual IR?
Setting the context to discard value names when not generating textual IR might give us a small performance gain, or might not. Only one way to find out… ;)

JohanEngelen:
Resolves #1749

### DIFF
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -84,6 +84,13 @@ CodeGenerator::CodeGenerator(llvm::LLVMContext &context, bool singleObj)
                  "configured properly");
     fatal();
   }
+
+#if LDC_LLVM_VER >= 309
+  // Set the context to discard value names when not generating textual IR.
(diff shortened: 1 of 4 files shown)


JohanEngelen:
Do we want a commandline flag to disable the discarding for bitcode debugging?

JohanEngelen:
No, haven't tested that yet.
I think the impact will not be that large without caching though. I think LLVM internally (like DMD internally), the actual string data of local variables is rarely looked at when producing binary output. I _very_ quickly glanced at our own source code where we set local variable names: if create strings with `Twine`s, then that speeds up compilation because LLVM will not evaluate those.
With caching, I think the speedup boost may be larger.

In any case, it's a relatively simple change and I don't see how it can hurt performance. Great catch finding the option!

PR #1761, opened 2016-09-14
CodeView: fix display of dynamic array arguments

rainers:
see https://github.com/ldc-developers/ldc/issues/1716

Is this relevant for gdb, too?

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -1022,7 +1022,17 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
                                                Flags                // flags
                                                );
 #else
-  if (!fromNested && vd->isParameter()) {
+  bool isParameter = vd->isParameter();
+  if (global.params.targetTriple->isWindowsMSVCEnvironment() &&
+      global.params.targetTriple->getArch() == llvm::Triple::x86_64) {
+    // arrays are not passed through ExplicitByvalRewrite, because it is not
+    //  considered an aggregate (see TargetABI::isAggregate)
+    // instead a local copy of the parameter is created in the local frame, so
+    // pretend it is a local
+    if (isParameter && type && type->ty == Tarray)
+      isParameter = false;
+  }
+  if (!fromNested && isParameter) {
     FuncDeclaration *fd = vd->parent->isFuncDeclaration();
     assert(fd);
     size_t argNo = 0;
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -947,7 +947,7 @@ void ldc::DIBuilder::EmitValue(llvm::Value *val, VarDeclaration *vd) {
 
 void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
                                        Type *type, bool isThisPtr,
-                                       bool fromNested,
+                                       bool rewrittenToLocal,
 #if LDC_LLVM_VER >= 306
                                        llvm::ArrayRef<int64_t> addr
 #else
@@ -979,7 +979,7 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
 
 #if LDC_LLVM_VER < 308
   unsigned tag;
-  if (!fromNested && vd->isParameter()) {
+  if (!rewrittenToLocal && vd->isParameter()) {
     tag = llvm::dwarf::DW_TAG_arg_variable;
   } else {
     tag = llvm::dwarf::DW_TAG_auto_variable;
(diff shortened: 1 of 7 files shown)


gen/dibuilder.cpp
```diff
@@ -1022,7 +1022,17 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
...
+  if (global.params.targetTriple->isWindowsMSVCEnvironment() &&
+      global.params.targetTriple->getArch() == llvm::Triple::x86_64) {
+    // arrays are not passed through ExplicitByvalRewrite, because it is not
+    //  considered an aggregate (see TargetABI::isAggregate)
+    // instead a local copy of the parameter is created in the local frame, so
+    // pretend it is a local
+    if (isParameter && type && type->ty == Tarray)
+      isParameter = false;
+  }
+  if (!fromNested && isParameter) {
```
kinke: All parameters, incl. implicit ones afaik (`this`, `_argptr`, `_arguments` etc.), are lvalues, so we should always have a in-memory local (`IrLocal::value` IIRC) and hence never have to refer to the LL parameter directly, although that may provide better debuggability for optimized builds, when the local is optimized away.

kinke: We could check whether an ABIRewrite is set for that parameter, and never use the LL parameter in that case (always taking the rewritten local instead). That wouldn't solve the slice issue for Win64 at hand, but it should prevent further debuginfo bugs.

klickverbot: We definitely need to do at least something along these lines. We can always add another primitive to the ABI interface if it turns out we need one. But special casing targets all over the code base like this isn't going to be viable in the long run.

rainers: Not sure I understand correctly, but the slices were the only type I've seen that failed to display correctly, even though almost all types get rewritten (struct values are an exception for example). I suspect that this happens because the LLVM codegen only changes the type of slices when applying the ABI.

I've added a check for the actual rewrite and now pass that information to EmitLocalVariable. This loses some information for other parameters, though (see change to test case nested.d).

gen/functions.cpp
```diff
@@ -721,8 +722,10 @@ void defineParameters(IrFuncTy &irFty, VarDeclarations &parameters) {
         assert(irparam->value->getType() == DtoPtrToType(paramType));
       } else {
         // Let the ABI transform the parameter back to an lvalue.
-        irparam->value =
+        auto Lvalue =
             irFty.getParamLVal(paramType, llArgIdx, irparam->value);
+        //rewrittenToLocal = Lvalue != irparam->value;
```
kinke: `irparam->arg->rewrite` is set if the parameter is rewritten, where we'll always want to use the local.

E.g., an argument of type `struct S { int a, b; }` on Win64 is rewritten to a LL `i64` integer, in order to pass it in a single register. We don't want to associate the debuginfo for that struct parameter with that LL integer parameter, but with the transformed-back local (`i64 -> S`).

kinke: I'm not sure if we gain something by using the LL parameter directly in case there's no rewrite.

kinke: Well I think that's even wrong, as modifications wouldn't show up correctly. `void foo(int a) { ++a; }` - if the debuginfo is associated with the LL param, it won't show up as incremented at the end of the function, would it?

gen/dibuilder.cpp
kinke: The slice issue is probably that LLVM expects a LL pointer to be associated with a debuginfo-struct type, not a 2x64-bit LL aggregate immediate (not-rewritten slice parameters on Win64).

gen/functions.cpp
rainers: Argh, that line was not meant to be commented out. That was a temporary change to compare old/new IRs. Amended the change to put the line back in.

> if the debuginfo is associated with the LL param, it won't show up as incremented at the end of the function, would it?

Actually it does because making it a parameter/local does not change the actual referred variable in the IR, but causes changes to the type by implicit ABI changes (LLVM assumes passing larger values by ref).

rainers:
> irparam->arg->rewrite is set if the parameter is rewritten, where we'll always want to use the local.

Do you think this is the better check? I'm still a bit on shaky ground when it comes to rewrites...

kinke:
Oh I now seem to slowly understand. So we always use `irparam->value` as parameter storage, i.e., the callee's local. And `createParameterVariable()` doesn't just seem to annotate the debug variable with a parameter index, but is also somehow affected by the LL parameter type. In that case, your check whether the local's lvalue is exactly the LL parameter value should be on the safe side - I guess only `ExplicitByvalRewrite` and the LLVM `byval` attribute behave that way.

PR #1762, opened 2016-09-15
Add a matching call to llvm.va_end when llvm.va_start was called.

### LINKED ISSUE #1744: `llvm.va_start` not matched with `llvm.va_end`
For this code:
```d
int return_two(...) {
  return 2;
}

int two_four() {
  return return_two() * 4;
}
```

with `-O3` it is compiled to:
```llvm
define i32 @_D3ipo10return_twoFYi({ i64, %object.TypeInfo** } %._arguments, ...) local_unnamed_addr #0 {
  %__va_list_mem = alloca { i32, i32, i8*, i8* }, align 8 ; [#uses = 1, size/byte = 24]
  %1 = bitcast { i32, i32, i8*, i8* }* %__va_list_mem to i8* ; [#uses = 1]
  call void @llvm.va_start(i8* %1)
  ret i32 2
}

define i32 @_D3ipo8two_fourFZi() local_unnamed_addr #0 {
  %1 = tail call i32 ({ i64, %object.TypeInfo** }, ...) @_D3ipo10return_twoFYi({ i64, %object.TypeInfo** } undef) #3 ; [#uses = 0]
  ret i32 8
}
```

Note that `two_four` still calls `return_two`. The problem is that we don't call `llvm.va_end` in `return_two`. When I add `  call void @llvm.va_end(i8* %1)` right before `ret`, the call to `return_two` is optimized out correctly in `two_four`.

Now, I don't know enough of `llvm.va_*` to know if this is a bug or not, but... at least it looks like a missed optimization opportunity.

JohanEngelen:
Resolves #1744

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -1011,6 +1011,17 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
     // copy _arguments to a memory location
     irFunc->_arguments =
         DtoAllocaDump(irFunc->_arguments, 0, "_arguments_mem");
+
+    // Push cleanup block that calls va_end to match the va_start call.
+    {
+      auto *vaendBB =
+          llvm::BasicBlock::Create(gIR->context(), "vaend", gIR->topfunc());
+      IRScope saveScope = gIR->scope();
+      gIR->scope() = IRScope(vaendBB);
+      gIR->ir->CreateCall(GET_INTRINSIC_DECL(vaend), llAp);
+      funcGen.scopes.pushCleanup(vaendBB, gIR->scopebb());
+      gIR->scope() = saveScope;
+    }
   }
 
   funcGen.pgo.emitCounterIncrement(fd->fbody);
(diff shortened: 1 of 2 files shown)


JohanEngelen:
ping

tests/codegen/vastart_vaend_gh1744.d
```diff
@@ -0,0 +1,54 @@
...
+        return;
+    }
+    if (a > 0)
+    {
+        throw new Exception("");
+        return;
+    }
+
+    // There are two control paths (normal return, exception resume) that
+    // should call va_end.
```
kinke: Isn't it 3? `a < 0`, `a > 0` (unwinding) and `a == 0`, the implicit return.

kinke: Oh, optimized to a single `a > 0`. ;)

gen/functions.cpp
```diff
@@ -1016,6 +1027,18 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   // output function body
   Statement_toIR(fd->fbody, gIR);
 
+  // D varargs: emit the cleanup block that calls va_end.
+  if (f->linkage == LINKd && f->varargs == 1) {
+    if (!funcGen.retBlock)
+      funcGen.retBlock = gIR->insertBB("return");
+
+    if (!gIR->scopereturned()) {
```
kinke: Please move this up before the retBlock-if, so that we don't create an empty BB if there's no need to.

tests/codegen/vastart_vaend_gh1744.d
JohanEngelen: No need for the (a < 0) part, I'll remove it.

kinke:
CircleCI: LLVM 4.0 apparently decorates the `llvm.va_start` arg with `nonnull`.

PR #1763, opened 2016-09-15
Move ldc-profdata source into the tools directory.

JohanEngelen:
Let's use the `utils` folder for stuff we need during building/testing, and the `tools` folder for tools that should be shipped with LDC.


JohanEngelen:
Good to merge?

PR #1764, opened 2016-09-15
Fix unused var warning.

JohanEngelen:
I chose this over `(void)...`.


JohanEngelen:
(annoyed by the warning, so merging)

PR #1768, opened 2016-09-17
CodeView: fix vector types void16 and void32

rainers:
This fixes vectors of void showing as delegates in debuggers (or asserting early).


JohanEngelen:
Nice.
To me it looks like you can replace "CodeView:" with "Debug info:". It's just that the testing is done only for codeview, but I imagine this also fixes the types for other debuggers.

LGTM.

rainers:
> To me it looks like you can replace "CodeView:" with "Debug info:"

True. I changed the commit message.

rainers:
Fixed the test for x86: 
* the little helper function was merged with other empty functions by the linker, so the debugger stopped too early
* with this fixed, it didn't have a standard stack frame, so walking up the stack to the main function didn't work. I guess no FPO records generated by LLVM.

PR #1769, opened 2016-09-17
Debuginfo: add inheritance information to classes

rainers:
uses inheritance information for base classes instead of listing base class members in derived classes

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -53,7 +53,16 @@ ldc::DIType getNullDIType() {
   return llvm::DIType();
 #endif
 }
+
+#if LDC_LLVM_VER >= 307
+llvm::DINodeArray getEmptyDINodeArray() {
+  return nullptr;
 }
+#else
+llvm::DIArray getEmptyDINodeArray() {
+  return llvm::DIArray();
+}
+#endif
 
 llvm::StringRef uniqueIdent(Type* t) {
 #if LDC_LLVM_VER >= 309
@@ -62,6 +71,8 @@ llvm::StringRef uniqueIdent(Type* t) {
 #endif
   return llvm::StringRef();
 }
+
+} // namespace
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -363,11 +374,8 @@ ldc::DIType ldc::DIBuilder::CreateMemberType(unsigned linnum, Type *type,
                                    );
 }
 
-void ldc::DIBuilder::AddBaseFields(ClassDeclaration *sd, ldc::DIFile file,
-                                   llvm::SmallVector<LLMetadata *, 16> &elems) {
-  if (sd->baseClass)
-    AddBaseFields(sd->baseClass, file, elems);
-
+void ldc::DIBuilder::AddFields(AggregateDeclaration *sd, ldc::DIFile file,
+                               llvm::SmallVector<LLMetadata *, 16> &elems) {
   size_t narr = sd->fields.dim;
   elems.reserve(narr);
   for (auto vd : sd->fields) {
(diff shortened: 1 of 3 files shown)


kinke:
I did a quick check on Linux with gdb.

Master:
```
(gdb) info locals
dc = 0x7ffff7ed7000
(gdb) print dc
$1 = (DerivedClass *) 0x7ffff7ed7000
(gdb) print *dc
$2 = {baseMember = 0, derivedMember = 0}
(gdb) print dc.baseMember
$3 = 3
(gdb) print dc.derivedMember
$4 = 7
```

This PR:
```
(gdb) info locals
dc = 0x7ffff7ed7000
(gdb) print dc
$1 = (DerivedClass *) 0x7ffff7ed7000
(gdb) print *dc
$2 = {<BaseClass> = {<Object> = {<No data fields>}, baseMember = 0}, derivedMember = 0}
(gdb) print dc.baseMember
$3 = 3
(gdb) print dc.derivedMember
$4 = 7
```

So the correct member values don't show up when displaying the whole class, but inheritance seems to be working nicely.

JohanEngelen:
@rainers Could you name the CodeView files with "_cv" or "_codeview" at the end (instead of at the start)? That's nicer if we add similar files for other debuggers.

rainers:
> So the correct member values don't show up when displaying the whole class, but inheritance seems to be working nicely.

Ouch, incorrect class members! But that's no change from the master print-out.

rainers:
> @rainers Could you name the CodeView files with "_cv" or "_codeview" at the end (instead of at the start)? That's nicer if we add similar files for other debuggers.

Can do. What whould be the others? _dwarf? We probably need to distinguish gdb and lldb, so _cdb would be appropriate for the codeview files. Maybe we can later also test other windows debuggers (e.g. mago), so using the debugger in the name instead of the debug info format should be best.

JohanEngelen:
I also think adding the debugger name as prefix is better.
Tbh, I don't feel too much for duplication a lot of tests for each debugger. I'd rather check the LLVM IR (note the LLVM comment below: "it's worth remembering that debuggers are not the only consumers of debug info").  But for the cases where the IR output has to be different for different debuggers, it's nice to add these specific tests. 

Related: I think we should add "Debugger Tuning" flags (http://clang.llvm.org/docs/UsersManual.html#controlling-debugger-tuning); I may submit a PR for this. It would replace the `if(target->Windows()){}` for debuginfo, with `if(getDebuggerTuning() != CDB)`, and set the default debugger tuning depending on the target platform. LLVM reads:
```cpp
  /// Identify a debugger for "tuning" the debug info.
  ///
  /// The "debugger tuning" concept allows us to present a more intuitive
  /// interface that unpacks into different sets of defaults for the various
  /// individual feature-flag settings, that suit the preferences of the
  /// various debuggers.  However, it's worth remembering that debuggers are
  /// not the only consumers of debug info, and some variations in DWARF might
  /// better be treated as target/platform issues. Fundamentally,
  /// o if the feature is useful (or not) to a particular debugger, regardless
  ///   of the target, that's a tuning decision;
  /// o if the feature is useful (or not) on a particular platform, regardless
  ///   of the debugger, that's a target decision.
  /// It's not impossible to see both factors in some specific case.
  ///
  /// The "tuning" should be used to set defaults for individual feature flags
  /// in DwarfDebug; if a given feature has a more specific command-line option,
  /// that option should take precedence over the tuning.
  enum class DebuggerKind {
    Default,  // No specific tuning requested.
    GDB,      // Tune debug info for gdb.
    LLDB,     // Tune debug info for lldb.
    SCE       // Tune debug info for SCE targets (e.g. PS4).
  };
```

So perhaps we should try to add "CDB" in there asap :)

JohanEngelen:
(btw, before I forget: when merging this PR, please first squash the commits such that each commit still builds (annoying when individual commits break compilation during bisecting))

rainers:
> So perhaps we should try to add "CDB" in there asap :)

I don't think cdb is a common target because nobody uses it (it could be remoted by some IDE though, I remember DLangIDE used it in the past), windbg would be the more common debugger that probably has similar capabilities. The overwhelming more common debugger is likely to be Visual Studio.

I don't oppose adding debugger tuning flags, but so far we don't have special tuning per debugger on Windows, it's all just about the CodeView debug format. For nice D support I'd expect to support mago features that you would disable with -gc.

rainers:
> (btw, before I forget: when merging this PR, please first squash the commits such that each commit still builds (annoying when individual commits break compilation during bisecting))

Done.

JohanEngelen:
> I don't oppose adding debugger tuning flags, but so far we don't have special tuning per debugger on Windows, it's all just about the CodeView debug format.

I meant this in `dibuilder.cpp`:
```d
    if (global.params.targetTriple->isWindowsMSVCEnvironment()) {
      // VS debugger does not support DW_ATE_UTF for char
```
It'd be very nice to replace the `targetTriple->isWindowsMSVCEnvironment()` with a function that checks for the VS debugger (under the hood it could be doing the same). From my testing it appears that LLDB also does not support DW_ATE_UTF.

PR #1771, opened 2016-09-17
Fix inferred output filename

kinke:
Base it on the first source filename, not on the last.


JohanEngelen:
Can you add a test for this?

Something like
`linking/test_inferred_filename.d`
```d
// If it works on Windows, it will work on other platforms too, and it makes it easier to do the regexp match because of the ".exe" extension
// REQUIRES: Windows
// RUN: %ldc %s %S/inputs/foo.d -vv | FileCheck %s

// CHECK: Linking with:
// CHECK-NEXT: 'test_inferred_filename.exe'

module modulename;
void main() { }
```

kinke:
I was going to add something similar, but then wasn't sure how to make it work on Windows and on other OS' because of the .exe extension, and I disliked that the .exe would be created in the working directory without removing it (and I guess the working directory could be a src directory, leaving the src tree modified).

Anyway, the 2 new LLVM-3.7+-only errors (`codegen/attr_weak.d` - `@weak` function not overriden - and a `runnable/cppa.d` crash on Linux x64) suggest other bugs wrt. linkage. The object files order for the linker isn't reversed anymore (i.e., corresponds to the source files order), and this appears to break things.

JohanEngelen:
> (and I guess the working directory could be a src directory, leaving the src tree modified)

The working dir is explicitly set by Lit. We already have a few tests that write to that, no problem. (it's in a subdir from where the tests are started)

kinke:
That `@weak` apparently doesn't really work with master either, at least on Windows:
```
ldc2.exe -c ..\ldc\tests\codegen\attr_weak.d ..\ldc\tests\codegen\inputs\attr_weak_input.d
link.exe attr_weak.obj attr_weak_input.obj lib\druntime-ldc.lib lib\phobos2-ldc.lib
```
=> `attr_weak.exe` asserts in `attr_weak.d:20`, regardless of object file order for linker.

kinke:
It gets pretty bad when trying to compile both modules at once with `-singleobj` (using current master again):

`ldc2.exe -c -output-ll ..\ldc\tests\codegen\attr_weak.d ..\ldc\tests\codegen\inputs\attr_weak_input.d -ofsingle_normal.ll`:

```LL
define i32 @return_two() #0 comdat {
  ret i32 1
                                                  ; No predecessors!
  ret i32 2
}

define weak i32 @return_seven() #0 comdat {
  ret i32 7
                                                  ; No predecessors!
  ret i32 1
}
```

Reversed source file order:
```LL
define weak i32 @return_two() #0 comdat {
  ret i32 2
                                                  ; No predecessors!
  ret i32 1
}

define i32 @return_seven() #0 comdat {
  ret i32 1
                                                  ; No predecessors!
  ret i32 7
}
```

kinke:
Okay the only way which doesn't assert is the one which is tested: compiling both modules with `-O3` AND specifying the ..._input.obj file before the other one when linking.

kinke:
I guess this has to do with comdats (hence no issue for LLVM < 3.7), and LDC using the `any` selection, which seems not to care about `weak` when picking a comdat. I recall reading somewhere that all comdats with a given name can only feature one selection kind, so we can't use `any` for weak symbols and stronger ones for non-weak.
I guess for our testcase the order just happens to be right when optimizing and linking in a specific order - at least on Windows.

kinke:
@rainers: I recall some COFF-weak-linking tricks on your part in (the C parts of) druntime. Could this be an LLVM issue with missing support for weak symbols in COFF objects?

JohanEngelen:
@kinke I remember when we looked at comdats, that I commented that perhaps `@weak` should be removed on Windows as there is no such thing on that platform. Perhaps we really just have to remove it.

kinke:
Hmm, it may be a general COFF limitation, but on the other hand, the `codegen/attr_weak.d` test also fails on Linux x64 with LLVM 3.7+, so it seems to be sensitive to the object files order for the linker there too.

JohanEngelen:
We want to be able to use `@weak` for overriding symbols in a library, and to tell the compiler that it should not inline nor reason about the library function while compiling the library.
So perhaps the tests are wrong, in that they assume `@weak` has too much power. I think for the library linking option, it works well.    The name "weak" may be bad for this.

rainers:
> @rainers: I recall some MSVC-weak-linking tricks on your part in (the C parts of) druntime. Could this be an LLVM issue with missing support for weak symbols in COFF objects?

In druntime we use the /ALTERNATENAME directive embedded into the object file. This seems the closest way to get something similar to weak symbols with COFF. The weak implementaion would be renamed to something like `return_two$weak` and the directive "/ALTERNATENAME:return_two=return_two$weak" would make the linker choose this if not implemented elsewhere.
The trouble with this approach is that it doesn't work too well within libraries, because the directive needs to be pulled in as part of another symbol definition referenced elsewhere.

kinke:
I think I figured why `runnable/cppa.d` crashed. It's good that this PR uncovered this nasty issue (see commit message).

kinke:
> I wonder whether we really need to hack in externalObjectFiles – it seems like it would be unnecessary duplication given that we rely on the index order anyway.

It did have a purpose, but I decided putting the external objects to the end of the list was unintuitive anyway, so I've amended the first commit (incl. commit message), and hacked the `weak` test.

tests/codegen/attr_weak.d
```diff
@@ -1,21 +1,20 @@
-// Test linking+running a program with @weak functions
+// Test linking+running a program with @weak function
 
-// RUN: %ldc -O3 %S/inputs/attr_weak_input.d -c -of=%t%obj
-// RUN: %ldc -O3 %t%obj -run %s
+// RUN: %ldc -O3 %S/inputs/attr_weak_input.d -c -of=%T/attr_weak_input%obj
+// RUN: %ldc -O3 %T/attr_weak_input%obj %s -of=%t%exe
+// RUN: %t%exe
```
kinke: `-run bla.d` is special in the sense that `bla.d` will be injected before all other files in the command line => `bla.d %T/attr_weak_input%obj`, so the new object file would override definitions in the existing object file, and this test wouldn't work.

JohanEngelen:
@kinke I really wish you wouldn't cram so many changes into one PR. It makes the history harder to trace, harder to revert things, testing, bisecting, etc. Now for example, without submitting a separate and quick PR, our appveyor testing is broken until we merge this monster PR. If you aren't already, I recommend using a tool like SourceTree to make selective committing/branching/etc. easier. 
Thanks so much.

kinke:
The AppVeyor fix could be simply cherry-picked onto master. I was hoping this would get merged pretty soon though, that's why I haven't done so already. And this is really not a monster PR. ;)

driver/linker.cpp
```diff
@@ -672,27 +682,26 @@ int linkObjToBinary() {
 int createStaticLibrary() {
   Logger::println("*** Creating static library ***");
 
-  const bool isTargetWindows =
-      global.params.targetTriple->isWindowsMSVCEnvironment();
+  const bool isCOFF = global.params.mscoff;
 
   // find archiver
-  std::string tool(isTargetWindows ? "lib.exe" : getArchiver());
+  std::string tool(isCOFF ? "lib.exe" : getArchiver());
```
JohanEngelen: All these renamings don't make sense. "COFF" does not mean "Windows".

kinke: I know that, but the previous names were inexact. MinGW builds for Windows use GCC.

JohanEngelen: So call it `isTargetMSVC`.
Other systems use COFF too.

klickverbot: If I remember correctly, MinGW actually uses COFF too.

kinke: Renamed, I've never heard of COFF in non-MSVC contexts before (and I'm pretty sure the MinGW objects aren't usable with the MS linker, at least ~2 years ago).

klickverbot: MinGW/MSVC incompatibilities notwithstanding, it actually is COFF:

```
$ gcc --version
gcc.exe (Rev2, Built by MSYS2 project) 6.2.0
$ gcc -c test.c
$ llvm-objdump -d test.o
test.o:	file format COFF-x86-64 
[…]
```

kinke:
I already began thinking about using `global.params.mscoff` as shortcut for that awfully long `global.params.targetTriple->isWindowsMSVCEnvironment()` (in combination with the 80 cols clang-format setting). Especially since it's exactly equivalent at the moment, and I don't really expect MS to push their object file format on non-Windows platforms. Let's please not dive into such bikeshedding, time is spent more wisely for other stuff.

As to splitting this into separate PRs - I've already cherry-picked the AppVeyor fix. The other separated issue was the thunks, which accumulate to about 10 modified lines (and verifying that it works is way simpler with the first commit, breaking `runnable/cppa.d`; no extra test case needed). And I didn't expect a debate about that mini refactoring in the last commit, included as part of supporting some more command-line files, a byproduct of having to go over that `linker.cpp` file.

kinke:
I'm pretty sure it's not **MS** COFF. Amend is on its way.

kinke:
Yeah, except that the front-end uses it to identify an MSVC target (C++ mangling and interface vtable special cases). Don't ask me to open an upstream PR renaming that variable and fixing the comment though. ;)

kinke:
Redundant AppVeyor fix removed and 2 new fixes incorporated (`-oq` regression introduced by #1726).

kinke:
The added bash tests are [here](https://github.com/ldc-developers/dmd-testsuite/commit/b02f47a8fd7f4c74553745e3a5d88776ec8e0438). I'm no bash expert at all, so if it can be shortened, be sure to let me know.

kinke:
Apparently there's a permission issue for the bash tests when removing the object files on non-Windows. I guess it has to do with how the output directory is created.

kinke:
*lol* The permission issue is that the .sh script doesn't have the executable permission. That's what you get when you try to do that on Windows. :D

kinke:
I've refactored the bash script a bit and updated the link accordingly.

kinke:
I'm going to merge this this weekend if there's no further feedback.

kinke:
Thanks!

PR #1772, opened 2016-09-19
fix build with LLVM 4.0

rainers:
At least on Windows, library debuginfomsf is needed by debuginfocodeview.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ endif()
 #
 
 find_package(LLVM 3.5 REQUIRED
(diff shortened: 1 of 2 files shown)


JohanEngelen:
@rainers Do you already notice that LLVM 4.0 has better debuginfo on Windows, or not yet?

JohanEngelen:
@rainers Perhaps for extra motivation: Ethan from Remedy told me that debug info is critical for their internal D usage.

rainers:
> Do you already notice that LLVM 4.0 has better debuginfo on Windows, or not yet?

I am hoping for improvements but didn't have time to investigate it more closely yet. At least, there don't seem to be regressions...

> Ethan from Remedy told me that debug info is critical for their internal D usage.

I know from Manu, too. Unfortunately, the info emitted for local variables is not compatible with mago yet.

PR #1773, opened 2016-09-19
ir2obj cache: input all optimization settings into the hash.

JohanEngelen:
This fixes ir2obj caching bugs (incorrect cache hits) when using some of the more exotic command line flags.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -17,6 +17,9 @@
 
 namespace opts {
 
+// This vector is filled by parseCommandLine in main.cpp.
+llvm::SmallVector<const char *, 32> allArguments;
+
 /* Option parser that defaults to zero when no explicit number is given.
  * i.e.:  -cov    --> value = 0
  *        -cov=9  --> value = 9
@@ -198,7 +201,7 @@ static cl::opt<bool, true> unittest("unittest",
 
 cl::opt<std::string>
     ir2objCacheDir("ir2obj-cache", cl::desc("Use <cache dir> to cache object files for whole IR modules (experimental)"),
-            cl::value_desc("cache dir"), cl::Prefix);
+            cl::value_desc("cache dir"));
 
 static StringsAdapter strImpPathStore("J", global.params.fileImppath);
 static cl::list<std::string, StringsAdapter>
--- a/driver/ir2obj_cache.cpp
+++ b/driver/ir2obj_cache.cpp
@@ -127,6 +127,104 @@ void storeCacheFileName(llvm::StringRef cacheObjectHash,
   llvm::sys::path::append(filePath, llvm::Twine("ircache_") + cacheObjectHash +
                                         "." + global.obj_ext);
 }
+
+// Output to `hash_os` all commandline flags, and try to skip the ones that have
+// no influence on the object code output. The cmdline flags need to be added
+// to the ir2obj cache hash to uniquely identify the object file output.
+// Because the compiler version is part of the hash, differences in the
+// default settings between compiler versions are already taken care of.
+// (Note: config and response files may also add compiler flags.)
+void outputIR2ObjRelevantCmdlineArgs(llvm::raw_ostream &hash_os)
+{
+  // Use a "whitelist" of cmdline args that do not need to be added to the hash,
+  // and add all others. There is no harm (other than missed cache
+  // opportunities) in adding commandline arguments that also change the hashed
+  // IR, which simplifies the code here.
+  // The code does not deal well with options specified without equals sign, and
+  // will add those to the hash, resulting in missed cache opportunities.
+  for (const char *arg : opts::allArguments) {
+    if (!arg || !arg[0])
+      continue;
+
+    // Out of pre-caution, all arguments that are not prefixed with '-' are
+    // added to the hash. Such an argument could be a source file "foo.d", but
+    // also a value for the previous argument when the equals sign is omitted,
+    // for example: "-code-model default" becomes "-code-model" "default".
+    // It results in missed cache opportunities. :(
+    if (arg[0] == '-') {
+      if (arg[1] == 'O') {
+        // We deal with -O later ("-O" and "-O3" should hash equally, "" and
+        // "-O0" too)
+        continue;
+      }
+      if (arg[1] == 'c' && !arg[2])
+        continue;
+      // All options starting with these characters can be ignored (LLVM does
+      // not have options starting with capitals)
+      if (arg[1] == 'D' || arg[1] == 'H' || arg[1] == 'I' || arg[1] == 'J' ||
+          arg[1] == 'L' || arg[1] == 'X')
+        continue;
+      if (arg[1] == 'd' || arg[1] == 'v' || arg[1] == 'w') {
+        // LLVM options are long, so short options starting with 'v' or 'w' can
+        // be ignored.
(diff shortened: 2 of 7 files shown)


JohanEngelen:
Yes we would still get false positives on LLVM flags. I don't know how to handle those. Or, perhaps we should just hash all commandlines, after removing the ones that match a "whitelist" of flags.

JohanEngelen:
> Are you hashing pre-optimisation but caching afterwards?

Yes.
The optimization flags are not reflected in the IR _pre_ optimization. So we have to manually add them to the hash.

JohanEngelen:
> Won't we still get false positives on LLVM-defined flags?

The LLVM options are a major pain. I cannot find any way to list all the _values_ of all our commandline options (from `llvm::cl`). So I would have to code my own argument parser... just for being able to get equal hashes for these things:
`-O` == `-O3` (option aliases)
`<nothing specified>` == `-O0` (default settings)
`-of=something` == `-of=somethingelse` (irrelevant options)
`-mtriple=abc` == `-mtriple abc`  (space or equals shouldn't matter)
etc.

I've thought about it before, and now it resurfaces: perhaps we should only allow LLVM options when prefixed with `-mllvm` like Clang. That would eliminate the need for `hideLLVMOptions()` in main.cpp and the small hack needed for `-color`. (https://stackoverflow.com/questions/34147464/how-to-avoid-llvms-support-commandline-leaking-library-arguments shows how to do it)

JohanEngelen:
@klickverbot I wrote a simple cmdline "parser" such that common flags don't result in cache miss hits. It should be much more robust against exotic compile flag IR changes.

Acceptable?

driver/cl_options.cpp
```diff
@@ -193,7 +196,7 @@ static cl::opt<bool, true> unittest("unittest",
 
 cl::opt<std::string>
     ir2objCacheDir("ir2obj-cache", cl::desc("Use <cache dir> to cache object files for whole IR modules (experimental)"),
-            cl::value_desc("cache dir"), cl::Prefix);
+            cl::value_desc("cache dir"));
```
JohanEngelen: this is an unrelated small bug fix, btw

driver/ir2obj_cache.cpp
```diff
@@ -127,20 +127,83 @@ void storeCacheFileName(llvm::StringRef cacheObjectHash,
...
+        continue;
+      // All  "-ir2..." options can be ignored
+      if (arg[1] == 'i' && arg[2] == 'r' && arg[3] == '2')
+        continue;
+      // All effects of -d-version... are already included in the IR hash.
+      if (strncmp(arg+1, "d-version", 9) == 0)
+        continue;
+      // All effects of -unittest are already included in the IR hash.
+      if (strcmp(arg + 1, "unittest") == 0) {
+        break;
```
JohanEngelen: bug!

JohanEngelen: fixed.

JohanEngelen:
ping @klickverbot Green and ready for your approval.

driver/ir2obj_cache.cpp
```diff
@@ -127,20 +127,87 @@ void storeCacheFileName(llvm::StringRef cacheObjectHash,
...
+        // We deal with -O later ("-O" and "-O3" should hash equally, "" and
+        // "-O0" too)
+        continue;
+      }
+      if (arg[1] == 'c' && !arg[2])
+        continue;
+      // All options starting with these characters can be ignored (LLVM does
+      // not have options starting with capitals)
+      if (arg[1] == 'D' || arg[1] == 'H' || arg[1] == 'I' || arg[1] == 'J' ||
+          arg[1] == 'L' || arg[1] == 'X')
```
thewilsonator: Is it really safe to ignore -I and -J? These can be used to implement module variables e.g. -I=linux_impl/ v.s. -I=windows_impl/

JohanEngelen: Yes, but those differences would show up in the hashed IR already.
We only need to add cmdline options that change object file output, but that don't change the IR that is hashed already.

JohanEngelen:
Merging this in the next days, if no further comments are added.

driver/ir2obj_cache.cpp
```diff
@@ -127,20 +127,87 @@ void storeCacheFileName(llvm::StringRef cacheObjectHash,
...
+      // All  "-ir2..." options can be ignored
+      if (arg[1] == 'i' && arg[2] == 'r' && arg[3] == '2')
+        continue;
+      // All effects of -d-version... are already included in the IR hash.
+      if (strncmp(arg+1, "d-version", 9) == 0)
+        continue;
+      // All effects of -unittest are already included in the IR hash.
+      if (strcmp(arg + 1, "unittest") == 0) {
+        continue;
+      }
```
kinke: I'd suggest excluding `-lib` too. No reason not to use the cached objects when outputting a static library.

driver/ir2obj_cache.cpp
```diff
@@ -127,20 +127,87 @@ void storeCacheFileName(llvm::StringRef cacheObjectHash,
...
+      if (strcmp(arg + 1, "run") == 0) {
+        break;
+      }
+    }
 
-  // Let hash depend on the compiler version:
-  hash_os << global.ldc_version << global.version << global.llvm_version
-          << ldc::built_with_Dcompiler_version;
+    // If we reach here, add the argument to the hash.
+    hash_os << arg;
```
kinke: Scenario: project is a static lib. I initially compile it with 3 .d source files. I then add a new file without touching the other 3 ->the 3 cached objects aren't used because there's a new file in the command line.

Only relevant for non-singleobj builds, i.e., static libraries nowadays, but still not ideal. Adhoc I can't come up with a reason for any file in the command line to be included in the hash, so maybe ignoring all args not beginning with `-` is safe.

JohanEngelen: Ah... good thing I wrote comments, this explains it:
```
    // Out of pre-caution, all arguments that are not prefixed with '-' are
    // added to the hash. Such an argument could be a source file "foo.d", but
    // also a value for the previous argument when the equals sign is omitted,
    // for example: "-code-model default" becomes "-code-model" "default".
```
Basically, without fully parsing all the arguments again, I don't know whether an argument that does not begin with `-` is a file or is the parameter of the preceding argument.

driver/ir2obj_cache.cpp
JohanEngelen: OK, done.

driver/ir2obj_cache.cpp
```diff
@@ -127,20 +127,87 @@ void storeCacheFileName(llvm::StringRef cacheObjectHash,
...
+// (Note: config and response files may also add compiler flags.)
+void outputIR2ObjRelevantCmdlineArgs(llvm::raw_ostream &hash_os)
+{
+  // Use a "whitelist" of cmdline args that do not need to be added to the hash,
+  // and add all others. There is no harm (other than missed cache
+  // opportunities) in adding commandline arguments that also change the hashed
+  // IR, which simplifies the code here.
+  // The code does not deal well with options specified without equals sign, and
+  // will add those to the hash, resulting in missed cache opportunities.
+  for (const char *arg : opts::allArguments) {
```
JohanEngelen: nice, done.

PR #1775, opened 2016-09-20
Lit: add testing with lldb

JohanEngelen:
I don't know what happened, but I can no longer debug much with LLDB. Local variables don't show up anymore, and global variables have a `<could not resolve type>`.
I have debugged with lldb before where these variables did show up correctly, so... I don't know.
I have the same problem with LDC0.17 and LDC1.1.0. Perhaps I dreamed it, or it was on Linux, where things worked better.

So: added testing with LLDB.
Testing works similar as with CDB (great stuff @rainers ): lines prefixed with "// LLDB:" are output to a `%t.lldb` file, which is then inserted into LLDB for execution. LLDB output goes into `%t.out.txt` which is then inserted into FileCheck to do the string matching.

### DIFF
--- a/tests/debuginfo/lldb.d
+++ b/tests/debuginfo/lldb.d
@@ -1,0 +1,29 @@
+// REQUIRES: atleast_llvm308
+// REQUIRES: lldb, sed
+// RUN: %ldc -g -of=%t%exe %s \
+// RUN: && sed -e "/^\\/\\/ LLDB:/!d" -e "s,// LLDB:,," %s > %t.lldb \
+// RUN: && %lldb %t%exe -s %t.lldb > %t.out.txt 2>&1 \
+// RUN: && FileCheck %s -check-prefix=CHECK < %t.out.txt
+
+int globalvar = 123;
+
+void main()
+{
+    int a = 42;
+    return;
+}
+
+// CHECK: Current executable set to {{.*}}lldb
+// LLDB: break set --file lldb.d --line 13
+// LLDB: run
+// CHECK:      void main()
+// CHECK-NEXT: {
+// CHECK-NEXT: int a = 42;
+// CHECK-NEXT: -> {{[0-9]+}} return;
+// CHECK-NEXT: }
+// LLDB: frame variable
+// check for frame variable name 'a', broken atm
+// LLDB: target variable
+// check for global variable name 'globalvar', broken atm
+// LLDB: continue
+// LLDB: exit
--- a/tests/lit.site.cfg.in
+++ b/tests/lit.site.cfg.in
@@ -108,3 +108,9 @@ if (platform.system() == 'Windows') and (config.default_target_bits == 64):
 if (platform.system() == 'Windows') and os.path.isfile( cdb ):
     config.available_features.add('cdb')
     config.substitutions.append( ('%cdb', '"' + string.replace( cdb, '\\', '\\\\') + '"') )
+
+# Add lldb substitution
+if platform.system() == 'Darwin':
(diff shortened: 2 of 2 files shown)


JohanEngelen:
OK, at least with `-gc` things work a little better: i see local variables with `-gc`.
Perhaps LLDB tries to reckognize the `DW_LANG_D` language, sees that it doesn't know a whole lot about it and then things don't show up.

JohanEngelen:
> Perhaps LLDB tries to reckognize the DW_LANG_D language, sees that it doesn't know a whole lot about it and then things don't show up.

Yep.
Found the culprit of missing LLDB functionality: https://reviews.llvm.org/D24794

JohanEngelen:
My plan is to keep this one test LLDB running on the CI system, so that we know what testing commands work (troubles on the Travis atm, but it runs fine on my Mac). And in the meanwhile, I'll try to mimic the CDB tests for LLDB with `REQUIRES: lldb40` (or something like that, it corresponds to lldb trunk version at the moment), that I can run from time to time on my own machine and that we can "release" onto the CI once the Mac lldb version is updated.

JohanEngelen:
Ah, the hanging test is a codesigning issue probably (opens a dialog).

rainers:
Looks good. I was just about to add something for gdb, too.

> OK, at least with -gc things work a little better: i see local variables with -gc

The dmd-testsuite gdb tests work slightly better with `-g` instead of `-gc`.

tests/debuginfo/lldb.d
```diff
@@ -0,0 +1,31 @@
+// REQUIRES: atleast_llvm308
+// REQUIRES: lldb, sed
+
+// RUN: %ldc -g -of=%t%exe %s \
+// RUN: && sed -e "/^\\/\\/ LLDB:/!d" -e "s,// LLDB:,," %s > %t.lldb \
```
rainers: AFAICT you don't need \ and &&, just a new RUN: line will execute the commands in sequence.

JohanEngelen: Yeah perhaps you are right. In the past I had issues with tests running in parallel, so I started using the `\ &&` thing, but it may have been cross-file instead of within a file.

rainers:
> OK, at least with -gc things work a little better: i see local variables with -gc.

Maybe it is also DWARF-4 that is confusing lldb. There is now a hidden switch that allows restriction to DWARF-2 with `-dwarf-version=2`.

JohanEngelen:
LLDB knows about the D language, but then finds that it doesn't know any details about it. So it just does nothing. It's fixed by https://reviews.llvm.org/D24794

JohanEngelen:
This PR is stuck on getting LLDB codesigned on TravisCI Mac, btw.

I hope to find the energy to install lldb trunk on CircleCI, so we can test things there. I started a discussion on adding D demangling support on LLDB's mailing list, and some nice ideas are floating around.

JohanEngelen:
Codesigning hint: perhaps something like this works? (https://github.com/Homebrew/homebrew-versions/blob/master/llvm38.rb#L8)
`quiet_system "/usr/bin/codesign", "-f", "-s", "lldb_codesign", "--dryrun", "llvm_check"`

JohanEngelen:
LLDB 4.0 will treat D as if it is C (instead of doing nothing): http://llvm.org/viewvc/llvm-project?view=revision&revision=287917

9il:
> LLDB 4.0 will treat D as if it is C (instead of doing nothing): http://llvm.org/viewvc/llvm-project?view=revision&revision=287917

Thanks!

PR #1776, opened 2016-09-21
add basic gdb test

rainers:
Similar to the cdb tests.

### DIFF
--- a/tests/debuginfo/strings_gdb.d
+++ b/tests/debuginfo/strings_gdb.d
@@ -1,0 +1,49 @@
+// REQUIRES: atleast_llvm308
+// REQUIRES: Linux
+// RUN: %ldc -g -of=%t %s
+// RUN: sed -e "/^\\/\\/ GDB:/!d" -e "s,// GDB:,," %s \
+// RUN:    | gdb %t >%t.out
+// RUN: FileCheck %s -check-prefix=CHECK < %t.out
+
+int main(string[] args)
+{
+    string[] nargs = args;
+    string ns = "a";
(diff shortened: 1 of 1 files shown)


rainers:
> Does this duplicate the tests in dmd-testsuite?

A bit. runnable/gdb1.d covers the argument passing, but all you can check there is a single regex match within the full output.
BTW: we could enable all but one gdb tests in the dmd-testsuite with a small patch to d_do_test.

rainers:
> So you figured out the issue? We can just try and upstream the patch as well.

The general problem is that the initial source file is not the compiled module, but __entrymain.d, so that breakpoint operation don't find the lines. The patch adds a command to switch the current source file.

I had the impression that the actual object file order is under discussion, so this could change again soon.

rainers:
> Users would run into this too, so we should definitely look into fixing this on our side.

I guess we should to set the DW_AT_main_subprogram attribute, but I couldn't find support for it in LLVM. Is there a way to pass DWARF attributes to the object file writer?

It probably works in dmd because "main" is the only function with line number info. LDC produces bad entries for empty Loc() information (containing "." as filename), I think avoiding that could help here too.

rainers:
> The general problem is that the initial source file is not the compiled module, but __entrymain.d

Just figured that this is solved by using -g instead of -gc, so it seems gdb already knows about the correct main function depending on the source language. See https://github.com/ldc-developers/ldc/pull/1784

rainers:
Closing as it doesn't add a lot to the dmd-testsuite cases.

PR #1784, opened 2016-09-24
enable gdb tests in the dmd-testsuite

rainers:
Another attempt at getting these running. gdb works much better if `-g` is used instead of `-gc` when compiling, as gdb then uses the correct initial source file and interprets global `uint` variables correctly.

@kinke I hope it's ok that I abuse the gdb branch by merging current dmd-testsuite's ldc.

One known issue is "optimized-out" variable in gdb10311.d.

see also https://github.com/ldc-developers/ldc/pull/1658

### DIFF
--- a/tests/d2/CMakeLists.txt
+++ b/tests/d2/CMakeLists.txt
@@ -17,11 +17,28 @@ function(add_testsuite config_name dflags model)
     # testsuite build system provides no way to run the test cases with a
     # given set of flags without trying all combinations of them.
     add_test(NAME ${name}
-        COMMAND make -k -C ${PROJECT_SOURCE_DIR}/tests/d2/dmd-testsuite RESULTS_DIR=${outdir} DMD=${LDMD_EXE_FULL} DFLAGS=${dflags} MODEL=${model} quick
+        COMMAND make -k -C ${PROJECT_SOURCE_DIR}/tests/d2/dmd-testsuite RESULTS_DIR=${outdir} DMD=${LDMD_EXE_FULL} DFLAGS=${dflags} MODEL=${model} GDB_FLAGS=${gdb_flags} quick
     )
     set_tests_properties(${name} PROPERTIES DEPENDS clean-${name})
 endfunction()
 
+if((${CMAKE_SYSTEM_NAME} MATCHES "Linux") AND (${LDC_LLVM_VER} GREATER 307))
+    execute_process(COMMAND gdb --version
+                    COMMAND head -n 1
+                    OUTPUT_VARIABLE GDB_VERSION
+                    OUTPUT_STRIP_TRAILING_WHITESPACE)
+    string(REGEX REPLACE "[^0-9]*([0-9]+[0-9.]*).*" "\\1" GDB_VERSION "${GDB_VERSION}")
+    message(STATUS "GDB ${GDB_VERSION} detected")
+    if(GDB_VERSION VERSION_LESS "7.6.1")
+        set(gdb_flags "ON_NOTLS")
+    else()
+        set(gdb_flags "ON")
+    endif()
+else()
+    set(gdb_flags "OFF")
+endif()
+
+string(REGEX REPLACE "[^0-9]*([0-9]+[0-9.]*).*" "\\1" GDB_VERSION "${GDB_VERSION}")
 # Would like to specify the "-release" flag for release builds, but some of the
 # tests (e.g. 'testdstress') depend on contracts and invariants being active.
 # Need a solution integrated with d_do_test.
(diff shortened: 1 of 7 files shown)


kinke:
Oh really cool that you found such an easy fix! :+1: I'll probably have used `-g` in my manual tests before, without realizing `-gc` was actually used. Recycling the dmd-testsuite branch is obviously fine.

rainers:
Unfortunately it doesn't work on the build servers, probably because they use ancient gdb versions without D support (travis 7.4 of 2012, circle 7.8 of 2014). My attempts at updating these on travis failed so far. Any ideas?

rainers:
Not sure. The breakpoints seem to work, but symbol lookup fails. Here's a random example 
https://travis-ci.org/ldc-developers/ldc/jobs/162413364#L3307:
```
Breakpoint 1 at 0x409c24: file gdb10311.d, line 19.
warning: no loadable sections found in added symbol-file system-supplied DSO at 0x2aaaaaacd000
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Breakpoint 1, D main () at gdb10311.d:20
20	}
RESULT=/home/travis/build/ldc-developers/ldc/dmd-testsuite-debug/runnable/gdb10311_0.gdb:4: Error in sourced command file:
No symbol "x" in current context.
```

rainers:
> Shouldn't the basic things work even without D support? Or does the D language tag cause it to give up entirely?

I've installed a VM similar to travis locally and figured that gdb 7.4 doesn't seem to support DWARF-4 debug info. If I switch to DWARF-2, variable display works.
Should we add a switch to restrict the DWARF version?

kinke:
Good catch Rainer.

rainers:
> Good catch(es), Rainer.

Thanks. It's not been perfect pleasure, though. I hope we can reduce the pain in the near future, e.g. by reducing the number of versions supported or using newer OS versions for integration builds.

tests/d2/CMakeLists.txt
```diff
@@ -6,30 +6,53 @@ elseif(${ptr_size} MATCHES "^8$")
...
         COMMAND ${CMAKE_COMMAND} -E remove_directory ${outdir})
 
+    set(all_dflags ${dflags} ${gdb_dflags})
+
     # The DFLAGS environment variable read by LDMD is used because the DMD
     # testsuite build system provides no way to run the test cases with a
     # given set of flags without trying all combinations of them.
     add_test(NAME ${name}
-        COMMAND make -k -C ${PROJECT_SOURCE_DIR}/tests/d2/dmd-testsuite RESULTS_DIR=${outdir} DMD=${LDMD_EXE_FULL} DFLAGS=${dflags} MODEL=${model} quick
+        COMMAND make -k -C ${PROJECT_SOURCE_DIR}/tests/d2/dmd-testsuite RESULTS_DIR=${outdir} DMD=${LDMD_EXE_FULL} DFLAGS=${all_dflags} MODEL=${model} GDB_FLAGS=${gdb_flags} quick
```
rainers: Two many quotes already in all_dflags. string(CONCAT) will help, hopefully.

kinke:
Travis is a bit slow. ;) Their beta environment, since October 2015, is based on Ubuntu 14.04 with gdb v7.7. Unfortunately, it's not yet available for their container-based (`sudo`-less) infrastructure, which does improve the timings considerably IIRC (Johan did a test a few months ago).

kinke:
Thanks for enduring the pain! :)

JohanEngelen:
For these long strings of trial-and-error things, I much prefer people to use the "Approve" button so that the author can clean up / squash the commit history and merge it himself. It makes the git history much easier to read and to work with, thanks.

kinke:
If you had cared to check, I did squash the commits.

JohanEngelen:
Excellent, GH doesn't clearly say that that's what happened.

Also: no tests for the -dwarf-version option.

rainers:
> If you had cared to check, I did squash the commits.

Thanks for merging and squashing. I didn't expect it to pass so soon... :)

rainers:
> Also: no tests for the -dwarf-version option.

True, no explicit test, but the gdb tests on travis and circle-ci are run with this option.

JohanEngelen:
> True, no explicit test, but the gdb tests on travis and circle-ci are run with this option.

But those could be adjusted some time later and then gone is the testing. Also, the current code does not add the Dwarf version when none is specified, but this is not tested (I think it will help in passing bitcode files on the cmdline).
(It's not a big deal in this case, but I keep hammering on tests, because we have so few already. Also, often thinking about and adding tests helps discover corner cases and helps in making things more robust.)

PR #1786, opened 2016-09-26
Dcompute

thewilsonator:
DCompute adds capability to target CUDA and SPRIV backends of llvm.

Several misc comments.
- ldc will segfault if -sinlgeobj is passed and the first module on the cmd line is marked `@compute`, I am yet to think of a solution for that.
- I feel most of the `if(!dct)` that stop PGO and DI generation could be done more elegantly, but I don't know how they are implemented, so suggestions are welcome.
- It is possible I have added unused functions
- Help and suggestions for TODOs much appreciated
- ldc is less ignorant of addrspaces now, but I might have missed some places where they are not taken into account.
- clang format may need to be run at some point.


9il:
Can a `@kernel` function be used as normal function? This is important to reduce code dublication

thewilsonator:
From `@compute` code yes, not from the host though.

9il:
Can functions between different `@compute` be inlined?

thewilsonator:
In case you were slightly confused `@compute` code is code that is compiled for the device. This includes `@kernels` and other function defined in the file. Compilation mode is effectively -singleobj. AFAIK all optimisations are backend specific. SPRIV optimisations are performed by the "interpreter" and the .ptx goes through nvidia's assembler although does use the llvm optimisation pipeline. 

It would be best if the comments in this thread were specifically about the PR.

thewilsonator:
CircleCI is failing because it's trying to build with llvm4.0, which it can't find.

kinke:
Very interesting; SPRIV = SPIR-V. ;)
Do you know if that could be used to compile regular graphics shaders for OpenGL/Vulkan too, or is it just for CUDA/OpenCL?

thewilsonator:
Currently Khronos's SPIRV-LLVM only has "intrinsics"\* for OpenCL, So if they added the intrinsics and we added the correct metadata then, in theory yes. At the moment, no.

\*by which I mean magically (c++) mangled function names

gen/functions.cpp
```diff
@@ -733,7 +733,7 @@ void defineParameters(IrFuncTy &irFty, VarDeclarations &parameters) {
       ++llArgIdx;
     }
 
-    if (global.params.symdebug)
+    if (global.params.symdebug && !gDComputeTarget)
```
JohanEngelen: this check can be removed completely, because it is already checked in DIBuilder::EmitLocalVariable (like other debuginfo functions)

gen/functions.cpp
```diff
@@ -869,7 +869,9 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   IrFunction *irFunc = getIrFunc(fd);
 
   // debug info
-  irFunc->diSubprogram = gIR->DBuilder.EmitSubProgram(fd);
+  if (!gDComputeTarget) {
```
JohanEngelen: this check should be moved inside of EmitSubProgram

gen/llvmhelpers.cpp
```diff
@@ -912,8 +912,10 @@ void DtoVarDeclaration(VarDeclaration *vd) {
     }
 
     irLocal->value = allocainst;
-
-    gIR->DBuilder.EmitLocalVariable(allocainst, vd);
+    if (!gDComputeTarget) {
```
JohanEngelen: move check inside EmitLocalVariable

gen/pgo.cpp
```diff
@@ -884,7 +889,9 @@ void CodeGenPGO::applyFunctionAttributes(llvm::Function *Fn) {
 void CodeGenPGO::emitCounterIncrement(const RootObject *S) const {
   if (!global.params.genInstrProf || !RegionCounterMap || !emitInstrumentation)
     return;
-
+  //Never generate PGO code for dcompute
+  if (gDComputeTarget)
```
JohanEngelen: emitInstrumentation should be false for gDComputeTarget, so this check is not needed.

gen/pgo.cpp
```diff
@@ -1089,6 +1096,9 @@ void CodeGenPGO::emitIndirectCallPGO(llvm::Instruction *callSite,
 
 void CodeGenPGO::valueProfile(uint32_t valueKind, llvm::Instruction *valueSite,
                               llvm::Value *value, bool ptrCastNeeded) {
+  //Never generate PGO code for dcompute
+  if (gDComputeTarget)
```
JohanEngelen: check is not needed. For DCompute functions, `emitInstrumentation == false` and `ProfRecord` should be false too.

gen/statements.cpp
```diff
@@ -117,13 +119,13 @@ class ToIRVisitor : public Visitor {
   void visit(ReturnStatement *stmt) LLVM_OVERRIDE {
     IF_LOG Logger::println("ReturnStatement::toIR(): %s", stmt->loc.toChars());
     LOG_SCOPE;
-
+
```
JohanEngelen: clang format this file

gen/pgo.cpp
thewilsonator: You could want to do a PGO build that also runs Kernels.

gen/statements.cpp
```diff
@@ -954,8 +986,9 @@ class ToIRVisitor : public Visitor {
       // For PGO instrumentation, we need to add counters /before/ the case
       // statement bodies, because the counters should only count the jumps
       // directly from the switch statement and not "goto default", etc.
+      // never generate PGO code for dcompute.
       llvm::SwitchInst *si;
-      if (!global.params.genInstrProf) {
+      if (!global.params.genInstrProf || dct) {
```
JohanEngelen: tell @johanengelen that this is a bug: instrumentation profiling setting is done per function, not globally.

gen/statements.cpp
```diff
@@ -988,7 +988,7 @@ class ToIRVisitor : public Visitor {
       // directly from the switch statement and not "goto default", etc.
       // never generate PGO code for dcompute.
       llvm::SwitchInst *si;
-      if (!global.params.genInstrProf || dct) {
+      if (!global.params.genInstrProf && !dct) {
```
JohanEngelen: I think the logic was correct:  if   (we don't generate PGO instrumentation)  OR  (we can't generate PGO instrumentation, in other words `dct != nullptr`)  then do this

PR #1789, opened 2016-09-27
rename codeview tests to *_cdb.d

rainers:
as suggested in https://github.com/ldc-developers/ldc/pull/1769#issuecomment-248641890


JohanEngelen:
The files also contain their own names sometimes, e.g.
```
// CDB: ld /f cvbaseclass*
// enable case sensitive symbol lookup
// CDB: .symopt-1
// CDB: bp `cvbaseclass.d:27`
```

rainers:
> The files also contain their own names sometimes, e.g.

Ouch, shouldn't have thought it is that easy (and actually run the tests locally). Updated and also increased required LLVM version to 3.9 to not pretend that CodeView information existed bfore that.

tests/debuginfo/basictypes_cdb.d
```diff
@@ -1,10 +1,11 @@
-// REQUIRES: atleast_llvm308
+// REQUIRES: atleast_llvm309
 // REQUIRES: Windows
 // REQUIRES: cdb
 // RUN: %ldc -g -of=%t.exe %s
 // RUN: sed -e "/^\\/\\/ CDB:/!d" -e "s,// CDB:,," %s \
 // RUN:    | %cdb -snul -lines -y . %t.exe >%t.out
 // RUN: FileCheck %s -check-prefix=CHECK -check-prefix=%arch < %t.out
+module basictypes;
```
JohanEngelen: Different from the filename, but I think that's actually a good thing: later on you test that indeed the module name (and not the filename!) is used for the type name.

rainers: I was moving a bit to and fro about whether to add the _cdb to the module name or not. Should I remove the suffix from the module names in the other files, too? They have some very generic names, though, that might be confused with aother symbols.

JohanEngelen: (this is all repurposing the test cases to do more than what they were intended for. Perhaps a separate file is better for this, but ok.)

rainers:
> I would actually also remove module ... in one of the files, and check that the filename is used instead to print a variable name.

Done.

rainers:
Rebased to please CirleCI.

PR #1793, opened 2016-09-27
[wip] Add initial ThinLTO support (LLVM >= 3.9)

JohanEngelen:
For information about ThinLTO, read: http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html

All that is needed inside LDC when ThinLTO is enabled:
- output bitcode instead of object code (but keep the `.o` extension)
- add a module summary index to the bitcode file

**What commandline flag name shall we give this?** 

ThinLTO needs linker support, and testing it on our CI systems is nigh impossible at the moment. :/
Tested with XCode 8:
```d
// file b.d
extern(C) void doesNothing() {}
```

```d
// file a.d
extern(C) void doesNothing(); // declaration only

void main() {
    for (ulong i = 0; i < 100_000_000; ++i) {
        doesNothing();
    }
}
```

Separate compilation (the usual):
```
> ../bin/ldc2 -c b.d -O3 -of=b.o
> ../bin/ldc2 a.d b.o -of=a -O3
> time ./a
./a  1.77s user 0.01s system 98% cpu 1.802 total
```

Separate compilation with ThinLTO:
```
> ../bin/ldc2 -c b.d -O3 -of=bthin.o -thinlto
> ../bin/ldc2 a.d bthin.o -of=athin -O3 -thinlto
> time ./athin
./athin  0.00s user 0.00s system 61% cpu 0.006 total
```

BAM!

### DIFF
--- a/driver/toobj.cpp
+++ b/driver/toobj.cpp
@@ -19,6 +19,9 @@
 #include "gen/programs.h"
 #include "llvm/IR/AssemblyAnnotationWriter.h"
 #include "llvm/IR/Verifier.h"
+#if LDC_LLVM_VER >= 309
+#include "llvm/Analysis/ModuleSummaryAnalysis.h"
+#endif
 #include "llvm/Bitcode/ReaderWriter.h"
 #if LDC_LLVM_VER >= 307
 #include "llvm/IR/LegacyPassManager.h"
@@ -54,6 +57,14 @@ using LLErrorInfo = std::string;
 static llvm::cl::opt<bool>
     NoIntegratedAssembler("no-integrated-as", llvm::cl::Hidden,
                           llvm::cl::desc("Disable integrated assembler"));
+
+#if LDC_LLVM_VER >= 309
+static llvm::cl::opt<bool>
+    EnableThinLTO("thinlto",
+                  llvm::cl::desc("Enable ThinLTO (requires linker support)"));
+#else
+const bool EnableThinLTO = false;
+#endif
 
 // based on llc code, University of Illinois Open Source License
 static void codegenModule(llvm::TargetMachine &Target, llvm::Module &m,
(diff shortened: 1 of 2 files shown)


kinke:
> BAM!

:D - definitely interesting.

JohanEngelen:
> The real benefit comes from being able to parallelise large LTO builds and make the incremental, though. Your example would have been trivial to achieve with current LTO.

Yes, it's just an example to show that it works. Current LTO (you mean, outputting bitcode and feeding that back into LDC right?) would be a little more work than just passing `-thinlto` as build flag (LDC does not recognize bitcode files in disguise with `.o` extension).
The (machine codegen part of the) build is already parallel without extra changes needed, and incremental builds+speedup are obtained by setting `generate ThinLTO hash` to true (I think for that case we should just disable our own hashing+cache btw).

The next step is getting this to work on Linux (build compiler plugin, and feed that to the linker automatically).

JohanEngelen:
Ideas for the cmdline flag?   `-thinlto` is good?

JohanEngelen:
>  and incremental builds+speedup are obtained by setting generate ThinLTO hash to true

... and setting a linkerflag to point to a cache dir.

kinke:
> `-thinlto` is good?

Sounds good to me.

JohanEngelen:
Clang has the option `-flto=full` and `-flto=thin`. The linkers (XCode 8 linker, and gold linker with LTO plugin installed) can deal with both, it seems. So we could just mimic Clang for the cmdline option (perhaps without the "f").

JohanEngelen:
Looks like it. Haven't looked into it. (I don't know how the linker selects whether to use full LTO or thin LTO)

JohanEngelen:
reminder to myself:  Keep the inline asm workaround for LLVM3.9; should be fixed for LLVM4.0
https://llvm.org/bugs/show_bug.cgi?id=30610

PR #1795, opened 2016-09-28
dcompute: Enable linking with libLLVMSPRIVlib.a

thewilsonator:


CMakeLists.txt
```diff
@@ -368,6 +376,8 @@ set(LDC_CXX_SOURCE_FILES
     ${GEN_HDR}
     ${IR_SRC}
     ${IR_HDR}
+    ${DCOMPUTE_SRC}
```
klickverbot: I don't think having a kitchen-sink `dcompute` directory makes much sense (looking at the combined PR); the files should mostly just go into `gen`. We might want to start organising the code in `gen/` into smaller, more manageable units (e.g. `gen/abi`, etc.), but having `dcompute` as a unit of encapsulation doesn't seem like the natural thing to do. After all, your work touches many different parts of the compilation process.

CMakeLists.txt
```diff
@@ -17,10 +17,25 @@ if (NOT D_COMPILER_DMD_COMPAT)
...
+        else()
+            set(${variable} "${${variable}} ${value}" PARENT_SCOPE)
+        endif()
+    endforeach(variable)
+endfunction()
+
+option(LDC_WITH_DCOMPUTE_SPIRV "Enable Dcompute to generate SPIRV kernels for targetting OpenCL, requires a khronos llvm")
+mark_as_advanced(LDC_WITH_DCOMPUTE_SPIRV)
+if(LDC_WITH_DCOMPUTE_SPIRV)
+    append( EXTRA_LLVM_MODULES sprivlib ) #yes it is libLLVMSPIRVLib
```
thewilsonator: Because the library is libLLVMSPIRVLib.a because Khronos decided to add a redundant lib at the end when compared to all the other llvm libs. So someone doesn't accidentally change it to `spriv`. Style point noted.

CMakeLists.txt
```diff
@@ -17,10 +17,25 @@ if (NOT D_COMPILER_DMD_COMPAT)
...
+    foreach(variable ${ARGN})
+        if(${variable} STREQUAL "")
+            set(${variable} "${value}" PARENT_SCOPE)
+        else()
+            set(${variable} "${${variable}} ${value}" PARENT_SCOPE)
+        endif()
+    endforeach(variable)
+endfunction()
+
+option(LDC_WITH_DCOMPUTE_SPIRV "Enable Dcompute to generate SPIRV kernels for targetting OpenCL, requires a khronos llvm")
```
thewilsonator: Noted.

CMakeLists.txt
klickverbot: A vague, flippant comment certainly doesn't help with avoiding accidental changes. We also need to add a SPIR-V CI build anyway for this to have any chance of staying in shape.

thewilsonator: I put the comment in because I felt that the library name was unintuitive with the redundant extra `lib` on the end and wanted to make that clear. I can remove it.

klickverbot: Remove it, or rephrase it so it is clearer. Currently, it just adds to the confusion, especially since the name of the LLVM module (`spirvlib`) has nothing to do with the name of the .a file (`libLLVMSPIRVLib`).

CMakeLists.txt
```diff
@@ -17,10 +17,25 @@ if (NOT D_COMPILER_DMD_COMPAT)
...
+        else()
+            set(${variable} "${${variable}} ${value}" PARENT_SCOPE)
+        endif()
+    endforeach(variable)
+endfunction()
+
+option(LDC_WITH_DCOMPUTE_SPIRV "Enable Dcompute to generate SPIR-V kernels for targetting OpenCL, requires a Khronos llvm")
+mark_as_advanced(LDC_WITH_DCOMPUTE_SPIRV)
+if(LDC_WITH_DCOMPUTE_SPIRV)
+    append(EXTRA_LLVM_MODULES sprivlib) #The .a is libLLVMSPIRVLib with the extra "Lib" on the end so need sprivlib not spirv
```
klickverbot: Sorry, that's still confusing, as pointed out earlier. The comment suggests that `spirvlib` (or the fact that it ends in `lib`) somehow influences the name of the library linked. That's not the case. `spirvlib` is the name of the LLVMBuild component, so we need to use that to look up the library using `llvm-config`/`LLVMConfig.cmake`, period. There is no conceptual link between `spirvlib` and `libLLVMSPIRVLib`. Even if the name of the latter was changed to `libLLVMSPIRVFactoryBuilderLibraryProvider`, we'd still use `spirvlib` if that's what the LLVM component is called internally. (The fact that LLVMConfig.cmake – but not llvm-config – builds the default library name from the upper-cased component name by default is irrelevant.)

I'm going on about this minor detail because it is an example for the most harmful kind of comments: The explanation is seemingly plausible, but is actually incorrect backwards reasoning. In such cases, readers would be much better off without a comment in the first place – yes, they'd need to look up the details themselves, but at least wouldn't walk away with wrong assumptions.

thewilsonator: OK, will remove.

thewilsonator:
It should hopefully be as simple as pointing the llvm source to https://github.com/KhronosGroup/SPIRV-LLVM/tree/khronos/spirv-3.6.1 or
https://github.com/KhronosGroup/SPIRV-LLVM/tree/khronos/spirv-3.8 but I would assume that building llvm from source each time would put waaay to much pressure. Is there a way to cache the generated binaries?

JohanEngelen:
CircleCI is currently used for testing LLVM trunk, and we shouldn't lose that. Getting SPIRV testing on CircleCI needs someone to figure out how to do multiple build environments on CircleCI.

JohanEngelen:
Is `sprivlib` correct or is it a typo? (order of R and I)
(Also note the title of the PR)

thewilsonator:
Oooh yeah, wtf is going on with my build, it is passing but no -lLLVMSPIRVLib appears on my build command.

thewilsonator:
The module is called `spirvlib` according to `llvm-config --components`. But it's still not working, is anybody able to get this to work? Otherwise I will have to resort to appending libLLVMSPIRVLib.a (or win equivalent) to ${LLVM_LIBRARIES}, which I really don't want to have to.

thewilsonator:
> Are there binary packages for these LLVM builds available somewhere?
Not that I can find.
> So what does llvm-config --libnames spirvlib give you?
`libLLVMSPIRVLib.a` and its dependencies.
and `llvm-config --libs spirvlib`
gives `-lLLVMSPIRVLib` and dependencies.

I think I must be going crazy.

thewilsonator:
Hmm I didn't mean to sync on this. Dunno what happend, the joys of hard reset + force push I guess?

thewilsonator:
SPIRV is a valid target now taking this redundant.

PR #1797, opened 2016-09-29
dcompute: [NFC] Refactor emission of debug info.

thewilsonator:
Better annotate reasons for not emitting debug info. Also make it
easier to change conditions under which DI should be emitted.

This will make it easier when more dcompute related code is added, as
there cannot be any DI emission for dcompute (SPIRV does allow some
extra debug info but that is not in the current scope of dcompute yet, and done
a very different way, ~~not sure about CUDA~~ CUDA appears to support dwarf, with some extensions).

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -74,6 +74,8 @@ llvm::StringRef uniqueIdent(Type* t) {
 
 } // namespace
 
+bool mustEmitDebugInfo() { return global.params.symdebug; }
+
 ////////////////////////////////////////////////////////////////////////////////
 
 // get the module the symbol is in, or - for template instances - the current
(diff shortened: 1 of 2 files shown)


JohanEngelen:
I think `mustEmitDebugInfo` should be a member of DIBuilder.  (In the future, it could be very convenient to have the outcome depend on a member variable of DIBuilder)
Otherwise, it LGTM.

kinke:
Hmm, how about making `DIBuilder` an interface (with default methods doing nothing) and the actual implementation a subclass? Then we could just use a `DIBuilder` base instance to skip debuginfo generation and get rid of these repeated checks.

thewilsonator:
If we had multiple type of debug out (dwarf, spirv,pdb,...)  then I would agree, but as it stands I don't think the extra indirection is worth it.

thewilsonator:
Wow that std.digest.sha unittest is HUUGE, I wonder why the change to a method would push it to over 10 minutes running time (thus timing out travis).

thewilsonator:
@JohanEngelen ping

JohanEngelen:
I think @kinke's idea makes sense, perhaps also for PGO stuff.

> If we had multiple type of debug out (dwarf, spirv,pdb,...) then I would agree

But with dcompute, we will soon have that second debug output, right? ;-)
Pulling so you can continue with the work.

PR #1800, opened 2016-09-30
dcompute: ldc.attributes @kernel and @compute

thewilsonator:
This will also require addition of 

``` D
///Readability aliases for compute
enum deviceOnly = 0;
enum hostAndDevice = 1;
///Mark a module to be compiled for dcompute.
struct compute {
     int alsoProduceHostCode;
}

///Mark a function as a compute API entry point
private struct _kernel {}
immutable kernel = _kernel();
```

to ldc.attributes.


thewilsonator:
Done.

9il:
@thewilsonator Could you please describe this attributes? I am worry about ability to reduce code duplication between common code and OpenCL code.

thewilsonator:
`@compute` marks a module for compilation for OpenCL/CUDA exclusive of the host.
`@kernel` marks a function that it should be callable from the host.

I think code duplication is somewhat inevitable, as the function signatures
(out of necessity) and style of code (both the code-model and optimisation).

Could possibly add another for functions for both host and device, to reduce 
duplication, with the same restrictions of the device.

thewilsonator:
expression->toInteger() works for bools right? the exp->toBoolean takes a scope and returns en Expression *, so I'm a bit confused.

thewilsonator:
Thats llvm 4.0 stuff thats failing.

9il:
If the devise is a CPU then a device function should be callable on the host. In addition @compute code should be able to call a generic functions from normal modules, if they can be inlined or instantiated for a device, even if this device is GPU.

thewilsonator:
>If the device is a CPU then a device function should be callable on the host. 

No. `@kernel`s are only accessible through the compute API (OpenCL/CUDA) other device function are device only, (above changes notwithstanding) and are not available to the host (even if the host is the device). Looking at mir, it would seem most of the auxiliary (i.e. non-kernel functions), would still be accessible to the host with the 6d67aa8 change.

> In addition @compute code should be able to call a generic functions from normal modules, if they can be inlined or instantiated for a device, even if this device is GPU.

This will be harder, as it would mean traversing the call graph and recursively trying to codegen function, I may do it at a later time. It may work automagically for templates, depending on what gets instantiated where, but don't count on it.

ZombineDev:
@thewilsonator Correct me if I am wrong, but shouldn't kernel functions be a subset of the functions callable from the host? I.e. device can use only kernel functions, whereas the host should be able to call any fuction. Obviously, this implies that kernel functions would get codegen-ed twice - first for the device intermediate representation (e.g. SPIR-V for OpenCL, Vulkan and OpenGL 4.5+ and PTX for CUDA, or whatever they use) and a second time for the host.

IMO, code sharing between host and device is very important strategically for D. It is ok if this doesn't work for the first iteration of the implementation, but the design should support seamless integration, similar to C++AMP, i.e. writing host and device functions in the same module. Anyway, this discussion is probably better for the newsgroup.

thewilsonator:
Kernel function are the ONLY functions that exist on the *device* that are callable by the host, and can only be called via the compute API.

 You can have functions that were compiled for both the host AND the device with `@compute(true)`, to reduced duplication, but you can't have, in the same module, some functions exclusively for the host and some exclusively for the device, and definitely not intermingled like SYCL. i.e. it's an all or nothing approach, all for the device(`@compute(false)`) all for the host (no attributes) or all for both host and device (`@compute(true)`), but NOT half and half.

Re:Vulkan, unfortunately we are limited by Khronos's llvm fork here that only compiles for OpenCL, due to all the whacky metadata required, and that fact that the SPIR-V, back end is not a proper backend. I agree that it would be awesome though!

ZombineDev:
Why is this limitation needed? What's prevents mixing normal functions and functions marked as `@compute(true)` in the same module?

thewilsonator:
By normal I assume you mean host only? Because, as I said its all or nothing. Compilation is done per module, and the module (as a whole) is either compiled for the host, the device or both. Doing it per function would be a PITA, and greatly increase the size of the changes necessary (read reviewing time and effort).

Doing it like this 
- avoids having the user specify per function wether it is host, device or both which would get tedious (and thus error prone, since they already have to specify it per module, no need to do it twice)
- provides a logical distinction between what is host only, device only, or shared between both. This makes for a simpler approach, and this domain is already very complex.
- will make issuing compile time errors (as opposed to linker errors) easier, by checking what module it was defined in, as opposed to inspecting the call graph (like some mark and recurse kind of algorithm, or something else), for invalid function use, use of globals, use of gc, function pointers etc.

thewilsonator:
Are there any further comments/concerns/considerations? 
I am under the assumption that `BoolExp->toInteger()` works the way I think it does, i.e. b ? 1 : 0.

All other not already open dcompute PRs depend on this PR.

gen/uda.cpp
```diff
@@ -304,34 +310,58 @@ void applyFuncDeclUDAs(FuncDeclaration *decl, IrFunction *irFunc) {
   }
 }
 
-/// Checks whether 'sym' has the @ldc.attributes._weak() UDA applied.
-bool hasWeakUDA(Dsymbol *sym) {
+StructLiteralExp *namedAttr(Dsymbol *sym, std::string name, bool errorIfFound,
+                            const char *errmsg, ArrayParam<Type *> elemTypes) {
```
thewilsonator: done

thewilsonator:
@JohanEngelen any further changes necessary?

JohanEngelen:
:( should have done squash+merge

PR #1801, opened 2016-10-01
dcompute: addrspace awareness

thewilsonator:
LDC currently assumes that there is only one addrspace, zero.
dcompute requires the use of multiple AS to deal with global/shared
generic pointer. This PR addresses this issue.

There are only two root uses (bar asserts) that I have identified, (there may
be more, but libmir/dcompute build fine with #1786 HEAD~3) which is
generating an invalid  bitcast, because the types are not equal, but
for the purposes of copying do not require, and cannot be done with, a
bitcast.

Most suspicious type comparisons seem to resort to using DtoBitCast. 
Further testing will be required to see if any resulting from 
`gIR->ir->CreateBitCast(...)` are valid in dcompute 
(i.e. not classes,aa,typeinfo etc.)

### DIFF
--- a/gen/tollvm.cpp
+++ b/gen/tollvm.cpp
@@ -531,7 +531,7 @@ void DtoAlignedStore(LLValue *src, LLValue *dst) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-LLType *StripAddrSpaces(LLType *t)
+LLType *stripAddrSpaces(LLType *t)
 {
   // Fastpath for normal compilation.
   if(gIR->dcomputetarget == nullptr)
@@ -549,7 +549,7 @@ LLType *StripAddrSpaces(LLType *t)
 }
 
 LLValue *DtoBitCast(LLValue *v, LLType *t, const llvm::Twine &name) {
-  if (StripAddrSpaces(v->getType()) == t) {
+  if (stripAddrSpaces(v->getType()) == t) {
     return v;
   }
   assert(!isaStruct(t));
--- a/gen/tollvm.cpp
+++ b/gen/tollvm.cpp
@@ -549,6 +549,12 @@ LLType *stripAddrSpaces(LLType *t)
 }
 
 LLValue *DtoBitCast(LLValue *v, LLType *t, const llvm::Twine &name) {
+  // Strip addrspace qualifications from v before comparing types by pointer
+  // equality. This avoids the case where the pointer in { T addrspace(n)* }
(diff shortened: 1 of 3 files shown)


thewilsonator:
This is needed because `struct Pointer!(uint n ,T) {T* ptr; alias ptr this;}` is artificially turned into the LLVM IR type `{ T addrspace(n)*}` (as opposed to `{ T* }`) and when the we do `Pointer!(1 ,float) p; *p=0;` when the pointer is GEP'd, the type of the GEP'd pointer is not equal to the type of `p.ptr`.
 So ldc, tries to BitCast the GEP'd pointer across addrspace's, thus generating invalid IR, as casting across addrspaces requires as addrspace cast (but this is not desired). 

By considering pointer types equal by ignoring the addrspace we avert this problem.

This should not cause any breakage as, as per the wiki, the only way to deal in addrspaces is with inline IR, and this is (OS, arch, ... ) specific.

gen/tollvm.cpp
```diff
@@ -531,16 +531,33 @@ void DtoAlignedStore(LLValue *src, LLValue *dst) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+LLType *StripAddrSpaces(LLType *t)
```
JohanEngelen: our functions start with lowercase.

JohanEngelen:
Can you work some of the explanation you added here as comments in the code?  (so that some of it remains!)

thewilsonator:
Done

gen/tollvm.h
```diff
@@ -40,8 +40,19 @@ LLPointerType *DtoPtrToType(Type *t);
 
 LLType *voidToI8(LLType *t);
 LLType *i1ToI8(LLType *t);
+
 // Removes all addrspace qualifications. float addrspace(1)** -> float**
+// Used to avoid invalid bitcasts between addrspaces when comparing pointer
```
JohanEngelen: I think it's better to add this explanation to the callsite where you call `stripAddrSpaces` for this purpose.

thewilsonator:
@JohanEngelen Thanks, that reads much better. Is there anything else?

JohanEngelen:
I think it's fine. I don't like it so much that the address stripping function is always called, but I hope the performance impact is zero.

thewilsonator:
> but I hope the performance impact is zero.

due to the way dcompute stuff is always done after host stuff, any branch predictor worth its salt should do fine with it.

Issue #1774, opened 2016-09-19
LDC master, built with itself, cannot compile druntime/Phobos (Error: template <foo> cannot deduce function...)


ximion:
Hello!
Due to an error we currently have an LDC Git snapshot in Debian.
(see https://tracker.debian.org/pkg/ldc )
The current built copy in the archive was build using Git revision 216c112.
Unfortunately, building a newer LDC Git version fails with
```
[ 11%] Generating gccbuiltins_s390.di
cd /build/ldc-1.1.0git20160919.77fac2d/build-static/runtime && ../bin/gen_gccbuiltins /build/ldc-1.1.0git20160919.77fac2d/build-static/runtime/gccbuiltins_s390.di s390
[ 11%] Generating std/array-debug.o
cd /build/ldc-1.1.0git20160919.77fac2d && /build/ldc-1.1.0git20160919.77fac2d/build-static/bin/ldc2 --output-o -c -I/build/ldc-1.1.0git20160919.77fac2d/runtime/druntime/src -I/build/ldc-1.1.0git20160919.77fac2d/runtime/druntime/src/gc /build/ldc-1.1.0git20160919.77fac2d/runtime/phobos/std/array.d -of/build/ldc-1.1.0git20160919.77fac2d/build-static/runtime/std/array-debug.o -w -g -link-debuglib -I/build/ldc-1.1.0git20160919.77fac2d/runtime/phobos
[ 11%] Generating std/ascii-debug.o
cd /build/ldc-1.1.0git20160919.77fac2d && /build/ldc-1.1.0git20160919.77fac2d/build-static/bin/ldc2 --output-o -c -I/build/ldc-1.1.0git20160919.77fac2d/runtime/druntime/src -I/build/ldc-1.1.0git20160919.77fac2d/runtime/druntime/src/gc /build/ldc-1.1.0git20160919.77fac2d/runtime/phobos/std/ascii.d -of/build/ldc-1.1.0git20160919.77fac2d/build-static/runtime/std/ascii-debug.o -w -g -link-debuglib -I/build/ldc-1.1.0git20160919.77fac2d/runtime/phobos
[ 11%] Generating std/base64-debug.o
[...]

PR #1803, opened 2016-10-02
Fix issue #1774

kinke:
Running a cleanup with a dummy target BB and later replacing all uses of that dummy with the actual BB is buggy, as the cleanup keeps track of its exit targets (via pointers) which aren't updated.


kinke:
Adding a test should be non-trivial. :]

PR #1804, opened 2016-10-03
CodeView: arguments passed by reference not shown

rainers:
Size 0 caused references to be encoded as "near32", not "__ptr64".

Arguments passed by reference still need to be actually used in the called function to be shown, otherwise they seemed to be optimized away.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -999,7 +999,15 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
     return; // unsupported
 
   if (vd->storage_class & (STCref | STCout)) {
+#if LDC_LLVM_VER >= 308
+    auto vt = type ? type : vd->type;
+    auto T = DtoType(vt);
+    TD = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type, TD,
+                                      getTypeAllocSize(T) * 8, // size (bits)
+                                      DtoAlignment(vt) * 8);   // align (bits)
+#else
     TD = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type, TD);
+#endif
   }
 
   // get variable description
--- a/tests/debuginfo/args_cdb.d
+++ b/tests/debuginfo/args_cdb.d
@@ -1,0 +1,257 @@
+// REQUIRES: atleast_llvm309
+// REQUIRES: Windows
+// REQUIRES: cdb
+// RUN: %ldc -g -of=%t.exe %s
+// RUN: sed -e "/^\\/\\/ CDB:/!d" -e "s,// CDB:,," %s \
+// RUN:    | %cdb -snul -lines -y . %t.exe >%t.out
+// RUN: FileCheck %s -check-prefix=CHECK -check-prefix=%arch < %t.out
+
+module args_cdb;
+import core.simd;
+
+// CDB: ld /f args_cdb*
+// enable case sensitive symbol lookup
+// CDB: .symopt-1
+
+int byValue(ubyte ub, ushort us, uint ui, ulong ul,
+            float f, double d, real r,
+            cdouble c, int delegate() dg, int function() fun,
+            int[] slice, float[int] aa, ubyte[16] fa,
+            float4 f4, double4 d4,
+            Interface ifc, TypeInfo_Class ti, typeof(null) np)
+{
+// CDB: bp `args_cdb.d:25`
+// CDB: g
+    // arguments implicitly passed by reference on x64 and not shown if unused
+    float cim = c.im + fa[7] + dg() + ifc.offset;
+    return 1;
+// CHECK: !args_cdb.byValue
+// CDB: dv /t
+
+// arguments not converted to locals come first:
+// x64: cdouble * c
+// x64: int delegate() * dg
+// "Internal implementation error for fa" with cdb on x64, ok in VS
+// x64: Interface * ifc
+
+// x86: unsigned char [16] fa
+// x86: Interface ifc
+
+// locals:
+// CHECK: unsigned char ub = 0x01
+// CHECK: unsigned short us = 2
+// CHECK: unsigned int ui = 3
+// CHECK: unsigned int64 ul = 4
+// CHECK: float f = 5
+// CHECK: double d = 6
+// CHECK: double r = 7
+// x86: cdouble c =
+// x86: int delegate() dg =
+// CHECK: <function> * fun = {{0x[0-9a-f`]*}}
+// CHECK: struct int[] slice =
+// CHECK: unsigned char * aa = {{0x[0-9a-f`]*}}
+// CHECK: float [4] f4 = float [4]
+// CHECK: double [4] d4 = double [4]
+// CHECK: struct TypeInfo_Class * ti = {{0x[0-9a-f`]*}}
+// noCHECK: <CLR type> np = <unknown base type 80000013> 
+
+// check arguments with indirections
+// CDB: ?? c
+// CHECK: cdouble
+// CHECK-NEXT: re : 8
+// CHECK-NEXT: im : 9
+
+// CDB: ?? dg
+// CHECK: int delegate()
+// CHECK-NEXT: context
+// CHECK-NEXT: funcptr
+// CHECK-SAME: args_cdb.main.__lambda
+
+// CDB: ?? fa[1]
+// "Internal implementation error for fa" on x64
+// no-x86: unsigned char 0x0e (displays 0xf6)
+
+// CDB: ?? ifc
+// CHECK: Interface
+// CHECK-NEXT: classinfo
+// CHECK-NEXT: vtbl
+// x64-NEXT: offset : 0x12
+// no-x86-NEXT: offset : 0x12 (displays 0)
+
+// CDB: ?? ti
+// CHECK: TypeInfo_Class
+// CHECK-NEXT: m_init : byte[]
+}
+
+int byPtr(ubyte* ub, ushort* us, uint* ui, ulong* ul,
+          float* f, double* d, real* r,
+          cdouble* c, int delegate()* dg, int function()* fun,
+          int[]* slice, float[int]* aa, ubyte[16]* fa,
+          float4* f4, double4* d4,
+          Interface* ifc, TypeInfo_Class* ti, typeof(null)* np)
+{
+// CDB: bp `args_cdb.d:94`
+// CDB: g
+    return 3;
+// CHECK: !args_cdb.byPtr
+// CDB: dv /t
+// CDB: ?? *ub
+// CHECK: unsigned char 0x01
+// CDB: ?? *us
+// CHECK: unsigned short 2
+// CDB: ?? *ui
+// CHECK: unsigned int 3
+// CDB: ?? *ul
+// CHECK: unsigned int64 4
+// CDB: ?? *f
+// CHECK: float 5
+// CDB: ?? *d
+// CHECK: double 6
+// CDB: ?? *r
+// CHECK: double 7
+// CDB: ?? *c
+// CHECK: cdouble
+// CHECK-NEXT: re : 8
+// CHECK-NEXT: im : 9
+// CDB: ?? *dg
+// CHECK: int delegate()
+// CHECK-NEXT: context
+// CHECK-NEXT: funcptr
+// CHECK-SAME: args_cdb.main.__lambda
+// CDB: ?? *fun
+// CHECK: <function> *
+// CDB: ?? *slice
+// CHECK: struct int[]
+// CHECK-NEXT: length : 2
+// CHECK-NEXT: ptr :
+// CHECK-SAME: 0n10
+// CDB: ?? *aa
+// CHECK: unsigned char * {{0x[0-9a-f`]*}}
+// CDB: ?? (*fa)[1]
+// CHECK: unsigned char 0x0e
+// CDB: ?? (*f4)[1]
+// CHECK: float 16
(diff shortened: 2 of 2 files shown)


gen/dibuilder.cpp
```diff
@@ -999,7 +999,11 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
     return; // unsupported
 
   if (vd->storage_class & (STCref | STCout)) {
-    TD = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type, TD);
+    TD = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type, TD
+#if LDC_LLVM_VER >= 308
+                                      , 8 * (global.params.is64bit ? 8 : 4)
```
kinke: I guess it's about the payload size, as LLVM should be able to figure out the pointer size itself.

rainers: Thanks, trying that now (tests pass locally).  clang seems to pass the pointer size, though: https://github.com/llvm-mirror/clang/blob/master/lib/CodeGen/CGDebugInfo.cpp#L756. Maybe it's just the alignment that's missing...

kinke: Seems to work better now (just one issue on Win32), although clang clearly passes the pointer size - very strange.

gen/dibuilder.cpp
```diff
@@ -999,7 +999,13 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
     return; // unsupported
 
   if (vd->storage_class & (STCref | STCout)) {
-    TD = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type, TD);
+    llvm::Type *T = DtoType(type ? type : vd->type);
+    TD = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type, TD
+#if LDC_LLVM_VER >= 308
+                                      , getTypeAllocSize(T) * 8 // size (bits)
+                                      , getABITypeAlign(T) * 8  // align (bits)
```
kinke: I think `DtoAlignment(type ? type : vd->type)` would make a better choice here, since the LL types don't feature any explicit alignment IIRC, so e.g. an explicit `align(N)` is only available via the D type.

tests/debuginfo/args_cdb.d
```diff
@@ -0,0 +1,231 @@
...
+// CHECK: struct TypeInfo * ti = {{0x[0-9a-f`]*}}
+// noCHECK: <CLR type> np = <unknown base type 80000013> 
+
+// arguments passed by reference are listed out of order
+// CDB: ?? dg
+// x86: int delegate()
+// x86-NEXT: context
+// x86-NEXT: funcptr
+// x86-SAME: args_cdb.main.__lambda
+// x64: int delegate() *
```
kinke: So that's still a bug for Win64, apparently due to `llvm::DIBuilder::createParameterVariable()` being used instead of `createAutoVariable()` - `rewrittenToLocal` is always false for `ExplicitByvalRewrite`, as the actual argument (pointer) is the parameter storage. I've just verified that structs > 64 bits passed by value are also represented by DI pointers (on Win64). `createParameterVariable()` seems to make use of the LL arg type somehow (at least checking whether it's a pointer or not) and not just use the specified `llvm::DIType`, leading to this wrong DI indirection.

Can't we just use `createAutoVariable()` for all parameters?

rainers: cdb shows references as pointers, but VS does not. These arguments implicitly passed by reference are even shown as values (without & on the type).


> Can't we just use createAutoVariable() for all parameters?

I think this will only work if we make an explicit copy, otherwise the debugger might not know about the implicit conversion to  reference.

tests/debuginfo/args_cdb.d
```diff
@@ -0,0 +1,231 @@
...
+            float f, double d, real r,
+            cdouble c, int delegate() dg, int function() fun,
+            int[] slice, float[int] aa, ubyte[16] fa,
+            float4 f4, double4 d4,
+            Interface ifc, TypeInfo ti, typeof(null) np)
+{
+// CDB: bp `args_cdb.d:25`
+// CDB: g
+    dg(); // delegate implicitly passed by reference on x64 and not shown if unused
+    float cim = c.im; // implicitly passed by reference on x64 and not shown if unused
```
rainers: While "c" can be evaluated in line 25, it is shown as "optimized away" in this line 26. This seems to be similar to the problem with gdb10311.d in the testsuite.

kinke: A while back, the IR was pretty ugly when it came to DI lifetime scopes. I imagine it's responsible for this.

rainers:
> Is that what you mean by 'implicitly passing by reference'? 

yes.

> I refer to it as 'explicit byval passing', emphasizing its (special) byval semantics.

I think this naming is confusing and has always made it hard to understand the ABI code for me. All arguments to byValue are passed explicitly by value. Some of these are implicitly converted to references, though.

> I don't wanna see struct Small small and struct Big * big for this:

This happens in cdb, but not in VS:

bigsmall

kinke:
> Some of these are implicitly converted to references.

I like that description. But I wouldn't like any 'reference' inside the name for the appropriate ABIRewrite, since I think it's important to stress that the callee operates with its private copy, not how it gets to that copy's lvalue (for all other ABIs, an implicit pointer computed by LLVM when using the `byval` keyword, here the explicit pointer provided by the caller).

> This happens in cdb, but not in VS

Yep I understood that in your previous post, but I still find it ugly having to adjust the 32/64 bit CDB tests due to ABI details. Can someone come up with any debugging advantage when decorating variables as parameters instead of locals? All I've seen so far, except for these pointer issues here, is an argument index displayed by gdb, but I don't see a real value in that index.

kinke:
`IndirectByvalRewrite` maybe, emphasizing the additional indirection?

rainers:
> IndirectByvalRewrite maybe, emphasizing the additional indirection?

Yes, better. I also considered "indirection" instead of "reference" to avoid confusion with explicit references.

> Can someone come up with any debugging advantage when decorating variables as parameters instead of locals?

It avoids the additional copy to the local stack (if you declare it as a reference, it will not be shown as passed by value). I'm not sure it will be optimized away if the argument is a reference.
Parameter can also be inspected before the function prolog.

rainers:
> IndirectByvalRewrite maybe, emphasizing the additional indirection?

We should be careful not to introduce meanings for "byVal" different from what the LLVM attribute means. I haven't fully understood why the latter doesn't work for the Win64 ABI.

rainers:
I amended a few tweaks to the test case:
- fixed size arrays and structs are also checked now aswell as class members
- I could simplify some x86/x64 checks by noting that displaying a pointer to a struct also lists it members
- structs and fixed size arrays passed by value display garbage on x86

kinke:
> We should be careful not to introduce meanings for "byVal" different from what the LLVM attribute means. I haven't fully understood why the latter doesn't work for the Win64 ABI.

Okay then let me explain. `void foo(Big big)`. If `Big` isn't passed in registers, it's to be passed in memory, by value. When using the `byval` keyword, the caller needs to pass a pointer to the original argument as LL arg (`Big* arg = &big`). LLVM then takes care of placing a bitcopy into the function parameters stack of the callee. The callee then accesses this copy via an implicit pointer (its LL parameter `Big* param = <pointer into parameters stack>`).
By-value semantics for aggregates > 64 bit on Win64 are different; nothing bigger than 64 bit is to be passed in the function parameters stack. The caller needs to allocate and memcpy the copy for the callee and then passes the pointer as actual argument. LDC violates this only wrt. dynamic arrays as you already noted.
But both methods - `byval` and IndirectByvalRewrite - share the same intention - passing a bitcopy (in memory) of the original D argument to the callee.

> It avoids the additional copy to the local stack (if you declare it as a reference, it will not be shown as passed by value). I'm not sure it will be optimized away if the argument is a reference.

What additional copy?! I haven't checked IR/asm lately, but I would have never thought that debug infos would imply additional copies.

rainers:
> Okay then let me explain.

Thanks for explanation. So the difference between LLVM byval and ExplicitByval is that the copy is placed in the "function parameter stack" by LLVM for the former, and in "local stack of the caller" for the latter? Is there any practical difference? Why does LLVM byval work for dynamic arrays?

Finally, I got the idea behind LLVM's naming "byval" now: a pointer is passed as the argument, but annotated to actually make a copy for value semantics.

> What additional copy?! 

The copy by the explicit rewrite:
```
define void @_D8bigsmall3fooFS8bigsmall5SmallS8bigsmall3BigZv(%bigsmall.Big* noalias nocapture align 16 %big, %bigsmall.Small %small_arg) #0 comdat !dbg !11 {
  %small = alloca %bigsmall.Small, align 4        ; [#uses = 2, size/byte = 4]
  store %bigsmall.Small %small_arg, %bigsmall.Small* %small, !dbg !27 ; [debug line = 4:6]
  call void @llvm.dbg.declare(metadata %bigsmall.Small* %small, metadata !15, metadata !28), !dbg !27 ; [debug line = 4:6] [debug variable = small]
  call void @llvm.dbg.declare(metadata %bigsmall.Big* %big, metadata !20, metadata !28), !dbg !27 ; [debug line = 4:6] [debug variable = big] 
```
The DI for variable "small" is declared on the copy, while for "big" it is the actual argument. If you declare "Big&" on the parameter, it will show in the debugger as an explicit reference. If you don't want that you have to make a copy similar to "small" to be able to declare the type of the "big" as "Big".

rainers:
BTW: this PR is about explicit references, the discussed implicit references just happen to show in the tests ;-)

kinke:
> So the difference between LLVM byval and ExplicitByval is that the copy is placed in the "function parameter stack" by LLVM for the former, and in "local stack of the caller" for the latter? Is there any practical difference?

Exactly. The difference is that in the first case, the pointer is implicit (based on arg index and previous args), while the pointer for Win64 needs to be explicit, as the copy can be allocated anywhere in the caller's stack. Additionally, Win64 requires the copy to be aligned at 16 bytes.
I don't know the primary motivations for Microsoft's choice here; it definitely simplifies the ABI though (compare that to System V, the ABI for all other x86_64 OS).

> Why does LLVM byval work for dynamic arrays?

We don't use byval for Win64 at all IIRC (but for most other ABIs incl. Win32). As dynamic arrays cannot be rewritten at the moment (that's the only reason for this Win64 ABI violation), we pass the LL struct (pair) directly as LL arg. From what I recall, LLVM then passes each member separately, i.e., as 2 separate 64-bit args (in registers if still available, otherwise resorting to the function parameters stack; potentially even the length in a register and the pointer in the function parameters stack :)).

> The copy by the explicit rewrite

That is no additional copy though. It's the **single** copy for the callee. There's generally 3 ways for a callee to get to its by-value parameter lvalue (always needs to be an lvalue, as it is a local for the callee, i.e., it may be modified):

1. If the original arg is passed without rewrite (most primitive types, but also dynamic arrays on Win64), the value (in most cases a register) is dumped to the callee stack, and that's the local then.
2. If the `byval` attribute is used (only valid for non-rewritten args), we use the implicit pointer (the LL param) as lvalue.
3. Otherwise, the ABIRewrite is responsible to provide the lvalue. As ExplicitByvalRewrite performs the copy on the caller's side, it simply returns the pointer to it (which happens to be the LL param).

That parameter lvalue is the final associated `IrParameter::value`, which we use for all parameter DI, independent of using `createParameterVariable()` or `createAutoVariable()`, and this is always correct. It's just that `createParameterVariable()` does something else too, leading to this DI indirection. I tried using `createAutoVariable()` and it works fine with cdb, displaying `struct Big big`.

> BTW: this PR is about explicit references

Yeah I know. Just general talk for more upcoming DI fixes. ;)

kinke:
Hmm, what if the intention behind this `create{Parameter,Auto}Variable()` separation is that the raw original params can be displayed (with the idea of passing the LL params directly, i.e., constant rvalues) separately from autos/locals (for which LLVM expects an lvalue)?

kinke:
Scratch that, apparently the only difference is really just the argument index annotation.

rainers:
Thanks for clarification. I hope I've digested that now.

I checked what ldc, dmd, cl and clang do for the bigsmall example for Win64:

- ldc: "small" shown, but wrong before storing the register to the stack, "big" available immediately because the DI contains the register. "big" declared as parameter, "small" as local.
- clang: "small" only shown correctly after storing the register to the stack, "big" always optimized away. Both declared as parameters.
- cl: before storing the parameter registers to the stack, "small" and "big" are shown, but wrong. After storing both are ok. No variables seem to be declared as "parameters".
- dmd: before storing the parameter registers to the stack, "small" and "big" are shown, but wrong. After storing both are ok. Both declared as "parameters".

Every backend uses different records, but only dmd and cl show parameters in the callstack.

kinke:
Thanks for checking, interesting comparison. We'll also need to check at least gdb too. With your recent `rewrittenToLocal = (localLval != llParam)` change, only `byval` args and some rewritten args (incl. ExplicitByvalRewrite) will be emitted as 'parameters'. I.e., a simple `int` param will show up as 'local'.

If MSVC itself doesn't declare any 'parameters', we could simply only declare 'locals' for MSVC targets. Thus working around the issues we're seeing (dynamic arrays; additional indirection for ExplicitByvalRewrite args with cdb/windbg), which appear to be caused either by LLVM's codeview library or the used debugger.

kinke:
gdb 7.11 doesn't require the 'locals' workaround, 'parameters' are fine (verified for a single int, a 2x32-bit struct rewritten to an `i64` and a 4x64-bit `byval` struct on Linux x64). I'll prepare a PR for this.

Issue #87, opened 2012-03-06
Debug info for implicit try/finally blocks


bioinfornatics:
dear,

Each time when i use std.stdio.File gdb print a message and i can't show the line.

```

25	    File f1 = File(bedOutFilePath, "w");
(gdb) n
26	/home/jonathan/Projets/dscience/examples/parser/.: No such file or directory.
	in /home/jonathan/Projets/dscience/examples/parser/.
(gdb) 
27	in /home/jonathan/Projets/dscience/examples/parser/.
(gdb) 
28	in /home/jonathan/Projets/dscience/examples/parser/.
```

they are no problem about io on the file. you can try:

```d
import std.stdio;
import std.string;

void main(){
    File f = File("myFile", "w");
    f.writeln( "something" );
    f.close();
}
```

build this:

```
$ ldc2 -g test.d
```

and debug:

```
$ gdb -q ./test
> r
> b 5
> n
6	/home/jonathan/.: No such file or directory.
	in /home/jonathan/.
```

kinke:
Still an issue.

kinke:
Fixed by https://github.com/ldc-developers/ldc/pull/1808.

PR #1808, opened 2016-10-03
Fix issue #87: do not use "." for missing debug line info

rainers:
instead try harder to find a suitable filename and line location.

In case of implicitly generated functions without any information, it now uses the line of the declaration.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -131,7 +131,10 @@ void ldc::DIBuilder::Declare(const Loc &loc, llvm::Value *var,
 }
 
 ldc::DIFile ldc::DIBuilder::CreateFile(Loc &loc) {
-  llvm::SmallString<128> path(loc.filename ? loc.filename : "");
+  const char* filename = loc.filename;
+  if (!filename)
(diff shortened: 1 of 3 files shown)


gen/dibuilder.cpp
```diff
@@ -930,23 +940,31 @@ void ldc::DIBuilder::EmitStopPoint(Loc &loc) {
...
-  if (!loc.linnum && IR->ir->getCurrentDebugLocation())
-    return;
+  if (!loc.linnum) {
+    if (IR->ir->getCurrentDebugLocation())
+      return;
+    // use the line of the enclosing symbol that has line number debug info
+    for (Dsymbol* sym = IR->func()->decl; sym && !linnum; sym = sym->parent)
+      linnum = sym->loc.linnum;
+    if (!linnum)
+      linnum = 1;
```
kinke: Thanks, looks good to me (particularly the test cases!). I'd just move this new fallback-linnum code outside the `#if` (note: no outer `if (!loc.linnum)` needed for the 5 lines above) to let LLVM < 3.7 benefit too. Feel free to merge afterwards.

PR #1809, opened 2016-10-04
Produce correct code for increment/decrement operations on complex types

LemonBoy:
As you can see I got tangled up while trying to get trough the forest of casts :)
I'm not very fond of the `toElem` call, nor to the fact I'm using `LLConstantFP` directly instead of `DtoConstFP` but, as stated before, I got lost while trying to figure the how to jump forth and back between the llvm and the frontend types.

Cheers :)

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -1364,7 +1364,8 @@ public:
     auto &PGO = gIR->funcGen().pgo;
     PGO.setCurrentStmt(e);
 
-    LLValue *const lval = DtoLVal(e->e1);
+    DValue *const dv = toElem(e->e1);
+    LLValue *const lval = DtoLVal(dv);
     toElem(e->e2);
 
     LLValue *val = DtoLoad(lval);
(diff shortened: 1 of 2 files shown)


kinke:
Thanks, looks good after a superficial first glance. :)

gen/toir.cpp
```diff
@@ -1387,6 +1387,17 @@ class ToElemVisitor : public Visitor {
       LLConstant *offset =
           e->op == TOKplusplus ? DtoConstUint(1) : DtoConstInt(-1);
       post = DtoGEP1(val, offset, false, "", p->scopebb());
+    } else if (e1type->iscomplex()) {
+      assert(e2type->iscomplex());
+      LLValue *re, *im, *one = LLConstantFP::get(DtoComplexBaseType(e1type), 1.0);
+      DValue *kz = toElem(e->e1);
```
JohanEngelen: I think it'd be good to move the `toElem` up to the `DtoLVal(e->e1)` call on line 1367. We had evaluation order problems in the past, so it's nice if things are grouped to make it clearer how evaluation happens.

gen/toir.cpp
```diff
@@ -1399,7 +1410,8 @@ class ToElemVisitor : public Visitor {
       llvm_unreachable("Unsupported type for PostExp.");
     }
 
-    DtoStore(post, lval);
+    if (post != nullptr)
+	DtoStore(post, lval);
```
JohanEngelen: An alternative is to do the `if (e1type->iscomplex()` check first and put this `post` business inside the else branch for all other types.
(you could also add a little comment to the definition of `post` to say what it is used for.)

JohanEngelen:
Thanks a lot for the fix.
A test is still missing.

gen/toir.cpp
kinke: Definitely true, otherwise `e1` is evaluated twice, something we definitely don't want. I.e., replace `LLValue *lval = DtoLVal(e->e1);` by `DValue *dv = toElem(e->e1); LLValue *lval = DtoLVal(dv);`.

LemonBoy:
Thanks for the review! I've adjusted the code and added a test :)

tests/codegen/complex_postexpr_gh1806.d
```diff
@@ -0,0 +1,14 @@
+// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
```
kinke: This is basically a no-op (well, it checks that it can be compiled).
There's no need to check the IR here, just running the code should do the trick - i.e., just use `// RUN: %ldc -run %s`. Including `creal` would be nice too, just to have a complete set.

kinke:
Btw, I guess this doesn't fix pre-in/decrements (`++x`). Not mandatory, but if you have some time, fixing that too would be nice. :)

LemonBoy:
Interestingly enough the prefix version gets desugared as `<var> {+=,-=} <unit>`, so nothing to adjust :)

kinke:
Excellent, thanks again!

PR #1810, opened 2016-10-06
CircleCI: build LDC with itself and test that version.

JohanEngelen:
This change makes sure that LDC can build itself. To protect against issues like: https://github.com/ldc-developers/ldc/issues/1774

The extra CircleCI test time is acceptable for me.
Edit: looks like it will add 7 minutes to the CircleCI build.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,5 +1,9 @@
 # Requires the BETA setting to use Ubunty 14.04 Trusty
 
+# It appears that the current directory is reset at the start of each `-` line,
+# so for example to do an out-of-source build, you have to chain operations
+# after `cd`: `cd build && cmake .. `.
+
 machine:
   environment:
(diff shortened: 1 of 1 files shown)


JohanEngelen:
The first build is done out-of-source, but I could not get `cd` to work without putting everything on one line (it looks like the current dir is reset at the start of each `-` line).

kinke:
Fine by me. 7 minutes is okay as the bottle-neck is still clearly AppVeyor of course. ;)

> it looks like the current dir is reset at the start of each `-` line

Might be worth noting at the beginning of the file, I'm sure it's going to bug us again.

JohanEngelen:
> Might be worth noting at the beginning of the file, I'm sure it's going to bug us again.

Done.

PR #1811, opened 2016-10-06
Apply TargetMachine options as function attributes in IR.

JohanEngelen:
This change is needed for LTO, as the linker/LTO-codegen otherwise would use the default TargetOptions.

~~Work-in-progress: will add tests when green.~~

Note: Clang does exactly the same.

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -18,6 +18,7 @@
 #include "mtype.h"
 #include "statement.h"
 #include "template.h"
+#include "driver/cl_options.h"
 #include "gen/abi.h"
 #include "gen/arrays.h"
 #include "gen/classes.h"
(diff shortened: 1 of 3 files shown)


JohanEngelen:
related LLVM-dev discussion: http://lists.llvm.org/pipermail/llvm-dev/2016-October/105663.html

JohanEngelen:
This is also needed for correct handling of .bc files on the cmdline.

JohanEngelen:
merging tomorrow if no comments.

JohanEngelen:
Agreed on the oddness, I looked for it but couldn't find it. Clang does this explicitly too, like I did.
There's some talk about these TargetOptions as attributes on LLVM mailing list / bug tracker related to LTO. Perhaps things will improve in the future, it's a little brittle for now.

PR #1812, opened 2016-10-07
Rename "ir2obj-cache" to "cache"

JohanEngelen:
Renames the (yet unreleased) commandline options `-ir2obj-cache*` to `-cache*`.
I am working on adding more caching to LDC, after which the cache is no longer just for IR->obj. Also, "ir2obj" is an implementation detail that the user shouldn't be bothered with.


JohanEngelen:
Hm, AppVeyor hit a std.file error, restarted the build

JohanEngelen:
Rebased to resolve merge conflict

kinke:
Fine by me if you already have ideas for further compilation caches and you think the renaming is useful.

PR #1816, opened 2016-10-08
Adapt emission of debuginfos for parameters

kinke:
Non-MSVC targets: emit all parameters (except for captured ones) as DI parameters.

MSVC targets: emit all parameters as DI locals to work around misc. issues discussed in PR #1804.

### DIFF
--- a/tests/debuginfo/args_cdb.d
+++ b/tests/debuginfo/args_cdb.d
@@ -28,16 +28,6 @@ int byValue(ubyte ub, ushort us, uint ui, ulong ul,
 // CHECK: !args_cdb.byValue
 // CDB: dv /t
 
-// arguments not converted to locals come first:
-// x64: cdouble * c
-// x64: int delegate() * dg
-// "Internal implementation error for fa" with cdb on x64, ok in VS
-// x64: Interface * ifc
-
-// x86: unsigned char [16] fa
-// x86: Interface ifc
-
-// locals:
 // CHECK: unsigned char ub = 0x01
 // CHECK: unsigned short us = 2
 // CHECK: unsigned int ui = 3
@@ -45,15 +35,22 @@ int byValue(ubyte ub, ushort us, uint ui, ulong ul,
 // CHECK: float f = 5
 // CHECK: double d = 6
 // CHECK: double r = 7
+// x64: cdouble * c =
 // x86: cdouble c =
+// x64: int delegate() * dg =
 // x86: int delegate() dg =
 // CHECK: <function> * fun = {{0x[0-9a-f`]*}}
-// CHECK: struct int[] slice =
+// x64: struct int[] * slice =
+// x86: struct int[] slice =
 // CHECK: unsigned char * aa = {{0x[0-9a-f`]*}}
+// "Internal implementation error for fa" with cdb on x64, ok in VS
+// x86: unsigned char [16] fa
 // CHECK: float [4] f4 = float [4]
 // CHECK: double [4] d4 = double [4]
+// x64: Interface * ifc
+// x86: Interface ifc
 // CHECK: struct TypeInfo_Class * ti = {{0x[0-9a-f`]*}}
-// noCHECK: <CLR type> np = <unknown base type 80000013> 
+// CHECK: void * np = {{0x[0`]*}}
 
 // check arguments with indirections
 // CDB: ?? c
@@ -68,7 +65,7 @@ int byValue(ubyte ub, ushort us, uint ui, ulong ul,
 // CHECK-SAME: args_cdb.main.__lambda
 
 // CDB: ?? fa[1]
-// "Internal implementation error for fa" on x64
+// "Internal implementation error for fa" with cdb on x64, ok in VS
 // no-x86: unsigned char 0x0e (displays 0xf6)
 
 // CDB: ?? ifc
@@ -90,7 +87,7 @@ int byPtr(ubyte* ub, ushort* us, uint* ui, ulong* ul,
           float4* f4, double4* d4,
           Interface* ifc, TypeInfo_Class* ti, typeof(null)* np)
 {
-// CDB: bp `args_cdb.d:94`
+// CDB: bp `args_cdb.d:91`
 // CDB: g
     return 3;
 // CHECK: !args_cdb.byPtr
@@ -141,7 +138,7 @@ int byPtr(ubyte* ub, ushort* us, uint* ui, ulong* ul,
 // CHECK-NEXT: m_init : byte[]
 // shows bad member values
 // CDB: ?? *np
-// noCHECK: <CLR type> np = <unknown base type 80000013> 
+// CHECK: void * {{0x[0`]*}}
 }
 
 int byRef(ref ubyte ub, ref ushort us, ref uint ui, ref ulong ul,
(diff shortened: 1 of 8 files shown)


tests/debuginfo/args_cdb.d
```diff
@@ -45,38 +36,39 @@ int byValue(ubyte ub, ushort us, uint ui, ulong ul,
...
-
-// check arguments with indirections
+// CHECK: void * np = {{0x[0`]*}}
+// CHECK: cdouble c =
+// CHECK: int delegate() dg =
+// x64: 'Internal implementation error for fa'
+// no-CHECK: unsigned char [16] fa
+// CHECK: Interface ifc
+
+// check arguments with indirections (their data is garbage)
```
kinke: In VS on Win64, `c`, `dg`, and `fa` show up as optimized away. `ifc` isn't, but its data is garbage. In the calling `main()` frame, they are all okay, where they are also declared as DI locals. So I guess this shows that the LLVM CodeView library checks the type of the associated LL value and does things differently depending on whether it's an alloca or something else (such as a LL function parameter).

rainers:
> In VS on Win64, c, dg, and fa show up as optimized away.

That happens in clang, too.

These parameters only have debug info for assigned registers with limited lifetime:
```
(00026F)  S_LOCAL: Local: 0000103D, c
(00027B)  S_DEFRANGE_REGISTER: rbx
	Range: [0000:00000082] - [0000:00000197], 0 Gaps
(00028B)  S_LOCAL: Local: 00001041, dg
(000298)  S_DEFRANGE_REGISTER: r14
	Range: [0000:00000089] - [0000:0000019F], 0 Gaps
(0002A8)  S_LOCAL: Local: 00001043, fa
(0002B5)  S_DEFRANGE_REGISTER: r13
	Range: [0000:0000009E] - [0000:0000019D], 0 Gaps
(0002C5)  S_LOCAL: Local: 00001035, ifc
(0002D3)  S_DEFRANGE_REGISTER: r8
	Range: [0000:0000009E] - [0000:000000A3], 0 Gaps
(0002E3)  S_DEFRANGE_REGISTER_REL: [rsp +  0040 ]
	Range: [0000:000000A3] - [0000:00000148], 0 Gaps
```
The breakpoint location in the cdb output (byValue+0x122) suggests that it is inside the range of these variables, though.

kinke:
Which LLVM version are you using, Rainer? I have 3.9. And how did you dump those DI details?

kinke:
Btw, on Win64 only the first 4 args are passed in registers. Due to the args order being reversed for `extern(D)`, out of these 4 troublesome args, only the pointer to the callee's `ifc` copy is passed in a register (3rd actual arg => R8). The other 3 pointers are passed in the function parameters stack, i.e., in the callee's frame.

kinke:
That register binding generally seems wrong. We surely don't want the pointer to be read from the register whenever accessing the DI, as the register might be quickly overwritten. Conceptually, we want the debugger to remember the initial pointer value and dereference that one when accessing the DI variable. We may have to add corresponding DWARF expressions (explicit dereferences or so) when declaring DI variables with LL values which aren't allocas, at least for MSVC targets.

rainers:
> Which LLVM version are you using, Rainer? I have 3.9. 

I am currently on 4.0 in the hope to not duplicate some bug fixes.

> And how did you dump those DI details?

I'm mostly using cvdump.exe from https://github.com/Microsoft/microsoft-pdb/tree/master/cvdump, but `llvm-readobj -codeview` can dump most of this, too (but it is very verbose).

rainers:
> So RBX, R14 and R13 shown by your dump should be utterly wrong.

If I check the memory at these addresses, they seem pretty ok. These registers have been loaded in the function prolog, though. So it seems the register records are ob, but confuse VS.

rainers:
> That register binding generally seems wrong. We surely don't want the pointer to be read from the register whenever accessing the DI, as the register might be quickly overwritten.

Please note that there is a lifetime associated with the register records. There can be a list of ranges similar to DWARF, e.g. `ifc` is addressed by a register first, but written to the stack later. This is what's supposed to be written in optimized mode by cl, too.

rainers:
> We may have to add corresponding DWARF expressions

Unfortunately not yet supported by the CodeView backend: https://github.com/llvm-mirror/llvm/blob/master/lib/CodeGen/AsmPrinter/CodeViewDebug.cpp#L947. That's why closure variables cannot be supported at the moment (also broken with gdb anyway).

rainers:
> So it seems the register records are ok, but confuse VS.

The current master version emits the same register records, but variables are marked as "Param":
```
(00007E)  S_LOCAL: Param: 00001005, c
(00008A)  S_DEFRANGE_REGISTER: rbx
	Range: [0000:00000000] - [0000:00000197], 0 Gaps
(00009A)  S_LOCAL: Param: 0000100B, dg
(0000A7)  S_DEFRANGE_REGISTER: r14
	Range: [0000:00000000] - [0000:0000019F], 0 Gaps
(0000B7)  S_LOCAL: Param: 0000100D, fa
(0000C4)  S_DEFRANGE_REGISTER: r13
	Range: [0000:00000000] - [0000:0000019D], 0 Gaps
(0000D4)  S_LOCAL: Param: 00001013, ifc
(0000E2)  S_DEFRANGE_REGISTER: r8
	Range: [0000:00000000] - [0000:000000A3], 0 Gaps
(0000F2)  S_DEFRANGE_REGISTER_REL: [rsp +  0040 ]
	Range: [0000:000000A3] - [0000:00000148], 0 Gaps
```
VS displays the variables correctly.

rainers:
BTW: I just checked the bigsmall example in C++: cdb also shows "big" as `struct Big * big = 0x000000d4acf8fda0`, but "small" as `struct Small small`.

I think, we are ok on master, we should not insist on beating C++ support.

kinke:
You're right, switching back to DI parameters for all (non-captured) parameters seems to fix things in most cases (especially for VS, which is looking pretty cool already!). So DI locals seem to require LL allocas whereas DI parameters seem a bit more flexible in this regard.

So we're back to that additional DI indirection for cdb on Win64, for dynamic arrays and args rewritten by ExplicitByvalRewrite, requiring some special-casing for the cdb tests.
More importantly, non-MSVC targets emit more DI parameters again too, not just args passed via `byval` (and maybe some rewritten args) as in master, completing the full backtrace with gdb etc.

rainers:
> You're right, switching back to DI parameters for all (non-captured) parameters seems to fix things in most cases (especially for VS, which is looking pretty cool already!).

Unfortunately there are no tests for indirections involving d4, f4, slice[]. These show bad values in VS now, I guess in cdb, too. There must have been a reason why I used the rewrittenToLocal parameter...

kinke:
Yeah but the previous `rewrittenToLocal = (llLval != rawLLParam)` doesn't work properly for all other ABIs. So I guess the problem with dynamic arrays and these LL vector types are most likely ABI violations on our part (MS states that `__m128` types are to be passed in memory too on Win64) in combination with incomplete/nonfinal CodeView support in LLVM.

kinke:
The SIMD types should be rather exotic, so I wouldn't focus on those. Dynamic arrays on the other hand are pretty ubiquitous. Declaring those as DI locals for Win64 may be a feasible workaround, but it's really ugly.

kinke:
Updated.

rainers:
Thanks, looks good now. Maybe we can tackle the dynamic arrays again when clang has caught up and forced some changes to LLVM.

rainers:
Hmmm, did merging this cause the build failure on master (travis OS/X)?
```
720: /bin/bash: /bin/bash: cannot execute binary file
720: make[2]: *** [/Users/travis/build/ldc-developers/ldc/dmd-testsuite-debug/fail_compilation/fail272.d.out] Error 126
```
Or is this a known spurious failure?

kinke:
> Or is this a known spurious failure?

I've retriggered the OSX job and it's working now. I've never seen that error though, so I'd say it was an **un**known spurious failure. OSX seems to suffer the most from these (mostly these 10 min timeouts).

PR #1817, opened 2016-10-09
Travis: use LLVM3.9 on OS X

JohanEngelen:
Let's see if this works (so that we can release with LLVM3.9 on OS X).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,10 @@ matrix:
       env: LLVM_VERSION=3.5.2 OPTS="-DTEST_COVERAGE=ON"
     - os: osx
       d: ldc
-      env: LLVM_CONFIG="llvm-config-3.8"
+      env: LLVM_VERSION=3.9.0
     - os: osx
       d: ldc
-      env: LLVM_CONFIG="llvm-config-3.8" OPTS="-DBUILD_SHARED_LIBS=ON"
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Homebrew does not have llvm3.9 yet.

See:
https://github.com/Homebrew/homebrew-core/pull/4579
https://github.com/Homebrew/homebrew-versions/pull/1384

JohanEngelen:
Instead of waiting for homebrew, I rewrote travis.yml to download the binaries from LLVM directly and cache them like we do for the Linux builds.  This appears to speed up the build a lot (~39min -> ~30min), possibly because LLVM's binaries have all assertions disabled?

JohanEngelen:
(green for mac with LLVM3.9 btw, nice :-))

kinke:
> possibly because LLVM's binaries have all assertions disabled

Seems like it, as we don't have assertions enabled on Linux as well. I must say I really prefer CI builds with enabled assertions; people keep complaining about LDC crashing, and most of these issues (and others!) would be caught by enabled assertions. I would even prefer alphas and betas to ship with assertions too.
Does CircleCI have assertions enabled or is it just AppVeyor now?

JohanEngelen:
Pretty sure that CircleCI also does not have assertions enabled...

Hm.

JohanEngelen:
Already merged in release-1.1.0 branch.

PR #1818, opened 2016-10-09
fix build against LLVM master: clEnumValEnd removed

rainers:
Not sure if has landed in circleCI already, but probably will soon.

### DIFF
--- a/gen/cl_helpers.h
+++ b/gen/cl_helpers.h
@@ -206,4 +206,16 @@ public:
 };
 }
 
+
+#if LDC_LLVM_VER >= 400
+#define clEnumValues llvm::cl::values
+#else
+template <typename DataType, typename... OptsTy>
+llvm::cl::ValuesClass<DataType> clEnumValues(const char *Arg, DataType Val,
+                                             const char *Desc,
+                                             OptsTy... Options) {
+  return llvm::cl::values(Arg, Val, Desc, Options..., clEnumValEnd);
+}
+#endif
+
 #endif
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -8,6 +8,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "gen/llvmhelpers.h"
+#include "gen/cl_helpers.h"
 #include "declaration.h"
 #include "expression.h"
 #include "gen/abi.h"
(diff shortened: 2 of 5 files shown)


gen/llvmhelpers.cpp
```diff
@@ -44,19 +45,24 @@
 
 #include "llvm/Support/CommandLine.h"
 
+#if LDC_LLVM_VER >= 400
+#define _clEnumValN llvm::clEnumValN
+#else
+#define _clEnumValN clEnumValN
+#endif
```
kinke: It has landed. :) Is this `_clEnumValN` really required? Would
```C++
#if LDC_LLVM_VER >= 400
using llvm::clEnumValN;
#endif
```
not work?

kinke:
Oh sry seems like I can't simply remove my test commit anymore. :(

kinke:
`llvm-profdata-4.0.cpp` needs to be synced as well.

rainers:
Updated.
I wonder what I have to do (or stop doing) to also build the profile and cache stuff on windows. Does it need a special build of LLVM? I only see the tests continue to fail...

kinke:
> I wonder what I have to do (or stop doing) to also build the profile and cache stuff on windows.

The `all` target includes those tools (at least with ninja).

gen/cl_helpers.h
```diff
@@ -206,4 +206,16 @@ class StringsAdapter {
 };
 }
 
+
+#if LDC_LLVM_VER >= 400
+#define clEnumValues llvm::cl::values
```
JohanEngelen: would `using clEnumValues = llvm::cl::values;` work?

gen/llvmhelpers.cpp
JohanEngelen: Could you rename `_clEnumValN` to ALL_CAPS ?

gen/cl_helpers.h
rainers: No, the compiler complains that using only works for types.

rainers:
> The all target includes those tools (at least with ninja).

Thanks, it helped me discover the outdated list of targets that my build script used. (With MULTILIB, I have to build the 32-bit libs with different environment settings.)

JohanEngelen:
@rainers I think you should just do whatever you think is best, and then merge this thing asap :-)

gen/llvmhelpers.cpp
```diff
@@ -44,19 +45,23 @@
 
 #include "llvm/Support/CommandLine.h"
 
+#if LDC_LLVM_VER >= 400
+// trick clEnumValN() into prepending the llvm namesspace 
+namespace cl { using OptionEnumValue = llvm::cl::OptionEnumValue; }
```
rainers: This workaround can be removed with the next update to LLVM: https://github.com/llvm-mirror/llvm/commit/15a3555c6d39d2b5926b306e4d24bce0cf4d390b

Issue #506, opened 2013-10-12
Array literal promotion logic is overly pessemistic


klickverbot:
Immutable array literals can be promoted to global constants if they are actually constants. `isConstLiteral`, the function that determines this, is currently overly conservative (see #503).

kinke:
Taken care of by https://github.com/ldc-developers/ldc/pull/1821.

PR #1821, opened 2016-10-10
Check StructLiteralExp elements in isConstLiteral.

LemonBoy:
Quick and dirty attempt at #506 :)

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -496,16 +496,37 @@ Expression *indexArrayLiteral(ArrayLiteralExp *ale, unsigned idx) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-bool isConstLiteral(ArrayLiteralExp *ale) {
-  // FIXME: This is overly pessemistic, isConst() always returns 0 e.g. for
-  // StructLiteralExps. Thus, we waste optimization potential (GitHub #506).
-  for (size_t i = 0; i < ale->elements->dim; ++i) {
-    // We have to check specifically for '1', as SymOffExp is classified as
-    // '2' and the address of a local variable is not an LLVM constant.
-    if (indexArrayLiteral(ale, i)->isConst() != 1) {
-      return false;
+bool isConstLiteral(Expression *e) {
+  // We have to check the return value of isConst specifically for '1',
+  // as SymOffExp is classified as '2' and the address of a local variable is
+  // not an LLVM constant.
+  //
+  // Examine the ArrayLiteralExps and the StructLiteralExps element by element
+  // as isConst always returns 0 on those.
+  switch (e->op) {
+  case TOKarrayliteral: {
+    auto ale = static_cast<ArrayLiteralExp *>(e);
+    for (auto el : *ale->elements) {
+      if (!isConstLiteral(el ? el : ale->basis))
+        return false;
     }
+  } break;
+
+  case TOKstructliteral: {
+    auto sle = static_cast<StructLiteralExp *>(e);
+    if (sle->sd->isNested())
+      return false;
+    for (auto el : *sle->elements) {
+      if (el && !isConstLiteral(el))
+        return false;
+    }
+  } break;
+
+  default:
+    if (e->isConst() != 1)
+      return false;
   }
+
   return true;
 }
 
--- a/gen/arrays.h
+++ b/gen/arrays.h
@@ -42,7 +42,7 @@ LLConstant *DtoConstSlice(LLConstant *dim, LLConstant *ptr,
 Expression *indexArrayLiteral(ArrayLiteralExp *ale, unsigned idx);
 
 /// Returns whether the array literal can be evaluated to a (LLVM) constant.
-bool isConstLiteral(ArrayLiteralExp *ale);
+bool isConstLiteral(Expression *e);
 
 /// Returns the constant for the given array literal expression.
 llvm::Constant *arrayLiteralToConst(IRState *p, ArrayLiteralExp *ale);
--- a/tests/codegen/const_struct.d
+++ b/tests/codegen/const_struct.d
@@ -1,0 +1,27 @@
+// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+// RUN: %ldc -run %s
+
+struct S0 { uint  x; }
+struct S1 { S0    y; this(this) { y.x = 1; } }
+struct S2 { S1[3] z; }
+
+struct C0 { int  *x; }
+
+void testNested() {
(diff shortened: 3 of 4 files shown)


JohanEngelen:
Test case needed! :)

gen/arrays.cpp
```diff
@@ -497,12 +497,21 @@ Expression *indexArrayLiteral(ArrayLiteralExp *ale, unsigned idx) {
 ////////////////////////////////////////////////////////////////////////////////
 
 bool isConstLiteral(ArrayLiteralExp *ale) {
-  // FIXME: This is overly pessemistic, isConst() always returns 0 e.g. for
-  // StructLiteralExps. Thus, we waste optimization potential (GitHub #506).
   for (size_t i = 0; i < ale->elements->dim; ++i) {
     // We have to check specifically for '1', as SymOffExp is classified as
     // '2' and the address of a local variable is not an LLVM constant.
```
JohanEngelen: the location of this comment is now awkward.
Also, you should add a comment that explains why the special case for structliterals is needed.

LemonBoy:
I added a small comment and fixed the indentation (the two-space indentation is quite peculiar, heh).
Also, what exactly am I supposed to test? :)

JohanEngelen:
For the test: you should test that the LLVM IR is emitted correctly. (i.e. a global constant for array of structs).  See https://wiki.dlang.org/LDC_Lit-based_testsuite.

For the comment, clarify it with something like: "We have to check _the return value of isConst_ specifically for '1',"

JohanEngelen:
> (the two-space indentation is quite peculiar, heh).

`clang-format` is your friend!

LemonBoy:
Hell, I hope I've got everything right this time! :)

LemonBoy:
Indeed, I've updated the patch!

JohanEngelen:
No test for the recursive case!
(also add a "negative" test: perhaps a nested struct thing where the nested one has one element that is not constant)

gen/arrays.cpp
```diff
@@ -496,16 +496,36 @@ Expression *indexArrayLiteral(ArrayLiteralExp *ale, unsigned idx) {
...
+  // as SymOffExp is classified as '2' and the address of a local variable is 
+  // not an LLVM constant.
+  //
+  // Examine the ArrayLiteralExps and the StructLiteralExps element by element
+  // as isConst always returns 0 on those.
+  switch (e->op) {
+    case TOKarrayliteral: {
+      auto ale = static_cast<ArrayLiteralExp *>(e);
+      for (size_t i = 0; i < ale->elements->dim; ++i) {
+        if (isConstLiteral(indexArrayLiteral(ale, i)) != 1)
```
klickverbot: `if (!isConstLiteral(…))`?

gen/arrays.cpp
```diff
@@ -496,16 +496,36 @@ Expression *indexArrayLiteral(ArrayLiteralExp *ale, unsigned idx) {
...
+  // We have to check the return value of isConst specifically for '1', 
+  // as SymOffExp is classified as '2' and the address of a local variable is 
+  // not an LLVM constant.
+  //
+  // Examine the ArrayLiteralExps and the StructLiteralExps element by element
+  // as isConst always returns 0 on those.
+  switch (e->op) {
+    case TOKarrayliteral: {
+      auto ale = static_cast<ArrayLiteralExp *>(e);
+      for (size_t i = 0; i < ale->elements->dim; ++i) {
```
klickverbot: You should actually be able to do `for (auto ale : *elements)` in LDC code.

gen/arrays.cpp
```diff
@@ -496,16 +496,36 @@ Expression *indexArrayLiteral(ArrayLiteralExp *ale, unsigned idx) {
...
+        if (isConstLiteral(indexArrayLiteral(ale, i)) != 1)
+          return false;
+      }
+    } break;
+
+    case TOKstructliteral: {
+       auto sle = static_cast<StructLiteralExp *>(e);
+       for (size_t i = 0; i < sle->elements->dim; ++i) {
```
klickverbot: Same comments as for the array literal branch apply.

LemonBoy:
Thanks for the review, I'll try to use more C++11 next time :)
Will add some tests whenever I find some time to fight with `lit`!

LemonBoy:
I hope the test's solid enough

kinke:
It's 3 dmd-testsuite tests - `runnable/{interpret,nested,sdtor}.d`.

LemonBoy:
Fixed, we shouldn't try to be too smart with non-POD structures.
Here's [1] a dustmited test case, if you observe the crash with `gdb` you can notice how, when the destructor is run, it first tries to get the address of `this` but gets zero instead. What follows is a not-so-nice zero-pointer deref.

PS: I've noticed that constant literal strings aren't considered as const by `isConst`...is this for some good reason ?

[1] https://gist.github.com/LemonBoy/3dc46189222593a4cb4d2b943cb97652

JohanEngelen:
> Fixed, we shouldn't try to be too smart with non-POD structures.

I think it is best to add a non-POD struct to the codegen test too then.

kinke:
At least in that dustmited gist, the problem with commit `eca43a2` isn't `this` (`&this` isn't null), but it's the captured variable. The initialization of the `X14838` struct literal (memcpy source for `k`) is wrong - both `S14838` literals are initialized with a null pointer, although they should get the nested context pointer in order to access captured `dtor`. What's destructed is the local `k`, not the literal.

kinke:
Yep, using `sle->sd->isNested()` instead of `!sle->sd->isPOD()` passes all tests on Win64.

gen/arrays.cpp
```diff
@@ -496,16 +496,37 @@ Expression *indexArrayLiteral(ArrayLiteralExp *ale, unsigned idx) {
...
+    } break;
+
+    case TOKstructliteral: {
+       auto sle = static_cast<StructLiteralExp *>(e);
+       if (!sle->sd->isPOD())
+         return false;
+       for (auto el : *sle->elements) {
+         if (el && !isConstLiteral(el))
+           return false;
+       }
```
kinke: There's a 3-spaces indentation here. ;)

LemonBoy:
Well, I've got that completely wrong then heh
The patch has been adjusted (and finally clang-format'ed), a test has been added, the whole PR has been rebased.
I'm toying around with the inclusion of `TOKstring` as a const element, still have to run the testsuite though.

JohanEngelen:
Can you squash and rebase?  CircleCI should be green then too.

kinke:
Thanks, much appreciated.

PR #1825, opened 2016-10-11
Allow alignment below sizeof(void*).

JohanEngelen:
Currently, alignment below the pointer size is ignored:
```d
align(1) ushort counter; // gets "align 8" in LLVM IR
```
So you cannot closely pack variables.

The minimum alignment is intended according to a39997d326f0d3da353d8b9f27ffd559e6fcc5d7.
Reverting a39997d326f0d3da353d8b9f27ffd559e6fcc5d7 to see which std.conv test fails.

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -846,10 +846,10 @@ void DtoResolveVariable(VarDeclaration *vd) {
                           linkage, nullptr, llName, vd->isThreadlocal());
     getIrGlobal(vd)->value = gvar;
 
-    // Set the alignment and use the target pointer size as lower bound.
-    unsigned alignment =
-        std::max(DtoAlignment(vd), gDataLayout->getPointerSize());
-    gvar->setAlignment(alignment);
+    // Set the alignment (it is important not to use type->alignsize because
+    // VarDeclarations can have an align() attribute independent of the type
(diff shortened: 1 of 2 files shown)


kinke:
Packing stack variables? If one really has to, it can be accomplished by a packed struct on the stack...

JohanEngelen:
Thanks for the hint @kinke . The tests were invalid so I've fixed them.
Fix also submitted upstream: https://github.com/dlang/phobos/pull/4852 .

kinke:
Just noticed there Phobos commits aren't the same. Could you cherry-pick the merged upstream one directly into the `ldc` branch and use that one instead?

JohanEngelen:
> Could you cherry-pick the merged upstream one directly into the ldc branch and use that one instead?

Nope, 2.071 frontend does not allow an expression inside `align(...)`.

kinke:
Do we need this in 1.1 then or should we delay this after the release?

JohanEngelen:
Why wait?
The merge conflict is trivially solved.

kinke:
> Why wait?

Well simply because there's no issue about this, so waiting for a few more weeks (I hope) until 1.1 is finally finalized wouldn't bother me, keeping our Phobos history cleaner.

Issue #1819, opened 2016-10-09
LDC usage with DUB is broken


rtbo:
This happens when building a static library, and is a regression introduced between `v1.1.0-beta2` and `v1.1.0-beta3`, more specifically in 6c03a3d72589f2d738c789760d23c534bc570b75.

Dub passes `-od=.dub/obj` and `-ofsome_relative_path/libfoo.a` to the compiler and the linker tries to output a file named `.dub/obj/some_relative_path/libfoo.a` and fails because the directory does not exists.

kinke:
If I understand correctly, the issue is that LDC doesn't **recursively** create non-existent directories. PR is on its way.

kinke:
Or does DMD not prepend the objdir for relative output lib filenames as LDC?

rtbo:
> Or does DMD not prepend the objdir for relative output lib filenames as LDC?

I'm not sure about DMD, but I think Dub does not expect that, because when it passes both `-od=.dub/obj` and `-ofsome_relative_path/libfoo.a` to LDC, the dir `some_relative_path` is created beforehand.

Maybe dub is wrong to pass the `-od` switch in this case. I think it's meant as a directory where LDC can place some garbage object files before linking. 
Whether dub is wrong or not, LDC usage with Dub looks broken with this new behaviour.

I'm not sure about DMD because DUB doesn't handle DMD and LDC the same way (sorry, what follows is not digestive)
```
$ dub build --vverbose --compiler=ldc2
...
ldc2 -check-printf-calls -lib -of.dub/build/library-debug-linux.posix-x86_64-ldc_0-292EEA13CB84EE6912C11893D2762E8D/liblibconfig-d.a -d-debug -g -w -oq -od=.dub/obj -d-version=Have_libconfig_d -d-version=Have_pegged -Isrc/ -I../../../home/remi/.dub/packages/pegged-0.3.2/pegged -Jtestdata src/config/config.d src/config/grammar.d src/config/package.d src/config/setting.d src/config/test.d src/config/util.d
Error: failed to create path to file: .dub/obj/.dub/build/library-debug-linux.posix-x86_64-ldc_0-292EEA13CB84EE6912C11893D2762E8D/liblibconfig-d.a
...
```
```
dub build --vverbose
...
[...]

PR #1830, opened 2016-10-14
Create static libs in -od directory for LDMD only

kinke:
Let LDC treat relative output paths as relative to the current working directory again (as it always used to until a few weeks ago). It's more intuitive and avoids breaking build systems/scripts using LDC directly.

Only LDMD continues to prepend the -od directory to the relative output path, for DMD compatibility.

Fixes issue #1819.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -151,6 +151,12 @@ cl::opt<bool> output_s("output-s", cl::desc("Write native assembly"));
 cl::opt<cl::boolOrDefault> output_o("output-o",
                                     cl::desc("Write native object"));
 
+static cl::opt<bool, true>
+    cleanupObjectFiles("cleanup-obj",
+                       cl::desc("Remove generated object files on success"),
+                       cl::ZeroOrMore, cl::ReallyHidden,
+                       cl::location(global.params.cleanupObjectFiles));
+
 // Disabling Red Zone
 cl::opt<bool, true>
     disableRedZone("disable-red-zone",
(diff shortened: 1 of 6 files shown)


driver/linker.cpp
```diff
@@ -40,6 +40,12 @@ static llvm::cl::opt<bool> staticFlag(
         "Create a statically linked binary, including all system dependencies"),
     llvm::cl::ZeroOrMore);
 
+// used by LDMD
+static llvm::cl::opt<bool> createStaticLibInObjdir(
+    "create-static-lib-in-objdir",
+    llvm::cl::desc("Create static library in -od directory (DMD-compliant)"),
+    llvm::cl::ZeroOrMore, llvm::cl::ReallyHidden);
```
kinke: I went for that additional (invisible) switch to simplify things. The other option, prepending the objdir manually to the relative `-of` path for LDC's command line, gets tricky if the user doesn't specify the `-of` option, as we'd need to figure out the default output path (based on the first final object file, i.e., first object or source file in the files list, additionally depending on `-singleobj`...).

kinke:
Also works around another issue which came up in #1819 - DMD has no object file collision issues for static libs, LDMD has.

driver/cl_options.cpp
```diff
@@ -151,6 +151,10 @@ cl::opt<bool> output_s("output-s", cl::desc("Write native assembly"));
 cl::opt<cl::boolOrDefault> output_o("output-o",
                                     cl::desc("Write native object"));
 
+static cl::opt<bool, true> cleanupObjectFiles(
+    "cleanup-obj", cl::desc("Remove generated object files on success"),
+    cl::ZeroOrMore, cl::location(global.params.cleanupObjectFiles));
```
kinke: Maybe `-cleanup-o` as it's more consistent with `-output-o`?

klickverbot: I'd say: Make it really hidden for now, and tackle the rest after the release.

PR #1832, opened 2016-10-15
libconfig-d

rtbo:
Hello

Pull-request according the project ideas page in order to remove dependency on libconfig.
Dependencies are managed by 2 submodules (`libconfig-d` and `Pegged`) in a `3rdparty` directory, and necessary files are directly compiled in `ldc2.o`
So it should be transparent for users.

I setup Travis and AppVeyor to build this branch.
- https://travis-ci.org/rtbo/ldc
- https://ci.appveyor.com/project/rtbo/ldc

Travis reports error when building on OSX: `dyld: Library not loaded: /opt/local/lib/libconfig.9.dylib` during CMake configure step. 
This dylib is not needed anymore (that's quite the purpose, isn't it?) but I can't tell where cmake is pulling this dependency from. Please give me indications.
In AppVeyor ldc emits warning that it can't find `default.switches` in the config file. I don't know how the config file looks like under windows, but I can remove the warning if it is normal. Otherwise please give me indication of what is wrong. Finally, windows build stops at this error: `Error: module md is in file 'std\digest\md.d' which cannot be read`. I don't think this is related to this PR, but can be wrong of course.


kinke:
Very nice, thanks for doing this.

> I can't tell where cmake is pulling this dependency from. Please give me indications.

````
dyld: Library not loaded: /opt/local/lib/libconfig.9.dylib
  Referenced from: /Users/travis/dlang/ldc-1.0.0/bin/ldmd2
  Reason: image not found
```

I.e., it's the used D host compiler (LDMD 1.0) which is linked dynamically against libconfig.

> In AppVeyor ldc emits warning that it can't find `default.switches` in the config file.

That's definitely not what's supposed to be happening. It's also the reason why the imports fail; no `-I` directories are set. The config file doesn't really differ on Windows. It's encoded in UTF8 without BOM afaik, but it uses Windows newlines (`\r\n`). The path to it may use backslashes instead of forward ones on Windows. So I guess one of these is causing the failure.

driver/configfile.h
```diff
@@ -16,31 +16,32 @@
...
 
 private:
   bool locate();
 
-  config_t *cfg;
-  std::string pathstr;
 
-  s_vector switches;
+  std::string pathstr;
+  ConfigData data;
```
kinke: How about getting rid of the `ConfigData` helper struct and implementing `ConfigFile` in D? Apparently all what would need to be done is converting `ConfigFile` to a struct and storing `pathstr` as C string.

rtbo: Yes, that's what I wanted to do at the start. I've chosen a safer path that modifies as little code as possible, but I prefer your option.
Any clue on how to call `exe_path::getBinDir()` from D? (it returns a `std::string`)

rtbo:
For OSX, I'll reintegrate the `brew install libconfig` in `.travis.yml`. It should do the trick.

rtbo:
It should work now.
AppVeyor still complains about a failed test (`std.file-debug`)

kinke:
I've retriggered AppVeyor. Unfortunately, `std.file` sometimes fails due to 2-seconds file timestamp granularity (FAT file systems).

CMakeLists.txt
```diff
@@ -315,6 +306,23 @@ endif()
...
+    3rdparty/pegged/pegged/dynamic/peg.d
+    3rdparty/pegged/pegged/dynamic/grammar.d
+    3rdparty/pegged/pegged/peg.d
+    3rdparty/pegged/pegged/parser.d
+    3rdparty/libconfig-d/src/config/config.d
+    3rdparty/libconfig-d/src/config/grammar.d
+    3rdparty/libconfig-d/src/config/package.d
+    3rdparty/libconfig-d/src/config/setting.d
+    3rdparty/libconfig-d/src/config/util.d
+)
```
kinke: In case those are all modules anyway, I'd prefer something compact and resilient like:

```
file(GLOB_RECURSE 3RDPARTY_SRC_D
    3rdparty/pegged/pegged/*.d
    3rdparty/libconfig-d/src/*.d
)
```

Oh, I've just seen that you excluded libconfig-d's `test.d` in this list, but that's only enabled for unittest builds, so it doesn't really count. ;)

driver/configfile.h
```diff
@@ -15,32 +15,26 @@
...
 private:
-  bool locate();
-
-  config_t *cfg;
-  std::string pathstr;
-
-  s_vector switches;
+  const char *pathcstr;
+  s_iterator switches_b;
+  s_iterator switches_e;
```
kinke: Please initialize them to nullptr here in the header and get rid of the explicit default ctor.

driver/configfile.d
```diff
@@ -0,0 +1,335 @@
+module driver.configfile;
+
+
+struct ConfigFile
```
kinke: `extern (C++)` - otherwise all fields are off by ptr size (monitor for D structs).

driver/configfile.d
```diff
@@ -0,0 +1,335 @@
...
+    
+    version(Windows)
+    {
+        // bindings missing in LDC LTS
+        import core.sys.windows.windows : HRESULT, HWND, HANDLE, DWORD, LPSTR, LPWSTR;
+
+        extern(Windows)
+        HRESULT SHGetFolderPathA(HWND, int, HANDLE, DWORD, LPSTR);
+        extern(Windows)
+        HRESULT SHGetFolderPathW(HWND, int, HANDLE, DWORD, LPWSTR);
```
kinke: I'd prefer only declaring actually used stuff - no *W and only 3 of the enums below.

kinke:
Are there any functional changes due to the C++ -> D transition of almost all of `ConfigFile` or did you take care to preserve existing behavior?

CMakeLists.txt
rtbo: So I don't change it?
Or I glob and explicit remove afterwards what is not needed?
Pegged has a lot of files that we don't need at all, so I would prefer keep it this way.
Another possibility is to make a `3rdparty[.lib][.a]` in `3rdparty/CMakeLists.txt`

driver/configfile.h
rtbo: done

driver/configfile.d
rtbo: done

rtbo:
Behavior (and C++ interface) should be the same.
I took care to locate `ldc2.conf` in the same order.

JohanEngelen:
Is there anything still preventing this from merging?

kinke:
I'd merge it if we agree that it should be part of 1.1, otherwise I'd wait (just to simplify bringing branch release-1.1 up to speed, i.e., one less commit to be reverted). If @klickverbot agrees that 1.1 should include all of master except for OSX (and the runtime functions type check, which depends on it). that is. If David is unsure whether other commits are responsible for the spurious Travis OSX failures, I'd tend to branch off a test branch (master with the 2 commits reversed) and trigger the OSX Travis jobs manually at least 20 times or so and see whether it's really just his OSX commit.

rainers:
I'm a bit late to the discussion, but there is one thing that bugs me: while it is good that we change the implementation of the config file reading to something we can actively fix in case of existing and maybe future problems, it's a bit unfortunate to replace one external dependency with another.
Would it make sense to extract just the parser (e.g. no writing required) into a few files to include directly into the LDC source?

kinke:
For me the primary advantages are:

* no runtime dependency (as opposed to linking libconfig dynamically, which is at least the case for our previous OSX releases)
* 'built-in' compilation dependencies as submodules (e.g., on Linux we currently require the libconfig-dev package; on Windows the lib needs to be built externally and the path then passed to LDC's CMake etc.)
* license

rainers:
I agree, it's a good improvement. It was just a thought when I stumbled over some complaints that implementing an adhoc parser would be pretty simple, especially if we stretch the syntax slightly by reusing the D Lexer, e.g. for number parsing. But as I never got to do it, I must not complain about an alternative implementation ;-)

JohanEngelen:
> trigger the OSX Travis jobs manually at least 20 times or so and see whether it's really just his OSX commit

Currently doing this for #1817 (LLVM 3.9).

kinke:
> Currently doing this for #1817 (LLVM 3.9).

@JohanEngelen: I don't understand. That branch is 20 commits behind master and does contain David's OSX commit...

JohanEngelen:
@kinke Should have checked that it contains the OSX stuff... dumb, sorry.

JohanEngelen:
Let's move forward on this?

JohanEngelen:
Being on a constrained internet connection for some time made me more aware of the size of things. This PR pulls in 2 external submodules, and especially the Pegged repo is quite large and pulls in even more submodules. This will increase git clone time and download size and also our release source archive size, adding a lot of files that we don't need.
I now feel much more for @rainers suggestion of stripping all of it down to a few files. From the CMake script addition, it looks like we would only have to copy 9 `.d` files into our repo and two license files. (similar to how we include FileCheck and profile-rt, instead of svn checkout their repositories)

kinke:
How about we just use a simple JSON file and parse it via `std.json` or so?

rtbo:
FWIW:
Not that I want to scuttle my own work, but I tend to agree that JSON would be simpler to handle.
`libconfig` has nice structuring features, but all we need is to select a set of switches.

kinke:
Closing in favor of #2016. Thanks Remi for hanging in there! :)

PR #1833, opened 2016-10-15
Re-check existing runtime function declarations for matching type

kinke:
I hope the revert isn't required anymore due to David's latest OSX work. We'll see.

### DIFF
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -127,22 +127,24 @@ llvm::Function *getRuntimeFunction(const Loc &loc, llvm::Module &target,
                                    const char *name) {
   checkForImplicitGCCall(loc, name);
 
-  if (!M) {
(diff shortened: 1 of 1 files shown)


kinke:
This consistently (at least twice out of 2 runs) produces a [segfault](https://travis-ci.org/ldc-developers/ldc/jobs/167953810#L5933) for the `core.thread` (release only) unittests for the OSX (static libs) Travis job. As this patch doesn't modify any behavior (just bailing out if the user provides a runtime function with incompatible signature), it's most likely a symptom of another bug that we should fix for v1.1.

Pinging @klickverbot and @JohanEngelen as you guys have OSX afaik.

kinke:
Seems to be a spurious failure, it has happened once for my other PR too. Never the shared lib job though, always the static one.

PR #1835, opened 2016-10-16
ELF: Use magic linker symbol names for begin/end of .minfo.

JohanEngelen:
I don't have Linux, but tested it on Weka's buildserver. The dot in the section name does not work (`__start_.minfo`), so I renamed the section to `__minfo` (only for ELF).

The previous bracketing of the .minfo section does not work when ThinLTO is employed. It appears that LLVM IR puts no guarantees on bracketing layout of symbols in custom sections, or it is an LLVM bug. Either way, we can rely on linker magic to fix the ThinLTO case.

@klickverbot Looks good?

### DIFF
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -101,9 +101,9 @@ enum class RegistryStyle {
   /// global.
   legacyLinkedList,
 
-  /// Module references are emitted into the .minfo section, bracketed with
-  /// extra sections to access the .minfo range. Global constructors/
-  /// destructors make sure _d_dso_registry is invoked once per ELF object.
+  /// Module references are emitted into the .minfo section. Global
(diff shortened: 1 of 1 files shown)


JohanEngelen:
> Seems like the symbol visibility is off,

I removed the `Hidden` visibility, because OS X also didn't have it. Gotta get that back in then.

JohanEngelen:
OK, green now.

JohanEngelen:
Should I put it behind an `if(LTO)` check then?

JohanEngelen:
> you might want to consider moving the global references into buildRegisterDSO now that they are no longer order-sensitive

I'll save that refactoring for later if you don't mind. (looks like creating the parameters and calling buildRegisterDSO can be moved to a new function)

PR #1837, opened 2016-10-17
Implement single-thread fences via the LDC_fence pragma.

LemonBoy:
Think `std::atomic_signal_fence` [1], but for D.
I tried implementing this as an extra and optional argument for `LDC_fence` but then discovered the hard way that it would break the bootstrap cycle.

[1] http://en.cppreference.com/w/cpp/atomic/atomic_signal_fence

### DIFF
--- a/gen/tocall.cpp
+++ b/gen/tocall.cpp
@@ -329,11 +329,18 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
 
   // fence instruction
   if (fndecl->llvmInternal == LLVMfence) {
-    if (e->arguments->dim != 1) {
-      e->error("fence instruction expects 1 arguments");
+    if (e->arguments->dim < 1 || e->arguments->dim > 2) {
+      e->error("fence instruction expects 1 (or 2) arguments");
       fatal();
     }
-    p->ir->CreateFence(llvm::AtomicOrdering((*e->arguments)[0]->toInteger()));
+    auto atomicOrdering =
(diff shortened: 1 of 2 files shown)


JohanEngelen:
(tip: when you make modifications to druntime or phobos, it helps if you provide links to those changes)

JohanEngelen:
> but then discovered the hard way that it would break the bootstrap cycle.

Can you elaborate?

LemonBoy:
I haven't committed the changes to `druntime` yet, I wanted some feedback first on the backend-part.
Wrt the bootstrapping cycle, if you add another argument to `llvm_memory_fence` then this [1] check fails when compiling druntime, at least that's what I've observed.

[1] https://github.com/ldc-developers/ldc/blob/b17e1485edf57dde46496c4991d2430566bec712/gen/tocall.cpp#L332

JohanEngelen:
I think it's nicer if you add the second argument to the llvm_memory_fence.
Why not allow 1 or 2 arguments to llvm_memory_fence?

LemonBoy:
It is indeed nicer, but afaics the problem lies in the build method that uses the host `ldc` to compile the new runtime, making the aforementioned check fail.
One could sidestep this problem by adding a version check (to be removed at a later stage), though.

JohanEngelen:
The host compiler is not building druntime. The aforementioned check fails, because you have not modified it? ;)

JohanEngelen:
Can you add a simple IR test for this? (so that we'll notice when the enum values inside LLVM change)

PR #1838, opened 2016-10-18
Consider string literals as constant in isConstLiteral.

LemonBoy:
Let's see if the test suite likes this.

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -522,6 +522,10 @@ bool isConstLiteral(Expression *e) {
     }
   } break;
 
+  // isConst also returns 0 for string literals that are obviously constant.
+  case TOKstring:
+    return true;
+
   default:
     if (e->isConst() != 1)
       return false;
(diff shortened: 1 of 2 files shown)


JohanEngelen:
`Expression.isConst` is `final` and refers to constfold.d `isConst(Expression e)`: https://github.com/ldc-developers/ldc/blob/9f998a398d6b992d6cc0de10bc3ad69140e5a3a0/ddmd/constfold.d#L49-L77.

Tampering with that function is probably bad for the DMDfe constfolding. But the implementation looks kind-of unfit for our purpose, no? 
Can you check what happens with constant pointers (pointer to function or global var), and why is it only returning 1 for TOKint64 and not for, say, TOKchar ?  I'm confused.

Something unrelated: does `isConst(Expression e)` need fixing for `@weak` functions? (pointer to such a function is incorrectly labeled as `isConst`?) Or do I completely misunderstand it?

JohanEngelen:
I think all that's missing is a testcase? ;)

PR #1845, opened 2016-10-19
Implement -finstrument-functions.

### LINKED ISSUE #1839: Add function instrumentation and profiling options
This tool https://github.com/bisco/uftrace.git needs function profiling information for generating nice profile graphs. Also look at this CppCon 2016 presentation video https://youtu.be/LNav5qvyK7I

@JohanEngelen says it should be easy to implement in LDC :-) We need something equivalent to -pg option of GCC.

LemonBoy:
Closes #1839.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -455,6 +455,11 @@ cl::opt<std::string> usefileInstrProf(
     cl::ValueRequired);
 #endif
 
+cl::opt<bool>
+    instrumentFunctions("finstrument-functions",
+                        cl::desc("Instrument function entry and exit with "
+                                 "GCC-compatible profiling calls"));
+
 static cl::extrahelp footer(
     "\n"
     "-d-debug can also be specified without options, in which case it enables "
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -962,6 +962,8 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   // debug info - after all allocas, but before any llvm.dbg.declare etc
   gIR->DBuilder.EmitFuncStart(fd);
 
+  emitInstrumentationFnEnter(fd);
+
   // this hack makes sure the frame pointer elimination optimization is
   // disabled.
   // this this eliminates a bunch of inline asm related issues.
@@ -1076,6 +1078,8 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   } else if (!gIR->scopereturned()) {
     // llvm requires all basic blocks to end with a TerminatorInst but DMD does
     // not put a return statement in automatically, so we do it here.
+
+    emitInstrumentationFnLeave(fd);
 
     // pass the previous block into this block
     gIR->DBuilder.EmitStopPoint(fd->endloc);
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -21,6 +21,7 @@
 #include "ir/irfunction.h"
 #include "ir/irtype.h"
 #include "ir/irtypefunction.h"
+#include "driver/cl_options.h"
 #include "ldcbindings.h"
 #include "mars.h"
 #include "module.h"
@@ -324,6 +325,16 @@ static void buildRuntimeModule() {
       Attr_1_2_NoCapture(Attr_1_NoCapture, 2, llvm::Attribute::NoCapture),
       Attr_1_3_NoCapture(Attr_1_NoCapture, 3, llvm::Attribute::NoCapture),
       Attr_1_4_NoCapture(Attr_1_NoCapture, 4, llvm::Attribute::NoCapture);
+
+  //////////////////////////////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////////////
+
+  // void __cyg_profile_func_enter(void *callee, void *caller)
+  // void __cyg_profile_func_exit(void *callee, void *caller)
+  createFwdDecl(LINKc, voidTy,
+                {"__cyg_profile_func_exit", "__cyg_profile_func_enter"},
+                {voidPtrTy, voidPtrTy}, {}, Attr_NoUnwind);
 
   //////////////////////////////////////////////////////////////////////////////
   //////////////////////////////////////////////////////////////////////////////
@@ -738,4 +749,29 @@ static void buildRuntimeModule() {
       break;
     }
   }
+}
+
+static void emitInstrumentationFn(const char *name) {
+  LLFunction *fn = getRuntimeFunction(Loc(), gIR->module, name);
+
+  // Grab the address of the calling function
+  auto *caller =
+      gIR->ir->CreateCall(GET_INTRINSIC_DECL(returnaddress), DtoConstInt(1));
+  auto callee = DtoBitCast(gIR->topfunc(), getVoidPtrType());
+
+#if LDC_LLVM_VER >= 307
+  gIR->ir->CreateCall(fn, {callee, caller});
+#else
+  gIR->ir->CreateCall2(fn, callee, caller);
+#endif
+}
+
+void emitInstrumentationFnEnter(FuncDeclaration *decl) {
+  if (opts::instrumentFunctions && decl->emitInstrumentation)
+    emitInstrumentationFn("__cyg_profile_func_enter");
+}
+
+void emitInstrumentationFnLeave(FuncDeclaration *decl) {
+  if (opts::instrumentFunctions && decl->emitInstrumentation)
+    emitInstrumentationFn("__cyg_profile_func_exit");
 }
--- a/tests/codegen/instrumentation.d
+++ b/tests/codegen/instrumentation.d
@@ -1,0 +1,29 @@
+// RUN: %ldc -c -output-ll -finstrument-functions -of=%t.ll %s && FileCheck %s < %t.ll
+
+void fun0 () {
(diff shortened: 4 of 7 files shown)


driver/cl_options.cpp
```diff
@@ -455,6 +455,11 @@ cl::opt<std::string> usefileInstrProf(
     cl::ValueRequired);
 #endif
 
+cl::opt<bool, true> instrumentFunctions(
+    "finstrument-functions",
+    cl::desc("Instrument function entry and exit with profiling calls"),
```
klickverbot: Maybe say "GCC-compatible profiling calls" to be clear?

gen/tollvm.cpp
```diff
@@ -782,3 +782,32 @@ LLValue *DtoAggrPaint(LLValue *aggr, LLType *as) {
...
+
+  // Generate the prototype to call either:
+  // void __cyg_profile_func_enter (void *callee, void *caller)
+  // void __cyg_profile_func_exit (void *callee, void *caller)
+  LLFunction *fn = gIR->module.getFunction(fnName);
+  if (!fn) {
+    LLType *VoidTy = LLType::getVoidTy(gIR->context());
+    LLType *Tys[] = {VoidPtrTy, VoidPtrTy};
+    auto fty = LLFunctionType::get(VoidTy, Tys, false);
+    fn = LLFunction::Create(fty, LLGlobalValue::ExternalLinkage, fnName,
```
klickverbot: Are there any LLVM attributes that could allow more optimisations here, or do we need to allow arbitrary hook functions?

driver/cl_options.cpp
LemonBoy: That's what GCC says in the `--help`, verbatim, and Clang also says `Generate calls to instrument function entry and exit`, so I think that there's no need to further clarify the description.

gen/tollvm.cpp
LemonBoy: We can't make any assumption about what's done in the functions, better play it safe.

driver/cl_options.cpp
JohanEngelen: Adding "GCC" may help, because you;d have to link with a GCC-compatible runtime I guess? (On Windows, it would help folks figure out what's wrong?)

gen/tollvm.cpp
JohanEngelen: I think we usually put the definition of these "runtime" functions in runtime.cpp. 
Attributes you may think about are "nocapture" "readonly" or something like that. Runtime.cpp has a bunch of that.
I had trouble finding where these cyg_profile functions are defined, did you find a link to an implementation?

tests/codegen/instrumentation.d
```diff
@@ -0,0 +1,16 @@
...
+  // CHECK-NEXT: call void @__cyg_profile_func_enter
+  // CHECK-NEXT: %2 = call i8* @llvm.returnaddress(i32 1)
+  // CHECK-NEXT: call void @__cyg_profile_func_exit
+  // CHECK-NEXT: ret void
+  return;
+}
+
+pragma(LDC_profile_instr, false)
+void fun1 () {
+  // CHECK: ret void
```
JohanEngelen: this doesn't check for the absense of __cyg_profile calls

tests/codegen/instrumentation.d
```diff
@@ -0,0 +1,16 @@
+// RUN: %ldc -c -output-ll -finstrument-functions -of=%t.ll %s && FileCheck %s < %t.ll
+
+void fun0 () {
+  // CHECK: %1 = call i8* @llvm.returnaddress(i32 1)
```
JohanEngelen: You could add some simple code to the function to test that the profiling function calls enclose any "body" code.

Also you could add a function with more than one exit path, and a function with an exception throwing exit path, or a non-trivial destructor call. Again to check where the profiling calls are placed.

gen/tollvm.cpp
JohanEngelen: Are these profiling functions ever used in a debugging context? What I mean is: is it OK if the compiler reorders things, e.g.:
```d
void foo() {
    __cyg_profile_func_start(...);
    smth();
    __cyg_profile_func_exit(...);
}
```
optimized (for whatever reason) to:
```d
void foo() {
    __cyg_profile_func_start(...);
    __cyg_profile_func_exit(...);
    smth();
}
```
To prevent that, perhaps an attribute is needed on the function declarations (or perhaps the absence of attributes means that LLVM assumes the worst).

tests/codegen/instrumentation.d
LemonBoy: My bad, I'm still convinced that lit's smarter than it really is :)

gen/tollvm.cpp
LemonBoy: I wasn't 100% sure about where to put this, I can move it to `runtime.cpp` if that's deemed a best place. I'll also have a look at what those attributes are for and which ones are applicable.

WRT the definition of the symbols, `LD_DEBUG` shows 
`binding file ./test [0] to /usr/lib/libc.so.6 [0]: normal symbol `__cyg_profile_func_enter' [GLIBC_2.2.5]`
so it's `libc` that provides the stubs...we might provide such stubs in the druntime, no ?

driver/cl_options.cpp
LemonBoy: Indeed, the stubbed out functions are actually defined in `libc.so` and it didn't occur to me that the MSVC libc doesn't have those (doesn't it ?).

JohanEngelen:
Great stuff btw :-)

ddmd/globals.h
```diff
@@ -202,6 +202,7 @@ struct Param
     OUTPUTFLAG output_bc;
     OUTPUTFLAG output_s;
     OUTPUTFLAG output_o;
+    bool instrumentFunctions;
```
JohanEngelen: this must be added to the struct in globals.d too. May fix the CI tester failures.

(it'd be so nice if there was a tool to check types that are declared in both C++ and D!)

gen/tollvm.cpp
JohanEngelen: I think in `runtime.cpp` is best:  for example, you forgot to explicitly set the C-calling convention, and there may be other things that end up forgotten. The `createFwdDecl` function was invented for this ;-)

tests/codegen/instrumentation.d
```diff
@@ -0,0 +1,33 @@
...
+  // CHECK-NEXT: call void @__cyg_profile_func_exit
+  // CHECK-NEXT: ret void
+  return;
+}
+
+pragma(LDC_profile_instr, false)
+int fun1 (int x) {
+  // CHECK: %x = alloca i32, align 4
+  // CHECK-NEXT: store i32 %x_arg, i32* %x
+  // CHECK-NEXT: ret i32 42
```
JohanEngelen: to check the absence of __cyg_profile_func_{enter,exit}, you should just do a `// CHECK-NOT: __cyg_profile_func_`.

Because you only want to check that for this one particular function, you should add `CHECK-LABEL` checks for each function, such that the file is "sectioned".

tests/codegen/instrumentation.d
```diff
@@ -0,0 +1,33 @@
...
+  return 42;
+}
+
+bool fun2 (int x) {
+  // CHECK: %1 = call i8* @llvm.returnaddress(i32 1)
+  // CHECK-NEXT: call void @__cyg_profile_func_enter
+  if (x < 10)
+    // CHECK: if:
+    // CHECK-NEXT: %{{.}} = call i8* @llvm.returnaddress(i32 1)
+    // CHECK-NEXT: call void @__cyg_profile_func_exit
```
JohanEngelen: how about
```
// CHECK: call void @__cyg_profile_func_exit
// CHECK-NEXT: ret
```
here, and at the end of the functions?

ddmd/globals.h
JohanEngelen: actually: better to not add it to globals.param, and to add it to cl_options.h.

gen/functions.cpp
```diff
@@ -973,6 +973,9 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   // debug info - after all allocas, but before any llvm.dbg.declare etc
   gIR->DBuilder.EmitFuncStart(fd);
 
+  if (opts::instrumentFunctions && fd->emitInstrumentation)
+    emitInstrumentationFn(false);
```
JohanEngelen: This is a little nitpicky / subjective, but how about adding two helper functions `emitInstrumentationFuncStart(FuncDeclaration *fd)` and `emitInstrumentationFuncEnd` in runtime.h?
Note that other `emitSomething` functions do the check whether to actually do something inside, instead of outside like you do here.

```cpp
inline void emitInstrumentationFuncStart(FuncDeclaration *fd) {
  if (opts::instrumentFunctions && fd->emitInstrumentation)
    emitInstrumentationFn(false);
}
```

gen/runtime.cpp
```diff
@@ -327,6 +327,16 @@ static void buildRuntimeModule() {
   //////////////////////////////////////////////////////////////////////////////
   //////////////////////////////////////////////////////////////////////////////
 
+  // void __cyg_profile_func_enter(void *callee, void *caller)
+  // void __cyg_profile_func_exit(void *callee, void *caller)
+  createFwdDecl(LINKc, voidTy,
+                {"__cyg_profile_func_exit", "__cyg_profile_func_enter"},
+                {voidPtrTy, voidPtrTy});
```
JohanEngelen: as we discussed, please add the nounwind attribute here.

JohanEngelen:
#1961

Issue #1829, opened 2016-10-14
Dangling pointers after replacing globals due to initializer type mismatch


nordlow:
When I replace dmd with ldmd2 in the last call, ldmd2 crashes as

CC=c++ ldmd2 -ofdmd -m64 -vtls -J. -J../res -L-lstdc++ -version=MARS -wi -O -release -inline access.d aggregate.d aliasthis.d apply.d argtypes.d arrayop.d arraytypes.d attrib.d builtin.d canthrow.d clone.d complex.d cond.d constfold.d cppmangle.d ctfeexpr.d dcast.d dclass.d declaration.d delegatize.d denum.d dimport.d dinifile.d dinterpret.d dmacro.d dmangle.d dmodule.d doc.d dscope.d dstruct.d dsymbol.d dtemplate.d dversion.d entity.d errors.d escape.d expression.d func.d globals.d hdrgen.d id.d identifier.d impcnvtab.d imphint.d init.d inline.d intrange.d json.d lexer.d lib.d link.d mars.d mtype.d nogc.d nspace.d opover.d optimize.d parse.d sapply.d sideeffect.d statement.d staticassert.d target.d tokens.d traits.d utf.d visitor.d typinf.d utils.d statementsem.d safe.d objc_stubs.d libelf.d scanelf.d irstate.d toelfdebug.d toctype.d glue.d gluelayer.d todt.d tocsym.d toir.d dmsc.d tocvdebug.d backend/bcomplex.d backend/cc.d backend/cdef.d backend/cgcv.d backend/code.d backend/cv4.d backend/dt.d backend/el.d backend/global.d backend/obj.d backend/oper.d backend/outbuf.d backend/rtlsym.d backend/ty.d backend/type.d tk/dlist.d root/aav.d root/array.d root/ctfloat.d root/file.d root/filename.d root/man.d root/outbuffer.d root/port.d root/response.d root/rmem.d root/rootobject.d root/speller.d root/stringtable.d newdelete.o glue.a backend.a
tk/dlist.d(51): list_inited is thread local
backend/cgcv.d(26): ftdbname is thread local
0  ldc2            0x00000000013af6a5
1  ldc2            0x00000000013ae05e
2  ldc2            0x00000000013ae19a
3  libpthread.so.0 0x00007f3ce59443d0
4  ldc2            0x00000000012fcd94
5  ldc2            0x0000000001350a42
6  ldc2            0x00000000013532fd
7  ldc2            0x0000000001315d2a
[...]

PR #1846, opened 2016-10-21
Fix union layout and initialization

kinke:


kinke:
Doesn't fix #1829.

kinke:
The _init type should now be always identical to the normal type (the used fields for the normal type are based on the field initializers). Only the initializer of globals may use a different type, based on which fields are set.
Getting rid of _init types may work around #1829 in that particular case, as that `config` global is default-initialized IIRC, so the type wouldn't change and the global wouldn't need to be replaced.
But #1829 is a more general issue - either a bug in LLVM 3.9 (unlikely I guess) or caused by us keeping around a reference to the old, replaced global, and using it after the replacement.

ir/irtypeaggr.cpp
```diff
@@ -53,110 +54,112 @@ void AggrTypeBuilder::addType(llvm::Type *type, unsigned size) {
 }
 
 void AggrTypeBuilder::addAggregate(AggregateDeclaration *ad) {
+  addAggregate(ad, nullptr, /* addAliases = */ true);
+}
+
+namespace {
+enum FieldPriority {
```
klickverbot: `enum class`? <3

ir/irtypeaggr.cpp
```diff
@@ -53,110 +54,112 @@ void AggrTypeBuilder::addType(llvm::Type *type, unsigned size) {
 }
 
 void AggrTypeBuilder::addAggregate(AggregateDeclaration *ad) {
+  addAggregate(ad, nullptr, /* addAliases = */ true);
```
klickverbot: Just make the parameter an enum then to make it self-documenting?

ir/irtypeaggr.cpp
```diff
@@ -53,110 +54,112 @@ void AggrTypeBuilder::addType(llvm::Type *type, unsigned size) {
...
-          size_t v_begin = vd->offset;
-          size_t v_end = v_begin + vd->type->size();
+        size_t v_begin = vd->offset;
+        size_t v_end = v_begin + vd->type->size();
 
-          if (v_begin >= f_end || v_end <= f_begin) {
-            continue;
+        if (v_begin < f_end && v_end > f_begin) {
+          if (addAliases && v_begin == f_begin &&
+              DtoMemType(data[i]->type) == DtoMemType(field->type)) {
```
klickverbot: Couldn't this just be `vd`? In fact, if you loop over all of `data` anyway, can't you just turn the whole thing into a range-based for loop?

ir/irtypeaggr.cpp
```diff
@@ -53,110 +54,112 @@ void AggrTypeBuilder::addType(llvm::Type *type, unsigned size) {
...
-  for (size_t index = 0; index < n; ++index) {
-    VarDeclaration *field = ad->fields[index];
+  // list of pairs: alias => actual field (same offset, same LL type)
+  llvm::SmallVector<std::pair<VarDeclaration *, VarDeclaration *>, 16> aliases;
 
-    // _init is !null for explicit inits
-    if (field->_init != nullptr && !field->_init->isVoidInitializer()) {
-      IF_LOG Logger::println("adding explicit initializer for struct field %s",
-                             field->toChars());
+  unsigned errors = global.errors;
```
klickverbot: (nitpick in old code: `const`)

tests/codegen/union.d
```diff
@@ -0,0 +1,58 @@
...
+
+struct S
+{
+    char c;              // default initializer: 0xff
+    uint ui;
+    bool[2] bools;       // make sure the 2 bools are extended to 2 bytes
+    char[2][4] multidim; // multidimensional init based on a single 0xff char
+}
+// CHECK-DAG: %union.S                     = type { i8, [3 x i8], i32, [2 x i8], [4 x [2 x i8]], [2 x i8] }
+// CHECK-DAG: @_D5union1S6__initZ          = constant %union.S { i8 -1, [3 x i8] zeroinitializer, i32 0, [2 x i8] zeroinitializer, [4 x [2 x i8]] {{\[}}[2 x i8] c"\FF\FF", [2 x i8] c"\FF\FF", [2 x i8] c"\FF\FF", [2 x i8] c"\FF\FF"], [2 x i8] zeroinitializer }
```
kinke: It was really necessary to escape the `[[` (FileCheck variable matcher) in `[4 x [2 x i8]] [[2 x i8] c"\FF\FF", ...]` via `{{\[}}[`. ;(

kinke:
Alright, the 2nd commit fixes #1829 for DMD at least, as expected. It should restrict that bug to only a few exotic cases.

ir/irtypeaggr.cpp
kinke: I'd have had to implement the < and -- operators then, so I changed my initial `enum class` to `enum`. ;)

ir/iraggr.cpp
```diff
@@ -167,23 +172,30 @@ IrAggr::createInitializerConstant(const VarInitMap &explicitInitializers,
...
+
+  // get LL field types
+  llvm::SmallVector<llvm::Type *, 16> types;
+  types.reserve(constants.size());
+  for (auto c : constants)
+    types.push_back(c->getType());
+
+  auto llStructType = getLLStructType();
+  bool notCompatible = (types.size() != llStructType->getNumElements());
+  if (!notCompatible) {
```
klickverbot: Style nitpick: I'd remove the negation here and call the variable just `compatible` –  my brain parsed `!notCompatible` parsed in interesting ways when just skimming the source.  (Or at least, `incompatible`.)

kinke:
Thanks for the blazingly fast review. ;)

> compile time pessimisations

Based on the admittedly vague CI results, there seem to be no significant changes.

> nicely abstract away all the bool zext stuff

I was thinking the same, it's a pain. Using `i8` generally and, in case it's really worth the trouble, truncating right before the actual operations (and zexting immediately again when storing) should decomplicate things substantially. That was quite a nasty bug and invisible due to the rather wild bitcasting...

kinke:
I also had to zext scalar bools to prevent type mismatches. Furthermore, I've had a look at how the dynamic struct literals are generated, noticed an issue and fixed + refactored that part too. I hope this PR now fixes our union-related bugs for good and gets stuff like #1719 working again.

kinke:
> Was there an actual bug – as in, wrong codegen – anywhere?

I did a quick test with bools with an intermediate commit, which seemed to suggest that LLVM didn't pack scalar `i1` and `[N x i1]` struct fields, using a byte for each `i1` instead. So that behavior for constant initializers may have been okay, although not desirable.

gen/toir.cpp
```diff
@@ -76,58 +75,49 @@ static LLValue *write_zeroes(LLValue *mem, unsigned start, unsigned end) {
...
+
+  unsigned offset = 0;
+  for (const auto &d : data) {
+    const auto vd = d.field;
+    const auto expr = d.expr;
 
     // initialize any padding so struct comparisons work
-    if (vd->offset != offset)
+    if (vd->offset != offset) {
+      assert(vd->offset > offset);
```
kinke: Damn, this assertion doesn't hold for the code in https://github.com/ldc-developers/ldc/issues/1324. :(

kinke:
I've amended the last commit to match DMD behavior for [DMD issue 16471](https://issues.dlang.org/show_bug.cgi?id=16471) referenced by #1324 (which is resolved by this, as we depend on a front-end fix).

Issue #1822, opened 2016-10-10
Calling abstract method


Rikarin:
Hello,
I did this

```d
class Child : Parent {
final override void dispose() { super.dispose(); }
}

class Parent {
abstract void dispose();
}
```

and it thrown this against any useful error.

```
0  ldc2            0x00000000013af6a5
1  ldc2            0x00000000013ae05e
2  ldc2            0x00000000013ae19a
3  libpthread.so.0 0x00007fb60cc0d080
4  ldc2            0x0000000000696134
5  ldc2            0x000000000066929d
6  ldc2            0x000000000066951f
7  ldc2            0x0000000000668caa
8  ldc2            0x000000000068a112
9  ldc2            0x000000000068b104
10 ldc2            0x000000000068b104
11 ldc2            0x000000000068d924
12 ldc2            0x000000000064418b
13 ldc2            0x0000000000623c43
14 ldc2            0x0000000000623c43
15 ldc2            0x0000000000624393
16 ldc2            0x0000000000623c43
17 ldc2            0x0000000000623c43
18 ldc2            0x0000000000624cae
19 ldc2            0x00000000006232e9
20 ldc2            0x00000000006db25c
21 ldc2            0x000000000068642f
22 ldc2            0x000000000058a1c9
23 ldc2            0x000000000068847b
24 ldc2            0x00000000013e2fff _D2rt6dmain211_d_run_mainUiPPaPUAAaZiZ6runAllMFZ9__lambda1MFZv + 15
25 ldc2            0x00000000013e2fc4
26 ldc2            0x00000000013e2eed
27 libc.so.6       0x00007fb60c363291 __libc_start_main + 241
28 ldc2            0x000000000049c751
```

I have LDC 1.1.0 beta 3

kinke:
Thanks, finally something reproducible from your invalid-code collection. :) DMD only complains when linking, LDC dies in a DValue ctor assertion.

KevinBrogan:
It only crashes if the child class is defined before the base class.

Crashes
===
```
class Derived : Base
{
	override void coreDump()
	{
		super.coreDump();
	}
}

class Base
{
	abstract void coreDump();
}
```

Doesn't Crash
===
```
class Base
{
	abstract void coreDump();
}

class Derived : Base
{
	override void coreDump()
	{
		super.coreDump();
	}
}
```

kinke:
Thx for that detail.

kinke:
Fixed by #1854.

PR #1854, opened 2016-10-25
Produce linking errors when calling abstract functions

kinke:
Fixes issue #1822.

`DtoResolveFunction()` explicitly excludes body-less abstract functions from being automatically declared. If we want to emit linking errors like DMD, we need a LL function for the call (we previously hadn't for abstract functions, so it was null and LDC crashed). Thus make sure the function is resolved and declared by invoking `DtoDeclareFunction()`.


kinke:
A test (probably in dmd-testsuite) will follow once merged.

kinke:
> document the behaviour of the different Dto{Resolve, Declare, Define} functions appropriately

I'd rather leave that to you, as I don't have a sufficient overview of this matter.

PR #1855, opened 2016-10-26
Only emit llvm.used _once_ in an LLVM module

JohanEngelen:
Before, when compiling multiple D modules into one LLVM module, we would emit llvm.used multiple times. All except the first would actually be recognized as llvm.used (the others would be renamed to llvm.used.* which is not recognized by LLVM).
This fixes that.

(For testing, I used the testcase that would start bugging for ThinLTO. I used the OS X triple and moduleinfo stuff, because I tried but did not find another way to get things to be added to llvm.used. Symbols referenced in inline asm are not added to llvm.used, surprisingly, but LLVM parses the assembly to find such references I think, so not a bug?)

### DIFF
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -73,6 +73,26 @@ void emitLinkerOptions(IRState &irs, llvm::Module &M, llvm::LLVMContext &ctx) {
 #endif
   }
 }
+
+void emitLLVMUsedArray(IRState &irs) {
+  if (irs.usedArray.empty()) {
+    return;
+  }
+
+  auto *i8PtrType = llvm::Type::getInt8PtrTy(irs.context());
+
+  // Convert all elements to i8* (the expected type for llvm.used)
+  for (auto &elem : irs.usedArray) {
+    elem = llvm::ConstantExpr::getBitCast(elem, i8PtrType);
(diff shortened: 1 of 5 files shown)


driver/codegenerator.cpp
```diff
@@ -73,6 +73,26 @@ void emitLinkerOptions(IRState &irs, llvm::Module &M, llvm::LLVMContext &ctx) {
...
+  }
+
+  auto *i8PtrType = llvm::Type::getInt8PtrTy(irs.context());
+
+  // Convert all elements to i8* (the expected type for llvm.used)
+  for (auto &elem : irs.usedArray) {
+    elem = llvm::ConstantExpr::getBitCast(elem, i8PtrType);
+  }
+
+  auto *arrayType = llvm::ArrayType::get(i8PtrType, irs.usedArray.size());
```
klickverbot: Nitpick: Using `auto *` and `auto` for pointers right next to each other is confusing – just stick with one. (Personally, I prefer `auto`, but YMMV.)

JohanEngelen: OK. I'm going to use `auto *` if you don't mind , I like that it shows pointerness.
Otherwise OK?

PR #1856, opened 2016-10-26
Map `export` visibility to LLVM DLL storage classes

### LINKED ISSUE #1745: Shared Library Not Exporting Symbols on Windows
It doesn't seem as though anything is being exported, even "gc_setProxy" which rt_LoadLibrary() from core.runtime tries to access to link the garbage collectors. This also includes functions with "export", not  actually being exported.

kinke:
Trying to fix #1745.

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -538,6 +538,12 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
 
   func->setCallingConv(gABI->callingConv(func->getFunctionType(), link, fdecl));
 
+  if (global.params.isWindows && fdecl->isExport()) {
+    func->setDLLStorageClass(fdecl->isImportedSymbol()
+                                 ? LLGlobalValue::DLLImportStorageClass
+                                 : LLGlobalValue::DLLExportStorageClass);
+  }
+
   IF_LOG Logger::cout() << "func = " << *func << std::endl;
 
   // add func to IRFunc
@@ -906,11 +912,12 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
 
   const auto f = static_cast<TypeFunction *>(fd->type->toBasetype());
   IrFuncTy &irFty = irFunc->irFty;
-  llvm::Function *func = irFunc->func;;
+  llvm::Function *func = irFunc->func;
 
   const auto lwc = lowerFuncLinkage(fd);
   if (linkageAvailableExternally) {
     func->setLinkage(llvm::GlobalValue::AvailableExternallyLinkage);
+    func->setDLLStorageClass(llvm::GlobalValue::DefaultStorageClass);
     // Assert that we are not overriding a linkage type that disallows inlining
     assert(lwc.first != llvm::GlobalValue::WeakAnyLinkage &&
            lwc.first != llvm::GlobalValue::ExternalWeakLinkage &&
@@ -918,6 +925,8 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   } else {
     setLinkage(lwc, func);
   }
+
+  assert(!func->hasDLLImportStorageClass());
 
   // On x86_64, always set 'uwtable' for System V ABI compatibility.
   // TODO: Find a better place for this.
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -930,8 +930,8 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
 
   // On x86_64, always set 'uwtable' for System V ABI compatibility.
   // TODO: Find a better place for this.
-  // TODO: Is this required for Win64 as well?
-  if (global.params.targetTriple->getArch() == llvm::Triple::x86_64) {
+  if (global.params.targetTriple->getArch() == llvm::Triple::x86_64 &&
+      !global.params.isWindows) {
     func->addFnAttr(LLAttribute::UWTable);
   }
   if (opts::sanitize != opts::None) {
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -851,6 +851,14 @@ void DtoResolveVariable(VarDeclaration *vd) {
     // as well).
     gvar->setAlignment(DtoAlignment(vd));
 
+    /* TODO: set DLL storage class when `export` is fixed (an attribute)
+    if (global.params.isWindows && vd->isExport()) {
+      auto c = vd->isImportedSymbol() ? LLGlobalValue::DLLImportStorageClass
+                                      : LLGlobalValue::DLLExportStorageClass;
+      gvar->setDLLStorageClass(c);
+    }
+    */
+
     applyVarDeclUDAs(vd, gvar);
 
     IF_LOG Logger::cout() << *gvar << '\n';
--- a/gen/declarations.cpp
+++ b/gen/declarations.cpp
@@ -356,6 +356,7 @@ public:
           setLinkage(lwc, newGvar);
 
           newGvar->setAlignment(gvar->getAlignment());
+          newGvar->setDLLStorageClass(gvar->getDLLStorageClass());
           applyVarDeclUDAs(decl, newGvar);
           newGvar->takeName(gvar);
 
@@ -367,6 +368,8 @@ public:
           gvar = newGvar;
           irGlobal->value = newGvar;
         }
+
+        assert(!gvar->hasDLLImportStorageClass());
 
         // Now, set the initializer.
         assert(!irGlobal->constInit);
--- a/tests/codegen/export.d
+++ b/tests/codegen/export.d
@@ -1,0 +1,19 @@
+// RUN: %ldc -output-ll -of=%t.ll %s
+// RUN: FileCheck %s < %t.ll
+
+// REQUIRES: Windows
+
+export
+{
+    // CHECK-DAG: define dllexport {{.*}}_D6export11exportedFooFZv
+    void exportedFoo() {}
+
+    // CHECK-DAG: declare dllimport {{.*}}_D6export11importedFooFZv
+    void importedFoo();
+}
+
+void bar()
+{
+    exportedFoo();
+    importedFoo();
+}
--- a/tests/codegen/export_crossModuleInlining.d
+++ b/tests/codegen/export_crossModuleInlining.d
@@ -1,0 +1,19 @@
+// Make sure exported functions can be cross-module inlined without exporting the local function copy.
+
+// REQUIRES: atleast_llvm307
+
+// RUN: %ldc -O -release -enable-cross-module-inlining -output-ll -of=%t.ll -I%S/inputs %s
+// RUN: FileCheck %s < %t.ll
+
+import export2;
+
+// CHECK-NOT: _D7export23fooFZi
+
+// CHECK: define {{.*}}_D26export_crossModuleInlining3barFZi
+int bar()
+{
+    // CHECK-NEXT: ret i32 666
+    return foo();
+}
+
+// CHECK-NOT: _D7export23fooFZi
--- a/tests/codegen/inputs/export2.d
+++ b/tests/codegen/inputs/export2.d
@@ -1,0 +1,1 @@
+export int foo() { return 666; }


thewilsonator:
How does this interact with whatever Benjamin Thaut is doing about making export be an attribute?

kinke:
First time I hear about that, but then I've only discovered it's actually a visibility 30 mins ago. ;) As long as the `DSymbol::isExport()` API then returns true if an `export` attribute is applied, we shouldn't have to care.

thewilsonator:
He gave a DConf talk about it, might be worth a watch. IIRC the problem was you can't tell if you're supposed be importing or exporting.

kinke:
> IIRC the problem was you can't tell if you're supposed be importing or exporting.

Yep, so I came up with some logic to be tested and reviewed here. It basically boils down to that the `export` symbol (global/function) is exported if it's part of (declared in or template-instantiated in) a compiled (and not just imported) module and otherwise imported.

thewilsonator:
Seems reasonable. I just wanted to note that it will probably be superseded by Ben's work.

kinke:
That's fine, but people already need something for v1.1. :)

kinke:
With the current logic, all modules of a binary should be compiled with a single ldc2/ldmd2 call. Otherwise, other objects (in the same binary) import the symbol via `dllimport`, i.e., access it via an unnecessary additional indirection (at least on Windows). I've just tested lnking independently compiled objects, and the MS linker emits a warning in this case:

```
current.obj : warning LNK4217: locally defined symbol _D8current24foo2FZv imported in function _Dmain
```

kinke:
On Windows, the MS linker now outputs the following line for each binary:

```
   Creating library current.lib and object current.exp
```

This is because the functions `gc_clrProxy` and `gc_setProxy` are always exported, and so the linker generates import .lib and that .exp file. This leads to an output-matching test failure (`runnable/test13613.d`)...

kinke:
LLVM docs say that `uwtable` is normally required for System V x86_64, but doesn't mention Windows.

kinke:
I don't know what those 2 functions are used for. We don't use them on the compiler side, and druntime only tries to get the exported functions in `rt/dmain2.d` (Windows only), there's no other reference anywhere. So if they were moved to a dedicated module (from `gc.proxy` and/or `gcstub.gc`), that object would never be pulled in by anyone.
Can you shed some light on this?

rainers:
> It's looking as if that's used to make DLLs (with statically linked-in druntime/GC) use the main executable's/prior DLL's GC right after loading the lib via rt_loadLibrary(), by passing the current GC pointer to the new library (and using exported functions in order not to call the executable's gc_setProxy(), but the DLL's).

Yes that's about it. It's an early (read: incomplete and very limited) attempt at using a shared GC. I think it's almost unusable but in very limited cases (single-threaded with leaking initializer memory, no other shared resources including the C runtime if it is statically linked, etc.). I don't think anybody is using this, but just in case it has to stay around longer until full shared runtime is available by Benjamin's work.

> I don't see an alternative to exporting those 2 functions for each binary linked statically against druntime. 

DMD built executables do not export it by default, you have to drag in these symbols by a def-file. So LDC can do the same. I wouldn't even mind if we just exclude these in LDC.

kinke:
Thanks for explaining.

> I wouldn't even mind if we just exclude these in LDC.

Sounds perfectly fine to me too, moving them to their own module isn't worth it IMO. If there really is someone using this with LDC, he/she must have exported them explicitly already and so know about this. Will update the PR accordingly.

rainers:
>>I wouldn't even mind if we just exclude these in LDC.

> Sounds perfectly fine to me too

Not sure if this is what you plan to do: the easiest non-breaking thing to do is to just remove `export` from the two functions. The example in the Wiki also doesn't use the functions being exported: https://wiki.dlang.org/Win32_DLLs_in_D#D_code_calling_D_code_in_DLLs

kinke:
I previously excluded both functions entirely for `version (LDC)`, now just the `export` scope is commented out. So this basically mimicks DMD, the functions are there, but need to be exported manually (well, DMD also exports them if the DLL itself is referencing the functions as they are then pulled in when linking).

rainers:
> It basically boils down to that the export symbol (global/function) is exported if it's part of (defined in or template-instantiated in) a compiled (and not just imported) module and otherwise imported.

IIRC this is indeed what Benjamin proposed. DMD is currently a bit different, though. It writes the export-directive if the function or variable is actually written to the object file, but it assumes it imported if a function has no body or a global variable has no initializer.

rainers:
BTW: I approve the druntime change, feel free to merge. The CircleCI failure should go away if rebased.

gen/llvmhelpers.cpp
```diff
@@ -856,6 +856,13 @@ void DtoResolveVariable(VarDeclaration *vd) {
     // as well).
     gvar->setAlignment(DtoAlignment(vd));
 
+    if (vd->isExport()) {
+      auto c = (vd->storage_class & STCextern) || vd->inNonRoot()
```
rainers: Why not use `vd->isImportedSymbol()` to get DMD compatible behavior? Same for functions?

kinke: I found it pretty weak. A function is 'imported' if it has the `export` visibility and no body. Similarly, a global must have no initializer to be 'imported'. So this relies on special DLL import headers stripped off bodies and initializers, something I wanted to avoid.

gen/functions.cpp
```diff
@@ -917,11 +923,12 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
 
   const auto f = static_cast<TypeFunction *>(fd->type->toBasetype());
   IrFuncTy &irFty = irFunc->irFty;
-  llvm::Function *func = irFunc->func;;
+  llvm::Function *func = irFunc->func;
 
   const auto lwc = lowerFuncLinkage(fd);
   if (linkageAvailableExternally) {
     func->setLinkage(llvm::GlobalValue::AvailableExternallyLinkage);
+    func->setDLLStorageClass(llvm::GlobalValue::DefaultStorageClass);
```
kinke: This is the biggest question mark of this PR - how to handle inlined functions. I'm not sure the current logic removing any DLL storage class works, i.e., if it actually gets used (by the call sites potentially explicitly specifying dllexport/dllimport, I haven't checked that yet). I'm only sure that inlined `export` functions shouldn't be exported. ;) It may make sense not to inline any `export` functions in the first place.

I also don't know if dllexport/dllimport has any effect on non-Windows platforms.

kinke:
druntime patch merged, PR rebased.

gen/llvmhelpers.cpp
rainers: > I found it pretty weak. A function is 'imported' if it has the export visibility and no body. Similarly, a global must have no initializer to be 'imported'.

Sure, it's not great. But I suspect changing semantics as proposed will require some efforts to still allow separate compilation (i.e. using libraries): If two separately compiled modules with one importing the other are to be linked into the same binary, you'll get missing symbols due to __imp_... symbols missing from the import table (that doesn't exist). @Ingrater is currently working on supporting it in dmd. I hope it will be possible to have this in LDC, too.

kinke: I just did a quick experiment:

current.d:
```D
import current2;
int main() { return foo(); }
```

current2.d:
```D
import cc;
int foo() { return bar(); }
```

cc.d:
```D
export int bar() { return 0; }
```

All 3 modules are compiled separately. Then when linking all objects at once, things work, although the MS linker warns about a local import of `bar` in the current2 object.

The linking error you mentioned arises if current2 and cc are packed to a static library, and then linking the current object with that library, as the cc object (and thus exported `bar`) isn't pulled in. :/ As long as both objects in the static lib have been compiled separately, that is.

gen/functions.cpp
kinke: The call sites specify neither dllexport nor dllimport, that's good.

gen/llvmhelpers.cpp
rainers: I guess it'll also happen if bar is called directly from current. Also, no need to use a static library, just using the separately compiled object files should expose the problem, too.

gen/functions.cpp
rainers: Sounds good. Copies of exported functions for inlining should not be exported, it would break ODR.

> Or are exports on non-Windows platforms a totally different thing?

I'm not a non-Windows expert, but AFAICT almost all symbol are currently "exported" on linux. It is handled by the "visibility", though.

gen/llvmhelpers.cpp
rainers: > I guess it'll also happen if bar is called directly from current. Also, no need to use a static library, just using the separately compiled object files should expose the problem, too.

Just tried that with dmd using its export semantic: the linker issues a warning, but seems to automatically match the import symbol to the exported and generate an "import table fake entry":
```
c:\tmp\d\ldc\exp>c:\l\d\dmd.nightly\dmd2\windows\bin\dmd.exe -c -m64 current2.d

c:\tmp\d\ldc\exp>c:\l\d\dmd.nightly\dmd2\windows\bin\dmd.exe -m64 -version=Import current.d current2.obj
   Creating library current.lib and object current.exp
current.obj : warning LNK4217: locally defined symbol _D8current23fooFZi imported in function _Dmain
```

kinke: Yeah right, DMD behaves exactly the same here. I've tested it with `cc.d` modified to:

```D
version (cc_Export)
    export int bar() { return 0; }
else
    export int bar();
```

and specifying that version when compiling `cc.d` separately with DMD.

So this looks pretty ready to me. We're compatible with DMD in the sense that `export`ed `extern` globals and body-less functions are always imported. But unlike DMD, one can get away without special .di headers or version-blocks clutter. When compiling modules separately, linking warnings (performance overhead due to local imports) or even errors (exporting object not pulled in from static lib by a `dllimport` reference) may result, but that's the case for DMD too.

JohanEngelen:
Nice progress.

> The call sites specify neither dllexport nor dllimport, that's good. And cross-module-inlining works nicely for the current-current2-cc example above - main() is reduced to a return 0, bar isn't exported (as long as compiling current.d alone without cc.d - ldc2 -O -release -enable-cross-module-inlining current.d).

Make sure you add tests for these, so it doesn't regress!

kinke:
Rebased and `tests/codegen/export_crossModuleInlining.d` amended.

kinke:
The Travis OSX failure is the unrelated spurious core.thread segfault.

gen/llvmhelpers.cpp
rainers: > errors (exporting object not pulled in from static lib by a dllimport reference) may result, but that's the case for DMD too.

Yes, but errors occur due to different conditions:
- dmd will miss the __imp_var symbol for variables that have no initializer (in the user code facing di file)
- ldc will miss the __imp_var symbol for variable declarations in files that are not compiled together with the file using the symbol

BTW: optlink seems to drag in the respective exported symbol from a library, too.

I don't think this is a blocker, but I'm also unsure LDC should "invent" an intermediate solution when dmd is going to change in the same direction soon, but slightly different again. IIRC the generation of dllimport-indirections for variables will be limited to the compilation with -shared, for example.

Ingrater:
I would strongly advice against implementing this now into ldc. The current state in dmd is so broken that nobody uses it. I could not find a single D project on github which uses export. For more details on how broken export is please see my dconf 2016 talk: https://youtu.be/MQRHxI2SrYM?list=PL3jwVPmk_PRyTWWtTAZyvmjDF4pm6EX6z

If you implement this feature parity with dmd now please also know that this will be in vain, because export will be changed from a visibility level to a attribute once my dll support pull request goes in. Which means that you would have to pretty much do the changes all over again. I can understand that you want to support dlls as much as dmd does but dmds dll support is so bad that you should rather not do that.

kinke:
I went through the presentation yesterday, and that convinced me to hold this back, as there's much more complexity to it than I initially thought. I just wanted to give people something for LDC 1.1, as the export topic keeps coming up.

So should we wait for the proper fix, which may still take a while judging by DMD history, or provide a minimalistic temporary solution for the vast majority of current use cases, which is exporting a handful of callbacks? E.g., just using `dllexport` for all `export` globals and functions with initializer/body, on Windows only, and no `dllimport` at all?

rainers:
I think exporting free functions is fine. If you want to deal with data, you are likely to fail with dmd, too. (IIRC Digital Mars implib doesn't distinguish between functions and data and always generates the indirect jump!).
I would suggest to just use ` vd->isImportedSymbol()` for now. It's compatible with dmd for functions which covers the use case of DLLs that just expose a C interface.

kinke:
Done, see commit msg.

rainers:
Sorry for the delay, will merge...

kinke:
No worries, glad it's in now.

PR #1859, opened 2016-10-27
fix build against current LLVM master @ 9b4e8286785f

rainers:
This will probably be necessary for CircleCI soon.


JohanEngelen:
Nice!

Issue #1844, opened 2016-10-19
Implement `-gline-tables-only`


JohanEngelen:
The flag can help in reducing object file sizes and compile speed, while maintaining debuggable executables that `perf` is also able to use. See e.g. http://clang.llvm.org/docs/UsersManual.html#using-sampling-profilers

Should be easy to implement (iirc it's just passing a true instead of a false to LLVM somewhere in the DIBuilder code).

JohanEngelen:
Implemented in #1861 !

PR #1861, opened 2016-10-27
impl -gline-tables-only

rtbo:
implementation of -gline-tables-only flag per #1844

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -74,22 +74,13 @@ llvm::StringRef uniqueIdent(Type* t) {
 
 } // namespace
 
-bool ldc::DIBuilder::mustEmitAnyDebugInfo() {
-  // for -g -gc and -gline-tables-only
-  return global.params.symdebug > 0;
-}
 
 bool ldc::DIBuilder::mustEmitFullDebugInfo() {
   // only for -g and -gc
   return global.params.symdebug == 1 || global.params.symdebug == 2;
 }
 
-bool ldc::DIBuilder::mustEmitVariablesDebugInfo() {
-  // only for -g and -gc
-  return global.params.symdebug == 1 || global.params.symdebug == 2;
-}
-
-bool ldc::DIBuilder::mustEmitLocationDebugInfo() {
+bool ldc::DIBuilder::mustEmitLocationsDebugInfo() {
   // for -g -gc and -gline-tables-only
   return global.params.symdebug > 0;
 }
@@ -747,7 +738,7 @@ DebugEmissionKind getDebugEmissionKind()
 
 
 void ldc::DIBuilder::EmitCompileUnit(Module *m) {
-  if (!mustEmitLocationDebugInfo()) {
+  if (!mustEmitLocationsDebugInfo()) {
     return;
   }
 
@@ -792,7 +783,7 @@ void ldc::DIBuilder::EmitCompileUnit(Module *m) {
 }
 
 ldc::DISubprogram ldc::DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
-  if (!mustEmitLocationDebugInfo()) {
+  if (!mustEmitLocationsDebugInfo()) {
 #if LDC_LLVM_VER >= 307
     return nullptr;
 #else
@@ -840,7 +831,7 @@ ldc::DISubprogram ldc::DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
 
 ldc::DISubprogram ldc::DIBuilder::EmitThunk(llvm::Function *Thunk,
                                             FuncDeclaration *fd) {
-  if (!mustEmitLocationDebugInfo()) {
+  if (!mustEmitLocationsDebugInfo()) {
 #if LDC_LLVM_VER >= 307
     return nullptr;
 #else
@@ -889,7 +880,7 @@ ldc::DISubprogram ldc::DIBuilder::EmitThunk(llvm::Function *Thunk,
 
 ldc::DISubprogram ldc::DIBuilder::EmitModuleCTor(llvm::Function *Fn,
                                                  llvm::StringRef prettyname) {
-  if (!mustEmitLocationDebugInfo()) {
+  if (!mustEmitLocationsDebugInfo()) {
 #if LDC_LLVM_VER >= 307
     return nullptr;
 #else
@@ -913,10 +904,9 @@ ldc::DISubprogram ldc::DIBuilder::EmitModuleCTor(llvm::Function *Fn,
   auto paramsArray = DBuilder.getOrCreateArray(params);
 #endif
 #if LDC_LLVM_VER >= 308
-  ldc::DISubroutineType DIFnType = DBuilder.createSubroutineType(paramsArray);
+   ldc::DISubroutineType DIFnType = DBuilder.createSubroutineType(paramsArray);
 #else
-  ldc::DISubroutineType DIFnType =
-      DBuilder.createSubroutineType(file, paramsArray);
+   ldc::DISubroutineType DIFnType = DBuilder.createSubroutineType(file, paramsArray);
 #endif
 
   // FIXME: duplicates?
@@ -944,7 +934,7 @@ ldc::DISubprogram ldc::DIBuilder::EmitModuleCTor(llvm::Function *Fn,
 }
 
 void ldc::DIBuilder::EmitFuncStart(FuncDeclaration *fd) {
-  if (!mustEmitLocationDebugInfo())
+  if (!mustEmitLocationsDebugInfo())
     return;
 
   Logger::println("D to dwarf funcstart");
@@ -955,7 +945,7 @@ void ldc::DIBuilder::EmitFuncStart(FuncDeclaration *fd) {
 }
 
 void ldc::DIBuilder::EmitFuncEnd(FuncDeclaration *fd) {
-  if (!mustEmitLocationDebugInfo())
+  if (!mustEmitLocationsDebugInfo())
     return;
 
   Logger::println("D to dwarf funcend");
@@ -966,7 +956,7 @@ void ldc::DIBuilder::EmitFuncEnd(FuncDeclaration *fd) {
 }
 
 void ldc::DIBuilder::EmitBlockStart(Loc &loc) {
-  if (!mustEmitLocationDebugInfo())
+  if (!mustEmitLocationsDebugInfo())
     return;
 
   Logger::println("D to dwarf block start");
@@ -987,7 +977,7 @@ void ldc::DIBuilder::EmitBlockStart(Loc &loc) {
 }
 
 void ldc::DIBuilder::EmitBlockEnd() {
-  if (!mustEmitLocationDebugInfo())
+  if (!mustEmitLocationsDebugInfo())
     return;
 
   Logger::println("D to dwarf block end");
@@ -999,7 +989,7 @@ void ldc::DIBuilder::EmitBlockEnd() {
 }
 
 void ldc::DIBuilder::EmitStopPoint(Loc &loc) {
-  if (!mustEmitLocationDebugInfo())
+  if (!mustEmitLocationsDebugInfo())
     return;
 
   // If we already have a location set and the current loc is invalid
@@ -1037,7 +1027,7 @@ void ldc::DIBuilder::EmitValue(llvm::Value *val, VarDeclaration *vd) {
     return;
 
   ldc::DILocalVariable debugVariable = sub->second;
-  if (!mustEmitVariablesDebugInfo() || !debugVariable)
+  if (!mustEmitFullDebugInfo() || !debugVariable)
     return;
 
   llvm::Instruction *instr =
@@ -1061,7 +1051,7 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
                                        llvm::ArrayRef<llvm::Value *> addr
 #endif
                                        ) {
-  if (!mustEmitVariablesDebugInfo())
+  if (!mustEmitFullDebugInfo())
     return;
 
   Logger::println("D to dwarf local variable");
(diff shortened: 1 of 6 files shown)


LemonBoy:
The tests are failing due to this [llvm change](https://reviews.llvm.org/D18612).
Using `auto` would help the readability.

gen/dibuilder.cpp
```diff
@@ -697,6 +697,37 @@ ldc::DIType ldc::DIBuilder::CreateTypeDescription(Type *type, bool derefclass) {
...
+      return llvm::DICompileUnit::LineTablesOnly;
+    case 2:
+    case 3:
+      return llvm::DICompileUnit::FullDebug;
+    default:
+      break;
+  }
+  llvm_unreachable("unknown DebugEmissionKind");
+#else
+  return global.params.symdebug < 2 ?
```
LemonBoy: This makes the level 0 emit the line informations too. As you can see [here](https://reviews.llvm.org/D18808) it should be possible to have the same effect by setting `EmitDebugInfo` to false.

JohanEngelen: Level 0 should not output any debug info, also not line numbers, so that needs to be fixed.

JohanEngelen:
We have to be careful if we are changing the meaning of the (integer) value of `global.params.symdebug`, which is a parameter possibly used by the DDMD FE.

Have you thought a little about what the behavior should be of cmdline parameters and orderings like this?
`-g -gline-tables-only`
`-g -gline-tables-only -g`
`-gline-tables-only -g`
`-gline-tables-only -gc`
`-gline-tables-only -gc -gline-tables-only`

rtbo:
In regards to cmdline parameters ordering it is the latest of the `-g*` params that prevail.
No change from previous behavior when there was only `-g` and `-gc`.
A warning or error would be more appropriate I guess in case of multiple `-g*` params given.

rtbo:
there are still a few lines of code like this in `driver/linker.cpp`:
```d
  // because of a LLVM bug, see LDC issue 442
  if (global.params.symdebug) {
    args.push_back("/LARGEADDRESSAWARE:NO");
  } else {
    args.push_back("/LARGEADDRESSAWARE");
  }

  // output debug information
  if (global.params.symdebug) {
    args.push_back("/DEBUG");
  }
```
Not sure which args must be passed to the linker with `-gline-tables-only`.

JohanEngelen:
> A warning or error would be more appropriate I guess in case of multiple -g* params given.

I think it can be helpful to allow multiple `-g` flags without complaining about it. I just wanted to check that you had thought about what the behavior should be (in contrast to what it is at the moment). I think Clang does the same as what you are doing btw, so should be fine.

This still needs a number of tests!

rtbo:
CircleCI fails due to this [change in llvm](https://reviews.llvm.org/D25073).
Unrelated to `-gline-tables-only`.

rtbo:
> This still needs a number of tests!

Tested with `llvm-3.6.2` and `llvm-3.9.1`, this simple test do not pass
```d
// RUN: %ldc -gline-tables-only --output-ll -of%t.ll %s && FileCheck %s < %t.ll
// Checks that ldc with -gline-tables-only do not emit debug info

int main()
{
    immutable int fact=6;
    int res=1;
    for(int i=1; i<=fact; i++) res *= i;
    return res;
}

// CHECK-NOT: DW_TAG
```

Also tested the 2 following tests, adapted from clang, none of them pass:
```d
// RUN: %ldc -gline-tables-only --output-ll -of%t.ll %s && FileCheck %s < %t.ll
// Checks that ldc with "-gline-tables-only" emits metadata for
// compile unit, subprogram and file.

int main() {
  // CHECK: ret i32 0, !dbg
  return 0;
}

// CHECK: !llvm.dbg.cu = !{!0}
// CHECK: !DICompileUnit(
// CHECK: !DISubprogram(
// CHECK: !DIFile(
```
```d
// RUN: %ldc -gline-tables-only --output-ll -c %s -of%t.ll && FileCheck %s < %t.ll
// Checks that ldc with "-gline-tables-only" doesn't emit debug info
// for variables and types.

// CHECK-NOT: DW_TAG_variable
int global = 42;

// CHECK-NOT: DW_TAG_typedef
// CHECK-NOT: DW_TAG_const_type
// CHECK-NOT: DW_TAG_pointer_type
// CHECK-NOT: DW_TAG_array_type
alias constCharPtrArray = const(char)*[10];

// CHECK-NOT: DW_TAG_structure_type
struct S {
  // CHECK-NOT: DW_TAG_member
  char a;
  double b;
  constCharPtrArray c;
}

// CHECK-NOT: DW_TAG_enumerator
// CHECK-NOT: DW_TAG_enumeration_type
enum E { ZERO = 0, ONE = 1 }

// CHECK-NOT: DILocalVariable
int sum(int p, int q) {
  int r = p + q;
  S s;
  E e;
  return r;
}
```

JohanEngelen:
Need help with those tests @rtbo ?

rtbo:
> Need help with those tests @rtbo ?

I've figured out for the dwarf tags. `tests/debuginfo/gline_tables_only2.d` is still failing the same.

rtbo:
Oops, cadbac08a0452d0b16e4a6b40d0e767ac790689c also strips almost all metadata, which comes back to a release build.

rtbo:
Better now. 
Still have noise dwarf emission for functions return types (makes `debuginfo/gline_tables_only2.d` fail).

rtbo:
Passes with llvm-3.9 only

rtbo:
Reason of test failure is that emitted metadata in IR code is far different form one version of LLVM to another.
Here is metadata emitted with llvm-3.9.1:
```llvm
!llvm.module.flags = !{!0, !1}
!llvm.dbg.cu = !{!3}
!llvm.ident = !{!5}

!0 = !{i32 2, !"Debug Info Version", i32 3}
!1 = !{i32 6, !"Linker Options", !2}
!2 = !{}
!3 = distinct !DICompileUnit(language: DW_LANG_D, file: !4, producer: "LDC (http://wiki.dlang.org/LDC)", isOptimized: false, runtimeVersion: 1, emissionKind: LineTablesOnly, enums: !2)
!4 = !DIFile(filename: "factorial.d", directory: "/opt/dev/test/factorial")
!5 = !{!"ldc version e9a486"}
!6 = distinct !DISubprogram(name: "D main", linkageName: "_Dmain", scope: null, file: !4, line: 2, type: !7, isLocal: false, isDefinition: true, scopeLine: 2, flags: DIFlagPrototyped, isOptimized: false, unit: !3, variables: !2)
!7 = !DISubroutineType(types: !2)
!8 = !DILocation(line: 4, column: 5, scope: !6)
!9 = !DILocation(line: 5, column: 5, scope: !6)
!10 = !DILocation(line: 6, column: 9, scope: !11)
!11 = distinct !DILexicalBlock(scope: !6, file: !4, line: 6, column: 5)
!12 = !DILocation(line: 6, column: 5, scope: !13)
!13 = distinct !DILexicalBlock(scope: !11, file: !4, line: 6, column: 5)
!14 = !DILocation(line: 7, column: 9, scope: !13)
!15 = !DILocation(line: 8, column: 5, scope: !6)
!16 = distinct !DISubprogram(name: "__entrypoint.main", linkageName: "main", scope: null, file: !17, line: 6, type: !7, isLocal: false, isDefinition: true, scopeLine: 6, flags: DIFlagPrototyped, isOptimized: false, unit: !3, variables: !2)
!17 = !DIFile(filename: "__entrypoint.d", directory: "/opt/dev/test/factorial")
!18 = !DILocation(line: 6, column: 15, scope: !16)
!19 = !DILocation(line: 8, column: 15, scope: !16)
```
and here for same code with llvm-3.6.2:
```llvm
!llvm.module.flags = !{!0, !1}
!llvm.ident = !{!3}

!0 = !{i32 2, !"Debug Info Version", i32 2}
!1 = !{i32 6, !"Linker Options", !2}
!2 = !{}
!3 = !{!"ldc version e9a486"}
!4 = !MDLocation(line: 4, column: 5, scope: !5)
!5 = !{!"0x2e\00D main\00D main\00_Dmain\002\000\001\000\000\00256\000\002", !6, null, !7, null, i32 ({ i64, { i64, i8* }* })* @_Dmain, null, null, !2} ; [ DW_TAG_subprogram ] [line 2] [def] [D main]
!6 = !{!"factorial.d", !"/opt/dev/test/factorial"}
!7 = !{!"0x15\00\000\000\000\000\000\000", null, null, null, !2, null, null, null} ; [ DW_TAG_subroutine_type ] [line 0, size 0, align 0, offset 0] [from ]
!8 = !MDLocation(line: 5, column: 5, scope: !5)
!9 = !MDLocation(line: 6, column: 9, scope: !10)
!10 = !{!"0xb\006\005\000", !6, !5}               ; [ DW_TAG_lexical_block ] [/opt/dev/test/factorial/factorial.d]
!11 = !MDLocation(line: 6, column: 5, scope: !12)
!12 = !{!"0xb\006\005\001", !6, !10}              ; [ DW_TAG_lexical_block ] [/opt/dev/test/factorial/factorial.d]
!13 = !MDLocation(line: 7, column: 9, scope: !12)
!14 = !MDLocation(line: 8, column: 5, scope: !5)
!15 = !MDLocation(line: 6, column: 15, scope: !16)
!16 = !{!"0x2e\00__entrypoint.main\00__entrypoint.main\00main\006\000\001\000\000\00256\000\006", !17, null, !7, null, i32 (i32, i8**)* @main, null, null, !2} ; [ DW_TAG_subprogram ] [line 6] [def] [__entrypoint.main]
!17 = !{!"__entrypoint.d", !"/opt/dev/test/factorial"}
!18 = !MDLocation(line: 8, column: 15, scope: !16)
```
I limited the tests to llvm >= 3.9 at the moment. (Shortest path, I admit).
Is there a good reason to make more tests for other llvm versions? My point is that a regression will be seen in the llvm-3.9 tests anyway.
And are the tests for 3.9 enough in the first place?

JohanEngelen:
I think testing for >=3.9 is fine.

gen/dibuilder.cpp
```diff
@@ -699,8 +721,39 @@ ldc::DIType ldc::DIBuilder::CreateTypeDescription(Type *type, bool derefclass) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+#if LDC_LLVM_VER >= 309
+typedef llvm::DICompileUnit::DebugEmissionKind DebugEmissionKind;
+#else
+typedef llvm::DIBuilder::DebugEmissionKind DebugEmissionKind;
+#endif
+
+DebugEmissionKind getDebugEmissionKind()
```
kinke: You may be able to get rid of the typedef (`using` is preferred btw) when using `auto` as return type here.

gen/dibuilder.cpp
```diff
@@ -724,19 +777,28 @@ void ldc::DIBuilder::EmitCompileUnit(Module *m) {
   IR->module.addModuleFlag(llvm::Module::Warning, "Debug Info Version",
                            llvm::DEBUG_METADATA_VERSION);
 
+
```
kinke: Hm?

gen/dibuilder.cpp
```diff
@@ -857,10 +920,9 @@ ldc::DISubprogram ldc::DIBuilder::EmitModuleCTor(llvm::Function *Fn,
   auto paramsArray = DBuilder.getOrCreateArray(params);
 #endif
 #if LDC_LLVM_VER >= 308
-  ldc::DISubroutineType DIFnType = DBuilder.createSubroutineType(paramsArray);
+   ldc::DISubroutineType DIFnType = DBuilder.createSubroutineType(paramsArray);
```
kinke: [My] clang-format disagrees.

tests/debuginfo/gline_tables_only2.d
```diff
@@ -0,0 +1,18 @@
...
+int main() {
+  // CHECK: ret i32 0, !dbg
+  return 0;
+}
+
+// CHECK: !llvm.dbg.cu = !{
+// CHECK: !DICompileUnit(
+// CHECK: !DISubprogram(
+// CHECK: !DIFile(
+// CHECK-NOT: DW_AT
```
kinke: Missing trailing newline (I know I'm pedantic). Are you aware that the non-existence of DW_AT is restricted to all lines following the previous `!DIFile(` match?

gen/dibuilder.cpp
rtbo: `auto` return type needs a tail expression, which makes code messy with the `switch`.
`using` is used in f65b73ce098ab5e10628c81a5bfbb0ec0e696be1.

gen/dibuilder.cpp
rtbo: Testing code vestige. Correction sneaked in 3761775fe3427137e5e2881e1b390fba83d5e4b7.

gen/dibuilder.cpp
rtbo: bd35c9600dcfd551ebc913aa37bf4325163d9931

gen/dibuilder.cpp
```diff
@@ -857,10 +920,9 @@ ldc::DISubprogram ldc::DIBuilder::EmitModuleCTor(llvm::Function *Fn,
...
 #else
-  ldc::DISubroutineType DIFnType =
-      DBuilder.createSubroutineType(file, paramsArray);
+   ldc::DISubroutineType DIFnType = DBuilder.createSubroutineType(file, paramsArray);
```
rtbo: bd35c9600dcfd551ebc913aa37bf4325163d9931

gen/dibuilder.cpp
```diff
@@ -1174,7 +1236,7 @@ void ldc::DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *llVar,
 }
 
 void ldc::DIBuilder::Finalize() {
-  if (!mustEmitDebugInfo())
+  if (!mustEmitFullDebugInfo() && !mustEmitLocationsDebugInfo())
```
rtbo: Intent was to make obvious that Finalize must be called for all debug info cases. But yes, it comes to the same. ba5002a074ee6341eee6c94a8a37e2f5deecbfc4

tests/debuginfo/gline_tables_only2.d
rtbo: No, I was not aware. Thanks for pointing it out. 
9c122f2ba3258600025d2561b8e2cf29580b0b34.

JohanEngelen:
Great stuff.
Nice work on the tests.

(squashed and merged)

Issue #1860, opened 2016-10-27
ICE on Android/ARM, related to recent LTO change in master


joakim-noah:
I've been building natively on Android/ARM with the latest llvm 3.9.0 release, I ran into this ICE when building druntime synced against latest master:

```
LLVM ERROR: Cannot select: 0xb592b708: f64 = bitcast 0xb592b3f0
  0xb592b3f0: i64 = build_pair 0xb592b090, 0xb592b5e8
    0xb592b090: i32,ch = CopyFromReg 0xb597de18, Register:i32 %vreg16
      0xb592b1b0: i32 = Register %vreg16
    0xb592b5e8: i32,ch = CopyFromReg 0xb592b090:1, Register:i32 %vreg17
      0xb592b0d8: i32 = Register %vreg17
In function: _d_array_init_double
```

I bisected and found that the issue is #1811, reverting that patch has master passing all stdlib and dmd tests again (there's an unrelated failure in the dmd testsuite because of the recently added test16225, which adds the `-m64` flag and doesn't work because I don't compile in the ARM64 backend on ARM32).

This is the command run on my Android tablet, same as on any other platform:

```
/data/data/com.termux/files/home/ldc/build71/bin/ldc2 --output-o -c -I/data/data/com.termux/files/home/ldc/runtime/druntime/src -I/data/data/com.termux/files/home/ldc/runtime/druntime/src/gc /data/data/com.termux/files/home/ldc/runtime/druntime/src/ldc/arrayinit.d -of/data/data/com.termux/files/home/ldc/build71/runtime/src/ldc/arrayinit-unittest.o -w -O3 -release -unittest -disable-invariants
```

You should be able to reproduce by using an ldc compiled against llvm with an ARM backend and adding the flag `-mtriple=armv7-none-linux-android` or `-mtriple=armv7-unknown-linux-gnueabi`.  If not, let me know.

JohanEngelen:
What if you comment-out setting the "use-soft-float" attribute? It wasn't clear how to set it for the SoftFP ABI type. Or perhaps things work when you add:
```diff
  case llvm::FloatABI::Default:
    break;
  case llvm::FloatABI::Soft:
+  case llvm::FloatABI::SoftFP:
     func.addFnAttr("use-soft-float", "true");
     break;
```

joakim-noah:
Yeah, commenting that line out fixes it and gets all stdlib tests passing again.

JohanEngelen:
[...]

PR #1863, opened 2016-10-28
Set `+soft-float` feature instead of `use-soft-float` attribute.

JohanEngelen:
Fix issue #1860

UseSoftFloat was removed from the TargetOptions in LLVM3.7. It's replacement (I believe) is the "+soft-float" target feature (in the target feature string). Inside LLVM, the "use-soft-float" attribute is upgraded to the "+soft-float" feature, so we can do that ourselves when creating the target machine.

(note that at the point where we add the attributes to the function, we have no information whether our TargetMachine was created with FloatABI:SoftFP or not)

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -598,6 +598,8 @@ createTargetMachine(std::string targetTriple, std::string arch, std::string cpu,
   case FloatABI::Soft:
 #if LDC_LLVM_VER < 307
     targetOptions.UseSoftFloat = true;
+#else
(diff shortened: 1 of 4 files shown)


joakim-noah:
This patch fixes the issue on Android/ARM: druntime compiles again and the full stdlib and dmd test suites pass.

lyrachord:
@see #2406

JohanEngelen:
Related: https://reviews.llvm.org/D8771

PR #1865, opened 2016-11-01
Add code for logging within D code.

JohanEngelen:
This adds the D counterpart of logger.{h,cpp}.

(Not currently used in D source, but I am already using it in my ccaching branch.)

### DIFF
--- a/gen/logger.cpp
+++ b/gen/logger.cpp
@@ -42,29 +42,32 @@ void Stream::writeValue(std::ostream &OS, const llvm::Value &V) {
   }
 }
 
+// This variable is pulled out of the Logger namespace to work around D bug 15576.
+// https://issues.dlang.org/show_bug.cgi?id=15576
+bool _Logger_enabled;
+
 namespace Logger {
 static std::string indent_str;
-bool _enabled;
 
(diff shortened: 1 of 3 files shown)


JohanEngelen:
Could someone help me with the Windows linker error? 
Is the code wrong (`extern (C++, Logger) extern __gshared bool _enabled;`) or is it incorrectly mangled by the compiler to match the C++ definition (`namespace Logger { bool _enabled; }`) ?
Thanks!

kinke:
I'm currently at work, but a mangling issue seems possible - the `static` storage class seems off in this case.

JohanEngelen:
I too think the mangling is off. __gshared is needed (otherwise things crash on OS X at least), it's the default on C's side right (non thread-local)?

OK, so I need a work around :)  `-mtriple` to the rescue!

JohanEngelen:
Bug found: https://issues.dlang.org/show_bug.cgi?id=15576

JohanEngelen:
I've worked around the D compiler bug with extern(C++, namespace) by pulling Logger::_enabled out of the Logger namespace.

kinke:
https://github.com/dlang/dmd/pull/6233

PR #1868, opened 2016-11-04
Win64: Get rid of 2 obsolete fields in TypeInfo_Struct

kinke:
The required changes to druntime turned out a bit more extensive than I had hoped for at the beginning, but since the job is already done, I thought I'd just open a PR.

### DIFF
--- a/gen/typinf.cpp
+++ b/gen/typinf.cpp
@@ -309,12 +309,13 @@ public:
     const auto structTypeInfoType = getStructTypeInfoType();
     const auto structTypeInfoDecl = Type::typeinfostruct;
 
-    // On x86_64, class TypeInfo_Struct contains 2 additional fields
-    // (m_arg1/m_arg2) which are used for the X86_64 System V ABI varargs
-    // implementation. They are not present on any other cpu/os.
-    const bool isX86_64 =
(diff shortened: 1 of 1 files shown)


gen/typinf.cpp
```diff
@@ -515,7 +514,7 @@ class LLVMDefineVisitor : public Visitor {
     // uint m_align;
     b.push_uint(DtoAlignment(tc));
 
-    if (global.params.is64bit) {
+    if (isSystemVx64) {
```
kinke: I'm rather suprised this hasn't bugged anyone until now. In druntime, the fields were only there for `version (X86_64)`, so this must have been broken for all other 64-bit targets.

kinke:
@rainers: Do you think it'd make sense to upstream the druntime portion?

rainers:
> @rainers: Do you think it'd make sense to upstream the druntime portion?

Took a while until I found the actual removed members because they don't show up in the diffs.IMO it's unlikely worth the trouble to change upstream dmd and druntime. Unsure about removing them in LDC as it just increases the differences...

kinke:
This has happened as of LDC v1.22.

Issue #1843, opened 2016-10-19
Forced lowercase outputfile extension


JohanEngelen:
`bin/ldc2 -c -output-ll test.d -of=tesT.LL`
outputs the file `testT.ll` on my system (OS X). Note the lowercase `ll`. (same for -output-s)

rtbo:
It is not only lowercase, the extension is forced to `ll` or `s` whatever the extension provided on the command line.
I believe it has something to do with [this line](https://github.com/ldc-developers/ldc/blob/master/ddmd/dmodule.d#L396).

kinke:
Nope that line isn't that important. The thing is that we have a single filename per object (-of or inferred) while we may output multiple files per module (.o/.bc/.ll/.s), that's why the other extensions are currently enforced. One potential fix may consist in using the filename directly if a single file is produced (but then the inferred filename needs to be adapted based on what single file is emitted, as currently the default object extension is used IIRC). It's a bit messy.

kinke:
Fixed by https://github.com/ldc-developers/ldc/pull/1869.

PR #1840, opened 2016-10-19
Add LTO support (full and thin), with `-flto=thin|full`.

JohanEngelen:
LTO needs linker support: I am only aware of support on OS X and Linux (through the LLVMgold plugin).

I chose the cmdline option `flto=thin|full` that mimics Clang's option (also Martin was looking for `flto` in the issue #693 he reported).

Resolves #693 .

For OS X, additional logic could be added later to point to a newer libLTO.dylib (see clang's `darwin::Linker::AddLinkArgs`).

PR #1869, opened 2016-11-05
Don't necessarily enforce file extension of .bc, .ll, .s files

kinke:
Use the output filename (-of or inferred) directly if a single file is emitted per compiled module.

Fixes issue #1843.

### DIFF
--- a/driver/toobj.cpp
+++ b/driver/toobj.cpp
@@ -347,11 +347,11 @@ public:
   }
 };
 
-void writeObjectFile(llvm::Module *m, std::string &filename) {
-  IF_LOG Logger::println("Writing object file to: %s", filename.c_str());
+void writeObjectFile(llvm::Module *m, const char *filename) {
+  IF_LOG Logger::println("Writing object file to: %s", filename);
   LLErrorInfo errinfo;
   {
-    llvm::raw_fd_ostream out(filename.c_str(), errinfo, llvm::sys::fs::F_None);
+    llvm::raw_fd_ostream out(filename, errinfo, llvm::sys::fs::F_None);
 #if LDC_LLVM_VER >= 306
     if (!errinfo)
 #else
@@ -361,7 +361,7 @@ void writeObjectFile(llvm::Module *m, std::string &filename) {
       codegenModule(*gTargetMachine, *m, out,
                     llvm::TargetMachine::CGFT_ObjectFile);
     } else {
-      error(Loc(), "cannot write object file '%s': %s", filename.c_str(),
+      error(Loc(), "cannot write object file '%s': %s", filename,
             ERRORINFO_STRING(errinfo));
       fatal();
     }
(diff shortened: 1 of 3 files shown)


thewilsonator:
IIUC you mean if more than one of `-output-{bc,ll,s,o}` is specified them still force the extensions, but not for object files (I need different extensions for object files, don't change this)?
And you mean llvm modules, not d modules,yes? 
I'm just trying to think how this will interoperate with my dcompute PRs which will have multiple files being outputted, but they will each come from one distinct llvm module.

Anyway if this is to be merged please also merge `ldc-developers:master` into `ldc-developers:dcompute`, so that I can test properly.

thewilsonator:
This will have merge conflicts with #1840

kinke:
> IIUC you mean if more than one of -output-{bc,ll,s,o} is specified them still force the extensions, but not for object files

Exactly, as the diff clearly shows. Object filenames aren't touched, only bc/ll/s.

> And you mean llvm modules, not d modules,yes?

Yep, i.e., a single or multiple (-singleobj) D modules compiled to an 'object' (incl. bc/ll/s).

In essence, this only affects the output filename of a bc/ll/s file if (1) a single `-output-{bc,ll,s}` flag is used, (2) `-output-o` isn't used (also preventing the generation of a static lib/executable/shared lib, as `-of` would apply to that name then) and (3) `-of` is specified (implying -singleobj), i.e., a single 'object' file with that name is to be produced. E.g., `ldc2 -output-ll -of=bla.IR foo.d [bar.d]`.

JohanEngelen:
Is there a way we can test this?
Perhaps `%ldc ... -output-s -of=%T/test.A  &&  %remove %T/test.A`, relying on failure of `%remove` when the file is not found?

kinke:
I added a tiny test, but opted for a textual IR file and an existence check via FileCheck.

driver/toobj.cpp
```diff
@@ -440,11 +437,23 @@ void writeModule(llvm::Module *m, std::string filename) {
...
+    llvm::sys::path::replace_extension(buffer, ext);
+    return buffer.str();
+  };
+
   // write LLVM bitcode
   if (global.params.output_bc || (doLTO && outputObj)) {
-    LLPath bcpath(filename);
-    if (global.params.output_bc)
-      llvm::sys::path::replace_extension(bcpath, global.bc_ext);
+    std::string bcpath = doLTO ? filename : replaceExtensionWith(global.bc_ext);
```
kinke: @JohanEngelen: I guess this fixes a potential issue wrt. `-flto=<something> -output-o -output-bc`, as I guess the regular object filename is required for the bitcode file in the LTO case.

driver/toobj.cpp
```diff
@@ -440,11 +437,24 @@ void writeModule(llvm::Module *m, std::string filename) {
...
+    return buffer.str();
+  };
+
   // write LLVM bitcode
   if (global.params.output_bc || (doLTO && outputObj)) {
-    LLPath bcpath(filename);
-    if (global.params.output_bc)
-      llvm::sys::path::replace_extension(bcpath, global.bc_ext);
+    std::string bcpath =
+        (doLTO && outputObj) ? filename : replaceExtensionWith(global.bc_ext);
```
kinke: @JohanEngelen: I guess this fixes a potential issue wrt. `-flto=<something> -output-o -output-bc`, as I guess the regular object filename is required for the bitcode file in the LTO case.

kinke: I think it's okay not outputting the redundant .bc copy in the LTO case for this exotic cmdline-flags constellation.

JohanEngelen:
Cool.

PR #1871, opened 2016-11-08
Add ccache-like source-level caching.

JohanEngelen:
This required a bunch of front-end changes. Shall I put all of it in IN_LLVM version blocks?

The caching is conservative. For example, it is only enabled for builds that output a single object file, and the order of the commandlines should not have changed (`-c -g` and `-g -c` get different cache entries). I think this is a very common use case. Other use cases can be added later.

Depends/includes https://github.com/ldc-developers/ldc/pull/1865


driver/cache.d
```diff
@@ -0,0 +1,326 @@
...
+}
+
+struct CachingState
+{
+    const(char)* dateUsed; // the date string used by calculateModulesHash.
+    const(char)* timeUsed; // the time string used by calculateModulesHash.
+}
+CachingState state;
+
+// Checks whether caching is still valid after doing semantic analysis. This
```
thewilsonator: Did you forget to finish writing a more descriptive comment?

JohanEngelen: thanks

JohanEngelen:
For testing, I added the %copy and %remove substitutions to our lit config. On Windows, what are the commands that we can use for those? `copy` gives the error: `command not found`.

kinke:
Both `copy` and `del` on Windows aren't stand-alone executables, but shell intrinsics => `cmd.exe /c copy ...`.

JohanEngelen:
@kinke excellent, thanks

kinke:
Wow, it doesn't seem to bother checking in %PATH%. The full path is `C:\Windows\System32\cmd.exe`.

JohanEngelen:
@kinke I'm surprised too. I'll fire-up my Windows machine and try to figure out what works.

JohanEngelen:
After fiddling with it for too long, I decided to simply add two test utilities `copyfile` and `removefile` that do exactly what is needed without hassle.

JohanEngelen:
OK, green now. Thanks for reviewing and comment on the front-end diff (#ifdef?)

tests/caching/source_caching_nonexistant.d
```diff
@@ -0,0 +1,26 @@
+// Test that caching works correctly if new files appear compared to the cached build (new files appear earlier in the search path).
+
+// The first few RUN commands make sure that the testrun starts with a clean slate:
+// 1. An empty cache.
+// 2. testdir/exist.d does not exist.
+
+// RUN: %ldc -c -cache=%T/exist1 %s -I%T -I%S/inputs/nonexist -of=%t%obj \
+// RUN: && %prunecache -f --max-bytes=1 %T/exist1  \
+// RUN: && removefile %T/exist.d \
```
klickverbot: `cmake -E …` should be useful for this as well.

PR #1872, opened 2016-11-09
Implement the betterC switch like DMD does.

LemonBoy:
Namely, don't emit the ModuleInfo.
Additionally, for the following code we emit less symbols than dmd (`28` vs `38`) and the resulting executable actually runs as we don't emit any `dso` section (which explains the former) and no code that fondles with `_d_dso_registry` (which explains the latter).

```D                     
module main;

extern(C) __gshared int __dmd_personality_v0;
extern(C) __gshared void* _d_dso_registry;

extern (C) int printf(in char*, ...);

extern (C) int main()
{
    printf("asd\n");
    return 0;
}
```

### DIFF
--- a/tests/linking/betterc.d
+++ b/tests/linking/betterc.d
@@ -1,0 +1,15 @@
+// RUN: %ldc -betterC -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+// RUN: %ldc -betterC -defaultlib= -run %s
+
+// CHECK-NOT: ModuleInfoZ
+// CHECK-NOT: ModuleRefZ
+// CHECK-NOT: call void @ldc.register_dso
+version (CRuntime_Microsoft) {
+    extern(C) int mainCRTStartup() {
+        return 0;
+    }
+}
+
+extern (C) int main() {
+    return 0;
+}
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -437,6 +437,11 @@ cl::opt<bool, true>
              cl::desc("implement http://wiki.dlang.org/DIP25 (experimental)"),
              cl::location(global.params.useDIP25));
 
+cl::opt<bool, true> betterC(
+    "betterC",
+    cl::desc("omit generating some runtime information and helper functions"),
+    cl::location(global.params.betterC));
+
 cl::opt<unsigned char, true, CoverageParser> coverageAnalysis(
     "cov", cl::desc("Compile-in code coverage analysis\n(use -cov=n for n% "
                     "minimum required coverage)"),
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -722,8 +722,8 @@ void codegenModule(IRState *irs, Module *m) {
   }
 
   // Skip emission of all the additional module metadata if requested by the
-  // user.
-  if (!m->noModuleInfo) {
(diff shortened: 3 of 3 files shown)


JohanEngelen:
No tests!

Can you try to include a test that shows that the resulting object file can be linked _without_ druntime?

9il:
Other issues:
1. switch errors
2. asserts
3. cast between arrays
4. Type info
5. Check that DRuntime is 100% does not required by generated code.

LemonBoy:
IR + Linking test has been attached.
@9il This PR implements `betterC` a'la dmd, it's not an effort to have a better `betterC` :) See [here](https://issues.dlang.org/show_bug.cgi?id=11881) for more details.

9il:
> @9il This PR implements betterC a'la dmd, it's not an effort to have a better betterC :) See here for more details.

OK )

kinke:
@9il: Is it just that you'd like `betterC` to imply `-release`, getting rid of all assertions, bounds checks, contracts etc.?

tests/linking/betterc.d
```diff
@@ -0,0 +1,13 @@
+// RUN: %ldc -betterC -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+// RUN: %ldc -betterC -defaultlib= -run %s
+
+// CHECK-NOT: ModuleInfoZ
+// CHECK-NOT: ModuleRefZ
+// CHECK-NOT: call void @ldc.register_dso
+extern(C) __gshared int __dmd_personality_v0;
```
klickverbot: Why would we require __dmd_personality_v0 in the first place?

LemonBoy: That's very bare metal we're running on then heh
Adding a stub declaration in a `version(CRuntime_Microsoft)` would do the trick, no ?

kinke: Yep, good idea.

tests/linking/betterc.d
```diff
@@ -0,0 +1,16 @@
...
+// RUN: %ldc -betterC -defaultlib= -run %s
+
+// CHECK-NOT: ModuleInfoZ
+// CHECK-NOT: ModuleRefZ
+// CHECK-NOT: call void @ldc.register_dso
+version (CRuntime_Microsoft) {
+  extern(C) void mainCRTStartup() { }
+}
+
+extern(C) __gshared void* _d_dso_registry;
```
JohanEngelen: Why is this needed? Doesn't that mean that more code emission needs to be disabled with `-betterC`?

9il:
> @9il: Is it just that you'd like -betterC to imply -release, getting rid of all assertions, bounds checks, contracts etc.?

betterC (in my vision) should verify that compiled code does not depend on Druntime. A code can still use some generic code from Phobos. assertions in non release mode should be replaced with inlined checks that use libc to exit from the program.

kinke:
Please document the new switch in [LDMD's usage output](https://github.com/ldc-developers/ldc/blob/master/driver/ldmd.cpp#L156) too.

kinke:
I amended the LDMD addition and pushed it directly to master (https://github.com/ldc-developers/ldc/commit/62a22520282a7f45980a1b959f1958771cdc062d), so that it can be part of beta 4.

PR #1875, opened 2016-11-12
[dcompute] toobj and optimiser

thewilsonator:
* dont optimise SPIR-V as turning GEPs into extracts causes the back end to assert.
* Always use `llvm::TargetMachine::CGFT_AssemblyFile` for ptx, if we don't create target machine will return null
* `llvm::createSPIRVWriterPass(out)->runOnModule(m)` is the way to make output happen for SPIR-V as there is no target machine for SPIR-V.


thewilsonator:
fixing the extra commits, #1879

PR #1878, opened 2016-11-13
[dcompute] ABIs for nvptx and spirv

thewilsonator:


thewilsonator:
I don't think I need to do anything special for these ABIs, and C++ mangling seems to be the only issue that I found, an that can be fixed with extern(C++) or pragma mangle.

gen/abi-spirv.cpp
```diff
@@ -0,0 +1,34 @@
...
+    if (hasKernelAttr(fdecl))
+      return llvm::CallingConv::SPIR_KERNEL;
+    else
+      return llvm::CallingConv::SPIR_FUNC;
+  }
+  // Since SPIR-V is an intermediate format and is subject to no optimisations
+  // inside llvm, but are subject to further optimisations post-translation
+  // don't bother doing anything too fancy.
+  bool passByVal(Type *t) override {
+    return t->size() <= 16; //enough for a float4
```
kinke: I think you'd better `return DtoIsInMemoryOnly(t)` here (true for all structs and static arrays). Please have a look at the comment of the abstract function to understand the special LLVM `byval` semantics. In essence, the IR would currently pass all types <= 16 bytes in memory, and all bigger types directly (in registers if available), so I guess the intent was the opposite.
Note that `returnInArg()` is tightly related to this (I'd recommend something like `return !tf->isref && DtoIsInMemoryOnly(tf->next)` there).

thewilsonator: Thanks, I'll probably disallow ref as a function return type and parameter because it does not play well with addrspaces, but in another PR that actually deals with that.

I'm still not entirely convinced that having byVal for small structs is a good idea (In dcompute all "pointers" are structs that contain a pointer (i.e. { float addrspace(1)* }). In particular pointer arguments to kernels with byVal attribute is going fail because pointers to pointers are disallowed in earlier versions of OpenCL.

For the same reasons for disallowing ref,`fooStruct = foo();` will fail if `fooStruct` is in an addrspace other than 0 e.g. if `fooStruct` is in shared memory if using sret.

thewilsonator:
pining @JohanEngelen 
CircleCI fails due to the reasons for PR #1881 
I have done as @kinke suggested and w.r.t passByVal and returnInArg
This will most likely need to be changed later, what to I'm not sure at the moment, because I need some real user feedback. 

Any way if there is no more concerns then please merge this and sync with master so that CircleCI works.

gen/abi-nvptx.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "gen/abi.h"
+#include "gen/uda.h"
+#include "ddmd/declaration.h"
+#include "tollvm.h"
+
+struct NVPTXTargetABI : TargetABI {
+  llvm::CallingConv::ID callingConv(llvm::FunctionType *ft, LINK l,
+                                    FuncDeclaration *fdecl = nullptr) override {
+    if (hasKernelAttr(fdecl))
```
kinke: I think a prior `assert(fdecl);` wouldn't hurt (as it defaults no null), for SPIRV as well.

gen/abi-nvptx.h
```diff
@@ -0,0 +1,21 @@
+//===-- gen/abi-nvptx.h - PPC64 ABI description ----------------*- C++ -*-===//
```
kinke: ~~PPC64~~, same for the other header.

gen/abi-nvptx.h
```diff
@@ -0,0 +1,21 @@
...
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+//
+// The ABI implementation used by dcompute for targetting NVPTX for cuda
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LDC_GEN_ABI_SPIRV_H
+#define LDC_GEN_ABI_SPIRV_H
```
kinke: `LDC_GEN_ABI_NVPTX_H`, otherwise the 2 new headers are mutually exclusive!

kinke:
> In dcompute all "pointers" are structs that contain a pointer (i.e. { float addrspace(1)* })

`DtoIsInMemoryOnly()` uses the D type, so the LL parameters representing D pointer parameters shouldn't get the `byval` attribute and be passed directly as LL struct `{ float addrspace(1)* }` then, right?

thewilsonator:
Thanks for the review!

>`DtoIsInMemoryOnly()` uses the D type, so the LL parameters representing D pointer parameters shouldn't get the byval attribute and be passed directly as LL struct `{ float addrspace(1)* }` now, right?

It would, but it is a struct on the D side as well (`struct Pointer(T,uint n) { T* ptr; alias ptr this;}` ), which is why it ends up as a struct on the LL side. It would be nice to fold the  `{ float addrspace(1)* }` to `float addrspace(1)*` but then I'd have to repair all the GEPs that access it, which would be nice since it would allow us to optimise SPIRV (GEP ->extract = triggered assert in the backend), but its no great loss because SPIRV is supposed to be optimised by the consumer.

I need to define its location (place in module hierarchy of ldc.*) and special case it in `DtoIsInMemoryOnly` but i'll do that in the next PR ( https://wiki.dlang.org/LDC_CUDA_and_SPIRV#Pull_Requests target + codegenmanager).

kinke:
Okay, so you're using different D struct wrappers for pointers to be able to discriminate between the various address spaces, correct? Nothing wrong with that, but did you consider using a different semantic for existing D keywords such as `__gshared` (global memory), `shared` (wavefront) etc. for dcompute modules?

When sticking with D struct wrappers, don't worry about optimizing away the LL struct and the (noop) GEP, as it really shouldn't make a difference in the end except for a maybe slightly more verbose IR. We do keep D structs in memory on the IR level, but we so do for all lvalues (incl. pointer locals and params) inside a function too, so that shouldn't make a big difference either (and even less when optimizing).
Passing the struct wrappers via `byval` and returning them via sret is not desirable though, so those should be passed directly on the IR level. How do you plan to handle other structs and static arrays passed by value? I guess using byval for those and making sure the bitcopy ends up in the thread's private memory makes sense.

thewilsonator:
>Okay, so you're using different D struct wrappers for pointers to be able to discriminate between the various address spaces, correct? Nothing wrong with that, but did you consider using a different semantic for existing D keywords such as __gshared (global memory), shared (wavefront) etc. for dcompute modules?

Wont work because __gshared is not part of the type, you'd need to constantly cast away shared to do anything, and I need more addrspaces (5) than I have keywords usable.

> Did you verify the byval bitcopies end up in the thread's private memory?

I get `foo_args` that are copied, so yes.

kinke:
Yes, the keywords thing would entail hacking the front-end to allow for a dcompute mode, in which existing keywords could be used as additional, purely cosmetical type modifiers, allowing for something like: `void kernel(float[] privateData, ref const __gshared(GlobalStruct) globalData, ref shared(WFStruct) wavefrontData, immutable(float)[] constData)`.

I merged master into dcompute. Wrt. adding the struct wrapper exceptions to `DtoIsInMemoryOnly()`, I would rather recommend adding them to the ABIs instead (in `passByVal()`, and using that in `returnInArg()`). `DtoIsInMemoryOnly()` is often used, but not in a 100% consistent way, i.e., there are places where D structs are still assumed to be in memory.

thewilsonator:
Yeah I'm trying to avoid changing the front end if at all possible.
> Wrt. adding the struct wrapper exceptions to DtoIsInMemoryOnly(), I would rather recommend adding them to the ABIs instead (in passByVal(), and using that in returnInArg())

Will Do.

>DtoIsInMemoryOnly() is often used, but not in a 100% consistent way, i.e., there are places where D structs are assumed to be in memory. Otherwise, you'd need to replace every GEP+load by an extract etc.

It's actually the other way around, i.e. extracts cause the backing to assert, hence the choice to leave SPIR-V modules completely unoptimised (they do get optimised by the consumer). It may be worth writing an unoptimisation pass, but that can wait.  

Thanks heaps!

Issue #1876, opened 2016-11-12
nightly Windows build does not work with file access tracking


rainers:
I'm trying to integrate LDC with Visual D using file tracking for automatic dependency handling. Using the nightly build from https://github.com/ldc-developers/ldc/releases/tag/LDC-Win64-master causes an error, though:
```
C:\tmp>"C:\Program Files (x86)\MSBuild\14.0\Bin\amd64\Tracker.exe" /i c:\tmp /d "C:\Program Files (x86)\MSBuild\14.0\Bin\amd64\filetracker64.dll" /c c:\tmp\ldc-x64\bin\ldc2.exe
FileTracker : error FTK0001: could not attach tracking. Not enough storage is available to process this command.
```
This does not happen with my own built ldc2.exe or ldmd2.exe. Fortunately, it also doesn't happen with 1.1.0-beta-3. It also does not happen with the 32-bit built ldc2.exe.
It is unlikely that it actually has something to do with the size of the executable (the appveyor build is the only to include a lot of target architectures), as it also happens with ldmd2.exe.

kinke:
AppVeyor uses the `RelWithDebInfo` CMake config, so featuring debug infos. Its LLVM is built using `Release`, but includes assertions, so the LDC assertions are enabled too. All targets are enabled for that LLVM.

> Not enough storage is available to process this command.

Are you sure it's not just your HDD/SSD filling up?

rainers:
> AppVeyor uses the RelWithDebInfo CMake config, so featuring debug infos. Its LLVM is built using Release, but includes assertions, so the LDC assertions are enabled too. All targets are enabled for that LLVM.

My local ninja build is RelWithDebInfo with assertions for both LLVM and LDC. I also tried the debug build by VS. Both can be run with the file tracker. Maybe there is something about the other targets?

> Are you sure it's not just your HDD/SSD filling up?

Pretty sure. I even started the Sysinternals process monitor to find the difference to cl.exe before noticing that it works with other builds of LDC. I couldn't spot any real difference in the logs, though.

kinke:
I can reproduce it - my own (`RelWithDebInfo`) builds work fine, an older 64-bit AppVeyor one doesn't. A msgbox with `application was unable to start correctly (0xc0000142)` pops up; that error is apparently related to failed DLL initialization. My builds use the dynamic MSVCRT libs, so does AppVeyor...

rainers:
[...]

PR #1880, opened 2016-11-13
fix issue 1876: MS file tracker fails to run for executables built with debug info

rainers:
passing /LARGEADDRESSAWARE:NO is a workaround for DWARF debug info, it causes trouble with CodeView debug info.

See https://github.com/ldc-developers/ldc/issues/1876

### DIFF
--- a/driver/linker.cpp
+++ b/driver/linker.cpp
@@ -688,15 +688,9 @@ static int linkObjToBinaryMSVC(bool sharedLib) {
   args.push_back("/NOLOGO");
 
   // specify that the image will contain a table of safe exception handlers
-  // (32bit only)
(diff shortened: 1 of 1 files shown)


driver/linker.cpp
```diff
@@ -694,7 +694,8 @@ static int linkObjToBinaryMSVC(bool sharedLib) {
   }
 
   // because of a LLVM bug, see LDC issue 442
-  if (global.params.symdebug) {
+  if (global.params.symdebug &&
+      !global.params.targetTriple->isWindowsMSVCEnvironment()) {
```
rainers: I should have read the function name... /LARGEADDRESSAWARE now added only for 32-bit unconditionally.

PR #1882, opened 2016-11-13
Consider pointer literals as constant in isConstLiteral.

JohanEngelen:
Similar to #1838 .

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -526,6 +526,19 @@ bool isConstLiteral(Expression *e) {
   case TOKstring:
     return true;
 
+  case TOKsymoff: {
+    // Note: dllimported symbols are not link-time constant.
+    auto soe = static_cast<SymOffExp *>(e);
+    if (VarDeclaration *vd = soe->var->isVarDeclaration()) {
+       return vd->isDataseg() && !vd->isImportedSymbol();
+    }
+    if (FuncDeclaration *fd = soe->var->isFuncDeclaration()) {
+        return !fd->isImportedSymbol();
+    }
+    // Assume the symbol is non-const if we can't prove it is const.
+    return false;
+  } break;
+
   default:
     if (e->isConst() != 1)
       return false;
--- a/tests/codegen/const_struct_export.d
+++ b/tests/codegen/const_struct_export.d
@@ -1,0 +1,20 @@
+// RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+// CHECK:     @.immutablearray{{.*}} = internal constant [2 x void ()*] {{.*}}exportedFunction
+// CHECK-NOT: @.immutablearray{{.*}} [2 x void ()*] {{.*}}importedFunction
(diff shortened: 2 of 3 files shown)


gen/arrays.cpp
```diff
@@ -526,6 +526,14 @@ bool isConstLiteral(Expression *e) {
   case TOKstring:
     return true;
 
+  case TOKsymoff: {
+    auto soe = static_cast<SymOffExp *>(e);
+    if (VarDeclaration *vd = soe->var->isVarDeclaration()) {
+      if (!vd->isDataseg())
+        return false;
+    }
```
kinke: So if it's not a variable, it's always const? I'd recommend using `return vd->isDataseg()` in the outer `if` to clarify things, and falling through to the `default` case (maybe with an explicit comment), unless you are absolutely sure either `Declaration *soe->var` is always a variable (and add a corresponding assert) or `soe` really is always const otherwise.

JohanEngelen: If it's not a variable, it's a function symbol, whose address is link-time constant.
I played a little with how to write this piece, and the result is the PR. The default `return true` of the function is awkward, giving rise to the funny controlflow for this case. I thought it'd be better to only return `true` for the cases where we know it is valid, but then implemented the opposite logic to reduce the number of operations involved. After your comment, I'll rewrite the original. It's less error-prone.

kinke: > If it's not a variable, it's a function symbol, whose address is link-time constant.

Yeah that's what I would have thought until a few days ago too, where the `dllimport` stuff came along. An imported `export` symbol's address isn't a link-time constant, it's a pointer set up at runtime.

JohanEngelen: I'm rewriting the code, but it turns out that for taking the address of an definition-only `export` function, the token is not `TOKsymoff`, but `TOKaddress`.

kinke: Thanks, looks better now.

> auto barptr = &bar; // ICE

That's one of the complexities Benjamin Thaut mentions in his talk. Globals like this need to be initialized separately at runtime, that's what he does in his PR.

Another thing he mentions are TLS globals. I don't remember whether `isDataseg()` is true for those - if so, they aren't constant either.
I hate to damp your enthusiasm, but for

tests/codegen/const_struct_export.d
```diff
@@ -0,0 +1,20 @@
...
+
+export void exportedFunction() {}
+export void importedFunction();
+export immutable int exportedVariable = 1;
+export immutable int importedVariable;
+
+void foo () {
+    immutable auto exportedFuncs = [ &exportedFunction, &exportedFunction ];
+    immutable auto importedFuncs = [ &importedFunction, &importedFunction ];
+    // CHECK: store void ()* @_D19const_struct_export16importedFunctionFZv
```
kinke: `... @{{.*}}_D19...` => annoying 32-bit MSVC mangling (`@"\01__D19const_struct_export16importedFunctionFZv"`)...

JohanEngelen: Fixed it.

JohanEngelen:
Do we want this in 1.2?

PR #1886, opened 2016-11-15
Partial cherry-pick from DMD master to remove import cycle.

JohanEngelen:
Had to add a fwd reference to break the cycle.
This fixes the build for dmd 2.072 (travis.yml is also updated).


JohanEngelen:
It's not an actual cherry pick, but the commit is https://github.com/dlang/dmd/commit/1d0ab8b9c136e46bf449c506ca25d2c8a784f7b9#diff-b4674e7b5d3a44178526afdefc9aa368
I'll add it upon merge.

PR #1887, opened 2016-11-17
For non-tagged commits, change version string to e.g. "1.1.0git-c693e264"

JohanEngelen:
This changes the version string from "c693e264" to "1.1.0git-c693e264", same as what LLVM does.
I find this makes it easier to see what version people are using. The version number is the number of the version currently being developed (same as what LLVM does).

Note: this only concerns non-tagged commits, nothing changes for the version string in releases.


JohanEngelen:
Btw, I tried to get "dirty" in the version string when changes were made relative to the checked-out commit.  Do more people want this? (it'll mean adding more stuff to the cmake script, but it's easy)

kinke:
Good idea. Wrt. 'dirty', it would also be nice as long as the CMake Voodoo is only a couple of lines IMO.

JohanEngelen:
Added "-dirty" too.

kinke:
Looks good, from the AppVeyor build:
```
LDC - the LLVM D compiler (1.1.0git-8a336e62):
  based on DMD v2.071.2 and LLVM 3.9.0git-fbbabf3
```

Nitpick: does your LLVM also use 7 chars for the hash? Then I'd prefer we use the same, not 8.

JohanEngelen:
We used to use 6 hash chars. (Git itself has some algorithm for determining how many chars to use I think (adds a char when it could be ambiguous)). `git describe` shows 8 chars. SourceTree shows 7 chars. Github shows 7.
Let's do 7 and be happy ;)

Issue #1864, opened 2016-10-28
ldc 1.1.0 beta2/beta3 linux-arm Segmentation fault


pisetta:
Hi,
I tested that on a Raspberry Pi 3 with ldc2-1.1.0-beta2-linux-arm and ldc2-1.1.0-beta3-linux-arm both segfaulted with this code.

```
import std.stdio;

struct Aggr(alias V) {
    alias Value = V;
}

template templateDecl( Vs... ) {
    void templateDecl() {
        foreach( v; Vs )
            static if ( is( v : Aggr!A, alias A ) )
                write( A );
        writeln();
    }
}

alias templateInst(alias V) = templateDecl!( Aggr!V );

void main() {
    string var = "var";
    templateInst!var();
}
```

I will test if x86 and x86_64 also segfaults soon.

Gianni Pisetta

pisetta:
Hi,
tested the code with linux-x86 and linux-x86_64 and both segfaulted. Attached there is the output of both, if it can be of any help.

Gianni Pisetta
[output-x86.txt](https://github.com/ldc-developers/ldc/files/558388/output-x86.txt)
[output-x86_64.txt](https://github.com/ldc-developers/ldc/files/558387/output-x86_64.txt)

LemonBoy:
The code fails [here](https://github.com/ldc-developers/ldc/blob/master/gen/nested.cpp#L45), `fd` is `<filename>.templateDecl!(Aggr!(var)).templateDecl` whose parent is `null` instead of `main`.
dmd fails with `function D main is a nested function and cannot be accessed from segf.templateDecl!(Aggr!(var)).templateDecl` so it could be considered as a bug in the frontend (the template instantiation is in `main` and I'd expect `main` to be the topmost parent), otherwise a tentative solution would be to change [this line](https://github.com/ldc-developers/ldc/blob/master/gen/nested.cpp#L46) into `!fd || fd->isStatic()`,

kinke:
It may be a front-end bug. Still, I think your proposed change (replace `!fd` with `!vdparent` though) absolutely makes sense. Also note the `assert(vdparent);` a few lines above.

LemonBoy:
Hmm, in this case `vdparent` is not null, the assertion (that today decided it was right about time to fire) fires when checking `fd`:
`ldc2: ../gen/nested.cpp:52: DValue* DtoNestedVariable(Loc&, Type*, VarDeclaration*, bool): Assertion `fd' failed.`

[...]

PR #1888, opened 2016-11-17
Fix check if nested function can access outer function frame

kinke:
The previous check wouldn't check intermediate aggregates for static-ness, that was one problem. The other was the assertion that the outer function can be reached as long as there are no static functions inbetween, which isn't always the case, as issue #1864 clearly shows, which is resolved by this.

### DIFF
--- a/gen/nested.cpp
+++ b/gen/nested.cpp
@@ -42,14 +42,13 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
 
   // Check whether we can access the needed frame
   FuncDeclaration *fd = irfunc->decl;
-  while (fd != vdparent) {
-    if (fd->isStatic()) {
-      error(loc, "function %s cannot access frame of function %s",
-            irfunc->decl->toPrettyChars(), vdparent->toPrettyChars());
-      return new DLValue(astype, llvm::UndefValue::get(DtoPtrToType(astype)));
-    }
-    fd = getParentFunc(fd, false);
-    assert(fd);
+  while (fd && fd != vdparent) {
+    fd = getParentFunc(fd);
+  }
+  if (!fd) {
+    error(loc, "function %s cannot access frame of function %s",
+          irfunc->decl->toPrettyChars(), vdparent->toPrettyChars());
+    return new DLValue(astype, llvm::UndefValue::get(DtoPtrToType(astype)));
   }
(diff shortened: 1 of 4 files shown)


gen/nested.cpp
```diff
@@ -42,14 +42,13 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
...
-  while (fd != vdparent) {
-    if (fd->isStatic()) {
-      error(loc, "function %s cannot access frame of function %s",
-            irfunc->decl->toPrettyChars(), vdparent->toPrettyChars());
-      return new DLValue(astype, llvm::UndefValue::get(DtoPtrToType(astype)));
-    }
-    fd = getParentFunc(fd, false);
-    assert(fd);
+  while (fd && fd != vdparent) {
+    fd = getParentFunc(fd);
```
kinke: `fd` will be null if there are static functions or aggregates in the parents hierarchy from and including `fd` up to the next outer function (if any, otherwise null as well).

JohanEngelen:
The testcase is missing from the PR.

kinke:
It'd be a dmd-testsuite one (so added later), if it actually makes sense - I think it's really a front-end bug and should work, so I'm rather reluctant to add one.

JohanEngelen:
Whereever the bug is, we need a testcase for this.

kinke:
My point is that I'm reluctant to add an output-matching fail-compilation test for something that looks like it should be working. I.e., a test making sure that something produces an error (instead of crashing), although it looks like valid code.

JohanEngelen:
Ah, I misunderstood. So the problem is probably a rejects-valid type of bug?
You seem to have fixed two bugs, none of the two can be tested?

kinke:
Probably, yes, I'm not sure. I can obviously add a test for the time being, so that we at least get notified when it works (if ever).

The other thing wrt. static aggregate parents/function literals is probably hard/impossible to test, as I guess the front-end rejects the code earlier and we never get into a problematic `DtoNestedVariable()` call. I think that check is only performed to make sure we can later really access all the frames in LDC's nested context code via later usages of `getParentFunc()`.

JohanEngelen:
The changes LGTM.

I prefer a test for the time being, but leave it to you.

kinke:
I've added the test.

PR #1889, opened 2016-11-18
[1.1] Master minus OSX shared libs

kinke:


kinke:
@JohanEngelen: I implicitly also reverted your https://github.com/ldc-developers/ldc/pull/1889/commits/eb0f97aaf16e1f35b9b23e4b5589680686fe8f62 during the revert fix. Is it a fix, needed for LLVM 3.9 or something like that?

kinke:
`tests/codegen/llvm_used_1.d`: there's no `llvm.used` array at all anymore. Was that specific to the OSX rework?

kinke:
Well in the reverted version, the ModuleInfo ptr is added in `build_module_ref()`, which isn't used for OSX: https://github.com/kinke/ldc/commit/bb2b648d8e286c82cc55219a2d2ac067e361d432#diff-8432a5b8ca0dee535fe5e1196e83b8c6L621

Looking at it now, I should have used a non-Android Linux triple instead (I removed the IR checks as quick workaround).

kinke:
Number of successful OSX runs: 1 (to be updated by manual retriggering)

JohanEngelen:
> I implicitly also reverted your eb0f97a during the revert fix-up

It is needed for LTO on Linux (previous behavior relies on LLVM ordering user segments in order of their appearance, which is not valid when doing LTO).

kinke:
Alright, so that'd need to be backported. The vibe.d issue(s) should have top priority for 1.1. Apart from that, it's looking good - I think the 31 successful OSX runs in a row clearly indicate that the spurious OSX core.thread unittest segfault is related to the OSX shared libs work.

JohanEngelen:
I'm OK with merging this, backporting the LTO fix (it is strange that CircleCI does not error on the LTO module constructor/destructor test), and cherry-picking onto the release-1.1.0 branch from then on.

JohanEngelen:
> I think the 31 successful OSX runs in a row clearly indicate that the spurious OSX core.thread unittest segfault is related to the OSX shared libs work.

Another data point: ldc master also shows the core.thread failure with LLVM3.9. (Note that this PR ups OSX testing to LLVM3.9)

JohanEngelen:
@klickverbot Same here.

kinke:
I removed the temporary commit reconfiguring Travis for 4 OSX builds.

Issue #1741, opened 2016-09-04
Assertion failed: "Initializer for composite element doesn't match!"


JohanEngelen:
Steps to reproduce:
- `dub init -t vibe.d test`
- `dub build -b debug --combined --compiler=ldc2`

Fails ~~for LDC built with LLVM 3.9.0 and LLVM 4.0 (trunk),~~ with the LLVM assert:

```
type mismatch for entry # 6 in vtbl initializer
%vibe.core.task.Task (%vibe.core.task.TaskFiber*)*
{ i8*, i64 } (%vibe.core.task.TaskFiber*)*
Assertion failed: (V[I]->getType() == T->getTypeAtIndex(I) && "Initializer for composite element doesn't match!"), function ConstantAggregate, file ../lib/IR/Constants.cpp, line 881.
```

~~Works fine with LLVM 3.8.~~

I haven't had time to narrow this down to something workable, but thought it needed reporting.

JohanEngelen:
vibe version: vibe-d-0.7.29

JohanEngelen:
dustmiting...

JohanEngelen:
File a.d 
```d
import task;

Task runTask;
```

File task.d
```d
class Super
{

    this(void delegate())
    {
    }

}

struct Task
{
    shared(Child) m_fiber;
}

class Child : Super
{
    this(void delegate() fun)
    {
        super(fun);
    }

    @property task()
    {
        return Task();
    }
}
```

`ldc2 -c task.d a.d` errors with  (LLVM 3.9):
```
type mismatch for entry # 5 in vtbl initializer
%task.Task (%task.Child*)*
i8* (%task.Child*)*
Assertion failed: (V[I]->getType() == T->getTypeAtIndex(I) && "Initializer for composite element doesn't match!"), function ConstantAggregate, file ../lib/IR/Constants.cpp, line 883.
```

The order of the files on the cmdline is important!
(compiles successfully with LLVM 3.8)

klickverbot:
Is this a 3.8/3.9 mismatch issue or just down to LLVM assertions being enabled or not?

JohanEngelen:
Hmm, I initially found this using the 1.1.0 binary package built with LLVM3.8 and 3.9. Apparently one has assertions on, the other doesn't. Doh.

So, indeed, LLVM 3.8 also fails.

klickverbot:
[...]

Issue #662, opened 2014-07-01
Check unnecessary vtbl emissions


redstar:
Reported by temtaine in news group (http://forum.dlang.org/post/yephpwnmvkcpyjnpbzti@forum.dlang.org). Look at the following modules:

A.d

```
import B;
```

B.d

```
import C;
import D;

import std.range;

class Font {
public:
    mixin RCounted;

    auto makeTextData(string s) {
        // split text by spaces
        auto arr = s.splitter.array;
    }    
}

class Engine {
    RC!Font font;
}
```

C.d

```
import A;
```

D.d

```
template RCounted() {
        void release() { this.destroy; }
}

struct RC(T) {
    ~this() {  _rc_ptr.release; }
    T _rc_ptr;
}
```

`ldc2 -c A.d B.d C.d D.d` works but `ldc2 -c A.d` produces an error message.

kinke:
The original symptoms should be worked around by #1709.

> When we codegen object.destroy!(Font) (dmd does this too), we call IrTypeClass:get() for B.Font and that generates the vtbl for that class (I don't think dmd does that).

Still needs to be looked into.

kinke:
Vtables are emitted once into the CU owning the class declaration. They are needed for the class init symbol, which is also emitted in that CU, so definitely not superfluous.

The testcase (nowadays requiring `import std.algorithm, std.array;`) works with LDC v1.27.

PR #1891, opened 2016-11-19
Opaque vtbl

kinke:
This gets Johan's reduced testcase in issue #1741 to compile on Linux x64, but was actually an experiment I did a couple of weeks ago wrt. issue #662. Not sure if opaqueness here is desirable, as it just hides potential function type mismatches; the primary aim was getting rid of additional semantic analysis for the function types of all virtual functions.

### DIFF
--- a/tests/compilable/gh1741.d
+++ b/tests/compilable/gh1741.d
@@ -1,0 +1,13 @@
+// Explicitly target Linux x86_64.
+// REQUIRES: target_X86
+// RUN: %ldc -mtriple=x86_64-pc-linux-gnu -c %s %S/inputs/gh1741b.d
+
+struct S
+{
+    C c;
+}
+
+class C
+{
+    @property s() { return S(); }
+}
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -390,7 +390,6 @@ LLValue *DtoVirtualFunctionPointer(DValue *inst, FuncDeclaration *fdecl,
 
   LLValue *funcval = vthis;
   // get the vtbl for objects
-  stripModifiers(inst->type->toBasetype())->ctype->isClass()->getVtblType(true);
   funcval = DtoGEPi(funcval, 0, 0);
   // load vtbl ptr
   funcval = DtoLoad(funcval);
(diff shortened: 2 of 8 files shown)


kinke:
When trying to build a dub project template using vibe.d on Linux x64, it now fails with the error `../../.dub/packages/vibe-d-0.7.30/source/vibe/http/server.d(1640,32): Error: variable vibe.http.server.handleHTTPConnection.tls_stream variables cannot be of type void`; LDC previously hit the assertion.

kinke:
The above issue is https://github.com/rejectedsoftware/vibe.d/issues/1611. When adding the OpenSSL dependency manually, a vibe.d project template can be compiled and run successfully with LDC/LDMD and dub v0.9.24-2 on Ubuntu 16.04 (x86_64).

kinke:
When working around https://github.com/rejectedsoftware/vibe.d/issues/1629, a vibe.d project template also works on Win64 with both LDC2 and LDMD2.

kinke:
I think we need this in v1.1. We can't afford not to support vibe.d.

JohanEngelen:
Test case is missing.

(In the future, perhaps it's worthwhile trying to emit vtables (`available_externally`) for some optimization potential.)

kinke:
Test case is a bit tricky. The crash is only triggered by assertions, so neither for Travis nor for Circle. It's only CI-reproducible when targeting non-Windows for AppVeyor, which seems a bit awkward.

JohanEngelen:
(not having assertions in Travis/CircleCI is really bad :( :( )

JohanEngelen:
Still good to add a test though, even if it cannot trigger on CI. Also, I don't mind having tests with explicit `-target ...` things in it; one of our strengths is being a cross-compiler. (note: clang/llvm has a ton of such tests)

kinke:
Minified test amended (I verified it still asserts with master).

tests/compilable/gh1741.d
```diff
@@ -0,0 +1,12 @@
+// Explicitly target Linux x86_64.
+// RUN: %ldc -mtriple=x86_64-pc-linux-gnu -c %s %S/inputs/gh1741b.d
```
JohanEngelen: small comment: you can add `// REQUIRES: target_X86` so that it doesn't fail for the (non-existent?) builds that do not support x86_64.

JohanEngelen: btw, it's quite incredible just how simple some testcases are that failed to compile before.

kinke: Thanks, that may help some ARM folks etc. - done.

kinke:
Yes, please. :)

kinke:
Note that we previously cast the looked-up function address anyway, we just verified the function types (for `!NDEBUG`) when building the vtable constant, where we would potentially run into forward-referencing issues like #1741 and possibly have to do additional semantic analysis. Additionally, there's no named LLVM struct type for the vtable of each class anymore, rendering the IR and especially the `-vv` log more compact (`-vv`: -30% lines for a printf-hello-world due to TypeInfos from `object.d` IIRC).

PR #1893, opened 2016-11-23
Implement different cache recovery mechanisms.

JohanEngelen:
The current retrieval mechanism (retrieve file from cache upon cache hit) is LLVM `create_link`, which creates a symbolic link on Unix/OSX and creates a hard link on Windows.  When symlinked cache files keep being used without the compiler knowing about it (e.g. they are only used by the linker), the cache pruning algorithm will delete the cached file when it becomes old.

`-cache-retrieval={copy|hardlink|link|symlink}`

This PR implements 4 different cache retrieval methods: copy, symlink, hardlink, "link" (previous behavior). "copy" is the safest, as it will work for all kinds of setups, so I've set that as the default. 
The downside is more disk space usage and longer cache retrieval time, especially for large files. The recommendation is to use "hardlink" when it works for the user's setup (doesn't work across filesystems, for example).

### DIFF
--- a/driver/cache.cpp
+++ b/driver/cache.cpp
@@ -55,6 +55,61 @@
 #include <io.h>
 #endif
 
+#if LDC_POSIX
+#include <unistd.h>
+// Returns true upon error.
+static bool createHardLink(const char *to, const char *from) {
+  return link(to, from) == -1;
+}
+// Returns true upon error.
+static bool createSymLink(const char *to, const char *from) {
+  return symlink(to, from) == -1;
+}
+#elif _WIN32
+#include <windows.h>
+namespace llvm {
+namespace sys {
+namespace path {
+// Fwd declaration to an internal LLVM function.
+std::error_code widenPath(const llvm::Twine &Path8,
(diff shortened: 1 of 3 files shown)


JohanEngelen:
It'd be good to have this in 1.1.0 too.
Caching can feel broken without it, especially when used over a long time period.

JohanEngelen:
@kinke Hope to get this into beta4 too.  The gist of this PR is to back-off by default what caching does:  let's have it copy files instead of symlinking by default.
(I somehow had the default wrong, will change it in 2 seconds)

JohanEngelen:
fixed

kinke:
Sure, fine by me.

JohanEngelen:
CircleCI failure: https://github.com/ldc-developers/ldc/pull/1895

nordlow:
What about adding yet another alternative that first tries to do a hardlink and then does either symlink or copy?

JohanEngelen:
@nordlow Sounds like a good idea, interested in adding it?

PR #1895, opened 2016-11-24
Fix LLM 4.0 build.

JohanEngelen:


JohanEngelen:
Added the milestone 1.1.0, so that we can provide LLVM4.0 binaries aswell if people are up for something adventurous.

PR #1896, opened 2016-11-25
Implement -ffast-math.

### LINKED ISSUE #1874: Compiler switch for applying @fastmath throughout a compilation unit
Opening an issue here as [requested](http://forum.dlang.org/post/szxizxxebxqtqyeerflt@forum.dlang.org).

Basically, a significant amount of performance sensitive numerical code is written under the assumption of associative and finite floating point arithmetic rather than strict IEEE compliance. In many real world cases this is a matter of performance difference by a factor of 3 or more due to the resulting vectorization opportunities. While the existing @fastmath attribute is very useful and sometimes function-level control is the appropriate choice, having a compiler switch analogous to the --ffast-math of GCC/Clang is very convenient for code that does not require IEEE compliance anywhere (in my line of work, most of the code I write). It would also greatly simplify benchmarking the effects of changing the floating point model quickly.

JohanEngelen:
Resolves #1874.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -14,6 +14,7 @@
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Target/TargetMachine.h"
 #include "llvm/IR/DataLayout.h"
+#include "llvm/IR/Operator.h"
 
 namespace opts {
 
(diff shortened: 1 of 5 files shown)


JohanEngelen:
?  ffast-math is already defined? I don't have this problem locally (OSX, LLVM3.9). Hmmm, the Hexagon backend defines "ffast-math"...

thewilsonator:
Should be easily solved by renaming "fast-math" in hideLLVMOptions.

JohanEngelen:
Includes/depends on #1919

PR #1898, opened 2016-11-27
Suffix the LTO plugin packaged with LDC with "-ldc", to prevent installation conflicts

### LINKED ISSUE #1897: ldc shouldn't install LLVMgold.so
I was building ldc 1.1.0 beta4 for Fedora and noticed that it installs a /usr/lib/LLVMgold.so file which it clearly shouldn't as it's part of llvm. This only seems to be happening on 64 bit architectures for some reason -- llvm's LLVMgold.so is already in /usr/lib64/ and then ldc tries to install it to /usr/lib/ (note the lib64 vs lib).

https://kojipkgs.fedoraproject.org//work/tasks/7209/16647209/build.log has the full build log (short lived link)

JohanEngelen:
Should resolve #1897


kinke:
Please see my edited https://github.com/ldc-developers/ldc/issues/1897#issuecomment-263132169.

kinke:
Sure.

Issue #1901, opened 2016-11-28
libLTO in bin


John-Colvin:
`libLTO-ldc.dylib` is a library, why is it being installed in `bin`, not in `lib`?

thewilsonator:
Because it's a plugin for the linker.

John-Colvin:
The reason I ask is that homebrew only allows executables in `bin`.

John-Colvin:
https://github.com/Homebrew/homebrew-core/pull/7308

JohanEngelen:
See https://github.com/ldc-developers/ldc/issues/1897

I'll work on a fix today.

kinke:
Fixed by #1902. It will be installed into `lib`, unless you specify a suffix via CMake, e.g., `-DLIB_SUFFIX=64` => `lib64`.

PR #1902, opened 2016-11-28
Fix #1897 and #1901

### LINKED ISSUE #1897: ldc shouldn't install LLVMgold.so

JohanEngelen:
Make "LIB_SUFFIX" a global cmake option (not just for the runtime) and use it also to define the LTO plugin install/search location.


JohanEngelen:
#1897  #1901

JohanEngelen:
I did not rename "LIB_SUFFIX" to "LDC_LIBDIR_SUFFIX" because I am not sure whether anybody is already using LIB_SUFFIX in his LDC build setup.

kinke:
The hardcoded #define is pretty ugly for multilib builds which may rename the lib directories (or cases where a user downloads a package and has to rename the dirs manually for his/her system setup and only later finding out that LDC uses a hardcoded relative path). So I'd prefer David's suggestion wrt. a default `-flto-binary` in the config file and hope I don't make you angry again.

JohanEngelen:
It's just extra work... feel free to implement it.

JohanEngelen:
(I can't please everybody, and it takes more and more time to think of the implications of adding it to ldc.conf, thinking about all use cases, fallbacks, etc.  :'(  I'm sorry, I'm already tired, and this is extra tiring... :( )

JohanEngelen:
(So I'll have a go at it later today......)

JohanEngelen:
Done.
Waiting for something else to fail :(  Argh, buildsystem stuff...

kinke:
Thanks, looks good after a first glance, but I'll have to recheck it more thoroughly. Let's make sure we get rid of this issue for good. ;)

JohanEngelen:
Already found the first problem :(  Fixed it.

JohanEngelen:
@John-Colvin @kalev  Is it possible for you to test this PR ?

kalev:
Thanks for the fixes! I tried the pull request and it works fine here on Fedora, putting LLVMgold-ldc.so in /usr/lib64.

I am not convinced about the approach though. I think most linux distros would probably prefer to just have a dependency on llvm libraries and avoid making a copy of LLVMgold.so. Why is it necessary to make a copy of it instead of just depending on llvm libraries? Is it to cut down the number of deps ldc has?

sambrightman:
@kalev see #1897 - to avoid requiring an LLVM install in order to use LTO. Presumably this would be easier if LLVM released the plugin separately.

kalev:
Ahh, I see. I think this doesn't actually work, because LLVMgold.so at least on my Fedora 25 is linked against shared llvm libraries:

$ readelf -d build/lib64/LLVMgold-ldc.so | grep NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libLLVM-3.8.so]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]

Since LLVMgold-ldc.so is linked with libLLVM-3.8.so, the copying trick doesn't avoid the dep on llvm libraries at all. Instead, this makes ldc pull in llvm-libs (a package that includes both libLLVM-3.8.so and LLVMgold.so) + ldc also installs another copy of LLVMgold.so.

I don't think the llvm dependency is a problem at all from a Linux distro's perspective, but the copying seems a bit unnecessary, given that LLVMgold.so is going to be installed anyway for ldc.

Would it be possible for ldc to just use the system LLVMgold.so if installed? And for upstream-distributed binary builds maybe have a build option that would make a copy of it?

kinke:
Thanks @kalev for your input. So I guess the most sensible thing to do is not to copy the file and instead require an installed/homebrew LLVM compatible with the version LDC was built against when using LTO (and keeping on trying to autodetect the path to `LLVMgold.so`), at least for guys doing a manual `make install` LDC installation/homebrew.
For the binary packages, I guess `libLLVM.so` has another load of dependencies, and shipping with a whole bunch of LLVM libs just for LTO doesn't seem like an attractive option to me, so relying on a compatible external `LLVMgold.so` there too makes the most sense to me.

JohanEngelen:
It's probably also good to add the path to the plugin in the LLVM installation to the search locations for the plugin.  (note that in that case, the LDC release packages will contain a search location corresponding to the location of LLVM when the release package was built...)

kalev:
Thanks for the fixes. I would search /usr/lib64/LLVMgold.so first and /usr/lib/LLVMgold.so second. This way, a 64 bit ldc installed on a multilib Fedora that has both 32 bit and 64 bit llvm libraries installed would find the right LLVMgold.so first.

kalev:
> It may or may not work so well when, e.g. with LDC built with LLVM4.0, generating LLVM4.0 IR+plugin options and putting that into a LLVM3.8 LTO plugin. That's another big reason why I added it to LDC's package.

That shouldn't be a problem at all for Fedora (and for other linux distros as well, I assume) since everything that links with llvm shared libraries gets rebuilt when llvm gets updated.

JohanEngelen:
Thanks @kalev , I've added the search paths.

(btw, I am mostly thinking about manual builds of LDC with newer LLVM versions, possibly not the system LLVM version. That person will just have to use `-DLDC_INSTALL_LTOPLUGIN=ON`.)

JohanEngelen:
ThinLTO doesn't work with mixed versions because it's being actively developed and the module summary format is changing: you'll see a version mismatch error.
Full LTO is working I think, because the binary format of LLVM IR did not change. I believe there are checks for that, so if it stops to work, you'll see an LLVM error I think.

Using 3.9 plugin with 4.0(trunk) LDC gives:
```
❯ ldc2 -flto=thin -O3 ./tests/linking/inputs/thinlto_ctor.d -run ./tests/linking/thinlto_modulecdtors_2.d
Invalid summary version 2, 1 expected
error: can't create ModuleSummaryIndexObjectFile for buffer: Corrupted bitcode
clang: error: unable to execute command: Segmentation fault: 11
clang: error: linker command failed due to signal (use -v to see invocation)
Error: /usr/bin/gcc failed with status: 254
```
but `-flto=full` works.
The error message could be improved, but that's LLVM territory.

Same test with 4.0 plugin and 3.9 LDC works (thinlto).

JohanEngelen:
> So should we ship the library with the Linux binary releases?

I think this is not possible: on Linux, the plugin depends on `/usr/include/plugin-api.h` which is probably different for different distributions.

kinke:
Sounds good, let's rely on an external plugin on Linux. That `plugin-api.h` header is part of `binutils` btw. That whole LTO plugin mess is hopefully only a temporary solution until the LLVM linker `lld` can be used.

kinke:
Btw, from https://reviews.llvm.org/rL288487:

> LLD used to take 11.73 seconds to link Clang. Now it is 6.94 seconds. MSVC link takes 83.02 seconds.

And from https://reviews.llvm.org/rL288373:

> As a result, LLD can now link a Chromium binary (output size 1.59 GB) in 10.28 seconds on my machine with ICF enabled. Compared to gold which takes 40.94 seconds to do the same thing, this is an amazing number.

Definitely interesting.

Issue #1904, opened 2016-11-29
'Invalid operand types for ICmp instruction' on ppc64


kalev:
On ppc64, ldc 1.1.0 beta4 builds fine when boostrapping with older ldc 0.17.2. When building 1.1.0 beta4 with itself, I get the following error. This is with llvm 3.9 on Fedora rawhide.

Full build log available at https://kojipkgs.fedoraproject.org//work/tasks/338/16670338/build.log (short lived link)

```
[ 11%] Generating src/core/math.o
cd /builddir/build/BUILD/ldc-1.1.0-beta4-src && /builddir/build/BUILD/ldc-1.1.0-beta4-src/build/bin/ldc2 --output-o -c -I/builddir/build/BUILD/ldc-1.1.0-beta4-src/runtime/druntime/src -I/builddir/build/BUILD/ldc-1.1.0-beta4-src/runtime/druntime/src/gc /builddir/build/BUILD/ldc-1.1.0-beta4-src/runtime/druntime/src/core/math.d -of/builddir/build/BUILD/ldc-1.1.0-beta4-src/build/runtime/src/core/math.o -w -relocation-model=pic -O3 -release -disable-invariants -d-version=Shared
#0 0x00003fff7b63a478 llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/lib64/libLLVM-3.9.so+0x84a478)
#1 0x00003fff7b63a958 (/lib64/libLLVM-3.9.so+0x84a958)
#2 0x00003fff7b638124 llvm::sys::RunSignalHandlers() (/lib64/libLLVM-3.9.so+0x848124)
#3 0x00003fff7b638304 (/lib64/libLLVM-3.9.so+0x848304)
#4 0x00003fff7dc004d8 (linux-vdso64.so.1+0x4d8)
#5 0x0000000010060b24 RootObject::compare(RootObject*) (/builddir/build/BUILD/ldc-1.1.0-beta4-src/build/bin/ldc2+0x10060b24)
#6 0x00000100082e6350
#7 0x00000000102f07d0 bool __gnu_cxx::__ops::_Iter_comp_iter<bool (*)(CaseStatement*, CaseStatement*)>::operator()<CaseStatement**, CaseStatement**>(CaseStatement**, CaseStatement**) /usr/include/c++/6.2.1/bits/predefined_ops.h:123:0
[...]

PR #1905, opened 2016-11-29
Attempt to fix PPC/PPC64 ABI issues

kinke:
Continuation of https://github.com/ldc-developers/ldc/pull/1683.


kinke:
Kai mentioned that LDC was able to compile itself back then, but that one test still failed. I assume it's the one wrt. reversed sorting, which was also a problem for ARM until the arguments order has been reversed for `extern(D)`.

kinke:
Note that https://github.com/ldc-developers/ldc/pull/1680 was merged, i.e., Kai's fixes are in ltsmaster, but not in master.

gen/abi-mips64.cpp
```diff
@@ -44,6 +48,10 @@ struct MIPS64TargetABI : TargetABI {
   }
 
   void rewriteFunctionType(TypeFunction *tf, IrFuncTy &fty) override {
+    if (!fty.ret->byref) {
+      rewriteArgument(fty, *fty.ret);
+    }
```
kinke: Note that `MIPS64TargetABI::rewriteArgument()` is empty and needs to be implemented, so this ABI has probably never been properly tested. Almost all ABIs rewrite a non-sret return value just like a regular argument of the same type.

kinke:
@joakim-noah, @smolt: It would be nice if you could verify everything still works.

joakim-noah:
Compiling ltsmaster natively on Android/ARM with the ARM portion of this PR, then will try the full PR with master.  That'll cover ARM soft-float, somebody with an armhf board will have to test that separately.

kalev:
I tested the pull request and it fixes the ppc64 issue, https://github.com/ldc-developers/ldc/issues/1904 for me. Thanks!

I ran it through Fedora's armv7hl, x86_64, ppc64le, ppc64, i686 builders and ldc can successfully build itself on all of these now.

kinke:
Excellent, thank you very much!

joakim-noah:
Full stdlib and dmd testsuite passes with this PR on Android with ltsmaster and master.

kinke:
See https://github.com/ldc-developers/ldc/pull/865 and https://github.com/dlang/dmd/pull/5232 as to why druntime and the compiler itself currently depend on the reversed order.

PR #1910, opened 2016-12-03
Fix TypeInfo_Struct constants for all non-x86_64 64-bit targets

kinke:
Fields `m_arg{1,2}` are only declared for x86_64 targets, so there was a size mismatch. In the best case, this lead to `m_RTInfo` containing a TypeInfo pointer instead of the `hasPointers` flag in the least significant bit (=> no `hasPointers`). `hasPointers` is also specified in `m_flags` though, so I'm not sure `m_RTInfo` is actually used.


kinke:
The beauty was that the member count check applied for x86_64 only (correct), but adding the 2 additional constants was only guarded by a check whether the target is any 64-bit one. ;)
Too much casting hid this error until now I guess (incl. no out-of-bounds memcpy etc), so the bug would presumably have manifested only for precise GC.

PR #1911, opened 2016-12-03
Refactor code to hide direct IrFunction->func usage and add convenient functions

Hardcode84:
These changes are required by my current runtimeCompile work but I think they can be useful by itself. Main idea is to replace direct IrFunction->func usage with getLLFunc in all places where we want to do a call or take function address. So if we need to replace function call with call to another function we only need to hijack getLLFunc. Also a bunch of other small refactorings.

### DIFF
--- a/ir/irfunction.h
+++ b/ir/irfunction.h
@@ -34,7 +34,25 @@ struct IrFunction {
   void setNeverInline();
   void setAlwaysInline();
 
-  llvm::Function *func = nullptr;
+  void setLLVMFunc(llvm::Function *function);
+
+  /// Get 'real' llvm function, this function should be used if your need to
+  /// access to function IR or attributes
+  llvm::Function *getLLVMFunc() const;
+  llvm::CallingConv::ID getCallingConv() const;
+  llvm::FunctionType *getLLVMFuncType() const;
+
+#if LDC_LLVM_VER >= 307
+  bool hasLLVMPersonalityFn() const;
+  void setLLVMPersonalityFn(llvm::Constant *personality);
+#endif
+
+  llvm::StringRef getLLVMFuncName() const;
+
+  /// Get llvm function suitable for a calling, this function should be used
+  /// if you need to make a call or to take function address
+  llvm::Function *getCallee() const;
+
   FuncDeclaration *decl = nullptr;
   TypeFunction *type = nullptr;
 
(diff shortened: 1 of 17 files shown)


ir/irfunction.h
```diff
@@ -75,5 +75,17 @@ struct IrFunction {
...
+llvm::Function *getLLFunc(FuncDeclaration *decl, bool create = false);
+
+llvm::StringRef getLLFuncName(FuncDeclaration *decl);
+#if LDC_LLVM_VER >= 308
+void setLLFuncSubprogram(FuncDeclaration *decl, llvm::DISubprogram *SP);
+#endif
+llvm::FunctionType *getLLFuncType(IrFunction *irFunc);
+#if LDC_LLVM_VER >= 307
+bool hasLLFuncPersonalityFn(const IrFunction *irFunc);
+void setLLFuncPersonalityFn(IrFunction *irFunc, llvm::Constant *personality);
```
kinke: Sorry for not providing any feedback so far. ;)
Are all of these starting with `getLLFuncName()` required too, or could they be used just as well via `getLLFunc(decl)->getName()`?

JohanEngelen:
Can you come up with a more descriptive name than `getLLFunc`?  Iiuc, after your JIT work is added, `IrFunction` will have two functions: the function with the user's code and the JIT shell around it. I've tried to come up with a better name but haven't found one yet.
(also, I like `getLLVMFunction` better)

ir/irfunction.h
Hardcode84: The idea was to use getLLFunc only in places where we want to call a function or take it address. With my feature it can return jit wrapper instead of real function and getName will  return unexpected result.

Hardcode84:
Yeah, `getLLFunc` may not be the best name. Ideally this name should reflect it main usage (creating calls). But with `getLLVMFunction` other functions names will become way too verbose from my opinion: `getLLVMFunctionName()`, `setLLVMFunctionSubprogram()`.

JohanEngelen:
Verbosity of function names is great (LLVM's source is great to read, DMD's isn't).

Hardcode84:
Any more comments except functions naming?

JohanEngelen:
Otherwise LGTM.

(The `assert(nullptr != func);` is a bit awkward and backwards to me. `assert(func != nullptr);` or `assert(func);`?)

kinke:
I'd rather see the right place for the JIT abstraction in `IrFunction` by hiding `func` and providing the corresponding `getLLFunc*()` etc. methods there, otherwise I see a big potential for `IrFunction::func` being used some day instead of the new free functions, which could remain as convenience forwarders to the new `IrFunction` methods.

Hardcode84:
Thanks, I'll try to update this PR tomorrow.

Hardcode84:
Updated. `IrFunction->func` is now private. There are now `getLLVMFunc()` which which should be used when we want to access to original function (in `DtoDefineFunction` and `applyFuncDeclUDAs`) and `getCallee()` which should be used to create call to function or get it address.

kinke:
Thanks, looks much better to me now.

As to final nitpicks, LDC uses a lot of free glue functions `Dto*()` taking some front-end arguments such as `FuncDeclaration`s and returning LLVM (or LDC-IR) types, so I'd suggest renaming `getCallee()` to `DtoCallee()`, `getLLVMFunc()` to `DtoFunc()` etc.

Hardcode84:
Renamed free functions to Dto*, added comments.

kinke:
Thanks. Please check whether you're okay with the commit I added (to be squashed when merging).

Hardcode84:
Looks good, Thanks.

Hardcode84:
@kinke can you please merge this PR. After that I'll be able to create PR with current state of my jit work.

kinke:
I've been waiting for another member to approve (Johan, David?). I'll merge tomorrow if there's no more feedback. Thanks for bearing with the delays. ;)

JohanEngelen:
Let's see that JIT stuff!

PR #1912, opened 2016-12-04
CMake cleanup and `-L-Wl,...` fix

kinke:


kinke:
After this I think we're good to go with beta6.

kinke:
This only makes sure the suffices are consistent: `<module>[-debug][_<MULTILIB_SUFFIX>].o`, `lib<[MULTI]LIB_SUFFIX>/libdruntime-ldc[-unittest][-debug].a`, `make` target `druntime-ldc[-unittest][-debug][_<MULTILIB_SUFFIX>]`, CTest tests `<module>[-debug][_<MULTILIB_SUFFIX>]` etc. In practice, this only leads to a consistent `-32` => `_32` (=> consistent regex matching) and doesn't tamper with the logic that the (32-bit) multilib tests are only generated for 64-bit hosts.

kinke:
As to actual fixes, the first 2 commits are indeed fixes, the second one addressing a reported (but misunderstood IIRC) issue wrt. `-L-Wl,...` being unusable (=> `-Xlinker -Wl,...` error). The only fix in the 3rd cleanup commit is the missing test dependency.

kinke:
One issue left - no `MULTILIB_SUFFIX` in dmd-testsuite CMake script.

kinke:
Merging as reviewing this thing isn't something I'd be keen on either. ;) David's concerns wrt. the suffix aren't well-founded IMO, as `MULTILIB_SUFFIX` [is](https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L26) either `32` or `64` (if not hacked by the user), so the only change in that regard is a consistent underscore instead of a mix of underscore and dash (which I found annoying when doing a multilib build on Windows a few months ago, which requires building the 64-bit build targets and tests in a 64-bit shell, and the 32-bit ones in a 32-bit shell).

kinke:
Believe me, it *is* a fix. E.g., `-L-Wl,-force_load,libdruntime-ldc.a` did not work, making perfect sense: `ldc2 -L-Wl,-force_load,libdruntime-ldc.a` => `gcc -Xlinker -Wl,-force_load,libdruntime-ldc.a` => `ld -Wl,-force_load,libdruntime-ldc.a`.
It is now `gcc -Wl,-force_load,libdruntime-ldc.a` => `ld -force_load libdruntime-ldc.a`, and that works.

kinke:
One can't concatenate multiple args via commas using `-L` unless I totally misunderstood something. `ldc2 -Larg1,arg2` => `gcc -Xlinker arg1,arg2` => `ld arg1,arg2`.

PR #1914, opened 2016-12-07
Remove the 16MB limit on static array sizes.

JohanEngelen:
The limit is a workaround for an optlink bug, thus unimportant for LDC, see
https://issues.dlang.org/show_bug.cgi?id=14859
https://forum.dlang.org/post/lxzyzcatydfkxkoxfccw@forum.dlang.org


JohanEngelen:
The DMD workaround is there to avoid a "cannot inline core.checkedint.mulu" error.

JohanEngelen:
A simpler change would be to just increase the limit to, say, 16TB. That would keep the diff with upstream supersmall.

Issue #1915, opened 2016-12-08
Missing etc.linux.memoryerror.registerMemoryErrorHandler


Kozzi11:
When trying to compile workspace-d with ldc I get this error:

`undefined reference to _D3etc5linux11memoryerror26registerMemoryErrorHandlerFZb`

LDC libs:
`nm -D /usr/lib/libldruntime.so  | grep registerMemory`
`nm -D /usr/lib/liblphobos2.so  | grep registerMemory`

DMD libs:
`nm -D /usr/lib/libphobos2.so  | grep registerMemory
00000000004bd8b8 T _D3etc5linux11memoryerror26registerMemoryErrorHandlerFZb
00000000004bd910 T _D3etc5linux11memoryerror28deregisterMemoryErrorHandlerFZb`

kinke:
Thanks, confirmed, it's a more general issue with our CMake script having an outdated whitelist of directories from which the runtime modules are pulled in. I'll work on a more robust black-list approach instead.

kinke:
That particular `etc.linux.memoryerror` module causes additional, non-trivial trouble, see #1917. I'm pretty sure I recently read something about it being scheduled for deprecation (from Walter), so I won't put any more effort into it.

0xEAB:
just run into this, doesn’t seem deprecated yet

kinke:
This has been reworked in druntmie v2.111: https://github.com/dlang/dmd/pull/20643

The new `registerMemoryAssertHandler()` function works with all compilers, and on most architectures. Will land with #4877.

PR #1917, opened 2016-12-08
CMake: Include missing modules in druntime/Phobos

kinke:
Fixes issue #1915. Using exclusively blacklists and no more whitelists (especially for subdirectories) should be way more robust wrt. future restructurings.
I also got rid of obsolete existing blacklist entries and removed a hopefully obsolete workaround wrt. disabling invariants for some parts of druntime.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -58,6 +58,10 @@ file(GLOB_RECURSE DRUNTIME_D ${RUNTIME_DIR}/src/*.d)
 list(REMOVE_ITEM DRUNTIME_D ${RUNTIME_DIR}/src/test_runner.d)
 # remove unsupported etc/linux/memoryerror.d (see issue #1915)
 list(REMOVE_ITEM DRUNTIME_D ${RUNTIME_DIR}/src/etc/linux/memoryerror.d)
+# FIXME: Remove all modules in core/stdcpp/ due to alpha quality.
+#        See PR #1917; partly fixed upstream.
+file(GLOB_RECURSE DRUNTIME_D_STDCPP ${RUNTIME_DIR}/src/core/stdcpp/*.d)
+list(REMOVE_ITEM DRUNTIME_D ${DRUNTIME_D_STDCPP})
 # remove all modules in gcstub/
 file(GLOB_RECURSE DRUNTIME_D_GCSTUB ${RUNTIME_DIR}/src/gcstub/*.d)
 list(REMOVE_ITEM DRUNTIME_D ${DRUNTIME_D_GCSTUB})
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -1,6 +1,6 @@
 project(runtime)
 
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8.5)
 
 #
 # Main configuration.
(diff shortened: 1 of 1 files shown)


kinke:
I'm kind of reluctant of adding that particular module as hack for the release just to close that issue early. On Win64, we are talking about 18 additional modules for the `all` target, i.e., a total of 9 for druntime/Phobos (debug + release), incl. `core.stdcpp.*` etc.

kinke:
Reason for the (current) CI errors: `core.stdcpp.typeinfo` contains a C-style static array declaration leading to a warning, thus the object file isn't generated. The declaration is fixed upstream.

kinke:
After excluding `core.stdcpp.*` due to misc. issues, the only new module causing trouble is - of course - the requested `etc.linux.memoryerror`. ;)

kinke:
See https://github.com/dlang/druntime/pull/1709 wrt. `core.stdcpp.*`.

runtime/CMakeLists.txt
```diff
@@ -69,120 +72,85 @@ list(REMOVE_ITEM DCRT_D
...
     if(APPLE)
-        list(APPEND CORE_D_SYS ${CORE_D_OSX})
+        list(APPEND DRUNTIME_D ${DRUNTIME_D_OSX})
+    elseif(${CMAKE_SYSTEM} MATCHES "FreeBSD")
+        list(APPEND DRUNTIME_D ${DRUNTIME_D_FREEBSD})
+    elseif(${CMAKE_SYSTEM} MATCHES "Linux")
+        list(APPEND DRUNTIME_D ${DRUNTIME_D_LINUX})
+    elseif(${CMAKE_SYSTEM} MATCHES "SunOS")
+        list(APPEND DRUNTIME_D ${DRUNTIME_D_SOLARIS})
+    # TODO: Android (Bionic)
```
kinke: @joakim-noah: Can you help out here wrt. detecting a native CMake build on Android?

kinke:
Wrt. the `etc.linux.memoryerror` unittest, it fails when throwing a new `NullPointerError` in `sigsegvUserspaceProcess()` after the SIGSEGV signal: `Fatal error in EH code: _Unwind_RaiseException failed with reason code: 5` (END_OF_STACK). Putting the code into a `main()` function instead of a unittest block doesn't change anything.

runtime/CMakeLists.txt
joakim-noah: I don't think CMake supports such builds well, as it doesn't have any built-in support for native Android compiles and [the package I'm using is just compiled similar to linux/ARM](https://github.com/termux/termux-packages/tree/master/packages/cmake).  So `CMAKE_SYSTEM_NAME` and `CMAKE_SYSTEM` are set to the same as on any linux, ie `Linux` and the same with the kernel version appended.

I suggest you just add the bionic folder for linux too.  It will not affect linux/glibc because its declarations are versioned out and Android isn't bothered by both `linux/` and `bionic/` being compiled.

kinke: Thanks. You do set CMAKE_SYSTEM_NAME to `Android` in your `build.sh` script though, but if that isn't standard, we may as well just include the 2 Bionic modules for Linux generally. My only worries were that they would then show up as tested as part of the unittests, although they are versioned out. But oh well, they only contain a single declaration each... ;)

kinke:
When radically reducing the signal handler to
```D
void sigsegvDataHandler()
{
    asm
    {
        naked;
        push RSI;   // return address (original RIP).
        // Parameter address is already set as RDI.
        jmp sigsegvUserspaceProcess;
    }
}
```
the exception is thrown correctly.
This then works:
```D
import etc.linux.memoryerror, core.stdc.stdio;

int* getNull() { return null; }
void setNull() { *getNull() = 42; }

void main()
{
    registerMemoryErrorHandler();

    try
    {
        //*getNull() = 42;
	setNull();
    }
    catch (NullPointerError)
    {
        printf("Caught\n");
    }

    deregisterMemoryErrorHandler();
}
```
But the commented-out code doesn't (unhandled `NullPointerError`), as it isn't `invoke`d like the function call, it's just an assignment happening to result in a thrown `Error`.

kinke:
As I recall reading somewhere that Walter plans to deprecate `etc.linux.memoryerror`, I simply keep on excluding that module now in this PR.
`https://github.com/dlang/druntime/pull/1709` has been merged, meaning that we should soon be able to include the `core.stdcpp.*` modules on Windows without running into linking errors for the druntime testrunners.

PR #1919, opened 2016-12-13
Refactor the machinery to add commandline options that clash with LLVM's options [NFC]

JohanEngelen:
This generalizes the piece of code that adds the "enable-color" option, such that it is easier to add other LLVM-clashing options (such as -ffast-math, #1896 ).

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -421,18 +421,6 @@ cl::opt<unsigned, true> nestedTemplateDepth(
         "(experimental) set maximum number of nested template instantiations"),
     cl::location(global.params.nestedTmpl), cl::init(500));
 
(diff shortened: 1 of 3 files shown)


thewilsonator:
Looks good. The only comment I have is that `cl::createClashingOptions` doesn't make it immediately obvious that it also hides a bunch of llvm options.

JohanEngelen:
> The only comment I have is that cl::createClashingOptions doesn't make it immediately obvious that it also hides a bunch of llvm options.

Before, it was the other way around:  `hideOptions` would also call `createOptions`. I replaced one dragon with another :(. I'll revert to a previous iteration where I just had two independent separate functions.

thewilsonator:
LGTM.

PR #1926, opened 2016-12-18
SCOPE_EXIT helper

Hardcode84:
SCOPE_EXIT helper and usage example in functions.cpp.
My SCOPE_EXIT helper that was discussed in gitter.


gen/functions.cpp
```diff
@@ -25,6 +25,7 @@
 #include "gen/dvalue.h"
 #include "gen/funcgenstate.h"
 #include "gen/function-inlining.h"
+#include "gen/helpers.h"
```
ZombineDev: `scope_guard.h` would be a more meaningful name.

gen/helpers.h
```diff
@@ -0,0 +1,84 @@
...
+
+  operator bool() const {
+    return flag;
+  }
+};
+
+template<typename Func>
+struct ScopeExit {
+  Func func;
+  Ownership active = false;
```
ZombineDev: Why is this necessary? It seems to me that if you remove it, the class will work the same way.

gen/helpers.h
```diff
@@ -0,0 +1,84 @@
...
+  return ScopeExit<typename std::decay<Func>::type>(std::forward<Func>(func));
+}
+
+}
+
+#define LDC_STRINGIZE2(a,b) a##b
+#define LDC_STRINGIZE(a,b) LDC_STRINGIZE2(a,b)
+#define LDC_UNNAME_VAR(basename) LDC_STRINGIZE(basename, __LINE__)
+
+#define SCOPE_EXIT auto LDC_UNNAME_VAR(scope_exit) = details::ScopeExitTag{} << [&]()
```
ZombineDev: Personally, I think this going a bit to far, if it's just for the syntax sugar. And I say this, after having watched Alexandrescu's talk.

gen/helpers.h
Hardcode84: We are returning ScopeExit struct from operator<< function and there can be a temporary copy which shouldn't call lambda.

gen/helpers.h
JohanEngelen: I also think this is going too far, but am OK with keeping it.

gen/functions.cpp
JohanEngelen: @Hardcode84 Can you rename the file to `scope_exit.h`?  (or `scope_guard` if you change the class name too)    I'll merge after.

Hardcode84: Done

JohanEngelen:
Thanks @Hardcode84

PR #1927, opened 2016-12-18
Fix #1924

### LINKED ISSUE #1924: returned array-of-array literals share address
```D
auto make2d()
{
    return [[0.0]];
}

void main()
{
    auto a = make2d();
    auto b = make2d();
    assert (a.ptr !is b.ptr);
    assert (a[0].ptr !is b[0].ptr); // FAILS!!!
}
```

I'm 99% sure this bug wasn't present in 1.1.0-b2, but it's definitely there in b6. Seems likely to have been caused by https://github.com/ldc-developers/ldc/pull/1821/files

This is a pretty nasty bug that could cause all manner of pain. It currently breaks https://github.com/DlangScience/dstats tests.

JohanEngelen:
Back off on what we consider as "constant literals": (some) dynamic array literals are not-const because they are expected to be newly allocated.

~~There is room for optimization, but it's non-trivial.~~ Fixed (needed for our testsuite!).

Resolves #1924

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -496,7 +496,7 @@ Expression *indexArrayLiteral(ArrayLiteralExp *ale, unsigned idx) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-bool isConstLiteral(Expression *e) {
+bool isConstLiteral(Expression *e, bool immutableType) {
   // We have to check the return value of isConst specifically for '1',
   // as SymOffExp is classified as '2' and the address of a local variable is
   // not an LLVM constant.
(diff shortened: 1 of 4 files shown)


John-Colvin:
`const a = [1,2,3];` should be a constant, just like `immutable` would.

JohanEngelen:
> Ideally const a = [1,2,3]; should be a constant, just like immutable would.

Yep. Probably as easy as
```diff
- if (arrayType->isImmutable() && isConstLiteral(e, true)) {
+ if ((arrayType->isImmutable() || arrayType->isConst()) && isConstLiteral(e, true)) {
```
in `toir.cpp`.
But that's for future work. Let's not introduce a new bug while trying to get this fixed! ;)

John-Colvin:
Here's one ldc currently gets right, but should probably have a test for nonetheless:
```D
mixin template A()
{
	auto a = [1, 2, 3];
}

void main()
{
	mixin A!() a0;
	mixin A!() a1;
	assert (a0.a.ptr !is a1.a.ptr);
}
```

JohanEngelen:
Thanks, added the testcase.

JohanEngelen:
close+reopen to trigger CircleCI

JohanEngelen:
cherry-picked onto release-1.1.0 too

PR #1932, opened 2016-12-19
Add ldc.exception.getCurrentException

JohanEngelen:
This adds functionality to retrieve the currently active exception. (functionality requested by Weka)

`ldc.exception.getCurrentException` returns the current "active" exception, or null when no exception is active.
An exception is considered active from when it is thrown until its catch
handler exits. Example:
```d
static import ldc.exception;
void testexample()
{
    void checkException(FooException ex)
    {
        assert(ex is ldc.exception.getCurrentException());
    }
    try
    {
        checkException(null);
        auto ex = new FooException;
        scope (exit)
            checkException(ex);
        throw ex;
    }
    catch (FooException ex)
    {
        checkException(ex);
    }
}
```

This is implemented by:
- calling `_d_eh_exit_catch` (new function) at the end of catch handlers (like C++'s `__cxa_end_catch`).
- Adding exceptions to a `currentExceptionStack` when they are thrown.
- Removing exceptions from the stack when the catch handler exits.
There is some additional logic to make things work well when chaining exceptions and other fun complications; see the eh_getcurrent.d test file.


JohanEngelen:
(close + reopen to retrigger CircleCI)

kinke:
MSVC already uses a stack IIRC.
You may need to adjust the stack on fiber context switches too (`_d_eh_swapContext()`).

JohanEngelen:
Yeah, it's a work in progress.

PR #1936, opened 2016-12-20
Fix LLVM trunk build.

JohanEngelen:


rainers:
In addition, DIBuilder::createGlobalVariable has been renamed to createGlobalVariableExpression

JohanEngelen:
Ah, indeed, more LLVM refactorings are happening today.

PR #1937, opened 2016-12-20
Sync command-line options and descriptions with DMD

### LINKED ISSUE #1923: Description of -w/-wi flags is misleading
Using LDC 1.1.0 Beta6

It looks like when compiling a D file that has warnings, supplying the -w flag will make the compilation fail - AFAIR this hasn't happened with previous releases, and the documentation suggests that this will only show warnings and not stop the compilation.

So, is the documentation or the LDC behavior wrong?

kinke:
Resolves #1923.

### DIFF
--- a/driver/ldmd.cpp
+++ b/driver/ldmd.cpp
@@ -154,6 +154,7 @@ Usage:\n\
   @cmdfile       read arguments from cmdfile\n\
   -allinst       generate code for all template instantiations\n\
   -betterC       omit generating some runtime information and helper functions\n\
+  -boundscheck=[on|safeonly|off]   bounds checks on, in @safe only, or off\n\
   -c             do not link\n\
   -color[=on|off]   force colored console output on or off\n\
   -conf=path     use config file at path\n\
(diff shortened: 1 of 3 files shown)


driver/ldmd.cpp
```diff
@@ -175,53 +176,58 @@ Usage:\n\
...
+  -gc            add symbolic debug info, optimize for non D debuggers\n\
+  -gs            always emit stack frame\n"
+#if 0
+"  -gx            add stack stomp code\n"
+#endif
+"  -H             generate 'header' file\n\
   -Hddirectory   write 'header' file to directory\n\
   -Hffilename    write 'header' file to filename\n\
-  --help         print help\n\
+  --help         print help and exit\n\
```
wilzbach: DMD supports `-h` as well since 2.072

kinke: I don't see it in the DMD 2.072.2 usage output and neither in [current source](https://github.com/dlang/dmd/blob/master/src/mars.d#L138)...

wilzbach: > I don't see it in the DMD 2.072.2 usage output and neither in current source. But there have been some more tweaks in 2.072.

Yeah it's hidden, because Walter and Andrei were a bit against it, but they couldn't say no against printing the help page in case someone typed `-h` as it's common behavior in the Unix world. So I am not sure whether you want to support/display it or not.

The current `dmd` frontend does accept `-h`:

https://github.com/dlang/dmd/blob/master/src/mars.d#L913

And the PRs with the discussion:

https://github.com/dlang/dmd/pull/5844/files
https://github.com/dlang/dmd/pull/5806

kinke: Ah thanks, but the 2.072 changes will be reflected in a later PR, this one is still aimed for LDC 1.1.

Issue #1941, opened 2016-12-23
Issue with main.cpp::parseCommandLine


xiren7:
When command line arguments of ldc2 passed by file, via something like "ldc2 @path/to/file.rsp" used by dub.

The two function called by parseCommandLine would not get the real "argc" and "argv", only get "argv[1] == @path/to/file.rsp":

```C
  const char *explicitConfFile = tryGetExplicitConfFile(argc, argv);
  std::string cfg_triple = tryGetExplicitTriple(argc, argv).getTriple();
```

So the actual problem is when ldc2 used with dub, if the host system is x86_64, and when run this command "dub -a=x86", the tryGetExplicitTriple will always return x86_64(via default value) instead of i686 triple.

kinke:
Thanks for the detailed analysis, fixed now.

PR #1942, opened 2016-12-23
Improve support for response files in command line

kinke:
Fix issue #1941 by expanding response files **before** temptatively parsing an explicit `-conf=` and/or target triple for the config file lookup. After injecting the switches from the config file right after argv[0], expand again to catch any response files specified in the config file switches.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -231,48 +231,50 @@ void initFromPathString(const char *&dest, const cl::opt<std::string> &src) {
   }
 }
 
-const char *tryGetExplicitConfFile(int argc, char **argv) {
-  // begin at the back => use latest -conf= specification
-  for (int i = argc - 1; i >= 1; --i) {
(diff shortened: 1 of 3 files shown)


JohanEngelen:
Any way to test this?
(perhaps with `LOG` commands and checking `-vv` output)

kinke:
A test without additional logging would require 3 files - a dummy D source, a separate config file with different switches than the default one, and a response file. I don't think that's worth it, especially since this command line parsing code isn't subject to frequent changes.

JohanEngelen:
IMO, every bug fix needs a test. The number of files is not a very strong argument against testing.
Because this code is looked at very infrequently, knowledge about it is lost quickly and then it is easily broken by innocent changes...

kinke:
With the additional `-v` output in #1950, the additional config file isn't required anymore, and I can get away with D source + response file.

PR #1952, opened 2017-01-06
[dcompute] irtypestruct

thewilsonator:
Transform `ldc.dcomptetypes.Pointer!(uint n,T)` into `{ T
addrspace(gIR->dcomputetarget->mapping[n])* }`

### DIFF
--- a/gen/dcomputetypes.cpp
+++ b/gen/dcomputetypes.cpp
@@ -38,13 +38,16 @@ bool isFromLDC_DComputeTypes(Dsymbol *sym) {
   return true;
 }
 
-std::pair<int, Type *> isDComputeTypesPointer(StructDeclaration *sd) {
-  if (!isFromLDC_DComputeTypes(sd) || strcmp(sd->ident->string, "Pointer"))
-    return notDComputeTypesPointer;
+DcomputePointer::DcomputePointer(StructDeclaration *sd)
+{
+  if (!isFromLDC_DComputeTypes(sd) || strcmp(sd->ident->string, "Pointer")) {
+    addrspace = -1;
+    type = nullptr;
+    return;
+  }
 
   TemplateInstance *ti = sd->isInstantiated();
-  int addrspace = isExpression((*ti->tiargs)[0])->toInteger();
-  Type *T = isType((*ti->tiargs)[1]);
-
-  return std::make_pair(addrspace, T);
+  addrspace = isExpression((*ti->tiargs)[0])->toInteger();
+  type = isType((*ti->tiargs)[1]);
 }
+
--- a/gen/dcomputetypes.cpp
+++ b/gen/dcomputetypes.cpp
@@ -39,7 +39,7 @@ bool isFromLDC_DComputeTypes(Dsymbol *sym) {
 }
 
 std::pair<int, Type *> isDComputeTypesPointer(StructDeclaration *sd) {
-  if (!isFromLDC_DComputeTypes(sd) || !strcmp(sd->ident->string, "Pointer"))
+  if (!isFromLDC_DComputeTypes(sd) || strcmp(sd->ident->string, "Pointer"))
     return notDComputeTypesPointer;
 
   TemplateInstance *ti = sd->isInstantiated();
--- a/gen/dcomputetypes.cpp
+++ b/gen/dcomputetypes.cpp
@@ -51,3 +51,8 @@ DcomputePointer::DcomputePointer(StructDeclaration *sd)
   type = isType((*ti->tiargs)[1]);
 }
 
+DcomputePointer::DcomputePointer()
+{
+  addrspace = -1;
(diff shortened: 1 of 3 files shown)


thewilsonator:
I've just had to upgrade OS and my dev tools died as a result (no make ;( ) so I apologise for abusing the CI and not doing the tests myself.

JohanEngelen:
> I apologise for abusing the CI and not doing the tests myself.

No need to apologize. Just start the title of your PRs with "[WIP]" (work in progress), and only remove that once your PR is really ready.

thewilsonator:
Its all good. It was only for 2b6abc4 . Should be good now as it passes.

thewilsonator:
Like this?

ir/irtypestruct.cpp
```diff
@@ -54,16 +54,15 @@ IrTypeStruct *IrTypeStruct::get(StructDeclaration *sd) {
 
   // For ldc.dcomptetypes.Pointer!(uint n,T),
   // emit { T addrspace(gIR->dcomputetarget->mapping[n])* }
-  std::pair<int, Type *> p;
-  if (gIR->dcomputetarget &&
-      ((p = isDComputeTypesPointer(sd)) != notDComputeTypesPointer)) {
+  DcomputePointer p(sd);
+  if (gIR->dcomputetarget && p) {
```
JohanEngelen: can you rewrite this such that it only does work when `gIR->dcomputetarget != nullptr`, like before?

JohanEngelen: (also, I think you can come up with a better name than `p` !)

thewilsonator: But naming thing is _hard_. Why do you think my project is called "DCompute"? ;)
I'm surprised you're not complaining about `x` 6 lines down, which I have changed to `body`.

But seriously, I like giving (short lived) pointer like variables names like `p`. If you want it to be something else, please edit this PR, 'cause I can't think of anything else.

gen/dcomputetypes.cpp
```diff
@@ -34,3 +37,17 @@ bool isFromLDC_DComputeTypes(Dsymbol *sym) {
     return false;
   return true;
 }
+
+DcomputePointer::DcomputePointer(StructDeclaration *sd)
+{
+  if (!isFromLDC_DComputeTypes(sd) || strcmp(sd->ident->string, "Pointer")) {
```
klickverbot: This should probably use the identifier pool instead of string comparison.

ir/irtypestruct.cpp
```diff
@@ -49,12 +52,29 @@ IrTypeStruct *IrTypeStruct::get(StructDeclaration *sd) {
...
-  builder.addAggregate(sd);
-  builder.addTailPadding(sd->structsize);
-  isaStruct(t->type)->setBody(builder.defaultTypes(), t->packed);
-  t->varGEPIndices = builder.varGEPIndices();
+  // For ldc.dcomptetypes.Pointer!(uint n,T),
+  // emit { T addrspace(gIR->dcomputetarget->mapping[n])* }
+  DcomputePointer p;
+  if (gIR->dcomputetarget && (p = DcomputePointer(sd))) {
 
+    // Translate the virtual dcompute address space into the real one for
```
klickverbot: Sometimes, a need for comments really just indicates insufficient names. This is a good example – if the variable was called `virtualToRealAddressSpaceMap` (or `->virtualToRealAS(…);`, etc.), this comment would be entirely unnecessary (although maybe still nice at the place of declaration).

gen/dcomputetypes.cpp
thewilsonator: Yes I intend to optimise all of my (ab)uses of `strcmp` at some point, but I am not sure of the best way of doing it as they are not predefs and I would need to insert them into the IdPool, but some of them are used from multiple places, and I don't want to insert them twice (or more). 
Any suggestions welcome.

thewilsonator:
That was [Johan's suggestion](https://gitter.im/ldc-developers/main?at=58a7f4ae0ad50ac3152fc456).
I intend to clean up `gen/dcomputetarget.cpp:TargetOCL::addKernelMetadata()` to use this approach (whatever it ends up being). And it would feel reasonably natural there, but I'm not particularly attached to it.

JohanEngelen:
> an llvm::Optional return value

I like it :)

thewilsonator:
Thats a good idea, I was thinking about doing something similar for "magic modules" `ldc.attributes` `ldc.dcomputetypes`,  to be able simplify and optimise `isLdcAttibutes` and `isFromLDC_DComputeTypes`.

thewilsonator:
True, will open an issue for this so we don't forget.

thewilsonator:
Any further comments?

thewilsonator:
Thanks! Will do.

PR #1953, opened 2017-01-06
[dcompute] codegen

thewilsonator:
Adapt the declaration and statement visitors for dcompute.

Don’t emit the `__initZ` symbols or typeinfo ~~and issue errors on
unsupported language concepts (classes/interfaces exceptions asm global
variables)~~(do this with a semantic analysis pass in a different PR).

Don’t always set `uwtable` when compiling `@compute` code if on non-windows x64.

Transform try finally for lowering `scope(exit);`

Implement codegen time conditional compilation with a call in an if
statement to a special function “__dcompute_reflect”


thewilsonator:
About the conditional compilation hack: its a hack, but I think it is the easiest and simplest solution to the need to select between intrinsics for different dcompute targets. The alternative option is to have proper versions (useable with `version(FOO)` statements) would require redoing  semantic analysis (and possibly parsing) of modules marked `@compute` and then doing one codegen pass for each.

This solution is
* faster. Only one pass through the frontend for each source module.
* smaller. All 20 lines of it.
* still covers all use cases. \*
* avoids a combinatorial explosion due to the way D implements `version` statements.

\* there are three cases for requirements of a feature 
1. You don't need it. No problem.
2. You absolutely require it (i.e. you can't substitute for it). Then your code won't be able to run on unsupported hardware anyway. Again, no problem.
3. It is desired because it's faster. But is only available on newer version of hardware however there exists a fallback method. You can "version out" supported and unsupported with this method.

thewilsonator:
Dont pull yet I need to figure out how to do TryFinally correctly.

gen/declarations.cpp
```diff
@@ -148,10 +148,6 @@ class CodegenVisitor : public Visitor {
                            decl->toPrettyChars());
     LOG_SCOPE
 
-    if (irs->dcomputetarget)  {
```
JohanEngelen: It it perhaps still good to add `assert(!irs->dcomputetarget)` to places like this.

gen/declarations.cpp
```diff
@@ -410,6 +420,12 @@ class CodegenVisitor : public Visitor {
     // don't touch function aliases, they don't contribute any new symbols
     if (!decl->isFuncAliasDeclaration()) {
       DtoDefineFunction(decl);
+      if (irs->dcomputetarget) {
```
klickverbot: Shouldn't this be in `DtoDefineFunction` like all the other ancillary function definition logic, or does it somehow specifically apply to being called from this visitor?

thewilsonator: I think this was the first location I tried and it worked, so possibly. I don't think I quite understand the operations of the whole `DtoResolveFunction`,`DtoDeclareFunction`,`DtoDefineFunction` pipeline. 
I'll move it there.

thewilsonator:
All green, any further comments?

gen/functions.cpp
```diff
@@ -972,7 +974,7 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   // this gets erased when the function is complete, so alignment etc does not
   // matter at all
   llvm::Instruction *allocaPoint = new llvm::AllocaInst(
-      LLType::getInt32Ty(gIR->context()), "alloca point", beginbb);
+      LLType::getInt32Ty(gIR->context()), "allocaPoint", beginbb);
```
klickverbot: Huh? This should never make it into "finished" IR anyway.

gen/statements.cpp
```diff
@@ -750,8 +772,14 @@ class ToIRVisitor : public Visitor {
     stmt->finalbody->accept(this);
     irs->DBuilder.EmitBlockEnd();
 
-    CleanupCursor cleanupBefore = irs->funcGen().scopes.currentCleanupScope();
-    irs->funcGen().scopes.pushCleanup(finallybb, irs->scopebb());
+    CleanupCursor cleanupBefore;
+    // For @compute code, don't emit any exception handling as there are no
+    // exceptions anyway.
+    const bool compute_code = !!irs->dcomputetarget;
```
klickverbot: Please stick to `camelCase` in LDC code, as upstream `dlang` does as well. (LLVM mostly uses `CamelCase` for values, which admittedly makes somewhat of a mess.)

gen/functions.cpp
thewilsonator: There was a segfault related to this with the SPIRV-LLVM that I was using, I think to do with the fact that it contained a space (I have had other problems with spaces with it) although I can't quite remember why. This fixed it, alough I'm not sure if its still a problem.

klickverbot: It doesn't matter anyway precisely because the BB is removed once codegen for a function is finished – I was just curious as to what was going on.

PR #1954, opened 2017-01-06
Dcompute main

thewilsonator:
Emit modules to the dcompute code generator if marked `@compute`

Remove version conditions for NVPTX{64} as these are now targeted by
dcompute.

Add the version specifier D_DCompute.


thewilsonator:
If there are no Objections I'll merge this tomorrow morning (GMT+8) so that I have a basis to start the lit tests (for both ldc and llvm, to make sure I haven't broken anything I haven't meant to in the process of going from 3.6.1 -> 4.0.0). 
Once I verified that all is good I'll make a release of the llvm and do a PR for ldc's cmake and CI infrastructure.

driver/main.cpp
```diff
@@ -1020,14 +1017,26 @@ void codegenModules(Modules &modules) {
       Module *const m = modules[i];
       if (global.params.verbose)
         fprintf(global.stdmsg, "code      %s\n", m->toChars());
-
-      cg.emit(m);
+      int atCompute = hasComputeAttr(m);
+      if (atCompute == 0 || atCompute == 2)
```
klickverbot: Perhaps add an `enum class DcomputeMode` (or some better name) instead of using magic numbers? `if (computeAttr != DcomputeMode::deviceOnly)` reads much, much better than `if (atCompute == 0 || atCompute == 2)`.

driver/dcomputecodegenerator.cpp
```diff
@@ -50,20 +51,20 @@ DComputeCodeGenManager::createComputeTarget(const std::string &s) {
...
-  for (int i = 0; i < dcomputeTargets.size(); i++) {
-    targets.push_back(createComputeTarget(dcomputeTargets[i]));
+  for (d_size_t i = 0; i < opts::dcomputeTargets.size(); i++) {
+    targets.push_back(createComputeTarget(opts::dcomputeTargets[i]));
   }
 }
 
 void DComputeCodeGenManager::emit(Module *m) {
-  for (int i = 0; i < targets.size(); i++) {
+  for (d_size_t i = 0; i < targets.size(); i++) {
```
JohanEngelen: can you make these range-based?
Or `for (d_size_t i = 0, e = targets.size(); i < e; i++)`, so that it is instantly clear that the targets vector is not changing in size?

driver/main.cpp
JohanEngelen: You should replace these magic numbers (0 and 2) with enum values for clarity.

driver/main.cpp
```diff
@@ -1020,14 +1017,26 @@ void codegenModules(Modules &modules) {
...
+      if (atCompute)
+        compute_modules.push_back(m);
+        
 
       if (global.errors)
         fatal();
     }
-  }
 
+    IF_LOG Logger::println("number of Modules for computecodgenning %d",
```
JohanEngelen: fix the typos in string.

driver/main.cpp
```diff
@@ -704,14 +706,6 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("SPARC64");
     registerPredefinedFloatABI("SPARC_SoftFloat", "SPARC_HardFloat");
     break;
-  case llvm::Triple::nvptx:
```
klickverbot: As far as I can see, this is largely orthogonal to DCompute – why the change?

driver/dcomputecodegenerator.cpp
```diff
@@ -50,20 +51,20 @@ DComputeCodeGenManager::createComputeTarget(const std::string &s) {
 }
 
 DComputeCodeGenManager::DComputeCodeGenManager(llvm::LLVMContext &c) : ctx(c) {
-  for (int i = 0; i < dcomputeTargets.size(); i++) {
-    targets.push_back(createComputeTarget(dcomputeTargets[i]));
+  for (d_size_t i = 0; i < opts::dcomputeTargets.size(); i++) {
```
klickverbot: Use `size_t` (`<cstdint>`) for host data. `d_size_t` is for compatibility with the D `size_t` type.

klickverbot: Also, most loops can easily be replaced with a range-based for loop.

JohanEngelen:
Nice to see we have similar comments :-)

driver/main.cpp
thewilsonator: Well, apart from through dcompute we have no proper support for nvptx. Proper support would require correct handling of addrspaces and emission of the correct magic metadata, both of which are now done through dcompute.

i.e. it didn't work before, it does work now (through dcompute) but not is a way that is compatible with just saying `-mtriple=nvptx-nvidia-cuda` . I think while harmless (since
nobody used if before) it is misleading, because if dcompute gains popularity like I hope it does,
then people may try out ldc, see that it appears to support PTX as a target and give up when it fails. I'll put it back for now, but I think that it should go at some point.

driver/main.cpp
```diff
@@ -1007,6 +1002,8 @@ void codegenModules(Modules &modules) {
   // Generate one or more object/IR/bitcode files.
```
thewilsonator: good catch, totally missed that comment.

driver/main.cpp
```diff
@@ -1045,14 +1050,29 @@ void codegenModules(Modules &modules) {
...
+        
 
       if (global.errors)
         fatal();
     }
-  }
 
+    IF_LOG Logger::println("number of Modules for DCompute: %d",
+                            compute_modules.size());
+    if (compute_modules.size()) {
```
klickverbot: Nitpick: `!computeModules.empty()`.

driver/main.cpp
```diff
@@ -1029,9 +1032,11 @@ void addDefaultVersionIdentifiers() {
 }
 
 void codegenModules(Modules &modules) {
-  // Generate one or more object/IR/bitcode files.
+  // Generate one or more object/IR/bitcode files/DCompute kernels.
   if (global.params.obj && !modules.empty()) {
     ldc::CodeGenerator cg(getGlobalContext(), global.params.oneobj);
+    DComputeCodeGenManager dccg(getGlobalContext());
+    std::vector<Module *> compute_modules;
```
klickverbot: Nitpick: `computeModules`

driver/main.cpp
```diff
@@ -1045,14 +1050,29 @@ void codegenModules(Modules &modules) {
       Module *const m = modules[i];
       if (global.params.verbose)
         fprintf(global.stdmsg, "code      %s\n", m->toChars());
-
-      cg.emit(m);
+      auto atCompute = hasComputeAttr(m);
```
klickverbot: Nitpick: `const`

thewilsonator:
This build will fail because the signature `hasComputeAttr` has changed and is used in  `LDCExtraSemanticAnalysis` and I foolishly used the web UI to resolve a merge conflict so I don't locally have the last merge commit and don't know how to get it.

thewilsonator:
See my edited comment above.

thewilsonator:
#2013

Issue #1955, opened 2017-01-06
ARM: Bus Error using synchronized blocks (1.1.0-beta6)


smolt:
1.1.0-beta6 build on Raspbian Pi 3 is getting Bus Errors in four dmd-testsuite tests:

- runnable/eh.d
- runnable/test23.d
- runnable/test34.d
- runnable/test4.d

```
LDC - the LLVM D compiler (1.1.0-beta6):
  based on DMD v2.071.2 and LLVM 3.9.1
  built with LDC - the LLVM D compiler (4a9853)
  Default target: armv7l-unknown-linux-gnueabihf
  Host CPU: (unknown)
  http://dlang.org - http://wiki.dlang.org/LDC

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_be - AArch64 (big endian)
    arm        - ARM
    arm64      - ARM64 (little endian)
    armeb      - ARM (big endian)
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
```

gdb shows all cases happen in code generated for synchronized blocks, something messed up in D_CRITICAL_SECTION, probably the pthread_mutex_t struct is misaligned.

Working back it appears the Bus Errors started in commit 8d06f23 "Allow align(1) for global variables". I'll have to study it, but my guess is that there was an existing alignment problem and this commit just revealed it.

smolt:
Ok, it is a misaligned `pthread_mutext_t` contained in a D_CRITICAL_SECTION.

Just this generates a bus error:
```D
void main()
{
  synchronized
  {
  }
}
```
Reason: ddmd/statement.d for `synchronized` creates byte array used as a D_CRITICAL_SECTION struct in druntime, and byte array gets alignment of 1 now (was 4).  CPUs with strict alignment don't like this.
```D
            /* Generate our own critical section, then rewrite as:
             *  __gshared byte[CriticalSection.sizeof] critsec;
             *  _d_criticalenter(critsec.ptr);
             *  try { body } finally { _d_criticalexit(critsec.ptr); }
             */
```
Not sure of fix yet.  Seems like there could be a `Target.critsecalignment()` similar to `Target.critsecsize()` used to set alignment of byte array.  Temporary fix would be to undo 8d06f23.

rainers:
I suspect alignment to target.ptrsize would be good enough here. There is even a pointer prepended to the D_CRITICAL_SECTION inside the byte array.

thewilsonator:
[...]

PR #1956, opened 2017-01-07
Fix alignment and size of target critical section

kinke:
The missing alignment is a front-end bug.

For POSIX targets, the critical section size was assumed to be identical to the host compiler's, which generally isn't true when cross-compiling.

`DtoMutexType()` wasn't used anywhere, so I removed it.


gen/target.cpp
```diff
@@ -64,31 +64,49 @@ unsigned Target::alignsize(Type *type) {
...
+  if (global.params.isWindows)
+    return is64bit ? 40 : 24;
+
+  // POSIX: sizeof(pthread_mutex_t)
+  const auto &triple = *global.params.targetTriple;
+  const auto arch = triple.getArch();
+  switch (triple.getOS()) {
+  case llvm::Triple::Linux:
+    if (triple.getEnvironment() == llvm::Triple::Android)
+      return Target::ptrsize;
```
kinke: druntime: [a 32-bit integer for CRuntime_Bionic](https://github.com/ldc-developers/druntime/blob/ldc/src/core/sys/posix/sys/types.d#L754)

gen/target.cpp
```diff
@@ -64,31 +64,49 @@ unsigned Target::alignsize(Type *type) {
...
+    if (arch == llvm::Triple::aarch64 || arch == llvm::Triple::aarch64_be)
+      return 48;
+    return is64bit ? 40 : 24;
+
+  case llvm::Triple::MacOSX:
+    return is64bit ? 56 : 40;
```
kinke: druntime: [+ prepended c_long](https://github.com/ldc-developers/druntime/blob/ldc/src/core/sys/posix/sys/types.d#L590) which seemed odd to me. @JohanEngelen: please investigate.

gen/target.cpp
```diff
@@ -64,31 +64,49 @@ unsigned Target::alignsize(Type *type) {
...
+
+  case llvm::Triple::FreeBSD:
+  case llvm::Triple::NetBSD:
+  case llvm::Triple::OpenBSD:
+  case llvm::Triple::DragonFly:
+    return Target::ptrsize;
+
+  case llvm::Triple::Solaris:
+    return is64bit ? 40 : 24;
```
kinke: druntime: [a 24-bytes struct](https://github.com/ldc-developers/druntime/blob/ldc/src/core/sys/posix/sys/types.d#L686). I assumed it'd be 40 bytes for 64-bit, same as Windows and Linux.

kinke:
Btw, somewhat off-topic, but apparently `pthread_mutex_t` was easily beaten by a huge margin in 2012, both on OSX and Linux: http://www.mr-edd.co.uk/blog/sad_state_of_osx_pthread_mutex_t

gen/target.cpp
JohanEngelen: I check my system header files (OS X 10.11.6, El Capitan): druntime's definition is correct.

kinke: Thanks, so overall `sizeof(pthread_mutex_t)` 64 and 44 bytes?

JohanEngelen: Indeed.

```cpp
#include <iostream>
#include <mutex>
int main() {
    std::cout << sizeof(pthread_mutex_t) << std::endl;
    return 0;
}
```

Outputs:  64.
Output when compiled with `-m32`:  44.

kinke: Thanks, amended.

kinke:
Upstream PR: https://github.com/dlang/dmd/pull/6408

kinke:
Slightly reworked (relying on druntime definitions without further assumptions from my side ;)) after the upstream PR has already been merged.

gen/target.cpp
```diff
@@ -64,30 +64,52 @@ unsigned Target::alignsize(Type *type) {
...
-    global.params.targetTriple->getOS() == llvm::Triple::NetBSD ||
-    global.params.targetTriple->getOS() == llvm::Triple::OpenBSD ||
-    global.params.targetTriple->getOS() == llvm::Triple::DragonFly
-#endif
-     ) {
-    return sizeof(size_t);
+  const bool is64bit = global.params.is64bit;
+
+  // Windows: sizeof(CRITICAL_SECTION)
+  if (global.params.isWindows)
```
JohanEngelen: Can you change this to look at the target triple instead?
Even if `isWindows` is set correctly when cross-compiling from, say, Linux to Windows, I find it much clearer to use `params.targetTriple` like you did below.

The rest looks good, nice.
(no test?)

kinke:
Test added, which shows that it doesn't work; the alignment seems to get lost somewhere on the way...

gen/target.cpp
kinke: `global.params.isWindows` is a (shorter) front-end synonym for `global.params.targetTriple->isOSWindows()`, just as `global.params.is64bit` used above is a shorthand for `global.params.targetTriple->isArch64Bit()`. As `Target` is a front-end class, I see no reason to use the clumsier explicit triple variant.

kinke:
Fixed by setting the alignment after `VarDeclaration.semantic(Scope)`, as it resets `VarDeclaration.alignment`.

gen/target.cpp
JohanEngelen: The reason is simple: clarity. Why not add it to the switch below?

JohanEngelen:
Nice proof of the importance of tests :)

gen/target.cpp
kinke: It's as clear as can be to me. The OS enum for Windows is Win32, and I hate it, as it covers Win64 as well (that's what I find unclear!). Additionally, the switch is there to figure out the target-specific `pthread_mutex_t` size on POSIX, while Windows is once again a special case.

kinke:
> Haven't we started supporting align(…) on globals? Wouldn't it be cleaner to just synthesise the corresponding AST fragment then?

I don't get your point. The front-end change does set the explicit `align(D_CRITICAL_SECTION.alignof)`, it just has to do it after semantic(), as that pass overrides it.

kinke:
> Nice proof of the importance of tests :)

Yep, once again. ;) It's a pretty important test, as it should notify us when druntime goes out of sync or support for a new platform is added to LDC.

kinke:
It may be worth thinking about aligning the D_CRITICAL_SECTION buffers at cache lines, so that ideally each critical section gets its own cache line. I'd introduce `Target::critsecalign()` and swap the field order (`struct D_CRITICAL_SECTION { Mutex mtx; D_CRITICAL_SECTION* next; }`). The 64-bit OSX size (64+8 bytes) is unfortunate for x86_64 with 64-bytes cache lines, but by putting the native mutex as first field, it'd have its own cache line and the D chain pointer the next one.

JohanEngelen:
David means to use put the `VariableDeclaration` in a `AlignDeclaration`, so that semantic analysis correctly sets the alignment, instead of modifying it after.

kinke:
Ah thanks, that does seem cleaner, amended.

gen/target.cpp
```diff
@@ -64,30 +64,52 @@ unsigned Target::alignsize(Type *type) {
...
+
+  case llvm::Triple::FreeBSD:
+  case llvm::Triple::NetBSD:
+  case llvm::Triple::OpenBSD:
+  case llvm::Triple::DragonFly:
+    return Target::ptrsize;
+
+  case llvm::Triple::Solaris:
+    return 24;
   }
```
smolt: Add `default: break;` to get rid of gcc -Wswitch warnings foreach Triple enumeration value not handled?

smolt:
Confirmed this resolves #1955, that is once it is merged into release-1.1.0 branch.

kinke:
Thanks @smolt. Any chance you can shed some light on the lit-tests issues on ARM? I'm just curious whether there are actual issues to be solved; in that case I'd try to work on those for v1.1, otherwise I'd say 1.1 can be released this month. In that regard, would you be able and willing to create the ARM release package Kai used to provide?

smolt:
> Any chance you can shed some light on the lit-tests issues on ARM? I'm just curious whether there are actual issues to be solved;

Yeah, there are only 5 lit tests that fail on linux arm, but at a glance they are not codegen issues.
```
    LDC :: codegen/align.d
    LDC :: codegen/asm_output.d
    LDC :: codegen/inlining_leakdefinitions.d
    LDC :: codegen/inlining_templates.d
    LDC :: codegen/llvm_used_1.d
```

Looks like align.d fails because arm abi rewrites struct args as vectors.  This test could be revised for arm.  Remaining tests are doing x86 asm specific stuff which could be disabled or have an arm asm path.  I can make an issue with these and provide more details.

> in that case I'd try to work on those for v1.1, otherwise I'd say 1.1 can be released this month. In that regard, would you be able and willing to create the ARM release package Kai used to provide?

Sure, are scripts to make an arm release in the ldc-developers repo?  I'd be building it on a Raspberry Pi.

JohanEngelen:
> Remaining tests are doing x86 asm specific stuff which could be disabled or have an arm asm path.

To quickly solve such issues, you can add `// REQUIRES: target_X86` and add `-mtriple=x86_64-linux-gnu` to the LDC invocation. I'll see if I can add those where they are needed. The `inlining_templates.d` test does not do X86 specific stuff though.

smolt:
> The inlining_templates.d test does not do X86 specific stuff though.

@JohanEngelen It fails with this: `/media/pi/TranquilityBase/ldc-catchup/ldc/tests/codegen/inputs/inlinables.d(100): Error: inline asm is not supported for the "armv7l" architecture`

PR #1957, opened 2017-01-07
Cache: Add files atomically to the cache (copy to temp file + rename)

JohanEngelen:
This PR prevents poisoned cache entries when LDC is aborted during writing to the cache.


JohanEngelen:
This assumes that the cache location filesystem supports renaming. I think it is justified.

JohanEngelen:
Question: should errors for adding items to the cache be fatal?

PR #1960, opened 2017-01-08
CMake: Build both static and shared sets of stdlibs

kinke:
Via `-DBUILD_SHARED_LIBS=BOTH`, addressing the build part of #1282.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -11,7 +11,7 @@ set(DMDFE_VERSION         ${D_VERSION}.${DMDFE_MINOR_VERSION}.${DMDFE_PATCH_VERS
 set(MULTILIB              OFF                                       CACHE BOOL    "Build both 32/64 bit runtime libraries")
 set(BUILD_BC_LIBS         OFF                                       CACHE BOOL    "Build the runtime as LLVM bitcode libraries")
 set(INCLUDE_INSTALL_DIR   ${CMAKE_INSTALL_PREFIX}/include/d         CACHE PATH    "Path to install D modules to")
-set(BUILD_SHARED_LIBS     OFF                                       CACHE STRING  "Whether to build the runtime as a shared library (ON|OFF|BOTH)")
+set(BUILD_SHARED_LIBS     AUTO                                      CACHE STRING  "Whether to build the runtime as a shared library (ON|OFF|BOTH)")
 set(D_FLAGS               -w                                        CACHE STRING  "Runtime build flags, separated by ;")
 set(D_FLAGS_DEBUG         -g;-link-debuglib                         CACHE STRING  "Runtime build flags (debug libraries), separated by ;")
 set(D_FLAGS_RELEASE       -O3;-release                              CACHE STRING  "Runtime build flags (release libraries), separated by ;")
@@ -31,11 +31,24 @@ else()
     set(MULTILIB_SUFFIX 64)
 endif()
 
+set(SHARED_LIBS_SUPPORTED OFF)
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux" OR
+   ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD" OR
+   APPLE)
+    set(SHARED_LIBS_SUPPORTED ON)
+endif()
+
+if(${BUILD_SHARED_LIBS} STREQUAL "AUTO")
+    if(SHARED_LIBS_SUPPORTED)
+        set(BUILD_SHARED_LIBS BOTH)
+    else()
+        set(BUILD_SHARED_LIBS OFF)
+    endif()
+endif()
+
 set(SHARED_LIB_SUFFIX "")
 if(NOT ${BUILD_SHARED_LIBS} STREQUAL "OFF")
-    if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Linux" AND
-       NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD" AND
-       NOT APPLE)
+    if(NOT SHARED_LIBS_SUPPORTED)
         message(FATAL_ERROR "Shared libraries (BUILD_SHARED_LIBS) are only supported on Linux, macOS and FreeBSD for the time being.")
     endif()
 
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -11,7 +11,7 @@ set(DMDFE_VERSION         ${D_VERSION}.${DMDFE_MINOR_VERSION}.${DMDFE_PATCH_VERS
 set(MULTILIB              OFF                                       CACHE BOOL    "Build both 32/64 bit runtime libraries")
 set(BUILD_BC_LIBS         OFF                                       CACHE BOOL    "Build the runtime as LLVM bitcode libraries")
 set(INCLUDE_INSTALL_DIR   ${CMAKE_INSTALL_PREFIX}/include/d         CACHE PATH    "Path to install D modules to")
-set(BUILD_SHARED_LIBS     OFF                                       CACHE BOOL    "Whether to build the runtime as a shared library")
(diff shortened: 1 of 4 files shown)


kinke:
With `-DBUILD_SHARED_LIBS=BOTH -DMULTILIB=ON` on Linux:

```
> ls lib lib32
lib:
libdruntime-ldc.a                       libdruntime-ldc-shared.so.2.0.71  libphobos2-ldc-debug-shared.so
libdruntime-ldc-debug.a                 libdruntime-ldc-shared.so.71      libphobos2-ldc-debug-shared.so.2.0.71
libdruntime-ldc-debug-shared.so         libldc-profile-rt.a               libphobos2-ldc-debug-shared.so.71
libdruntime-ldc-debug-shared.so.2.0.71  libldc-profile-rt-shared.a        libphobos2-ldc-shared.so
libdruntime-ldc-debug-shared.so.71      libphobos2-ldc.a                  libphobos2-ldc-shared.so.2.0.71
libdruntime-ldc-shared.so               libphobos2-ldc-debug.a            libphobos2-ldc-shared.so.71

lib32:
libdruntime-ldc.a                       libdruntime-ldc-shared.so.2.0.71  libphobos2-ldc-debug-shared.so
libdruntime-ldc-debug.a                 libdruntime-ldc-shared.so.71      libphobos2-ldc-debug-shared.so.2.0.71
libdruntime-ldc-debug-shared.so         libldc-profile-rt.a               libphobos2-ldc-debug-shared.so.71
libdruntime-ldc-debug-shared.so.2.0.71  libldc-profile-rt-shared.a        libphobos2-ldc-shared.so
libdruntime-ldc-debug-shared.so.71      libphobos2-ldc.a                  libphobos2-ldc-shared.so.2.0.71
libdruntime-ldc-shared.so               libphobos2-ldc-debug.a            libphobos2-ldc-shared.so.71
```

kinke:
This needs some testing on OSX wrt. merging the 32/64-bit libs.

kinke:
Pinging @JohanEngelen and @klickverbot - someone please test `make install` after CMake `-DBUILD_SHARED_LIBS=BOTH -DMULTILIB=ON` on OSX, in order to test merging the x86 and x86_64 libs into these 'fat' libs (that's the open WIP aspect about this PR).

kinke:
Setting it to default will double the tests (at least the unittests), so we'll need to slightly tweak the CI scripts.

If it wasn't for `-relocation-model=pic`, we wouldn't have to compile the Phobos modules (and tests!) twice. The druntime and profile-rt modules get an additional `-d-version=Shared`, so they need to be recompiled (but they are very quick in comparison anyway).

kinke:
Thanks - what's the point of these OSX 'fat' libs? Are they just a fancy way of cutting the number of lib files in half or does Apple enforce that format?

Wrt. profile-rt, static libs were a deliberate choice for `-link-sharedlib` too. Should they be shared instead or should `-d-version=Shared` be defined when compiling static `libprofile-rt-shared.a` for binaries linked to shared druntime/Phobos?

kinke:
Still awaiting an answer wrt. any advantages of OSX fat libs... ;)

> If it wasn't for -relocation-model=pic, we wouldn't have to compile the Phobos modules (and tests!) twice.

If we emit PIC by default on Linux anyway, we should exploit that and build the Phobos modules only once, greatly speeding up the build process. Making it a Linux special case will clutter the CMake m(adn)ess even more; ~~would emitting PIC by default on OSX be an option too?~~ oh, just seen that we already do that for OSX too. And LLVM defaults to PIC on Win64 as well.

kinke:
Building shared druntime/Phobos libs (incl. unittest libs and running those tests) in addition to the static libs and tests now only requires a few additional minutes on Travis.

kinke:
From the [Travis LLVM 3.8 MULTILIB=ON BUILD_SHARED_LIBS=BOTH](https://api.travis-ci.org/jobs/205992534/log.txt?deansi=true) overkill job:

```
3090/3090 Test #3090: druntime-test-shared ............................................................   Passed    2.47 sec

99% tests passed, 1 tests failed out of 3090

Total Test time (real) = 224.19 sec

The following tests FAILED:
	2859 - rt.util.typeinfo-debug-shared_32 (Failed)
Errors while running CTest


The command "ctest -j3 --output-on-failure -E "dmd-testsuite|lit-tests"" exited with 8.
store build cache

changes detected, packing new archive
.


The job exceeded the maximum time limit for jobs, and has been terminated.
```

So it only just just made it under 50 mins. :D More than 3.000 unittested druntime/Phobos modules (8x druntime/Phobos: release/debug, shared/static, x86/x86_64)...

kinke:
Travis:
```
Linking C shared library ../lib64/libdruntime-ldc-unittest-debug-shared.so
CMake Error: cmake_symlink_library: System Error: File exists
make[3]: *** [lib64/libdruntime-ldc-unittest-debug-shared.so.2.0.72] Error 1
```

Seems to happen if the job previously built shared libs too (1st run works, 2nd doesn't, 3rd does...). Any idea why the LDC build directory seems to be cached across jobs?!

kinke:
@klickverbot: Please test the merged libs on OSX again.

kinke:
I had another solution - all 9 libs hardcoded, until I realized this friggin `$<TARGET_LINKER_FILE:...>` ought to work. Damn CMake, I hate it (but sadly I also don't know a better alternative).

kinke:
Well now it looks like the merged shared libs don't work:

```
ld: warning: ignoring file /Users/travis/build/ldc-developers/ldc/lib/libdruntime-ldc-unittest-shared.dylib, file was built for x86_64 which is not the architecture being linked (i386): /Users/travis/build/ldc-developers/ldc/lib/libdruntime-ldc-unittest-shared.dylib`.
```

I won't pursue OSX further. I don't have it, Travis takes hours to finally process the OSX jobs, so someone working on OSX please fix the last issues, I'm done with this.

kinke:
> the unittest libs aren't merged

And never have been, so the unittests with `MULTILIB=ON` on OSX most likely never worked. Not tested by CI, so I won't put any more effort into it.

runtime/CMakeLists.txt
```diff
@@ -31,17 +31,31 @@ else()
...
+if(NOT ${BUILD_SHARED_LIBS} STREQUAL "OFF")
+    if(NOT SHARED_LIBS_SUPPORTED)
         message(FATAL_ERROR "Shared libraries (BUILD_SHARED_LIBS) are only supported on Linux, macOS and FreeBSD for the time being.")
     endif()
 
-    list(APPEND D_FLAGS -relocation-model=pic)
-    set(D_LIBRARY_TYPE SHARED)
-else()
-    set(D_LIBRARY_TYPE STATIC)
+    # Only use the `-shared` lib suffix if static libs are generated too
```
klickverbot: Wouldn't it be better to always use `-shared` for uniformity? (e.g. if somebody explicitly needs to link against druntime/Phobos in their custom build system, etc.)

kinke: > adding SHARED_LIB_SUFFIX (-shared) to all shared-lib targets and files for BUILD_SHARED_LIBS == ON too would complicate generating the default config file and misc. command lines in our CI scripts.

kinke: I'd suggest always adding `-shared` as soon as we have `-link-sharedlib` or something similar.

It's `-link-sharedlib` vs. `-defaultlib=druntime-ldc-shared,phobos2-ldc-shared -debuglib=druntime-ldc-debug-shared,phobos2-ldc-debug-shared`.

kinke:
As mentioned somewhere else, I suggest putting this, something like #1735 and #2016 into a single LDC release (1.3). Rationale being that they all affect the build system, and distro maintainers etc. should only have to adapt their scripts once.

kinke:
Any concerns? Otherwise I'll merge it in a couple of days. This should have been tested quite extensively by intermediate commits (with `-DBUILD_SHARED_LIBS=BOTH -DMULTILIB=ON`, on both Linux and OSX, with installing LDC etc.).

PR #1961, opened 2017-01-08
Fix -finstrument-functions. The wrong return address was used.

JohanEngelen:
Don't know whether this has ever worked, but... according to LLVM spec passing a non-zero argument to `llvm.returnaddress` is likely to be incorrect. More importantly: we need the return address of the current function (level 0), not of the calling function (level 1).

@LemonBoy did you ever test whether `uftrace` gives sane output for LDC programs?


JohanEngelen:
(Note: with this PR, our IR is identical to clang++'s output)

Issue #1933, opened 2016-12-19
Error: piece is larger than or outside of variable   call void @llvm.dbg.value(metadata i64 %29, i64 0, metadata !1479, metadata !7942), !dbg !7923


ximion:
Using LDC 1.1 Beta 6, all packages from Debian Unstable.

When compiling a certain part of Vibe.d, LDC fails with
```
piece is larger than or outside of variable
  call void @llvm.dbg.declare(metadata %router.MatchGraphBuilder.TerminalTag* %.frame.sroa.3, metadata !106, metadata !400), !dbg !401
!106 = !DILocalVariable(name: "builder", scope: !105, file: !4, line: 18, type: !51)
!400 = !DIExpression(DW_OP_bit_piece, 64, 192)
LLVM ERROR: Broken module found, compilation aborted!
```

A minimal testcase for this bug (likely a regression since this worked before) is attached.
[testcase.tar.gz](https://github.com/ldc-developers/ldc/files/662145/testcase.tar.gz)

This issue only happens when compiling with optimization.

kinke:
Thanks for the nice small testcase! I don't have time for this though atm...

JohanEngelen:
I cannot reproduce this on OS X. What LLVM version is LDC built with?
(please paste the full output of `../bin/ldc2 --version`)

thewilsonator:
I also couldn't repro on OS X.

ximion:
I am currently at work, but I have the same Debian system here, so except for the CPU the version output should be the same:
```
LDC - the LLVM D compiler (1.1.0):
  based on DMD v2.071.2 and LLVM 3.8.1
  built with LDC - the LLVM D compiler (1.0.0)
  Default target: x86_64-pc-linux-gnu
  Host CPU: ivybridge
  http://dlang.org - http://wiki.dlang.org/LDC

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_be - AArch64 (big endian)
    amdgcn     - AMD GCN GPUs
    arm        - ARM
    arm64      - ARM64 (little endian)
    armeb      - ARM (big endian)
    bpf        - BPF (host endian)
    bpfeb      - BPF (big endian)
    bpfel      - BPF (little endian)
    cpp        - C++ backend
    hexagon    - Hexagon
    mips       - Mips
    mips64     - Mips64 [experimental]
    mips64el   - Mips64el [experimental]
    mipsel     - Mipsel
    msp430     - MSP430 [experimental]
[...]

PR #1963, opened 2017-01-09
Put debuginfo on GEP instead of on the alloca with an offset.

JohanEngelen:
Resolves issue #1933 .

It's the same workaround for an LLVM bug (I think it's a bug) as #1646 .

~~Working on reducing the original testcase further so it doesn't rely on Phobos; will add later.~~ Original testcase added.

### DIFF
--- a/gen/nested.cpp
+++ b/gen/nested.cpp
@@ -56,20 +56,12 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
     return makeVarDValue(astype, vd);
   }
 
-  LLValue *dwarfValue = nullptr;
-#if LDC_LLVM_VER >= 306
-  std::vector<int64_t> dwarfAddr;
-#else
-  std::vector<LLValue *> dwarfAddr;
-#endif
-
   // get the nested context
   LLValue *ctx = nullptr;
   auto currentCtx = gIR->funcGen().nestedVar;
   if (currentCtx) {
     Logger::println("Using own nested context of current function");
     ctx = currentCtx;
-    dwarfValue = currentCtx;
   } else if (irfunc->decl->isMember2()) {
     Logger::println(
         "Current function is member of nested class, loading vthis");
@@ -84,10 +76,6 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
     Logger::println("Regular nested function, loading context arg");
 
     ctx = DtoLoad(irfunc->nestArg);
-    dwarfValue = irfunc->nestArg;
-    if (global.params.symdebug) {
-      gIR->DBuilder.OpDeref(dwarfAddr);
-    }
   }
 
   assert(ctx);
@@ -120,10 +108,6 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
     IF_LOG Logger::println("Same depth");
   } else {
     // Load frame pointer and index that...
-    if (dwarfValue && global.params.symdebug) {
-      gIR->DBuilder.OpOffset(dwarfAddr, val, vardepth);
-      gIR->DBuilder.OpDeref(dwarfAddr);
-    }
     IF_LOG Logger::println("Lower depth");
     val = DtoGEPi(val, 0, vardepth);
     IF_LOG Logger::cout() << "Frame index: " << *val << '\n';
@@ -135,27 +119,32 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
   const auto idx = irLocal->nestedIndex;
   assert(idx != -1 && "Nested context not yet resolved for variable.");
 
-  if (dwarfValue && global.params.symdebug) {
-    gIR->DBuilder.OpOffset(dwarfAddr, val, idx);
-  }
+#if LDC_LLVM_VER >= 306
+    LLSmallVector<int64_t, 2> dwarfAddrOps;
+#else
+    LLSmallVector<LLValue *, 2> dwarfAddrOps;
+#endif
 
-  val = DtoGEPi(val, 0, idx, vd->toChars());
+  LLValue *gep = DtoGEPi(val, 0, idx, vd->toChars());
+  val = gep;
   IF_LOG {
     Logger::cout() << "Addr: " << *val << '\n';
     Logger::cout() << "of type: " << *val->getType() << '\n';
   }
   if (!isSpecialRefVar(vd) && (byref || vd->isRef() || vd->isOut())) {
     val = DtoAlignedLoad(val);
-    // dwarfOpDeref(dwarfAddr);
     IF_LOG {
       Logger::cout() << "Was byref, now: " << *irLocal->value << '\n';
       Logger::cout() << "of type: " << *irLocal->value->getType() << '\n';
     }
   }
 
-  if (dwarfValue && global.params.symdebug) {
-    gIR->DBuilder.EmitLocalVariable(dwarfValue, vd, nullptr, false, true,
-                                    dwarfAddr);
+  if (global.params.symdebug) {
+    // Because we are passing a GEP instead of an alloca to
+    // llvm.dbg.declare, we have to make the address dereference explicit.
+    gIR->DBuilder.OpDeref(dwarfAddrOps);
+    gIR->DBuilder.EmitLocalVariable(gep, vd, nullptr, false, true,
+                                    dwarfAddrOps);
   }
 
   return makeVarDValue(astype, vd, val);
--- a/tests/compilable/gh1933.d
+++ b/tests/compilable/gh1933.d
@@ -1,0 +1,66 @@
+// See Github issue 1933.
+
+// RUN: %ldc -c -release -g -O0 %s
+// RUN: %ldc -c -release -g -O3 %s
+
+ptrdiff_t countUntil(T, N)(T haystack, N needle)
+{
+    ptrdiff_t result;
+    foreach (elem; haystack) {
+        if (elem == needle)
+            return result;
+        result++;
+    }
+    return -1;
+}
+
+bool foo(alias pred, N)(N haystack)
+{
+    foreach (elem; haystack) {
+        if (pred(elem))
+            return true;
+    }
+    return false;
+}
+
+struct MatchTree
+{
+    struct Tag
+    {
+    }
+
+    struct Terminal
+    {
+        string[] varNames;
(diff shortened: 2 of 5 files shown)


JohanEngelen:
Any comments?
(it seems to work, and it fixes the bug)

rainers:
Does this also fix https://github.com/ldc-developers/dmd-testsuite/blob/ldc/runnable/gdb10311.d which is currently disabled https://github.com/ldc-developers/dmd-testsuite/blob/ldc/Makefile#L156?
This test case also showed wrong values for the closure variable while executing the nested function.

gen/nested.cpp
```diff
@@ -154,8 +151,10 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
   }
 
   if (dwarfValue && global.params.symdebug) {
-    gIR->DBuilder.EmitLocalVariable(dwarfValue, vd, nullptr, false, true,
-                                    dwarfAddr);
+    // Because we are passing a GEP instead of an alloca to
+    // llvm.dbg.declare, we have to make the address dereference explicit.
+    gIR->DBuilder.OpDeref(dwarfAddr);
+    gIR->DBuilder.EmitLocalVariable(gep, vd, nullptr, false, true, dwarfAddr);
```
kinke: If going with this, please do it properly and get rid of `dwarfValue` and make sure the only DWARF op is this last deref (otherwise it's surely wrong if the variable is an outer frame, as there's additional offsetting and dereferencing going on).

JohanEngelen: I don't understand. Are you saying that the original code is wrong? (this PR is just a different way of doing the same thing as the original code is doing)

kinke: As far as I understand this, we previously used the nested frame alloca as `LLValue* dwarfValue` for the DI declaration. We then represented the actual variable location via DWARF expressions based on that alloca, such as loading the first frame field for the outermost frame address and then further offsetting into that outer frame to get the address of a nested variable in the outermost scope. That's why the code does every offset and load twice, once for the actual `LLValue*` and then via DWARF ops for the debuginfo declaration.
Now with your workaround consisting in passing the actual variable address directly, a GEP, in combination with an explicit DWARF deref instead of an alloca with a potentially complex DWARF expression, we don't need the previous DWARF ops [here](https://github.com/ldc-developers/ldc/pull/1963/files#diff-8f5a14f6e1ae81e4570abcb00d35ebc5R124).

I'd rather favor looking into what we may do wrong wrt. DWARF expressions instead of using the GEP, as the docs state it should be an alloca, I guess because that should make it easier to keep debug infos for variables when optimizations (such as alloca elisions) are enabled (and in fact this transformation of debuginfos seems to cause #1933). I did a very quick test with `DW_OP_bit_piece` instead of `DW_OP_plus` for the offset, but that quickly produced an LLVM assertion, and I didn't pursue it further (and am currently lacking time).

JohanEngelen: Ahhh, yes you are right, good catch. Will work on a fix tonight. Hope I can test things a little.....

Back then, I also tried using `DW_OP_bit_piece` but failed after trying for a long time. I think these are all LLVM bugs. Clang uses a different mechanism (defines a struct for each nested frame) and so I think the optimization passes are not tested very well for our fancy DW_OP constructs.

I also do not have much time at the moment. I just hope we can release soon. So I'm happy to fix the bug right now, and postpone looking into how to use LLVM properly / file LLVM bugs later.

gen/nested.cpp
```diff
@@ -135,27 +119,33 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
...
+  LLValue *gep = DtoGEPi(val, 0, idx, vd->toChars());
+  val = gep;
   IF_LOG {
     Logger::cout() << "Addr: " << *val << '\n';
     Logger::cout() << "of type: " << *val->getType() << '\n';
   }
   if (!isSpecialRefVar(vd) && (byref || vd->isRef() || vd->isOut())) {
     val = DtoAlignedLoad(val);
-    // dwarfOpDeref(dwarfAddr);
+    gIR->DBuilder.OpDeref(dwarfAddrOps);
```
kinke: Thanks for the clean-up, LGTM. Are you sure about this particular change? Just asking because we give ref/out variables a special reference-debuginfo type [here](https://github.com/ldc-developers/ldc/blob/master/gen/dibuilder.cpp#L1092).

kinke:
@rainers: No, the value is still `<optimized out>` with https://github.com/ldc-developers/ldc/pull/1963/commits/4b68abbd44528d06666fa0da8143faa66ef31bc3 on Ubuntu 16.04.

rainers:
> No, the value is (still?) <optimized out>

Too bad. Thanks for testing.

gen/nested.cpp
JohanEngelen: No, I am not sure at all about that change. Thanks for pointing out the code where the reference type is set for the debug info.

JohanEngelen: Fixed.

kinke:
I've added 2 more temptative fixes: https://github.com/kinke/ldc/commits/fix1933

Those will need some more testing though.

kinke:
The 2nd one above is redundant, as special-ref variables have the ref storage class anyway, so they are already correctly declared as references, and `foreach (ref e; range)` works.

Sadly, there's a regression wrt. captured `this` though for gdb:
```D
struct S
{
    int a;

    void foo()
    {
        void bar()
        {
            a = 456; // <-- breakpoint here
        }
        bar();
    }
}

void main()
{
    auto s = new S(123);
    s.foo();
    assert(s.a == 456);
}
```

gdb shows `this` and `a` as `<optimized out>` with this PR (and my first tentative fix as well), while it worked fine before:
```
(gdb) info locals
this = @0x7ffff7ed6000: {a = 123}
(gdb) print this
$1 = (S &) @0x7ffff7ed6000: {a = 123}
(gdb) print this.a
$2 = 123
(gdb) print a
$3 = 123
```

JohanEngelen:
:( 

Can we live with the regression and get this in, possibly with your extra fix, to speed-up release?  (I don't have much time to delve into this issue :/ )

kinke:
I can't really make my mind up whether this fix for `-O<N > 0> -g` (easily worked around with `-disable-verify`) is worth it at the expense of that regression, which happens for regular `-g` builds and should thus have a far more wide-spread impact.

JohanEngelen:
I'll try to test a little with lldb.
Perhaps this PR just completely breaks the generated debuginfo? (and by accident removes the LLVM assert).

kinke:
I don't think it's that bad. I assumed the mostly enabled gdb tests in dmd-testsuite would cover regular closure variables, but since `vthis` apparently isn't, I'm not so sure anymore.

JohanEngelen:
My line of reasoning was: better to have bad debug info :( , then to have LDC crash on a popular project (`-g -O` is quite common I think). The workaround `-disable-verify` is not well-known enough. I thought we had pretty much completely dysfunctional debuginfo in the previous release, iirc (llvm >= 3.8)?

kinke:
If `-O -g` is indeed rather common, it definitely makes sense to include this (+ my 2 tiny extensions) in 1.1.

The cherry-picks onto `release-1.1.0` were rather conservative, so I don't expect any troubles from the few hotfixes. @JohanEngelen @klickverbot @smolt: Enough time to prepare the final 1.1 release this week? And any plans wrt. moving to a newer front-end (2.073.0?)?

JohanEngelen:
OK, so let's get this in. Will you do the merge @kinke ?

Yes, let's finalize the 1.1.0 release. I can easily prepare the Mac packages this week.

> any plans wrt. moving to a newer front-end (2.073.0?)

I have not worked on it because I am working on getting Weka up to LDC 1.1.0. Because I think it is very valuable to test LDC master (or something close to it) on Weka's codebase, I did not yet work on moving to newer frontend versions. But once I am done with Weka->1.1.0, I will probably start working on moving LDC to 2.072 soon after, and possibly 2.073 immediately after.

JohanEngelen:
(about merging: i meant including your 2 extra fixes)

kinke:
Merged as part of #1984.

smolt:
> Enough time to prepare the final 1.1 release this week?

I won't probably have time until weekend if curious about me creating the armv7hf release.  I just looked at the ldc-scripts and they appear to support arm, so would try that.  Do you all make the final binary by having ldc recompiling itself?

JohanEngelen:
> make the final binary by having ldc recompiling itself

Yes :-)

kinke:
https://github.com/ldc-developers/ldc-scripts/commit/ddc0c5ce3f42ce912ee834187963367a27a964ff should take care of bootstrapping LDC with itself. https://github.com/ldc-developers/ldc-scripts/commit/6c6f8f29776a56861ee69a33191436578f48ed69 reduces the variables to set manually. ;)

The release draft is up and I've added the few bugfixes since beta6, feel free to improve it. I'll do the tagging after dinner, so that we can build & upload the packages over the next days/weekend and then finally publish it officially.

JohanEngelen:
nice

kinke:
> I will probably start working on moving LDC to 2.072 soon after

I've just started merging 2.072.2 into druntime for a start just so you all know.

smolt:
I have kicked off an armv7hf build of 1.1.0 on a Raspberry Pi 3 using ldc-scripts.  First time trying these scripts, might take me a couple cycles to get it right.  And Pi 3, though way fast than Pi 1, still is slow compared to x86 machines.  BTW, I am thinking of also building ldc in a Raspbian compatible way (armv6 instructions), which supports all Pi versions (earlier Pi 1 is not armv7).  Is there interest in including that binary?

smolt:
Something I just realized.  By building on Raspbian, the system gcc only generates armv6 instructions by default.  This means C/C++ portions of LDC, druntime, and phobos will not use armv7 instructions.  It will run ok because armv7 is superset of armv6, and D parts will use armv7 instructions.  I could play with C flags when I build LLVM and LDC, but have a feeling it might take some time to get right.

kinke:
If it defaults to armv6, I think an armv6 LDC package makes most sense. Including the `-mtriple=...` in CMake variables `D_COMPILER_FLAGS` (LDC D parts) and `D_FLAGS` (runtime libs) may be enough to do the job.

kinke:
Btw I've started merging the 2.072.2 ddmd front-end and am in the process of ironing out the most obvious compilation errors. I'll open a corresponding branch in the next couple of days for collaboration.

PR #1965, opened 2017-01-13
added Homebrew osx/linux

timotheecour:
now that ldc is fixed on linux (https://github.com/Linuxbrew/homebrew-core/issues/852). Moved to top since since should be the preferred way to install since it's cross-platform, doesn't require sudo and adds more flexibility (more recent versions) w all of brew's advantages

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,14 +33,14 @@ installing it might not be necessary. For several platforms, there
 are also stand-alone binary builds available at the
 [GitHub release page](https://github.com/ldc-developers/ldc/releases).
 
-| Distribution | Command               |
-| ------------ | --------------------- |
-| Arch Linux   | `pacman -S ldc`       |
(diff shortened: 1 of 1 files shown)


timotheecour:
(homebrew isn't really a distribution, so it sticks out a bit in the list top/bottom irrelevant)
will you merge if I move it back alphabetically but keep the change `Homebrew => Homebrew osx/linux` ? it's not obvious otherwise that it's usable on all linux platforms (most ppl would assume homebrew=OSX). More discoverability = more ppl using = faster upgrades and bug fixes, and more ppl switching to more recent versions of ldc (since homebrew tends to have more recent version). That's the advantage.

alternatively i could keep as is but add a line:
```
linuxbrew `brew install ldc`
```

if not, I won't insist further.

PR #1966, opened 2017-01-13
Properly mark lit tests that require X86. Should fix these tests on ARM.

JohanEngelen:


JohanEngelen:
@smolt :)

smolt:
Yes - now just `codegen/align.d` test failure remains.  I haven't learned the FileCheck stuff so not sure what to do.  With arm and aarch64, structs are rewritten as vectors of i32 or i64.  The IR for the first align.d failure looks like this, but CHECK-SAME is looking for "align 32 %" on arg_arg I assume.  Ideas?

```
define void @_D5align23passAndReturnOuterByValFS5align5OuterZS5align5Outer(%align.Out
er* noalias sret align 32 %.sret_arg, [4 x i64] %arg_arg) #0 comdat {
  %arg = alloca %align.Outer, align 32            ; [#uses = 2, size/byte = 32]
  %1 = bitcast %align.Outer* %arg to [4 x i64]*   ; [#uses = 1]
  store [4 x i64] %arg_arg, [4 x i64]* %1
  %2 = bitcast %align.Outer* %.sret_arg to i8*    ; [#uses = 1]
  %3 = bitcast %align.Outer* %arg to i8*          ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i32(i8* %2, i8* %3, i32 32, i32 1, i1 false)
  ret void
}
```

PR #1968, opened 2017-01-14
Fix alignment test for ARM.

JohanEngelen:
@smolt: Continuation of #1966 .

PR #1969, opened 2017-01-15
Make tests/codegen/align.d more robust wrt. target ABI

kinke:
An alternative to #1968, but yet untested on ARM (I forgot to add ARM when building my current LLVM...).

### DIFF
--- a/tests/codegen/align.d
+++ b/tests/codegen/align.d
@@ -3,8 +3,8 @@
 // Fails on Windows_x86, see https://github.com/ldc-developers/ldc/issues/1356
 // XFAIL: Windows_x86
 
-align(32) struct Outer { int a; }
-struct Inner { align(32) int a; }
+align(32) struct Outer { int a; bool b; double c = 0; long d; }
+struct Inner { align(32) int a; bool b; double c = 0; long d; }
 
 align(1) ubyte globalByte1;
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Doesn't work on ARM.

For x86 we do
```
define void @_D5align23passAndReturnOuterByValFS5align5OuterZS5align5Outer(%align.Outer* noalias sret align 32 %.sret_arg, %align.Outer* byval align 32 %arg) #0 {
```
for ARM we do
```
define void @_D5align23passAndReturnOuterByValFS5align5OuterZS5align5Outer(%align.Outer* noalias sret align 32 %.sret_arg, [4 x i64] %arg_arg) #0 {
  %arg = alloca %align.Outer, align 32            ; [#uses = 2, size/byte = 32]
  %1 = bitcast %align.Outer* %arg to [4 x i64]*   ; [#uses = 1]
  store [4 x i64] %arg_arg, [4 x i64]* %1
```

This PR does not change ARM's IR into the x86 one, so the test still fails.

kinke:
I was pretty sure the mix of floating-point and integers as well as some padding inbetween fields would prevent the rewrite, but the ARM TargetABI is a bit strange and rewrites all structs <= 64 bytes. Closing.

PR #1973, opened 2017-01-18
WIP: Runtime compile

Hardcode84:
Here is current state of my 'jit' work. Simple examples work but everything mostly untested (and I'm not sure how properly test it). Tested only on win64 and only with llvm 3.9. IR preparation for jit mostly in ```gen/runtimecompile.cpp```. Jit itself implemented as shared library in ```runtime/rtcompile-rt/cpp-so/```, it was mostly copypasted from llvm examples and I'm not sure it 100% correct. More details later...

Known issues:
* Attempt to use thread local variables from jitted code:
```
Relocation type not implemented yet!
UNREACHABLE executed at c:\ldc\llvm\lib\executionengine\runtimedyld\Targets/Runt
imeDyldCOFFX86_64.h:104!
```
* Throwing exceptions from jitted code - crash


driver/cl_options.cpp
```diff
@@ -481,6 +481,12 @@ cl::opt<LTOKind> ltoMode(
                    "Parallel importing and codegen (faster than 'full')")));
 #endif
 
+cl::opt<bool> enableRuntimeCompile(
+    "enable-runtime-compile",
+    cl::desc("Enable runtime compilation"),
+    cl::init(false));
+
```
thewilsonator: You should use external storage for this (`cl::location`, see `useDeprecated` for an example).

driver/main.cpp
```diff
@@ -434,6 +434,14 @@ void parseCommandLine(int argc, char **argv, Strings &sourceFiles,
   }
 #endif
 
+  if (enableRuntimeCompile) {
+#if defined(LDC_RUNTIME_COMPILE)
+    global.params.enableRuntimeCompile = true;
```
thewilsonator: See above comment on external storage.

thewilsonator:
Prefer to use the `isaFoo` functions from `gen/tollvm.h` instead of `llvm::dyn_cast` where possible.

runtime/rtcompile-rt/d/ldc/runtimecompile.d
```diff
@@ -0,0 +1,74 @@
+module ldc.runtimecompile;
+
+immutable runtimeCompile = _runtimeCompile();
+private struct _runtimeCompile {}
+
```
thewilsonator: Move this into `ldc.attributes` and import it. This obviates the need for some of the changes in `gen/uda.cpp`. Also this file should be part of `druntime/ldc` and possibly consider conditionally enabling it via a version (see `driver/main.cpp:registerPredefinedVersions()`).

Hardcode84:
Thanks
Now I'am thinking how to workaround exceptions issues on win64: https://llvm.org/bugs/show_bug.cgi?id=24233

PR #1974, opened 2017-01-18
Do not generate DI for non-integral enum types.

LemonBoy:
Found while working on DMD's DWARF output, I have no idea if the test works but you get the point :)

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -695,7 +695,9 @@ ldc::DIType ldc::DIBuilder::CreateTypeDescription(Type *type, bool derefclass) {
                                       8, 8, "typeof(null)");
   if (t->ty == Tvector)
     return CreateVectorType(type);
-  if (t->isintegral() || t->isfloating()) {
+  if (t->isfloating())
+    return CreateBasicType(type);
+  if (t->isintegral()) {
     if (type->ty == Tenum)
       return CreateEnumType(type);
     return CreateBasicType(type);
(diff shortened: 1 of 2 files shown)


kinke:
Is that (integral enums only) a general DWARF limitation?

LemonBoy:
Not quite as the DWARF specification doesn't impose any constraint on the `DW_AT_const` but it seems gdb doesn't like non-integral enums.

kinke:
Visual Studio doesn't like float enums either (`<Unable to read memory>`).

JohanEngelen:
Instead of testing that the debuginfo is not an enum, perhaps you can test that the debugger shows something sane?

ljmf00:
> Not quite as the DWARF specification doesn't impose any constraint on the `DW_AT_const` but it seems gdb doesn't like non-integral enums.

They do. `DW_AT_const_value` is required to be a numeric value.

> Each enumerator entry has a DW_AT_name attribute, whose value is a null-terminated string containing the name of the enumeration literal. Each enumerator entry also has a DW_AT_const_value attribute, whose value is the actual numeric value of the enumerator as represented on the target system.

ljmf00:
Any updates on this @LemonBoy ?

kinke:
@ljmf00: Only integral enums are represented as DI enums these days, non-integral ones as typedefs, see https://github.com/ldc-developers/ldc/blob/3701684d6f3a615c911310a1fb5069493b6240d5/gen/dibuilder.cpp#L343-L348

ljmf00:
> @ljmf00: Only integral enums are represented as DI enums these days, non-integral ones as typedefs, see
> 
> https://github.com/ldc-developers/ldc/blob/3701684d6f3a615c911310a1fb5069493b6240d5/gen/dibuilder.cpp#L343-L348

Thanks. I just searched for DWARF related PRs to see what was in progress and didn't notice this was superseded.

PR #1979, opened 2017-01-21
ltsmaster - Fix lit tests for ARM

smolt:
Cherry picked lit-test fixes for ARM from master so ltsmaster can pass all lit-tests too.


JohanEngelen:
How do you guys feel about releasing (quietly) 0.17.3 ?

smolt:
I haven't been following closely, so can't think of a reason not to.

kinke:
> How do you guys feel about releasing (quietly) 0.17.3 ?

I'm not convinced it's necessary. I see `ltsmaster` as a moving target for bootstrapping on new platforms, getting some backports (mainly ABI-related) from time to time. The binary packages are of no real use, since the user is better off with newer releases (same set of platforms; true only since v1.1-beta6 as the PowerPC ABI fixes were forgotten in master). And just tagging `ltsmaster` with new 0.17.x releases (and not really testing them) from time to time seems arbitrary to me.

Additionally, as the releases on the GitHube page are ordered by release date, I don't like 0.17.x releases showing up at the top before the actual current release. A guy in hurry may just click on the releases link and download a package from the topmost release (I probably would ;)).

JohanEngelen:
It's not about the binary packages, it's about the source package to download. I thought that distributions that build from scratch only use released versions of software, so they can't use ltsmaster. For automated systems, a versioned source download is much preferable over a git hash checkout, I reckon.
Yes it is arbitrary to tag ltsmaster with a new release, but in our case that's how it is with every release. To me, it looks like there are some important commits in ltsmaster that warrant a new pointrelease. 

The order in which things appear on Github should not influence release decisions.
(doing the release now, just before 1.1.0, will take care of this too)

JohanEngelen:
Shall I create the 0.17.3 release then?

kinke:
Please do, thanks.

JohanEngelen:
On the order of releases on the GH release page, I found this: https://stackoverflow.com/questions/22822586/swap-latest-release-on-github
Do we want to go that route? :S

kinke:
The 0.17.3 text is pretty short and there are no binary packages, so people will hopefully scroll down a lil' bit, so I don't see a huge problem with this.

JohanEngelen:
There will be some binaries, for example to do our travisCI testing.  So I added the sentence to explain that it's not the latest version.

PR #1980, opened 2017-01-22
[ltsmaster] Sync LLVM-related CMake stuff from master

kinke:
Hopefully fixing the AppVeyor builds.


kinke:
The AppVeyor linking errors are fixed now, but stdlib compilation later on dies due to LLVM assertions wrt. CodeView debuginfos (not surprising at all).

PR #1981, opened 2017-01-22
[CMake] Change runtimelib dependency to configure_file input file.

JohanEngelen:
Fixes the "Policy CMP0058" cmake warning.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -314,7 +314,7 @@ macro(dc input_d d_flags output_dir output_suffix outlist_o outlist_bc)
                 ${LDC_EXE}
                 ${LDC_EXE_FULL}
                 ${GCCBUILTINS}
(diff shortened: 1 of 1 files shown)


kinke:
This means that the runtime libs aren't recompiled after changing `etc/ldc2.conf` anymore, right?

JohanEngelen:
Current master: when `ldc2{_phobos}.conf.in` is altered, runtimelibs are rebuilt. Also when `bin/ldc2.conf` is altered.

I've changed this PR to: when `ldc2{_phobos}.conf.in` is altered (note the `.in`), runtimelibs are rebuilt. When `bin/ldc2.conf` is altered, nothing happens.
Is that acceptable? It's nice to get rid of the cmake warning. I looked into proper ways of adding configure_file output file as a dependency, but I haven't found a solution yet and don't want to spend more time on it now. So the current solution is not acceptable, just close this PR.

PR #1986, opened 2017-01-25
[CMake] Add option to build LDC using LTO.

JohanEngelen:
[CMake] Add option to build LDC using LTO: `-DLDC_BUILD_WITH_LTO={OFF, ON, FULL, THIN}`.

When `LDC_BUILD_WITH_LTO` is set, the script adds -flto switch for the C++ compiler and D compiler, but only if they can create an executable with it.

Tested on OSX with ThinLTO (both with and without ThinLTO LLVM build). Some extra work is probably needed on Linux. On Windows, it won't do anything as the cmdline flag checks will fail.
On Linux, we should also enable the thinlto cache if possible.
Quick measurements indicate ~3% percent increased performance for `-O0` and `-O3` compilation (with ThinLTO LLVM too). Future step is adding PGO ;-)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ include(FindDCompiler)
 include(CheckIncludeFile)
 include(CheckLibraryExists)
 include(CheckCXXCompilerFlag)
+include(CheckDSourceCompiles)
 
 # The script currently only supports the DMD-style commandline interface
 if (NOT D_COMPILER_DMD_COMPAT)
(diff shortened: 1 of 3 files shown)


JohanEngelen:
What do you guys think?

kinke:
3% is hardly noticeable and not worth 100+ CMake lines IMO. :/

JohanEngelen:
lol ?
I'm not going to argue the truism of LTO making sense.

I'll split the 50 lines of extra cmakescript to a separate file (PGO will be in there too) and merge.

kinke:
> What do you guys think?
> I'll split the 50 lines of extra cmakescript [...] and merge.

Then why ask in the first place? :P

PR #1987, opened 2017-01-29
Merge 2.072.2 front-end

kinke:
The part needing most attention was ddmd.root.ctfloat, ddmd.target (incl. gen/target.cpp) and ddmd.builtin. The front-end is now prepared for elaborate compile-time floating-point types to allow for proper cross-compilation.

This version still uses the host's `real` type for compile-time reals, except for MSVC hosts, which still use 64-bit doubles (when compiled with DMD host compiler too).

Some other changes:

* semantic*() of Statements extracted from statement.d to statementsem.d
* mangle() -> mangleToBuffer()
* Identifier::string -> toChars()
* Token::float80value => floatvalue
* Dsymbol::isAggregateMember() => isMember()
* BoolExp is no more
* ddmd.root.ctfloat: LDC-specific CTFE builtins


kinke:
The compiler itself can be compiled by DMD 2.072.2 on my Win64 box now. There's quickly a crash when building the libs though.

JohanEngelen:
Nice work.

JohanEngelen:
Omg the warning spam is awful... How come DMD devs don't use C++ warnings......

I am going to add the constructor declarations again.

kinke:
I don't think readding the ctors is worth it. The problem is that the D ctors cannot be used from C++, so the declarations are useless and only increase the upstream diff. Better disable that warning for gcc (no warnings with MS compiler).

JohanEngelen:
Not having an explicit C++ ctor means there will be implicit ones........... I am not going to disable a super useful warning. The DDMD code is terrible, and it needs to be fixed.

kinke:
With current state and the following tentative workaround for a new sort of reference-initialization, druntime-ldc-debug can be compiled and linked:

```diff
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -470,17 +470,23 @@ public:
       Declaration *d = static_cast<VarExp *>(e->e1)->var;
       if (d->storage_class & (STCref | STCout)) {
         Logger::println("performing ref variable initialization");
-        // Note that the variable value is accessed directly (instead
-        // of via getLVal(), which would perform a load from the
-        // uninitialized location), and that rhs is stored as an l-value!
-        DSpecialRefValue *lhs = toElem(e->e1)->isSpecialRef();
-        assert(lhs);
+
+        DValue *lhs = toElem(e->e1);
         DValue *rhs = toElem(e->e2);

-        // We shouldn't really need makeLValue() here, but the 2.063
-        // frontend generates ref variables initialized from function
-        // calls.
-        DtoStore(makeLValue(e->loc, rhs), lhs->getRefStorage());
+        if (auto specialRef = lhs->isSpecialRef()) {
+          // We shouldn't really need makeLValue() here, but the 2.063
+          // frontend generates ref variables initialized from function
+          // calls.
+          LLValue *rhsLVal = makeLValue(e->loc, rhs);
+          DtoStore(rhsLVal, specialRef->getRefStorage());
+        } else {
+          LLValue *lhsLVal = DtoLVal(lhs);
+          if (!DtoIsInMemoryOnly(rhs->type))
+            DtoStore(DtoRVal(rhs), lhsLVal);
+          else
+            DtoMemCpy(lhsLVal, DtoLVal(rhs));
+        }

         result = lhs;
         return;
```

kinke:
Wrt. the remaining severe reference-init issue:

```D
bool[int] links() {
  bool[int] r;
  return r;
}

void foo(int tid, bool linked) {
  links[tid] = linked;
}
```

This currently crashes. `links[tid] = linked` is rewritten to `bool[int] __aatmp54 = links();) , ((ref bool __aaval55 = linked;)) , __aatmp54[tid] = __aaval55` (same as before). But the `__aaval55` reference to `linked` doesn't have the `STCforeach` storage class (besides `STCref`) anymore (although I haven't checked the storage class with LDC v1.1). As such, we don't recognize it as special-reference-variable (`DtoIsSpecialRef()`) anymore.

Not sure how to detect them properly now. @klickverbot: Any chance you know of some upstream changes in this regard? The new storage classes are STCref, STCinit, STCctfe and STCtemp.

kinke:
master's reference variable has storage class STCforeach instead of STCctfe.

kinke:
Alright, building compiler & libs should be working now. The first dmd-testsuite-debug results on Win64 are very promising, just a few failures. The debug unittests can almost be built as well; there are some `<at>safe`ty-related errors in `core.atomic` and `std.conv`.

kinke:
Update: OSX and the unittests should build successfully too now (Travis keeps running with a single job at a time...). Work on fixing the test failures can thus begin.

kinke:
There appears to be a C++ mangling issue on Linux x86_64 wrt. `alias real_t = real` (not mangled as `long double`) preventing a successful self-compilation (CircleCI).

kinke:
Oh wow, all unittests pass on CircleCI; 7 dmd-testsuite failures (release+debug each) remain. On Win64, unittests fail for `std.outbuffer` + `std.json` and there's 2 additional dmd-testsuite failures.

kinke:
I'll rework the LDMD command-line parser and -forwarder as side-product of fixing the remaining LDMD command-line issues (for Linux & OSX; some more work required to get Windows green).

kinke:
Yeah I know. The code diverged quite a bit (D vs. C++, clang-format, re-ordering upstream...), so I decided that translating in-place is worth it (and leads to more thorough testing of LDC's command-line parsing and command-line semantics as nice side-effect).
I think there's no real alternative to checking out manually where upstream's mars.d has changed when upgrading to a newer front-end in order to adapt LDC/LDMD accordingly (modified descriptions, new/removed command-line options...). Diffing the parsing code side-by-side may have been convenient in the past, but it's not anymore.

kinke:
We have the first green runs now. :)

kinke:
`runnable/test13613.d` most likely also fails on non-Windows, but only AppVeyor tests with assertions enabled. The issue is that an aggregate contains a field twice (two VarDeclarations with the same name and offset).
On Win32, the mangling of naked functions still doesn't work correctly; `extern(C++)` shouldn't get an extra leading underscore.
I haven't checked the 2 unittest failures on Win64 yet.

kinke:
The unittests now pass on Win64 too (=> https://github.com/dlang/phobos/pull/5097 + relaxed `std.json` unittest). So only that duplicate-fields LDC assertion remains.

kinke:
That duplicate fields issue was a pretty nasty one. The test (runnable/test13613.d) only triggers an LDC assertion when generating `S1.init`; it otherwise passes - although the LL struct contains one extra field (significantly diverging front-end `S1.sizeof` and actual LL struct size)!
LDC 1.1 is fine.

ddmd/aggregate.d
```diff
@@ -252,6 +252,17 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
...
+            {
+                for (size_t i = 0; i < ad.fields.dim; i++)
+                {
+                    if (ad.fields[i] == v)
+                    {
+                        // already a field
+                        return 0;
+                    }
+                }
+            }
```
kinke: @rainers: I don't know if you're a DMD compiler expert too, but if you are, it may be worth checking out if this bug (duplicate fields (2x `y`) for `struct S1` in `runnable/test13613.d`) haunts the DMD backend too.

rainers:
> I don't know if you're a DMD compiler expert too, but if you are, it may be worth checking out if this bug (duplicate fields (2x y) for struct S1 in runnable/test13613.d) haunts the DMD backend too.

Probably no longer an expert, but this issue is also in dmd. The problem is that determineFields is called recursively on S1, and after the second call has completed the fields, the first (outer) call still continues adding remaining fields.

`pragma(msg,typeof(S1.tupleof));` yields `(int, void*, void*, void*)`, so definitely worth upstreaming.

kinke:
AppVeyor's finally green too now.

ZombineDev:
It seems Iain has hit the same issue when upgrading GDC's frontend version and resurrected one of Kenji's PRs here: https://github.com/dlang/dmd/pull/6531.

kinke:
Thanks @ZombineDev for the establishing the cross-reference.

I'd say the job is as good as done; the duplicate fields issue should get an additional proper IR test, and Iain may fix the issue at its root upstream, but that's about it.

Wrt. the improved cross-compilability of D reals (see commit msg for details), I think this is the best we can have as long as we don't have a software `real_t`, for which we'd need libmpfr or some other library providing the required CTFE builtins (LLVM's APFloat doesn't).
So this supersedes #1317. In there, I used `real_t` as wrapper around a software `llvm::APFloat`, in target real precision. Due to missing pow(), exp() etc., I fell back to converting to host real (or double if target real was 64 bits) and letting the host C runtime perform the CTFE functions before converting back to target real precision. So firstly, CTFE used a mixed precision (target real precision for binops, possibly host real precision for the math functions). Secondly, wrapping a C++ type with complex lifetime (and default ctor...) in a D struct, with loads of binops and special functions, to be embedded in some front-end unions etc. didn't exactly make it easy and elegant, and never worked for target real precision > 64-bit (double-free).

kinke:
Wrt. LDMD, it's at least 400 lines less crappy code, and I actually don't find it that crappy anymore now (sure, strcmp() isn't very elegant...). The direct translation also prevents subtle issues like `ldmd2 -enable-contracts -release a.d` => `ldc2 -release -enable-contracts a.d` (non-supported args were appended after all switches directly recognized by LDMD; `-enable-contracts` after `-release` should re-activate them...). Wrt. LDC now supporting multiple occurrences for more command-line options, I think that scenario is not too uncommon, due to a chain of args (e.g., CMake default args + project args + user args), so I see that as another advantage. And wrt. future adaptations due to new command-line options, I'm absolutely sure this is an improvement after just having had to compare the logic side-by-side myself and then deciding that translating in-place was the way to tackle this.
I completely forgot about the header comment in `ldmd.cpp`; I tend to constantly overlook those as I usually hop from one function to the next...

I'd be happy to merge. Is there any point in sticking with 2.072 for LDC v1.2? Otherwise I'd simply move on to 2.073.1 to close the gap earlier.

kinke:
Löl, that was quick. ;)

kinke:
;) Any opinion on 2.072 vs. 2.073?

kinke:
An immediate 2.072 release cycle (just front-end + libs upgrade, no new features except for improved cross-compilation) while working on 2.073 and adding new features there sounds good to me.

JohanEngelen:
@klickverbot @kinke Very much in favor of a quick 2.072 release. I am testing LDC master on Weka's code now (merging things, then build, let's see)

(Before I forget: I do have a comments about our release mechanism: I think it is much better to branch "release-1.2.x" and do point releases from there only by tagging (instead of creating branches like 1.2.0, 1.2.1). Also, the ldc-packaging scripts should check out a release tag, instead of a branch.)

kinke:
+1 for LLVM-style `release-1.2` branch and tag-checkout. To be honest, I thought that's already how we handle it [apparently not really noticing the trailing `.0` in `release-1.1.0`] - until you recently created the `release-0.17.3` branch. ;)

JohanEngelen:
Lol, yeah, similar here. I actually noticed it when I bumped the version number in the release-1.1.0 branch and pushed it... Oops, quickly reverted that, you probably didn't see it ;-)
Good that we are on the same line :-)  (LLVM-style is what I meant indeed)

JohanEngelen:
> +1 for LLVM-style release-1.2 branch (and tag-checkout).

@kinke I modified ldc-scripts such that it does tag-checkout. Works on my Mac, can you test the MSVC changes? Thanks.

PR #1993, opened 2017-02-02
[dcompute] semantic analysis

thewilsonator:
Semantic analysis for dcompute and tests, now with correctly behaving submodules!


thewilsonator:
Hmm, making the `codeProduction` field of `ldc.attributes.compute` an enum has (unsurprisingly) changed it's type and therefore in uda.cpp
```C++
int hasComputeAttr(Dsymbol *sym) {

  auto sle = getMagicAttribute(sym, attr::compute);
  if (!sle)
    return 0;

  checkStructElems(sle, {Type::tint32}); // <----- Errors

  if (!sym->isModule())
    sym->error("@ldc.attributes.compute can only be applied to modules");

  return 1 + (*sle->elements)[0]->toInteger();
}
```

fails.

Suggestions for fixing it?
change it to verify that the type is an enum from `ldc.attributes` by the name `CompileFor`?
(no other enum exist in `ldc.attributes` and I don't know how to synthesise the required type given `checkStructElems` checks for equality by identity)

JohanEngelen:
This should work for enums (enums can be checked with Type::tint32):
```diff
void checkStructElems(StructLiteralExp *sle, ArrayParam<Type *> elemTypes) {
  if (sle->elements->dim != elemTypes.size()) {
    sle->error(
        "unexpected field count in 'ldc.attributes.%s'; does druntime not "
        "match compiler version?",
        sle->sd->ident->string);
    fatal();
  }

  for (size_t i = 0; i < sle->elements->dim; ++i) {
-    if ((*sle->elements)[i]->type != elemTypes[i]) {
+    if ((*sle->elements)[i]->type->toBasetype() != elemTypes[i]) {
      sle->error("invalid field type in 'ldc.attributes.%s'; does druntime not "
                 "match compiler version?",
                 sle->sd->ident->string);
      fatal();
    }
  }
}
```

thewilsonator:
What working directory is used when running the (semantic) test suite? So that I can correctly import from `./inputs/whatever` see the appveyor failure.

JohanEngelen:
Look at how things are done here with test imports: https://github.com/ldc-developers/ldc/blob/master/tests/codegen/inlining_disablecross.d
Add `-I%S` to LDC commandline.

thewilsonator:
Ahh thanks.

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,192 @@
...
+  }
+  void visit(CallExp *e) override {
+    // SynchronizedStatement is lowered to
+    //    Critsec __critsec105; // 105 == line number
+    //    _d_criticalenter(& __critsec105); <--
+    //    ...                                 |
+    //    _d_criticalexit( & __critsec105);   |
+    // So we intercept it with the CallExp ----
+
+    if (!strncmp(e->toChars(),"_d_criticalenter",16)) {
```
klickverbot: Style thing: I would use llvm::StringRef or even just std::string - str(n)cmp is so 1990s. ;)

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+
+  void visit(IfStatement *stmt) override {
+    if (stmt->condition->op == TOKcall) {
+      auto ce = (CallExp *)stmt->condition;
+      if (ce->f && ce->f->ident && ! strcmp(ce->f->ident->string,
+                                            "__dcompute_reflect")) {
+        auto arg1 = (DComputeTarget::ID)(*ce->arguments)[0]->toInteger();
+        if (arg1 == DComputeTarget::Host)
```
klickverbot: Braces please - the comment looks really funky otherwise.

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include "gen/recursivevisitor.h"
+#include "gen/uda.h"
+#include "gen/dcomputetarget.h"
+#include "gen/logger.h"
+#include "ddmd/declaration.h"
+#include "ddmd/module.h"
+#include "ddmd/identifier.h"
+#include "id.h"
+
+struct DComputeSemantic : public StoppableVisitor {
```
klickverbot: Seems like this doesn't do any analysis on its own apart from filtering disallowed constructs - maybe `DComputeRestrictionVerifier` (or something less clunky along those lines) would be more descriptive?

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,192 @@
+//===-- dcomputevailditychecker.cpp ---------------------------------------===//
```
klickverbot: File name.

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,192 @@
...
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+//
+// Validation for @compute code:
+//      enforce: @nogc, nothrow, all function calls are to modules that are also
+//          @compute. The enforcemnt of nothrow is simpler because all functions
+//          are assumed to not throw. We only need to check for ThrowStatement
```
klickverbot: So should `@compute` imply `nothrow`?

thewilsonator: It does, but not in the way that affects e.g. mangling. No functions can throw, and no functions can call other functions that are not also in a module that is `@compute`. I do it like this because I'm already doing a semantic analysis and so fold the requirement to check for `nothrow` into this pass. Yes, I probably could call the dmdfe nothrow semantic visitor, but I'm already visiting so whats the point? This will be faster and no less correct.

gen/semantic-dcompute.cpp
thewilsonator: Will do, I just want to see if the last commit fixed the problem

gen/semantic-dcompute.cpp
thewilsonator: I went with strncmp because otherwise I'd have to do some substringing and length checking (I think, not sure how substr behaves if you try yo substring beyond the length of the string). I think this is cleaner despite its "retroness". If you really want me to change it I will, but i think it's better like this.

gen/semantic-dcompute.cpp
JohanEngelen: I think the term semantic analysis also includes validity checking. (analyzes the code, and then tells you whether it was valid or not)
(also, in the future, someone may want to store dcompute sema results for codegen purposes)

I would use `DComputeSemanticAnalysis` or `...Analyzer` btw.

gen/semantic-dcompute.cpp
thewilsonator: In addition this disallows throwing `Error`s as well as opposed to just `Exception`s, which is what `nothrow` implies. 
CUDA transforms (the C/C++ macro) `assert` into a call to `__assertFail` (can't remember the exact spelling), I'm not sure about openCL. That is the only `Error` that will likely be used on device side code anyway.

tests/semantic/dcompute.d
```diff
@@ -0,0 +1,90 @@
...
+    switch(s)
+    {
+        default:
+            break;
+    }
+
+    //CHECK: Error: can only call functions from other @compute modules in @compute code
+    somefunc();
+    if (__dcompute_reflect(0,0))
+        //CHECK-NOT: Error: can only call functions from other @compute modules in @compute code
```
JohanEngelen: The CHECK-NOTs should be as general as possible, I think you can do `CHECK-NOT: Error` because FileCheck first matches the CHECKs before the CHECK-NOTs. (so it will check if there is no CHECK-NOT match between CHECK matches, which is what's needed here)

(You know how to run lit locally, right?  Go to the tests folder of your build dir and run `./runlit -v .` or `./runlit -v semantic/dcompute.d`)

gen/semantic.d
```diff
@@ -0,0 +1,28 @@
...
+import ddmd.arraytypes;
+import ddmd.dsymbol;
+import ddmd.dmodule;
+
+extern(C++) void dcomputeSemanticAnalysis(Module m);
+extern(C++) int hasComputeAttr(Dsymbol m);
+
+extern(C++) void extraLDCSpecificSemanticAnalysis(ref Modules modules)
+{
+    for (size_t i = 0; i < modules.dim; i++)
```
JohanEngelen: nit: ranged for?

tests/semantic/dcompute.d
```diff
@@ -0,0 +1,90 @@
+// RUN: not %ldc -o- -I%S %s 2>&1 | FileCheck %s
+
+
+@compute(CompileFor.deviceOnly) module dcompute;
+import ldc.attributes;
+import inputs.notatcompute : somefunc;
+
+extern(C) bool perhaps();
+extern(C) bool __dcompute_reflect(int,int);
+//CH ECK: Error: interfaces and classes not allowed in @compute code
```
JohanEngelen: there are a few disabled checks like this, is that intentional?

JohanEngelen:
For the error message checking: include the line number that the error should report. You can do that with e.g. `[[@LINE+2]]` to match the current line number plus two. See e.g. `codegen/attr_allocsize_diag.d`

JohanEngelen:
This is great btw. :)
 We can use the same extraLDCSpecificSemanticAnalysis method for a bunch of other error messages that we only report during codegen but that should really be reported for `-o-`.

tests/semantic/dcompute.d
thewilsonator: Yes perhaps these should be `XFAIL` instead. Perhaps this is frontend weirdness that I am not expecting. Speaking of erring on classes / ifaces, I should also disallow declaring (non global) class / iface variables.

The `-vv` output :
```
argeting 'x86_64-apple-macosx' (CPU 'core2' with features '')
Building type: real
Building type: uint
Building type: ulong
Building type: int
Type::getTypeInfo(): int
insnTemplate = ret
dcomputeSema: dcompute: dcompute.object
dcomputeSema: dcompute: dcompute.ldc
dcomputeSema: dcompute: __anonymous
dcomputeSema: dcompute: extern ()
dcomputeSema: dcompute: extern ()
dcomputeSema: dcompute: dcompute.I  <---- no error given
dcomputeSema: dcompute: dcompute.C <---- ditto
dcomputeSema: dcompute: dcompute.c
./dcompute.d(17): Error: variable dcompute.c global variables not allowed in @compute code
dcomputeSema: dcompute: dcompute.func
./dcompute.d(22): Error: variable dcompute.func.foo associative arrays not allowed in @compute code
./dcompute.d(24): Error: array literal in @compute code not allowed
./dcompute.d(26): Error: cannot use 'new' in @compute code
./dcompute.d(28): Error: cannot use 'delete' in @compute code
./dcompute.d(32): Error: setting 'length' in @compute code not allowed
./dcompute.d(34): Error: cannot use operator ~= in @compute code
./dcompute.d(36): Error: cannot use operator ~ in @compute code
./dcompute.d(38): Error: typeinfo not available in @compute code
./dcompute.d(40): Error: cannot use 'synchronized' in @compute code
./dcompute.d(42): Error: string literals not allowed in @compue code
./dcompute.d(44): Error: cannot switch on strings in @compute code
./dcompute.d(51): Error: can only call functions from other @compute modules in @compute code
./dcompute.d(57): Error: no exceptions in @compute code
./dcompute.d(67): Error: no exceptions in @compute code
./dcompute.d(83): Error: asm not allowed in @compute code
dcomputeSema: dcompute: dcompute.func1
dcomputeSema: dcompute: dcompute.func2
dcomputeSema: dcompute: lib <----------- this is the pragma(lib, "bar");
dcomputeSema: dcompute: object.RTInfo!(I)  <--- not sure what to do here
dcomputeSema: dcompute: object.RTInfo!(C) < prob ignore, it would be duplicate error
```

tests/semantic/dcompute.d
thewilsonator: So you're saying that `CHECK-NOT: Error:` won't eat the `CHECK`s that come after it?

tests/semantic/dcompute.d
JohanEngelen: The errors do not appear because some XXXDeclarations are not visited, because they are not routed through visit(Declaration*). For example, InterfaceDeclaration --> 
See visitor.h:
```d
    virtual void visit(AggregateDeclaration *s) { visit((ScopeDsymbol *)s); }
    virtual void visit(StructDeclaration *s) { visit((AggregateDeclaration *)s); }
    virtual void visit(UnionDeclaration *s) { visit((StructDeclaration *)s); }
    virtual void visit(ClassDeclaration *s) { visit((AggregateDeclaration *)s); }
    virtual void visit(InterfaceDeclaration *s) { visit((ClassDeclaration *)s); }
```

I think adding this to RecursiveWalker will fix it:
```d
  void visit(ScopeDsymbol *s) override {
    call_visitor(s) || recurse(s->members);
  }
```

JohanEngelen: `XFAIL` works different from what you think, btw (it's a Lit directive, not a FileCheck directive)
https://wiki.dlang.org/LDC_Lit-based_testsuite#Fail_tests

thewilsonator: Thanks! Also added support for `PragmaDeclaration` and `PragmaStatement`.

JohanEngelen:
pgo.cpp, MapRegionCounters needs a `void visit(Dsymbol *) override {}` . You're finding bugs in other parts of LDC code....

thewilsonator:
Wonderful, will "fix".

gen/pgo.cpp
```diff
@@ -309,6 +309,8 @@ struct MapRegionCounters : public StoppableVisitor {
     Hash.combine(PGOHash::OrOrExpr);
   }
 
+  void visit(Dsymbol *) override {}
```
JohanEngelen: move this to the other noop overrides around line 182.

thewilsonator:
@JohanEngelen Ping. Anything else?

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,197 @@
...
+struct DComputeSemanticAnalyser : public StoppableVisitor {
+  // InterfaceDeclaration
+  // ClassDeclaration
+  // VarDeclaration : isDataSeg (global variable)
+  // PragmaDeclaration : pragma(lib, "...")
+  // nogc
+  // nothrow
+  // no typeid
+  // no string switches
+  // no synchronized
```
JohanEngelen: Fix these comments to mean something.

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,197 @@
...
+  void visit(InterfaceDeclaration *decl) override {
+    decl->error("interfaces and classes not allowed in @compute code");
+    stop = true;
+  }
+  void visit(ClassDeclaration *decl) override {
+    decl->error("interfaces and classes not allowed in @compute code");
+    stop = true;
+  }
+  void visit(VarDeclaration *decl) override {
+    // Don't print multiple errors for 'synchronized'
```
JohanEngelen: clarify further, or refer to other comment about synchronized

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,197 @@
...
+      decl->error("global variables not allowed in @compute code");
+      stop = true;
+      return;
+    }
+
+    if (decl->type->ty == Taarray) {
+      decl->error("associative arrays not allowed in @compute code");
+      stop = true;
+    } else if (decl->type->ty == Tclass) // includes interfaces
+    {
```
JohanEngelen: clang format

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,197 @@
...
+  void visit(PragmaDeclaration *decl) override {
+    if (decl->ident == Id::lib) {
+      decl->error(
+          "linking additional libraries not supported in @compute code");
+      stop = true;
+    }
+  }
+
+  // Nogc enforcement.
+  // No need to check AssocArrayLiteral because AA's are benned anyway
```
JohanEngelen: typo (banned)

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,197 @@
...
+  }
+  void visit(CompoundAsmStatement *e) override {
+    e->error("asm not allowed in @compute code");
+    stop = true;
+  }
+  void visit(AsmStatement *e) override {
+    e->error("asm not allowed in @compute code");
+    stop = true;
+  }
+  // nothrow
```
JohanEngelen: improve comment.

whitespace

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+  void visit(IfStatement *stmt) override {
+    if (stmt->condition->op == TOKcall) {
+      auto ce = (CallExp *)stmt->condition;
+      if (ce->f && ce->f->ident &&
+          !strcmp(ce->f->ident->string, "__dcompute_reflect"))
+      {
+        auto arg1 = (DComputeTarget::ID)(*ce->arguments)[0]->toInteger();
+        if (arg1 == DComputeTarget::Host)
+          // Alllow code explicily for host to bypass the call only @conpute
```
JohanEngelen: typos

gen/semantic.d
```diff
@@ -0,0 +1,27 @@
...
+
+extern(C++) void extraLDCSpecificSemanticAnalysis(ref Modules modules)
+{
+    foreach(m; modules[])
+    {
+        if (hasComputeAttr(m))
+            dcomputeSemanticAnalysis(m);
+    }
+    
+}
```
JohanEngelen: EOL

tests/semantic/dcompute.d
```diff
@@ -0,0 +1,94 @@
...
+        func2();
+
+    //CHECK: dcompute.d([[@LINE+1]]): Error: asm not allowed in @compute code
+    asm {ret;}
+}
+
+void func1() {}
+void func2() {}
+
+//XFAIL: Error: linking additional libraries not supported in @compute code
```
JohanEngelen: XFAIL?

gen/semantic-dcompute.cpp
thewilsonator: Believe it or not that is clang formatted. Will change anyway.

thewilsonator:
@JohanEngelen Done, all green.

gen/semantic-dcompute.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+  void visit(IfStatement *stmt) override {
+    if (stmt->condition->op == TOKcall) {
+      auto ce = (CallExp *)stmt->condition;
+      if (ce->f && ce->f->ident &&
+          !strcmp(ce->f->ident->string, "__dcompute_reflect")) {
+        auto arg1 = (DComputeTarget::ID)(*ce->arguments)[0]->toInteger();
+        if (arg1 == DComputeTarget::Host)
+          // Allow code explicily for host to bypass the call only @compute
+          // restriction below.
```
JohanEngelen: I had to read this comment several times. How about
`Code inside an if(__dcompute_reflect)-statement is explicitly allowed to be non-@compute compatible code. Thus, the if-statement body's code should not be checked for @compute semantics and the recursive visitor should stop here.`
Also, move this comment to before the if(..."__dcompute_reflect")`

thewilsonator: done.

PR #1994, opened 2017-02-02
To check magic LDC UDAs parameter types, convert them to their basetype first.

JohanEngelen:
With this, you can check enums with Type::tint32, instead of constructing the enum type. For example: `checkStructElems(sle, {Type::tint32})` for this UDA:
```
enum E {a,b,c};
struct exampleUDA
{
   E e;
}
```

Needed for e.g. #1993


thewilsonator:
Once this passes and you merge it into master, please then merge master back into dcompute.

PR #1998, opened 2017-02-10
Hide more LLVM flags

klickverbot:
This cleans up the output for LLVM 3.9.1 on Linux with the default target set enabled.

`clang-format` makes the diff a bit hard to read, unfortunately.


kinke:
Too bad, rebasing isn't possible via online editor.

Issue #2001, opened 2017-02-16
inconsistent behavior of ldmd with -of, missing extension


BBasile:
`dmd somefile.d -lib -ofdir/somelib` outputs `somelib.a` in `dir`
`ldmd somefile.d -lib -ofdir/somelib` outputs `somelib` in `dir`

DMD adds a extension when no `.` is present in the filename part of the `-of` argument.

timotheecour:
* behavior of dmd/ldmd
* arguably dmd's behavior is too magical though

BBasile:
I don't thing it's a good idea at all to change DMD. LDMD should layer its behavior on DMD. DMD way of handling this case is good because extensions are system specific. In my example the same command line can be used on win or on linux, in one case to prduce a _.lib_, in the other a _.a_. This is something my IDE relies on BTW.

kinke:
Thanks for noticing, although I don't think any tool should rely on automatic file extensions.

kinke:
Fixed by #2002.

PR #2002, opened 2017-02-17
LDMD: Append default file extension if -of doesn't contain any

kinke:
Resolves issue #2001.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -325,7 +325,7 @@ cl::opt<std::string>
 #endif
 
 cl::opt<llvm::Reloc::Model> mRelocModel(
-    "relocation-model", cl::desc("Relocation model"),
+    "relocation-model", cl::desc("Relocation model"), cl::ZeroOrMore,
 #if LDC_LLVM_VER < 309
     cl::init(llvm::Reloc::Default),
 #endif
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -59,6 +59,9 @@ cl::list<std::string> runargs(
         "Runs the resulting program, passing the remaining arguments to it"),
     cl::Positional, cl::PositionalEatsArgs);
 
+cl::opt<bool> invokedByLDMD("ldmd", cl::desc("Invoked by LDMD?"),
+                            cl::ZeroOrMore, cl::ReallyHidden);
+
 static cl::opt<ubyte, true> useDeprecated(
     cl::desc("Allow deprecated code/language features:"), cl::ZeroOrMore,
     clEnumValues(clEnumValN(0, "de", "Do not allow deprecated features"),
--- a/runtime/profile-rt/DefineBuildProfileRT.cmake
+++ b/runtime/profile-rt/DefineBuildProfileRT.cmake
@@ -16,8 +16,6 @@ if (LDC_WITH_PGO)
         if (NOT (LDC_LLVM_VER GREATER 309))
             set(PROFRT_EXTRA_LDFLAGS "Ws2_32.lib")
         endif()
-    else()
-        set(PROFRT_EXTRA_FLAGS "-fPIC -O3")
     endif()
 
     CHECK_CXX_SOURCE_COMPILES("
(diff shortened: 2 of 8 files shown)


thewilsonator:
Would be good to have tests for this.

kinke:
I wanted to make sure all agree on this (not just for LDMD, but for LDC too) before adding the PITA tests.

thewilsonator:
Do we not have the ability to do tests for only LDMD? or are you wanting to do this for both and hence waiting?

BBasile:
the tests should verify this:

| command | Posix output | Windows Output |
| - | - | - |
| file.d -of=file | file| file.exe
| file.d -of=file.plugin | file.plugin| file.plugin
| file.d -lib -of=file.plugin | file.plugin| file.plugin
| file.d -lib -of=file | file.a| file.lib

kinke:
> Do we not have the ability to do tests for only LDMD?

Of course, e.g., dmd-testsuite uses LDMD exclusively.

> or are you wanting to do this for both and hence waiting?

This PR does it for both, as it's implemented in LDC, regardless of being invoked via LDMD or not. We could also keep the previous behavior of not tampering with the `-of` value (for LDC, not LDMD) in case some other tool relied on that, although I think that's highly unlikely as the static/dynamic libs or Windows executables are barely usable without extension.

BBasile:
The corner case is just when a dot is present in the `-of` argument. Unfortunately I don't speak cpp but I have the feeling that you always add the right extension, which must not be done. The right extension is only added when no extension is already present (row 2 & 3).

BBasile:
Also a question: You seem make the change in a file that's not specific to LDMD. Does the fix only applies to LDMD ? (it should not affect LDC2).

kinke:
Don't worry and don't rely on your feelings :], the extension is only added if the filename part doesn't contain a dot (see PR title...). DMD uses the same logic, and `FileName::defaultExt()` is implemented in the DMD front-end (https://github.com/ldc-developers/ldc/blob/master/ddmd/root/filename.d#L409).

kinke:
> Also a question

I already explained 2 posts further up...

JohanEngelen:
I think LDC should do what other compilers do: no automatic file extension adding.

kinke:
> I think LDC should do what other compilers do: no automatic file extension adding.

Which ones did you check? If gcc and clang don't add anything, I very much agree.

JohanEngelen:
I checked clang.

driver/cl_options.cpp
```diff
@@ -59,6 +59,9 @@ cl::list<std::string> runargs(
         "Runs the resulting program, passing the remaining arguments to it"),
     cl::Positional, cl::PositionalEatsArgs);
 
+cl::opt<bool> ldmd("ldmd", cl::desc("Invoked by LDMD?"), cl::ZeroOrMore,
```
JohanEngelen: could you rename the `ldmd` variable to `invokedByDMD` or something else more descriptive?

kinke:
https://github.com/ldc-developers/ldc/blob/master/driver/main.cpp#L1031 - shouldn't the default `-shared`` output extension be `.dylib` instead of `.so` on OSX (Darwin)?

kinke:
I'd say let's put this and (adapted) #2021 into 1.2 and then release the first alpha asap.

runtime/profile-rt/DefineBuildProfileRT.cmake
```diff
@@ -16,8 +16,6 @@ if (LDC_WITH_PGO)
         if (NOT (LDC_LLVM_VER GREATER 309))
             set(PROFRT_EXTRA_LDFLAGS "Ws2_32.lib")
         endif()
-    else()
-        set(PROFRT_EXTRA_FLAGS "-fPIC -O3")
```
kinke: Appropriate optimizations (for MSVC too) are already enforced - we rewrite all `CMAKE_C_FLAGS_{DEBUG,MINSIZEREL,RELWITHDEBINFO}` with ..._RELEASE in `runtime/CMakeLists.txt` (as the C parts of druntime/Phobos are always compiled in release too).

Issue #1997, opened 2017-02-10
Use travis_wait instead of spamming CI log with verbose output


klickverbot:
Verbose make/ctest output is currently only used to avoid false positives from the Travis build stall detection. However, this leads to huge build logs, and there is now an alternative:

https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

JohanEngelen:
Personally, I find the verbose output useful to check whether a test actually ran.

klickverbot:
The output for actually building the various object files is entirely unnecessary, though. (For the tests, one could argue that the counter is sufficient, but I see your point.)

klickverbot:
I find the CI build log to be far too verbose to get an overview of what's happening (> 8k lines!).

kinke:
+1 for hiding the objects progress only; I also prefer seeing which tests were run. I'll already be happy as soon as I don't need to click on "View raw log" anymore if the log is once again too long to be fully displayed. ;)

kinke:
With ninja replacing make for Travis, the objects progress has been reduced considerably, so I think this can be closed now.

klickverbot:
I suppose so (although I still think the test count would be enough ;)).

PR #2003, opened 2017-02-17
[WIP] Reduce Travis log verbosity

kinke:
Wrt. issue #1997.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,17 +80,17 @@ install:
 script:
   - cmake --version
   - cmake -DLLVM_CONFIG=$(which ${LLVM_CONFIG}) $OPTS .
-  - travis_wait make -j3 > /dev/null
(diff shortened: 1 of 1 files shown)


kinke:
As far as I can tell, we cannot use `travis_wait`, as it echos the output of both stdout and stderr (at the end), and redirecting the internal command manually isn't possible. Or rather, I didn't find any switch preventing `make` from displaying the progress. So we'd have to come up with a manual solution similar to https://github.com/bioconda/bioconda-recipes/commit/0ea6c5dc486c6ae4d01a9120199086351216425e. If the command is successful, it should echo stderr only (for warnings, decprecations etc.); if it fails, stdout would probably be useful as well.

kinke:
Closing in favor of #2066 (ninja).

PR #2005, opened 2017-02-18
Initial support for targeting RISC-V architectures: defining version identifiers

e-y-e:
This PR adds the RISCV32 and RISCV64 version identifiers to ddmd/cond.d as reserved identifiers, and allows selecting these architectures as the target.

LLVM 4.0 has defined the architectures as `llvm::Triple::riscv32` and `llvm::Triple::riscv64`, but does not yet have support for producing binaries for RISC-V. riscv/riscv-llvm has support for producing binaries, but defines 32 bit as `llvm::Triple::riscv` instead. Therefore, a cmake option was added for building with riscv-llvm (`-DRISCV_LLVM_DEV=ON`). The plan is to remove this option once LLVM has full RISC-V support.

The next step is determining the features of the target CPU.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -454,6 +454,16 @@ if(GENERATE_OFFTI)
     append("-DGENERATE_OFFTI" CMAKE_CXX_FLAGS)
 endif()
 
+option(RISCV_LLVM_DEV, "full RISC-V support with riscv-llvm")
+mark_as_advanced(RISCV_LLVM_DEV)
+
+#
+# Enable building with riscv-llvm, for full RISC-V support.
+#
+if(RISCV_LLVM_DEV)
+    append("-DRISCV_LLVM_DEV" CMAKE_CXX_FLAGS)
+endif()
+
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -88,6 +88,9 @@ set(LIB_SUFFIX "" CACHE STRING "Appended to the library installation directory.
 option(GENERATE_OFFTI "generate complete ClassInfo.offTi arrays")
 mark_as_advanced(GENERATE_OFFTI)
 
+option(RISCV_LLVM_DEV, "full RISC-V support with riscv-llvm")
+mark_as_advanced(RISCV_LLVM_DEV)
```
JohanEngelen: can you move all of this down to where the option is used?

e-y-e: Done.

JohanEngelen:
Thanks :)

For testing, you can use something like
```
// REQUIRES: target_RISCV

// RUN: %ldc -mtriple=...
```

TravisCI has LLVM 4.0 built with RISCV support.

PR #2006, opened 2017-02-18
Merge 2.073.2 front-end

kinke:


kinke:
The compiler itself can be built on Win64, but compiling the libs fails for each (?) module due to 'recursive template expansion'.

kinke:
Alright, compiler and libs build successfully now. The single remaining unittest compilation error is due to some sort of forward-referencing or other symbol resolution failure for std.regex. The first dmd-testsuite results are promising.
`mars.d` has changed quite a bit, primarily due to the new DMD `-mcpu` switch, with obviously incompatible values compared to LDC; a mapping to LLVM CPU and/or CPU features will probably be required.

driver/main.cpp
```diff
@@ -772,6 +775,15 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("D_PIC");
   }
 
+  if (arch == llvm::Triple::x86 || arch == llvm::Triple::x86_64) {
+    if (traitsTargetHasFeature("sse2")) {
+      VersionCondition::addPredefinedGlobalIdent("D_SIMD");
```
klickverbot: `D_SIMD` refers to the DMD-style `__simd` opcode interface.

kinke: Ah okay, and D_AVX isn't documented at all, but I guess it's just an extension for additional opcodes.

kinke:
This concludes propagating the `mars.d` changes to LDC/LDMD. On Win64, 5 dmd-testsuite-debug failures remain, 3 of which seem invalid.

kinke:
Update:

* unittests: `std.regex.internal.shiftor` fails to compile its unittests ('template not defined'), all others compile & work (at least on Win64).
* lit-tests: work.
* dmd-testsuite: work, except for:
  * `runnable/b16278.d`: missing symbol on Win64/OSX (debug only), links on Linux (?)
  * `runnable/test15862.d`: crash ('illegal instruction')

kinke:
`runnable/test15862.d` shows that LDC still lacks propagating immutable constants to LL constants. The 4 failing asserts check that repeated calls to pure functions returning indirections to immutable stuff return pointers to the same LL constant.

kinke:
So DMD caches the first result for all later detected identical invokations of strongly pure functions inside a function? Apparently only when optimizing. Shouldn't/couldn't this be done in the front-end already? At least optionally, in case the compile time/memory footprint impact could become noticeable.

kinke:
Apparently only remaining issue: `std.regex.internal.shiftor` unittest compile error.

kinke:
Wow, that `std.regex` thing seems really ugly.

`alias std.regex.internal.kickstart.Kickstart = std.regex.internal.kickstart.ShiftOr` not being available in module `std.regex.internal.shiftor` may be okay (?), as it's only imported locally in a struct in imported `std.regex.internal.ir`.

When trying to work around this via explicit import:
```diff
diff --git a/std/regex/internal/shiftor.d b/std/regex/internal/shiftor.d
index 0a549f4..f3aca3a 100644
--- a/std/regex/internal/shiftor.d
+++ b/std/regex/internal/shiftor.d
@@ -6,7 +6,7 @@ module std.regex.internal.shiftor;

 package(std.regex):

-import std.regex.internal.ir;
+import std.regex.internal.ir, std.regex.internal.kickstart : Kickstart;
 import std.range.primitives, std.utf;

 //utility for shiftOr, returns a minimum number of bytes to test in a Char
```

it then fails with

> C:\LDC\ldc\runtime\phobos\std\regex\internal\shiftor.d(29): Error: class std.regex.internal.shiftor.ShiftOr!char.ShiftOr base type must be class or interface, not Kickstart!char

which makes perfect sense, as the (very similar) base std.regex.internal.**kickstart**.ShiftOr!char is a struct, not a class!!

kinke:
Upstream PR: https://github.com/dlang/phobos/pull/5243

Awaiting some green now. :)

kinke:
Upgraded to 2.073.2 (very few fixes only). The first of the 2 front-end fixes originated from this PR (https://github.com/ldc-developers/ldc/commit/06e88fe2191363d45891eb025c365555c73b257c), the other one was already cherry-picked into master (https://github.com/ldc-developers/ldc/commit/b159de0819baf4450d5b1f2d0c29b5ef7e17df7f).

dushibaiyu:
will it marge 073.2 into ldc 1.2.0 release?

dushibaiyu:
@klickverbot  Thanks.

kinke:
Great that the OSX issue seems to be understood now (although it'll still happen for the shared libs as I understand).
We'll need it for LDC 1.2 too though, right?

kinke:
I see, since 2.072. I don't remember whether it was only the static-libs job spuriously failing since 2.072, but we'll find out.

kinke:
Looking good. I'll cherry-pick the fix onto master before merging 2.073.2 so that 1.2 doesn't (yet) need its own `ldc-release-1.2.x` druntime branch.

PR #2009, opened 2017-02-19
Linux: Always build C parts of standard library as PIC

### LINKED ISSUE #2008: druntime C parts of binary releases not built with PIC
On a fresh Ubuntu 16.10 image:

```
ubuntu@ubuntu:~$ echo 'void main() { import std.stdio; writeln("Hello World!"); }' > test.d
ubuntu@ubuntu:~$ ldc2-1.1.0-linux-x86_64/bin/ldc2 test.d
/usr/bin/ld: /home/ubuntu/ldc2-1.1.0-linux-x86_64/bin/../lib/libdruntime-ldc.a(errno.c.o): relocation R_X86_64_PC32 against symbol `__errno_location@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
```

In Ubuntu 16.10 (and newer), they patched the linker to make position-independent code the default even for executables (to enable ASLR, …). As a consequence, we chose to default to PIC on Linux as well. However, this applies only to the `ldc2` output and not the C parts of the runtime – since we build on an old Ubuntu version, `gcc` still defaults to non-PIC code. As a result, `libdruntime-ldc.a` is incompatible with the implicit PIE default of the system linker.

We need to push out a 1.1.1 release with this issue fixed.

klickverbot:
GitHub: Fixes #2008.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -351,6 +351,24 @@ macro(compile_phobos2 d_flags lib_suffix path_suffix outlist_o outlist_bc)
     endforeach()
 endmacro()
 
+# Sets the common properties for all library targets.
+function(set_common_library_properties target)
+    set_target_properties(${target} PROPERTIES
+        VERSION ${DMDFE_VERSION}
+        SOVERSION ${DMDFE_PATCH_VERSION}
+        LINKER_LANGUAGE C
+    )
+
+    # ldc2 defaults to position-independent code on Linux to match the implicit
+    # linker default on Ubuntu 16.10 and above. As we might be building on an
+    # older system (e.g. binary packages), we need to make sure the C parts are
+    # built as PIC as well.
+    if(${CMAKE_SYSTEM} MATCHES "Linux")
+        set_target_properties(${target} PROPERTIES
+            POSITION_INDEPENDENT_CODE ON)
+    endif()
+endfunction()
+
 # Builds a copy of druntime/Phobos from the source files gathered above. The
 # names of the added library targets are appended to outlist_targets.
 macro(build_runtime d_flags c_flags ld_flags lib_suffix path_suffix outlist_targets)
@@ -365,15 +383,13 @@ macro(build_runtime d_flags c_flags ld_flags lib_suffix path_suffix outlist_targ
     set_target_properties(
         druntime-ldc${target_suffix} PROPERTIES
         OUTPUT_NAME                 druntime-ldc${lib_suffix}
-        VERSION                     ${DMDFE_VERSION}
-        SOVERSION                   ${DMDFE_PATCH_VERSION}
-        LINKER_LANGUAGE             C
         ARCHIVE_OUTPUT_DIRECTORY    ${output_path}
         LIBRARY_OUTPUT_DIRECTORY    ${output_path}
         RUNTIME_OUTPUT_DIRECTORY    ${output_path}
         COMPILE_FLAGS               "${c_flags}"
         LINK_FLAGS                  "${ld_flags}"
     )
+    set_common_library_properties(druntime-ldc${target_suffix})
 
     # When building a shared library, we need to link in all the default
     # libraries otherwise implicitly added by LDC to make it loadable from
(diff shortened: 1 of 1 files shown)


JohanEngelen:
lgtm

kinke:
Defaulting to PIC on Linux should make LTO there even more useful, right? Johan, if you have some time, extending the ldc-scripts to use LTO when building the Linux/OSX release packages would be nice (let's actually use `LDC_BUILD_WITH_LTO`!).

JohanEngelen:
Will do, LDC_BUILD_WITH_LTO reduces the OSX binary size by _a lot_.

JohanEngelen:
@klickverbot I just made a 1.1.x branch.

JohanEngelen:
OK, agreed.

JohanEngelen:
Already working on it, shall I make a release 1.2.x branch already?

JohanEngelen:
(lol, chatting through github doesn't work so well :))

kinke:
Did you test this? It doesn't seem to work when trying to build a shared lib in #2002 (and linking in static druntime).

> /usr/bin/ld: /home/travis/build/ldc-developers/ldc/lib/libdruntime-ldc.a(errno.c.o): relocation R_X86_64_PC32 against undefined symbol `__errno_location@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC

kinke:
From CMake docs:

> The POSITION_INDEPENDENT_CODE property determines whether position independent executables or shared libraries will be created.

So it may have no effect for static library targets (or rather, the compiler flags for the C modules it depends on). How about building all libs as PIC, on each platform, explicitly via D and C compiler flags?

kinke:
Yep, this appears to be the same issue, triggered by linking against static druntime/Phobos (with non-PIC C parts) when building a shared lib.
https://github.com/ldc-developers/ldc/pull/2002/commits/288dba89695bca92a446c2f65795c4d6181e0d46 hopefully fixes this by compiling everything related to runtime libs as PIC.

kinke:
It clearly isn't, that target-property-trick apparently doesn't work for static libs. My variant now does work, but shows new problems for 32-bit Windows (Phobos-hello-world crashes).

kinke:
> You also disregard...

I don't disregard the fact that this fix fixed the non-PICness of the C parts in shared libs (if the host compiler didn't default to `-fPIC`, i.e., older Debian and Ubuntu).

But the C parts in static libs are still clearly non-PIC on those systems, as the snap package and my new dmd-testsuite tests (building a shared lib and linking against **static** druntime/Phobos for non-`BUILD_SHARED_LIBS=ON` - not in master yet...) show (Ubuntu 14.04 far snap IIRC, 12.04 for Travis). I can also give it a shot on Ubuntu 16.04 sometime to see if it's affected too (I guess it is).

kinke:
> That never was an issue.

Re-read the issue, I stand corrected. The issue is really getting strange now though.

> That isn't supposed to work, and wasn't ever.

Oh, so having a fully self-sufficient .so with no other dependencies (and it's private GC etc.) isn't supported and shared runtime libs are mandatory?

kinke:
Apparently 2.8.12. I just found

> CMake 2.8.9 and higher prefer instead to use the POSITION_INDEPENDENT_CODE target property to determine what targets should be position independent

[here](https://cmake.org/cmake/help/v3.1/policy/CMP0018.html).

kinke:
I don't know - their docs sadly only go back to 3.x, no 2.8.x docs anymore... I mentioned that I hate it, right? ;)

kinke:
I'll try to update Travis' CMake (https://github.com/ldc-developers/ldc/pull/2002/commits/5bd9cf44b9f2a4d8aa9e4bab3ed3f871e1b38258, new version is 3.2.3) incl. bumping the required CMake version (hoping that 2.8.9 is indeed sufficient). The target property is the way more elegant way and gets rid of the Win32 issue.

kinke:
CircleCI uses 2.8.12.2, and that works. PICness is tested when linking against static druntime/Phobos (in case of `BUILD_SHARED_LIBS != ON`) when generating a tiny shared object as part of the #2001 tests (just checking the automatically appended file extension for `ldmd2 -shared -of=sharedLib ...`).

kinke:
Travis works too now with CMake 3.2.3. Stock CMake for Travis OSX is 3.6.2 btw.

PR #2014, opened 2017-02-24
[dcompute] merge master into dcompute

thewilsonator:
Primarily because my ldc is segfaulting trying to do `opts::hideLLVMOptions()` and I want to be ahead of #1998 before I try do any fixing.


thewilsonator:
Hmmm, was thinking it would be good to update druntime first but https://github.com/ldc-developers/druntime/compare/dcompute...ldc-developers:ldc shows 361(!) unmerged commits, which is weird because it does show the commits I want i.e. everything from `Merge tag 'v2.072.2' into ldc-merge-2.072` (f1a06df 28 Jan) onwards.

As for the confilcts: 
`ddmd/mars.d`: is me adding the extra semantic pass. Action: ours.
`driver/cl_options.cpp`: is #1998 Action: theirs. I still need to add the SPIRV passes to the ignore array, but that comes later.
`gen/statements.cpp`: is my codegen conditional compilation hack. There is raw var declaration above it which seems to have been removed at some point, so remove that keep the hack.
`runtime/druntime`: is just because it is a submodule. My only addition to that is https://github.com/ldc-developers/druntime/pull/87 which is `lcd.attributes` and `ldc.dcomputetypes`.

Suggestions and help greatly appreciated.

kinke:
> which is weird because it does show the commits I want i.e. everything from Merge tag 'v2.072.2' into ldc-merge-2.072 (f1a06df 28 Jan) onwards.

Yeah well master is at 2.072.2 level now while ltsmaster is still at 2.071.2, so of course there's plenty of new stuff in the libs and LDC itself too. No idea what you mean by `Action: ours.` and `Action: theirs`. I resolved all actual conflicts in master already, so the few conflicts you see now are due to ltsmaster-specific divergences.

thewilsonator:
> Yeah well master is at 2.072.2 level now while dcompute is still at 2.071.2, so of course there's plenty of new stuff in the libs and LDC itself too.

Oh, some them are from like march last year, and I thought it was some sort of space-time discontinuum. I got confused and thought that your commit would be in place of all the others.

> No idea what you mean by Action: ours. and Action: theirs. I resolved all actual conflicts in master already, so the few conflicts you see now are due to dcompute-specific divergences.

OK will merge druntime and see if that makes if easier to resolve.

PR #2015, opened 2017-02-25
[dcompute] compilation and behaviour fixes

thewilsonator:
Update `ident->string` to `ident-toChars()`

Fix `__dcompute _reflect` code generation

Actually target CUDA

Disable debug info.

17421c4 is NFC, but I could not for the life of me get the insert on line 378 of asmsstmt.cpp to not crash with `set<string>`, however this works and should be of comparable performance.

Don't pass `Pointer!(n,T)` in memory.

### DIFF
--- a/gen/semantic-dcompute.cpp
+++ b/gen/semantic-dcompute.cpp
@@ -140,7 +140,7 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
     if (stmt->condition->op == TOKcall) {
       auto ce = (CallExp *)stmt->condition;
       if (ce->f && ce->f->ident &&
-          !strcmp(ce->f->ident->string, "__dcompute_reflect"))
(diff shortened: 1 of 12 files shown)


gen/semantic-dcompute.cpp
```diff
@@ -25,9 +25,46 @@
...
 #include "ddmd/identifier.h"
+#include "ddmd/template.h"
 #include "id.h"
 
+// In @compute code only calls to other function in @compute core are allowed.
+// However, a @kernel function taking a template alias function parameter is
+// allowed, but while the alias appears in the symbol table of the module of the
+// template declaration, it's module of origin is the module at the point of
+// instansiation so we need to check for that.
+bool isNonComputeCallExpVaild(CallExp *ce,FuncDeclaration *currentKernel) {
```
JohanEngelen: spelling...

PR #2016, opened 2017-02-25
configuration parser

rtbo:
Hi guys,

Following #1832 and #2007, yet another config file pull request. 
Hopefully the last one!


.travis.yml
```diff
@@ -71,7 +70,7 @@ before_install:
     export LLVM_CONFIG="llvm-$LLVM_VERSION/bin/llvm-config";
 install:
   - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then export CC="gcc-4.9"; export CXX="g++-4.9"; fi
-  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then brew update; brew install libconfig; fi;
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then brew update; fi;
```
klickverbot: You can delete the whole line, then.

kinke:
I like it, thanks. Some unittests wouldn't hurt; I guess we'll have to switch from reading directly from file to reading the full config string then.

rtbo:
> Some unittests wouldn't hurt

I Agree. I don't think any build target exists yet to run `unittest` blocks in the ldc executable.

JohanEngelen:
> run unittest blocks in the ldc executable.

I am _very strongly_ against adding unittests inside the source. It drowns code and promotes only shallow unittesting. Adding a general unittest framework to our repo would be nice, but for now perhaps you can just add some lit tests with `%ldc -conf=...` .

kinke:
I like the separate LDC unittest executable. As building it only recompiles the D modules, it doesn't even take a minute on our CI systems.

kinke:
Any remaining concerns?

JohanEngelen:
It'd be good to get this in 1.2-alpha too

kinke:
> It'd be good to get this in 1.2-alpha too

I don't think so. See https://github.com/ldc-developers/ldc/pull/1960#issuecomment-284041163.

kinke:
It doesn't. But it gets rid of the libconfig runtime dependency and as such distro maintainers should get rid of it too. In order not to annoy them with changes in 1.2 **and** 1.3 and to accelerate the 1.2 release (didn't we say one month ago that we're going to put out an alpha/beta immediately?), I'd prefer to delay it to 1.3, whose release cycle can theoretically (very theoretically given the very limited current manpower) be started these days anyway.

PR #2021, opened 2017-03-03
Cherrypick DMD PR #6582.

JohanEngelen:
Fix dlang Issue 17237 - Wrong alignment for 256-bit vectors.

https://forum.dlang.org/post/dehxjddmuyzoyqkbdzwn@forum.dlang.org
https://issues.dlang.org/show_bug.cgi?id=17237

### DIFF
--- a/ddmd/aggregate.d
+++ b/ddmd/aggregate.d
@@ -640,15 +640,7 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
         if (!isunion)
             *nextoffset = ofs;
 
-        if (alignment == STRUCTALIGN_DEFAULT)
-        {
-            if ((global.params.is64bit || global.params.isOSX) && memalignsize == 16)
-            {
-            }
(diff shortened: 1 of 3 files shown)


kinke:
Please wait before merging this, I asked Iain for more details upstream (https://github.com/dlang/dmd/pull/6582).

kinke:
Looks like I messed up something when merging the front-end - we have 2 `Target::alignsize()` implementations (`ddmd/target.d` and `gen/target.cpp`), and no linker complains. :(

JohanEngelen:
I thought to fully apply the diff, even if we don't use some code of it. But perhaps you prefer to not artificially add to the diff with the "based-upon dlang version"?

JohanEngelen:
Fixed.

JohanEngelen:
Ah lol. Confusion :)
I thought you wanted not the full diff. I have no preference, let me know.  I'll add the git commit hash too then.

PR #2025, opened 2017-03-07
Fix LDMD regression when generating response file

kinke:
It wrongly started with argv[0] (path to ldc2 binary).

Found while using dub, which easily generates command lines > 75K (Windows limit: 32K). Luckily before the first alpha/beta. ;)

### DIFF
--- a/driver/ldmd.cpp
+++ b/driver/ldmd.cpp
@@ -576,7 +576,7 @@ void translateArgs(size_t originalArgc, char **originalArgv,
       }
     } else {
       const auto ext = ls::path::extension(p);
-      if (ext.equals_lower("exe")) {
+      if (ext.equals_lower(".exe")) {
         // should be for Windows targets only
         ldcArgs.push_back(concat("-of=", p));
         continue;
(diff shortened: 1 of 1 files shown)


kinke:
Programmatically generating a huge LDMD command line (response file) via bash as part of a dmd-testsuite test sounds feasible.

kinke:
[Test](https://github.com/ldc-developers/dmd-testsuite/commit/6add0c447d6be7ec66b6bf3f1b210142a9b930bf) added.

driver/ldmd.cpp
```diff
@@ -576,7 +576,7 @@ void translateArgs(size_t originalArgc, char **originalArgv,
       }
     } else {
       const auto ext = ls::path::extension(p);
-      if (ext.equals_lower("exe")) {
+      if (ext.equals_lower(".exe")) {
```
klickverbot: I suppose this is untested as well?

kinke: Yep. I refuse to add a test for this though, as otherwise we'd have to test every code path checking for a hardcoded file extension (no dot in DMD front-end, with dot in LLVM...). ;)

kinke:
> Wouldn't it be better to add a source file at the end of the response file as well to make sure its contents are fully forwarded?

Then I'll need to check the beginning too. Could set 2 versions, one in the first line, the other one in the last, and check for them both with a static assert inside the source file... not motivated enough right now though. ;)

PR #2026, opened 2017-03-07
dmd-testsuite: Improve LDMD response file test

kinke:


kinke:
Yeah well, there was the .sh file permissions issue again, requiring yet another dmd-testsuite commit. The permissions have no effect on my Windows box, so the tests were green here. :/

PR #2027, opened 2017-03-08
support int vector multiplication

MartinNowak:
- turns out clang/llvm allow those directly and don't have intrinsics

See [_mm_mullo_epi32](http://clang.llvm.org/doxygen/smmintrin_8h_source.html#l00581) for example.
Could also add div and others, but those hardly map to dedicated SIMD instructions.

Mostly interested in supporting https://github.com/MartinNowak/druntime/blob/7ac278655505d79d8d94ef2ffbc5200691e54918/src/core/internal/arrayop.d#L165.

### DIFF
--- a/ddmd/expression.d
+++ b/ddmd/expression.d
@@ -8227,9 +8227,6 @@ extern (C++) class BinAssignExp : BinExp
 
         int isvector = type.toBasetype().ty == Tvector;
 
-        if (op == TOKmulass && isvector && !e2.type.isfloating() && (cast(TypeVector)type.toBasetype()).elementType().size(loc) != 2)
-            return incompatibleTypes(); // Only short[8] and ushort[8] work with multiply
-
         if (op == TOKdivass && isvector && !e1.type.isfloating())
             return incompatibleTypes();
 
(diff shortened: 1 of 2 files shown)


yebblies:
If it's going to be different to dmd then shouldn't it be decided in Target?  Or is this already ldc-modified code?

MartinNowak:
> If it's going to be different to dmd then shouldn't it be decided in Target? Or is this already ldc-modified code?

Seems like this isn't target specific, but compiler specific. If there is a policy how to deal with those, let me know or just update the PR.

yebblies:
Basically it all gets stuck in target.  Maybe we'll eventually split it into something more formal but currently...

See eg https://github.com/MartinNowak/ldc/blob/1bb274608bb13b871670d9ff94e06dbbbba70523/ddmd/target.d#L96 for a similar type of check (that I assume it a no-op in ldc).

JohanEngelen:
Closely related: https://github.com/ldc-developers/ldc/pull/1485

thewilsonator:
@MartinNowak add `// RUN: %ldc -c %s` as the first line, so the test is actually run.

kinke:
@thewilsonator et al: It's just much more efficient editing the PR directly instead of posting suggestions (which take only a few seconds less than editing), having the author read it and waiting for the adaptations. I'll do in this case.

kinke:
Wrt. the test, I'd prefer [Johan's test](https://github.com/ldc-developers/ldc/pull/1485/files#diff-26c32ff086697dc8b80f55751c42dbcc), guarded by a `REQUIRES: target_X86` (as it targets Haswell) and maybe extended by a binassign test, as it checks generated assembly too. Not sure with which PR to continue... ;)

MartinNowak:
Made a dmd PR https://github.com/dlang/dmd/pull/6723 @JohanEngelen.

JohanEngelen:
Nice!

PR #2029, opened 2017-03-11
CI for dcompute

thewilsonator:
replace OS X 3.9.0 (no shared libs) with OS X 5.0.0 with SPIRV support to enable testing for DCompute.


thewilsonator:
Hmm, it didn't fail horrifically, but it timed out. Any ideas how to fix the timing out? This is possibly due to the fact the the llvm used is not release, and therefore not as fast, but I'd like to use a debug version if possible for proper backtraces.

kinke:
It didn't just time out by a little bit. It only managed to compile a few unittest modules after having built LDC. And the Travis OSX VMs are already damn slow and sometimes already time out.
So I recommend either adding a new special OSX Travis job **only** performing the dcompute tests you're interested in (lit-tests?), without dmd-testsuite, compiling and running the stdlib unittests etc., or trying how far a Release LLVM gets you.

thewilsonator:
Thanks, good ideas. I used OS X because thats what I'm on and I've yet to figure out (not that I've put any effort into it) how to do cross-compilation releases for LLVM. But if that works we can hopefully use it to do LLVM5.0 general testing.

kinke:
Frequently updated prebuilt LLVM master is available at least for CircleCI. Johan switched from LLVM `master` to `release_40` in https://github.com/ldc-developers/ldc/commit/1621bed1f445c384167b440befabcc1079671088, so by reverting that you should be able to use LLVM 5.0 (in release config) for CircleCI. Is SPIR-V support already in LLVM master or is your build something special except for debug config?

thewilsonator:
Yes mine is special. The only thing in master (for a while) are the triple enums. there isn't even a stub like for RISC-V.

thewilsonator:
Urgh that merging didn't turn out how I planned. I think I should have done the merge as a separate PR and then rebased on top of that. Is there any way to do that?

JohanEngelen:
I think in general, rebasing a PR is easier than merging stuff into it. You could still rebase this one on the point where you want it to be merged.

thewilsonator:
Wont merging the undesired commits on to the target branch via another PR remove them from this one? 

Anyhow I'll fix that tomorrow. It looks like writing tests for OpenCL is going to be a while lot of fun ;( 

Any idea whats going on with the ddoc tests spamming the logs?

JohanEngelen:
It doesn't make sense to merge master into a PR. You should merge master into dcompute branch, and then rebase this PR on top.

thewilsonator:
PTX works on travis 3.9+ appyeyor and circle, just gotta figure out this segfault for SPIRV.

.travis.yml
```diff
@@ -29,6 +29,7 @@ matrix:
 
 cache:
   directories:
+    - llvm-5.0.0
```
thewilsonator: Hmm still isn't working. I take it then that this will only download the release once and won't take into account that the release has updates. Is there a way to still check if the release has been updated?

RealTimeDeployment: At this time it is stripped down and cleaned up. I want to rewrite it with correct global symbols when i get the time. No it will not make call again on its own as it is it corrects and is finished. For continous the appropriate time to make the call should be specified by distributor. Prefered with this very basic model in my opinion would be for embedded device after charge cycle completed. This is only a subset. Honestly as it is it is for correcting incorrect target value and  there is no need to keep rechecking. Perhaps a flag for Temp_i _max not equal to Target_i_max would invoke inspection. I would never want it to keep checking as a monitor on an embedded device. One reason is expansion could reach infinity without limits and two it might race with changing conditions. If device has a constant power supply perhaps it might function as a pmu, but i worry about heat and as heat rises expansion rises and continous growth would occur requiring thermal management to override. As weight increases by heat this sunset sees it as loss and increases current. I dont want to overheat hardware. Which is were tensorflow comes in as tensor is used for monitoring and management. Tensorflow is working on a complete pip. Using gcc and characterization.

thewilsonator:
The travis failure is a timeout on the 3.7.1 multilib.

PR #2030, opened 2017-03-11
Enable cross-static-lib generation for LLVM 3.9+

kinke:
By directly integrating LLVM's `llvm-lib.exe` driver to generate static libs for MSVC targets and a heavily stripped-down version of the `llvm-ar` tool for the other targets.

For the moment, just testing if it basically works on Linux and OSX too (works fine here on Win64; able to produce libs for x86 and x86_64 Linux...) and which older LLVM versions aren't supported. I removed the `llvm-ar` logic for modifying an existing lib (insertion order...), I guess it'll have to be put in again for the final implementation. :/


kinke:
Replacing objects in existing libs (and keeping old members) re-added.

kinke:
`ldc2` executable size delta for my RelWithDebInfo LDC with RelWithDebInfo LLVM 3.9.1 (X86 targets only) on Win64, LLVM linked statically: +86 KB (25.2M vs. 25.3M).

kinke:
The 3rd commit seems to work fine (has to be tested manually, AppVeyor uses a preset environment) and paves the way for integrating LLD. Enough for tonight though. ;)

kinke:
Interesting - this time the spurious OSX failure/segfault wasn't `core.thread`, but `std.net.curl`.

kinke:
Any concerns? To me, this is currently mostly a shortcut for guys having to cross-link externally. Handling a single static lib instead of a multitude of objects is just more convenient. And for MSVC targets, I noticed that the static lib file size for a Phobos-hello-world is reduced by a very rough 10% [IIRC] compared to VS 2015 `lib.exe` (for both 32/64-bit targets).

PR #2031, opened 2017-03-13
Travis LLVM 4.0 + CircleCI LLVM trunk

JohanEngelen:


JohanEngelen:
hm, did LLVM 4.0 become that much slower such that OS X jobs on Travis now always time out? :(

kinke:
The last 2 OSX jobs apparently only ran for 42-43 mins. Travis either reduced the timeout (previously ~50 mins) or the VMs took an eternity to initialize or something like that.

thewilsonator:
It could be that the time required to fetch the new LLVM is counted but not displayed.

kinke:
... and the rest is the VM startup (~2min 43s), checkable under 'worker information' in line 5. Linux takes less than a second, but oh well.
OSX was already on the verge before and sometimes timed out. We'd need a couple more minutes or ideally more usable OSX workers; they're by far the slowest of our CI slaves.

kinke:
The OSX workers are supposed to have 2 CPU cores now too (was a single one when I last checked), see [here](https://docs.travis-ci.com/user/ci-environment#Virtualization-environments), but I guess Travis only updated the description as I didn't notice any performance boost at all.

Wrt. the new linker warnings when using LLVM 4.0 (`[LLVM] object file (...) was built for newer OSX version (10.12) than being linked (10.11)`), those could be fixed by switching to a newer OSX image, see [here](https://docs.travis-ci.com/user/osx-ci-environment#OS-X-Version).

And btw, Ubuntu 14.04 would now be available (still as beta) for the Linux container environments too, see [here](https://docs.travis-ci.com/user/trusty-ci-environment).

JohanEngelen:
I've reverted the LLVM upgrade for OS X, so that we can get this in quicker.

JohanEngelen:
@kinke Nice that you looked up the recent Travis capabilities.

JohanEngelen:
CircleCI failure is fixed by #2039

PR #2032, opened 2017-03-14
Fix LLVM 5.0 compilation

JohanEngelen:


JohanEngelen:
@thewilsonator Have a look at the comment: in your dcompute branch, you can add code that correctly sets the address space on debuginfo for pointers.

thewilsonator:
Thanks for the heads up. I don't support debug info properly yet and the change must be fairly recent because my version works. I'm going to recompile LLVM anyway as the OS X build slave timed out on a debug release (hoping optimisation will fix it), so feel free to merge this and I'll merge master rebuild everything.

PR #2041, opened 2017-03-18
Add -mscrtlib cmdline option and support -static on Windows

kinke:
The default MS C runtime library doesn't depend on the `LINK_WITH_MSVCRT` CMake variable anymore.

The user can freely choose among the 4 variants libcmt[d] / msvcrt[d] via -mscrtlib or choose between static/dynamic release variant via -static.

~~As LINK_WITH_MSVCRT defaulted to OFF and -static defaults to false, produced Windows executables/DLLs will now be linked against the runtime DLLs by default (and thus shrink in size).~~ LDC keeps on defaulting to the static release C runtime, so `-static=false` or `-mscrtlib=msvcrt[d]` must be used explicitly in order to link against the runtime DLLs.

### DIFF
--- a/driver/ldmd.cpp
+++ b/driver/ldmd.cpp
@@ -210,11 +210,9 @@ Where:\n\
 "  -map             generate linker .map file\n"
 #endif
 "  -mcpu=<id>       generate instructions for architecture identified by 'id'\n\
-  -mcpu=?          list all architecture options\n"
-#if 0
-"  -mscrtlib=<name> MS C runtime library to reference from main/WinMain/DllMain\n"
-#endif
-"  -mv=<package.module>=<filespec>  use <filespec> as source file for <package.module>\n\
+  -mcpu=?          list all architecture options\n\
+  -mscrtlib=<name> MS C runtime library to reference from main/WinMain/DllMain\n\
+  -mv=<package.module>=<filespec>  use <filespec> as source file for <package.module>\n\
   -noboundscheck   no array bounds checking (deprecated, use -boundscheck=off)\n\
   -O               optimize\n\
   -o-              do not write object file\n\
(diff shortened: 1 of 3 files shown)


kinke:
Pinging @Rainers. DMD does a bit differently by apparently embedding the reference(s) in the object file containing the entry point; I simply use command-line options for the MS linker.

kinke:
My tests with a simple Phobos-hello-world worked fine, but apparently we need some more libs, as dmd-testsuite's `d_do_test` executable cannot be linked due to missing `execv*()` and `tzset()` symbols. Maybe Rainer knows what's missing. I tried including `[lib]ucrt[d]` as default lib too, but that didn't help.

kinke:
Found it, the POSIX names are in `oldnames.lib`.

kinke:
@rainers: When linking against the dynamic release variant (by default in an intermediate commit), a few tests segfaulted in EH code on my system: core.thread-debug, runnable\eh.d, runnable\test4.d, runnable\testdstress.d.

rainers:
> Pinging @Rainers. DMD does it a bit differently by apparently embedding the reference(s) in the object file containing the entry point; I simply use command-line options for the MS linker.

The new VC project integration of Visual D relies on embedded library references as the link step is D agnostic. If you link with C/C++ code that has embedded references, there is no issue, but linking only D files compiled with LDC will not work. Wouldn't it make sense to do what dmd does, i.e. embed those references in the object file that contains main, WinMain or DllMain? Is it feasible?

The usual Visual D projects also do not link through LDC, though they already work similar as proposed here (currently passes library name and /NOD:libcmt).

kinke:
> The new VC project integration of Visual D relies on embedded library references as the link step is D agnostic.

Ah, so that's the reason. Does this imply that DMD also adds references to druntime/Phobos? Even the path to them?

rainers:
> Does this imply that DMD also adds references to druntime/Phobos? Even the path to them?

Yes, a reference to phobos (it includes druntime) is also added, but without path. The library search path is usually given by the config file sc.ini.

kinke:
I think the lib reference embedded in the main object doesn't cover all use cases. I really like using LDC as linker driver on Windows, and having the ability to quickly switch to debug and/or shared C runtime by relinking without recompiling the main module seems to provide some value to me.
This still allows LDC to simultaneously embed a reference if the main module happens to be compiled in the same command line, to allow for seamless Visual D integration and (L)DMD compatibility.

PR #2044, opened 2017-03-19
Continue with #1735 (-link-sharedlib)

kinke:
~~Requires/includes #1960 and~~ offers a simple cmdline switch `-link-sharedlib` to choose between static and shared druntime/Phobos.

The only breaking change is that `-link-sharedlib` defaults to true when generating a shared library via `-shared` and NOT using `-static`, so the libs specified via `-defaultlib` (or legacy `-debuglib`) will get a `-shared` suffix in the linker command line. Adding an explicit `-link-sharedlib=false` to the config file should restore full legacy behavior.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,6 +96,9 @@ script:
     if [[ "${OPTS}" == *-DMULTILIB?ON* ]]; then
       ninja -j2 phobos2-ldc-unittest-debug phobos2-ldc-unittest phobos2-ldc-unittest-debug_32 phobos2-ldc-unittest_32;
       ninja -j3 druntime-ldc-unittest-debug druntime-ldc-unittest druntime-ldc-unittest-debug_32 druntime-ldc-unittest_32;
+    elif [[ "${OPTS}" == *-DBUILD_SHARED_LIBS?ON* ]]; then
+      ninja -j2 phobos2-ldc-unittest-debug-shared phobos2-ldc-unittest-shared;
+      ninja -j3 druntime-ldc-unittest-debug-shared druntime-ldc-unittest-shared;
     else
       ninja -j2 phobos2-ldc-unittest-debug phobos2-ldc-unittest;
       ninja -j3 druntime-ldc-unittest-debug druntime-ldc-unittest;
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -86,6 +86,10 @@ static cl::opt<bool, true>
     createSharedLib("shared", cl::desc("Create shared library (DLL)"),
                     cl::ZeroOrMore, cl::location(global.params.dll));
 
+cl::opt<bool> staticFlag("static", cl::ZeroOrMore,
+                         cl::desc("Create a statically linked binary, "
+                                  "including all system dependencies"));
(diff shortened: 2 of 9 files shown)


JohanEngelen:
How about `-link-sharedstdlib` or some other name that more clearly indicates that it only concerns druntime and Phobos?

kinke:
`-link-debuglib` isn't any conciser.

kinke:
Oh and of course there is (and will remain) another breaking change wrt. `-link-debuglib` without `-debuglib=` (highly unlikely to have none of the latter as we had it in the config file). Previously, this would result in no default libs being linked against, whereas now a `-debug` suffix is appended to the `-defaultlib=...` names, rendering `-debuglib` obsolete in almost all cases.

kinke:
Required lib filename scheme to work out of the box with a config file defining `-defaultlib=<stem>`: `<stem>[-link-debuglib ? "-debug" : ""][-link-sharedlib ? "-shared" : ""]`, i.e., the one used in our runtimes CMake script (`libdruntime-ldc.a, libdruntime-ldc-debug.a, libdruntime-ldc-shared.so, libdruntime-ldc-debug-shared.so`).

kinke:
This combination resolves #1282.

kinke:
2 Travis jobs now hit:
```
Linking C shared library ../lib/libphobos2-ldc-shared.so
std/xml.o: file not recognized: File truncated
```

Seems like it's trying to link it although the object is still being compiled. Wtf, make/CMake bug?

kinke:
Any concerns or do all agree this is a sane solution?

kinke:
@klickverbot: This will need your approval or another form of feedback. ;)

kinke:
The initial reason I went for the `-shared` suffix is that no suffix doesn't work on Windows when using a single libs directory (static lib `phobos2-ldc.lib`, import lib for `phobos2-ldc.dll` also `phobos2-ldc.lib`) and I wanted consistent default library names across platforms.

Wrt. cross-compilation and configs: we already support multiple sections in `ldc2.conf` based on the target triple, so e.g. copying the `default` section and adjusting the default switches (other libs directory, external cross-linker via `-gcc=...`, possibly different `-defaultlib` etc.) already works (used when cross-compiling from Win64 to Win32). Surely not the most elegant and non-redundant way, but very simple and powerful.

No potential additional flavours come to mind right now.

kinke:
We could also add an additional `-stdlib-shared-suffix` option defaulting to `-shared` but overridable by distros only shipping with shared libs (to an empty suffix in their config file to disable the effect of `-stdlib-shared` or whatever).

kinke:
Fine by me, but we should think carefully about how to name that switch. I guess there's enough confusion about the terms 'runtime library' (druntime only?), 'standard library' (Phobos only?) and 'default library' already (mea culpa too). Wrt. the existing `-defaultlib=druntime-ldc,phobos2-ldc`, I'm nowadays inclined towards naming the 2 switches `-defaultlib-{debug,shared}`, what do you think?

kinke:
Superseded once more, by #2443 this time.

Issue #2048, opened 2017-03-21
Emit `uwtable` IR attribute like Clang does


klickverbot:
See https://bugs.llvm.org/show_bug.cgi?id=32335#c7 – apparently the best we can do for now is to copy Clang's defaults (`clang::ToolChain::IsUnwindTablesDefault()`).

kinke:
Mea culpa, I recently removed that for Win64 as no errors popped up when omitting it (and our comment only suggested it's needed for Linux x86_64).

kinke:
Fixed by #2049.

PR #2049, opened 2017-03-21
Re-add uwtable function attribute on Win64

kinke:
Apparently needed for x86_64 in general and NetBSD (possibly other BSD flavours too) based on `clang::ToolChain::IsUnwindTablesDefault()`.

Resolves issue #2048.

### DIFF
--- a/gen/abi.h
+++ b/gen/abi.h
@@ -107,6 +107,14 @@ struct TargetABI {
     return name;
   }
 
+  /// Returns true if all functions require the LLVM uwtable attribute.
+  virtual bool needsUnwindTables() {
+    // Condensed logic of Clang implementations of
+    // `clang::ToolChain::IsUnwindTablesDefault()` based on early Clang 5.0.
+    return global.params.targetTriple->getArch() == llvm::Triple::x86_64 ||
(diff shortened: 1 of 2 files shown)


kinke:
I wouldn't have found out about NetBSD if I hadn't checked the actual implementations in the subclasses. There's one more special case for Windows but it doesn't concern us (MachO with MSVC toolchain?) and a comment about missing uwtable support for non-x86 Windows targets etc. Wrt. clang, I thought mentioning it in the commit description is enough; feel free to add a code comment directly if you prefer it more explicit.

JohanEngelen:
Would be good to add tests for this, so that it is not so easily removed.

kinke:
Comment added myself. ;) - IMO enough to prevent accidental removal of the attribute. Testing this would be overkill to me.

kinke:
Seriously, compiling a dummy module 10 times and checking whether uwtable is there for x86_64 on Linux, OSX, Windows, NetBSD and on x86 NetBSD but NOT there for all other platforms sounds incredibly dumb and superfluous to me. Especially since LLVM 4.0 on Win64 does care about it and fails if it's missing.

PR #2050, opened 2017-03-21
Fix LLVM 5.0 build.

JohanEngelen:


JohanEngelen:
Hmpf. Something is broken in LLVM repo regarding the cxxflags output of llvm-config :/

Issue #2042, opened 2017-03-19
get_git_head_revision cmake function searches beyond base directory of source tree


WebDrake:
The `get_git_head_revision` function defined in `cmake/Modules/GetGitRevisionDescription.cmake` will continue to search recursively through parent directories for a `.git` directory even if the base directory of the source tree does not contain one.

This means that if (say) a source tarball is unzipped in a subdirectory of a (separate) git tree (example: someone using git to version their home directory, or in my case <https://github.com/ldc-developers/ldc2.snap/issues/21#issuecomment-287575769>, as part of the build process of a snap package), the `git describe` call will pick up on git version history not related to the actual LDC code, and corrupt the version info accordingly.

If it's not possible to straightforwardly prevent recursion beyond the base directory of the source tree (or if this recursion is a desired feature), it would be helpful if there was a clearly documented way to disable getting & comparing to the git revision description (e.g. describe how to avoid the git override in the `Version information` section of `CMakeLists.txt`).

PR #2052, opened 2017-03-23
Add option to avoid recursing in get_git_head_revision cmake function

WebDrake:
~~**Add IGNORE_GIT_REVISION cmake option to ignore git revision description**~~

~~This option allows users to override the default behaviour of comparing `LDC_VERSION` to the output of `git describe` and preferring the latter in the case of any difference.~~

~~This should obviously be used with caution but allows the user to avoid issues that can arise when e.g. building from a source tarball that has been extracted in a subdirectory of an independent git tree (!).  This is unfortunately a real possibility: apart from folks who use git to version their home directory [1], it's also proven an issue when trying to use source tarballs to build the LDC snap package [2].~~

~~[1]https://stackoverflow.com/questions/21266948/use-git-to-manage-home-directory/42946460~~
~~[2]https://github.com/ldc-developers/ldc2.snap/issues/21#issuecomment-287575769~~

~~This provides a workaround for https://github.com/ldc-developers/ldc/issues/2042.~~

Since the `get_git_head_revision` function is only ever invoked from the base directory of the LDC source tree, it does not need to recurse in order to get the correct git revision for LDC.  This patch therefore adds a parameter that can be set to `TRUE` or `FALSE` depending on whether one wants to use recursion or not, and sets it to false in when calling `get_git_head_revision` in LDC's own `CMakeLists.txt`.

This should prevent issues that arise when building from source tarballs that have been extracted into a subdirectory of another git tree (!).  This is unfortunately a real possibility: apart from folks using git to
version their home directory [1], it has also proven an issue when using source tarballs to build the LDC snap package [2].

[1]https://stackoverflow.com/questions/21266948/use-git-to-manage-home-directory/42946460
[2]https://github.com/ldc-developers/ldc2.snap/issues/21#issuecomment-287575769

Fixes https://github.com/ldc-developers/ldc/issues/2042.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,7 +302,7 @@ if(NOT TAG MATCHES "NOTFOUND")
         set(LDC_VERSION "${TAG}")
     endif()
 else()
-    get_git_head_revision(REFSPEC HASH)
+    get_git_head_revision(REFSPEC HASH FALSE)
     if(NOT HASH STREQUAL "GITDIR-NOTFOUND")
         # Append git hash to LDC_VERSION
         string(SUBSTRING "${HASH}" 0 7 LDC_VERSION_HASH)
--- a/cmake/Modules/GetGitRevisionDescription.cmake
+++ b/cmake/Modules/GetGitRevisionDescription.cmake
@@ -39,9 +39,18 @@ set(__get_git_revision_description YES)
 # to find the path to this module rather than the path to a calling list file
 get_filename_component(_gitdescmoddir ${CMAKE_CURRENT_LIST_FILE} PATH)
 
-function(get_git_head_revision _refspecvar _hashvar)
+function(get_git_head_revision _refspecvar _hashvar _use_recursion)
 	set(GIT_PARENT_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
 	set(GIT_DIR "${GIT_PARENT_DIR}/.git")
+	# if we're not recursing and GIT_DIR does not exist,
+	# we are not in git
+	if(NOT _use_recursion AND NOT EXISTS "${GIT_DIR}")
+		set(${_refspecvar} "GITDIR-NOTFOUND" PARENT_SCOPE)
+		set(${_hashvar} "GITDIR-NOTFOUND" PARENT_SCOPE)
+		return()
+	endif()
+	# otherwise, either GIT_DIR exists, or we recurse
+	# until we find it or we reach the root dir
 	while(NOT EXISTS "${GIT_DIR}")	# .git dir not found, search parent directories
 		set(GIT_PREVIOUS_PARENT "${GIT_PARENT_DIR}")
 		get_filename_component(GIT_PARENT_DIR ${GIT_PARENT_DIR} PATH)


WebDrake:
Note, this is another patch that ideally I'd like to provide to the 0.17.x and 1.1.x release series as well as 1.2.x.

WebDrake:
OK, I'll tweak tomorrow.  What do you suggest regarding submitting this patch against other release series?

WebDrake:
Rebased and retargeted against master (following @kinke's suggestion in the other PR) and renamed the parameter:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e0191cd..de642ee 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,7 +297,7 @@ set(LDC_D_GENERATED
 #
 # Gather source files.
 #
-if(NOT IGNORE_GIT_REVISION)
+if(NOT LDC_IGNORE_GIT_REVISION)
     include(GetGitRevisionDescription)
     git_get_exact_tag(TAG)
     if(NOT TAG MATCHES "NOTFOUND")
```

JohanEngelen:
Isn't it better to just fix the `get_git_head_revision` CMake function? It explicitly does a search for `.git` in parent directories, although this is not necessary for us. Won't simply removing that functionality work?

WebDrake:
Hmmm, I did consider that.  One reason I didn't touch it was that AFAICS it's a copy of a 3rd-party cmake utility script, and patching the function directly might have created some issues if we ever wanted to integrate an updated version of that script.

JohanEngelen:
We could submit an upstream patch ;-)

WebDrake:
> We could submit an upstream patch ;-)

Tell you what, I'll try to take some time to create an alternative patch which either provides an option to disable recursion, or allows custom definition of the highest-level directory to recurse to.  That ought to be viable to send upstream.  (I would imagine the original script allows recursion all the way to `/` for a reason.)

JohanEngelen:
I am not sure, and don't want you to spend a lot of time on this needlessly. I think it's OK to just modify the script and be happy :)  
To make sure that future updates of the script (if ever) won't overwrite the modifications, you could change the functionname slightly (e.g. add `_ldc`) so things would clearly break.

WebDrake:
Let's see what I can do in the next day or two.  At the end of the day the change is 99% for my convenience, so I'm happy to do the legwork to get it right ;-)

JohanEngelen:
Cool :)

WebDrake:
I've updated the patch with what I hope is a much more reliable solution.  The `get_git_head_revision` function now accepts a 3rd parameter which can be `TRUE` or `FALSE` in order to indicate whether or not to recurse to higher directories.  `CMakeLists.txt` sets this parameter to `FALSE` when calling the function.

This should fix https://github.com/ldc-developers/ldc/issues/2042 while also being a much more potentially upstreamable change.

WebDrake:
Note, the effects of the new patch are:

* the git revision will always be taken into account when building from a git clone;

* the git revision will never be used when building from a source tarball.

RealTimeDeployment:
Did tensor flow set expected type to only tensor flow?  They made a lot of changes today that I am worried are not correct, specifically in the string. I was attempting to send messages but I think they may have been taken as pull requests. I think they may have broken bzl specifically the string argument and allowed inputs. Also think they used the "security" note that wasn't wanted. I think they pushed it all together into one tether than seeing it as different possible "templates" for different usages. The pushed tensors as always the only input argument instead of understanding that it was a separate instance. Only for training tensors.

WebDrake:
Actually, I think it may be worth tweaking to use a parameter the user can set instead of just hardcoding `FALSE`. Thoughts...?

rpavlik:
FYI, I think I fixed this upstream in https://github.com/rpavlik/cmake-modules/commit/43a50aab1f2d453900972ef67bf03b25f6acb800 by not going above CMAKE_SOURCE_DIR. See https://github.com/rpavlik/cmake-modules/issues/58

Issue #2046, opened 2017-03-20
ppc64/ppc64le builds fail with "Error: number '1e-300' is not representable"


kalev:
1.2.0 beta1 fails to build on Fedora ppc64 and ppc64le builders with:

```
[ 13%] Generating src/core/stdc/fenv.o
cd /builddir/build/BUILD/ldc-1.2.0-beta1-src && /builddir/build/BUILD/ldc-1.2.0-beta1-src/build/bin/ldc2 --output-o -c -I/builddir/build/BUILD/ldc-1.2.0-beta1-src/runtime/druntime/src -I/builddir/build/BUILD/ldc-1.2.0-beta1-src/runtime/druntime/src/gc /builddir/build/BUILD/ldc-1.2.0-beta1-src/runtime/druntime/src/core/stdc/fenv.d -of/builddir/build/BUILD/ldc-1.2.0-beta1-src/build/runtime/src/core/stdc/fenv.o -w -relocation-model=pic -O3 -release -d-version=Shared
/builddir/build/BUILD/ldc-1.2.0-beta1-src/runtime/druntime/src/core/stdc/fenv.d(399): Error: number '1e-300' is not representable
/builddir/build/BUILD/ldc-1.2.0-beta1-src/runtime/druntime/src/core/stdc/fenv.d(400): Error: number '1e-300' is not representable
make[2]: *** [runtime/CMakeFiles/druntime-ldc.dir/build.make:325: runtime/src/core/stdc/fenv.o] Error 1
```

Full build logs available at https://kojipkgs.fedoraproject.org//work/tasks/922/18460922/build.log (short lived link)

kinke:
This originates from [here](https://github.com/ldc-developers/ldc/blob/5c5831a32f56dbc2e3aba142d7b3d77bebf587d1/ddmd/lexer.d#L2089-L2135) and should be the case if the C runtime's `strtold("1e300", null)` and/or `strtod("1e300", null)` leave behind an `errno == ERANGE` (setting `*isOutOfRange` [here](https://github.com/ldc-developers/ldc/blob/master/ddmd/root/ctfloat.d#L121-L138) or returning true [here](https://github.com/ldc-developers/ldc/blob/master/ddmd/root/port.d#L91-L112)).
Both options seem odd as decimal exponents up to 307 shouldn't be a problem for double precision. Maybe the `strtold()` signature is wrong in assuming D `real` (double precision for LDC on PPC) is C `long double` (that strange doubledouble?).

[...]

PR #2054, opened 2017-03-25
Let LLVM parse floating-point literals instead of host C runtime

kinke:
Rendering the parsing consistent across all platforms and resolving stuff like #2046. Visual Studio before 2015 couldn't handle hex literals and had some rounding quirks too.


kinke:
First commit should go upstream: https://github.com/dlang/dmd/pull/6653

kinke:
1st commit merged upstream, so I included a reference to it in this PR's first commit.

Issue #2059, opened 2017-04-03
enhancement: LLVM 5.0 supports "contract" FMF


JohanEngelen:
LLVM 5.0 gained the `contract` FastMathFlag, that enables fusing ops without having to enable all kinds of other fastmath options. (see e.g. https://reviews.llvm.org/D31169)

First step is for us to extend the magic `@ldc.attributes.llvmFastMathFlags` attribute implementation to support `"contract"`.  (the implementation is trivial, and probably less work than typing this issue :( )

Perhaps we can add a convenience attribute similar to `@fastmath`, something like `@fusedmath`. I don't know whether `contract` it is useful enough without the other FMFs. Hope @9il could do some testing ;-)

JohanEngelen:
For reference: https://reviews.llvm.org/D31169

JohanEngelen:
@9il  So, with some very rudimentary testing, the `contract` flag gives you operator fusing, but not reassociation. Is that helpful for some of your algorithms?

9il:
Yes

kinke:
Implemented by https://github.com/ldc-developers/ldc/pull/2060.

PR #2060, opened 2017-04-03
Implement @ldc.attributes.llvmFastMathFlag("contract") magic UDA.

JohanEngelen:
Resolves issue #2059

### DIFF
--- a/gen/uda.cpp
+++ b/gen/uda.cpp
@@ -215,6 +215,10 @@ void applyAttrLLVMFastMathFlag(StructLiteralExp *sle, IrFunction *irFunc) {
     irFunc->FMF.clear();
   } else if (value == "fast") {
     irFunc->FMF.setUnsafeAlgebra();
+#if LDC_LLVM_VER >= 500
+  } else if (value == "contract") {
+    irFunc->FMF.setAllowContract(true);
+#endif
   } else if (value == "nnan") {
     irFunc->FMF.setNoNaNs();
   } else if (value == "ninf") {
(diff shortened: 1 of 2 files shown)


JohanEngelen:
Circleci fails probably because its LLVM is a few days outdated

JohanEngelen:
Failing test: https://github.com/ldc-developers/ldc/pull/2063

gen/uda.cpp
```diff
@@ -215,6 +215,10 @@ void applyAttrLLVMFastMathFlag(StructLiteralExp *sle, IrFunction *irFunc) {
     irFunc->FMF.clear();
   } else if (value == "fast") {
     irFunc->FMF.setUnsafeAlgebra();
+#if LDC_LLVM_VER >= 500
+  } else if (value == "contract") {
+    irFunc->FMF.setAllowContract(true);
```
JohanEngelen: What do you guys think would be good for pre-LLVM5? Give the standard warning `ignoring unrecognized flag parameter` below, or add a special `LLVM < 5.0 does not support "contract"` ?

klickverbot: I'd do the latter, possibly even something like "compiler needs to be built against LLVM 5.0+ to support 'contract'" or so to make abundantly clear what is going on.

JohanEngelen: thanks, done.

JohanEngelen:
@klickverbot Do you have any further comments?

kinke:
Unrelated, just the usual annoying LLVM trunk breakages.

PR #2065, opened 2017-04-08
Support Visual Studio 2017 and use it for CI

kinke:


kinke:
I didn't realize the lit-tests do test the MSVC auto-detection, as the spawned processes don't inherit the full parent environment. The auto-detection works with the 2017 Build Tools in the default install path (on my box) but apparently still fails with full-blown VS 2017 for AppVeyor.
DMD fails to invoke the 2017 linker (changed directory structure...), so use LDMD as host compiler for the x86 job too.

kinke:
The good news is that VS 2017 seems to work out of the box - once it is detected. ;)
Detecting VS 2017 with that annoying MS helper binary works for AppVeyor (when invoked in the initial environment), but apparently not in the special environment set up for the lit-tests (only inheriting the parent's PATH or something along these lines).

rainers:
> Detecting VS 2017 with that annoying MS helper binary

Not sure what vswhere does, but I use registry key SOFTWARE\Microsoft\VisualStudio\SxS\VS7, value "15.0" in the Visual D installer to detect the VS installation. "VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt" contains the version of the VC tool chain.

kinke:
Thanks Rainer, that seems to work and also gets the lit-tests to run; I'll throw away the MS binary. AppVeyor have also updated their VS 2017 image today (now with VS 15.1 (26403) and WinSDK v10.0.15063).

PR #2066, opened 2017-04-08
Travis: Replace make by ninja

kinke:
Ever since building both static and shared runtime libs (and static/shared Phobos sharing their object files), the CI systems spuriously failed when generating the libraries, e.g., `std/xml.o: file not recognized: File truncated`.

~~Try writing to a temporary object file and rename it as soon as it is fully written.~~ Switching from make to ninja apparently fixes this problem and also reduces the Travis log verbosity (as Travis supports line-updates, simulating an interactive shell experience, kinda neat).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ addons:
     - libcurl3:i386
     - cmake
     - cmake-data
+    - ninja-build
 before_install:
   -
     if [ "${TRAVIS_OS_NAME}" = "linux" ]; then
@@ -77,6 +78,7 @@ before_install:
     export LLVM_CONFIG="llvm-$LLVM_VERSION/bin/llvm-config";
 install:
   - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then export CC="gcc-4.9"; export CXX="g++-4.9"; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then brew update; brew install ninja; fi;
   - eval "${DC} --version"
   - pip install --user lit
   - python -c "import lit; lit.main();" --version | head -n 1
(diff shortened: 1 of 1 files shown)


kinke:
It was an experiment showing that there are indeed conflicting file accesses. The move to ninja seems to get rid of this issue.

kinke:
My best guess is that it's a CMake/make bug wrt. simultaneous linking of two libraries sharing some object files. It has occurred for some of the latest compiled Phobos object files only. With make, there's an output like `Scanning dependencies of library...` which I don't see with ninja; I suspected it to have something to do with it (make just scanning the filesystem if all dependant files are there?), that's why I tried the .tmp-object-then-rename thingy. That then had the rather unexpected effect of emphasizing the issue; IIRC, Circle and at least 3 Travis jobs failed for that single commit (when trying to write/rename? the object file, not when linking).

As to missing dependencies, I think I did nothing wrong by collecting all object targets [here](https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L477) and then using this `phobos2_o` set for both [static](https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L485) and [shared](https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L495) libs.

kinke:
Looking at the Makefiles produced by CMake on Linux, there are multiple identical rules for the same object - i.e., one in `runtime/CMakeFiles/phobos2-ldc.dir/build.make` and another in `runtime/CMakeFiles/phobos2-ldc-shared.dir/build.make`. Each object is a custom CMake target...

kinke:
So I just recommend requiring ninja for parallel builds starting with 1.3 (although it's actually only needed when generating both static and shared runtime libs). 1.3 will require mods to ldc-scripts etc. anyway (no libconfig, shared libs, LLVM for MSVC hosts compiled with /MT etc), so it seems like a good time to clean up all sorts of build-related stuff.

kinke:
The simplest approach for now is to restrict the (new) combined static+shared runtime builds to the ninja generator, see 2nd commit.

kinke:
Good to go? I hate those random failures... ;)

kinke:
CMake fix in #2075 to allow for `make` testing.

kinke:
Updated, now simply make -> ninja replacement with more compact Travis logs.

kinke:
The multilib jobs are mostly close to or already over the edge => #2074 disables the 64-bit dmd-testsuite (tested by all non-multilib jobs after all) and should thus improve the current situation.
Travis' timings vary wildly lately, e.g., yesterday a plain normal LLVM 3.8 job nearly timed out (took 48 mins or so and should normally take less than 30).

PR #2068, opened 2017-04-10
Fix compilation error with DMD 2.074

kinke:
TypeInfo.init is now @disabled. Retain backwards-compatibility for ltsmaster.

Note: already used by the Travis LLVM 3.5 job.


kinke:
Right, upstream already has a solution => cherry-picked.

PR #2071, opened 2017-04-13
Fix LLVM 5.0 compilation.

JohanEngelen:
This fixes compilation with LLVM trunk.

However, there were apparently some more changes related to attributes, and now LDC cannot compile certain files any more.
For example, we now get things like:
```llvm
; Function Attrs: norecurse nounwind readnone
define uwtable "less-precise-fpmad"="false" "no-frame-pointer-elim"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "target-cpu"="core2" "unsafe-fp-math"="false" i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #2 {
```
instead of 
```llvm
; Function Attrs: norecurse nounwind readnone uwtable
```

I haven't been able to figure out where to fix things yet.


thewilsonator:
w.r.t the alloca addrspace I think the only place that it will be non-zero normally is for standard stack allocation on NVPTX where the stack addrspace is 5.

Is it already honoured for inline IR? Because that is the only other place that a non zero address space would arise.

gen/attributes.cpp
```diff
@@ -67,8 +67,8 @@ AttrSet::AttrSet(const AttrSet &base, unsigned index, LLAttribute attribute)
 AttrSet
 AttrSet::extractFunctionAndReturnAttributes(const llvm::Function *function) {
   auto old = function->getAttributes();
-  LLAttributeSet existingAttrs[] = {old.getFnAttributes(),
-                                    old.getRetAttributes()};
+  llvm::AttributeSet existingAttrs[] = {old.getFnAttributes(),
+                                        old.getRetAttributes()};
   AttrSet r(LLAttributeSet::get(gIR->context(), existingAttrs));
```
thewilsonator: (for llvm 5.0) You want to use the `AttributeList::get` form from line 312 of Attributes.h
```C++
  static AttributeList get(LLVMContext &C, AttributeSet FnAttrs,
                           AttributeSet RetAttrs,
                           ArrayRef<AttributeSet> ArgAttrs);
```

thewilsonator:
```C++
AttrSet r(LLAttributeSet::get(gIR->context(),old.getFnAttributes(),
                                  old.getRetAttributes(), {}));
```
worked for me.

JohanEngelen: Thanks, that static function got added very recently, nice

JohanEngelen:
addrspace:  Clang sets the address space to what the datalayout says:
```cpp
  const DataLayout &DL = M->getDataLayout();
  unsigned AddrSpace = DL.getAllocaAddrSpace();

```

thewilsonator:
Ah thats easy then, I'll try to find out what the NVPTX data layout string should be.

thewilsonator:
It seems the AllocaInst change was for the AMDGPU backend (see https://reviews.llvm.org/D31042?id=) and doesn't concern NVPTX at all, and the stack address space for OpenCL is 0 so this shouldn't affect us at all.

thewilsonator:
Eh, what? 
```
E: Unable to locate package llvm-5.0
E: Couldn't find any package by regex 'llvm-5.0'
E: Unable to locate package llvm-5.0-dev
E: Couldn't find any package by regex 'llvm-5.0-dev'
E: Unable to locate package clang-5.0
E: Couldn't find any package by regex 'clang-5.0'

sudo apt-get install llvm-5.0 llvm-5.0-dev clang-5.0 returned exit code 100
```

JohanEngelen:
addrspace LLVM commit: https://reviews.llvm.org/rL299888

JohanEngelen:
Nasty little LLVM change: http://llvm.org/viewvc/llvm-project?view=revision&revision=300367

PR #2076, opened 2017-04-22
Merge 2.074 front-end + libs

kinke:


ddmd/target.d
```diff
@@ -179,6 +190,13 @@ struct Target
             realalignsize = 2;
             reverseCppOverloads = true;
             c_longsize = 4;
+            if (ptrsize == 4)
```
JohanEngelen: Let's version this out.

kinke: This code is only used by DMD, we have our own Target struct at the top of the file.

ddmd/aggregate.d
```diff
@@ -251,18 +259,6 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
...
-            {
-                for (size_t i = 0; i < ad.fields.dim; i++)
-                {
-                    if (ad.fields[i] == v)
-                    {
-                        // already a field
-                        return 0;
-                    }
-                }
-            }
```
kinke: This will need to be checked, i.e., whether a proper fix preventing the same field from being added multiple times to the fields list has made it into upstream 2.074.

kinke:
First upstream PR: https://github.com/dlang/druntime/pull/1815

kinke:
Linux: 4 distinct failing unittest modules left, the rest passes.

gen/statements.cpp
```diff
@@ -899,15 +899,20 @@ class ToIRVisitor : public Visitor {
       stringTableSlice = DtoConstSlice(arrLen, arrPtr);
     } else {
       for (auto cs : *cases) {
-        if (cs->exp->op == TOKvar) {
-          const auto vd =
-              static_cast<VarExp *>(cs->exp)->var->isVarDeclaration();
+        // skip over casts
+        auto ce = cs->exp;
+        while (ce->op == TOKcast)
+          ce = static_cast<CastExp *>(ce)->e1;
```
klickverbot: Is there any chance that this drops a narrowing conversion? (Sorry if this is totally off, haven't studied the code closely.)

kinke: Nope. It's only used to check for a nested `VarExp`; `cs->exp`, the original expression, is evaluated in all cases (variable/constant).

kinke:
2nd upstream PR: https://github.com/dlang/druntime/pull/1816

kinke:
Next one: https://github.com/dlang/phobos/pull/5345

kinke:
Another one: https://github.com/dlang/phobos/pull/5346

kinke:
Update:

* There are mangling issues when using LDC 0.17.2 as host compiler (Travis LLVM 3.6 job).
* OSX isn't tested anymore due to #2077, but the previous compile error should be fixed (non-nogc fiber migration check).
* Linux and Windows only have 2 failing unittest modules left:
  * A `core.thread` [unittest fails](https://github.com/ldc-developers/druntime/blob/ldc-merge-2.074/src/core/memory.d#L899) in release only (`-O2` and higher), apparently because `pureMalloc(size_t.max - 2)` doesn't return `null`. I quickly checked the IR for a (working) stand-alone test, and both non-optimized and optimized IR looked fine.
  * `std.math` fails a bunch of `ieeeFlags` [checks](https://github.com/ldc-developers/phobos/blob/ldc-merge-2.074/std/math.d#L4945-L4989). We already disabled a few of those (right above) due to potential instruction reordering, but these new tests are more elaborate (using delegates), so I don't think LLVM interferes here even for debug builds, but I haven't digged deeper.

kinke:
With https://github.com/ldc-developers/phobos/commit/2220a3a959090333005306134e20dd5e576fdb2a, all `std.math-debug` tests work on my Win64 box except for a [rounding test](https://github.com/ldc-developers/phobos/blob/ldc-merge-2.074/std/math.d#L5616-L5654).

kinke:
Alright, first green runs - all but the Travis 2 `-DBUILD_SHARED_LIBS=ON` jobs should pass now.

kinke:
The shared libs issue (on Linux at least) is due to linking errors wrt. this [particular nested struct](https://github.com/ldc-developers/phobos/blob/ldc-merge-2.074/std/conv.d#L5955):
```
/home/travis/build/ldc-developers/ldc/lib/libphobos2-ldc.so: undefined reference to `_D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5emptyMFNaNbNdNiNfZb'
```

kinke:
Upgraded to 2.074.1 and synced with master.

kinke:
As to the final issue: not even a Phobos hello-world program can be linked when using the shared runtime libs due to undefined members of [this nested struct](https://github.com/ldc-developers/phobos/blob/ldc-merge-2.074/std/conv.d#L5955-L6019):
```
echo "import std.stdio; void main() { writefln(\"Hello LDC2 %d\", (void*).sizeof); }" > hello.d
bin/ldc2 hello.d -link-debuglib
/home/martin/build-ldc/lib/libphobos2-ldc-debug.so: undefined reference to `_D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5emptyMFNaNbNdNiNfZb'
/home/martin/build-ldc/lib/libphobos2-ldc-debug.so: undefined reference to `_D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5frontMFNaNbNdNiNfZa'
/home/martin/build-ldc/lib/libphobos2-ldc-debug.so: undefined reference to `_D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result6lengthMFNaNbNdNiNfZm'
/home/martin/build-ldc/lib/libphobos2-ldc-debug.so: undefined reference to `_D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result8popFrontMFNaNbNiNfZv'
/home/martin/build-ldc/lib/libphobos2-ldc-debug.so: undefined reference to `_D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZS3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result'
```

Any ideas why only the shared libs are affected?

kinke:
Static:
```
$ readelf -Ws build-ldc/lib/libphobos2-ldc.a | grep _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result
   886: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5emptyMFNaNbNdNiNfZb
   887: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5frontMFNaNbNdNiNfZa
   888: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result6lengthMFNaNbNdNiNfZm
   889: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result8popFrontMFNaNbNiNfZv
```
vs. shared:
```
$ readelf -W --dyn-syms build-ldc/lib/libphobos2-ldc-shared.so | grep _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result
    57: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5emptyMFNaNbNdNiNfZb
   233: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5frontMFNaNbNdNiNfZa
   255: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result6lengthMFNaNbNdNiNfZm
   298: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result8popFrontMFNaNbNiNfZv
$ readelf -Ws build-ldc/lib/libphobos2-ldc-shared.so | grep _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result
    57: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5emptyMFNaNbNdNiNfZb
   233: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5frontMFNaNbNdNiNfZa
   255: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result6lengthMFNaNbNdNiNfZm
   298: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result8popFrontMFNaNbNiNfZv
  3318: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5emptyMFNaNbNdNiNfZb
  9136: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result5frontMFNaNbNdNiNfZa
  9696: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result6lengthMFNaNbNdNiNfZm
 11076: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _D3std4conv47__T7toCharsVii10TaVE3std5ascii10LetterCasei1TkZ7toCharsFNaNbNiNfkZ6Result8popFrontMFNaNbNiNfZv
```

kinke:
Oh and btw, both shared and static Phobos libs consist of the same D object files.
Any suggestions? I'm stuck.

JohanEngelen:
So the function is not codegenned, I've run into similar issues with Weka's codebase where some templates are culled too aggressively.  But also because of failing to deduce attributes in certain cases but not in others: https://issues.dlang.org/show_bug.cgi?id=17541. 
Can you check if there isn't another symbol defined for that function but without the attributes?

kinke:
> So the function is not codegenned

But then why is the linker not complaining when feeding it the static lib with exactly the same D objects?

JohanEngelen:
Conjecture:  in the static case, the linker sees all symbols and knows it can strip the symbol in which the undef thing is referenced. In the sharedlib case, perhaps the linker cannot make as many assumptions, because the shared lib could be replaced with another and then behavior could be different when certain symbols are available or not? Really, I don't know.

kinke:
Thanks anway - I tested the `needsCodegen()` hack from #2168, commenting out [this](https://github.com/ldc-developers/ldc/blob/merge-2.074/ddmd/dtemplate.d#L7227-L7235), and hello-world can then be linked successfully.

JohanEngelen:
This is the first time I see that happening in Phobos stuff. That's very good news: see if it happens with dmd too, dustmite and submit bugreport. The weka stuff has been a bit elusive so far. (or I've lost track in the confusion of the weird complex bugs I submitted...)

kinke:
DMD 2.074.0 has no problems with linking the above hello-world with its combined (druntime + Phobos) `libphobos2.so` on Linux x64. That would have been shocking. ;)

JohanEngelen:
lol heh true. (thought they are doing static linking)  Very strange that our front-end culls, and theirs doesn't though....

kinke:
> Very strange that our front-end culls, and theirs doesn't though

I have a strong assumption - they seem to build druntime & Phobos in a single command line, all D files at once: https://github.com/dlang/phobos/blob/master/posix.mak#L321
Whereas we compile each object separately. And that culling basically checks whether a template is (also) instantiated in some imported non-root module. So if the complete imported source is also specified on the command line, this problem can't occur.

kinke:
So this should basically be due to (supposedly fixed) https://issues.dlang.org/show_bug.cgi?id=2500.

kinke:
> thought they are doing static linking

They do by default but ship with static and shared libs, so I caught the linker line via `-v` and linked against the shared lib manually.

kinke:
All green now after the template culling hack.

CMakeLists.txt
```diff
@@ -817,6 +819,9 @@ add_subdirectory(tests)
 
 install(PROGRAMS ${LDC_EXE_FULL} DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
 install(PROGRAMS ${LDMD_EXE_FULL} DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
+if(NOT CMAKE_INSTALL_PREFIX STREQUAL "/usr")
```
klickverbot: Huh? This smells massively.

kinke: I didn't want to install the bash script system-wide. It's more of an internal script by itself, useful for local builds and release packages, but not system-wide availability IMO.
Correct context: https://github.com/ldc-developers/ldc/pull/2198

kinke:
I created a test branch `merge-2.074-2` with the template culling hack reverted and #2231 merged. As expected, this also fixes the linking issues with shared Phobos. The Phobos unittest runners can also be linked successfully although the runtime modules are compiled separately. See https://travis-ci.org/ldc-developers/ldc/builds/257913800. And both Semaphore jobs successfully compiled Phobos; that may have been a fluke or the #2231 Semaphore issue has been fixed in 2.074.

I'd prefer using #2231 (maybe merged as part of 2.074 if the Semaphore issue reproducibly vanishes) and expose the weakened template culling hack as command-line option. Thoughts?

kinke:
> if the Semaphore issue reproducibly vanishes

Well, no Phobos compilation failures after 10 jobs, so I think it's safe to assume 2.074 gets rid of that last issue of #2231.
[The random failures when linking the testrunners are due to parallel `ninja` invokations when running the `build-*-test-runner*` CTest tests in parallel. Unfortunately, Semaphore doesn't support an in-source config file, so I can't switch the global config to `ninja all all-test-runners` as long as #2231 isn't merged.]

PR #2082, opened 2017-04-28
[wip] try using non-LTO-build of LDC on Mac.

JohanEngelen:
an attempt to fix the mac-binaries that don't work with pre-8 XCode.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,12 @@ before_install:
 install:
   - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then export CC="gcc-4.9"; export CXX="g++-4.9"; fi
   - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then brew update; brew install ninja; fi;
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then
+      wget -O ldc.tar.xz https://github.com/ldc-developers/ldc/releases/download/v1.2.0/ldc2-1.2.0-osx-x86_64-noLTO.tar.xz;
+      mkdir ldc-noLTO;
+      tar -xvf ldc.tar.xz --strip 1 -C ldc-noLTO;
+      export DMD="ldc-noLTO/bin/ldmd2"
+    fi;
   - eval "${DC} --version"
   - pip install --user lit
   - python -c "import lit; lit.main();" --version | head -n 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ install:
       wget -O ldc.tar.xz https://github.com/ldc-developers/ldc/releases/download/v1.2.0/ldc2-1.2.0-osx-x86_64-noLTO.tar.xz;
       mkdir ldc-noLTO;
       tar -xvf ldc.tar.xz --strip 1 -C ldc-noLTO;
-      export DMD="ldc-noLTO/bin/ldmd2"
+      export DMD="ldc-noLTO/bin/ldmd2";
     fi;
   - eval "${DC} --version"
   - pip install --user lit


JohanEngelen:
prior:  binary package built with LTO enabled (XCode 8) does not work on Travis (XCode 7).
test result: binary package built _without_ LTO enabled _does_ work on Travis

kinke:
Hmm, does that mean that when building LDC itself using LTO, we need to make sure first that the clang host compiler for the C parts uses a compatible LLVM bitcode version to the LLVM version the produced LDC uses?

JohanEngelen:
The C++ compiler and LDC compiler have to be somewhat compatible in bitcode yes, but: older bitcode versions are supported by newer LLVM versions. LDC will specifically direct the linker to use the libLTO-ldc.dylib that is packaged with LDC. This means that if the C++ compiler produces an older bitcode version than LDC, things still work. (that's how it works at the moment on my machine for example)

But the LDC LTO build is not supposed to leave behind any bitcode anywhere. Somehow, this is not true and when executing LDC some bitcode interpretation/compilation is required. I don't know yet which part that is. I think it is related to XCode supporting generating executables that can run on different architectures, and somehow we enable that?

JohanEngelen:
When `(C++ LLVM version) <= (LDC LLVM version)` LTO (should) work fine (save for issues with different default triples between the compilers :( ).

When `(C++ LLVM version) > (LDC LLVM version)` things do not work automatically, because we tell the linker to use the older LDC LLVM version's code. This is "fixed" by removing/renaming `libLTO-ldc.dylib` (or `LLVMgold-ldc.so` on Linux), such that the system default LTO linker plugin is used.

kinke:
It happens when building the dummy executable [here](https://github.com/ldc-developers/ldc/blob/master/cmake/Modules/ExtractDMDSystemLinker.cmake#L33) which seems to suggest that the linker doesn't get the host LDC's `libLTO-ldc.dylib`.

JohanEngelen:
When using our CMake option `-DLDC_BUILD_WITH_LTO=...`, it's added for Mac here: https://github.com/ldc-developers/ldc/blob/master/cmake/Modules/HandleLTOPGOBuildOptions.cmake#L37-L44
For Linux, extra work may be needed.

(Note it also works when `D_COMPILER_FLAGS` contains `-flto=...` because then the linker line output with `-v`contains `-lto_library ...../libLTO-ldc.dylib`.)

kinke:
Wait, it's added for `CMAKE_{EXE,SHARED}_LINKER_FLAGS`, but those aren't used when executing the process directly in that CMake script. But I would assume that invoking the host LTO-enabled-LDC would automatically forward its .dylib to the linker. Debugging Travis a bit by showing the actually used command line and enabling verbose output for the linker would certainly help.

JohanEngelen:
The travis script fails at a part where LTO is not enabled right?  `ldc2 -v dummy.d` should work, without any bitcode/LTO stuff happening, and so we also shouldn't pass LTO things to the linker.

kinke:
Well we use it to determine the linker flags to be used for the gcc/clang custom command when linking a D executable manually [here](https://github.com/ldc-developers/ldc/blob/master/CMakeLists.txt#L595), so it better contain the LTO stuff as well if LDC is to be built with LTO, right?

JohanEngelen:
Wait, I am confused, I think we are discussing two things here.
The CMake script works fine in creating an LTO-optimized LDC: otherwise I couldn't use it on my machine. However, somehow, the output LDC cannot be transfered to another machine, and simply doing `ldc2 -v dummy.d` fails with a bitcode error, although there shouldn't be any bitcode around anywhere in that command.  I am not checking whether somehow bitcode is seeping into druntime/phobos/whatever libs we ship with LDC.

kinke:
Ah yeah sorry, `D_COMPILER_FLAGS` apparently is really empty, I thought it'd contain LTO flags.
I guess building LDC with LTO enabled shouldn't have any effect on generated druntime/Phobos but apparently there is, and that's the issue.

JohanEngelen:
Exactly.
The file sizes of druntime/phobos/profilert are different for LTO-LDC and normal-LDC, so that's bad.

kinke:
Hmm, but what I desribed above would be another issue, i.e., when linking LDC manually (default for OSX), we only use `-flto=...` from `DDMD_DFLAGS` (set in `HandleLTOPGOBuildOptions.cmake`) directly for host clang, i.e., it doesn't get the path to host LDC's .dylib. The LTO flag should be used for `ldmd2 -v dummy.d` only in that case, so that we extract all required flags from the linker command line produced by the host LDC.

JohanEngelen:
So indeed we build (parts of) the shipped libs with LTO enabled, that's bad:
```
[736/1264] /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc  -DLDC_WITH_PGO -I../. -I../ddmd -I../ddmd/root -Iddmd -I../ -isystem /Users/johan/llvm/trunk/install/include -flto=thin -O3 -DNDEBUG     -DCOMPILER_RT_HAS_ATOMICS=1 -DCOMPILER_RT_HAS_FCNTL_LCK=1 -MMD -MT runtime/CMakeFiles/ldc-profile-rt.dir/profile-rt/profile-rt-50/InstrProfilingValue.c.o -MF runtime/CMakeFiles/ldc-profile-rt.dir/profile-rt/profile-rt-50/InstrProfilingValue.c.o.d -o runtime/CMakeFiles/ldc-profile-rt.dir/profile-rt/profile-rt-50/InstrProfilingValue.c.o   -c ../runtime/profile-rt/profile-rt-50/InstrProfilingValue.c
```

JohanEngelen:
All our C/CXX flags that we use for building LDC itself are also used for compiling the C/CXX source files in the runtime libs. Instead of using `CMAKE_CXX_FLAGS`, we should be using `EXTRA_CXXFLAGS` right? (e.g. for `-DOPAQUE_VTBLS` etc.)
And then for LTO, instead of adding it globally to `CMAKE_CXX_FLAGS CMAKE_C_FLAGS CMAKE_EXE_LINKER_FLAGS CMAKE_SHARED_LINKER_FLAGS`, it should be added locally.

JohanEngelen:
Yes, I think you are right about adding the LTO flag to the `ldmd2 -v dummy.d` is good.

kinke:
Afaik, we only use CMAKE_C_FLAGS for the runtimes (and mostly if not exclusively CMAKE_CXX_FLAGS for LDC), so using `CMAKE_CXX_FLAGS` for LDC-specific flags may be okay (although not very nice).
Yes, more local LTO flags sound good and should fix both issues.

JohanEngelen:
I'm working on it.
I am reworking things into using `LDC_CXX_FLAGS` (we already have it!) instead of `CMAKE_CXX_FLAGS`. It's not a big deal: there is only one cpp file in the runtime (profile-rt), but still.

JohanEngelen:
@klickverbot Was it intentional not to include `DDMD_DFLAGS` here https://github.com/ldc-developers/ldc/blob/master/cmake/Modules/ExtractDMDSystemLinker.cmake#L28-L30 ?

kinke:
It probably wasn't, and there should be no harm in including it (in addition to `DDMD_LFLAGS` as well!), it's used in the non-manually-linking case for all D executables [here](https://github.com/ldc-developers/ldc/blob/master/CMakeLists.txt#L599).

JohanEngelen:
I think I got things to work correctly now. Testing whether cross-C++-D LTO still works for the LDC code...

PR #2085, opened 2017-04-30
Travis: use LLVM 4.0 on Mac.

JohanEngelen:


JohanEngelen:
Let's test with the version we ship with :)

PR #2091, opened 2017-05-03
Implement -fsave-optimization-record

### LINKED ISSUE #2089: Enhancement: add flag to save LLVM optimization record
Implement `-fsave-optimization-record` similar as to what Clang has now. Looks like a very straightforward addition to LDC.

See here: http://llvm.org/devmtg/2016-11/Slides/Nemet-Compiler-assistedPerformanceAnalysis.pdf

Matt Godbolt's Compiler Explorer has support for it: http://lists.llvm.org/pipermail/llvm-dev/2017-May/112578.html

Implementation hints:
- Use `LLVMContext::setDiagnosticsOutputFile`
- For inspiration: look at Clang source of how setDiagnosticsOutputFile is called, and what functionality is behind `fsave-optimization-record`.

JohanEngelen:
Resolves #2089 .

### DIFF
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -13,17 +13,63 @@
 #include "mars.h"
 #include "module.h"
 #include "scope.h"
+#include "driver/cl_options.h"
 #include "driver/linker.h"
 #include "driver/toobj.h"
 #include "gen/logger.h"
 #include "gen/modules.h"
 #include "gen/runtime.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Support/ToolOutputFile.h"
+#include "llvm/Support/YAMLTraits.h"
 
 /// The module with the frontend-generated C main() definition.
 extern Module *g_entrypointModule;
 
 /// The module that contains the actual D main() (_Dmain) definition.
 extern Module *g_dMainModule;
+
+namespace {
+
+std::unique_ptr<llvm::tool_output_file>
+createAndSetDiagnosticsOutputFile(IRState &irs, llvm::LLVMContext &ctx,
+                                  llvm::StringRef filename) {
+  std::unique_ptr<llvm::tool_output_file> diagnosticsOutputFile;
+
+#if LDC_LLVM_VER >= 400
+  // Set LLVM Diagnostics outputfile if requested
+  if (opts::saveOptimizationRecord.getNumOccurrences() > 0) {
+    llvm::SmallString<128> diagnosticsFilename;
+    if (!opts::saveOptimizationRecord.empty()) {
+      diagnosticsFilename = opts::saveOptimizationRecord.getValue();
+    } else {
+      diagnosticsFilename = filename;
+      llvm::sys::path::replace_extension(diagnosticsFilename, "opt.yaml");
+    }
+
+    std::error_code EC;
+    diagnosticsOutputFile = llvm::make_unique<llvm::tool_output_file>(
+        diagnosticsFilename, EC, llvm::sys::fs::F_None);
+    if (EC) {
+      irs.dmodule->error("Could not create file %s: %s",
+                         diagnosticsFilename.c_str(), EC.message().c_str());
+      fatal();
+    }
+
+    ctx.setDiagnosticsOutputFile(
+        llvm::make_unique<llvm::yaml::Output>(diagnosticsOutputFile->os()));
+
+    // If there is instrumentation data available, also output function hotness
+    if (!global.params.genInstrProf && global.params.datafileInstrProf)
+      ctx.setDiagnosticHotnessRequested(true);
+  }
+#endif
+
+  return diagnosticsOutputFile;
+}
+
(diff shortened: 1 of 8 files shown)


driver/codegenerator.cpp
```diff
@@ -188,7 +234,14 @@ void CodeGenerator::writeAndFreeLLModule(const char *filename) {
       {llvm::MDString::get(ir_->context(), Version)};
   IdentMetadata->addOperand(llvm::MDNode::get(ir_->context(), IdentNode));
 
+  std::unique_ptr<llvm::tool_output_file> diagnosticsOutputFile =
+      createAndSetDiagnosticsOutputFile(*ir_, context_, filename);
```
JohanEngelen: I am not sure about this. Other suggestions for where else to place this?

JohanEngelen:
Any concerns?

kinke:
Looks good to me and seems like a very neat feature. As soon as it works online, be sure to announce it. ;) @rainers: VisualD supporting this would be pretty darn cool. :]

rainers:
> VisualD supporting this would be pretty darn cool. :]

Adding the option is easy, but I guess you mean having the visual feedback as in the slides. Not so easy ;-)

JohanEngelen:
Having the visual feedback inside Visual Studio is quite some work I imagine. But using llvm/utils/opt-viewer/opt-viewer.py generates a nice webpage that shows the optimization remarks.

rainers:
>  But using llvm/utils/opt-viewer/opt-viewer.py generates a nice webpage that shows the optimization remarks.

I tried that but the python script failed with undecipherable errors (after installing the missing modules):
```
C:\Users\rainer\Documents\Visual Studio 14\Projects\ConsoleApp5\x64\Debug>C:\l\Python34\python.exe C:\s\d\ldc\llvm\utils\opt-viewer\opt-viewer.py opt.report html
For faster parsing, you may want to install libYAML for PyYAML
For faster parsing, you may want to install libYAML for PyYAML
For faster parsing, you may want to install libYAML for PyYAML
For faster parsing, you may want to install libYAML for PyYAML
For faster parsing, you may want to install libYAML for PyYAML
For faster parsing, you may want to install libYAML for PyYAML
For faster parsing, you may want to install libYAML for PyYAML
For faster parsing, you may want to install libYAML for PyYAML
For faster parsing, you may want to install libYAML for PyYAML
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "C:\l\Python34\lib\multiprocessing\pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "C:\l\Python34\lib\multiprocessing\pool.py", line 44, in mapstar
    return list(map(*args))
  File "C:\s\d\ldc\llvm\utils\opt-viewer\optrecord.py", line 178, in get_remarks
    if not hasattr(remark, 'DebugLoc') or remark.key in all_remarks:
  File "C:\s\d\ldc\llvm\utils\opt-viewer\optrecord.py", line 120, in key
    for (key, value) in arg.iteritems():
AttributeError: 'dict' object has no attribute 'iteritems'
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\s\d\ldc\llvm\utils\opt-viewer\opt-viewer.py", line 229, in <module>
    all_remarks, file_remarks, should_display_hotness = optrecord.gather_results(pmap, args.yaml_files)
  File "C:\s\d\ldc\llvm\utils\opt-viewer\optrecord.py", line 195, in gather_results
    remarks = pmap(get_remarks, filenames)
  File "C:\l\Python34\lib\multiprocessing\pool.py", line 260, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "C:\l\Python34\lib\multiprocessing\pool.py", line 599, in get
    raise self._value
AttributeError: 'dict' object has no attribute 'iteritems'
```
Any idea what's wrong here? If it worked, the build script could generate the html files and provide a clickable link that opens them within Visual Studio.

JohanEngelen:
Looks like an LLVM bug (python2 vs python3, works for me with python2). Does the answer here fix it? `https://stackoverflow.com/questions/37803745/attributeerror-dict-object-has-no-attribute-iteritems-flask-sqlalchemy-err`

rainers:
Thanks, using Python 2.7 works.

One problem I noticed is that the files in the report have the directory stripped from the name, so the source code annotation is not shown for code generated from imported templates or modules in sub packages (without listing all possible directories with -source-dir). Is there an LLVM option that allows keeping the full path?

JohanEngelen:
Could also be that we don't give the full path to LLVM somewhere. The full path is passed in debuginfo, right? Just checking.

rainers:
Yes, debug info seems to be fine.

rainers:
The directory is a separate property in DIFile record, though, e.g. `!5 = !DIFile(filename: "srcname.d", directory: "C:\5Cs\5Cd\5Cldc\5Cldc\5Ctests\5Cdebuginfo") `. Maybe it's not used when generating the report or it dislikes back-slashes.

PR #2092, opened 2017-05-04
Set `noalias` LLVM attribute for `immutable` pointer parameters.

JohanEngelen:


JohanEngelen:
@kinke I am not sure about the byval stuff. Could you have a look too?

kinke:
Lgtm, it should be fine for `byval` args. Basically, if a D `struct S` is to be passed with `byval`, the LL caller passes an `S*` `&arg`, LLVM performs a memcpy onto the callee parameters stack, and the callee's LL parameter is then a `S* param` pointing to that copy in the stack. I.e., the callee's LL param is definitely not aliasable by another param, but `byval` should already imply that (and so adding `noalias` explicitly shouldn't be needed but apparently also doesn't do any harm).

JohanEngelen:
Exactly what I thought. (applying it on byval too simplifies the logic in the function: we can apply it in one place, instead of in two places)

kinke:
Hmm on the downside, it'll render our textual IR more verbose (if it's really already implied for `byval`).

JohanEngelen:
@kinke I'm having trouble generating a function that has a `byval` param, help! ;)

JohanEngelen:
Quickly checking the optimized output, and `byval` seems to imply `noalias`.

kinke:
(Passing things (except vectors) bigger than `2 * size_t.sizeof` by value will almost always result in `byval`).

JohanEngelen:
Next step would be optimizing `immutable T[]`, immutable slices... Maybe this: http://llvm.org/docs/LangRef.html#noalias-and-alias-scope-metadata  but that's a lot to do for every load...

gen/functions.cpp
```diff
@@ -176,6 +179,9 @@ llvm::FunctionType *DtoFunctionType(Type *type, IrFuncTy &irFty, Type *thistype,
       } else {
         // Add sext/zext as needed.
         attrs.add(DtoShouldExtend(loweredDType));
+        if ((loweredDType->ty == Tpointer) && loweredDType->isImmutable()) {
+          attrs.add(llvm::Attribute::NoAlias);
+        }
```
kinke: You could relax that to checking whether the pointee is immutable, the pointer itself doesn't have to, right?

JohanEngelen: Thanks, good point.

kinke:
Yeah I looked it up too when I saw the PR. ;)

> Another is when the two pointers are only ever used for reading memory.

JohanEngelen:
I'll add an extra line in the comments.
Works for classes now too.

JohanEngelen:
done

tests/codegen/immutable_reference_loading.d
```diff
@@ -0,0 +1,147 @@
...
+    // CHECK: ret
+}
+
+// CHECK-LABEL: define{{.*}} @immutptr_ref
+auto immutptr_ref(ref immutable float* a)
+{
+    // CHECK: = load
+    auto temp = *a;
+    clobber();
+    // CHECK: = load
```
kinke: Doesn't sound right, does it?

kinke:
Adding immutable slices should be straight-forward and possibly nicely pay off for speed, as it will affect all D strings. :)

tests/codegen/immutable_reference_loading.d
JohanEngelen: ah you are right. transitive immutable (confused by C++ mindset)
This falls under the same topic as immutable slices. We have to put noalias on `immutable T**` loads.
I'll remove this test for now.

JohanEngelen:
I'll save the immutable slices and `immutable T**` stuff for another time.

kinke:
I hoped a `noalias` for the LL slice aggregate/pair param would do it. I think 'any pointer based on <param>' includes a pointer obtained by GEPing the 2nd component of the pair, see [here](http://llvm.org/docs/LangRef.html#pointeraliasing) (`The “based on” relationship is transitive.`).

JohanEngelen:
Yeah, the last section of that text is about what we need. I think we need to adorn metadata to the load instructions (https://github.com/ldc-developers/ldc/pull/2092#issuecomment-299515191)

JohanEngelen:
I am no longer sure this is the right path. I think it is valid to add noalias, but I do not know if there isn't a better, more correct solution. LLVM has a bunch of intrinsics to indicate invariance. I am investigating and will also write to the mailing list when i get around to it. Interesting stuff!

JohanEngelen:
(close+reopen to rerun CI; since then, I've learned a little more about LLVM's invariant attributes, so this PR may take a different route. But I'm curious to see if it still works correctly, because this enables aggressive optimization and relies on `immutable` really meaning `immutable`)

JohanEngelen:
This optimization is invalid in general, because `immutable` does not mean `immutable`:

1. can cast an immutable object to an immutable ubyte[]
2. read from the ubyte[]
3. call class method on the object, such that monitor field is modified
4. read from ubyte[] again --> must treat ubyte[] as mutable data here

Thus code like this cannot be optimized well:
```
ubyte foo(immutable ubyte[] arr) {
   auto temp = arr[2];
   callFunction();
   return temp + arr[2]; // second read from memory cannot be optimized (elided nor reordered), because it just might be the monitor field of some object
}
```

JohanEngelen:
it's a WIP. Work on the D spec is on its way to change the spec and clarify what optimizations _are_ possible with `immutable`.

Issue #2094, opened 2017-05-05
strange struct code gen


arBmind:
https://godbolt.org/g/9G8pG3

```d
private struct P { float x, y; }

auto len(P p) {
    return p.x*p.x + p.y*p.y;
}

auto len(float x, float y) {
    return x*x + y*y;
}
```

The assembler for the struct variant:
```asm
pure nothrow @nogc @safe float example.len(example.P):
        movq    rax, xmm0
        movd    xmm1, eax
        mulss   xmm1, xmm1
        shr     rax, 32
        movd    xmm0, eax
        mulss   xmm0, xmm0
        addss   xmm0, xmm1
        ret
```

The route through the register `rax` seems wrong. The code without the `struct` is fine.

You might want to take a look. Thank you for the good work!

kinke:
Thanks for the report. Clang's result: https://godbolt.org/g/jF71l9

kinke:
According to the unoptimized IR, the only difference is that LDC passes the struct IR-wise as `double` (in order for it to end up in ABI-dictated  `xmm0`; may already be done in DMD's `toArgTypes()`), while clang represents it as vector `<2 x float>`.

**LDC**:
```
define float @_D7current3lenFS7current1PZf(double %p_arg) #0 comdat {
  %p = alloca %current.P, align 4                 ; [#uses = 5, size/byte = 8]
  %1 = bitcast %current.P* %p to double*          ; [#uses = 1]
  store double %p_arg, double* %1
  %2 = getelementptr inbounds %current.P, %current.P* %p, i32 0, i32 0 ; [#uses = 1, type = float*]
  %3 = load float, float* %2                      ; [#uses = 1]
  %4 = getelementptr inbounds %current.P, %current.P* %p, i32 0, i32 0 ; [#uses = 1, type = float*]
  %5 = load float, float* %4                      ; [#uses = 1]
  %6 = fmul float %3, %5                          ; [#uses = 1]
  %7 = getelementptr inbounds %current.P, %current.P* %p, i32 0, i32 1 ; [#uses = 1, type = float*]
  %8 = load float, float* %7                      ; [#uses = 1]
  %9 = getelementptr inbounds %current.P, %current.P* %p, i32 0, i32 1 ; [#uses = 1, type = float*]
  %10 = load float, float* %9                     ; [#uses = 1]
  %11 = fmul float %8, %10                        ; [#uses = 1]
  %12 = fadd float %6, %11                        ; [#uses = 1]
[...]

PR #2095, opened 2017-05-05
[WIP] Non-Windows x86_64: Rewrite HFAs to fp vectors

kinke:
Fixes suboptimal codegen exposed by issue #2094.


gen/abi-x86-64.cpp
```diff
@@ -102,7 +102,34 @@ bool passByVal(Type *ty) {
...
+  LLType *dmdLLType = dmd_abi::getAbiType(ty);
+
+  if (dmdLLType) {
+    // Check for Homogeneous Floating-point Aggregates, and rewrite to a vector
+    // type in that case.
+    if (ty->ty == Tcomplex32) {
+      return llvm::VectorType::get(LLType::getFloatTy(gIR->context()), 2);
+    }
+    if (ty->ty == Tstruct) {
+      // TODO: check restrictions
```
klickverbot: Huh?

kinke: That's the remaining WIP part after the alignment fixes today and most likely the cause of the C(++) interop regressions, i.e., figuring out the size threshold for an HFA-to-vector rewrite or potential other restrictions. I started with a limit of 8 components, to account for `<8 x float>` (AVX). Haven't checked yet into what e.g. `<8 x double>` translates register-wise for non-AVX2 code; and then we need to keep track of the register usage manually for the SysV ABI...

kinke:
I've had another look at the SysV ABI specs; it's less optimal than I thought, e.g., `struct Vector { float x, y, z; }` and `struct ComplexDouble { double re, im; }` are to be passed in two XMM registers. The only exception are the C(++) `__m<NNN>` types, i.e., in our case D vector types.

So a complex type defined in a library (Phobos) should probably look like `struct { double2 values; }` in order to pass it in a single register.
Unfortunately, allowing direct access to the components via something like
```
struct ComplexDouble
{
    union
    {
        double2 values;
        struct { double re, im; }
     }
}
```
should prevent that optimization if I understood correctly.

Other problems: in master, a 3-floats struct is rewritten as 2xdouble, thx to `toArgTypes()`, which means that the load actually (a) assumes 8-bytes alignment while the actual one is 4 (but x86_64 only, so probably no problem) and (b) reads 4 bytes more than actually allowed...

thewilsonator:
Would having the definition as 
```
struct ComplexDouble
{
    double2 values;
    @ property double re() { return values.__array[0];}
    @ property void re(double d) { values.__array[0] = d;}
    @ property double im() { return values.__array[1];}
    @ property void im(double d) { values.__array[1] = d;}
}
```
not hinder the opts?

thewilsonator:
Hmm, the other thing to consider with redefining phobos' complex as vectors is: would it interfere with (re)vectorisation (considering `re` and `im` are involved in different operations most of the time)?

jpf91:
> Hmm, the other thing to consider with redefining phobos' complex as vectors is: would it interfere with (re)vectorisation (considering re and im are involved in different operations most of the time)?

I'm not sure why exactly vectorisation should be a problem, but `involved in different operations` is an important point. I don't know the X86 ABI that well, but maybe some AArch64 insights can help:

On AArch64, float registers are 16 byte wide, so a complex double can fit into one register. Nevertheless, the ABI defines that aggregates with multiple same-type floating point members are passed in multiple float registers, one per member. E.g. `struct {double re, im;}` is passed in two registers. Only explicitly annotated 'multi-value' types - Short-vectors such as `double2` - can be passed in one register. The ABI explicitly maps the C `double _Complex` type to two registers.

I guess the rationale is exactly what @thewilsonator mentioned: If you do independent operations on `re` and `im` you usually don't want to use vector operations (you'd have to always use operand vectors with one neutral element). If the neutral element is `0`, normal 8 byte loads can work, however AArch64 always loads to the lower bits. If you want the lower 8 bytes to be `0` you have to do extra work. Same if you need `1` as a neutral element. So most of the time `re` and `im` should live in independent registers and then requiring them to be repacked into one register when calling a function will hurt performance (as long as not all 8 FP registers are already used).

TLDR: I'd simply follow whatever the C ABI does for `double _complex` types. This gives C compatibility for free and usually the C ABI designers already chose the best performing solution.

kinke:
> usually the C ABI designers already chose the best performing solution.

Until some years ago, I would have thought the same. But seeing stuff like returning an empty struct in C++ (no fields) for SysV x64 requiring a hidden sret pointer to the super-duper 1 byte dummy result allocated on the caller's stack (i.e., waste a general-purpose register for the most likely unused pointer) made me more realistic. ;)

Wrt. passing 64-bit real/imaginary parts in separate registers vs. repacking and later unpacking, that might pay off for complex numbers, but I imagine it to be a different kind of story for a 2D vector (without `double2` usage, just a regular struct with 2 doubles), and of course for the rather popular 3/4D float/double vectors.
I just found it surprising, as once you pass a struct with 8 doubles, you're using up all 8 dedicated XMM registers already, while they would all fit into a single ZMM (AVX-512) register just fine.

That aside, the actual issue I'm trying to solve here is a LLVM quirk (IMO), further complicating the SysV ABI. While the SysV algorithm works via classifying each eightbyte (integer, floating-point, memory etc.), we apparently need to do further discriminations between FP eightbytes (e.g., don't rewrite a struct with 2 floats as single double, but as a vector of 2 floats with matching alignment) to get optimal optimization later on.

kinke:
[Oh and `toArgTypes()` doesn't support vector types > 128 bits as struct members - a `double4` is passed in a register, a `struct { double4 bla; }` in memory.]

Issue #115, opened 2012-06-07
stack trace missing on assert failure


thelastmammoth:
the stack trace appears in dmd, not ldmd2/ldc (even with various combination of debug flags):
(i'm on osx if that's relevant).

main4.d:

``` d
int fun(int x){ assert(x<0); return x+x;}
void main(){int x=fun(10);}
```

```
ldmd2 main4 && ./main4                                                                                                  
core.exception.AssertError@main4.d(1): Assertion failure
```

==> no stack trace

```
dmd main4 && ./main4                                                                                                
core.exception.AssertError@main4(1): Assertion failure
----------------
5   main4                               0x00000001095d2576 _d_assertm + 42
6   main4                               0x00000001095c2477 void main4.__assert(int) + 23
7   main4                               0x00000001095c2491 int main4.fun(int) + 25
8   main4                               0x00000001095c245a _Dmain + 14
9   main4                               0x00000001095d2ea9 extern (C) int rt.dmain2.main(int, char**).void runMain() + 29
10  main4                               0x00000001095d285e extern (C) int rt.dmain2.main(int, char**).void tryExec(scope void delegate()) + 38
11  main4                               0x00000001095d2ef6 extern (C) int rt.dmain2.main(int, char**).void runAll() + 58
12  main4                               0x00000001095d285e extern (C) int rt.dmain2.main(int, char**).void tryExec(scope void delegate()) + 38
13  main4                               0x00000001095d27e9 main + 237
14  main4                               0x00000001095c2444 start + 52
15  ???                                 0x0000000000000001 0x0 + 1
----------------
```

dmakarov:
I submitted a pull request https://github.com/ldc-developers/druntime/pull/18 that fixes this issue.

redstar:
I got now a stack trace, but the symbols are still missing (compiled with -g):

    core.exception.AssertError@main4.d(1): Assertion failure
    ----------------
    ./main4() [0x401ddc]
    ./main4() [0x401e02]
    ./main4() [0x40f194]
    ./main4() [0x40f069]
    ./main4() [0x40f105]
    ./main4() [0x40f069]
    ./main4() [0x40efc8]
    ./main4() [0x401f28]
    /lib64/libc.so.6(__libc_start_main+0xf5) [0x7fa85308fdc5]

Do you have an idea why?

redstar:
I get a stack trace now but the symbols are still missing (compiled with -g):

[...]

PR #2096, opened 2017-05-06
Don't eliminate frame pointer by default for debuginfo builds

kinke:
As long as no optimization level (i.e., 0) is used. Should make the OSX backtraces work by default for debug builds and may improve the debuginfo experience in general.


kinke:
I don't follow. According to #115, you OSX guys need it for back traces, and you considered the fp elim to be enabled by default for debuginfo builds as bug, but now you suggest fixing up druntime wherever it's using EBP/RBP instead? That's way out of my scope here.

kinke:
2nd commit directly pushed to master as https://github.com/ldc-developers/ldc/commit/a38cae7831cb255d4169930504ef5b807b992d21; closing.

PR #2097, opened 2017-05-06
druntime: Use _Unwind_Backtrace for more accurate backtraces on Darwin

klickverbot:
There are two things "wrong" with this patch:

 * The libunwind headers should be moved to some proper place in `sys`, and de-duplicated from `rt.*`.

 * The changes aren't marked up with `version (LDC)`.

However, both of these hardly make the current mess worse, which already fails on both counts. What should really happen is a proper refactor of these in upstream druntime, including provisions for our special cases (although we could just move DMD to use name-based skipping as well, in case its inliner ever catches up).


kinke:
About time to get this fixed for our OSX users. ;) - Thanks. I guess filtering out everything above and incl. the topmost _d_throw_exception is outside the scope of this. ;)

kinke:
Feel free to merge once 'green', seems definitely worth to be included in the next 1.3 beta.

PR #2102, opened 2017-05-07
toconstelem: Implement array literal VectorExps

### LINKED ISSUE #2101: F f = F.init fails
```
LDC - the LLVM D compiler (1.3.0git-a969bcf):
  based on DMD v2.073.2 and LLVM 4.0.0
  built with LDC - the LLVM D compiler (0.17.5git-64a274a)
  Default target: x86_64-apple-darwin16.5.0
  Host CPU: haswell
```
```d
alias F = __vector(double[2]);
struct S
{
	F f = F.init;
}

void main()
{
}
```

```
Global variable initializer type does not match global variable type!
%test3.S* @_D5test31S6__initZ
```

klickverbot:
Previously, we would always try to splat the given expression across the
whole vector. This hasn't been noticed earlier as DMD mostly generates
ArrayInitializers when initializing vectors to array literals (but
notably not when explicitly assigning VectorType.init).

GitHub: Fixes #2101.

### DIFF
--- a/gen/toconstelem.cpp
+++ b/gen/toconstelem.cpp
@@ -157,8 +157,8 @@ public:
                          e->type->toChars());
     LOG_SCOPE;
 
-    Type * const t = e->type->toBasetype();
-    Type * const cty = t->nextOf()->toBasetype();
+    Type *const t = e->type->toBasetype();
+    Type *const cty = t->nextOf()->toBasetype();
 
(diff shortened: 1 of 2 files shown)


kinke:
That was quick. ;) - Adding the testcase to `tests/compilable` wouldn't hurt I think.

JohanEngelen:
IR test please :)

PR #2104, opened 2017-05-07
Add switch '-Xcc <ccflag>'. Fix #2093.

### LINKED ISSUE #2093: Passing options directly to CC
Is there a way to pass a switch to the gcc executable (or whatever is in $CC or -gcc=) in a way that LDC don't precede it with `-Xlinker`?

I'm specifically talking about `--sysroot=` switch now. What LDC calls is:
```
my-gcc ...blablabla... -Xlinker --sysroot=/directory/of/sysroot ...blablabla...
```
which seems to do a different thing than:
```
my-gcc ...blablabla... --sysroot=/directory/of/sysroot ...blablabla...
```

The first one causes `my-gcc` to fail. Apparently, it doesn't override the default sysroot that `gcc` was built with, which causes `gcc` to pass the `--sysroot` switch to `ld` twice: once with the default path and once from `-Xlinker --sysroot=`. `ld` seems to pick up the first one only which is not the one I want.

So, having a way to pass a switch straight to `gcc` would be helpful.

epi:
Implemented as discussed in comments to #2093.

There's `assert(optnum < this->size() && "Invalid option index")` in `cl::list::getPosition`, and `size()` doesn't seem to be present for `opts::linkerSwitches`, so this assert shouldn't compile.
In what circumstances are asserts enabled when building LDC? I've tried both Release and Debug builds and it compiled.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -285,10 +285,13 @@ cl::list<std::string> transitions(
     cl::value_desc("idents"), cl::CommaSeparated);
 
 static StringsAdapter linkSwitchStore("L", global.params.linkswitches);
-static cl::list<std::string, StringsAdapter>
+cl::list<std::string, StringsAdapter>
     linkerSwitches("L", cl::desc("Pass <linkerflag> to the linker"),
                    cl::value_desc("linkerflag"), cl::location(linkSwitchStore),
                    cl::Prefix);
+
+cl::list<std::string> ccSwitches("Xcc", cl::Hidden,
+    cl::desc("Pass <ccflag> to GCC/Clang"), cl::value_desc("ccflag"));
 
 cl::opt<std::string>
     moduleDeps("deps",
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -289,7 +289,8 @@ cl::list<std::string> linkerSwitches("L",
     cl::value_desc("linkerflag"), cl::Prefix);
 
 cl::list<std::string> ccSwitches("Xcc", cl::CommaSeparated,
-    cl::desc("Pass <ccflag> to GCC/Clang"), cl::value_desc("ccflag"));
+    cl::desc("Pass <ccflag> to GCC/Clang for linking"),
+    cl::value_desc("ccflag"));
 
 cl::opt<std::string>
     moduleDeps("deps",
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -290,7 +290,7 @@ cl::list<std::string, StringsAdapter>
                    cl::value_desc("linkerflag"), cl::location(linkSwitchStore),
                    cl::Prefix);
 
-cl::list<std::string> ccSwitches("Xcc", cl::Hidden,
+cl::list<std::string> ccSwitches("Xcc", cl::CommaSeparated,
     cl::desc("Pass <ccflag> to GCC/Clang"), cl::value_desc("ccflag"));
 
 cl::opt<std::string>
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -284,11 +284,9 @@ cl::list<std::string> transitions(
         "Help with language change identified by <idents>, use ? for list"),
     cl::value_desc("idents"), cl::CommaSeparated);
 
-static StringsAdapter linkSwitchStore("L", global.params.linkswitches);
-cl::list<std::string, StringsAdapter>
-    linkerSwitches("L", cl::desc("Pass <linkerflag> to the linker"),
-                   cl::value_desc("linkerflag"), cl::location(linkSwitchStore),
-                   cl::Prefix);
+cl::list<std::string> linkerSwitches("L",
+    cl::desc("Pass <linkerflag> to the linker"),
+    cl::value_desc("linkerflag"), cl::Prefix);
 
 cl::list<std::string> ccSwitches("Xcc", cl::CommaSeparated,
     cl::desc("Pass <ccflag> to GCC/Clang"), cl::value_desc("ccflag"));
--- a/tests/linking/linker_switches.d
+++ b/tests/linking/linker_switches.d
@@ -1,0 +1,13 @@
+// Test if global order of flags passed with -Xcc, -L and pragma(lib) is preserved
+
+// REQUIRES: Linux
+
+// RUN: %ldc %s -of=%t -Xcc -DOPT1,-DOPT2 -L-L/usr/lib -L--defsym -Lfoo=5 -Xcc -DOPT3 -v | FileCheck %s
+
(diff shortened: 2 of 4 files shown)


kinke:
Thanks, looks quite alright after a first glance.

> In what circumstances are asserts enabled when building LDC?

This is governed by whether LLVM was built with assertions or not. Our AppVeyour CI uses LLVM/LDC assertions (`LLVM_ENABLE_ASSERTIONS=ON` for LLVM CMake).

epi:
That's exactly where AppVeyor build fails: `error C2039: 'size': is not a member of 'llvm::cl::list<std::string,opts::StringsAdapter,llvm::cl::parser<std::string>>'`

I'll probably solve it by using `cl::list` with built-in storage for `linkerSwitches`, instead of `global.params.linkswitches`.

Is there a recommended way I could add a unittest for `-Xcc` (and `-Xcc` vs. `-L`)?

kinke:
> I'll probably solve it by using cl::list with built-in storage for linkerSwitches, instead of global.params.linkswitches.

That won't work like that, as the front-end does tamper with `global.params.linkswitches`, e.g., when encountering a `pragma(lib, mylib)` on non-Windows IIRC.

kinke:
Wrt. testing, a unittest isn't feasible, but an integration test is, something like using `-v` to display the gcc command line and matching that one (on non-Windows platforms only as we invoke the linker directly there), see [this example](https://github.com/ldc-developers/ldc/blob/master/tests/linking/fulllto_1.d).

driver/cl_options.cpp
```diff
@@ -285,11 +285,14 @@ cl::list<std::string> transitions(
...
 
 static StringsAdapter linkSwitchStore("L", global.params.linkswitches);
-static cl::list<std::string, StringsAdapter>
+cl::list<std::string, StringsAdapter>
     linkerSwitches("L", cl::desc("Pass <linkerflag> to the linker"),
                    cl::value_desc("linkerflag"), cl::location(linkSwitchStore),
                    cl::Prefix);
 
+cl::list<std::string> ccSwitches("Xcc", cl::Hidden,
+    cl::desc("Pass <ccflag> to GCC/Clang"), cl::value_desc("ccflag"));
```
JohanEngelen: Some ideas:
- I think this should not be `cl::Hidden`. It is a useful flag and we don't want everybody to only find out about it after reading the issue tracker ;-)
- Perhaps add `cl::CommaSeparated`

epi:
@kinke
> > I'll probably solve it by using cl::list with built-in storage for linkerSwitches, instead of global.params.linkswitches.

> That won't work like that, as the front-end does tamper with global.params.linkswitches, e.g., when encountering a pragma(lib, mylib) on non-Windows IIRC.

I'm aware of that. Contents of `global.params.linkswitches` can be added after `linkerSwitches` and `ccSwitches`, without changing current behaviour.

driver/cl_options.cpp
epi: Good point.

driver/cl_options.cpp
```diff
@@ -284,11 +284,12 @@ cl::list<std::string> transitions(
...
-static cl::list<std::string, StringsAdapter>
-    linkerSwitches("L", cl::desc("Pass <linkerflag> to the linker"),
-                   cl::value_desc("linkerflag"), cl::location(linkSwitchStore),
-                   cl::Prefix);
+cl::list<std::string> linkerSwitches("L",
+    cl::desc("Pass <linkerflag> to the linker"),
+    cl::value_desc("linkerflag"), cl::Prefix);
+
+cl::list<std::string> ccSwitches("Xcc", cl::CommaSeparated,
+    cl::desc("Pass <ccflag> to GCC/Clang"), cl::value_desc("ccflag"));
```
klickverbot: Maybe add "for linking" or something so it is clear when they are invoked?

JohanEngelen: (thought I was the only one thinking that :-))

JohanEngelen:
@epi Need help adding the test?
You can go into the `.../build/tests/` directory, and run `./runlit.py -v .` to run all lit tests. See https://wiki.dlang.org/LDC_Lit-based_testsuite

epi:
@JohanEngelen Thanks, I found this page and successfully tried running existing tests. I'll submit the test of -Xcc shortly, just need to get some urgent things at work done first. :)

tests/linking/linker_switches.d
```diff
@@ -0,0 +1,13 @@
+// Test if global order of flags passed with -Xcc, -L and pragma(lib) is preserved
+
+// REQUIRES: Linux
```
epi: I suppose this test should also be run for Darwin, but not Windows. Grepping in lit I found that something like `REQUIRES-ANY: Linux Darwin` seems to be the way to achieve it, but it only makes the test unsupported under Linux.

epi: I suppose this test should also be run for Darwin, but not Windows. Grepping in lit I found that something like `REQUIRES-ANY: Linux Darwin` seems to be the way to achieve it, but it only makes this test unsupported under Linux.

JohanEngelen: IIRC the CI systems run latest Lit. In newer versions of Lit, you can do boolean expressions in `REQUIRES` and friends.
Maybe in this case you can just use `UNSUPPORTED: Windows`.

See also http://llvm.org/docs/TestingGuide.html#constraining-test-execution

epi: Thanks. This is what I was looking for.

kinke:
@epi: Almost there, the MS linker flags are still based exclusively on `global.params.linkswitches` though, see https://github.com/ldc-developers/ldc/blob/master/driver/linker.cpp#L591-L603.

epi:
@kinke: Oops. :) So the question is what `-Xcc` should do with MS linker - should it work exactly like `-L`, be ignored or disallowed?

kinke:
> on non-Windows platforms only as we invoke the linker directly there

So just no `-Xcc` support for MSVC targets.

kinke:
Travis OSX [complains](https://travis-ci.org/ldc-developers/ldc/jobs/231548516#L7259) about unknown `--defsym` for its `ld`. I'm also surprised `gcc -Xlinker foo=5` apparently works. ;)

epi:
@kinke: Seems that my attempt to find the most innocent flags has failed. :) I don't have access to any OSX machine, so I could keep guessing, but I think for this test it doesn't matter if the linker succeeds or not, as long as flags are passed to it as expected.
I'm thinking of something along the lines of:
```
// RUN: /bin/sh -c '%ldc %s -of=%t -Xcc -DOPT1,-DOPT2 -L-L/usr/lib -L--defsym -Lfoo=5 -Xcc -DOPT3 -v || true' | FileCheck %s
```

kinke:
Sounds good! You may also want to redirect stderr for the LDC command, just for tidiness.

JohanEngelen:
Another thing you can try is using `not` and pass something that the linker is bound to fail on.

kinke:
Thanks, it'll be in 1.3. :)

PR #2111, opened 2017-05-13
Fix build against llvm50 master

rainers:


kinke:
The problem with Circle is that the [official LLVM](http://apt.llvm.org/) repo still has a version that is a month old for Ubuntu 12.04 (and all Ubuntu releases older than 16.04, the other ones are brand-new).

JohanEngelen:
Perhaps we need to spend the time :'( :'(  to try CircleCI 2.0, I think there we can run Ubuntu 16. https://circleci.com/blog/ubuntu-12-04-precise-build-image-end-of-life-warning/

JohanEngelen:
Thanks @rainers . I hit an assert when compiling std/uni.d. 
```
FAILED: cd /Users/johan/ldc/johan && /Users/johan/ldc/johan/buildtrunk/bin/ldc2 --output-o -c -I/Users/johan/ldc/johan/runtime/druntime/src -I/Users/johan/ldc/johan/runtime/druntime/src/gc /Users/johan/ldc/johan/runtime/phobos/std/uni.d -of/Users/johan/ldc/johan/buildtrunk/runtime/std/uni-debug.o -w -g -link-debuglib -I/Users/johan/ldc/johan/runtime/phobos
Assertion failed: ((!OldAlign || !NewAlign || OldAlign == NewAlign) && "Attempt to change alignment!"), function addAttributes, file ../lib/IR/Attributes.cpp, line 1062.
```
Haven't looked further into it yet.

JohanEngelen:
The assert was due to an LLVM bug. It's been fixed: http://llvm.org/viewvc/llvm-project?view=revision&revision=303458

PR #2113, opened 2017-05-14
Allow multiple occurrences of all command-line flags

kinke:
Except for `-sanitize` to make clear that they cannot be combined.

### DIFF
--- a/driver/cache.cpp
+++ b/driver/cache.cpp
@@ -114,8 +114,7 @@ namespace {
 
 // Options for the cache pruning algorithm
 llvm::cl::opt<bool> pruneEnabled("cache-prune",
-                                 llvm::cl::desc("Enable cache pruning."),
-                                 llvm::cl::ZeroOrMore);
+                                 llvm::cl::desc("Enable cache pruning."));
 llvm::cl::opt<unsigned long long> pruneSizeLimitInBytes(
     "cache-prune-maxbytes",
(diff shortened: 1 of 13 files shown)


JohanEngelen:
I don't think it is a good idea to allow zeroOrMore for all these params. For example, `-cache-prune -cache-prune` == `-cache-prune`, that's understandable and OK. But things like `-cache-prune-maxbytes=10 -cache-prune-maxbytes=20` == `-cache-prune-maxbytes=20` is not OK in my opinion and should give an error. I believe it is common practice to report conflicting cmdline switches, instead of using the last one.

kinke:
I had some doubts about the rather special `-cache*` options as well. But for almost all other options I think it does make sense, based on our own CMake experience with (C++) flags coming in from CMake defaults, LLVM, user and numerous hardcoded ones; the MS compiler warns about some duplicates/conflicts IIRC, and that makes us having to delete related options first before adding new ones etc.

JohanEngelen:
The `-cache*` options are not a special case, my point was for all flags that accept values but not a list. Quickly checked GCC and Clang: they indeed do allow overriding, `-O3 -std=c++11 -O0 -O3 -std=c++03` is accepted fine. So I guess we can do the same.

kinke:
Reworked to allow `-cache-prune*` options max once (only allowing to override `-cache-retrieval`).

PR #2115, opened 2017-05-14
[WiP] Use pooled identifiers for magic UDAs and dcompute

### LINKED ISSUE #2011: Pooled identifiers for LDC specific symbols & modules
We should extend upstream with a mechanism for adding backend-specific members to Identifier (probably also refactoring DMD-specific stuff into that). Also do something similar for "magic modules"  like `ldc.attributes` to allow detection by pointer comparison.

Should help optimise `gen/uda.cpp` and `gen/dcomputetypes.cpp`.

thewilsonator:
Fixes #2011


thewilsonator:
Urgh, I meant to do this on my fork. Ignore this

Issue #2105, opened 2017-05-10
`-output-ll -of test.s` outputs `test.ll` and `test.s`


JohanEngelen:
When specifying an output file together with `-output-ll`, LLVM IR is written to the specified output file:
`ldc2 -c -output-ll test.d -of dummy.blah` creates the file `dummy.blah` and writes LLVM IR to that file.
 But if the output filename ends with `.s`, 
`ldc2 -c -output-ll test.d -of dummy.s`
_two_ files are created: `dummy.ll` and `dummy.s`. Clearly a bug, and I do not know how to write LLVM IR to a `.s` file. (one of the effects of this bug is that we cannot see LLVM IR output at d.godbolt.org)

kinke:
This is a 'feature': https://github.com/ldc-developers/ldc/blob/master/driver/main.cpp#L533-L558

It recognizes the output extension as assembly and thus enables `-output-s`. As we then have 2 output files for that module (but only a single output filename), they each get the default extension for uniqueness, and so the .ll contains the IR and .s the assembly.

kinke:
The autodetection should at least not enforce this behavior as it currently does, so that an explicit `-output-s=false` could be used under such rather special circumstances.

kinke:
Fixed by #2118.

JohanEngelen:
Thanks @kinke

PR #2118, opened 2017-05-14
Fix object file type autodetection (issue #2105)

kinke:


kinke:
@JohanEngelen: Are you fine with this or do you insist on a test? ;)

driver/main.cpp
```diff
@@ -530,29 +530,24 @@ void parseCommandLine(int argc, char **argv, Strings &sourceFiles,
 
   // LDC output determination
 
-  // if we don't link, autodetect target from extension
-  if (!global.params.link && !global.params.lib && global.params.objname) {
+  // if we don't link and there's no `-output-*` switch but an `-of` one,
+  // autodetect type of desired 'object' file from file extension
+  if (!global.params.link && !global.params.lib && global.params.objname &&
+      global.params.output_o == OUTPUTFLAGdefault) {
```
kinke: Some more context: `global.params.output_o` equals `OUTPUTFLAGdefault` here if `-output-o` is omitted (neither enabled nor disabled) AND none of the others are enabled (so omitted or explicitly disabled).

JohanEngelen: Thanks, this helped :)

JohanEngelen:
So if I understand correctly, with this PR:
`ldc2 test.d -output-ll -of=test.s` will only output IR to the file `test.s`, right?

Yeah, tests would help. Tedious, I know, so perhaps you can spend the time elsewhere.

kinke:
> ldc2 test.d -output-ll -of=test.s will only output IR to the file test.s, right?

Yep, but that's no change in behavior, that was already the case, just verified with 1.3.0-beta1.

JohanEngelen:
Ah I meant with `-c`. (surprised things works without `-c` !).
So: `ldc2 test.d -c -output-ll -of=test.s`.  If that only outputs `test.s` with LLVM IR, I'm happy.

kinke:
> (surprised things works without -c !).

I used it a couple of times in lit tests IIRC. It makes sense though, doesn't it? Only write IR to the output file, skip codegen & linking.
Yep, adding `-c` makes no difference with this PR anymore if there's at least one (non-disabled) `-output-*` option, just verified. ;)

JohanEngelen:
Clang is complaining about the code:
```
../driver/main.cpp:546:32: warning: comparison of constant 'BOU_FALSE' (2) with expression of type 'bool' is always true [-Wtautological-constant-out-of-range-compare]
    } else if (opts::output_ll != cl::BOU_FALSE &&
               ~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~
../driver/main.cpp:550:32: warning: comparison of constant 'BOU_FALSE' (2) with expression of type 'bool' is always true [-Wtautological-constant-out-of-range-compare]
    } else if (opts::output_bc != cl::BOU_FALSE &&
               ~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~
../driver/main.cpp:554:31: warning: comparison of constant 'BOU_FALSE' (2) with expression of type 'bool' is always true [-Wtautological-constant-out-of-range-compare]
    } else if (opts::output_s != cl::BOU_FALSE &&
               ~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~
```

kinke:
Thanks Johan, fixed in 6c6379d534ef44cf38cafed5707f0d39310e820d.

PR #2120, opened 2017-05-18
Remove magic +1 constant for parameter indexing.

JohanEngelen:
LLVM may change the indexing offset in the future. See http://llvm.org/viewvc/llvm-project?view=revision&revision=302060

We currently have an attribute bug when using LLVM trunk, where we hit an assertion with alignment attributes. Could be an LLVM bug, but also could be a bug on our end. This PR removes one potential source of bugs.


ir/irfuncty.cpp
```diff
@@ -99,17 +99,19 @@ LLValue *IrFuncTy::getParamLVal(Type *dty, size_t idx, LLValue *val) {
...
 
 // handle implicit args
 #define ADD_PA(X)                                                              \
   if (X) {                                                                     \
-    newAttrs.add(idx, (X)->attrs);                                             \
+    newAttrs.addToParam(idx, (X)->attrs);                                      \
     idx++;                                                                     \
   }
 
-  ADD_PA(ret)
```
JohanEngelen: this here was a latent bug. If `ret == nullptr`, the indexing was off-by-one for the rest of the function...

JohanEngelen:
Pretty sure it is a bug in LLVM. This line https://github.com/llvm-mirror/llvm/blob/4679522422c140f4008e793fd89e4dea9139e6bb/lib/IR/Attributes.cpp#L1061 uses the Index with +1 offset for parameters and passes it to getParamAlignment that expects 0-based ArgNo parameter indexing. So it's comparing the alignment of two different parameter slots.

PR #2125, opened 2017-05-21
[dcompute] targets & ABIs

thewilsonator:


gen/dcompute/abi-rewrites.h
```diff
@@ -0,0 +1,39 @@
...
+#include "gen/tollvm.h"
+#include "gen/dcompute/druntime.h"
+
+struct DComputePointerRewrite : ABIRewrite {
+  LLType* type(Type* t) override {
+    auto ptr = toDcomputePointer(static_cast<TypeStruct*>(t)->sym);
+    return ptr->toLLVMType();
+  }
+  LLValue *getLVal(Type *dty, LLValue *v) override {
+    return DtoBitCast(v,this->type(dty));
```
thewilsonator: I'm trying to remove the enclosing struct, is this the right way to do it?
As in move the GEP to the other side of the function call boundary so that the Pointer's alias this member is passed and not the struct so that the `function({ i32 addrspace(3)*})` becomes `function( i32 addrspace(3)*)`.

thewilsonator: Added a TODO, will investigate later.

PR #2129, opened 2017-05-21
Dcompute target fix

thewilsonator:
Fix travis 3.5. Still not sure why 3.6 would fail. Also fix the `-Wreorder` warnings.


gen/dcompute/targetOCL.cpp
```diff
@@ -105,8 +107,8 @@ class TargetOCL : public DComputeTarget {
     llvm::SmallVector<llvm::Metadata *, 8> kernelMDArgs;
     kernelMDArgs.push_back(llvm::ConstantAsMetadata::get(llf));
     // MDNode for the kernel argument address space qualifiers.
-    llvm::SmallVector<llvm::SmallVector<llvm::Metadata *, 8>,6> paramArgs;
-    llvm::SmallVector<const char*,6> args = {
+    llvm::SmallVector<llvm::Metadata *, 8>[count_KernArgMD] paramArgs;
+    const char*[count_KernArgMD] args = {
```
klickverbot: By the way, you don't need the horribly C89-like `count_KernArgMD` crutch – a plain array without length specified will work just fine with range-based for.

gen/dcompute/targetOCL.cpp
```diff
@@ -115,10 +117,8 @@ class TargetOCL : public DComputeTarget {
       "kernel_arg_name"
     };
       
-    for (auto &str : args) {
-      llvm::SmallVector<llvm::Metadata *, 8> tmp;
-      tmp.push_back(llvm::MDString::get(ctx, str));
-      paramArgs.push_back(tmp);
+    for (int i=0; i<count_KernArgMD; i++) {
```
klickverbot: After fixing the build, please `clang-format` all the `dcompute` code. (I don't want to do it myself as it would break all your merges.)

gen/dcompute/targetOCL.cpp
thewilsonator: Thanks! I didn't know that.

klickverbot: It's kind of magical, and also kind of scary (since it looks like it shouldn't work).

thewilsonator:
@klickverbot Any idea why travis and appveyor aren't running at all?

thewilsonator:
Closing and reopening to retriever.

thewilsonator:
Nope, still a flat line.

thewilsonator:
3.6 and 3.6 travis now pass. Circle fails due to the issue fixed by #2128, semaphore fails with the std.file test. Merging now.

PR #2130, opened 2017-05-21
Debug info fixes for class types

Syniurge:
Fix the `DIType` produced for static arrays of classes and pointer to classes, and fix the size of class types in `DIBuilder::CreateCompositeType`.

Ex.: `Array!FuncDeclaration` was causing an infinite recursion crash in GDB, because for its « `FuncDeclaration[SMALLARRAYCAP] smallarray` » field `DIBuilder::CreateSArrayType()` had DIBuilder return a value type instead of a reference type for FuncDeclaration. 

---
With these two simple fixes debugging LDC 1.x+Calypso built by LDC went from a painful experience into a smooth one, I've had no other crash/garbage value so far.

(BTW sorry to have put Calypso's development in slumber, it's back at full blast now, and the transition to DDMD isn't going too bad)

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -440,6 +440,8 @@ ldc::DIType ldc::DIBuilder::CreateCompositeType(Type *type) {
   // Use the actual type associated with the declaration, ignoring any
   // const/wrappers.
   LLType *T = DtoType(sd->type);
+  if (t->ty == Tclass)
+    T = llvm::cast<llvm::PointerType>(T)->getElementType();
   IrTypeAggr *ir = sd->type->ctype->isAggr();
   assert(ir);
 
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -276,7 +276,7 @@ ldc::DIType ldc::DIBuilder::CreateEnumType(Type *type) {
       getTypeAllocSize(T) * 8,               // size (bits)
       getABITypeAlign(T) * 8,                // align (bits)
       DBuilder.getOrCreateArray(subscripts), // subscripts
-      CreateTypeDescription(te->sym->memtype, false));
+      CreateTypeDescription(te->sym->memtype));
 }
 
 ldc::DIType ldc::DIBuilder::CreatePointerType(Type *type) {
@@ -296,7 +296,7 @@ ldc::DIType ldc::DIBuilder::CreatePointerType(Type *type) {
   const llvm::Optional<unsigned> DWARFAddressSpace = llvm::None;
 #endif
 
-  return DBuilder.createPointerType(CreateTypeDescription(nt, false),
+  return DBuilder.createPointerType(CreateTypeDescription(nt),
                                     getTypeAllocSize(T) * 8, // size (bits)
                                     getABITypeAlign(T) * 8,  // align (bits)
 #if LDC_LLVM_VER >= 500
@@ -323,7 +323,7 @@ ldc::DIType ldc::DIBuilder::CreateVectorType(Type *type) {
   return DBuilder.createVectorType(
       getTypeAllocSize(T) * 8,              // size (bits)
       getABITypeAlign(T) * 8,               // align (bits)
-      CreateTypeDescription(te, false),     // element type
+      CreateTypeDescription(te),            // element type
       DBuilder.getOrCreateArray(subscripts) // subscripts
       );
 }
@@ -381,7 +381,7 @@ ldc::DIType ldc::DIBuilder::CreateMemberType(unsigned linnum, Type *type,
   llvm::Type *T = DtoType(t);
 
   // find base type
-  ldc::DIType basetype = CreateTypeDescription(t, true);
+  ldc::DIType basetype = CreateTypeDescription(t);
 
   auto Flags = DIFlagZero;
   switch (prot) {
@@ -600,7 +600,7 @@ ldc::DIType ldc::DIBuilder::CreateSArrayType(Type *type) {
   return DBuilder.createArrayType(
       getTypeAllocSize(T) * 8,              // size (bits)
       getABITypeAlign(T) * 8,               // align (bits)
-      CreateTypeDescription(t, false),      // element type
+      CreateTypeDescription(t),             // element type
       DBuilder.getOrCreateArray(subscripts) // subscripts
       );
 }
@@ -618,7 +618,7 @@ ldc::DISubroutineType ldc::DIBuilder::CreateFunctionType(Type *type) {
   Type *retType = t->next;
 
   // Create "dummy" subroutine type for the return type
-  LLMetadata *params = {CreateTypeDescription(retType, true)};
+  LLMetadata *params = {CreateTypeDescription(retType)};
 #if LDC_LLVM_VER == 305
   auto paramsArray = DBuilder.getOrCreateArray(params);
 #else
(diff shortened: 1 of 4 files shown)


Syniurge:
Added a test case modeled after the CDB tests.

Is it okay to make it a `lit` test like the ones for CDB? The GDB output is slightly different for DMD or GDC-built executables, for instance when built by DMD 2.074 base classes do not appear while printing a derived class object, and type names are fully qualified.

> Did you look into the currently disabled GDB tests in dmd-testsuite? Maybe some of them are fixed by this.

I couldn't find one that involve classes, and isn't this the only test that gets disabled? :
```make
# LDC: even without optimizations "x optimized-out"
DISABLED_TESTS += gdb10311
```

JohanEngelen:
(I am a big fan of lit tests.)

kinke:
@Syniurge: Having issues with the old gdb used by Travis?

Syniurge:
@kinke Yep, the test works fine on my Ubuntu 12.04 debootstrap with GDB 7.4 so no idea why it's still failing. I added a cat so that Travis finally let me know what's going on.

kinke:
I don't know in which direction you are going, but if you have to weaken your tests to make them work with 7.4, I'd rather look into getting a newer gdb from some PPA or so instead.

Syniurge:
Actually GDB 7.4 works fine, the problem is when LDC is built inside an Ubuntu 12.04 environment, it gets the global variable address wrong:

```d
uv gvar;
static this() { gvar = new xyz(12, 34.56); }

int main()
{
    (...)
    xyz* ptr = cast(xyz*) &gvar;
    (...)
}
```
```gdb
(gdb) p 'classtypes_gdb.gvar'
Cannot access memory at address 0x7ffff864aa38
(gdb) p *ptr
$3 = (struct xyz *) 0x7ffff7eed000
(gdb) p **ptr
$4 = {<> = {<> = {<No data fields>}, i = 12}, f = 34.5600014, d = nan(0xc000000000000)}
```
Whereas `p 'classtypes_gdb.gvar'` works if LDC is built inside Ubuntu 16.04. Should I investigate further? This isn't really related to the issue I was trying to fix.

Syniurge:
Oh, it works if I make gvar a __gshared. I re-enabled the gvar checks, Travis should be green now.

kinke:
I think I managed to screw up the comment fix ;) - so gdb 7.4 is fine, but the host Ubuntu version makes the difference? Did your tests involve the same LLVM version across 12.04 and 16.04?

Syniurge:
It was LLVM 3.9.1 for both of them, http://releases.llvm.org/3.9.1/clang+llvm-3.9.1-x86_64-linux-gnu-ubuntu-16.04.tar.xz in 16.04 and http://releases.llvm.org/3.9.1/clang+llvm-3.9.1-x86_64-linux-gnu-debian8.tar.xz in 12.04:

```bash
# LDC built in 16.04
$ ../../build.llvm391.1604/bin/ldc2 -g -dwarf-version=2 classtypes_gdb.d

$ schroot -c precise_amd64
(precise_amd64)$ gdb --version
GNU gdb (Ubuntu/Linaro 7.4-2012.02-0ubuntu2) 7.4-2012.02

(precise_amd64)$ gdb classtypes_gdb --batch -x classtypes_gdb.gdb
Breakpoint 1 at 0x402232: file classtypes_gdb.d, line 37.
warning: no loadable sections found in added symbol-file system-supplied DSO at 0x7ffff7ffa000
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Breakpoint 1, D main () at classtypes_gdb.d:41
warning: Source file is more recent than executable.
41          return 0;
$1 = (struct xyz *) 0x7ffff7eed020
$2 = {<> = {<> = {<No data fields>}, i = 99}, f = 88.7699966, d = 624.351}
$3 = (struct xyz *) 0x7ffff7eed020
$4 = {<> = {<> = {<No data fields>}, i = 99}, f = 88.7699966, d = 624.351}
$5 = {0x0, 0x0, 0x7ffff7eed040, 0x0}
$6 = {<> = {<> = {<No data fields>}, i = 2}, f = 2, d = 0.98699999999999999}
$7 = (struct uv *) 0x7ffff7eed000
$8 = {<> = {<No data fields>}, i = 12}
[Inferior 1 (process 8905) exited normally]

# LDC built in 12.04
(precise_amd64)$ ../../build.llvm391.1204/bin/ldc2 -g -dwarf-version=2 classtypes_gdb.d
(precise_amd64)$ gdb classtypes_gdb --batch -x classtypes_gdb.gdb
Breakpoint 1 at 0x409902: file classtypes_gdb.d, line 37.
warning: no loadable sections found in added symbol-file system-supplied DSO at 0x7ffff7ffa000
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Breakpoint 1, D main () at classtypes_gdb.d:41
41          return 0;
$1 = (struct xyz *) 0x7ffff7eed020
$2 = {<> = {<> = {<No data fields>}, i = 99}, f = 88.7699966, d = 624.351}
$3 = (struct xyz *) 0x7ffff7eed020
$4 = {<> = {<> = {<No data fields>}, i = 99}, f = 88.7699966, d = 624.351}
$5 = {0x0, 0x0, 0x7ffff7eed040, 0x0}
$6 = {<> = {<> = {<No data fields>}, i = 2}, f = 2, d = 0.98699999999999999}
classtypes_gdb.gdb:9: Error in sourced command file:
Cannot access memory at address 0x7ffff861dd70
```
(btw for `p &'classtypes_gdb.gvar'` in the first executable I get `(struct uv **) 0x7ffff7fec718`)

kinke:
Thanks for the details. So you built the first (working) executable outside chroot. Does building it inside make a difference so that we can blame an older C runtime or so? ;)

Syniurge:
It's not ABI-compatible with libstdc++ from 12.04:
../../build.llvm391.1604/bin/ldc2: /lib/x86_64-linux-gnu/libtinfo.so.5: no version information available (required by ../../build.llvm391.1604/bin/ldc2)
../../build.llvm391.1604/bin/ldc2: relocation error: ../../build.llvm391.1604/bin/ldc2: symbol std::_V2::error_category::_M_message[abi:cxx11](int) const, version GLIBCXX_3.4.21 not defined in file libstdc++.so.6 with link time reference
```
the ubuntu-toolchain-r/test PPA didn't help, copying the ones from 16.04 or building libstdc++ from the gcc 5.4 source result in segfault, so I don't know what prevent it from running inside 12.04 but kinda gave up.

The cause of the incorrect TLS debug info isn't gcc itself either since the result is the same with clang.

**Comment by Syniurge** [comment]
(disregard my previous message)

It's the same result if I run LDC built in 16.04 from the 12.04 chroot. The cause is not gcc either since it's the same result with clang, and I had to build libstdc++ from source to make ldc2 from 16.04 run. So yes, it looks like it's one of these:

```bash
(precise_amd64)$ LD_LIBRARY_PATH="$(pwd)/../../build.llvm391.1604/bin" ldd ../../build.llvm391.1604/bin/ldc2
        linux-vdso.so.1 =>  (0x00007ffc7cfa5000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fede03fb000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fede01f3000)
        libtinfo.so.5 => /lib/x86_64-linux-gnu/libtinfo.so.5 (0x00007feddffcb000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007feddfdab000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007feddfaab000)
#not libstdc++.so.6 => ../../build.llvm391.1604/bin/libstdc++.so.6 (0x00007feddf72b000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007feddf513000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007feddf14b000)
        /lib64/ld-linux-x86-64.so.2 (0x0000560456a02000)
```

kinke:
Thanks a lot, especially for hanging in there wrt. the TLS issue. :)

Syniurge:
After having a quick look at how TLS variables are implemented it became easy to suspect the linker and indeed generating the .o file in 12.04 but doing the linking in 16.04 fixes it, so LDC is fine and blaming GDB in tests/d2/CMakeLists.txt is wrong:

```cmake
    if(GDB_VERSION VERSION_LESS "7.6.1")
        set(gdb_flags "NOTLS")
    else()
        set(gdb_flags "ON")
    endif()
```

PR #2133, opened 2017-05-23
tests/d2: use cmake variable instead of hardcoded make

nrTQgc:


nrTQgc:
FreeBSD and NetBSD has make and gmake. We should use gmake. How can I select gmake in this case?
By the way is there auto builds for LDC on FreeBSD?

Issue #2131, opened 2017-05-22
Suboptimal code for `bool |= bool`


JohanEngelen:
We create slightly bad code for `core.checkedint.mulu` calls.
Not sure whether this is an LLVM issue or bad codegen on our part. Our code makes sure that `overflow` is a 1 or a 0, regardless of the value that was passed in; which seems unnecessary.

See: https://godbolt.org/g/0jUhgs

```d
import core.checkedint;
alias T = ulong;
T foo(T x, T y, ref bool overflow)
{
   return mulu(x, y, overflow);
}
```

LDC: (`-O3 -enable-cross-module-inlining`)
```asm
ulong example.foo(ulong, ulong, ref bool):
        mov     rax, rdx
        mul     rsi
        seto    cl
        mov     dl, byte ptr [rdi]
        and     dl, 1
        or      dl, cl
        mov     byte ptr [rdi], dl
        ret
```

GDC: (`-O3`)
```asm
ulong example.foo(ulong, ulong, ref bool):
        mov     rax, rdi
        mov     rcx, rdx
        mul     rsi
        seto    dl
        or      BYTE PTR [rcx], dl
        ret
```

JohanEngelen:
Simpler test case:
```d
void foo(bool x, ref bool o)
{
  o |= x;
}
```

```asm
void example.foo(bool, ref bool):
        or      sil, byte ptr [rdi]
        and     sil, 1
        mov     byte ptr [rdi], sil
        ret
```

Unoptimized IR:
```llvm
define void @_D8overflow3fooFbKbZv(i8* dereferenceable(1) %o, i1 %x_arg) #0 {
  %x = alloca i8, align 1                         ; [#uses = 2, size/byte = 1]
  %1 = zext i1 %x_arg to i8                       ; [#uses = 1]
  store i8 %1, i8* %x
  %2 = load i8, i8* %x                            ; [#uses = 1]
  %3 = trunc i8 %2 to i1                          ; [#uses = 1]
  %4 = zext i1 %3 to i32                          ; [#uses = 1]
  %5 = load i8, i8* %o                            ; [#uses = 1]
  %6 = trunc i8 %5 to i1                          ; [#uses = 1]
  %7 = zext i1 %6 to i32                          ; [#uses = 1]
  %8 = or i32 %7, %4                              ; [#uses = 1]
  %9 = icmp ne i32 %8, 0                          ; [#uses = 1]
[...]

PR #2135, opened 2017-05-24
Add range metadata when loading bools

kinke:
When optimizing the code, analogous to clang, fixing one aspect of issue #2131.

### DIFF
--- a/gen/dvalue.cpp
+++ b/gen/dvalue.cpp
@@ -13,7 +13,9 @@
 #include "gen/llvm.h"
 #include "gen/llvmhelpers.h"
 #include "gen/logger.h"
+#include "gen/optimizer.h"
 #include "gen/tollvm.h"
+#include "llvm/IR/MDBuilder.h"
 
 namespace {
 bool isDefinedInFuncEntryBB(LLValue *v) {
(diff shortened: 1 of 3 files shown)


kinke:
Johan's seemingly simpler testcase `void foo(bool x, ref bool o) { o |= x; }` still doesn't work, i.e., `x` is still ANDed with 1 in optimized assembly; I guess we do need `zeroext` for bool params after all.

JohanEngelen:
For the testcase, I like `void foo(bool x, ref bool o) { o |= x; }` much better because it doesn't depend on any imported module or cross-module inlining.
(in my tests, zeroext didn't matter)

kinke:
`zeroext` is necessary. There's no explicit `and` anymore in the optimized IR already after the first commit, but there is in assembly unless `zeroext` is used in the 2nd commit. Switched to the simpler test case as that's working properly now too.

kinke:
> This can wait if you don't feel particularly inspired

Yep, not feeling too inspired about refactoring this tiny check for an active `-O` switch into something bigger. Maybe something for another time though (incl. caching of that metadata node?).

Good to go into 1.3?

JohanEngelen:
lgtm

PR #2142, opened 2017-05-26
Integrate LLD for MSVC targets; prepare for other targets

kinke:


kinke:
Alright, I just tested directly cross-compiling and -linking from Linux x64 to Win64. It works without much hassle.

1. Build this PR, using CMake option `-DLDC_WITH_LLD=ON`, with a recent LLVM incl. LLD (I used the current `release_40` branch for both LLVM and LLD on Ubuntu 16.04)
2. Copy over system libs and object files from a Visual C++ installation. I used Visual C++ 2015 x64 Update 3 with the latest Win10 SDK. I merged all .{lib,obj} files in most directories specified in the LIB environment variable (VS tools command prompt) into a single new directory, **renamed all those files to lowercase** (most Linux filesystems are case-sensitive) and packed them to an archive. Raw size nearly 0.5 GB, [.tar.xz](https://dl.dropboxusercontent.com/s/6js68fbccj9b0m5/libs_msvc14_x64.tar.xz?dl=0) about 32 MB.
3. Copy over LDC runtime files, e.g., from the latest Win64 [release package](https://github.com/ldc-developers/ldc/releases/) (all files in the `lib64` subdirectory).
4. Tweak the LDC config and add a section like this:
```
x86_64-pc-windows-msvc:
{
    // default switches injected before all explicit command-line switches
    switches = [
        "-I/home/martin/ldc/runtime/druntime/src",
        "-I/home/martin/ldc/runtime/profile-rt/d",
        "-I/home/martin/ldc/runtime/phobos",
        "-L-L/home/martin/Downloads/libs_msvc14_x64",
        "-L-L/home/martin/Downloads/ldc2-1.3.0-beta1-win64-msvc/lib64", 
        "-defaultlib=phobos2-ldc,druntime-ldc",
        "-debuglib=phobos2-ldc-debug,druntime-ldc-debug"
    ];
};
```

You'll then be able to produce Win64 executables and DLLs by simply using the `-mtriple=x86_64-pc-windows-msvc` switch. I produced a Phobos-hello-world and a dmd-testsuite `runnable\eh.d` executable, copied them over to my Windows box and they worked.

kinke:
This also means that if we are allowed to redistribute the MS libs, we could provide a self-sufficient package for Windows, similar to the DMD package, without relying on any Visual C++ installation (and well basically dependening on **nothing at all** except for Windows these days - no libconfig, no shared C runtime, no external linker & libs, ...). A self-sufficient Win64-only package (no 32-bit libs) would be a ~50 MB download, just to give you an idea.

kinke:
> Do you know what the licensing situation is on the MS libraries?

Nope, no idea.

> How big is the impact of LDC_WITH_LLD=ON?

Size-wise for the `ldc2` executable currently +7.5% with just the COFF output capability (Win64, `RelWithDebInfo`, LLVM & LLD 3.9.1). Mach-O and ELF still to come.
It's just a quick hack for now so that I don't need to check the LLVM version and LLD availability.

kinke:
I got rid of `LDC_WITH_LLD` as user-provided CMake option and try to detect LLVM 3.9+ incl. LLD headers & libs now.
AppVeyor produces LLD-enabled LDC builds and so does Travis for LLVM 4.0 (Linux at least), but LLD isn't tested as it's opt-in via `-internal-linker` for now (still experimental, no .pdb support etc.).

CMakeLists.txt
```diff
@@ -446,12 +446,23 @@ if(GENERATE_OFFTI)
...
 endif()
 
-option(RISCV_LLVM_DEV, "full RISC-V support with riscv-llvm")
-mark_as_advanced(RISCV_LLVM_DEV)
+#
+# LLD integration (requires LLVM >= 3.9 with LLD headers & libs)
+#
+set(LDC_WITH_LLD OFF)
+if(LDC_LLVM_VER GREATER 308)
+    if(EXISTS "${LLVM_INCLUDE_DIRS}/lld/Driver/Driver.h")
```
klickverbot: Maybe just use the standard CMake header search functionality? `${LLVM_INCLUDE_DIRS}/…` (with the plural) looks quite wrong.

driver/linker-msvc.cpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+    args.push_back(str);
+  };
+
+  for (const auto &str : opts::linkerSwitches) {
+    addSwitch(str);
+  }
+
+  for (unsigned i = 0; i < global.params.linkswitches->dim; i++) {
+    addSwitch(global.params.linkswitches->data[i]);
```
klickverbot: This is probably code that existed before, but you can use `(*global.params.linkswitches)[i]` or our range-based for wrapper.

CMakeLists.txt
kinke: `CHECK_INCLUDE_FILE()` didn't work (although there's a previous line with `include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})` - no extra include dirs appear to be used for the check).

CMakeLists.txt
```diff
@@ -546,6 +557,13 @@ add_custom_target(${LDMD_EXE} ALL DEPENDS ${LDMD_EXE_FULL})
 # Figure out how to link the main LDC executable, for which we need to take the
 # LLVM flags into account.
 set(LDC_LINKERFLAG_LIST "${SANITIZE_LDFLAGS};${LLVM_LIBRARIES};${LLVM_LDFLAGS}")
+if(LDC_WITH_LLD)
+    if(MSVC)
+        list(APPEND LDC_LINKERFLAG_LIST lldCOFF.lib lldCore.lib lldDriver.lib)
+    else()
+        set(LDC_LINKERFLAG_LIST "-llldCOFF;-llldCore;-llldDriver;${LDC_LINKERFLAG_LIST}")
```
kinke: Not that I've seen, in fact I didn't find anything about how to embed LLD via CMake. On Linux the order matters (even inbetween these 3 new libs), so I put them in front. As the LLD libs don't have the `LLVM` prefix, we can't specify them in `find_package()`...

CMakeLists.txt
klickverbot: Does CMAKE_REQUIRED_INCLUDES do the trick?

kinke:
Yesterday I tried integrating LLD for ELF and Mach-O targets. Sadly, there's an unknown number of conflicting command-line options then (which only show up after linking, when running LDC; additionally, only the first conflicting option is shown as I had to find out). I renamed the `-dwarf-version` one, but then `-march` popped up and I stopped, as I guess that wasn't all.

CMakeLists.txt
kinke: Yep it does.

driver/linker.cpp
```diff
@@ -20,19 +20,27 @@
...
 
-static llvm::cl::opt<bool>
+static llvm::cl::opt<llvm::cl::boolOrDefault>
     staticFlag("static", llvm::cl::ZeroOrMore,
                llvm::cl::desc("Create a statically linked binary, including "
                               "all system dependencies"));
 
+#if LDC_WITH_LLD
+static llvm::cl::opt<bool>
+    useInternalLinker("internal-linker", llvm::cl::ZeroOrMore, llvm::cl::Hidden,
```
kinke: Renamed.

CMakeLists.txt
kinke: Hmm it doesn't for Travis (LLVM 4.0). :/

kinke: gcc needs an additional `-std=c++11` flag; fixed now.

JohanEngelen:
(about cmdline options: I think it'd be best to someday go the Clang route: put all LLVM options behind a `-mllvm` switch)

JohanEngelen:
Very cool work btw!

kinke:
The problem is that LLD's [ELF/LTO.cpp](https://github.com/llvm-mirror/lld/blob/release_40/ELF/LTO.cpp#L19) drags in an [LLVM header](https://github.com/llvm-mirror/llvm/blob/master/include/llvm/CodeGen/CommandFlags.h) with quite a number of command-line options, quite a few colliding with our duplicates, but also quite a number of new options, at least some would also make sense for LDC.

The include has been removed in LLD master, but it's there for 3.9 and 4.0.

kinke:
I quickly looked into using that shared LLVM header ourselves (in a separate .cpp, so that it's only pulled in if it's not already via LLD), but what's required is a hack. The actually required shared utility functions (setting up a TargetMachine's TargetOptions according to all codegen command-line options etc. - very handy) are all marked as static, so one can't export them. LLD has its own little .cpp file including the LLVM header and augmenting it with their `external` wrapper. They use it to prevent conflicts inside LLD itself. ;)
Unfortunately, for us this means that we'd have to forward to these `external` LLD wrappers for `LDC_WITH_LLD`, and otherwise include the LLVM header and implement the wrappers ourselves.

CMakeLists.txt
kinke: Damn, I missed that the first Travis OSX job now uses the SPIRV LLVM which apparently doesn't feature LLD. So all good, LLD is dectected correctly for the other job.

Laeeth:
licensing
https://www.visualstudio.com/en-us/productinfo/2015-redistribution-vs#visual-c-runtime

or you could just include a little call to download and install vcredist* as a script in the repo.  (older versions usually require logging in with MS account, unfortunately, but curl should pick up latest one ).  I don't know if vcredist installer works on wine.

kinke:
We don't need to redistribute the DLLs (easy enough to install and only needed with non-default `-mscrtlib=msvcrt[d]`); it's about the static and import .lib files.

MartinNowak:
They do support static linking, so one would think that their license supported some form of redistribution.
But non of those lib files are in [Visual Studio 2015 Redistribution List](https://www.visualstudio.com/license-terms/mt228262/).

d-random-contributor:
They allow redistribution of libraries after they are run through linker :)

PR #2143, opened 2017-05-27
[dcomptue] semantic analysis & lit tests

thewilsonator:


thewilsonator:
The following lit tests fail on LLVM 5.0 (Circle and Travis OSX/SPIRV) and are unrelated
```
PGO/indirect_calls.d
linking/thinlto_asm_x86.d
linking/thinlto_1.d
linking/thinlto_modulecdtors.d
linking/thinlto_modulecdtors_2.d
```
And the following of the dmd test suite fail on Circle and are unrelated.
```
runnable/gdb4149.d
runnable/sdtor.d
```

Semaphore fails to install packages correctly.

nordlow:
Title is mis-spelled :)

gen/uda.h
```diff
@@ -28,12 +28,12 @@ void applyVarDeclUDAs(VarDeclaration *decl, llvm::GlobalVariable *gvar);
...
-enum class DComputeCompileFor
+/// Must match ldc.dcompute.Compilefor + 1 == DComputeCompileFor
+enum class DComputeCompileFor : int
 {
   hostOnly = 0,
   deviceOnly = 1,
   hostAndDevice = 2
 };
-DComputeCompileFor hasComputeAttr(Dsymbol *sym);
+extern "C" DComputeCompileFor hasComputeAttr(Dsymbol *sym);
```
JohanEngelen: why was `extern "C"` added?

thewilsonator: Because I couldn't figure out how to get `enum class` mangling to match across C++/D.

tests/codegen/dcompute_cu_addrspaces.d
```diff
@@ -0,0 +1,37 @@
+// REQUIRES: atleast_llvm309
+// REQUIRES: target_NVPTX
+// RUN: %ldc -c -mdcompute-targets=cuda-350 -m64 -output-ll -output-o %s && FileCheck %s --check-prefix=LL < kernels_cuda350_64.ll && FileCheck %s --check-prefix=PTX < kernels_cuda350_64.ptx
```
JohanEngelen: You should not use the default output filename here, because it will interfere with parallel execution of tests.
I see that the output filename cannot be specified, so that should be fixed urgently to allow separate parallel compilation of D files.

gen/uda.h
JohanEngelen: ah, enum class mangling is broken? I didn't know. :(   (did you find the dlang bug report for it?)

tests/codegen/dcompute_cu_addrspaces.d
thewilsonator: Wait, I thought it was only in parallel for a given file? Even so it should be fine for now as there are only two test files but I will 
They currently can't be specified because they follow a programmatic naming. If they were able to be specified we would need to expose the name (prefix?) to the D side (via traits or something) so that the correct file can be loaded. The paralleleness of the builds (other then lit tests) should not be a problem because you need to build all the files that interact with dcompute at once anyway.

I will get around to it (early July-ish).

gen/uda.h
thewilsonator: 1) I actually can't remember, I'll look into it.
2) no.

tests/codegen/dcompute_cu_addrspaces.d
JohanEngelen: > you need to build all the files that interact with dcompute at once anyway.

Really?
CUDA or OpenCL modules cannot be merged?

thewilsonator: You mean on disk or the tests?
The tests are separate because we could figure out how to have multiple independent requires clauses apply to different run directives.

On disk not directly, when you load them you can link them at runtime with the compute API in question. 

But the main reason for that is preservation of compile time information to drive the APIs. That starts to get very confusing, if not impossible, you're missing information. Also I dont know what happens if you have duplicate symbols when you link modules.

JohanEngelen: Note that the whole parallel build argument is just a minor argument. Having the compile generate files with hardcoded filenames is just bad. You can use e.g. `Module::objfile->name->str` as a start and suffix that with the cuda/opencl versions etc.

thewilsonator: Yes, but which objfile? If you have N `@compute` modules you do NOT get N "object files", you get one per compute API version specified on the command line. Having the output names hardcoded predictable means that the user need not specify where to look for the file.

JohanEngelen: do what `-singleobj` does

thewilsonator: That would choose whatever the first d file on the command line is which possibly change if you were to rebuild all of the dcompute only files on a project.
It also needs to be available to D for maximum streamlining, i.e. the user doesn't care what it is named, only that the library knows what it is and can handle all related aspects for them. Perhaps another command line switch for the prefix, defaulting to "kernels" or some other known prefix?

JohanEngelen: `-singleobj` also obeys the `-of` switch.  So `-of=foo.o` could mean that `foo-cuda350.ptx` etc. is generated.
But another cmdline option is also a possibility. 
Note that it's not just the hardcoded prefix, but also the (implicitly) hardcoded directory (current directory).

> It also needs to be available to D for maximum streamlining

I guess you want that so you can do `auto ptx = import("kernels...")`?
This is another instance where it is a pain that it is not possible to pass values to D through the command line (like `-D` for C++)....

PR #2146, opened 2017-06-02
Prefer C runtime over host D compiler for CTFE real_t math

kinke:
In order to simplify bootstrapping LDC on new platforms where ltsmaster's Phobos doesn't support the native `real` type yet, such as AArch64.

### DIFF
--- a/ddmd/root/ctfloat.d
+++ b/ddmd/root/ctfloat.d
@@ -8,7 +8,13 @@
 
 module ddmd.root.ctfloat;
 
-static import core.math, core.stdc.math;
+version(IN_LLVM)
+{
+    static import core.stdc.tgmath;
+    version(DigitalMars) static import core.math; // yl2x
+}
+else
+    static import core.math, core.stdc.math;
 import core.stdc.errno;
 import core.stdc.stdio;
 import core.stdc.stdlib;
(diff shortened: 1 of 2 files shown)


kinke:
I'm not 100% comfortable with this. E.g., LDC uses the C runtime for `std.math.sqrt()` (indirectly via LLVM intrinsics on x86 at least), but does an additional check for `x < 0 => NaN`. The exact semantics need to be checked before merging this.

kinke:
I checked; sqrt() was the only exception, all other CTFE builtins using host runtime's `{core,std}.math` were previously using LLVM intrinsics (with master's druntime/Phobos at least) which boil down to C runtime calls, so there should be no change in behavior after the amended sqrt() fix.

thewilsonator:
Is there any reason this hasn't been pulled yet? 
All CI failures have since been fixed. 
I'll close & reopen to retrigger CI make sure it works and then pull.

kinke:
> Is there any reason this hasn't been pulled yet?

As soon as David's AArch64 support makes it into `ltsmaster`, this PR is obsolete, at least until a new FP type pops up (not very likely, especially since we don't use the PowerPC doubledouble type as D `real`).

kinke:
This is now obsolete, as current master's `dmd.root.ctfloat` makes almost exclusive use of the host compiler's C runtime via `core.stdc.math`. `std.math` is only used for `rint()` and `nearbyint()`, both of which are LLVM intrinsics (for both master and ltsmaster).

PR #2147, opened 2017-06-03
Remove dependence on std::to_string, because it is not available on all systems.

JohanEngelen:
This PR should enable native compilation on Android.

See https://gist.github.com/joakim-noah/4153662c68fb2ba9592c9cb0d7892624

@joakim-noah

### DIFF
--- a/gen/inlineir.cpp
+++ b/gen/inlineir.cpp
@@ -9,6 +9,7 @@
 #include "gen/irstate.h"
 #include "gen/logger.h"
 #include "gen/tollvm.h"
+#include "gen/to_string.h"
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/AsmParser/Parser.h"
@@ -56,7 +57,7 @@ DValue *DtoInlineIRExpr(Loc &loc, FuncDeclaration *fdecl,
   // always inlined, this name does not escape the current compiled module; not
   // even at -O0.
   static size_t namecounter = 0;
-  std::string mangled_name = "inline.ir." + std::to_string(namecounter++);
+  std::string mangled_name = "inline.ir." + ldc::to_string(namecounter++);
   TemplateInstance *tinst = fdecl->parent->isTemplateInstance();
   assert(tinst);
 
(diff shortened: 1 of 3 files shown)


joakim-noah:
Oh, I was thinking about submitting something like this.  Although, according to that NDK issue thread you linked, it looks like this will come in Android NDK r16 later this year.  Up to you, patch looks good to me, if you think it's worthwhile in general, but don't care if we just wait till NDK r16 also.

JohanEngelen:
https://github.com/android-ndk/ndk/issues/82

JohanEngelen:
Let's get this in. The bug report already had the milestone changed from r13 to r14, and now from r15 to r16. Hope it helps in working towards removing the need for you to maintain a separate branch.

joakim-noah:
The main reason I'm building separately is having to patch llvm to use a form of emulated TLS that is easy to pass to the GC.  If you want to get Android support into mainline, that's the only remaining blocker.

llvm has a form of emulated TLS for ELF now, modeled on the way gcc does it, we need to move to that.  However, that TLS data has to be registered with the GC and I haven't wanted to peer into the low-level details of how to do that.  Johannes said they got it working with gdc, so you could perhaps look at what they did if you're interested.  It might not be so bad, it's just that my appetite for looking into that is low.

If your interest in that topic is higher and want to chip in with Android, that'd be a better way to get this branch into mainline.

ZombineDev:
`to_string` is not available on Android even with Clang 4.0 (native compilation). I had patch one of the llvm tools, in order to get llvm to build, so that's much appreciated @JohanEngelen!

joakim-noah:
Termux is switching to libc++ in preparation for the NDK doing so, termux/termux-packages@d946e67, they just pushed an update that switched a bunch of my installed packages over.  I've built llvm and ldc, both with and without this PR, against libc++, no problem.  If this PR is not considered more widely applicable, feel free to revert: it's not needed for native compilation of ldc on Android anymore.

ZombineDev:
That's good to hear @joakim-noah, thanks for the update. I hope to have more time in the coming weeks to test ldc's aarch64 support on Android.

joakim-noah:
There may not be much left, seeing all the pulls David got merged in #2153.  I hope to have access to some `AArch64` hardware in the coming weeks, should be able to try it myself then.

PR #2148, opened 2017-06-03
Use LLVM headers for shared codegen command-line options

kinke:
I.e., [llvm/CodeGen/CommandFlags.h](https://github.com/llvm-mirror/llvm/blob/release_40/include/llvm/CodeGen/CommandFlags.h) which in turn includes [llvm/MC/MCTargetOptionsCommandFlags.h](https://github.com/llvm-mirror/llvm/blob/release_40/include/llvm/MC/MCTargetOptionsCommandFlags.h).

This gets rid of a few duplicates on our side and includes about 35 (depending on LLVM version) new command-line options. LLVM provides a helper function to set up the TargetOptions according to (most of) these options.
Newer LLVM versions may add new options and we'll automatically inherit them, including setting up the TargetOptions accordingly.

I did my best (TM) to remove a few unused/undesirable options and hide all remaining new ones except for `-fp-contract`. The lists will need to be tweaked from time to time.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -134,10 +134,6 @@ static cl::opt<ubyte, true> debugInfo(
         clEnumValN(3, "gline-tables-only", "Add line tables only")),
     cl::location(global.params.symdebug), cl::init(0));
 
-static cl::opt<unsigned, true>
-    dwarfVersion("dwarf-version", cl::desc("Dwarf version"), cl::ZeroOrMore,
-                 cl::location(global.params.dwarfVersion), cl::Hidden);
-
 cl::opt<bool> noAsm("noasm", cl::desc("Disallow use of inline assembler"),
                     cl::ZeroOrMore);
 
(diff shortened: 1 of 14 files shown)


kinke:
The plan is to later copy the LLD wrappers into this .cpp module and hoping that the linker will use these ones for LLD instead of the LLD object file (with conflicting command-line options). This approach should be scalable for other LLVM components including these shared headers.

driver/cl_options.cpp
```diff
@@ -595,6 +550,16 @@ void hideLLVMOptions() {
...
 
+      "thread-model", "exception-model", "enable-fp-mad",
+      "enable-unsafe-fp-math", "enable-no-infs-fp-math",
+      "enable-no-nans-fp-math", "enable-no-trapping-fp-math",
+      "denormal-fp-math", "soft-float", "recip", "nozero-initialized-in-bss",
+      "tailcallopt", "stack-symbol-ordering", "stack-alignment", "enable-pie",
+      "use-ctors", "use-init-array", "emulated-tls", "unique-section-names",
+      "jump-table-type", "meabi", "debugger-tune", "asm-instrumentation",
+      "mc-relax-all", "incremental-linker-compatible", "asm-show-inst", "fcfi",
+      "cfi-type", "cfi-enforcing", "cfi-func-name", "pie-copy-relocations",
```
kinke: That's the most interesting part, the new hidden options, incl. some fine-grained FP math options.

kinke:
Good to go now?

kinke:
Dammit, linker flags order issue on Linux (LLD libs must come after our objects...).

JohanEngelen:
It's a big change, and I did not do a very detailed review. It looks good to me.
I hope the semantics of all options stayed the same and that our tests have enough coverage for this.

PR #2149, opened 2017-06-03
Replace `-sanitize` with `-fsanitize` and add `-fsanitize-coverage`.

JohanEngelen:
This reworks the sanitizer interface to be identical with Clang's.
`-fsanitize-coverage` is only available from LLVM 4.0.

What's left to do is pass the correct runtime libraries to the linker: I'll update the cmake script to check whether the runtime sanitizer libs of compiler-rt are available somewhere. If so, copy them to LDC. (this should also remove the need to have a copy of profile-rt inside our repo)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,6 +334,7 @@ file(GLOB IR_HDR ir/*.h)
 set(DRV_SRC
     driver/cache.cpp
     driver/cl_options.cpp
+    driver/cl_options_sanitizers.cpp
     driver/codegenerator.cpp
     driver/configfile.cpp
     driver/dcomputecodegenerator.cpp
@@ -352,6 +353,7 @@ set(DRV_HDR
     driver/cache.h
     driver/cache_pruning.h
     driver/cl_options.h
+    driver/cl_options_sanitizers.h
     driver/codegenerator.h
     driver/configfile.h
     driver/dcomputecodegenerator.h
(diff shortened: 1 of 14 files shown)


kinke:
> this should also remove the need to have a copy of profile-rt inside our repo

Highly appreciated!

JohanEngelen:
> How would this integrate with shipping PGO-enabled LDC in distros?

When building LDC, you'd also have to build LLVM compiler-rt. It's easy to build from source, I don't know if there are pre-built dev packages available. I was a bit optimistic about "I'll update the cmake script to...", it will take a while. We need a longer discussion on how to integrate compiler-rt things; perhaps adding it as submodule is a workable solution... (https://github.com/llvm-mirror/compiler-rt)

I first want to get AddresssSanitizer working with our dynamic memory management (right now it only works for stack allocated memory).

With this PR, simple fuzzing tests work. The only real addition is the coverage sanitizer. You have to manually build libFuzzer from LLVM, and the other sanitizer libs if you want to use those.

JohanEngelen:
> perhaps adding it as submodule is a workable solution...

Err... So that would need a submodule for each LLVM version that we support.......

JohanEngelen:
Good to squash+merge?

kalev:
As for the distro question, we have compiler-rt in Fedora so it should be perfectly fine from our point of view, no need to do submodules.

PR #2156, opened 2017-06-05
Add indirect call promotion pass when PGO info is available.

JohanEngelen:
This pass was apparently implicit in LLVM < 5.0. Fixes test PGO/indirect_calls.d with LLVM 5.0.


thewilsonator:
Looks good, travis is all green. Yay!

PR #2158, opened 2017-06-06
Fix lto

thewilsonator:
LLVM asserts that the `ProfileSummaryInfo*` passed to `llvm:: buildModuleSummaryIndex` is not null.


JohanEngelen:
great, thanks.

thewilsonator:
That took an embarrassingly long time to debug, should have just rebuilt LLVM with assertions, would have caught it immediately. 

Could you have a look at the `PGO/indirect_calls.d` test on 5.0? Its the only one left making the SPIR-V tester not green.

Issue #2108, opened 2017-05-11
Global variable initializer type does not match global variable type


MrSmith33:
```
Global variable initializer type does not match global variable type!
%voxelman.graphics.font.textmesher.TextMesherParams* @_D8voxelman8graphics4font10textmesher16TextMesherParams6__initZ
LLVM ERROR: Broken module found, compilation aborted!
ldc2 failed with exit code 1.
```

Fails on 1.2.0 and 1.3.0-beta1 on Windows
Upd: And 1.1.1

P.S. Can we have debug release?

ennis:
The following code seems to trigger the same error:

	import std.stdio;

	struct Vec
	{
	    float v; 
	}

	struct Quat
	{
		union
		{
			Vec v;   // replacing this line with 'float v;' makes the error disappear
			struct
			{
				float x;
			}
		}

		@nogc static Quat identity() pure nothrow @property
		{
			Quat q;
			q.x = 1;  // commenting this line makes the error disappear
			return q;
		}
	}

	struct S 
	{
		Quat q = Quat.identity;
	}

	static Quat q = Quat.identity;

	void main()
	{
	}

Compiler output:

	Global variable initializer type does not match global variable type!
	%app.S* @_D3app1S6__initZ
	LLVM ERROR: Broken module found, compilation aborted!


Output of `ldc2 -version`:

	LDC - the LLVM D compiler (1.3.0-beta1):
	based on DMD v2.073.2 and LLVM 3.9.1
	built with LDC - the LLVM D compiler (1.3.0-beta1)
	Default target: x86_64-pc-windows-msvc
	Host CPU: haswell

ennis:
Reduced with dustmite:

	struct Vec
	{
	}

	struct Quat
	{
		union
		{
			Vec v;
			struct
			{
				float x;
			}
		}

		static identity() 
		{
			Quat q;
			q.x = 1;
			return q;
		}
	}

	struct S 
	{
		Quat q = Quat.identity;
	}

Also fails on 1.3.0-beta2

kinke:
Thank you very much @ennis.

kinke:
```D
struct Vec { float v; }

struct Quat
{
	union
	{
		Vec v;
		struct { float x; }
	}

	@nogc static Quat identity() pure nothrow @property
	{
		Quat q;
		q.x = 1; // q.v.v = 1;
		return q;
	}
}

struct S 
{
	Quat q = Quat.identity;
}
````

[...]

PR #2171, opened 2017-06-17
Fix issue #2108 (triggered by T.init initializing dominated members in nested unions)

kinke:


driver/codegenerator.cpp
```diff
@@ -216,6 +216,10 @@ void CodeGenerator::finishLLModule(Module *m) {
 }
 
 void CodeGenerator::writeAndFreeLLModule(const char *filename) {
+  // Issue #1829: make sure all replaced global variables are replaced
+  // everywhere.
+  ir_->replaceGlobals();
```
klickverbot: Hmm, this leaves me with a queasy feeling – so we still build up a schizophrenic IR tree, just not with freed memory anymore?

kinke: Yep, although most occurrences should be replaced already beforehand (when replacing the global); this 2nd replacement pass should only cover some (cached) leftovers due to later inlining etc. (wasn't able to produce a testcase yet which would crash before).

The nicer thing would be replacing `IrGlobal::value` by `IrGlobal::getValue()` which could look it up in the current `gIR` module, but that's not feasible due to it being inherited from `IrVar`, and `IrVar` and `IrGlobal` being used in a front-end union (and `IrGlobal` apparently quite often simply used as `IrVar`), so the cached `value` field is currently required.

So the feasible options to me were either registering each `IrGlobal` with its mangled name in `gIR` and replacing their `value`s when swapping out the global, or doing a second replacement pass right before finalizing the LL module to make sure we really replace all occurrences. I went for the latter as it's both easier and safer.

gen/declarations.cpp
```diff
@@ -266,45 +270,26 @@ class CodegenVisitor : public Visitor {
...
-          gvar->replaceAllUsesWith(newValue);
-
-          gvar->eraseFromParent();
-          gvar = newGvar;
-          irGlobal->value = newGvar;
-        }
+        setLinkage(lwc, gvar);
 
-        assert(!gvar->hasDLLImportStorageClass());
+        // In case of type mismatch, swap out the variable.
```
klickverbot: This is now out of sync with the below `// Now, set the initializer.` comment.

kinke: Fixed.

kinke:
Will merge in the new few days if there's no more feedback.

gen/irstate.h
```diff
@@ -199,6 +204,18 @@ struct IRState {
...
+  // If the types don't match, this entails creating a new helper global
+  // matching the initializer type and replacing all existing uses of globalVar
+  // by a bitcast pointer to the helper global's payload.
+  // Returns either the specified globalVar if the types match, or the bitcast
+  // pointer replacing globalVar (and resets globalVar to the new helper
+  // global).
+  llvm::Constant *setGlobalVarInitializer(llvm::GlobalVariable *&globalVar,
+                                          llvm::Constant *initializer);
+
+  void replaceGlobals();
```
klickverbot: A short comment here would be nice as well.

kinke: Amended and added my longer response above as commit message.

PR #2172, opened 2017-06-18
Switch Android onto the sectionELF style of module registry, but leave out the unused register_dso and friends

joakim-noah:
dmd moved from [the linked list at `_Dmodule_ref` to a bracketed section a couple years ago](https://github.com/dlang/druntime/commit/e91f59d2175c1056af5d2de59f4fd5dee7a1779d), so I've been reverting that druntime commit for ldc/Android till now.  Turns out I can use this `sectionELF` code just fine, as long as I take out the calls to `_d_dso_registry` and modify druntime to use the slightly different ldc symbols `__start___minfo` and `__stop___minfo`, druntime PR forthcoming.  Confirmed by cross-compiling the stdlib tests from linux/x64 to Android/ARM with ldc master, and running the stdlib and dmd testsuite natively with master and merge-2.074.

I didn't look very closely at `sectionELF` however, just hacked this in, so if David thinks anything else might cause issues, let me know.

### DIFF
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -120,8 +120,7 @@ RegistryStyle getModuleRegistryStyle() {
     return RegistryStyle::sectionDarwin;
   }
 
-  if ((t->isOSLinux() && t->getEnvironment() != llvm::Triple::Android) ||
-      t->isOSFreeBSD() ||
+  if (t->isOSLinux() || t->isOSFreeBSD() ||
 #if LDC_LLVM_VER > 305
       t->isOSNetBSD() || t->isOSOpenBSD() || t->isOSDragonFly()
 #else
@@ -376,7 +375,10 @@ void emitModuleRefToSection(RegistryStyle style, std::string moduleMangle,
                                                                : "__minfo");
   gIR->usedArray.push_back(thismref);
 
-  if (!isFirst) {
+  // Android doesn't need register_dso and friends- see rt.sections_android-
+  // so bail out here.
+  if (!isFirst ||
+      global.params.targetTriple->getEnvironment() == llvm::Triple::Android) {
     // Nothing left to do.
     return;
   }


gen/modules.cpp
```diff
@@ -463,7 +462,8 @@ void emitModuleRefToSection(RegistryStyle style, std::string moduleMangle,
     const auto bb = llvm::BasicBlock::Create(gIR->context(), "", dsoCtor);
     IRBuilder<> b{bb};
     LLValue *params[] = {b.getFalse(), minfoRefPtr};
-    b.CreateCall(registerDSO, params);
+    if (global.params.targetTriple->getEnvironment() != llvm::Triple::Android)
+      b.CreateCall(registerDSO, params);
```
klickverbot: Why would any of the code be needed in the executable? This still emits the registerDSO functions and ctors/dtors that do nothing.

joakim-noah: I thought about removing the call to `buildRegisterDSO` also, but figured I'd keep this patch minimal.  As for the ctor/dtor, I initially simply switched away from the linked list of module info and built it, saw the linker complain that it couldn't find `_d_dso_registry`, then took out the two calls that added it.  It took a couple minutes, as I wasn't about to dive into all the complexity of everything else you do here.

If you think I should just remove this entire last block instead on Android, I can do that.

joakim-noah:
`--gc-sections` always worked fine before and after this patch, but I make sure the `ld.bfd` linker is being used, because `gold` would reorder the TLS sections and I don't know of an easy way to make it keep those sections together, as the default `ld.bfd` linker script does.  If the issue is mostly with `gold`, that doesn't matter to me right now.

gen/modules.cpp
klickverbot: What I'm asking for is to just not call any of the stuff that actually generates executable code. This should amount to always taking the `!isFirst` early return above.

joakim-noah:
I'm unclear exactly how `--gc-sections` works, I'm just reporting what happens, ie the module info is always there, as the test runner uses it to run the unit tests.  I now see that the `__minfo` section is currently left in the final executable, while the `register_dso` section is generated for an object file compiled from a small D source file but not in the final executable, presumably removed by the linker.

I've not looked into why it keeps some sections and throws out others, let me know if you want me to investigate, if there's something I should look for.  I had the impression `gold` was much more aggressive at removing sections than `ld.bfd`, maybe that explains it.

gen/modules.cpp
joakim-noah: Doesn't seem like that'll work as [I need the `__start___minfo`/`__stop___minfo` symbols in druntime](https://github.com/dlang/druntime/commit/e91f59d2175c1056af5d2de59f4fd5dee7a1779d#diff-a4fd3d910955daa992bdd4b0e3400425R66), as mentioned above.  I'll investigate.

klickverbot: __start_minfo/__stop_minfo are generated by the linker. The code here merely uses them (to achieve something along the lines of what you are doing in druntime).

joakim-noah: OK, if you say so. :) I don't want to try and understand everything you did here, hence simply reverting that druntime commit on Android for so long.

I just tried exiting early in `emitModuleRefToSection` where you indicated and the druntime test runner builds and all modules pass, same with a small sample file.  I'll try the phobos test runner and dmd testsuite next and update this PR in the way you suggested once everything passes.

joakim-noah: Passes all the tests when done the way you wanted, so updated the PR.

joakim-noah:
Submitted the druntime fixes upstream, dlang/druntime#1851, would be nice to get these two pulls in before the final 1.3 release, so I can reduce the patches needed for my ldc Termux package for Android.

joakim-noah:
Thanks for the review, added comment and merged.  If you approve the upstream druntime pull linked above, I'll go ahead and merge it in the ldc branch.

Issue #2168, opened 2017-06-16
Performance regressions in LDC 1.3.0-beta[1|2]


jondegenhardt:
I'm seeing performance regressions in several of my standard benchmarks (https://github.com/eBay/tsv-utils-dlang/blob/master/docs/Performance.md). Specifically the `tsv-filter`, `tsv-summarize`, and `csv2tsv` tests. Performance is off 20-30% from previous LDC releases.

So far I've only tested OSX (Xcode 8.3.3).

I won't have time to diagnose further for several days, but I'll see if I can narrow this further sometime during the week of June 19. I don't think it's likely to be due to changes in Phobos, but I can check that and if it is specific to OSX. If a fix is created for issue #2161 (performance degradation with boundscheck=off) I can check that as well. Testing with this option removed did show improvement on one benchmark, but overall benchmarks were worse.

kinke:
According to your benchmark page, the numbers have been obtained using LDC 1.1, so have you compared 1.1 to the 1.3 betas, skipping 1.2? I'm asking since we switched to LLVM 4.0 (on non-Windows) starting with 1.2 and the regressions are most likely due to that, although we have only heard from general performance improvements (even > 10%) with LLVM 4.0 until now.

jondegenhardt:
It's possible that I skipped 1.2, don't remember off-hand. I'll test with 1.2 and update this report as soon as I can. It may be a couple days, I have some schedule conflicts as the moment.

I normally test every LDC release, though without updating the benchmarks page. And I've been using 1.2 since it came out, so it's likely I tested it also. However, I don't have any record of it.

JohanEngelen:
@jondegenhardt just a quick note: those numbers are kindof cool to report in a blog post somewhere :)  Did you also try with `-flto=thin` ?

JohanEngelen:
Can you add the LLVM version to that table? Thanks.

jondegenhardt:
@JohanEngelen Thanks! Thought the numbers might be of general interest. I did not try `-flto=thin`. I'll add the LLVM version to the table later today, can't right now.

kinke:
I would have liked this to correlate with #2161 of course, but sadly it doesn't. It may be worth trying to recompile with LDC 1.3 and 2.072.2 druntime/Phobos (hopefully compatible) and see whether we have to blame the libs, the compiler or both. ;)

kinke:
[...]

Issue #2161, opened 2017-06-10
Performance decrease with `-boundscheck=off`


JohanEngelen:
The following code performs (a lot) worse with `-boundscheck=off`  (`ldc2 -release -O3`):

```d
module cmpBench;

import std.algorithm;
import std.range;
import std.stdio;
import std.traits;

void insertionSort(Range, Less)(Range r, Less l)
if (hasLength!Range && isRandomAccessRange!Range && hasSlicing!Range)
{
   foreach (immutable i; 1 .. r.length)
   {
      bringToFront(
            r[0 .. i].assumeSorted!((a,b) => l(a, b)).upperBound(r[i]),
            r[i .. i + 1]);
   }
}

struct Pair(T, U = T)
{
   T f;
   U s;

   this(const T a, const U b)
   {
      f = a;
      s = b;
   }

   bool opEquals(ref const Pair r) const
   {
      return f == r.f && s == r.s;
   }
}

alias IntPair = Pair!int;
alias IntPairPair = Pair!IntPair;

bool lt(int l, int r)
{
   return l < r;
}

bool lt(T, U)(ref const Pair!(T, U) l, ref const Pair!(T, U) r)
{
   if (l.f != r.f) return lt(l.f, r.f);
   return lt(l.s, r.s);
}

struct La(T)
{
   bool opCall(ref const T l, ref const T r) const
   {
      return lt(l, r);
   }
}


struct Ld(T)
{
   bool opCall(ref const T l, ref const T r) const
   {
      return lt(r, l);
   }
}

void main()
{
   enum N = 10;
   enum S = 1000;

   IntPairPair[] v;
   v.reserve(24 * S);
   for (auto i = 0; i < S; i++)
   {
      v ~= IntPairPair(IntPair(i + 0, i + 1), IntPair(i + 2, i + 3));
      v ~= IntPairPair(IntPair(i + 0, i + 1), IntPair(i + 3, i + 2));
      v ~= IntPairPair(IntPair(i + 0, i + 2), IntPair(i + 1, i + 3));
      v ~= IntPairPair(IntPair(i + 0, i + 2), IntPair(i + 3, i + 1));
      v ~= IntPairPair(IntPair(i + 0, i + 3), IntPair(i + 1, i + 2));
      v ~= IntPairPair(IntPair(i + 0, i + 3), IntPair(i + 2, i + 1));

   }

   La!IntPairPair la;
   Ld!IntPairPair ld;
[...]

PR #2180, opened 2017-06-24
Make inlining threshold customizable via (existing) -inline-threshold

kinke:
By letting LLVM set up the inlining params (taking into account `-inline-threshold`) instead of hardcoding them.

The old hardcoded thresholds were identical to the LLVM 3.9 defaults, but they changed in LLVM 4.0:

* optLevel > 2: 275 => 250
* -Os: 75 => 50
* -Oz: 25 => 5

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -25,6 +25,9 @@
 #include "llvm/IR/Module.h"
 #include "llvm/IR/DataLayout.h"
 #include "llvm/ADT/Triple.h"
+#if LDC_LLVM_VER >= 400
+#include "llvm/Analysis/InlineCost.h"
+#endif
 #if LDC_LLVM_VER >= 307
 #include "llvm/Analysis/TargetTransformInfo.h"
 #endif
@@ -269,16 +272,12 @@ static void addOptimizationPasses(PassManagerBase &mpm,
 #endif
 
   if (willInline()) {
-    unsigned threshold = 225;
-    if (sizeLevel == 1) { // -Os
-      threshold = 75;
-    } else if (sizeLevel == 2) { // -Oz
-      threshold = 25;
-    }
-    if (optLevel > 2) {
-      threshold = 275;
-    }
-    builder.Inliner = createFunctionInliningPass(threshold);
+#if LDC_LLVM_VER >= 400
+    auto params = llvm::getInlineParams(optLevel, sizeLevel);
+    builder.Inliner = createFunctionInliningPass(params);
+#else
+    builder.Inliner = createFunctionInliningPass(optLevel, sizeLevel);
+#endif
   } else {
 #if LDC_LLVM_VER >= 400
     builder.Inliner = createAlwaysInlinerLegacyPass();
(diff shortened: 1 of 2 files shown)


kinke:
This (unexpectedly) fixes the performance regression #2161 for LLVM 4.0 for me on Linux x64. I compared the `-O3` IR for #2161, saw that the issue was definitely an inlining one, then looked at how we set up the according pass, noted the new thresholds in LLVM 4.0 source and just wanted to play around with `-inline-threshold` to get it to inline almost everything again. Instead, just recompiling the testcase was enough to obtain identically performing `-boundscheck=off|safeonly` executables again.

JohanEngelen:
Can you add a test for this?
Something that checks whether a function is/isn't inlined?  (we should add tests that check that passes are run or not, so that we detect changes in LLVM's passes like this; similar to how we found out that indirect-call-promotion was no longer happening)

JohanEngelen:
Very nice that you found the bug and fixed it btw!

kinke:
A lit test based on the testcase is certainly possible. I hope it's just not going to be too brittle. The problem seemed to be that too much inlining was going on already in deeper call levels due to the omitted bounds checks, and that seemed to be problematic for the outer calls in `main()` (due to iterative inlining I assume).

I'm slightly optimistic that this also affects #2168, as those may also be inlining issues caused by Phobos restructurings in 2.073.

JohanEngelen:
> A lit test based on the testcase is certainly possible. I hope it's just not going to be too brittle.

Yeah.. :/  
With all the restructuring and whatnot going on in LLVM, I'm just worried that our pass pipeline will break more often.

kinke:
Test case amended, which fails unfortunately for LLVM 3.8 and possibly 3.7, but works for the others.

kinke:
This also 'fixes' the `-O3` 32-bit std.file unittest issues for SemaphoreCI.

JohanEngelen:
( related to which passes are run: you can use `-debug-pass=Structure` for this: https://godbolt.org/g/MAJ76d .  The "Unnamed pass: implement Pass::getPassName()" is perhaps from one of the passes we added. :-) )

JohanEngelen:
LGTM

JohanEngelen:
The added test fails for LLVM 6.0 (OSX).
One of the instances of insertionSort has all calls inlined, but the other doesn't (neither with nor without boundscheck=off). The call that is not inlined is to `_D3std9algorithm8mutation__T16bringToFrontImplTSQBu5range__T11SortedRangeTAS6gh2161__T4PairTSQr__TQmTiTiZQtTQqZQzSQBm__T13insertionSortTQCkTSQCn__T2LaTQCyZQiZQBmFQDkQzZ9__lambda3ZQEpTQEfZQGdFNaNbNiNfQFwQEyZm` which is quite a big function so it is definitely not clear whether inlining is profitable. 

As @kinke feared, the test is probably too brittle. Do we want to keep it? Or change it to something else that tests the pipeline more directly?

thewilsonator:
Let's see what happens for a new LLVM for dcompute.

PR #2188, opened 2017-07-02
AddressSanitizer, Darwin: add support for linking with correct library when found in LDC's lib dir.

JohanEngelen:
This PR is only meant to change Darwin link line with -fsanitize=address. With this PR, I can use addresssanitizer successfully (and easily) on my Mac. 
The PR finds a `libldc_rt.asan_osx_dynamic.dylib` or `libclang_rt.asan_osx_dynamic.dylib` in LDC's lib dir, and falls back to the old behavior (passing `-fsanitize=address` to the linker driver) if nothing is found. So now the user can copy his self-built library to LDC's lib dir and be happy :-)
(the cmake script could copy+rename the library to the lib dir when the library is found in the LLVM binaries)

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -17,7 +17,7 @@
 #include "gen/optimizer.h"
 #include "llvm/ProfileData/InstrProf.h"
 #include "llvm/Support/FileSystem.h"
-#include <llvm/Support/Path.h>
+#include "llvm/Support/Path.h"
 #include "llvm/Target/TargetMachine.h"
 #include "llvm/Target/TargetOptions.h"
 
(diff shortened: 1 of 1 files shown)


JohanEngelen:
(the linux link improvement will probably come soon after)

JohanEngelen:
OK to merge?

driver/linker-gcc.cpp
```diff
@@ -17,6 +17,7 @@
 #include "gen/optimizer.h"
 #include "llvm/ProfileData/InstrProf.h"
 #include "llvm/Support/FileSystem.h"
+#include <llvm/Support/Path.h>
```
kinke: Please use quotes instead of brackets. Otherwise lgtm.

JohanEngelen: wow, good catch. a bad mistake

JohanEngelen: probably a copy-paste from exe_path.cpp :(

PR #2191, opened 2017-07-03
Additional -vv logging in DtoCreateNestedContext.

JohanEngelen:
Helps with tracking down the why of GC allocations.


thewilsonator:
LGTM.

PR #2193, opened 2017-07-06
Fix search of GCC used for MSP430 external assembly

luismarques:


driver/tool.cpp
```diff
@@ -65,7 +65,20 @@ std::string getProgram(const char *name, const llvm::cl::opt<std::string> *opt,
...
+std::string getGcc(const llvm::Triple *triple) {
+  if(triple != nullptr && triple->getArch() == llvm::Triple::msp430) {
+    std::string path = findProgramByName("msp430-gcc");
+
+    if(path.empty()) {
+      path = findProgramByName("msp430-elf-gcc");
+    }
+
+    assert(!path.empty() &&
+           "Could not find MSP430 GCC cross-compiler for external assembly");
```
thewilsonator: I'd change the `assert` to 
```
if (path.empty()) {
  error(Loc(), "Could not find MSP430 GCC cross-compiler for external assembly");
  fatal();
}
```
but otherwise looks good.

kinke:
I'd prefer letting the user set it via the `-gcc` switch. It can even be inserted in the `ldc2.conf` config file for the msp430 triple.

driver/tool.cpp
```diff
@@ -65,7 +65,22 @@ std::string getProgram(const char *name, const llvm::cl::opt<std::string> *opt,
 
 ////////////////////////////////////////////////////////////////////////////////
 
-std::string getGcc() {
+std::string getGcc(const llvm::Triple *triple) {
```
ZombineDev: Perhaps this could be generalized to:
(in pseudo-code)
```D
if (gcc_not_set_by_user && target != host)
{
    string path;
    auto allPossibleSuffixes = powerset(target.vendor, target.os, target.abi);
    foreach (suffixes; allPossibleSuffixes )
    {
        auto suffix = suffixes.join('-')
        if (suffix.length) suffix = '-' ~ suffix;
        if (auto p = findProgramByName("${arch}${suffix}-gcc"))
        { path = p; break; }
    }

    if (path.empty)
        error();
    else
        return path;
}
```

JohanEngelen: What's this strange language you are programming in??

(;-)  the birth of `driver/tool.d` ?)

ZombineDev: ;)

Issue #2189, opened 2017-07-02
Semaphore CI: Use custom LLVM build with enabled assertions and LTO plugin


kinke:
As suggested in #2074. And don't forget LLD.
Semaphore is based on Ubuntu 14.04.

kinke:
I used https://github.com/ldc-developers/ldc-scripts/pull/5 to build LLVM 4.0.1 with enabled assertions and use it for SemaphoreCI now, which already fails due to assertions, most if not all apparently caused by debuginfos: https://semaphoreci.com/ldc-developers/ldc/branches/master/builds/81

JohanEngelen:
Oh wow, this is pretty bad :( great that you found it.

kinke:
Assertions fixed; LTO plugin now included (https://github.com/ldc-developers/ldc-scripts/pull/5/commits/7f6f490fb59e0c7bb106335c387889fb56fc26ea) and working.

PR #2194, opened 2017-07-06
Add support for MSP430 / 16-bit targets

luismarques:
`global.params` doesn't have a flag for the condition of being a 16-bit target. We have two options: 1) add that flag, or 2) compute 16-bitness from the target triple each time `DtoSize_t` is called. Since adding fields to `struct Params` might be disruptive, I opted for the latter. Please say if you prefer the former.

I avoid calling `isArch16Bit` unless the target is not 64-bit, which should be the most common target, and therefore avoid unnecessary work. Having a flag for 16-bitness would be slightly cheapter, but I'm guessing the speed penalty is not noticeable.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -818,13 +818,10 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("BigEndian");
   }
 
-  // a generic 64bit version
+  // Set versions for arch bitwidth
   if (global.params.isLP64) {
     VersionCondition::addPredefinedGlobalIdent("D_LP64");
-  }
-
-  // 16 bit version for 16 bit targets
-  if(arch == llvm::Triple::msp430) {
+  } else if (global.params.targetTriple->isArch16Bit()) {
     VersionCondition::addPredefinedGlobalIdent("D_P16");
   }
 
@@ -909,7 +906,7 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("Posix");
     break;
   case llvm::Triple::UnknownOS:
-    if(arch == llvm::Triple::msp430)
+    if (arch == llvm::Triple::msp430)
       break;
     // fallthrough
   default:
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -767,6 +767,9 @@ void registerPredefinedTargetVersions() {
     registerPredefinedFloatABI("MIPS_SoftFloat", "MIPS_HardFloat");
     registerMipsABI();
     break;
+  case llvm::Triple::msp430:
+    VersionCondition::addPredefinedGlobalIdent("MSP430");
+    break;
 #if defined RISCV_LLVM_DEV || LDC_LLVM_VER >= 400
 #if defined RISCV_LLVM_DEV
   case llvm::Triple::riscv:
@@ -818,6 +821,11 @@ void registerPredefinedTargetVersions() {
   // a generic 64bit version
   if (global.params.isLP64) {
     VersionCondition::addPredefinedGlobalIdent("D_LP64");
+  }
+
+  // 16 bit version for 16 bit targets
+  if(arch == llvm::Triple::msp430) {
+    VersionCondition::addPredefinedGlobalIdent("D_P16");
   }
 
   if (gTargetMachine->getRelocationModel() == llvm::Reloc::PIC_) {
@@ -900,6 +908,10 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("AIX");
     VersionCondition::addPredefinedGlobalIdent("Posix");
     break;
+  case llvm::Triple::UnknownOS:
+    if(arch == llvm::Triple::msp430)
+      break;
+    // fallthrough
   default:
     switch (global.params.targetTriple->getEnvironment()) {
     case llvm::Triple::Android:
--- a/tests/codegen/ptr_16_bit.d
+++ b/tests/codegen/ptr_16_bit.d
@@ -5,5 +5,10 @@
 void* ptr;
 static assert(ptr.sizeof == 2);
 
+static assert (size_t.sizeof == 2);
+
 version(D_P16) {}
+else static assert(0);
+
+version(MSP430) {}
 else static assert(0);
--- a/tests/codegen/ptr_16_bit.d
+++ b/tests/codegen/ptr_16_bit.d
@@ -5,8 +5,6 @@
 void* ptr;
 static assert(ptr.sizeof == 2);
 
-static assert (size_t.sizeof == 2);
-
 version(D_P16) {}
 else static assert(0);
 
--- a/tests/codegen/ptr_16_bit.d
+++ b/tests/codegen/ptr_16_bit.d
@@ -1,0 +1,9 @@
+// REQUIRES: target_MSP430
+
+// RUN: %ldc -mtriple=msp430 -o- %s
+
+void* ptr;
+static assert(ptr.sizeof == 2);
+
+version(D_P16) {}
+else static assert(0);
--- a/gen/tollvm.cpp
+++ b/gen/tollvm.cpp
@@ -261,14 +261,11 @@ LLIntegerType *DtoSize_t() {
 
     if (triple->isArch64Bit()) {
       t = LLType::getInt64Ty(gIR->context());
-    }
-    else if (triple->isArch32Bit()) {
+    } else if (triple->isArch32Bit()) {
       t = LLType::getInt32Ty(gIR->context());
-    }
-    else if (triple->isArch16Bit()) {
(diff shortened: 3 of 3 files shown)


JohanEngelen:
Note that `t` is static, so the work is only done once anyway.

I'd prefer a more uniform version with `isArch64Bit(), isArch32Bit , isArch16Bit()`. What do you think?  (in that case you can even add an `else { llvm_unreachable("Unsupported size_t width");` so that things will break once 8 or 128bit becomes a thing!)

JohanEngelen:
To add a testcase for this, could you have a look at our lit-based tests? https://wiki.dlang.org/LDC_Lit-based_testsuite
(we don't want your work to support msp430 and other 16bit targets to regress!)

Perhaps you can use `-mtriple=i386-linux-gnu-code16` as triple and add `// REQUIRES: target_X86` to the testcase, so it's tested on our CI systems.

Let me know if you want help.

luismarques:
How about we do `getArchPointerBitWidth` and then generate an int of that size? I saw `getIntNTy` but I'm not sure that's public. I can check the details if you agree with that avenue.

JohanEngelen:
LLVM's `getArchPointerBitWidth` would be nicer, but it is not public.

luismarques:
@JohanEngelen Do you have experience running lit on macOS?

```
$ python -m pip install lit
Collecting lit
  Retrying (Retry(total=4, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', error(54, 'Connection reset by peer'))': /simple/lit/
  Retrying (Retry(total=3, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', error(54, 'Connection reset by peer'))': /simple/lit/
  Retrying (Retry(total=2, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', error(54, 'Connection reset by peer'))': /simple/lit/
  Retrying (Retry(total=1, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', error(54, 'Connection reset by peer'))': /simple/lit/
  Retrying (Retry(total=0, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', error(54, 'Connection reset by peer'))': /simple/lit/
  Could not find a version that satisfies the requirement lit (from versions: )
No matching distribution found for lit
```

luismarques:
@JohanEngelen Bah, it was Google's DNS server that was screwing things up...

luismarques:
@JohanEngelen:
> Perhaps you can use -mtriple=i386-linux-gnu-code16 as triple and add // REQUIRES: target_X86 to the testcase, so it's tested on our CI systems.

Your idea here is that that triple would specify a 16-bit target, right (x86 using 16 bit addresses)? But that triple matches `isArch32Bit`, not 16 bit. If that's your intent and it can't be worked around, can't we just target MSP430 in the testcase? Or would that be a problem for the CI system? (e.g., it uses an LLVM without all the targets).

JohanEngelen:
> Or would that be a problem for the CI system? (e.g., it uses an LLVM without all the targets).

That was my worry indeed.
Just checked the Travis CI log: MSP430 is there.  So you can write a test with `// REQUIRES: target_msp430` and it should be tested (should work on your system too, something `target_*`)

For lit tests:  `./runlit.py -v .`  or `./runlit.py -v codegen/something.d` from the build dir works wonders on OSX!

luismarques:
Due to the requirement to include the new test, I broadened the scope of this pull request to include the other changes needed for `-mtriple=msp430` to work. As part of those other changes, I defined a version identifier for 16-bit code called `D_P16`, by analogy with `D_LP64`. Now, `D_LP64` was an awful name because it means 64-bit in general and not C's LP64 in particular. I chose `D_P16` to mean pointers are 16-bit, but now I'm thinking if we should just call it `D_16`. In theory we could have a Harvard architecture where the native integer size is different from the native pointer size. That's one argument in favor of `D_P16`. Another argument would be consistency with `D_LP64`.

driver/main.cpp
```diff
@@ -767,6 +767,9 @@ void registerPredefinedTargetVersions() {
     registerPredefinedFloatABI("MIPS_SoftFloat", "MIPS_HardFloat");
     registerMipsABI();
     break;
+  case llvm::Triple::msp430:
+    VersionCondition::addPredefinedGlobalIdent("D_P16");
```
JohanEngelen: This should be moved down to where "D_LP64" is added, with a `triple->isArch16Bit()`.

driver/main.cpp
```diff
@@ -900,6 +903,9 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("AIX");
     VersionCondition::addPredefinedGlobalIdent("Posix");
     break;
+  case llvm::Triple::UnknownOS:
+    if(arch == llvm::Triple::msp430)
+      break;
```
JohanEngelen: can you add a `// fallthrough` comment, so that it is clear that that's intended?

tests/codegen/ptr_16_bit.d
```diff
@@ -0,0 +1,6 @@
+// REQUIRES: target_MSP430
+
+// RUN: %ldc -mtriple=msp430 -c -output-s %s
+
+void* ptr;
+static assert(ptr.sizeof == 2);
```
JohanEngelen: here you can also add a check that confirms `"D_16"` is defined.

tests/codegen/ptr_16_bit.d
```diff
@@ -0,0 +1,6 @@
+// REQUIRES: target_MSP430
+
+// RUN: %ldc -mtriple=msp430 -c -output-s %s
```
JohanEngelen: nit: perhaps `-o-` instead of `-c -output-s`?  Or are you explicitly checking that ASM output works? (clarify in comment)

JohanEngelen:
Two more comments:
- Perhaps a short forum discussion about adding a `D_16` version. 
- Please use clang-format to format your code.

driver/main.cpp
JohanEngelen: Do you want to add "MSP430" to the official list of reserved version identifiers?

luismarques: Yes, let's add MSP430.

JohanEngelen:
Pretty cool work btw :-)

luismarques:
I used clang-format and I amended my commit to include the formatting changes relevant to my diff. (I did not include a few changes to preexisting code).

Why are the CI builds failing?

The forum discussion about version=D_16 was so far unfruitful regarding the actual naming. Give me your opinion on D_P16 vs D_16 vs something else, and lets move on?

thewilsonator:
> Why are the CI builds failing?

Travis timed out on my dcompute build, unrelated.
Circle fails on a gdb test, it has done this for a while, unrelated.
Semaphore seems to have some issues with dwarf debug generation. Not sure why? Don't think this is related.

`D_LP64` is used to identify 64-bit mode in `object.d`.
So `D_P16` seems good to me.

kinke:
> Semaphore seems to have some issues with dwarf debug generation. Not sure why? Don't think this is related.

https://github.com/ldc-developers/ldc/issues/2189#issuecomment-313931008

JohanEngelen:
@luismarques I made some changes!  (your clang-format didn't work, or did you not commit it?; did it manually for you)

luismarques:
@JohanEngelen I used my local install of llvm/clang (/usr/local/Cellar/llvm/4.0.1/bin/clang-format). Despite being recent it complained that it didn't recognize the directive `#FixNamespaceComments: false` on LDC's `.clang-format` (I commented it out), so maybe something was misconfigured or I misused clang-format? Because I did push my commit.

@JohanEngelen In your extra testing commit you added a test for object.d's size_t sizeof. I had added that test originally, but I removed because size_t is incorrect in object.d, but changing it is problematic until the issue with the __LINE__ size is solved.

JohanEngelen:
clang-format: mine also complains (3.9), but still formats correctly. Don't know what went wrong in your setup.

Ah, size_t is in object.d.... sorry. Removing it.

JohanEngelen:
the unknown clang-format key is very new: https://reviews.llvm.org/rL296632
It still formats fine though, so no need to comment it out.

JohanEngelen:
@luismarques This all looks good to me.  Can you also submit PRs with dlang to get the version identifiers properly reserved? See https://github.com/dlang/dmd/pull/5850 and https://github.com/dlang/dmd/pull/5864

luismarques:
@JohanEngelen done.
<https://github.com/dlang/dmd/pull/6994>
<https://github.com/dlang/dlang.org/pull/1822>

JohanEngelen:
Thanks :)
Once the PR gets merged upstream, it'd be nice if you can cherry pick it onto LDC.

luismarques:
Okay. Thanks for all the help.

PR #2195, opened 2017-07-07
Dcompute fixes

thewilsonator:


gen/dibuilder.cpp
```diff
@@ -77,12 +77,18 @@ llvm::StringRef uniqueIdent(Type* t) {
 
 
 bool ldc::DIBuilder::mustEmitFullDebugInfo() {
-  // only for -g and -gc
+  // only for -g and -gc but not dcompute (yet)
```
JohanEngelen: best to add "TODO" here so it's findable  (same below)

thewilsonator: Good Idea.

JohanEngelen: and you should add a testcase that passes `-g` together with dcompute flags, testing that it doesn't break compilation, right?

JohanEngelen:
What is this fixing?  (looking at the fixes, I would expect 3 testcases)

thewilsonator:
Will do, I've got to disappear a typeid somewhere first.

thewilsonator:
Added tests.

PR #2201, opened 2017-07-11
Update LDC 0.17.x to LLVM 5.0

redstar:
This PR updates ltsmaster to use the upcoming LLVM 5.0 version.

I hope that I picked all relevant commits. It does not compile yet - that's something for the next evening.


gen/functions.cpp
```diff
@@ -391,13 +393,62 @@ void DtoResolveFunction(FuncDeclaration *fdecl) {
...
+                 lessPreciseFPMADOption ? "true" : "false");
+  func.addFnAttr("no-infs-fp-math", TO.NoInfsFPMath ? "true" : "false");
+  func.addFnAttr("no-nans-fp-math", TO.NoNaNsFPMath ? "true" : "false");
+#if LDC_LLVM_VER < 307
+  func.addFnAttr("use-soft-float", TO.UseSoftFloat ? "true" : "false");
+#endif
+
+  // Frame pointer elimination
+  func.addFnAttr("no-frame-pointer-elim",
+                 opts::disableFpElim ? "true" : "false");
```
thewilsonator: CI fails because "‘disableFpElim’ is not a member of ‘opts’". I think you forgot to cherrypick that too?

redstar: Yes, I am aware of this. It's only part of the problem. I think some files I picked are not needed (PGO for 0.17.x??). In general, it does not look too bad. I hope to fix it soon.

JohanEngelen:
There are a bunch of things that are not needed: for example LTO things, `createAndSetDiagnosticsOutputFile` and `archiver.cpp`.

gen/functions.cpp
redstar: Turned out to be a missing include.... :-)

JohanEngelen:
To make things a little easier to grasp, could you already merge the adding of harmless LLVM5.0 things like FileCheck and profile-rt?

JohanEngelen:
Frankly @redstar , I think it is much much easier to not cherry pick, and only fix the things that are broken, while looking at how they are done in master.  The current state of this PR adds many new things to 0.17.x that are not needed and are likely to be half-broken this way.

redstar:
@JohanEngelen You can be right. I think I'll give it a try to fix the compile errors only. Nevertheless I committed the first profile-rt update as this is indeed harmless.

JohanEngelen:
@redstar I've made my own branch, can we close this one?
https://github.com/ldc-developers/ldc/pull/2254

redstar:
@JohanEngelen For sure. Thanks for the work.

PR #2202, opened 2017-07-13
Remove prelude assert from ctors & dtors (Cherrypick dmd/6982)

thewilsonator:
This is a cherrypick of https://github.com/dlang/dmd/pull/6982

See also https://forum.dlang.org/thread/btddutbfirtdydmufwkj@forum.dlang.org

The asserts are almost completely useless as they are only ever triggered when trying to construct or destruct an object at the null location.

This is required to build DCompute without the use of the `-release`, as asserts and strings are not supported.

Any help pushing https://github.com/dlang/dmd/pull/6982 forward is appreciated.


kinke:
16 commits... ;) - looking into git in some more detail may be well worth the effort. In this trivial case, it'd be something like this, assuming you are on a clean local `invariants` branch:
```
git reset --hard origin/master [dangerous, make sure your working dir is clean]
git cherry-pick 34c8bab
git push -f
```

thewilsonator:
Heh, that got rid of some of them. I think some of those commits were on my master branch which invariants diverges from.

thewilsonator:
Hmm, any idea why the PGO test it failing?

JohanEngelen:
The PGO test is testing `// Explicitly test that the assert(this) branch is *not* profiled`.
Removing lines 120:121 and 158-159 doesn't work?

thewilsonator:
Ha I was just looking at that. I'll remove it.

thewilsonator:
Do you happen to know where it was being excluded from PGO (PGO is your thing right :) )? Thats dead code now.

thewilsonator:
The Circle failure is the GDB one, the semaphore failures are the dwarf/debug ones.
Merging.

PR #2205, opened 2017-07-14
Make the producer name in debuginfo more precise.

JohanEngelen:
Instead of `"LDC (http://wiki.dlang.org/LDC)"`, this PR sets the producer string to the full version of LDC, for example:
`"LDC 1.4.0git-c041205-dirty (LLVM 4.0.1svn-r307596)"`
or 
`"LDC 1.4.0 (LLVM 4.0.0)"`

Shall I also add the DMDfe version?

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -10,6 +10,7 @@
 #include "gen/dibuilder.h"
 
 #include "driver/cl_options.h"
+#include "driver/ldc-version.h"
 #include "gen/functions.h"
 #include "gen/irstate.h"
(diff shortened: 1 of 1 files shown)


rainers:
Does this end up somewhere in the object file or executable? I might be interested (for mago) if it can be found in CodeView debug info.

> Shall I also add the DMDfe version?

I guess the LDC version is very much coupled with a front end and runtime library version, so it doesn't really add much value.

JohanEngelen:
Don't know for sure if it is added to the object file, but I suppose so yes.

PR #2206, opened 2017-07-14
Fix LLVM 5.0 build

JohanEngelen:
Quite a significant change here: https://reviews.llvm.org/D21723

Shall we also update the identifiers in the SyncScopeID struct for LLVM>=5.0?
https://github.com/ldc-developers/druntime/blob/cc9335131218cf157af7e562309b64e5b0c51bb3/src/ldc/intrinsics.di#L357-L361


JohanEngelen:
@thewilsonator Another update required for the SPIRV build!

thewilsonator:
New one up, hot off my laptop.

thewilsonator:
> Quite a significant change here: https://reviews.llvm.org/D21723

That looks like a very interesting change very applicable for SPIRV. Will have to take a closer look.

>Shall we also update the identifiers in the SyncScopeID struct for LLVM>=5.0?

Yeah that would be great.

gen/tocall.cpp
```diff
@@ -332,11 +332,17 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
     }
     auto atomicOrdering =
         static_cast<llvm::AtomicOrdering>((*e->arguments)[0]->toInteger());
+#if LDC_LLVM_VER >= 500
+    llvm::SyncScope::ID scope = (e->arguments->dim == 2)
+                                    ? (*e->arguments)[1]->toInteger()
+                                    : llvm::SyncScope::System;
+#else
```
thewilsonator: I think you need to `#include "llvm/IR/LLVMContext.h"` at the top of this file so that `llvm::SyncScope` is in scope.

thewilsonator:
Other than the missing header and assuming the tests work, LGTM.

kinke:
It's not Circle's fault, it's the [LLVM APT repo](http://apt.llvm.org/) which only provides frequent updates (something like daily) for the 3 latest Ubuntu versions (16.04, 16.10 and 17.04) and only seldom updates for the older ones (trusty: June 21st, which is actually quite good, compared to precise with April 12th (!)).

kinke:
It may pay off to look into CircleCI's docker support and use a Ubuntu 16.04 image for it, so that its LLVM 5.0 is always up to speed.

PR #2207, opened 2017-07-14
Switch to special quiet NaN as init value for all FP types

kinke:
**Update**:

By letting LLVM truncate compile-time `real_t` values to float/double IR constants. We previously relied on the host, which turns signalling real_t NaNs to quiet float/double NaNs.

Signalling NaNs are characterized by their most-significant mantissa bit being 0 (and at least for D the 2nd-most-significant mantissa bit being set to distinguish them from infinity). The host truncation led to the MSB
getting set while keeping the rest of the mantissa (i.e., both MSBs set).

Here's the resulting bitpattern table in hex:

|           | float  |   double  |          x87 real |
| ------ | ----- | ---------- | --------------- |
| init old |  7f**e**00000   |7ff**c**000000000000  | 7fff**a**000000000000000 |
| init new |  7f**a**00000 |  7ff**4**000000000000 |  7fff**a**000000000000000 |
| qnan  |     7f**c**00000  | 7ff**8**000000000000 |  7fff**c**000000000000000 |

DMD 2.074.0 produces wrongly quiet float/double init values. It's known: https://forum.dlang.org/post/nsp1ql$ivu$1@digitalmars.com

### DIFF
--- a/ddmd/root/ctfloat.d
+++ b/ddmd/root/ctfloat.d
@@ -199,6 +199,13 @@ else
     static __gshared real_t one = real_t(1);
     static __gshared real_t minusone = real_t(-1);
     static __gshared real_t half = real_t(0.5);
+  version(IN_LLVM)
+  {
+    // Initialized via LLVM in C++.
+    static __gshared real_t initVal;
+    static __gshared real_t nan;
+    static __gshared real_t infinity;
+  }
 }
 
 version (IN_LLVM)
--- a/ddmd/root/ctfloat.d
+++ b/ddmd/root/ctfloat.d
@@ -110,6 +110,13 @@ extern (C++) struct CTFloat
         return !(r == r);
     }
 
+  version(IN_LLVM)
+  {
+    // LDC doesn't need isSNaN(). The upstream implementation is tailored for
+    // DMD/x86 and only supports double-precision and x87 real_t types.
+  }
+  else
+  {
     static bool isSNaN(real_t r)
     {
         static if (real_t.sizeof == 8)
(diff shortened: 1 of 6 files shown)


thewilsonator:
In `codegen/vector_init.d` on x64 the NaNs in the `__initZ` symbol seem to initialised to 

0x7FF4000000000000

instead of 

0x7FFC000000000000

Not sure why, considering it passes on x86. It is the only failure though. 

Should we be implementing `Target::RealProperties::[s]nan()` in C++ as well?

kinke:
The new value `0x7FF4000000000000` is a signalling NaN, whereas the tested one is a quiet NaN. The inconsistency between 32 and 64 bits is especially ugly (and must be caused by DMD [2.074.0], as that's the first compiler I used to generate the 1.3 LDC bootstrap compiler which then builds 1.3 final).

kinke:
[Changed the scope of the PR after figuring out what the actual issue is.]

kinke:
To make clear we don't need that buggy version assuming x87 real_t if size != 8.

kinke:
Win32 unexpectedly still fails with quiet `double.init`.

kinke:
To be more precise: the Win32 LDC compiler generates quiet ones; the 64-bit build with `-m32` signalling ones. Possibly something to do with SSE vs. x87.

kinke:
In the long term, using quiet FP inits (with the second-most mantissa bit set too, to distinguish them from normal quiet NaNs) for all 3 types may make more sense.

kinke:
Well, why wait => switched to quiet `{float,double,real}.init`.

kinke:
Any objections to a consistent special quiet NaN for all FP types?

JohanEngelen:
can't help judging here, maybe ask on the forums if you're unsure.

kinke:
No feedback in the forum - merging.

PR #2209, opened 2017-07-15
Upgrade to CircleCI 2.0

kinke:
Based on the latest `gcc` docker image (currently 7.1) which is (currently) based on Debian 9 (aka Stretch, released 9 days ago). Try using the LLVM nightly for Debian sid (unstable) as their stable branch is apparently still for Debian 8 (Jessie). That LLVM nightly is updated ~daily.


.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+          name: Install LDC host compiler
+          command: |
+            if [[ ! -e ldc2-$HOST_LDC_VERSION-linux-x86_64/bin/ldmd2 ]]; then
+              wget https://github.com/ldc-developers/ldc/releases/download/v$HOST_LDC_VERSION/ldc2-$HOST_LDC_VERSION-linux-x86_64.tar.xz
+              tar -xf ldc2-$HOST_LDC_VERSION-linux-x86_64.tar.xz
+            fi
+      - save_cache:
+          key: host-ldc-{{ .Environment.HOST_LDC_VERSION }}
+          paths:
+            - ldc2-$HOST_LDC_VERSION-linux-x86_64
```
kinke: They don't support user-defined environment variables in cache keys (see [here](https://discuss.circleci.com/t/cannot-use-circle-yml-environment-variables-in-cache-keys/10994)) and neither in paths yet, so the first host compiler isn't cached.

thewilsonator:
Looks good, there still that gdb failure there from ages ago.

kinke:
The `runnable/gdb4149.d` issue is still there (with gdb v7.12 now), which prevents the stdlib unittests from running, but everything else seems to work fine, incl. a bunch of new (helpful) warnings from gcc v7.1. The timings are also quite nice (25 mins without stdlib unittests, incl. building LDC twice) and definitely an improvement from CircleCI 1.0.

So for Linux x64, SemaphoreCI plays the role as speedy mulitlib test with enabled LLVM+LDC assertions in a conservative Ubuntu environment (Ubuntu 14.04, gcc 4.9), CircleCI tests LLVM trunk incl. LTO in a recent Debian environment, and Travis tests the rest in a soon-to-be Ubuntu 14.04 environment.

kinke:
More work to be figured out at some point:
```
The following tests FAILED:
	 25 - core.internal.convert (Failed)
	 63 - core.time (Failed)
	360 - std.math (Failed)
	416 - core.internal.convert-debug (Failed)
	807 - core.internal.convert-shared (Failed)
	845 - core.time-shared (Failed)
	1142 - std.math-shared (Failed)
	1198 - core.internal.convert-debug-shared (Failed)
```

3 out of 4 `std.datetime` unittests require about 110 secs each, the 4 GB memory seem to be limiting here. I'll reduce parallelization.

kinke:
> 3 out of 4 std.datetime unittests require about 110 secs each

It's not much better when using 2 jobs, it takes ~60 seconds then. With 2 CPU cores, a halved time wrt. 4 jobs makes perfect sense if it's NOT due to memory limitation, but a potential LLVM 5.0 regression (Semaphore with LLVM 4.0.1 requires about 3 seconds for `std.datetime`).

PR #2210, opened 2017-07-15
Travis: Switch to Trusty

kinke:
They're migrating to Trusty implicitly starting with July 18th.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: d
 sudo: false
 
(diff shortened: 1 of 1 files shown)


kinke:
Right, so we have 2 out of 6 Travis Linux jobs completing in under 17 minutes on a weekend?! ;)

JohanEngelen:
sweeet

kinke:
Apparently the timings were no flukes, almost identical ones (17-22 mins) again after merging, while the immediatetly preceding master commit took 23-42 mins. So I'm pretty sure the bad Travis timings lately were related to their deprecation process.

PR #2211, opened 2017-07-16
Drop support for LLVM < 3.7

kinke:


thewilsonator:
LGTM.

JohanEngelen:
Nice. Merge when you are happy with it @kinke 

IIRC, ArrayRef now accepts initializer lists, right?

thewilsonator:
ArrayRef has accepted initialiser lists for quite a while I think.

kinke:
Yep, good catch, it's guaranteed to be there for LLVM 3.7+; for 3.5 and 3.6 `LLVM_HAS_INITIALIZER_LISTS` had to be defined. So there may be room for some simplifications (as well as re-clang-formatting; I only reformatted a few places manually to keep the diff clearer).

PR #2214, opened 2017-07-16
Add search logic for AddressSanitizer library on Linuxes.

JohanEngelen:
Only tested on mac with -mtriple=x86_64-pc-linux, which does the correct thing.
As least on Ubuntu, the libs are called: `libclang_rt.asan-x86_64.a` and `libclang_rt.asan-x86_64.so`.

The shared lib support is halfbaked. I put in some code to make it easy for someone to make that work in the future.

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -203,12 +203,53 @@ bool addDarwinASanLinkFlags(std::vector<std::string> &args) {
   return false;
 }
 
+// Returns the arch name as used in the compiler_rt libs.
+// FIXME: implement correctly for non-x86 platforms (e.g. ARM)
+llvm::StringRef getCompilerRTArchName() {
+  return global.params.targetTriple->getArchName();
+}
+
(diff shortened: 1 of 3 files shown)


driver/linker-gcc.cpp
```diff
@@ -203,12 +203,53 @@ bool addDarwinASanLinkFlags(std::vector<std::string> &args) {
   return false;
 }
 
+// Returns the arch name as used in the compiler_rt libs.
+// FIXME: implement correctly for non-x86 platforms (e.g. ARM)
+llvm::StringRef getCompilerRTArchName() {
+  return global.params.targetTriple->getArchName();
+}
+
+bool addUnixlikeASanLinkFlags(std::vector<std::string> &args) {
```
JohanEngelen: I don't like the `addUnixlikeASanLinkFlags` name, any suggestions?

JohanEngelen:
Can I merge this?
It is working well together with the follow up #2219 that shows that TravisCI and CircleCI can now do ASan tests! (both Linux and OSX)

PR #2215, opened 2017-07-18
[dcompute] output kernels in object directory

thewilsonator:
Place the built kernels in the same directory as the object files. The build system should at least know where that is. This shifts the onus of informing the executable of where the kernels from the compiler to the build system.


JohanEngelen:
clang-format
and perhaps nicer variable names than `tmp`? (the length of 32 of tmp looks worrying, too)

thewilsonator:
the 32 is plenty for the moment
the longest name is `kernels_cuda350_64.ptx\0` which is 23.

JohanEngelen:
OK, perhaps add an assert on `len < 32` then.

JohanEngelen:
(and the null terminating character is added already by `sprintf`)

kinke:
> the null terminating character is added already by sprintf

Older MS versions didn't if the resulting string length matched exactly the buffer length.

JohanEngelen:
> Older MS versions didn't if the resulting string length matched exactly the buffer length.

How would it know the buffer length?

kinke:
Ah yeah right that was s**n**printf.

thewilsonator:
Thanks Kinke, but the size will never (currently) exceed 23 + the null terminator, thats why I went for the stack buffer. I dont the the SSO of `std::string` will cover that. Can we use an `llvm::SmallString`? I can't seem to find a `raw_foo_stream` for small string.

JohanEngelen:
you can use snprintf (note the extra 'n').

kinke:
No need to be overly pedantic IMO, the C++ streaming is ugly but safe and gets rid of the gcc warning about potential buffer overflow, that's the primary reason I opted for it.
And about small string optimization - if you really think this has a measurable impact on the compilation runtime, we could reserve some 128/256 bytes, but I thought this nano-optimization isn't worth it.

JohanEngelen:
(ah sry, didn't see you already used the stream thing. perfect :)

PR #2217, opened 2017-07-18
CMake: Add logic to copy runtime libraries from LLVM/Clang to LDC's lib dir

JohanEngelen:
Tested on OSX, with manually-built LLVM (checked out Clang and compiler-rt in the expected dirs inside the LLVM source dir before running cmake and building; so the total installed build includes LLVM, Clang, and compiler-rt).

We can use the same for profile-rt.

Perhaps we should revisit where we put our runtime libs. Clang puts them in a versioned+OS dir: `lib/clang/4.0.0/darwin/`.


JohanEngelen:
nice: works on Travis, so we can test address sanitizer there.

JohanEngelen:
Needed to run otool on Darwin to fix-up after the dylib rename.

JohanEngelen:
It's working on the CI platforms :-)

PR #2219, opened 2017-07-19
Add AddressSanitizer lit testing

JohanEngelen:
Depends on 
- [x] https://github.com/ldc-developers/ldc/pull/2217 
- [x] #2214


JohanEngelen:
See https://forum.dlang.org/post/zoiyldrfgjlpfhzhhtfn@forum.dlang.org for why I had to remove the adding of LLVM bin dir to the path in the lit configuration script. (when llvm-symbolizer is on the path, there is no line debug information in ASan output, unless dsymutil is run)

JohanEngelen:
Ah right, this of course also depends on https://github.com/ldc-developers/ldc/pull/2214

JohanEngelen:
Already worked on TravisCI OSX! :-) Now testing linux.

Windows: it looks like ASan should also work there, but I have never tried.

JohanEngelen:
Looks like CircleCI also has a debuginfo issue where line numbers are not showing :(
Perhaps I have to remove that check from the test then...

JohanEngelen:
> Perhaps I have to remove that check from the test then...

Nope, setting the llvm tools path (but not on Darwin!) fixed it. All green.

JohanEngelen:
@thewilsonator Would be great if you could build the LLVM SPIRV special build with compiler-rt too (you'll need to checkout clang too to get it built I believe). Also, libFuzzer would be nice... ;-)
1. checkout llvm
2. checkout compiler-rt in projects dir
3. checkout clang in tools dir
4. build llvm (this builds clang and compiler-rt too)
5. build libFuzzer (go to lib/Fuzzer and run build.sh) --> creates libFuzzer.a, copy it to LLVM's lib install dir
That way we would get ASan+Fuzz testing on the OSX SPIRV tester too.

thewilsonator:
Yay, I dont need to build clang, the sanitiser stuff ends up unser `./clang/6.0.0/lib/darwin/*`. I didn't run `build.sh` it does it automatically if you configure CMake correctly but ends up as `libLLVMFuzzer.a`. We will need to change our CMake as it is up to LLVM6.0 now.

thewilsonator:
To build without clang, disable all cmake `COMPILER_RT_*` options except COMPILER_RT_BUILD_SANITIZERS, COMPILER_RT_SANITIZERS_TO_BUILD and COMPILER_RT_DEFAULT_TARGET_TRIPLE.

JohanEngelen:
>  I didn't run build.sh it does it automatically if you configure CMake correctly but ends up as libLLVMFuzzer.a.

What's the config switch?

thewilsonator:
IDK, i just built it. Although when I tried to built ldc, ldc's "find llvm" cmake failed so I resorted to building as you described.

JohanEngelen:
libFuzzer is most likely going to be moved to compiler-rt, so it's exact location and name may change. I'll keep an eye out for it.

PR #2224, opened 2017-07-20
Add automatic libFuzzer linking + fuzz test

JohanEngelen:
This depends on first pulling
- [x] #2219

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -29,6 +29,10 @@ static llvm::cl::opt<std::string>
                               "LLVMgold.so (Unixes) or libLTO.dylib (Darwin))"),
                llvm::cl::value_desc("file"));
 
+static llvm::cl::opt<bool> linkNoCpp(
+    "link-no-cpp", llvm::cl::ZeroOrMore, llvm::cl::Hidden,
+    llvm::cl::desc("Disable automatic linking with the C++ standard library."));
+
 //////////////////////////////////////////////////////////////////////////////
 
 namespace {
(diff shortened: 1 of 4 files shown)


JohanEngelen:
Right now, only CircleCI has libFuzzer available. 
On my Mac, things work splendid as well :-)

JohanEngelen:
Added potential Android support (untested).

JohanEngelen:
@kinke Any way I can see what the build script is on Semaphore? (also the CMake output would be nice to see in the test report)

kinke:
The build script is what you see - the commands when clicking on a job. If you want access to it, sign up and let me know your account ID so I can add you to the project (apparently no GitHub integration). Beware of the `16 more setup commands` at the top of each job, these are the shared prepended commands and include the CMake output.

JohanEngelen:
>  Beware of the 16 more setup commands at the top of each job,

This.

JohanEngelen:
My ID = https://semaphoreci.com/johanengelen

JohanEngelen:
probably just `-DLDC_INSTALL_LLVM_RUNTIME_LIBS=ON` is needed.

JohanEngelen:
Should we have a way to disable linking with the C++ lib, so that the user has a chance to provide it himself if LDC is assuming the wrong one?

JohanEngelen:
@klickverbot What do you think about the C++ lib thing? How fragile is that?

JohanEngelen:
`-x c++` is recognized but doesn't work (on mac).
I'll add a hidden switch so people can turn off the adding of `-lc++`.

PR #2226, opened 2017-07-21
Revise LDMD

kinke:


driver/ldmd.cpp
```diff
@@ -75,9 +50,11 @@
 
 namespace ls = llvm::sys;
 
-// We reuse DMD's response file parsing routine for maximum compatibilty - it
+// We reuse DMD's response file parsing routine for maximum compatibility - it
 // handles quotes in a very peculiar way.
 int response_expand(size_t *pargc, char ***pargv);
+
+// in ddmd/root/man.d
 void browse(const char *url);
```
kinke: The only LDMD part implemented in D, otherwise it'd be C++ only.

CMakeLists.txt
```diff
@@ -704,7 +704,7 @@ build_d_executable(
     "${LDMD_D_SOURCE_FILES}"
     "$<TARGET_LINKER_FILE:LDMD_CXX_LIB>"
     "${LDMD_D_SOURCE_FILES}"
-    "LDMD_CXX_LIB;${LDC_LIB}"
+    "LDMD_CXX_LIB"
```
kinke: Linking LDMD doesn't depend on the LDC library.

PR #2236, opened 2017-07-24
Conditionally set version LDC_DCompute

thewilsonator:


driver/main.cpp
```diff
@@ -919,8 +919,11 @@ void registerPredefinedVersions() {
   VersionCondition::addPredefinedGlobalIdent("LDC");
   VersionCondition::addPredefinedGlobalIdent("all");
   VersionCondition::addPredefinedGlobalIdent("D_Version2");
+
 #if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX
-  VersionCondition::addPredefinedGlobalIdent("LDC_DCompute");
+  if (dcomputeTargets.size() != 0) {
```
ZombineDev: @thewilsonator how come this information is known only at run-time, but `LDC_LLVM_SUPPORTED_TARGET_SPIRV` or `LDC_LLVM_SUPPORTED_TARGET_NVPTX` is set at compile-time?

JohanEngelen: @ZombineDev The intended meaning of LDC_DCompute is "compiled with DCompute enabled", instead of "compiler supports DCompute". That's then similar to the meaning of e.g. `LDC_AddressSanitizer`.  You can debate about this of course, and perhaps people would like a `LDC_SupportsDCompute` version identifier too.

ZombineDev: Of course you're right, thanks for the quick explanation.

PR #2238, opened 2017-07-24
profile-rt for LLVM 6.0

thewilsonator:


JohanEngelen:
Thanks! For next time: we don't need the CMakeLists.txt file from compiler-rt. (it's a little confusing to have it in our repo, so let's remove it)

thewilsonator:
Oh whoops

PR #2239, opened 2017-07-24
CMake: also install the ASan and libFuzzer libraries

JohanEngelen:


CMakeLists.txt
```diff
@@ -754,6 +754,7 @@ function(copy_compilerrt_lib llvm_lib_name ldc_lib_name fixup_dylib)
         if (fixup_dylib)
             execute_process(COMMAND install_name_tool -id @rpath/${ldc_lib_name} ${PROJECT_BINARY_DIR}/lib${LIB_SUFFIX}/${ldc_lib_name})
         endif()
+        install(PROGRAMS ${PROJECT_BINARY_DIR}/lib${LIB_SUFFIX}/${ldc_lib_name} DESTINATION ${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX})
```
kinke: If the executable permission isn't required, `install(FILES ...)` suffices.

JohanEngelen: OK thanks, I changed it.
(I checked: the dylib from LLVM's build are marked +x, but at least on Mac that is not needed...)

kinke: Thx, I also checked on Linux, where the system .so's are 644 by default.

PR #2243, opened 2017-07-26
Integrate Polly the polyhedral optimiser

thewilsonator:


gen/uda.cpp
```diff
@@ -1,5 +1,5 @@
+#include "driver/cl_options.h"
```
JohanEngelen: move this down.
The `gen/uda.h` is intentionally put first and separated from the rest, such that it is proven that the .cpp's own header file can be included without any prerequisites.

gen/uda.cpp
```diff
@@ -393,8 +393,24 @@ void applyFuncDeclUDAs(FuncDeclaration *decl, IrFunction *irFunc) {
...
+    } else if (ident == Id::udaWeak) {
+      // @weak is applied elsewhere
+    } else if (ident == Id::udaKernel) {
+      // @kernel is applied elsewhere but warn if dcompute is
+      // either not supported or enabled
+#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX
+      if (!opts::dcomputeTargets.size())
+        sle->warning("dcompute not enabled: @kernel has no effect");
+#else
+      sle->warning("dcompute not supported: @kernel has no effect");
```
JohanEngelen: We should be careful adding warnings for things that may be fairly benign. There is no way for the user to turn off specific warnings (my attempt towards that direction was slain mercilessly by the dmd dictators, still frustrated about that), and thus it is already hard to build with warnings as errors. Adding these "no effect" warnings creates problems with -we for projects that should build with and without dcompute.

JohanEngelen: (I just document the cases where an attribute may have no effect, like for `@allocSize`)

thewilsonator: Fair enough.

thewilsonator:
I do realise that the test in its current state is more suited to `tests/compilable` but I do intend to test the IR once I figure out what it does.

JohanEngelen:
You should also add simple tests regarding the pollskip attribute: test all cases where it should and should not be added.
It is very unfortunate that you have to _dis_able polly rather than _en_able it. This is probably not going to work well with LTO and different compiler versions. Did you ask on the LLVM list whether there is a way to make polly opt-in, instead of opt-out ?

kinke:
[The druntime commit is already in master (https://github.com/ldc-developers/ldc/commit/6b7a12588fafd02a848678ed339f48bdf09a4503).]

kinke:
[Rebasing a PR to the current head - `git rebase origin/master` is preferred over merging master into PRs. Merging `master` is what I do for big PRs like merging a new front-end, as that entails updates to all submodules etc.]

kinke:
Are you using some weird git tool? ;)

thewilsonator:
IDK what happened. I think that might have happened because my master is ancient. Failing all else I'll close and regenerate the diff against master, going to sleep now though.

JohanEngelen:
Don't close!!! Just fixup things and force push to this branch. If you close and open new PRs, the discussion about it will be spread around different PRs, kind of annoying.

kinke:
In case you need some guidance, here's how I would do it:

1. Figure out the commits you actually want to push and note their hashes.
2. Create a new temporary branch, branching off current remote master: `git checkout -b tmp origin/master`.
3. Cherry-pick the commits from step 1, in creation order. You may need to adjust some if there are conflicts. `git cherry-pick <hash>`.
4. Make sure there's no src tree diff wrt. this branch (assuming this one is okay): `git diff <yourRemote>/polly2`.
5. All good, make sure your working dir is clean (`git status`) and then you can reset your local `polly2` branch to the temporary one: `git checkout polly2 && git reset --hard tmp`. A git branch is not much more than a named pointer to some commit in the git DAG, potentially tracking some remote branch.
6. Still on your local `polly2` branch, push-force it to your remote: `git push -f` (assuming your local one is tracking it, otherwise something like `git push -u <yourRemote> polly2`.
7. Remove the temporary branch: `git branch -D tmp`.

JohanEngelen:
See https://lists.llvm.org/pipermail/llvm-dev/2017-September/117063.html

thewilsonator:
We should move to MLIR if we want this

PR #2246, opened 2017-07-28
Travis: Reduce #parallel build jobs to reduce sporadic failures

kinke:
... due to killed processes when exhausting the worker's memory (4 GB **max** on Linux, unfortunately).


kinke:
I'm a bit reluctant to merge this, as build times seem to suffer, but accurate measurements aren't possible for Travis, the variance is very high [and the blissful period of Linux jobs taking ~20 mins was unfortunately very short], so this is more of a gut feeling. ;)

kinke:
Apparently not required anymore due to selective reduced parallelization for memory-intensive unittest modules.

Issue #2235, opened 2017-07-23
aligned struct segfaults garbage collector


KytoDragon:
I came upon this bug while trying to write a sample program for a different bug (assigning aligned structs in unions overwrites data located after the union).
```
module test.test;

align(4) struct Foo {
    void* y;
    byte z;
}

class Bar {
    
    Foo foo;
    byte x;

    void set(Foo f) {
        x = 99;
        foo = f;
    }
}

void main() {
    Bar bar = new Bar();
    Foo f;
    bar.set(f);
}
```
Stacktrace:
```
Segfault!
Tried to read from 0x0000000000000000
Unknown source: test.exe: _D2gc4impl12conservative2gc3Gcx7prepareMFNbZv
Unknown source: test.exe: _D2gc4impl12conservative2gc3Gcx11fullcollectMFNbbZm
Unknown source: test.exe: _D2gc4impl12conservative2gc14ConservativeGC157__T9runLockedS107_D2gc4impl12conservative2gc14ConservativeGC18fullCollectNoStackMFNbZ2goFNbPS2gc4impl12conservative2gc3GcxZmTPS2gc4impl12conservative2gc3GcxZ9runLockedMFNbKPS2gc4impl12conservative2gc3GcxZm
Unknown source: test.exe: gc_term
Unknown source: test.exe: rt_term
Unknown source: test.exe: _d_run_main
\D\Test_union\__entrypoint.d(8): __entrypoint.main
Unknown source: kernel32.dll: BaseThreadInitThunk
Unknown source: ntdll.dll: RtlUserThreadStart
```
Setting the alignment to 1 or 8 fixes it, setting it to 2 or 4 triggers it. Removing the member x, its assignment in set() or changing the order of members in either Foo or Bar fixes it.
Using Windows 7 x64 using LDC 1.3.0.
This does not happen using DMD 2.075.0

KytoDragon:
Here is my original problem with unions (which does not segfault).
```
module test.test;

align(4) struct Foo {
    void* y;
    byte z;
}

class Bar {
    
    union {
        bool b;
        Foo foo;
    }
    byte x;

    void set(Foo f) {
        x = 99;
        foo = f;
    }
}

void main() {
    Bar bar = new Bar();
    Foo f;
    bar.set(f);
    assert(bar.x == 99); // Fails
}
```

kinke:
> Using Windows 7 x64 using LDC 1.3.0.
> This does not happen using DMD 2.075.0

[...]

PR #2247, opened 2017-07-28
Fix issue #2235 - IR struct packedness

kinke:
Unnaturally aligned aggregates were potentially not marked as packed, leading to LLVM inserting additional padding and screwing up the memory layout.

### DIFF
--- a/ir/irtypeaggr.cpp
+++ b/ir/irtypeaggr.cpp
@@ -203,16 +203,24 @@ IrTypeAggr::IrTypeAggr(AggregateDeclaration *ad)
       aggr(ad) {}
 
 bool IrTypeAggr::isPacked(AggregateDeclaration *ad) {
-  if (ad->isUnionDeclaration()) {
-    return true;
-  }
-  for (unsigned i = 0; i < ad->fields.dim; i++) {
-    VarDeclaration *vd = static_cast<VarDeclaration *>(ad->fields.data[i]);
-    unsigned a = vd->type->alignsize() - 1;
-    if (((vd->offset + a) & ~a) != vd->offset) {
+  const auto aggregateSize = (ad->sizeok == SIZEOKdone ? ad->structsize : ~0u);
+
+  // For unions, only a subset of the fields are actually used for the IR type -
+  // don't care.
+  for (const auto field : ad->fields) {
+    // The aggregate's size and the field offset need to be multiples of the
+    // field's natural alignment, otherwise the aggregate type is unnaturally
+    // aligned, and LLVM would insert padding.
+    const auto naturalFieldAlignment = field->type->alignsize();
+    const auto mask = naturalFieldAlignment - 1;
+
+    // If the aggregate's size is unknown, any field with natural alignment > 1
+    // will make it packed.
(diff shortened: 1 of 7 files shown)


kinke:
Damn, hitting the assertion about unknown aggregate size. Using `AggregateDeclaration::size(Loc)` doesn't help.

tests/codegen/gh2235.d
```diff
@@ -0,0 +1,29 @@
+// RUN: %ldc -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+// RUN: %ldc -run %s
+
+// CHECK: %gh2235.Foo = type <{
```
JohanEngelen: Note to reviewers: the `< ... > ` mean that the struct is packed. (without them, it's a normal struct)

ir/irtypeaggr.cpp
```diff
@@ -203,16 +203,24 @@ IrTypeAggr::IrTypeAggr(AggregateDeclaration *ad)
...
-    return true;
-  }
-  for (unsigned i = 0; i < ad->fields.dim; i++) {
-    VarDeclaration *vd = static_cast<VarDeclaration *>(ad->fields.data[i]);
-    unsigned a = vd->type->alignsize() - 1;
-    if (((vd->offset + a) & ~a) != vd->offset) {
+  const auto aggregateSize = (ad->sizeok == SIZEOKdone ? ad->structsize : ~0u);
+
+  // For unions, only a subset of the fields are actually used for the IR type -
+  // don't care.
```
JohanEngelen: So now when 
```
union justOneField {
   int i;
}
```
This used to become a packed struct, but no longer is, right?

Does that change things with e.g. `justOneField[10]` (for example when the size of the union field is 5 bytes) ?

kinke: The size is determined by the front-end; we insert explicit trailing padding for naturally aligned aggregates too to make sure to match it. We just need to make sure the IR struct matches the memory layout chosen by the front-end, by taking into account that LLVM would insert implicit padding between members and pad the overall struct if natural alignment is violated. We have an assertion that the resulting LL type isn't larger than the front-end size, which was triggered by #2235.
So this just prevents all unions from automatically being represented as packed IR structs, there may be no need to (although the check would ideally only cover the union fields making it to the final IR type), and the same goes for aggregates with `align(1)` alignment.

klickverbot: I guess Johan was referring to e.g. `[8 x %justOneField]` on the IR level – could the padding for that change with the IR structs no longer being packed?

kinke: First off, unions aren't special in any way here, they are represented as IR structs as any other D aggregate, and they have a (padded) front-end size. When emitting it and there are trailing padding bytes after the last field, we emit an explicit i8 array for it. So the only way LLVM could pad the overall union further is if the D front-end size isn't a multiple of a field's natural alignment. And that's exactly what happend in #2235, the 8-bytes pointer alignment on x64 violated by the 12-bytes struct (due to outer align(4)), and that's exactly what my fix is about.

kinke: Something unclear still?

PR #2248, opened 2017-07-29
Fix a few compiler warnings

JohanEngelen:
This fixes the warning "default label in switch which covers all enumeration values". We will already get a warning when _not_ all enumeration values are covered.
I think with this PR, all warnings are resolved and we can use `-Werror` on our CI systems.

Let's not debate about this change too long. I looked into disabling the warning for only this piece of code, but it is cumbersome. So let's either apply this change, or remove `-Wcovered-switch-default` from our build flags. Thanks.

### DIFF
--- a/gen/dcompute/targetCUDA.cpp
+++ b/gen/dcompute/targetCUDA.cpp
@@ -13,6 +13,7 @@
 #include "gen/abi-nvptx.h"
 #include "gen/logger.h"
 #include "gen/optimizer.h"
+#include "gen/to_string.h"
 #include "llvm/Transforms/Scalar.h"
 #include "driver/targetmachine.h"
 #include <cstring>
@@ -47,12 +48,11 @@ public:
     // sm version?
   }
   void setGTargetMachine() override {
-    char buf[8];
-    bool is64 = global.params.is64bit;
-    snprintf(buf, sizeof(buf), "sm_%d", tversion / 10);
+    const bool is64 = global.params.is64bit;
+
     gTargetMachine = createTargetMachine(
         is64 ? "nvptx64-nvidia-cuda" : "nvptx-nvidia-cuda",
-        is64 ? "nvptx64" : "nvptx", buf, {},
+        is64 ? "nvptx64" : "nvptx", "sm_" + ldc::to_string(tversion / 10), {},
         is64 ? ExplicitBitness::M64 : ExplicitBitness::M32, ::FloatABI::Hard,
         llvm::Reloc::Static, llvm::CodeModel::Medium, codeGenOptLevel(), false,
         false);
--- a/gen/asm-x86.h
+++ b/gen/asm-x86.h
@@ -2508,7 +2508,7 @@ struct AsmProcessor {
       if (ptrtype == Byte_Ptr) {
         return false;
       }
-    // drop through
+    // fallthrough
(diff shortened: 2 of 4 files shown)


kinke:
> I think with this PR, all warnings are resolved and we can use -Werror on our CI systems.

Some of the [Circle](https://circleci.com/gh/ldc-developers/ldc/2895) warnings with LLVM 5.0:
```
/root/project/gen/dcompute/targetCUDA.cpp: In member function 'virtual void {anonymous}::TargetCUDA::setGTargetMachine()':
/root/project/gen/dcompute/targetCUDA.cpp:49:8: warning: '%d' directive output may be truncated writing between 1 and 10 bytes into a region of size 5 [-Wformat-truncation=]
   void setGTargetMachine() override {
        ^~~~~~~~~~~~~~~~~
/root/project/gen/dcompute/targetCUDA.cpp:49:8: note: directive argument in the range [-214748364, 214748364]
/root/project/gen/dcompute/targetCUDA.cpp:52:13: note: 'snprintf' output between 5 and 14 bytes into a destination of size 8
     snprintf(buf, sizeof(buf), "sm_%d", tversion / 10);
     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/root/project/gen/tocall.cpp: In function 'bool DtoLowerMagicIntrinsic(IRState*, FuncDeclaration*, CallExp*, DValue*&)':
/root/project/gen/tocall.cpp:338:37: warning: enumeral and non-enumeral type in conditional expression [-Wextra]
     llvm::SyncScope::ID scope = (e->arguments->dim == 2)
                                 ~~~~~~~~~~~~~~~~~~~~~~~~
                                     ? (*e->arguments)[1]->toInteger()
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                     : llvm::SyncScope::System;
                                     ~~~~~~
```

kinke:
Completely off-topic, but it'd be nice if you could look into the sporadic `ir2obj_cache_pruning2.d` failures plaguing Travis and AppVeyor from time to time: https://travis-ci.org/ldc-developers/ldc/jobs/258653490#L3431

JohanEngelen:
`ir2obj_cache_pruning2.d` failures : #2249

JohanEngelen:
I would like to turn on `-Werror` only for (a selection of) CI systems. Those compilers don't update that often I presume, but let's try? (We shouldn't enable it by default in the CMake script of course.)

gen/dcompute/targetCUDA.cpp
```diff
@@ -47,7 +47,7 @@ class TargetCUDA : public DComputeTarget {
     // sm version?
   }
   void setGTargetMachine() override {
-    char buf[8];
+    char buf[16];
```
klickverbot: Can we please just get rid of code like that? There is no reason to use C string manipulation here, especially for something that happens once.

JohanEngelen: yeah.
working on it.

klickverbot: (The PR is obviously fine as it is.)

JohanEngelen: fixed

PR #2249, opened 2017-07-29
Make cache pruning test more robust by sleeping longer between invokes.

JohanEngelen:
I think there is a spurious case where things happen right before / after second elapses. This should fix that.

@kinke Hope this fixes the spurious failures like https://travis-ci.org/ldc-developers/ldc/jobs/258653490#L3431

### DIFF
--- a/tests/linking/ir2obj_cache_pruning2.d
+++ b/tests/linking/ir2obj_cache_pruning2.d
@@ -33,6 +33,6 @@ void main()
     {
         // Sleep for 2 seconds, so we are sure that the cache object file timestamps are "aging".
         import core.thread;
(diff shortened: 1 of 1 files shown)


tests/linking/ir2obj_cache_pruning2.d
```diff
@@ -33,6 +33,6 @@ void main()
     {
         // Sleep for 2 seconds, so we are sure that the cache object file timestamps are "aging".
```
klickverbot: Comment is now outdated.

JohanEngelen: urf..

PR #2251, opened 2017-07-29
CircleCI: Adapt LLVM APT repo

kinke:


kinke:
They ([LLVM](http://apt.llvm.org/)) provide 3 Debian branches now, but exactly the one we'd need (Stretch) appears to be broken (and not updated) for nearly a week now, so switching back to Debian unstable, which seems to work (most of the time, some 404 errors as usual...).

PR #2252, opened 2017-07-30
Merge 2.075.1 front-end and stdlibs

kinke:
Just the quick first version which can be successfully built (on Win64 at least) after ~2.5 hours of work. :)


joakim-noah:
Built this branch and ran the stdlib tests natively on Android/ARM, had two issues:
* The tests for `std.uri` don't compile because it can't [implicitly convert `char[]` to `string`](https://github.com/dlang/phobos/blob/master/std/uri.d#L322), reverting the part of dlang/phobos#5409 that amended `std.uri.decode` got it to build again.  Strangely, the build error is reproducible on linux/x64 if I try to build the 2.075.0 version of `std.uri`'s tests using ldc 1.3, but not with dmd 2.073-5.  `decodeComponent` should have the same issue, but it's not tested, so no complaint.
* [`std.stdio` fails in this forked block](https://github.com/dlang/phobos/blob/master/std/stdio.d#L1363) if I run all the tests by simply invoking `./runtime/phobos2-test-runner`, but not if I just run that module with `./runtime/phobos2-test-runner std.stdio`.  This suggests some codegen/linking bug or an environment issue, but since that test has worked for years, not sure.  Maybe the recent switch from static constructors to lazy initialization, ie dlang/phobos#5421?  I'll look into this more.

I'm currently running the dmd-testsuite, will update with the results.

joakim-noah:
The following tests now fail in the dmd-testsuite on Android/ARM:
* runnable/sdtor
* runnable/ldc_github_420
* compilable/test17502
* fail_compilation/diag6373
* fail_compilation/ice13259
* fail_compilation/fail13902

kinke:
> The tests for std.uri don't compile because it can't implicitly convert char[] to string, reverting the part of dlang/phobos#5409 that amended std.uri.decode got it to build again. Strangely, the build error is reproducible on linux/x64 if I try to build the 2.075.0 version of std.uri's tests using ldc 1.3, but not with dmd 2.073-5.

DMD 2.074.0 errors out for this minimized testcase derived from failing https://github.com/ldc-developers/phobos/blob/ldc-merge-2.075/std/uri.d#L324...
```
string foo()
{
    char[] r;
    return r;
}
```

kinke:
But templatizing makes it work for DMD and LDC...

```D
string foo()()
{
    char[] r = "123".dup;
    return r;
}

int main()
{
    auto r = foo();
    return cast(int) r.length;
}
```

ZombineDev:
@kinke that's by design - mutable result of pure functions can be converted implicitly to immutable. Templating the function allows the compiler to infer purity.

kinke:
Thanks Petar. So I guess this is an additional check performed by LDC which needs to take this special case of legal implicit conversion into account.

I lost the colors in the console on Windows, so I probably screwed up somewhere.

ZombineDev:
@kinke I'm really confused as to why ldc differs from dmd here, as this feature has been in the language for quite a while and much more unittests in phobos should have been failing by now if ldc never supported this feature. For reference:
* http://forum.dlang.org/post/k7j1ta$2kv8$1@digitalmars.com
* https://issues.dlang.org/show_bug.cgi?id=8993
* https://github.com/dlang/dmd/pull/1700
* https://github.com/dlang/dmd/pull/3319
* https://wiki.dlang.org/DIP29 (perhaps should be marked as implemented?)

kinke:
This is the stack trace for the (front-end) error emitted for `-c -unittest ...\uri.d`:
```
ddmd.errors.verrorPrint(const ddmd.globals.Loc * p2, int p1, const char * ap, const char * format, char * header, const char * headerColor, const char * headerColor) Line 148	D
ddmd.expression.Expression.error(const char * format) Line 2643	D
ddmd.dcast.implicitCastTo.ImplicitCastTo.visit(ddmd.expression.Expression *) Line 104	D
ddmd.visitor.Visitor.visit(ddmd.expression.SymbolExp * e) Line 875	D
ddmd.visitor.Visitor.visit(ddmd.expression.VarExp * e) Line 885	D
ddmd.expression.VarExp.accept(ddmd.visitor.Visitor * v) Line 6711	D
ddmd.dcast.implicitCastTo(ddmd.expression.Expression * e, ddmd.dscope.Scope * sc, ddmd.mtype.Type * t) Line 167	D
ddmd.func.FuncDeclaration.semantic3(ddmd.dscope.Scope *) Line 3785	D
ddmd.dtemplate.TemplateInstance.semantic3(ddmd.dscope.Scope *) Line 6823	D
ddmd.dtemplate.TemplateInstance.trySemantic3(ddmd.dscope.Scope * sc2) Line 8438	D
ddmd.dtemplate.TemplateInstance.semantic(ddmd.dscope.Scope * sc, ddmd.root.array.Array!(ddmd.expression.Expression).Array * fargs) Line 9055	D
ddmd.dtemplate.functionResolve(ddmd.declaration.Match * fargs, ddmd.dsymbol.Dsymbol * tthis, ddmd.globals.Loc tiargs, ddmd.dscope.Scope * sc, ddmd.root.array.Array!(ddmd.root.rootobject.RootObject).Array * loc, ddmd.mtype.Type * dstart, ddmd.root.array.Array!(ddmd.expression.Expression).Array * dstart) Line 2915	D
ddmd.func.resolveFuncCall(ddmd.globals.Loc loc, ddmd.dscope.Scope * sc, ddmd.dsymbol.Dsymbol * s, ddmd.root.array.Array!(ddmd.root.rootobject.RootObject).Array * s, ddmd.mtype.Type * tiargs, ddmd.root.array.Array!(ddmd.expression.Expression).Array * tiargs, int tthis) Line 4305	D
ddmd.expression.CallExp.semantic(ddmd.dscope.Scope * sc) Line 10338	D
ddmd.expression.AssignExp.semantic(ddmd.dscope.Scope * sc) Line 12970	D
ddmd.statementsem.StatementSemanticVisitor.visit(ddmd.statement.ExpStatement *) Line 99	D
ddmd.statement.ExpStatement.accept(ddmd.visitor.Visitor * v) Line 709	D
ddmd.statementsem.StatementSemanticVisitor.visit(ddmd.statement.CompoundStatement *) Line 3581	D
ddmd.statement.CompoundStatement.accept(ddmd.visitor.Visitor * v) Line 896	D
ddmd.func.FuncDeclaration.semantic3(ddmd.dscope.Scope *) Line 1604	D
ddmd.dmodule.Module.semantic3(ddmd.dscope.Scope *) Line 1182	D
ddmd.mars.mars_mainBody(ddmd.root.array.Array!(const(char)*).Array * files, ddmd.root.array.Array!(const(char)*).Array * files) Line 1620	D
cppmain(int argc, char * * argv) Line 1069	C++
D main(...) Line 36	D
```

No obvious LDC specifics in there, so no idea why we diverge here.

kinke:
Okay, it's just that we have our own LDC-specific `core.stdc.stdlib.alloca`, which isn't pure yet as the upstream version... ;)

kinke:
Looking pretty good already, interpreting the [Travis LLVM 4.0.1 Linux x64 job](https://travis-ci.org/ldc-developers/ldc/jobs/260698757):

* All LDC and stdlib unittests are green.
* Single failing lit-test `codegen/vector_init.d`: vector related.
* dmd-testsuite failures:
  * vector related: `runnable/ldc_github_420.d`, `runnable/testargtypes.d`
  * `compilable/test17502.d`: possibly vector related
  * `runnable/sdtor.d`
  * `runnable/traits_getPointerBitmap.d`
  * `fail_compilation/{diag6373,ice13259}.d`: missing backticks in error msg

Note that I simply allowed all vector ops in new `Target::isVectorOpSupported()` for now.

kinke:
Update, remaining failures based on [Semaphore](https://semaphoreci.com/ldc-developers/ldc/branches/merge-2-075/builds/11) (with enabled assertions; Travis only shows 2 remaining dmd-testsuite failures) and my local tests on Win64:

* `runnable/sdtor.d(2842): Assertion failure`
* `runnable/test12.d`: LLVM assertion `/build/llvm/include/llvm/CodeGen/SelectionDAGNodes.h:948: llvm::SDValue::SDValue(llvm::SDNode*, unsigned int): Assertion '(!Node || !ResNo || ResNo < Node->getNumValues()) && "Invalid result number for the given node!"' failed.`
* `compilable/test17502.d`: LDC assertion `../gen/llvmhelpers.cpp:1845: DValue* makeVarDValue(Type*, VarDeclaration*, llvm::Value*): Assertion `isIrVarCreated(vd) && "Variable not resolved."' failed.`
* `compilable/test17352.d`: LLVM assertion `/build/llvm/lib/CodeGen/AsmPrinter/DwarfDebug.cpp:1213: virtual void llvm::DwarfDebug::endFunction(const llvm::MachineFunction*): Assertion '!FnScope || SP == FnScope->getScopeNode()' failed.`
* Phobos:
  * `std.datetime.stopwatch` features brittle unittests for a benchmarking function. With enabled optimizations, that code may run too fast (duration = 0), probably because it's all optimized away. Happens for Semaphore (Travis is too slow of course => no problems :D) and on my Win64 box.
  * Windows-specific: `std.datetime.timezone` fails with enabled optimizations on my Win64 box, due to an apparently invalid registry key handle when enumerating the time zones.

kinke:
`runnable/sdtor.d:2842` apparently fails due to a missed opportunity for in-place construction.

`compilable/test17502.d` fails due to an implicit `__result` variable in an out contract (no `out (result) { ... }`, but `out { assert(__result); }`) with the method return type being `auto`. That special result variable seems to be wrongly detected as nested variable, while it is/should actually be passed as by-ref argument when calling the contract.

gen/toir.cpp
```diff
@@ -2733,6 +2733,22 @@ bool toInPlaceConstruction(DLValue *lhs, Expression *rhs) {
       ToElemVisitor::call(gIR, ce, DtoLVal(lhs));
       return true;
     }
+    // DMD issue 17457: detect structliteral.ctor(args)
+    else if (ce->e1->op == TOKdotvar) {
```
klickverbot: `else` is meaningless as the previous `if` ends in `return` anyway. Ditching it also makes the comment indentation less awkward.

kinke: Glad these commits get actually looked at. ;) - Will address that tiny nit.

Any chance you could help out with the contract thing?

kinke:
Something really odd is going on here. `compilable/test17502.d` works fine here on Win64, when cross-compiling to Linux x64 too. But it doesn't work on a Linux host...

kinke:
@rainers: I'm getting linker errors with druntime 2.075 due to missing symbols `_{DP,TP}_{beg,end}` with latest VS 2017 15.3.5. `/NOOPTTLS` doesn't help, Google isn't helping. Does DMD emit these symbols?

kinke:
Progress => update:

(Non-critical) LLVM debuginfo assertion when compiling with `-g -enable-inlining` (from `runnable/xtest46.d`):
```
template map(fun...)
{
    auto map(double[] r)
    {
        struct Result
        {
            this(double[] input)
            {
            }
        }

        return Result(r);
    }
}


void test139()
{
    double[] x;
    alias typeof(map!"a"(x)) T;
    T a = void;
    auto b = map!"a"(x);
    auto c = [map!"a"(x)];
    T[3] d = void;
}
```

That's the only remaining issue on Linux (in my VM; Travis is green due to disabled assertions); CircleCI chokes on some std.process unittest (no exception thrown when starting a process with a non-existing working dir or something like that), Semaphore seems to have memory issues...
On OSX, there are 2 more issues. Objective-C support was refactored, and I probably introduced some regressions while adapting our code; a bunch of dmd-testsuite Objective-C tests fail. Additionally, a C++ mangling dmd-testsuite test was changed and now expects 2 leading underscores, while LDC apparently produces a single one (as for Linux).
For Windows, there are the mentioned druntime linker errors which I worked around locally for now. There's one Windows-specific failing unittest left, `std.datetime.timezone` (release only).

rainers:
> I'm getting linker errors with druntime 2.075 due to missing symbols _{DP,TP}_{beg,end} with latest VS 2017 15.3.5. /NOOPTTLS doesn't help, Google isn't helping. Does DMD emit these symbols?

These are used for precise scanning of the DATA/TLS segments with --DRT-scanDataSeg=precise. (see rt.sections_win64.initSections). The dmd implementation for COFF is here: https://github.com/dlang/dmd/blob/master/src/ddmd/backend/mscoffobj.c#L2457, which is invoked from the glue layer.

This needs something similar as the array of module info pointers. i.e. a segment in the linker output embraced by labels _{D,T}P_beg and _{D,T}P_end.

kinke:
The 2 OSX-specific issues should be fixed now, except for a small thingy wrt. mangling of naked C++ functions, will fix that tonight.

gen/target.cpp
```diff
@@ -226,7 +226,7 @@ Expression *Target::paintAsType(Expression *e, Type *type) {
     break;
 
   default:
-    llvm_unreachable("Unsupported source type");
+    assert(0);
```
JohanEngelen: why this change?

kinke: That's from the revert...

kinke:
I finally found the reason for the `std.datetime.timezone` unittest failure on Win64: it's the GC collecting a `std.windows.registry.Key` object while it's being iterated over.

kinke:
Hmm not sure if this isn't a more general issue.

What happens is that firstly the `baseKey` alloca is gotten rid of. It's only used for `.keyNames()`, which returns a new `std.windows.registry.KeyNameSequence` object storing a reference to its parent `baseKey`. The `keyNames()` result isn't stored in an alloca either; it's only required for `.opApply()`.
In (optimized) Phobos, [KeyNameSequence.opApply()](https://github.com/ldc-developers/phobos/blob/ldc/std/windows/registry.d#L1416) calls a little helper, `regProcessNthKey()`, which performs the iteration, but only gets the raw Windows HKEY handle, without its owning `Key` object (`baseKey`).
In the end, when GC is triggered during iteration, there are apparently no live `baseKey` references anywhere. The issue can be worked around by passing the owning `Key` instead of `HKEY` to `regProcessNthKey()`. I haven't looked at the `std.windows.registry` IR yet; I'd assume that a method `KeyNameSequence.opApply()` needs to keep `this` alive though to prevent issues like this.

kinke:
@klickverbot: Need your help here. Do we rely on all (really) live object references being somewhere in reachable memory or in a register at the time GC is run? I'd think that we had far more bug reports in that case wrt. optimized builds...

kinke:
Well okay here's another specialty of the issue at hand - the `Key` object itself isn't needed anymore and collecting would be fine if it wasn't for its destructor/finalizer, which closes the HKEY handle - only because of that, the lifetime needs to be expanded, so a Phobos bug does indeed make sense.

kinke:
The workaround works because passing the `Key` owner to `regProcessNthKey()` and referencing that in its [lambda](https://github.com/ldc-developers/phobos/blob/ldc/std/windows/registry.d#L750) leads to a reference to the `Key` object (which is indeed on the GC heap) in the nested frame of `regProcessNthKey`, which is guaranteed to be in memory.

kinke:
Exactly.

kinke:
> From what I gather, the Key is on the GC heap

In my testcase yes; in the failing unittest, it's actually a `scope Key`. It ought to work for an arbitrary object though.

kinke:
It's initialized with `scope baseKey = Registry.localMachine.getKey("...");`, with `Key.getKey()` returning a new `Key` object. So I guess `scope` doesn't have any effect here - we just use the returned ref for the vtable lookup of keyNames() and then pass it as `this`.

kinke:
That would indeed keep the object alive and hide this bug in the unittest. ;) - I just checked DMD's assembly without inlining, and it's apparently not calling the dtor at the end either.

rainers:
I think the registry classes are meant to be used with `scope`, otherwise the keys might linger forever.

`scope` seems to have silently changed its meaning with dmd 2.075, though. While 2.074 still generates the _d_newclass/_d_delclass pair, 2.075 omits the _d_delclass call. So I suspect this is a front end regression.

ZombineDev:
The `scope` storage class has two distinct effects when applied to classes / class variables, in addition to the meaning that DIP1000 patches - disallowing escaping reference to an object by assigning the reference to non-`scope` variables:

1. When a scope class variable is initialized (!= assigned) with the result of a `new` expression, the class object gets allocated on the stack and just like structs its destructor gets called at the end of the scope. A bit confusing fact (at least for people expecting them to behave like C++ class values) is that `scope` class variables are still reference types, meaning that only when a variable is initialized to a new object, a hidden temporary is created on the stack and its address is assigned to the variable. Otherwise they behave like regular class variables:
```D
{
    // initialization
    scope SomeClass a = new SomeClass();

    {
        // assignment
        scope MyClass b = a;
        // same as above: `c`'s type is inferred to be `MyClass`
        // and with `-dip1000` flag `c`'s storage class will be inferred as scope,
        // but `scope` in this context doesn't affect the RAII semantics -
        // you just can't assign the reference to a non-`scope` variable, like globals
        auto c = a;
    } // no dtor is called

} // `a`'s dtor is called
```

2. `scope` classes (as in `scope class Range { /* ... */ }`) can (at least should?) only be stack-allocated.

The two properties listed above haven't changed in recent releases at all - DIP1000 only now makes the `scope` class technique more safe. Before DIP1000, use of scope classes was discouraged in favor of `std.typecons.scoped` because they weren't safe, but also because you couldn't allocate them in-place when used as member fields, among other inconsistencies. However given how broadly they are used in some projects (like Sociomantic's Ocean library and ddmd), realistically speaking I don't think they will be deprecated.

About `scope baseKey = Registry.localMachine.getKey("...");`:
`baseKey` most certainly should **not** be destroyed when it goes out of scope - its just a plain reference to something allocated on the heap.

rainers:
> About scope baseKey = Registry.localMachine.getKey("...");:
> baseKey most certainly should not be destroyed when it goes out of scope - its just a plain reference to something allocated on the heap.

It used to get destroyed in 2.074. I've reduced it to
```
import core.stdc.stdio;

void foo()
{
	scope baseKey = Key.getKey();
}

int main(string[] argv)
{
	printf("main init\n");
	foo();
	printf("main exit\n");
	return 0;
}

class Key
{
private:
	this() { printf("ctor\n"); }
    ~this() { printf("dtor\n"); }

	static Key getKey() { return new Key(); }
}
```
This prints
```
main init
ctor
dtor
main exit
```
with dmd 2.074 and 
```
main init
ctor
main exit
dtor
```
with dmd 2.075.1, i.e. the class instance assigned to the scope variable is destroyed in the final collection, not in foo's scope. The 2.074 version is according to spec IMO: "... the destructor for an object is automatically called when the reference to it goes out of scope." It doesn't matter whether the obect was created with a new expression or by a factory function.

ZombineDev:
According to https://dlang.org/spec/attribute.html#scope:
> [...] scope function return values are not allowed.

ZombineDev:
@rainers the behavior change that you're observing was a bug fix. Let my dig-up the pull request...

rainers:
I suspect it changed with https://github.com/dlang/dmd/pull/6826

ZombineDev:
Here it is: https://github.com/dlang/dmd/pull/6826

rainers:
> [...] scope function return values are not allowed.

I think this refers to something like:
```
auto fun()
{
    scope key = new Key;
    return key;
}
```
not to assigning a function return value to a scope variable.

ZombineDev:
So yes, your example was changed for the better - the destructor shouldn't have been called. To bad that there may have been code that replied on the wrong behavior.

ZombineDev:
> I think this refers to something like [..]

Yes you're right about this, but I thought the wording in the spec implied that you can't return class values (like in C++), i.e. there're only class references.

ZombineDev:
I haven't looked at the Phobos unittest, but suspect (from the discussion here) that it may be wrong. In which case, I think the best course of action is to fix upstream phobos and for you to cherry-pick the fix for your 1.5/2.075.1 release.

rainers:
> So yes, your example was changed for the better - the destructor shouldn't have been called.

I don't think it's better. The destructor should be called, but escape analysis should prohibit misusage in safe code. 
BTW: the `fun` example actually compiles, even if marked safe. I think it should produce an error.

ZombineDev:
> BTW: the fun example actually compiles, even if marked safe. I think it should produce an error.

Yes, it should not compile.

> I don't think it's better. The destructor should be called, but escape analysis should prohibit misusage in safe code.

No, in general you can't know if the destructor should be called. Consider the following:
```D
class Connection
{
    this(string address) { /* connect to address */ }
    ~this() { /* disconnect */ }

    size_t write(in ubyte[] data);
    size_t read(ubyte[] buf);
}

class ConnectionPool
{
    Connection[string] cache;
    Connection get(string addr)
    {
        if (auto p = addr in cache)
            return *p;

         // assume that the memory is manually managed, not like the code written here for illustration purposes
        auto c = new Connection(addr);
        cache[addr] = c;
        return c;
    }

    ~this() { delete cache; /* assume manual destruction and deallocation */ }
}

RefCounted!ConnectionPool cp;

void clienCode()
{
    scope c = cp.get("some address");
    ubyte[] data = ...;
    c.write(data);
}

clienCode();
clienCode(); // would crash if Connection.~this() was called at the end of the scope.
```

rainers:
> scope c = cp.get("some address");

safe code should not allow this unless the function is a pure factory function (other conditions might work, too).

ZombineDev:
I disagree, the code is perfectly correct and also code that DIP1000 is meant to support.

rainers:
AFAICT DIP1000 is not consistent with the current meaning of scope. For example, https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md#scope-function-returns contradicts the spec https://dlang.org/spec/attribute.html#scope regarding scope on function return values.

Walter also added a workaround to allow scope delegates to keep their existing semantic with -dip1000.

ZombineDev:
> For example, https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md#scope-function-returns contradicts the spec https://dlang.org/spec/attribute.html#scope regarding scope on function return values.

How so? That part of the spec talks only about scope classes. With or without DIP1000 you can't return the address of a local variable. `scope return` is about allowing to return function parameters, which is safe as long as the caller knows that an argument passed to `scope return` parameter may be returned.

rainers:
DIP1000 is not the spec yet. Regarding classes the spec is pretty explicit that "scope is used to **guarantee** cleanup."

ZombineDev:
> Walter also added a workaround to allow scope delegates to keep their existing semantic with -dip1000.

I'm not sure if `scope` delegate parameters are now working safely, but the idea is that the closure environment (context) is captured and allocated on the stack of the caller and the callee receives essentially a pointer of local variable, which is safe as long as the pointer is not escaped.

The stack-allocation is a transparent optimization which should not affect the behavior of the caller or callee. Whether it happens or not can be thought of as implementation-defined behavior. DIP1000 just now enforces the safety.

> DIP1000 is not the spec yet

De jure it is not, de facto it is. The reason Walter did not put it there yet is that he wanted to right down all the changes to `@safe` as a coherent whole. I disagree that this is right approach, as it leads to confusion about the current meaning of `@safe` and `scope` in the meantime (several releases!) but that's where we're at today.

> Regarding classes the spec is pretty explicit that "scope is used to guarantee cleanup."

That's about initialization - `scope Class c = new Class()` - this does not affect assignment, otherwise you'll end up with use after free, as horrid behavior that this test case used to exhibit: https://github.com/dlang/dmd/pull/6826/files#diff-1147486c08c85586a3015dd70e5b3c9aR14

rainers:
> DIP1000 just now enforces the safety.

DIP1000 changes the `scope` from the delegate to the variable holding the delegate, that's why a workaround was necessary: https://github.com/dlang/dmd/blob/master/src/ddmd/mtype.d#L6222

> De jure it is not, de facto it is.

DIP1000 might be the future, but it is hidden behind -dip1000 which probably only few people use. It's still too broken, it can even causes code generation errors, e.g. https://issues.dlang.org/show_bug.cgi?id=17449, so I wouldn't call it the de facto spec yet.

> That's about initialization - scope Class c = new Class() - this does not affect assignment, 

There is nothing in the current spec that says so.

> otherwise you'll end up with use after free, as horrid behavior that this test case used to exhibit: https://github.com/dlang/dmd/pull/6826/files#diff-1147486c08c85586a3015dd70e5b3c9aR14

Unsafe, but not necessarily a bug. You only add `scope` to more variables because of DIP1000, though.

I think changing semantic without a changelog entry is a regression, so I'll file a bug report with the example above. Let's hear what the scope experts say...

rainers:
> I'll file a bug report

https://issues.dlang.org/show_bug.cgi?id=17869

BTW: the code in https://github.com/ldc-developers/ldc/pull/2252#issuecomment-333364175 does fail to compile with -dip1000.

kinke:
After https://github.com/dlang/phobos/pull/5749, we'll only have to take care of an apparent LLVM debuginfo inlining bug to be fully green. I compiled the [earlier code snippet](https://github.com/ldc-developers/ldc/pull/2252#issuecomment-331678565) with current LDC master and this PR using a common LLVM 5.0. LDC master works with `-g -enable-inlining`, this PR asserts:

```
/build/llvm/include/llvm/CodeGen/SelectionDAGNodes.h:948: llvm::SDValue::SDValue(llvm::SDNode*, unsigned int): Assertion `(!Node || !ResNo || ResNo < Node->getNumValues()) && "Invalid result number for the given node!"' failed.
```

The base IR (without inlining) only differs in one function which has changed due to https://github.com/ldc-developers/ldc/pull/2252/commits/befd05eb56591f5113d76d3c8f5b07840e66eb20 (in-place construction):

```diff
 define weak_odr void @_D7xtest4617__T3mapVAyaa1_61Z3mapFNaNbNiNfAdZS7xtest4617__T3mapVAyaa1_61Z3mapFAdZ6Result(%"xtest46.map!\22a\22.map.Result"* noalias sret align 8 %.sret_arg, { i64, double* } %r_arg) #0 comdat !dbg !46 {
   %r = alloca { i64, double* }, align 8           ; [#uses = 2, size/byte = 16]
-  %.structliteral = alloca %"xtest46.map!\22a\22.map.Result", align 8 ; [#uses = 3, size/byte = 8]
   store { i64, double* } %r_arg, { i64, double* }* %r, !dbg !51 ; [debug line = 3:10]
   call void @llvm.dbg.declare(metadata { i64, double* }* %r, metadata !50, metadata !36), !dbg !51 ; [debug line = 3:10] [debug variable = r]
-  %1 = getelementptr inbounds %"xtest46.map!\22a\22.map.Result", %"xtest46.map!\22a\22.map.Result"* %.structliteral, i32 0, i32 0 ; [#uses = 1, type = i8**]
+  %1 = getelementptr inbounds %"xtest46.map!\22a\22.map.Result", %"xtest46.map!\22a\22.map.Result"* %.sret_arg, i32 0, i32 0 ; [#uses = 1, type = i8**]
   store i8* null, i8** %1, !dbg !52               ; [debug line = 12:9]
   %2 = load { i64, double* }, { i64, double* }* %r, !dbg !52 ; [#uses = 1] [debug line = 12:9]
-  %3 = call %"xtest46.map!\22a\22.map.Result"* @_D7xtest4617__T3mapVAyaa1_61Z3mapFAdZ6Result6__ctorMFNaNbNcNiNfAdZS7xtest4617__T3mapVAyaa1_61Z3mapFAdZ6Result(%"xtest46.map!\22a\22.map.Result"* nonnull returned %.structliteral, { i64, double* } %2) #0, !dbg !52 ; [#uses = 0] [debug line = 12:9] [display name = xtest46.map!"a".map.Result.this]
-  %4 = bitcast %"xtest46.map!\22a\22.map.Result"* %.sret_arg to i8*, !dbg !52 ; [#uses = 1] [debug line = 12:9]
-  %5 = bitcast %"xtest46.map!\22a\22.map.Result"* %.structliteral to i8*, !dbg !52 ; [#uses = 1] [debug line = 12:9]
-  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %4, i8* %5, i64 8, i32 1, i1 false), !dbg !52 ; [debug line = 12:9]
+  %3 = call %"xtest46.map!\22a\22.map.Result"* @_D7xtest4617__T3mapVAyaa1_61Z3mapFAdZ6Result6__ctorMFNaNbNcNiNfAdZS7xtest4617__T3mapVAyaa1_61Z3mapFAdZ6Result(%"xtest46.map!\22a\22.map.Result"* nonnull returned %.sret_arg, { i64, double* } %2) #0, !dbg !52 ; [#uses = 0] [debug line = 12:9] [display name = xtest46.map!"a".map.Result.this]
   ret void, !dbg !52                              ; [debug line = 12:9]
 }
```

kinke:
https://bugs.llvm.org/show_bug.cgi?id=34834

kinke:
The debuginfos look okayish; all calls have line infos. What looks wrong/suboptimal is the line info for the implicit LL `ret` of the empty struct ctor (it's set to the line where the definition begins; the closing brace is 2 lines down).

kinke:
I'm leaning towards disabling `runnable/xtest46` for now for the CI systems using an LLVM with enabled assertions (i.e., AppVeyor and Semaphore) in order not to delay this any longer. Objections/better ideas?

kinke:
> to be resolved before the final release

I'm not that optimistic, but we'll see. ;) - I upgraded SemaphoreCI to LDC-LLVM 5.0.0 as well.

kinke:
Dammit, running out of (RAM?) disk space for Semaphore with LLVM 5, even when building with 2 jobs only...

joakim-noah:
Looks like all CIs pass, with the only failure on Travis with llvm 3.7.1 that times out for some reason.  Maybe it's time to merge this into master and let more people bang on it with a beta release?

JohanEngelen:
@kinke Great work!

joakim-noah:
Ran the master stdlib and dmd tests on Android/ARM, only issue is the `std.stdio` failure when running all the tests.  I need to look into that.

joakim-noah:
Rather than step through the `std.stdio` issue with a debugger, I kicked off some test builds on a linux/x64 VPS to see where it may be coming from.  Turns out it comes from upstream, as both dmd 2.074.1 and 2.075.1 show the same failure, though I cannot reproduce with ldc 1.4.  However, I think [Oleg saw it on a RPi 2 with ldc 1.4](http://forum.dlang.org/post/jgtkbgmyrfsledknefvr@forum.dlang.org), so maybe it depends on the toolchain used.

This doesn't surface in the CI for dmd or ldc because they don't just run the phobos test runner with no arguments so it runs all the tests at once, but pass each test to the runner separately and this problem doesn't show up that way.

I also looked at the issue where `phobos2-test-runner` hangs after running all the tests, which I've seen on linux/x64 for some time and Oleg saw on linux/armhf too.  Turns out it goes away if the druntime tests are not run, which I got [by changing this line to link against `libdruntime-ldc.a` instead](https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L811).  There's no reason to include the druntime tests in `phobos2-test-runner`, but we should probably figure out why it hangs before removing them.

Running all the tests from one run of the test runner is a good stress test of running a lot of the stdlib code, and it looks like we have a couple issues that are not exposed by our CI or upstream, because they don't run the test runner that way.

PR #2254, opened 2017-07-30
Update LDC 0.17.x to LLVM 5.0 (2nd try)

JohanEngelen:
@redstar I got `ltsmaster` to build with this PR. But there is a debuginfo LLVM assert when building the debug version of Phobos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,28 @@
+dist: trusty
 language: cpp
 sudo: false
 
 matrix:
   include:
     - os: linux
-      env: LLVM_VERSION=4.0.0
+      env: LLVM_VERSION=4.0.1
     - os: linux
-      env: LLVM_VERSION=3.9.0
+      env: LLVM_VERSION=3.9.1 OPTS="-DBUILD_SHARED_LIBS=ON"
     - os: linux
-      env: LLVM_VERSION=3.8.1
+      env: LLVM_VERSION=3.8.1 OPTS="-DBUILD_SHARED_LIBS=OFF"
     - os: linux
       env: LLVM_VERSION=3.7.1 OPTS="-DMULTILIB=ON"
-    - os: linux
-      env: LLVM_VERSION=3.6.2 OPTS="-DBUILD_SHARED_LIBS=ON"
-    - os: linux
-      env: LLVM_VERSION=3.5.2 OPTS="-DTEST_COVERAGE=ON"
     - os: osx
-      env: LLVM_CONFIG="llvm-config-3.8"
+      env: LLVM_VERSION=4.0.0 OPTS="-DBUILD_SHARED_LIBS=OFF"
   allow_failures:
     #- env: LLVM_VERSION=3.9
 
 cache:
   directories:
-    - llvm-4.0.0
+    - llvm-4.0.1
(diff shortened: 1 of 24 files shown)


joakim-noah:
Tried this PR out on Android/ARM with [the latest 5.0rc1 source release](http://prereleases.llvm.org/5.0.0/#rc1) built for ARM with my emulated TLS patch, all the stdlib tests pass and only one module, `runnable/testabi`, fails in the dmd-testuite, with the following output:
```
core.exception.AssertError@runnable/testabi.d(243): Assertion failure
Test1 inout ssss        Fail
```
So this PR is looking good, but when I tried llvm 5.0 rc1 with 1.4 master, whether natively or cross-compiling, I'm getting assertions and segfaults in ldc when building some stdlib modules with optimizations, usually something like this:
```
[ 48%] Generating objects-unittest/std/algorithm/comparison.o
LLVM ERROR: Cannot select: 0x55d609ae03d0: f64 = bitcast 0x55d6098ac8e0
   0x55d6098ac8e0: i64 = build_pair 0x55d6098acc88, 0x55d609ae0b20
      0x55d6098acc88: i32,ch = CopyFromReg 0x55d609ae8e58, Register:i32 %vreg67
         0x55d6098acf60: i32 = Register %vreg67
      0x55d609ae0b20: i32,ch = CopyFromReg 0x55d6098acc88:1, Register:i32 %vreg68
         0x55d6098ad030: i32 = Register %vreg68 
In function: _D3std9algorithm10comparison18__unittestL2015_49FNaNfZv
```
The debug stdlib test-runner builds compile fine and most tests pass, but anything at `-O2` or higher for certain modules breaks ldc.  I'm unable to reproduce these issues when building the stdlib for linux/x64.  I'll try to track down which optimization pass is now breaking on ARM, let me know if you have any insight.

joakim-noah:
Huh, I couldn't reproduce the issue with `runnable/testabi` by building it with the command shown in the log, but it was reproducible by running the whole testsuite again, so I thought of looking in the setup for that, and sure enough, [it adds an `-O` to `DFLAGS`](https://github.com/ldc-developers/ldc/blob/master/tests/d2/CMakeLists.txt#L52), which isn't output to `LastTest.log`.  So it turns out the issue is that some ABI data is being optimized away by llvm 5.0 now, need to figure out why.

As for the problems with ldc master, it looks like that error is in the `ARM Instruction Selection` codegen pass, which I guess means it was passed IR for which the ARM backend couldn't find an instruction?  I'll keep digging into that and file an issue if needed.

kinke:
> it adds an -O to DFLAGS, which isn't output to LastTest.log

Yep, I only fixed that recently in master. There are still issues when permutations come into play IIRC, I don't think the proper full command line is displayed for those yet (instead, the output for all run permutations is accumulated). Not very handy for `runnable/xtest46.d` with its 32 (used to be 64) permutations and lots of output. ;)

JohanEngelen:
Merge so that it's easier to make LLVM5.0 fixes?

joakim-noah:
Fine by me, only one optimization issue on Android/ARM, which can always be fixed later.

redstar:
Yes, let's merge it.

PR #2256, opened 2017-08-03
Add helper function `append` to runtime cmakescript

JohanEngelen:
The `append` function is missing when the runtime cmakescript is used directly (instead of included by the main LDC cmakescript).
With this PR, the `build-ldc-runtime.sh` works for me straight from my `build/bin` dir and specifying my git checkout as `LDC_SRC_DIR`, nice!

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -12,6 +12,17 @@ if(NOT LDC_EXE)
     endif()
 endif()
 
+# Helper function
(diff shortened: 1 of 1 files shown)


kinke:
I fixed this already in https://github.com/ldc-developers/ldc/pull/2253 superseding the script.

JohanEngelen:
Ah ok, so let's pull that?

Issue #2257, opened 2017-08-03
Error with inline assembly and `-fsanitize=address`


JohanEngelen:
Testcase:
```d
void foo()
{
    uint[4] a;
    asm pure nothrow @nogc {
        mov a, EAX;
        mov a+4, EBX;
    }
}
```

When compiled with `ldc2 -c -fsanitize=address test.d`, we get the error:
```
<inline asm>:1:16: error: unknown token in expression
        movl %ebx, 4+(%rcx)
                      ^
LLVM ERROR: Error parsing inline asm
```

Strangely, when I compile manually to LLVM IR and then use `llc` to obtain assembly output, everything works fine. So I can't reproduce it from LLVM IR onwards.

The inline asm looks like this in LLVM IR:
```llvm
  call void asm sideeffect "movl %eax, $0\0A\09movl %ebx, 4+$1", "=*m,=*m,~{memory}"([4 x i32]* %3, [4 x i32]* %3)
```
without `-fsanitize=address`, the $1 resolves to `ebp` plus an offset; with `-fsanitize=address`, the $1 resolves to a register without any offset and that's invalid (i.e. "4+(%rcx)" is not valid but "4+32(%rcx)" is). At least that's what I think is what's going on.

JohanEngelen:
(btw, this prevents compilation of druntime with `-fsanitize=address` :()

JohanEngelen:
Note: to get equal assembly output from `llc` as from `ldc2`, `llc` must be called with `-O0` explicit.

JohanEngelen:
Testcase without ASan:
```d
void foo()
{
    align(32) uint[4] a;
    asm pure nothrow @nogc {
        mov a+4, EBX;
    }
}
```

JohanEngelen:
LLVM needs some work to support it: https://lists.llvm.org/pipermail/llvm-dev/2017-August/116244.html

just-harry:
There's some inline assembly included by `windows.h` for 32-bit Windows that triggers this bug—which is a real drag, so I've been toying around with the code that translates DMD-style assembly to LLVM inline assembly, and I think I've come up with a bodge that workarounds the issue: https://github.com/ldc-developers/ldc/pull/4632

PR #2260, opened 2017-08-03
Fix inline assembly string with offset calculation.

JohanEngelen:
Resolves issue #2257

See https://lists.llvm.org/pipermail/llvm-dev/2017-August/116158.html

### DIFF
--- a/gen/asm-x86.h
+++ b/gen/asm-x86.h
@@ -21,6 +21,7 @@
 #include "id.h"
 #include "ldcbindings.h"
 #include "gen/llvmhelpers.h" // printLabelName
+#include "llvm/Target/TargetMachine.h"
 #include <cctype>
 
 #ifndef ASM_X86_64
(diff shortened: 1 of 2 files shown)


JohanEngelen:
Reproduces with:
`ldc2 -c -output-s`
```d
__gshared uint[2] constants = [0x0001_0203, 0x0001_0203];
void offconst()
{
    asm pure nothrow @nogc {
        movdqa XMM7, [constants + 4];
    }
}
```

JohanEngelen:
This is not a good fix. It also breaks this code:
```d
void offfunc()
{
    asm pure nothrow @nogc {
        movdqa XMM7, [offfunc + 4];
    }
}
```

JohanEngelen:
I think this is an LLVM bug, and I don't know how to work around it in all cases.
https://lists.llvm.org/pipermail/llvm-dev/2017-August/116220.html

JohanEngelen:
LLVM bug, so we shouldn't fix this here. I think the workaround will be very very hard to pull off. Let's not waste time on that. For now, I will just blacklist the particular function with ASan.

PR #2261, opened 2017-08-03
Add the -fsanitize-blacklist option for the sanitizers.

JohanEngelen:
Having fun with ASan :)
Blacklisting is necessary for some druntime functions, and LLVM/Clang already provides just the right functionality.

Will add the actual needed blacklist in a later PR.

### DIFF
--- a/driver/cl_options_sanitizers.cpp
+++ b/driver/cl_options_sanitizers.cpp
@@ -35,6 +35,8 @@ cl::list<std::string> fSanitizeBlacklist(
     cl::desc("Add <file> to the blacklist files for the sanitizers."),
     cl::value_desc("file"));
 
+std::unique_ptr<llvm::SpecialCaseList> sanitizerBlacklist;
+
 #ifdef ENABLE_COVERAGE_SANITIZER
 cl::list<std::string> fSanitizeCoverage(
     "fsanitize-coverage", cl::CommaSeparated,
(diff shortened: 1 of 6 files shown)


JohanEngelen:
Ah ofcourse. Windows mangling.

JohanEngelen:
Any concerns?

kinke:
Looks good after a quick glance, just one thing: we shouldn't use the IR function name, as the necessity to fix up the testcase for Win32 clearly shows. ;)

JohanEngelen:
I thought about using the "normal" function name, without the \01 added for Win32, but then I thought that because of different underscore conventions, the start of the mangled name is already different anyway so you'd have to use a * at the start already ("*_D...."). The blacklist for Clang is using mangled names, and I don't want to deviate from that.
What I can do is to shorten the StringRef by one if we detect a '\01' at the start of the string?

JohanEngelen:
(eager to get this in btw, it's annoying me with asan work)

JohanEngelen:
(also, I could get rid of the extra _ added for win32.......)

kinke:
I meant just using the regular D mangle, not the ugly stuff we have to add to make LLVM use that name for the object file.

JohanEngelen:
Yep indeed, I am already reworking things.  function.cpp should just check whether the FuncDecl is in the blacklist, without concerning itself about mangling/blacklist details. I've changed the lookup to use mangleExact (whose returned string is cached).

JohanEngelen:
(I wanted things to be as close as possible to the real mangled name in the object file, including hashing. But... I can think of other use cases where blacklisting using the non-hashed name makes more sense. So I'm happy with mangleExact too)

JohanEngelen:
The by-reference was intentional: the interface does not allow passing null pointers.
Inability of D code to bind to C++ class references leading to bad code :(

JohanEngelen:
I'll merge when Travis is green.

kinke:
> the interface does not allow passing null pointers.

Refs don't protect from passing nulls under the hood, the segfault is just a `FuncDeclaration *fd = nullptr; func(*fd)` away.

PR #2264, opened 2017-08-06
Don't use -m32/-m64 for external ARM assembler

kinke:
And merge the two out-of-sync copies for adding gcc target flags. The flag was omitted for gcc as linker already, but not for gcc as external assembler.

### DIFF
--- a/driver/tool.cpp
+++ b/driver/tool.cpp
@@ -74,22 +74,21 @@ std::string getGcc() {
 ////////////////////////////////////////////////////////////////////////////////
 
 void appendTargetArgsForGcc(std::vector<std::string> &args) {
+  using llvm::Triple;
+
+  const auto &triple = *global.params.targetTriple;
+  const auto arch64 = triple.get64BitArchVariant().getArch();
+  const auto arch32 = triple.get32BitArchVariant().getArch();
+
   // Only specify -m32/-m64 for architectures where the two variants actually
   // exist (as e.g. the GCC ARM toolchain doesn't recognize the switches).
-  const auto archVariant64 =
-      global.params.targetTriple->get64BitArchVariant().getArch();
-  const auto archVariant32 =
-      global.params.targetTriple->get32BitArchVariant().getArch();
-  if (archVariant64 == llvm::Triple::UnknownArch ||
-      archVariant32 == llvm::Triple::UnknownArch ||
-      archVariant64 == llvm::Triple::aarch64 ||
-      archVariant64 == llvm::Triple::aarch64_be) {
+  if (arch64 == Triple::UnknownArch || arch32 == Triple::UnknownArch ||
+      arch64 == Triple::aarch64 || arch64 == Triple::aarch64_be) {
     return;
   }
 
   // MIPS does not have -m32/-m64 but requires -mabi=.
-  if (archVariant64 == llvm::Triple::mips64 ||
-      archVariant64 == llvm::Triple::mips64el) {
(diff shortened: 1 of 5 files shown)


driver/tool.cpp
```diff
@@ -70,6 +71,52 @@ std::string getGcc() {
...
+  // exist (as e.g. the GCC ARM toolchain doesn't recognize the switches).
+  const auto archVariant64 =
+      global.params.targetTriple->get64BitArchVariant().getArch();
+  const auto archVariant32 =
+      global.params.targetTriple->get32BitArchVariant().getArch();
+  if (archVariant64 == llvm::Triple::UnknownArch ||
+      archVariant32 == llvm::Triple::UnknownArch ||
+      archVariant64 == llvm::Triple::aarch64 ||
+      archVariant64 == llvm::Triple::aarch64_be) {
+    return;
```
JohanEngelen: for easier reading of the control flow (at least for me), it's nicer to either do early returns like here (so also for MIPS), or to have one big if-elseif chain (so merge this one with the next).

JohanEngelen:
Maybe some `-v` tests?

kinke:
[Updated.]

JohanEngelen:
Cool, I guess without having the actual toolchain, it's annoying to add testing for -m32 being added or not. (we don't have the capability of doing a fake link, or outputting the linker line without executing it)

JohanEngelen:
LGTM.

(I restarted the osx travis builds)

kinke:
> I guess without having the actual toolchain, it's annoying to add testing for -m32 being added or not.

`-m32` / whatever isn't worth being tested. What's important is that an object can be externally assembled and linked to an executable. So this would test the -mabi thing for MIPS too and would have shown issues on ARM earlier.

JohanEngelen:
Because we don't have CI for ARM and MIPS and most of our devs are not working on such machines, we should try our best to test ARM/MIPS/... particulars on X86 machines by specifying -mtriple=... so that we can see things regress.

The test you added is ok and should have been added when the -no-integrated-as flag was first added. But the test is not testing whether `-no-integrated-as` does what it promises (use external assembler). The test only tests that LDC accepts the flag and compiles+links without erroring. The flag could just be ignored and the test would pass. (The dependency on Phobos is also not needed.)

I'm not asking you to fix all this right now. But keep it in mind next time.

kinke:
As I said, I don't see any reason for 'fixing' this, I'm convinced this generic way is the proper way of testing this (okay, checking that the file is actually assembled externally wouldn't hurt). A handful of selected tests for specific targets which have proven to be faulty in the past can never make up for the [current] lack of CI boxes and their full test converage. I think it's much more important to be able to run the testsuite on a particular system and get notified about hopefully the majority of issues which still need to be ironed out, instead of finding out about them one by one after trying more advanced features/codebases, which is nothing but frustrating.
[The dependency on Phobos was deliberate to make sure the produced object is compatible with both druntime and Phobos libs (architecture-wise), but druntime alone would have sufficed in retrospective, that's true.]

kinke:
[I should have made this point clearer earlier too: just testing that there's the `-m32/-m64` arg on x86, that it's NOT there for ARM targets, and that `-mabi` is used for MIPS targets instead isn't a meaningful test in the sense that it doesn't test the actual goal; it just tests our assumptions that these cmdline flags are required for those targets.]

Looking at it again, the reason for dragging in Phobos wasn't the architecture compatibility, but me wanting to feed the assembler with something a little bigger than just a main() with a printf() call. No real reason for it, I just felt like it. ;)

PR #2265, opened 2017-08-08
[trivial] LLVM reverted the change that required these version specific changes.

JohanEngelen:
https://llvm.org/viewvc/llvm-project?view=revision&revision=309286

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -139,11 +139,7 @@ void printVersion(llvm::raw_ostream &OS) {
   // redirecting stdout to a file.
   OS.flush();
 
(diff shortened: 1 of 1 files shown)


kinke:
During the release-candidate phase, right? %) - @thewilsonator: I guess LLVM master is also affected by this, so we'd probably need a new prebuilt LLVM again.

thewilsonator:
Will do. And i do use llvm master.

JohanEngelen:
Merging because it prevents any local development and also prevents CI testing.
@thewilsonator Please update the SPIRV build asap, otherwise we will have to revert to testing without SPIRV.

PR #2266, opened 2017-08-10
Replace hacky disabling of fp-elimination for functions with inline asm

kinke:
... by proper function attributes. Clang sets these 2 attributes if frame pointer elimination is to be fully disabled (see [here](https://github.com/llvm-mirror/clang/blob/b3d4dafef1d8deb0e748cc9625ff86c38735ca0d/lib/CodeGen/CGCall.cpp#L1706-L1707)).

[I only came across this by accident.]


kinke:
LLVM 3.9+ only apparently. :/

PR #2267, opened 2017-08-12
Issue 1876 - nightly builds don't run with tracker.exe

rainers:
Trying not to mix different C++ compilers and runtimes.


rainers:
Didn't help.

Issue #2179, opened 2017-06-23
Cmake problem : installation of /usr/share/bash-completion/completions does not take into account of CMAKE_INSTALL_PREFIX


aitzkora:
Hi,
When I configure the sources to install in my home directory, at the install level, cmake complains about
permissions for writing /usr/share/bash-completion/completions/ldc2, although I set CMAKE_INSTALL_PREFIX to a  user directory. When I do the install it stops because he cannot writes
in /usr/share/... Can you fix that please ?
Thanks, 
best regards

kinke:
Thanks for reporting, fixed by #2197.

Fatmice:
This is still a problem in origin/master as of ec9ffe29f108d5b2dd7c86e3390dbb44d7c5c79f.  I'm using Fedora 26.

I make a new folder call ldc-build outside of ldc, which has the git repo.
While in ldc-build, I issue this
`cmake -DCMAKE_INSTALL_PREFIX=../install/ldc -DINCLUDE_INSTALL_DIR=../install/ldc/import -DCONF_INST_DIR=../install/ldc/etc -DCMAKE_BUILD_TYPE=Release -DLIB_SUFFIX=64 -DLLVM_ROOT_DIR=../install/llvm -DD_COMPILER=../install/ldc-0.17.4-bootstrap/bin/ldmd2 ../ldc`

Here's the two important files, CMakeCache.txt and cmake_install.cmake in a zip.
[cmake.zip](https://github.com/ldc-developers/ldc/files/1214018/cmake.zip)

As you can see, line 82-93 in cmake_install.cmake, "/usr/share/bash-completion/completions" is still explicitly used as the destination for files in bash_completion.d
```
if("${CMAKE_INSTALL_COMPONENT}" STREQUAL "Unspecified" OR NOT CMAKE_INSTALL_COMPONENT)
     list(APPEND CMAKE_ABSOLUTE_DESTINATION_FILES "/usr/share/bash-completion/completions/")
     if(CMAKE_WARN_ON_ABSOLUTE_INSTALL_DESTINATION)
         message(WARNING "ABSOLUTE path INSTALL DESTINATION : ${CMAKE_ABSOLUTE_DESTINATION_FILES}")
     endif()
     if(CMAKE_ERROR_ON_ABSOLUTE_INSTALL_DESTINATION)
[...]

PR #2269, opened 2017-08-12
Installation improvements

Fatmice:
- Bash completion now installs relative to CMAKE_INSTALL_PREFIX
- Include paths and linker paths in ldc2.conf are populated by %%ldcbinarypath%%.  This allows self-contained installation.  This is compatible with "make DESTDIR=/path install" and can be used in combination with CMAKE_INSTALL_PREFIX
- Added optional variable INCLUDE_PATH, which defaults to "include/d" in runtime/CMakeList.txt, that can be set to modify the final path of INCLUDE_INSTALL_DIR
- @INCLUDE_PATH@ replaced @INCLUDE_INSTALL_DIR@ in ldc2.conf as the latter is now a combination of %%ldcbinarypath%%/../@INCLUDE_PATH@
- Added @LIB_SUFFIX@ to the linker path in ldc2.conf
Fixes #2197 #2179


Fatmice:
What to do about the CircleCI fail?
Also locally, my PR fail one test, the lit one.  It passes all but [two](https://gist.github.com/Fatmice/87624bdecce115452ceab5a60b63c474) of the lit tests.

kinke:
Don't worry about Circle, that's totally unrelated. Actually, none of your changes are tested by CI, they don't install LDC.

JohanEngelen:
@Fatmice The lit errors are unrelated. Are you able to run executables that are produced with `ldc2 -flto=thin` ? If not: please file a separate bug report and we'll try to figure out what's wrong there.

Fatmice:
No I'm never able to run the executable produced with -flto=thin.  The linking pass with no issue but always give something like this
Error: /tmp/app-9d64b31-56167b failed with status: -2
Error: message: Illegal instruction (core dumped)
Error: program received signal 2 (Interrupt)
I'm trying to figure out why it is, maybe it is because my llvm isn't built with lto=thin.  Fedora 26 automatically comes with a binutils that has the LLVMgold.so plugin, so that is probably why the linking was okay.  However, my laptop does not have a system installed of the LLVM that was referenced by ldc for building, only locally, and I know I built that llvm (4.0.1) without the lto=thin flag.  Maybe rebuilding llvm and and subsequently ldc will make the error it go away, but that is just a guess.

JohanEngelen:
> maybe it is because my llvm isn't built with lto=thin

That definitely will not solve the problem (It will cost you a lot of time to test and find out it doesn't help).
Try building LLVM while passing `-DLLVM_BINUTILS_INCDIR=/path/to/binutils/include -DLDC_INSTALL_LTOPLUGIN=ON` to CMake, such that a new `LLVMgold.so` plugin is created and copied to the LDC build. The correct include path will contain the file `plugin-api.h` . After that, running cmake for LDC again should copy `LLVMgold-ldc.so` into LDC's `lib` directory.

Fatmice:
I did this but it still fail the two lto tests.  I'll file a new bug report for this.

Fatmice:
Okay, I'll see about making what you want work.

Fatmice:
I have no objection for what you said but making everything relative to %%ldcbinarypath%% would break in the case of DESTDIR it not for what I implemented for @LDC2_INCLUDE_DIR@
But yes, if you would prefer the simpler case I can make that work...User would have to manually fix their include path in ldc2.conf if they move their DESTDIR installation.

Fatmice:
@klickverbot I've modified the changes to be compatible with existing usage while keeping what I added.  Anything else needed for merge approval?

Fatmice:
@klickverbot will this PR make it into 1.4 release?

kinke:
I'm in favor of postponing this; I think it's too big a change for beta1 -> final, and if I was a package maintainer, I'd be pissed off if it wasn't 100% backwards compatible.

Fatmice:
Okay, I can pare down to only the bash completion change to be relative to CMAKE_INSTALL_PREFIX.  The rest of the changes I can remove for personal use.  This will be more conservative and in line with what you are comfortable with.  Agreed?

Fatmice:
Nice all checks passed

runtime/CMakeLists.txt
```diff
@@ -35,7 +35,14 @@ set(DMDFE_VERSION         ${D_VERSION}.${DMDFE_MINOR_VERSION}.${DMDFE_PATCH_VERS
...
 set(BUILD_BC_LIBS         OFF                                 CACHE BOOL   "Build the runtime as LLVM bitcode libraries")
-set(INCLUDE_INSTALL_DIR   ${CMAKE_INSTALL_PREFIX}/include/d   CACHE PATH   "Path to install D modules to")
+if((NOT DEFINED INCLUDE_INSTALL_DIR) AND (NOT DEFINED CMAKE_INSTALL_PREFIX))
+	set(INCLUDE_INSTALL_DIR		"/usr/include/d")
+elseif((NOT DEFINED INCLUDE_INSTALL_DIR) AND (DEFINED CMAKE_INSTALL_PREFIX))
+	set(INCLUDE_INSTALL_DIR		${CMAKE_INSTALL_PREFIX}/include/d)
+else()
+	set(INCLUDE_INSTALL_DIR		${INCLUDE_INSTALL_DIR})
+endif()
+set(INCLUDE_INSTALL_DIR	  ${INCLUDE_INSTALL_DIR}	  	  	  CACHE PATH   "Path to install D modules to" FORCE)
```
kinke: How about
```
if(DEFINED CMAKE_INSTALL_PREFIX)
    set(INCLUDE_INSTALL_DIR_DEFAULT /usr/include/d)
else()
    set(INCLUDE_INSTALL_DIR_DEFAULT ${CMAKE_INSTALL_PREFIX}/include/d)
endif()
...
set(INCLUDE_INSTALL_DIR   ${INCLUDE_INSTALL_DIR_DEFAULT}   CACHE PATH   "Path to install D modules to")
```
That `..._DEFAULT` is used in other places as well, e.g., [here](https://github.com/ldc-developers/ldc/blob/master/CMakeLists.txt#L729-L734).

Fatmice: Yes, more to the point.  I'll fix.

Fatmice: Hm...on second thought.  What if -DINCLUDE_INSTALL_DIR was set though?  Wouldn't that simply override it with whatever is in INCLUDE_INSTALL_DIR_DEFAULT?

Fatmice:
```
if(NOT DEFINED CMAKE_INSTALL_PREFIX)
    set(INCLUDE_INSTALL_DIR_DEFAULT /usr/include/d)
else()
    set(INCLUDE_INSTALL_DIR_DEFAULT ${CMAKE_INSTALL_PREFIX}/include/d)
endif()
if(NOT DEFINED INCLUDE_INSTALL_DIR)
	set(INCLUDE_INSTALL_DIR   ${INCLUDE_INSTALL_DIR_DEFAULT}  CACHE PATH   "Path to install D modules to")
else()
	set(INCLUDE_INSTALL_DIR   ${INCLUDE_INSTALL_DIR}		  CACHE PATH   "Path to install D modules to")
endif()
```

kinke: CACHE variables (without `FORCE`) are only set if they aren't set already (e.g., on command line or from CMake cache for repeated cmake invokations), so what I wrote above is exactly what's required to let the user enforce his/her own `INCLUDE_INSTALL_DIR` on the command line.

Fatmice: Yes, you are right.  I forgot about the FORCE.

JohanEngelen:
( @Fatmice Instead of merging master into the branch, please rebase onto master. That will get rid of all the merge commits, clobbering your actual work. )

Fatmice:
> ( @Fatmice In the future: Instead of merging master into the branch, please rebase onto master. That will get rid of all the merge commits, clobbering your actual work. )

I should have forked before submitting this PR.
If I do git rebase origin master, the upstream will clobber my local (myldc) master

JohanEngelen:
( Ah yes I see. I generally recommend not to use the master in one's own fork. )

kinke:
Should be obsoleted by #2693. As `CMAKE_INSTALL_PREFIX` defaults to `/usr/local`, the `INCLUDE_INSTALL_DIR_DEFAULT` change shouldn't be necessary.

PR #2272, opened 2017-08-13
CMake: Include D source files in Visual Studio project

kinke:
Tested with Visual Studio 2017; works fine, and the `LDCShared` project/lib can still be built, the D files are simply ignored when building it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -386,10 +386,11 @@ set(LDC_D_SOURCE_FILES
 
 source_group("Source Files\\${DDMDFE_PATH}" FILES ${FE_SRC_D})
 source_group("Header Files\\${DDMDFE_PATH}" FILES ${FE_HDR})
-source_group("Source Files\\gen" FILES ${GEN_SRC})
+source_group("Source Files\\gen" FILES ${GEN_SRC} ${GEN_SRC_D})
 source_group("Header Files\\gen" FILES ${GEN_HDR})
-source_group("Source Files\\ir" FILES ${IR_SRC})
+source_group("Source Files\\ir" FILES ${IR_SRC} ${IR_SRC_D})
 source_group("Header Files\\ir" FILES ${IR_HDR})
+source_group("Source Files" FILES ${DRV_SRC_D})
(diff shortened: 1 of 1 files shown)


kinke:
@rainers: I use these projects for debugging LDC, by simply specifying a debug command line invoking my RelWithDebInfo LDC (built externally via ninja). I have Visual D 0.45, and D debugging isn't that bad this way, but I cannot choose a debug engine as for a Visual D project [I built the VisualD project for DMD a few days ago, and debugging that worked very well!]. Can you shed some light on how things could be improved for LDC, without having to maintain a VisualD project file with a complete file listing etc.?

kinke:
Well debugging has just improved a lot - LLVM's dump() seems to work now. I changed more things at the same time, so I don't know what fixed it. Using Visual Studio 2017 now and merge-2.075 with LDC-LLVM (incl. CodeView cherry-pick) as host compiler.

rainers:
> I use these projects for debugging LDC, by simply specifying a debug command line invoking my RelWithDebInfo LDC

I guess you mean "devenv ldc2.exe ...". That always uses the VS debug engine. For the C++ part that should work as usual, the D code needs some debug info fixups to work with that engine though (without the mago expression evaluator).
If you absolutely want to use the mago debug engine similarly just create a tiny Visual D project and switch the target executable in the debugger settings.

> Well debugging has just improved a lot for me - LLVM's dump() seems to work now.

Good to hear. I used dump() a lot, too, but don't remember having a problem with it.

kinke:
> I guess you mean "devenv ldc2.exe ...". That always uses the VS debug engine.

I meant just changing the debug command line of a project's properties.

> If you absolutely want to use the mago debug engine similarly just create a tiny Visual D project and switch the target executable in the debugger settings.

Yep, adding a dummy project should do, I was just wondering whether it could be enforced via a .vcxproj trick. And with CMake's [include_external_msproject](https://cmake.org/cmake/help/latest/command/include_external_msproject.html), I guess this could be automated, incl. setting the debug command line (configuring a .visualdproj.in file and adding that to the solution).

Issue #1020, opened 2015-08-13
incorrect error message when compiling out-of-line c++ method with it's declaration


MartinNowak:
```
cat > objc_stubs.d << EOF
struct Objc_ClassDeclaration
{
    // MARK: Objc_ClassDeclaration
    extern (C++) bool isInterface()
    {
        return false;
    }
}
EOF

cat > objc.di << EOF
struct Objc_ClassDeclaration
{
    // MARK: Objc_ClassDeclaration
    extern (C++) bool isInterface();
}
EOF

ldc2 -c -ofsingle.o objc_stubs.d objc.di
```

objc.di(4): Error: Function type does not match previously declared function with the same mangled name: _ZN21Objc_ClassDeclaration11isInterfaceEv

Simple to solve by not compiling the header.

kinke:
I can't reproduce this anymore. The signatures/LLVM function types should be identical anyway. So it was probably another issue which has probably been fixed in the meantime.

kinke:
Well, I don't know what I tested back then, but the issue is still here (but the additional `-singleobj` switch is needed). The issue is that there's a `objc.Objc_ClassDeclaration` and a `objc_stubs.Objc_ClassDeclaration` LL type. I guess the D namespace prefix needs to be omitted for `extern(C++)`.

rainers:
Here is an example (for Windows) that does not need multiple modules, template expansions are bad enough:
```
import std.datetime;

extern(Windows)
{
	struct FILETIME {
		DWORD dwLowDateTime;
		DWORD dwHighDateTime;
	}
	void GetSystemTimeAsFileTime(FILETIME*);
}

void fun()
{
	Clock.currStdTime;
	GetSystemTimeAsFileTime(null);
}
```
The problem is that std.datetime uses a forward declaration from core.sys.windows.winbase that references a different (though identical) FILETIME declaration.

kinke:
> I guess the D namespace prefix needs to be omitted for extern(C++).

A while back I looked into this. If I recall correctly, the problem is that a type doesn't have any linkage (makes sense, only its function members have one) and I didn't find a quick way to check if the type has been declared inside an `extern(C[++])` scope (in order to use identical LL type names for multiple declarations in different D modules).

jacob-carlborg:
Here's another use case that causes problems using LDC but works with DMD:

```d
extern (Objective-C) interface NSObject
{
[...]

PR #2273, opened 2017-08-13
[2.075] Be more forgiving wrt. multiple function declarations/definitions

kinke:
... sharing a common mangled name. [compilable/test17352.d](https://github.com/ldc-developers/dmd-testsuite/blob/ldc-merge-2.075/compilable/test17352.d) tests that the compiler accepts 2 identical templated function instantiations.
See https://issues.dlang.org/show_bug.cgi?id=17352.

Clashes for function *declarations* are more common, mainly wrt. multiple identically looking `extern(C++)` forward declarations in different D modules:

a.d:
```D
extern(C++) struct S {}
extern(C++) void foo(S s); // type: (a.S) => void
```

b.d:
```D
extern(C++) struct S {}
extern(C++) void foo(S s); // type: (b.S) => void
```

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -533,11 +533,10 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
     func = LLFunction::Create(functype, llvm::GlobalValue::ExternalLinkage,
                               mangledName, &gIR->module);
   } else if (func->getFunctionType() != functype) {
-    error(fdecl->loc,
-          "Function type does not match previously declared "
-          "function with the same mangled name: `%s`",
-          mangleExact(fdecl));
-    fatal();
+    warning(fdecl->loc,
+            "type of function `%s` does not match previously declared "
+            "function with the same mangled name: %s",
+            fdecl->toPrettyChars(), mangleExact(fdecl));
   }
 
   func->setCallingConv(gABI->callingConv(func->getFunctionType(), link, fdecl));
@@ -915,7 +914,16 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
     return;
   }
 
-  IrFunction *irFunc = getIrFunc(fd);
+  IrFunction *const irFunc = getIrFunc(fd);
+  llvm::Function *const func = irFunc->getLLVMFunc();
+
+  if (!func->empty()) {
+    warning(fd->loc,
+            "skipping definition of function `%s` due to previous definition "
+            "for the same mangled name: %s",
+            fd->toPrettyChars(), mangleExact(fd));
+    return;
+  }
 
   // debug info
   irFunc->diSubprogram = gIR->DBuilder.EmitSubProgram(fd);
@@ -934,7 +942,6 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
 
   const auto f = static_cast<TypeFunction *>(fd->type->toBasetype());
   IrFuncTy &irFty = irFunc->irFty;
-  llvm::Function *func = irFunc->getLLVMFunc();
 
   const auto lwc = lowerFuncLinkage(fd);
   if (linkageAvailableExternally) {


kinke:
Would fix #1020.

kinke:
A little more context: `compilable/test17352.d` failed due to a debuginfo mismatch. The templated function was first defined and associated with a specific dwarf subprogram. It was later re-defined a 2nd time; the associated subprogram for the function remained unchanged, but the dwarf metadata in the body referred to the 2nd subprogram as enclosing scope. So I guess only the body was replaced by the 2nd definition.

JohanEngelen:
If this is valid D code, then there shouldn't be any warnings, right? 

Not erroring on redefinition of functions seems very questionable. But this will only happen for extern(C) and extern(C++) functions, right? 

As you mention in #1020, if the frontend would store the linkage for types too, then the problem is much better solvable. (it already stores `extern(C++,{struct,class})` for structs and classes, and I believe it also stores `extern(C++)` already for classes)

kinke:
> But this will only happen for extern(C) and extern(C++) functions, right?

As the linked dmd-testsuite test shows, unfortunately clearly not. The extern(D) templates get each an alias to a local of the same type and name, so they are mangled identically... messy stuff. That's why I thought I'd like to warn at least [and then remembering your issue about warnings during codegen not being fatal for `-w` and ignored if not using `-wi`... ;)].

JohanEngelen:
> [and then remembering your issue about warnings during codegen not being fatal for -w and ignored if not using -wi... ;)]

Thanks. I still have a bad scar from not being able to add warning categorization to dmd to make -werror usable. :( :(

gen/functions.cpp
```diff
@@ -533,11 +533,10 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
...
   } else if (func->getFunctionType() != functype) {
-    error(fdecl->loc,
-          "Function type does not match previously declared "
-          "function with the same mangled name: `%s`",
-          mangleExact(fdecl));
-    fatal();
+    warning(fdecl->loc,
+            "type of function `%s` does not match previously declared "
+            "function with the same mangled name: %s",
+            fdecl->toPrettyChars(), mangleExact(fdecl));
```
JohanEngelen: Can you add the reason in comments about why we are continuing instead of returning here?

kinke:
@klickverbot: What's your opinion on this? I think a proper solution would be using a hashmap or something in the front-end, so that multiple identical template instantiations inside a module are folded to a single one. But then again with `-singleobj`, I guess we'd still hit it across modules...

ZombineDev:
> I think a proper solution would be using a hashmap or something in the front-end, so that multiple identical template instantiations inside a module are folded to a single one. 

That should already be the case, see:
https://github.com/ldc-developers/ldc/blob/v1.4.0-beta1/ddmd/dtemplate.d#L504

If that's not working (in the domain of a single compiler invocation) there must be a bug.

kinke:
@ZombineDev: DMD also just works around this in the backend, e.g., https://github.com/dlang/dmd/commit/76affab78fec0c9acb0253d9672c5c0bdf843dd4.

ZombineDev:
Yes, but that "should not happen" if the front-end is working as designed. (It's in the [spec](https://dlang.org/spec/template):

> 5. Multiple instantiations of a TemplateDeclaration with the same TemplateArgumentList all will refer to the same instantiation.

)
The relatively young age of the [PR](https://github.com/dlang/dmd/pull/6955) indicates that for the most part folding of identical template instances was already working, and what's left are corner cases which no one has had the time to deeply investigate.

kinke:
> What I mean by this is that ideally the problem should be fixed in the front-end and not worked around in all back-ends.

I can't agree more. ;) Doesn't change the fact that whenever DMD employs workarounds like that, all other compilers have to follow suit until it's properly fixed.

kinke:
The multiple definitions part is in master.

PR #2279, opened 2017-08-17
[MSVC] CMake: Prepare for LLVM 5.0

kinke:
Some LLVM libs were missing from the linker flags for current LLVM 5.0. It was only during troubleshooting that I noticed the MSVC special case in `FindLLVM.cmake`. Getting rid of it entirely made the issues disappear. I checked `llvm-config.exe`, and it's returning appropriate switches for the MS toolchain, no need for special processing.

This apparently also leads to `LLVM_CXXFLAGS` being used for the first time when building with MSVC. These include `/W4`, a level further up from CMake's default `/W3`, so I disabled a few more frequent warnings.

I also had to link in a lib from Microsoft's Debug Info Access SDK manually. As the VS command prompt unfortunately doesn't even include the DIA SDK dir in the `LIBPATH` environment variable, I even had to specify a full path, relying on the `VSINSTALLDIR` environment variable.

### DIFF
--- a/cmake/Modules/FindLLVM.cmake
+++ b/cmake/Modules/FindLLVM.cmake
@@ -52,74 +52,9 @@ macro(_LLVM_FAIL _msg)
 endmacro()
 
 
-if ((WIN32 AND NOT(MINGW OR CYGWIN)) OR NOT LLVM_CONFIG)
-    if (WIN32)
-        # A bit of a sanity check:
-        if( NOT EXISTS ${LLVM_ROOT_DIR}/include/llvm )
-            message(FATAL_ERROR "LLVM_ROOT_DIR (${LLVM_ROOT_DIR}) is not a valid LLVM install")
-        endif()
-        # We incorporate the CMake features provided by LLVM:
-        set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${LLVM_ROOT_DIR}/share/llvm/cmake;${LLVM_ROOT_DIR}/lib/cmake/llvm")
-        include(LLVMConfig)
-        # Set properties
-        set(LLVM_HOST_TARGET ${TARGET_TRIPLE})
-        set(LLVM_VERSION_STRING ${LLVM_PACKAGE_VERSION})
-        set(LLVM_CXXFLAGS ${LLVM_DEFINITIONS})
-        set(LLVM_LDFLAGS "")
-        list(REMOVE_ITEM LLVM_FIND_COMPONENTS "all-targets" index)
-        list(APPEND LLVM_FIND_COMPONENTS ${LLVM_TARGETS_TO_BUILD})
-        # Work around LLVM bug 21016
-        list(FIND LLVM_TARGETS_TO_BUILD "X86" TARGET_X86)
-        if(TARGET_X86 GREATER -1)
-            list(APPEND LLVM_FIND_COMPONENTS x86utils)
-        endif()
-        # Similar to the work around above, but for AArch64
-        list(FIND LLVM_TARGETS_TO_BUILD "AArch64" TARGET_AArch64)
-        if(TARGET_AArch64 GREATER -1)
-            list(APPEND LLVM_FIND_COMPONENTS AArch64Utils)
-        endif()
-        # Similar to the work around above, but for AMDGPU
-        list(FIND LLVM_TARGETS_TO_BUILD "AMDGPU" TARGET_AMDGPU)
-        if(TARGET_AMDGPU GREATER -1)
-            list(APPEND LLVM_FIND_COMPONENTS AMDGPUUtils)
-        endif()
-        if(${LLVM_VERSION_STRING} MATCHES "^3\\.[0-8][\\.0-9A-Za-z]*")
-            # Versions below 3.9 do not support components debuginfocodeview, globalisel
-            list(REMOVE_ITEM LLVM_FIND_COMPONENTS "debuginfocodeview" index)
-            list(REMOVE_ITEM LLVM_FIND_COMPONENTS "globalisel" index)
-        endif()
-        if(NOT ${LLVM_VERSION_STRING} MATCHES "^3\\.[0-7][\\.0-9A-Za-z]*")
-            # Versions beginning with 3.8 do not support component ipa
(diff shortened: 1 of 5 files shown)


kinke:
AppVeyor: `The command line is too long.` ...

rainers:
> I also had to link in a lib from Microsoft's Debug Info Access SDK manually.

I never had trouble with this when building with LLVM master. I guess that's because I'm not compiling with LLD integration. Does it have debug info support now?

cmake/Modules/FindLLVM.cmake
```diff
@@ -52,74 +52,9 @@ macro(_LLVM_FAIL _msg)
...
-            list(REMOVE_ITEM LLVM_FIND_COMPONENTS "globalisel" index)
-        endif()
-        if(NOT ${LLVM_VERSION_STRING} MATCHES "^3\\.[0-7][\\.0-9A-Za-z]*")
-            # Versions beginning with 3.8 do not support component ipa
-            list(REMOVE_ITEM LLVM_FIND_COMPONENTS "ipa" index)
-        endif()
-        if(${LLVM_VERSION_STRING} MATCHES "^3\\.[0-9][\\.0-9A-Za-z]*")
-            # Versions below 4.0 do not support component debuginfomsf
-            list(REMOVE_ITEM LLVM_FIND_COMPONENTS "debuginfomsf" index)
-        endif()
```
rainers: Removing this list is very much appreciated. I had to add the binaryformat library recently, too (probably caused by the selected targets).

kinke:
Yep binaryformat was missing for me too. I disabled the LLD integration (needs #2148 otherwise to work) but still needed `diaguids.lib` due to a few class and interface GUIDs.

kinke:
The remaining AppVeyor issue is unrelated btw, realizing that cost me some hours yesterday - they probably changed something in their image (and once again didn't mention anything in their blog/Twitter feed/...). Built this PR yesterday successfully with both LLVM 5.0 and 4.0.1.

kinke:
AppVeyor upgraded to the latest VS + WinSDK update and since then master is red. Upgrading myself right now...

kinke:
I have the same issue now, even after rebuilding LLVM. LDC segfaults during a GC collection triggered by a druntime module ctor, in `gc.impl.conservative.gc.Gcx.mark()` called by `rt.tlsgc.scan()`...

kinke:
Might have something to do with the `.tls` section disappearing from `ldc2.exe` when linking with VS2017 v15.3.1. @rainers: You're the expert for that kind of thing, right? ;)

rainers:
> Might have something to do with the .tls section disappearing from ldc2.exe when linking with VS2017 v15.3.1 (but the magic _tls_start/end symbols are still there and bracket a plausibly small range). 

Oh, boy. Sounds like this will break every program linked against/with v15.3.1 ;-(

>@rainers: You're the expert for that kind of thing, right? ;)

I'll have a look...

rainers:
The problem seems to be that .tls is no longer an image section, but only a segment in the .data section. TLS access seems to be ok, but _tls_start and _tls_end are now evaluated relative to the .data section:
```
int x = 7;

extern(C) extern
{
	int _tls_start;
	int _tls_end;
}

void[] initTLSRanges() nothrow @nogc
{
	auto pbeg = cast(void*)&_tls_start;
	auto pend = cast(void*)&_tls_end;
	return pbeg[0 .. pend - pbeg];
}

void main()
{
	printf("%p: %d\n", &x, x);
	
	auto r = initTLSRanges();
	printf("%p: %d\n", r.ptr, r.length);
}
```
prints
```
000001B0638D7470: 7
000001B0638E60E0: 233
```
i.e. the variable is not in the TLS range (but it's the range that is wrong here). Happens with dmd, too, if using V2017 v15.3.1. This seems to be a linker issue as exchanging the vc library with the one from VS2015 doesn't help.

kinke:
Thanks for the analysis. So if we were to work around that, the host compiler's druntime would have to be patched first, thanks MS, very much appreciated.
I can't even dowgrade to 15.2. Needed to uninstall VS2017 completely and download the 15.0 installer (no installers for the previous updates). I can only upgrade directly to 15.3.1 later on. And I thought the whole reason for the new VS2017 system with less registry keys was to enable different parallel installations...

rainers:
> So if we were to work around that, the host compiler's druntime would have to be patched first

Depends on the workaround. Maybe we can enforce moving the segment back into the .tls section.

>I can't even downgrade to 15.2.

I guess my VS2017 is broken now, too. (I mostly use 2013 and 2015, though).

They also changed the vcvars64.bat to always switch to %HOME%\source. That breaks a good deal of my batches, probably also some builds by Visual D.

kinke:
> Does it have debug info support now?

From https://llvm.org/docs/CMake.html#llvm-specific-variables:
```
LLVM_ENABLE_DIA_SDK:BOOL
    Enable building with MSVC DIA SDK for PDB debugging support. Available only with MSVC. Defaults to ON.
```

Don't know what that means exactly and don't have time to dig into it right now, but I take it as a good sign. ;)

kinke:
Btw, Microsoft released 15.3.**2** 4 days later. I don't dare testing that yet though.

> They also changed the vcvars64.bat to always switch to %HOME%\source. That breaks a good deal of my batches

Yep, mine too. Additionally, the `$(Platform)` MSBuild variable apparently isn't defined anymore...

rainers:
> Btw, Microsoft released 15.3.2 4 days later. I don't dare testing that yet though.

I tried it, no improvement for our issues. cv2pdb also doesn't work anymore...

rainers:
> So if we were to work around that, the host compiler's druntime would have to be patched first, thanks MS, very much appreciated.

I have not found any work around, so here is the runtime change: https://github.com/dlang/druntime/pull/1910

rainers:
> I have not found any work around, so here is the runtime change: dlang/druntime#1910

I stumbled over a workaround: there is a hidden linker switch /noopttls that restores the old behaveour. Seems to exist starting with VS2015.

kinke:
Oh wow, how did you find out about that one? No single Google hit... ;)

rainers:
> Oh wow, how did you find out about that one? No single Google hit... ;)

Staring at strings in link.exe. You can find a quite some more undocumented switches this way.

Issue #2186, opened 2017-06-29
ldc.conf and switch order


John-Colvin:
switches in ldc.conf are added to `allArguments` in `parseCommandLine`, deliberately inserted before any command-line switches. What was the motivation for this order? AFAICT `dmd` does it the other way around.

kinke:
Simple - being able to override the default switches in the config.

John-Colvin:
Some things work the other way around though: e.g. linker flags are normally processed left to right, so if I add a path for the linker to search in it will always be searched *after* the paths in the config. That wouldn't be so bad if those were paths that only contained Phobos (because can use defaultlib to override), but it's common to have `/usr/bin` or `/usr/local/bin` used, which means you can't use -L flags to override system libraries like you can with `dmd` and `gcc/clang`.

kinke:
Yep, the (few) list options obviously cannot be overridden and only appended to this way. For these few, the relative order only matters for `-L` and `-Xcc` afaict.
Selected libraries may be overridden by using their path as input (`ldc2 ... /path/to/libmine.a`), or, as a last resort, by specifying another config file with no lib paths.

Open for suggestions on how to improve this, but simply appending the default switches makes ~99% of those non-overridable (incl. `-defaultlib=` for the bare metal guys), so I'm against that.

kinke:
(E.g., we could filter out all default switches beginning with `-L-L` and append those to the command line.)

JohanEngelen:
@John-Colvin How about what @kinke proposed?  Filtering `-L-L` switches and adding those in front of ldc.conf switches?

kinke:
> generalizable to a "post-switches = [ ... ]" section in ldc.conf

Hmm, tasty, I like it, making the process fully transparent. `pre/post-switches`? `switches-prefix/suffix`?

JohanEngelen:
Do we have to be careful about backward compatibility? (i.e. we need to preserve the "switches" name?)

We can also add comments in the JSON making things clear if the names aren't.

kinke:
We can simply fall back to `switches` for the prefix ones if the new name isn't found.

JohanEngelen:
true. Go for it. Would be a nice addition :)

kinke:
> Go for it.

It's not like I don't do enough here already. Way too much of my spare time actually.

kinke:
Fixed by #2281.

PR #2281, opened 2017-08-20
Add `post-switches` to config file sections

kinke:
For switches to be appended after the user switches (or right before the first `-run` switch).

The sections inherit it (as well as `switches`) from the `default` section if it isn't overridden explicitly.

Fixes issue #2186.

### DIFF
--- a/driver/configfile.cpp
+++ b/driver/configfile.cpp
@@ -203,6 +203,17 @@ bool ConfigFile::read(const char *explicitConfFile, const char *section) {
 }
 
 void ConfigFile::extendCommandLine(llvm::SmallVectorImpl<const char *> &args) {
-  // insert switches from config file before all explicit ones
+  // insert 'switches' before all user switches
   args.insert(args.begin() + 1, switches.begin(), switches.end());
+
+  // append 'post-switches', but before a first potential '-run'
+  size_t runIndex = 0;
+  for (size_t i = 1; i < args.size(); ++i) {
+    if (strcmp(args[i], "-run") == 0 || strcmp(args[i], "--run") == 0) {
+      runIndex = i;
+      break;
+    }
+  }
+  args.insert(runIndex == 0 ? args.end() : args.begin() + runIndex,
+              postSwitches.begin(), postSwitches.end());
 }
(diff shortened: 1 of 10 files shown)


JohanEngelen:
It'd be very nice to get this into 1.4. Without it, it is a lot more work to use LTO/sanitizer builds of the standard library.
Biggest complaint is that tests are missing.

kinke:
Sure thing, but as there was no feedback, I didn't want to postpone beta1 further.

kinke:
Oh and wrt. the tests - as all config files use it, it's tested everywhere, at least that both sets of switches are used (and that the `post-switches` are inserted before `-run`). I don't think we output the full command line anywhere yet...

JohanEngelen:
I'm working on the tests.

JohanEngelen:
(pfiew, ok, the git pushing worked)

tests/driver/post_switches.d
```diff
@@ -0,0 +1,21 @@
+// RUN: not %ldc -I=%runtimedir/src -conf=%S/inputs/post_switches.conf %s -v -L-user-passed-switch | FileCheck %s
+
+// CHECK: -normal-switch
+// CHECK-SAME: -normal-two-switch
+// CHECK-SAME: -user-passed-switch
+// CHECK-SAME: -post-switch
+// CHECK-SAME: -post-two-switch
+
+
+// RUN: not %ldc -I=%runtimedir/src -conf=%S/inputs/post_switches.conf -v -L-user-passed-switch -run %s -L-after-run | FileCheck %s --check-prefix=WITHrUN
```
JohanEngelen: the `WITHrUN` is not a typo btw. Lit confuses `WITHRUN:` lines for `RUN:` lines.

JohanEngelen:
Is this good to merge?

driver/configfile.cpp
```diff
@@ -201,3 +201,19 @@ bool ConfigFile::read(const char *explicitConfFile, const char *section) {
...
 }
+
+void ConfigFile::extendCommandLine(llvm::SmallVectorImpl<const char *> &args) {
+  // insert 'switches' before all user switches
+  args.insert(args.begin() + 1, switches.begin(), switches.end());
+
+  // append 'post-switches', but before a first potential '-run'
+  size_t runIndex = 0;
+  for (size_t i = 1; i < args.size(); ++i) {
+    if (strcmp(args[i], "-run") == 0 || strcmp(args[i], "--run") == 0) {
```
klickverbot: In an ideal world, this would be `args[i] == "-runs"s || args[i] == "--run"s`. (Only made it into C++14, though, I think.)

PR #2282, opened 2017-08-20
Travis OSX: Save almost 2 minutes (by skipping brew)

kinke:
Installing an official pre-built ninja binary (73KB zipped) instead.


JohanEngelen:
nice :)

PR #2285, opened 2017-08-23
libFuzzer: support LLVM 6.0 compiler-rt fuzzer

JohanEngelen:
libFuzzer was moved to compiler-rt, and therefore its location and filename has changed.
Tested locally on OSX only.


JohanEngelen:
@thewilsonator Hm, it's strange that `lib/clang/6.0.0/lib/darwin/libclang_rt.asan_dynamic_osx.a` is found but not `lib/clang/6.0.0/lib/darwin/libclang_rt.fuzzer_osx.a` in the SPIRV LLVM package.

PR #2289, opened 2017-08-26
Improve Android support in runtime CMake script

kinke:
Based on @joakim-noah's command-line in https://wiki.dlang.org/Building_LDC_runtime_libraries.


joakim-noah:
Thanks, didn't see this quickly enough. :)

PR #2291, opened 2017-08-26
asm-x86, fix implicit fallthrough.

JohanEngelen:


JohanEngelen:
Yeah this whole file is a mess.
What's up with the while(1)'s ?

JohanEngelen:
the thing looks like a bug to me, but... i don't want to look into it atm. The other thing I thought is that it is a construct to easily disable it or smth.

JohanEngelen:
Maybe yeah. I'll clean up.

JohanEngelen:
be my guest ;-) ;) thanks

JohanEngelen:
#2296

PR #2292, opened 2017-08-26
An attempt to fix runnable/gdb4149.d test: correct debuginfo for `ref` parameters.

JohanEngelen:
We currently do not create an `alloca` for a `ref` parameter (this is in contrast to what clang does, and in contrast to what we do for pointer parameters). The `llvm.dbg.declare` call that attaches debuginfo to the parameter expects to receive a pointer to the ref parameter: in the case of `ref int` it expects to receive a pointer of type `int**`, which is what the alloca type would be if we would have created it.

I think a proper fix is to treat `ref` parameters simply as pointer parameters (or to teach LLVM to deal with function parameters passed directly into `llvm.dbg.declare`). And thus we would create an alloca, etc. That's not so easy and is a big change. So first I am trying this fix, where we say that the parameter is of type `int` instead of telling the debugger/users that it is `int&`.

(tested on OSX with LLVM 5.0 + LLDB 5.0)

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -121,6 +121,20 @@ void ldc::DIBuilder::Declare(const Loc &loc, llvm::Value *var,
   DBuilder.insertDeclare(var, divar, diexpr, debugLoc, IR->scopebb());
 }
 
+// Use this to tag debuginfo onto a non-alloca LLVM value, e.g. a function
+// parameter.
+void ldc::DIBuilder::DbgValue(const Loc &loc, llvm::Value *var,
+                              ldc::DILocalVariable divar,
+                              ldc::DIExpression diexpr) {
+  unsigned charnum = (loc.linnum ? loc.charnum : 0);
+  auto debugLoc = llvm::DebugLoc::get(loc.linnum, charnum, GetCurrentScope());
+  DBuilder.insertDbgValueIntrinsic(var,
+#if LDC_LLVM_VER < 600
+                                   0,
+#endif
+                                   divar, diexpr, debugLoc, IR->scopebb());
+}
+
 ldc::DIFile ldc::DIBuilder::CreateFile(Loc &loc) {
   const char* filename = loc.filename;
   if (!filename)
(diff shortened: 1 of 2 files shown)


JohanEngelen:
OK, so this indeed fixes the CircleCI failure. But it only works for LLVM >= 5.0.  A change in LLVM behavior it seems.

kinke:
> That's not so easy and is a big change.

I dare doubt that. ;) Thanks for digging into this!

JohanEngelen:
About the easiness, perhaps it's easy ;)
I tried removing this special case: https://github.com/ldc-developers/ldc/blob/master/gen/functions.cpp#L756  . That brings back the alloca's for byref params, but it adds incorrect bitcasting, and then the accesses to the alloca'd pointer are wrong (needs double indirection instead of that bitcast).
I was very confused to see clang doing an alloca for all function parameters, and then I saw that we do the same thing (somehow I had forgotten) for all parameters except `ref` ones (probably `out` too). Generally, I think it's best to do whatever clang does, so it'd be nice if we can move towards alloca for `ref` too.

(sidenote: i tested it with LLDB 5.0 which works better than OSX's current LLDB, so there is progress!)

JohanEngelen:
At `-O3`, the `@llvm.dbg.declare` is turned into a `@llvm.dbg.value`. I didn't fully understand the meaning of the latter, so I didn't dare use it directly here. 
But this here provides more info, and I think it is safe for us to use: http://nondot.org/sabre/LLVMNotes/DebugInfoVariableInfo.txt

JohanEngelen:
With using `llvm.dbg.value` directly, things work in LLDB (the reference parameter shows up as reference parameter).

JohanEngelen:
Breaks this code:
```d
void g(int[] xs)
{
    foreach(ref x; xs) {}
}
```
There, an alloca is used for the ref variable.

gen/dibuilder.cpp
```diff
@@ -1040,12 +1054,20 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
...
-    auto T = DtoType(type);
-    TD = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type, TD,
-                                      getTypeAllocSize(T) * 8, // size (bits)
-                                      DtoAlignment(type) * 8); // align (bits)
+    // Note: createReferenceType expects the size to be the size of a pointer,
+    // not the size of the type the reference refers to.
+    TD = DBuilder.createReferenceType(
+        llvm::dwarf::DW_TAG_reference_type, TD,
+        gDataLayout->getPointerSizeInBits(), // size (bits)
+        DtoAlignment(type) * 8);             // align (bits)
```
kinke: You sure about that? I think we already had it the other way around once, but I'm not sure. If the pointer size is to be used, then surely the alignment refers to the pointer as well, right?

kinke:
I'll try using this reference + debug.value approach for Win64 params rewritten by ExplicitByvalRewrite. It'll probably make a good choice for nested vars as well.

gen/dibuilder.cpp
JohanEngelen: I am sure about the pointer size, I checked clang source. The alignment I'm not sure about (would have to dig more in clang source). I think indeed clang also aligns with pointer alignment.

JohanEngelen: https://github.com/llvm-mirror/clang/blob/4f648cc4fd61eb8098386ddf465438485598cd92/lib/CodeGen/CGDebugInfo.cpp#L836-L856

JohanEngelen:
#2294

PR #2293, opened 2017-08-27
D dynamic compilation support

Hardcode84:
Here is current state of my 'jit' work. Simple examples work but everything mostly untested (and I'm not sure how properly test it). Tested on win64 and linux64. IR preparation for jit mostly in gen/runtimecompile.cpp. Jit itself implemented as shared library in runtime/rtcompile-rt/cpp-so/, it was mostly copypasted from llvm examples and I'm not sure it 100% correct.

Status:

Tests:
* I am using existing lit infrastructure, not sure it is best suitable here
* It is hard to check functions were actually dynamically compiled (if we compile them statically most of the tests still pass)
* Need more tests

Platform support:

Win64:
* Thread locals in dynamic code doesn't work
```text
Relocation type not implemented yet!
UNREACHABLE executed at c:\ldc\llvm\lib\executionengine\runtimedyld\Targets/Runt
imeDyldCOFFX86_64.h:104!
```
* Throwing and cathing exception in dynamic code doesnt work due to llvm bug (https://bugs.llvm.org//show_bug.cgi?id=24233)

Linux:
* Thread locals in dynamic code doesn't work
```text
LLVM ERROR: Cannot select: 0x1007ec8: i64 = X86ISD::WrapperRIP TargetGlobalTLSAddress:i64<i64* @_D12thread_local8threadIdm> 0 [TF=10]
  0x1007e60: i64 = TargetGlobalTLSAddress<i64* @_D12thread_local8threadIdm> 0 [TF=10]
```

MacOS:
* Thread locals doesn't work

Areas of improvement:
* Revise dynamic code optimization (copypasted from ldc/clang for now)
* Research hotpatching for dynamic code dispatch (does llvm support this?)
* Option to limit dynamic functions search depth
* Extend `@runtimeCompile` attribute to allow mark some functions always non-dynamic
* Get rid of thunks when calling dynamic functions from other dynamic functions
* Check dynamic compilation with code coverage (theoretically it should work)
* Check dynamic compilation with PGO (dynamic compiler can use PGO info collected offline)
* Research into jitted partial application, eg:
```
@rumtimeCompile int foo(int a, int b)
{
  if (b == 1)
  {
    return a * 42;
  }
  return a + 42;
}


// Generate another function which interprets 'b' as constant allow llvm optimizer to drop branch completely
auto fun = apply!foo(placeholder, 0);
return fun(5); // foo(5, 0)
```
* Throw exception when attempted to call dynamic functions without compilation


Hardcode84:
Related druntime PR: https://github.com/ldc-developers/druntime/pull/101

ddmd/globals.h
```diff
@@ -234,6 +234,8 @@ struct Param
     uint32_t hashThreshold; // MD5 hash symbols larger than this threshold (0 = no hashing)
 
     bool outputSourceLocations; // if true, output line tables.
+
+    bool enableRuntimeCompile;
```
thewilsonator: A small comment like the above would be nice.

.gitmodules
```diff
@@ -1,6 +1,7 @@
 [submodule "druntime"]
 	path = runtime/druntime
-	url = https://github.com/ldc-developers/druntime.git
+	url = https://github.com/Hardcode84/druntime.git
+	branch = runtime_compile
```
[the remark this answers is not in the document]
Hardcode84: done

Hardcode84:
So here some design overview of dynamic compilation feature

`@runtimeCompie` attribute can be applied to any function (including lambdas and class virtual methods)

Compiler part:
1. In `DtoDeclareFunction` read function uda's, if there is a `@runtimeCompie` attribute set `IrFunction::runtimeCompile = true`
2. If `IrFunction::runtimeCompile == true`, call for this function `declareRuntimeCompiledFunction` which will create thunk function `rtCompileFunc`
`rtCompileFunc` has same signature and attributes as the original function and any attemt to call or take address of the this function will be redirected to `rtCompileFunc` in `DtoCallee`
3. In `DtoDefineFunction` call `defineRuntimeCompiledFunction` for any function which have `runtimeCompile == true`
4. In `defineRuntimeCompiledFunction` create global variable which will store pointer to compiled function and create thunk function body which just calls this global var 
(TODO: It is possible to get rid of this global var, we just need to hotpatch code in runtime with compiled code address, is it worth it? how to do this in llvm?)
5. In `writeAndFreeLLModule` call function `generateBitcodeForRuntimeCompile` for each module
6. In `generateBitcodeForRuntimeCompile` first recursively search functions calls inside functions body, starting with `@runtimeCompie` functions.
After that we have two lists of functions:
* Directly marked by user `@runtimeCompie`
* Not marked by user but used by `@runtimeCompie` functions directly or indirectly used by dynamic functions and with body available
Both lists are subject to dynamic compilation. For the first list dynamic version will be always used. For the second static version will be used when called from static code and dynamic version when called from dynamic code.
Also we now have a list of symbols required by dynamic functions (variables and external functions)
(TODO: Option for limit recursion depth?)
7. Clone module via `llvm::CloneModule` keeping only functions marked for dynamic compilations (directly or indirectly)
8. Optimize dynamic functions thunks in cloned module, make thunk global var constant and initialize it with function (`fixupRtThunks`) (TODO: llvm optimizer can handle this but better to remove these thunks completely in dynamic code)
9. Create module data required for dynamic compilation
```
/// Symbol required by dynamic code
struct RtCompileSymList
{
  const char* name; // Symbol name
  void* sym;        // Symbol address
};

/// Dynamic function
struct RtCompileFuncList
{
  const char* name; // Function name
  void* func;       // Thunk global var address to store compiled function pointer
};

/// Runtime compiled functions info per module
/// organized in linked list 
/// starting with `RuntimeCompileModulesHeadName` (defined in runtime)
struct RtComileModuleList
{
  RtComileModuleList* next; // Next module
  const void* irData;       // Ir data
  int32 irDataSize;         // Ir data size in bytes
  RtCompileFuncList* funcList; // Dynamic functions
  i32 funcListSize;            // Dynamic functions count
  RtCompileSymList* symList;   // Symbols
  i32 symListSize;             // Symbols count
};
```
10. Add module constructor which registers `RtComileModuleList` into global linked list

Runtime part:
1. Defines runtime compiled moduleds list head
2. Defines `rtCompileProcessImplSo` function which do actual compilation
3. `rtCompileProcessImplSo` for each module in list:
4. Parses ir data into llvm module
5. Optimizes module (TODO: optimization setup copypasted from clang and LDC itself need to revise)
6. Add modules to jit (copypasted from llvm tutorials), resolve functions using `RtComileModuleList` data and update thunk vars

Hardcode84:
I tested this on linux and there is some issue with thread locals, except that everything seems to work.
I don't have device to test on macOS.

First comment updated.

ddmd/globals.h
```diff
@@ -234,6 +234,8 @@ struct Param
     uint32_t hashThreshold; // MD5 hash symbols larger than this threshold (0 = no hashing)
 
     bool outputSourceLocations; // if true, output line tables.
+
+    bool enableRuntimeCompile; // Enable dynamic compilation
```
JohanEngelen: this is not needed (we try to reduce the frontend diff (i.e. changes in `./ddmd/`) as much as possible). Instead you should define `cl::opt<bool> enableRuntimeCompile`, such that it does not use "external" storage.

runtime/jit-rt/DefineBuildJitRT.cmake
```diff
@@ -0,0 +1,93 @@
+if(LDC_RUNTIME_COMPILE)
+    find_package(LLVM 3.9 REQUIRED support core irreader executionengine passes target nativecodegen)
```
kinke: I guess this conflicts with https://github.com/ldc-developers/ldc/blob/master/CMakeLists.txt#L22 and thus prevents successful linking.

runtime/CMakeLists.txt
```diff
@@ -600,11 +601,17 @@ macro(build_all_runtime_variants d_flags c_flags ld_flags path_suffix outlist_ta
         get_target_suffix("" "${path_suffix}" target_suffix)
         set_common_library_properties(ldc-profile-rt${target_suffix})
     endif()
+
+    build_jit_runtime ("${d_flags}" "${c_flags}" "${ld_flags}" "${path_suffix}" ${outlist_targets})
```
kinke: Any reason you're not including `${D_FLAGS}` and maybe `${D_FLAGS_RELEASE}` here?

ddmd/globals.h
Hardcode84: done

runtime/CMakeLists.txt
Hardcode84: added

runtime/jit-rt/DefineBuildJitRT.cmake
```diff
@@ -1,5 +1,5 @@
 if(LDC_RUNTIME_COMPILE)
-    find_package(LLVM 3.9 REQUIRED support core irreader executionengine passes target nativecodegen)
+    find_package(LLVM 3.7 REQUIRED support core irreader executionengine passes target nativecodegen)
```
kinke: What I meant was that invoking find_package() twice probably doesn't work and that conditionally including your additional libs in the single invokation should, although at the cost of less modularity.

Hardcode84: I noticed another thing, it fail to find 32-bit llvm libraries when MULTILIB is ON, but as I understand compiler itself will still be 64 bit in this case. Does these machines have 32-bit llvm libs?

driver/linker-gcc.cpp
```diff
@@ -340,10 +340,12 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
     args.push_back("-lldc-profile-rt");
   }
 
-  if (global.params.enableRuntimeCompile) {
+#if defined(LDC_RUNTIME_COMPILE)
```
JohanEngelen: nitpick: to prevent all these `#if` things, you could create a function `isRuntimeCompileEnabled()` similar to `isUsingLTO()`.

runtime/jit-rt/DefineBuildJitRT.cmake
kinke: The compiler will be 64-bit, no 32-bit LLVM available and the 32-bit libs definitely shouldn't be needed; only the LDC executable needs to be linked to LLVM. So this would indicate that some compiler/linker flags leak into other build targets (e.g., the 32-bit shared runtime libs).

Hardcode84: I do dynamic compilation through llvm in my jit runtime shared library, which then linked to the target application, and if this application is 32-bit then jit runtime and llvm libs also have to be 32-bit.

kinke: Ah yes that makes perfect sense. I guess you'll have to build the native JIT runtime library exclusively then for now, excluding the 32-bit one for MULTILIB builds.

driver/linker-gcc.cpp
Hardcode84: Done

Hardcode84:
Another attempt to fix libraries list _almost_ worked. There is `llvm_map_components_to_libnames`  function but it is only available when we do `find_library(LLVM ... CONFIG)`. I tried to manually include `LLVMConfig.cmake` and `LLVM-Config.cmake` to get `llvm_map_components_to_libnames` and this worked everywhere except windows (it works on my local windows machine though).
I do not want to use explicit libraries list because different llvm versions return different libraries list from same  component list.
Also all jit test are failing on osx because they either can't find symbols in jitted code or just segfaults and I don't osx machine to debug this.

Hardcode84:
@kinke @klickverbot @thewilsonator @JohanEngelen 
I disabled some thread local and exceptions jit tests, remaining should be green now.
Current status:
* thread locals in jit broken on all platforms
* Exceptions in jit broken on windows but work on linux and osx

So, what are next steps?

kinke:
The next step is to congratulate you, nice progress, and to start annoying you with obligatory nitpicks. ;)

driver/cl_options.cpp
```diff
@@ -456,6 +456,13 @@ cl::opt<std::string>
                        cl::value_desc("prefix"));
 #endif
 
+#if defined(LDC_RUNTIME_COMPILE)
+cl::opt<bool> enableRuntimeCompile(
+    "enable-runtime-compile",
+    cl::desc("Enable runtime compilation"),
```
kinke: I don't know if that lengthy option is clear enough. I'd prefer to have JIT in there (just `-enable-jit`?). I'd also think that `@jit` for the UDA would be clear enough and less clumsy.

driver/cl_options.h
```diff
@@ -121,5 +121,12 @@ extern cl::opt<std::string> saveOptimizationRecord;
 extern cl::list<std::string> dcomputeTargets;
 extern cl::opt<std::string> dcomputeFilePrefix;
 #endif
+
+#if defined(LDC_RUNTIME_COMPILE)
+extern cl::opt<bool> enableRuntimeCompile;
+inline bool isRuntimeCompileEnabled() { return enableRuntimeCompile; }
+#else
+inline bool isRuntimeCompileEnabled() { return false; }
+#endif
```
kinke: I'd get rid of the extra function and just use a `constexpr enableRuntimeCompile` if the define is missing.

gen/irstate.h
```diff
@@ -219,6 +219,14 @@ struct IRState {
...
   void replaceGlobals();
 
+  struct RtCompiledFuncDesc {
+    llvm::GlobalVariable* thunkVar;
+    llvm::Function* thunkFunc;
+  };
+
+  std::map<llvm::Function*,RtCompiledFuncDesc> runtimeCompiledFunctions;
+  std::set<IrGlobal*> runtimeCompiledVars;
+
```
kinke: clang-format

runtime/jit-rt/DefineBuildJitRT.cmake
```diff
@@ -0,0 +1,98 @@
...
+    file(GLOB LDC_JITRT_D ${JITRT_DIR}/d/ldc/*.d)
+
+    # Choose the correct subfolder depending on the LLVM version
+    file(GLOB LDC_JITRT_CXX ${JITRT_DIR}/cpp/*.cpp)
+    file(GLOB LDC_JITRT_SO_CXX ${JITRT_DIR}/cpp-so/*.cpp)
+
+    # Set compiler-dependent flags
+    if(MSVC)
+        # Omit Default Library Name from the library, so it will work with both release and debug builds
+        set(JITRT_EXTRA_FLAGS "/Zl")
```
kinke: No `/O<N>` for MSVC?

runtime/jit-rt/cpp-so/compile.cpp
```diff
@@ -0,0 +1,380 @@
...
+  if (0x1 == name[0]) {
+    return name.substr(1);
+  }
+  return "_" + name;
+#else
+  return name;
+#endif
+}
+
+auto getSymbolInPorcess(const std::string& name)
```
kinke: P**r**ocess

runtime/jit-rt/cpp-so/compile.cpp
```diff
@@ -0,0 +1,380 @@
...
+#include "llvm/Support/DynamicLibrary.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/TargetSelect.h"
+#include "llvm/Support/Host.h"
+#include "llvm/Target/TargetMachine.h"
+#include "llvm/Bitcode/BitcodeReader.h"
+
+namespace {
+
+#pragma pack(push,1)
```
kinke: I thought gcc wouldn't support that syntax. It wouldn't be required if you replaced the `int ...Size` members below with `size_t`, so that all fields are of pointer-size and no unnatural alignment results.

runtime/jit-rt/cpp-so/compile.cpp
```diff
@@ -0,0 +1,380 @@
...
+  void* sym;
+};
+
+struct RtCompileVarList
+{
+  const char* name;
+  const void* init;
+};
+
+struct RtComileModuleList
```
kinke: RtCom**p**ile

runtime/jit-rt/cpp-so/context.h
```diff
@@ -0,0 +1,26 @@
...
+#define CONTEXT_H
+
+#include <cstddef> //size_t
+
+// must be synchronized with D source
+typedef void (*InterruptPointHandlerT)(void*, const char* action, const char* object);
+typedef void (*FatalHandlerT)(void*, const char* reason);
+typedef void (*DumpHandlerT)(void*, const char* str, std::size_t len);
+
+#pragma pack(push,1)
```
kinke: Superfluous afaict.

runtime/jit-rt/d/ldc/runtimecompile.d
```diff
@@ -0,0 +1,109 @@
+module ldc.runtimecompile;
+
+version(LDC_RuntimeCompilation)
+{
```
kinke: `version(...):` gets rid of the extra scope and makes it absolutely clear the whole module is only defined for that version.

runtime/jit-rt/d/ldc/runtimecompile.d
```diff
@@ -0,0 +1,109 @@
...
+  auto del = cast(DelType*)context;
+  assert(buff !is null);
+  (*del)(buff[0..len]);
+}
+
+
+// must be synchronized with cpp
+align(1) struct Context
+{
+align(1):
```
kinke: The outer `align(1)` is bogus, and the inner one can be safely omitted.

tests/runtimecompile/globals_types.d
```diff
@@ -0,0 +1,94 @@
...
+}
+
+struct Bar
+{
+  Foo f;
+  Foo* pf;
+  Bar* pb;
+}
+
+@runtimeCompile __gshared byte  i8  = 42 + 1;
```
kinke: A UDA can be applied to a scope: `@runtimeCompile { ... }`

driver/cl_options.h
Hardcode84: done

gen/irstate.h
Hardcode84: done

runtime/jit-rt/DefineBuildJitRT.cmake
Hardcode84: Actually, it wasn't needed as it was already in flags (and next -O3 too).  Removed O3

runtime/jit-rt/cpp-so/compile.cpp
Hardcode84: fixed

runtime/jit-rt/cpp-so/compile.cpp
Hardcode84: fixed

runtime/jit-rt/d/ldc/runtimecompile.d
Hardcode84: done

tests/runtimecompile/globals_types.d
Hardcode84: done

driver/cl_options.cpp
Hardcode84: As I said before it is not a true jit because user need to do explicit `compileDynamicCode()` call before  any dynamic functions usage. But if everyone okay with that, I can change name of course.

Hardcode84:
About thread locals issues, they actually can be workarounded by generating non-dynamic functions accessing thread locals and replacing direct thread locals access in dymanic code with calls to these functions.
But I am not sure I have enough llvm knowledge to do this automatically. And user also cannot do this because in current design all functions called from dynamic code are aggressively marked as dynamic.

Hardcode84:
TLS now work

runtime/jit-rt/cpp-so/compile.cpp
Hardcode84: Well, working with size_t-like types on llvm side is just too ugly. Let's stick to current solution for now?

runtime/jit-rt/d/ldc/runtimecompile.d
Hardcode84: done

runtime/jit-rt/cpp-so/context.h
Hardcode84: removed

Hardcode84:
Ok here is another issue:
```
int foo()
{
  return 42;
}

alias fptr = int function();

@runtimeCompile fptr get_foo_ptr()
{
  // We create jit version of each function used in @runtimeCompile functions when possible
  // so jit will have it's own version of foo()
  return &foo;
}

assert(foo() == get_foo_ptr()()); // success
assert(&foo is get_foo_ptr()); // failure, get_foo_ptr() returned pointer to jitted code
```
@kinke how critical is this?

kinke:
Well I guess such behavior (potentially multiple identities of functions) is to be expected from your feature. I wouldn't have expected TLS to work either, at least not at such an early stage already, so I don't think it's critical at all.

What use cases do you have in mind for this JIT work? Optimizing small performance-critical code parts for the used CPU, or something bigger?

Hardcode84:
> Optimizing small performance-critical code parts for the used CPU

For the initial version this will be main usecase but in the future I am planning to add more high level compilation controls features.
For now `@runtimeCompile` variables is a single example of them (but very limited):
```
@runtimeCompile _gshared int flag;

// In some 'normal' code
flag = 2;
compileDynamicCode();

@runtimeCompile void foo()
{
  // flag is constant to jitted code
  // this check and branch will be removed
  if (1 == flag)
  {
    // Some code
  }
  else if (2 == flag)
  {
    // Some other code
  }
}
```

Hardcode84:
And another issue: LDC sets `target-cpu` and `target-features` for all functions in `applyTargetMachineAttributes` and then replaces them with values from user attributes if any. I want to replace them with jit host values but respect user values if they were set explicitly.
@kinke any ideas how to do it properly?

Hardcode84:
More specific question - how to get `IrFunction` or `FuncDeclaration` from function name?

Hardcode84:
Ok, it works, but I had to add
```
  // List of functions with cpu or features attributes overriden by user
  std::vector<IrFunction *> targetCpuOrFeaturesOverriden;
```
field to `IRState`

Hardcode84:
This should be final version for now, do not expect major changes.

Also, we need to decide about naming: `@runtimeCompile` vs `@jit`, I am ok with both variants.

Hardcode84:
ping

gen/functions.cpp
```diff
@@ -566,6 +567,15 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
   applyTargetMachineAttributes(*func, *gTargetMachine);
   applyFuncDeclUDAs(fdecl, irFunc);
 
+  if(irFunc->runtimeCompile) {
+    declareRuntimeCompiledFunction(gIR, irFunc);
+  }
+
+  if (irFunc->targetCpuOverriden ||
+      irFunc->targetFeaturesOverriden) {
+    gIR->targetCpuOrFeaturesOverriden.push_back(irFunc);
```
kinke: Overrid**d**en.

kinke:
Pinging @klickverbot and @JohanEngelen about attribute naming.

@Hardcode84: A design overview in the source would be nice, please at least include a link to your [earlier post](https://github.com/ldc-developers/ldc/pull/2293#issuecomment-325820726) in a prominent spot.

I'm for including it then as experimental feature in v1.6.

Hardcode84:
@kinke thanks.

About API, there is `CompilerSettings` struct with `optLevel ` and `sizeLevel ` fields (which are same as compiler -O command line options), and both set default values to 0 (no optimization). This is consistent with the compiler command line but is this a good thing? Some people will probably forget to set them to some value and will run their jit code without optimizations.

Also, I added `%lib` substitution to lit tests (maps to `.lib` on windows and `.a` on other platforms).

andrewbenton:
@Hardcode84 I haven't worked on this at all, so I really have no vote, but if I did have one, I'd vote for @jit since it's shorter and simpler.

Hardcode84:
Investigating weird win32 issue in one of the recently added tests (writeln throws UTFException when called from jitted code but works fine when called from static code)

JohanEngelen:
> Also, I added %lib substitution to lit tests (maps to .lib on windows and .a on other platforms).

You can make a separate PR for that if you want to reduce the size of this PR. It's just cluttering the diffs. Also add it to the Wiki page when pulled, thanks!

JohanEngelen:
As for "jit" or "runtimeCompile", I don't have a preference (you could ask the forums, but not sure whether you'll get high-quality answers). Once it is decided on one or the other, please make the naming consistent throughout. There's also the option of naming it `runtimeCompile` (perhaps easier to grasp throughout codebase, and perhaps more true to the nature of things like, you mention), and providing an `@jit` alias for brevity.

Hardcode84:
Ok, about this `writeln` failure, jit fails to properly initialize `spec` variable in `formattedWrite`
```
auto spec = FormatSpec!Char(fmt);
```
Static compiler output:
```
00C71111  vmovups     ymm0,ymmword ptr [_D3std6format18__T10FormatSpecTaZ10FormatSpec6__initZ (0D0DB10h)]  
00C71119  vmovaps     ymmword ptr [spec],ymm0   // 00EBFB60
00C71122  vmovups     ymm0,ymmword ptr ds:[0D0DB20h]  
00C7112A  vmovups     ymmword ptr [esp+0D0h],ymm0  // 00EBFB70
```
Jit output:
```
008500B1  vmovups     ymm0,ymmword ptr ds:[0D0DB10h]  
008500B9  vmovaps     ymmword ptr [esp+40h],ymm0  
008500BF  mov         edi,dword ptr [ebp+18h]  
008500C2  vmovups     ymm0,ymmword ptr ds:[0D0DB10h] /// !!!
008500CA  vmovups     ymmword ptr [esp+50h],ymm0  
```
Also, I found that `addPassesToEmitFile` (which I use for debug dumps) outputs different assembly than I see in debugger (which is outputted by `addPassesToEmitMC` in jit)
And `addPassesToEmitFile` actually output correct assembly :(
And this is reproducible only on win32 with -O0 and avx enabled.

Hardcode84:
Reproducer:
```
struct Foo
{
  int[12] arr = [0,1,2,3,4,5,6,7,8,9,10,11];
}

Foo foo()
{
  auto f = Foo();
  return f;
}

@runtimeCompile Foo bar()
{
  auto f = Foo();
  return f;
}

void main(string[] args)
{
  compileDynamicCode();
  auto f1 = foo();
  auto f2 = bar();
  assert(f1 == f2);
}
```
Static:
```
00E91000  vmovups     ymm0,ymmword ptr ds:[0EED2D0h]  
00E91008  vmovups     ymmword ptr [eax+10h],ymm0  
00E9100D  vmovups     ymm0,ymmword ptr ds:[0EED2C0h]  
00E91015  vmovups     ymmword ptr [eax],ymm0  
00E91019  vzeroupper  
00E9101C  ret  
```
Jit:
```
008C0000  vmovups     ymm0,ymmword ptr ds:[0EED2F0h]  
008C0008  vmovups     ymmword ptr [eax+10h],ymm0  
008C000D  vmovups     ymm0,ymmword ptr ds:[0EED2F0h]  
008C0015  vmovups     ymmword ptr [eax],ymm0  
008C0019  vzeroupper  
008C001C  ret  
```
IR:
```
define x86_stdcallcc void @"\01_D4test3barFZS4test3Foo"(%test.Foo* inreg noalias align 4) #1 comdat {
  %2 = getelementptr inbounds %test.Foo, %test.Foo* %0, i32 0, i32 0
  %3 = bitcast [12 x i32]* %2 to i8*
  call void @llvm.memcpy.p0i8.p0i8.i32(i8* %3, i8* bitcast ([12 x i32]* @.arrayliteral.1 to i8*), i32 48, i32 1, i1 false)
  ret void
}
```

Hardcode84:
Current status:

* Exceptions doesn't work on windows if there are jitted functions in stack
* Thread local access issues wokarounded
* jit asm dump is disabled until I figure out good way to do it
* Invalid codegen in jit for struct initialization on win32 avx (have no idea why, added a test and disabled it on win32, other platforms seems fine)

About naming, I like the idea of `@jit` as alias to `@runtimeCompile`. Also it will be forward compatible, if we decide to do a true jit in the future (where `@runtimeCompile` will be need to be explicitly compiled, and `@jit` will be a true jit).

latest design document (including `@runtimeCompile` const variables and TLS workaround): https://gist.github.com/Hardcode84/fe986a3c1219e40a88ee4b67f44ea386

It is quite big to be included in cpp, maybe just put `.md` file somewhere in the repo?

Ping @kinke @JohanEngelen

gen/functions.cpp
Hardcode84: Fixed

Hardcode84:
https://github.com/ldc-developers/druntime/pull/105

Attribute renaming PR

JohanEngelen:
> It is quite big to be included in cpp, maybe just put .md file somewhere in the repo?

We could put such documentation in `./docs/` ? (that's what LLVM does)

tests/runtimecompile/struct_init.d
```diff
@@ -0,0 +1,76 @@
+
+// llvm generates overlapped simd reads and writes to init these
+// structs but fails on win32 for some reason
+// UNSUPPORTED: Windows_x86
```
JohanEngelen: Better to mark this XFAIL, so it's clear that it should be fixed at some point (and when it suddenly starts working we get a ping that we should update the documentation that it works).

Probably good to add a warning when targeting Windows_x86 with `-enable-runtime-compile`? (I understand that you don't want to disable it for that target, to enable testing things that do work?)

driver/cl_options.cpp
```diff
@@ -456,6 +456,19 @@ cl::opt<std::string>
...
+    cl::init(false));
+
+cl::opt<bool> runtimeCompileTlsWorkaround(
+    "runtime-compile-tls-workaround",
```
JohanEngelen: no test yet for this cmdline arg.

gen/runtimecompile.h
```diff
@@ -0,0 +1,13 @@
+#ifndef RUNTIMECOMPILE_H
```
JohanEngelen: Add the standard file header with copyright, etc.
(also prefix the define with `LDC_GEN_`)

gen/runtimecompile.cpp
```diff
@@ -0,0 +1,816 @@
+#include "runtimecompile.h"
```
JohanEngelen: add `gen/` to path

add the standard file header with copyright, etc.

runtime/jit-rt/cpp-so/callback_ostream.cpp
```diff
@@ -0,0 +1,11 @@
+
```
JohanEngelen: You should add copyright info to all these new files.
If you copied code from LLVM/Clang, mention that too (see e.g. `gen/pgo.h`)

runtime/jit-rt/cpp-so/optimizer.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include "context.h"
+#include "utils.h"
+#include "valueparser.h"
+
+namespace {
+
+void addOptimizationPasses(llvm::legacy::PassManagerBase &mpm,
+                           llvm::legacy::FunctionPassManager &fpm,
+                           unsigned optLevel, unsigned sizeLevel) {
+  //  if (!noVerify) {
```
JohanEngelen: don't leave commented out code, just remove it.
Instead, you could add TODO comments with references to other pieces of code as a hint/suggestion for future development.

Hardcode84:
About licenses, compiler licensed under BSD and phobos/druntime under Boost license, what license should I pick for jit runtime?

tests/runtimecompile/struct_init.d
Hardcode84: Made tests XFAIL

driver/cl_options.cpp
Hardcode84: Added simple test, checking compiler accepts this option

gen/runtimecompile.h
Hardcode84: Done

gen/runtimecompile.cpp
Hardcode84: Done

runtime/jit-rt/cpp-so/optimizer.cpp
Hardcode84: Made some cleanup here, I will think how to share optimization setup code between compiler and jit runtime.

Hardcode84:
Added document to /docs/

Hardcode84:
@kinke @klickverbot @JohanEngelen 

About naming again. I like @JohanEngelen idea of keeping `@runtimeCompile` with `@jit` alias. Also I want to separate `@runtimeCompile` for functions and `@runtimeCompileConst` for variables per @klickverbot advice in previous PR.

If everyone are ok with this, please merge druntime PR and I will update this PR.
ldc-developers/druntime#105

Also, I will rename command line option to `-enable-jit`. Unlike the code, there are no reasons to keep separate options for current implementation and possible future true jit.

kinke:
Sounds good to me. Any reason the jit UDAs aren't aliases, but separate symbols?

Hardcode84:
> Any reason the jit UDAs aren't aliases, but separate symbols?

No

Hardcode84:
I am ok with any name now :)

kinke:
Sounds like the best solution so far; I didn't realize I had more of a problem with the term **runtime**Compile (=> druntime in my brain) than with its actual length, as usage is clearly expected to be rare.

Hardcode84:
Updated renaming PR

https://github.com/ldc-developers/druntime/pull/105

Hardcode84:
druntime submodule update:
https://github.com/ldc-developers/ldc/pull/2404

kinke:
Why a separate PR? Just update the submodule in here.

Hardcode84:
ok

Hardcode84:
Renamed EVERYWHERE

kinke:
Sorry to bother you hopefully one last time, but there's still `docs/runtime_compile.md` left. ;)

CMakeLists.txt
```diff
@@ -383,6 +383,17 @@ message(STATUS "Building LDC with PGO support")
 append("-DLDC_WITH_PGO" LDC_CXXFLAGS)
 
 #
+# Enable Runtime compilation if supported for this platform and LLVM version.
+# LLVM >= 3.9 is required
+#
+set(LDC_DYNAMIC_COMPILE False)  # must be a valid Python boolean constant (case sensitive)
+if (NOT (LDC_LLVM_VER LESS 400))
+    message(STATUS "Building LDC with runtime compilation support")
```
klickverbot: "with dynamic compilation support" for consistency.

LICENSE
```diff
@@ -23,6 +23,9 @@ Libraries (lib/ and import/ in binary packages):
    Phobos (runtime/phobos), is distributed under the terms of the Boost
    Software License. See the individual source files for author information.
 
+ - The jit runtime library is distributed under the terms of the Boost
```
klickverbot: I'd make this "The ldc-jit-rt [or whatever library it really is] library is distributed under the terms of the Boost Software license (but additionally makes use of LLVM code)." or something like that for clarity – unless I'm missing something, people need to distribute the LLVM copyright notice when they are using dynamic compile support.

runtime/jit-rt/cpp-so/optimizer.cpp
```diff
@@ -0,0 +1,176 @@
...
+  assert(nullptr != init);
+  auto var = module.getGlobalVariable(name);
+  if (nullptr != var) {
+    auto type = var->getType()->getElementType();
+    auto initializer =
+        parseInitializer(context, module.getDataLayout(), type, init);
+    var->setConstant(true);
+    var->setInitializer(initializer);
+    var->setLinkage(llvm::GlobalValue::PrivateLinkage);
+    //    auto tempVar = new llvm::GlobalVariable(
```
klickverbot: What's the deal with this (and other) commented out sections?

runtime/jit-rt/d/ldc/dynamic_compile.d
```diff
@@ -0,0 +1,122 @@
+/**
+ * Contains dynamic compilation API.
+ *
+ * Copyright: Authors 2017
```
klickverbot: This should probably at least say something vague like "the LDC team", or your name.

CMakeLists.txt
Hardcode84: Done

LICENSE
Hardcode84: Done

runtime/jit-rt/cpp-so/optimizer.cpp
Hardcode84: Removed

runtime/jit-rt/d/ldc/dynamic_compile.d
Hardcode84: This was copypasted from somewhere, fixed to "the LDC team". But I don't like these 'authors' fields because everyone forget to update them.

Hardcode84:
More renamings and cleanups

kinke:
Thanks Ivan. I'll merge this tomorrow afternoon (CET) before tagging v1.6.0-beta1 if there are no objections.

kinke:
Feasible options being 1 or 93, I opted for the full bunch. I'll gladly let others make decisions like this in the future, I gave a 24+h notice.

Hardcode84:
Thanks for review. Actually, I wanted to ask about squashing but I forgot.

PR #2294, opened 2017-08-27
DI: Don't use llvm.dbg.declare with non-alloca storage

kinke:
An alternative to #2292.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -113,22 +113,23 @@ ldc::DIScope ldc::DIBuilder::GetCurrentScope() {
   return fn->diLexicalBlocks.top();
 }
 
-void ldc::DIBuilder::Declare(const Loc &loc, llvm::Value *var,
+// Use this to set the memory address for a debuginfo variable.
+// LLVM requires the storage to be an alloca.
+void ldc::DIBuilder::Declare(const Loc &loc, llvm::Value *storage,
                              ldc::DILocalVariable divar,
                              ldc::DIExpression diexpr) {
   unsigned charnum = (loc.linnum ? loc.charnum : 0);
   auto debugLoc = llvm::DebugLoc::get(loc.linnum, charnum, GetCurrentScope());
-  DBuilder.insertDeclare(var, divar, diexpr, debugLoc, IR->scopebb());
+  DBuilder.insertDeclare(storage, divar, diexpr, debugLoc, IR->scopebb());
 }
 
-// Use this to tag debuginfo onto a non-alloca LLVM value, e.g. a function
-// parameter.
-void ldc::DIBuilder::DbgValue(const Loc &loc, llvm::Value *var,
+// Use this to set the (current) value for a debuginfo variable.
+void ldc::DIBuilder::SetValue(const Loc &loc, llvm::Value *value,
                               ldc::DILocalVariable divar,
                               ldc::DIExpression diexpr) {
   unsigned charnum = (loc.linnum ? loc.charnum : 0);
   auto debugLoc = llvm::DebugLoc::get(loc.linnum, charnum, GetCurrentScope());
-  DBuilder.insertDbgValueIntrinsic(var,
+  DBuilder.insertDbgValueIntrinsic(value,
 #if LDC_LLVM_VER < 600
                                    0,
 #endif
@@ -1149,7 +1150,7 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
   variableMap[vd] = debugVariable;
 
   if (useDbgValueIntrinsic) {
-    DbgValue(vd->loc, ll, debugVariable,
+    SetValue(vd->loc, ll, debugVariable,
              addr.empty() ? DBuilder.createExpression()
                           : DBuilder.createExpression(addr));
   } else {
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -121,6 +121,20 @@ void ldc::DIBuilder::Declare(const Loc &loc, llvm::Value *var,
   DBuilder.insertDeclare(var, divar, diexpr, debugLoc, IR->scopebb());
 }
 
+// Use this to tag debuginfo onto a non-alloca LLVM value, e.g. a function
+// parameter.
+void ldc::DIBuilder::DbgValue(const Loc &loc, llvm::Value *var,
+                              ldc::DILocalVariable divar,
+                              ldc::DIExpression diexpr) {
+  unsigned charnum = (loc.linnum ? loc.charnum : 0);
+  auto debugLoc = llvm::DebugLoc::get(loc.linnum, charnum, GetCurrentScope());
+  DBuilder.insertDbgValueIntrinsic(var,
+#if LDC_LLVM_VER < 600
+                                   0,
+#endif
+                                   divar, diexpr, debugLoc, IR->scopebb());
+}
+
 ldc::DIFile ldc::DIBuilder::CreateFile(Loc &loc) {
   const char* filename = loc.filename;
   if (!filename)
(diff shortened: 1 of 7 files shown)


kinke:
It becomes more and more evident that LLVM doesn't like non-allocas for debuginfo params and locals. That's especially proving to be problematic for Win64 with its ExplicitByvalRewrite, where the storage for a parameter is an alloca of the call**er** and that pointer is passed as LL pointer parameter (and used directly for the DI local).
We may need to represent the params rewritten by ExplicitByvalRewrite as DI references to outer variables to fix this.

JohanEngelen:
Does this work for optimized debug info?  I read that metadata does not count as use of the alloca,
 so perhaps the alloca is stripped immediately and then the debuginfo is lost, see http://nondot.org/sabre/LLVMNotes/DebugInfoVariableInfo.txt .  I think we can use `@llvm.dbg.value` instead to point directly to the non-alloca variable.

kinke:
> I think we can use @llvm.dbg.value instead to point directly to the non-alloca variable.

This definitely sounds interesting. We have the same issue with nested variables, those aren't allocas either.

JohanEngelen:
Yeah, I am working on trying with llvm.dbg.value now.

kinke:
I hoped all that needed to be constant is the storage/lvalue of the variable. The [official DI docs](http://llvm.org/docs/SourceLevelDebugging.html) are, hmm, crappy; it shouldn't be that hard to associate a debuginfo variable with its (constant) LL lvalue?!

JohanEngelen:
> I hoped all that needed to be constant is the storage/lvalue of the variable. 

When the value is constant, it's storage (LLVM value) is also constant and all is well. When a value is not constant, on every write you'd get a new LLVM value (because of SSA), and then you'd have to retag that new value with llvm.dbg.value (because the LLVM value associated with the D variable has changed). But for non-constant things, we always use an alloca (right?), and we can use llvm.dbg.declare.

kinke:
Now based on reworked #2292, and the first results are fantastic, no issues with struct params > 64 bit anymore on Win64. I still have to check the nested vars, but I'm very optimistic.

gen/dibuilder.h
```diff
@@ -141,6 +141,8 @@ class DIBuilder {
   DIScope GetCurrentScope();
   void Declare(const Loc &loc, llvm::Value *var, ldc::DILocalVariable divar,
                ldc::DIExpression diexpr);
+  void DbgValue(const Loc &loc, llvm::Value *var, ldc::DILocalVariable divar,
```
klickverbot: Shouldn't this be just `Value` in line with `Declare` (or rename both)?

gen/dibuilder.cpp
```diff
@@ -1040,12 +1054,22 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
...
-  if (vd->isRef() || vd->isOut()) {
+  // llvm.dbg.declare only works properly with local allocas.
+  // Represent variables with non-alloca LL storage as DI references, like `ref`
+  // and `out` parameters. Their values are constant, so we don't have to attach
+  // the debug information to a memory location and can use llvm.dbg.value
+  // instead.
+  const bool storedInAlloca = llvm::dyn_cast<llvm::AllocaInst>(ll);
+  bool useDbgValueIntrinsic = false;
+  if (!storedInAlloca || vd->isRef() || vd->isOut()) {
+    useDbgValueIntrinsic = !storedInAlloca;
```
klickverbot: I find this a bit confusing to read, especially given the comment – is there a situation where `ref` and `out` actually do use an `alloca`?

kinke: See Johan's original comment in #2292. Apparently used for special-ref loop variables. It's a bit smelly and should be definitely looked into sometime.

klickverbot: Oh yes, they are fun – I called them `SpecialRefVar`s in our code precisely because they don't map to anything else in the D language. In terms of looking into things, I've been wondering for a while how much of the extra `DValue` metadata we could do away with given that we have LLVM IR type information and frontend type information already. The current design seems needlessly complicated, although I can't really pinpoint any specific points yet.

kinke:
`DbgValue()` renamed to `SetValue()` and comments slightly reworked. Nested variables definitely still need to be fixed though, working on it.

kinke:
Unfortunately, this approach apparently doesn't work for nested vars **at all**, neither on Win64 nor on Linux x64. The nested vars don't show up in the VS debugger anymore (previously they were shown, although each with some attached error message); in gdb the same (symbols still known, but not listed in `info locals`, `optimized out` when trying to print them), where they previously seemed to work pretty well. No LLVM assertions, no failing tests, but fully broken debuginfos for nested vars, argh.

kinke:
An example from [tests/debuginfo/nested.d](https://github.com/ldc-developers/ldc/blob/master/tests/debuginfo/nested.d#L21): `arg0` captured in `nested()` is represented as DI reference with a constant value (the `arg0` storage, a GEP into the parent frame):
```LL
%arg0 = getelementptr inbounds %nest.encloser, %nest.encloser* %2, i32 0, i32 0 ; [#uses = 1, type = i32*]
call void @llvm.dbg.value(metadata i32* %arg0, i64 0, metadata !28, metadata !16), !dbg !34 ; [debug line = 1:6] [debug variable = arg0]

!11 = !DIDerivedType(tag: DW_TAG_reference_type, baseType: !12, size: 64, align: 32)
!12 = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
!16 = !DIExpression()
!28 = !DILocalVariable(name: "arg0", scope: !24, file: !3, line: 1, type: !11)
```

This results in the following naked DWARF var without any value on Linux x64 (via `dwarfdump`):
```
< 2><0x000000a0>      DW_TAG_variable
                        DW_AT_name                  "arg0"
                        DW_AT_decl_file             0x00000001 /home/martin/current.d
                        DW_AT_decl_line             0x00000007
                        DW_AT_type                  <0x0000011c>
```

kinke:
I tried using `DW_OP_stack_value` as additional expression. Quoting official DWARF docs:

> The DW_OP_stack_value operation specifies that the object does not exist in memory but its value is nonetheless known and is at the top of the DWARF expression stack. In this form of location description, the DWARF expression represents the actual value of the object, rather than its location.

Result on Linux x64: no change at all, the vars still only feature a name, type and source location.

rainers:
> Unfortunately, this approach apparently doesn't work for nested vars at all, neither on Win64 nor on Linux x64.

Not sure if it is relevant to your approach, but non-trivial DWARF expressions were unsupported on Windows for ages until a month ago https://github.com/llvm-mirror/llvm/commit/2cd77a84868638c82bda5b620838e775dbf4997c added "rudimentary DwarfExpression support to CodeView".

kinke:
Thanks, but I guess an expression shouldn't be really necessary here; I mean it's working for immediate pointers from parameters (ref/out params, params rewritten by ExplicitByvalRewrite and params with the LLVM `byval` attribute), but `llvm.dbg.value` apparently (silently) hates a GEP value.

kinke:
Now trying to avoid all GEPs and using an alloca + more complex DWARF expression instead. Seems to work fine on Linux x64; the VS debugger on Win64 displays closure parameters and locals correctly, but the captured ones (in nested functions) are missing, possibly due to a too complex DWARF expression for LLVM 4.

JohanEngelen:
Related discussion on LLVM-dev, pointing out some of the limitations of the current debuginfo LLVM code: https://lists.llvm.org/pipermail/llvm-dev/2017-September/117141.html

kinke:
Yeah with the alloca+expression approach, it's #1933 all over again; LLVM < 3.9 fails for our testcase. We speculated about it being an LLVM bug back then, and apparently later other testcases came up which also failed with LLVM 3.9, so that we went ahead with the GEP + weird additional dereference approach.

Also considering that LLVM < 5 misused `DW_OP_plus` as `DW_OP_plus_uconst` (different 'signature', I checked the DWARF standard while trying to fix the CI builds) is not really making me feel any more comfortable about LLVM's debuginfos.

The original issue triggering this again, LLVM 5 complaining about the LL pointer param as DI storage for non-nested ref/out params, was definitely a bug on our side, as it was declared as reference, but we passed the reference rvalue instead of its lvalue.

I'll probably split off the fix as a separate PR, as this is still clearly WIP and I'm not sure the alloca+expression route is the proper way of handling this, or whether LLVM is supposed to swallow GEPs just fine and we're just having a hard time trying to work around it 'properly'.

JohanEngelen:
> I'll probably split off the fix as a separate PR

Sounds good.

kinke:
Obsoleted by #2315.

PR #2299, opened 2017-08-30
[MSVC] Error out if LLVM is built with MD[d]

thewilsonator:
So people like me don't waste days staring at linker errors.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,11 +145,9 @@ if(MSVC)
 
     # Link against the static MSVC runtime; CMake's C(++) flags apparently default to the dynamic one.
     # Host DMD/LDMD already defaults to linking against the static MSVC runtime.
-    # NOTE: Requires LLVM to be built with CMake variable LLVM_USE_CRT_{RELEASE,DEBUG,...}=MT[d].
-    if(${LLVM_CXXFLAGS} MATCHES "MD")
-        message(FATAL_ERROR "LLVM must be built with LLVM_USE_CRT_{RELEASE,DEBUG,...}=MT[d], not MD[d]")
+    if(${LLVM_CXXFLAGS} MATCHES "(^| )/MDd?( |$)")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -150,6 +150,11 @@ if(MSVC)
     # Link against the static MSVC runtime; CMake's C(++) flags apparently default to the dynamic one.
     # Host DMD/LDMD already defaults to linking against the static MSVC runtime.
     # NOTE: Requires LLVM to be built with CMake variable LLVM_USE_CRT_{RELEASE,DEBUG,...}=MT[d].
+    foreach(llvm_crt LLVM_USE_CRT_RELEASE LLVM_USE_CRT_DEBUG)
```
Hardcode84: There are also `LLVM_USE_CRT_RELWITHDEBINFO` and `LLVM_USE_CRT_MINSIZEREL` (basically for each possible value in `CMAKE_BUILD_TYPE`)

kinke: Yeah, but requiring to set them all is a PITA. Only the one for the currently selected CMake config is required.

thewilsonator: I hope that it will pick up on any that are set rather than require all to be set. I'm not sure if this is the right way to do it, my CMake knowledge is rather limited. I couldn't figure out if I should be using `if(DEFINED ...)` instead.

kinke: I just checked llvm-config, and its CXX flags include `/MT`. So checking that `LLVM_CXXFLAGS` includes `/MT[d]` should do the job.

thewilsonator: Ah, great.

kinke:
FWIW, I pushed an exact check and edited the Wiki page. Thanks for immediately working on a fix and thinking about other users. :)

thewilsonator:
Well I was rather determined to get it working, but I suspect that other people would have given up after the first couple of tries. Given how simple the fix is there is no reason to not generate a nicer error message. 

My regex-fu is pretty bad, so I'm going to assume that it is correct and I'll lave this for you to merge than you're satisfied that it works.

JohanEngelen:
is this good to merge?

PR #2305, opened 2017-09-02
Fix LLVM 6.0 compilation.

JohanEngelen:


JohanEngelen:
@thewilsonator And another update needed for LLVM6.0 on OSX.

PR #2308, opened 2017-09-04
Fix #2307

### LINKED ISSUE #2307: [DCompute] Fails when host code has thread local globals
DCcompute @ 9c50bd97633478bb901e534b4c708bf2da4be261 fails to build due to a thread local global during ISel for NVPTX despite the fact the the TLS is only used in host code. 

```
LLVM ERROR: Cannot select: 0x7f8f0d25f1a0: i64 = GlobalTLSAddress<i32 addrspace(1)* @_D8dcompute6driver5error6statusE8dcompute6driver5error6Status> 0
In function: _D8dcompute6driver4cuda8platform8Platform10initialiseFkZv
```

Things to note:

* I have no idea why the `addrspace(1)` is there, I ~think this is in the NVPTX backend (which would make it only slightly less weird), see below~  it is I built an LLVM find out.
* if the variable is made `__gshared` the NVPTX backend segfaults instead.
* files outputted with `-output-ll` can be built without problem with `llc`.
* `%"ldc.dcompute.Pointer!(cast(AddrSpace)1u, float).Pointer" = type { float* }` is  missing _its_ `addrspace(1)` but the ABI transformation to remove the outer aggregate have produced a `float addrspace(1)*`

Below is the backtrace of the segfault with `__gshared`:
```
frame #0: 0x00000001003534f4 ldc2`llvm::NVPTXTargetLowering::getPrototype(llvm::DataLayout const&, llvm::Type*, std::__1::vector<llvm::TargetLoweringBase::ArgListEntry, std::__1::allocator<llvm::TargetLoweringBase::ArgListEntry> > const&, llvm::SmallVectorImpl<llvm::ISD::OutputArg> const&, unsigned int, llvm::ImmutableCallSite) const + 948
    frame #1: 0x00000001003556cf ldc2`llvm::NVPTXTargetLowering::LowerCall(llvm::TargetLowering::CallLoweringInfo&, llvm::SmallVectorImpl<llvm::SDValue>&) const + 6927
    frame #2: 0x0000000100694f1a ldc2`llvm::TargetLowering::LowerCallTo(llvm::TargetLowering::CallLoweringInfo&) const + 6714
    frame #3: 0x00000001006aeab9 ldc2`llvm::SelectionDAGBuilder::lowerInvokable(llvm::TargetLowering::CallLoweringInfo&, llvm::BasicBlock const*) + 953
    frame #4: 0x000000010069ccb4 ldc2`llvm::SelectionDAGBuilder::LowerCallTo(llvm::ImmutableCallSite, llvm::SDValue, bool, llvm::BasicBlock const*) + 1652
    frame #5: 0x0000000100687e1d ldc2`llvm::SelectionDAGBuilder::visitCall(llvm::CallInst const&) + 1357
    frame #6: 0x000000010067ce59 ldc2`llvm::SelectionDAGBuilder::visit(llvm::Instruction const&) + 121
    frame #7: 0x0000000100706b05 ldc2`llvm::SelectionDAGISel::SelectBasicBlock(llvm::ilist_iterator<llvm::ilist_detail::node_options<llvm::Instruction, false, false, void>, false, true>, llvm::ilist_iterator<llvm::ilist_detail::node_options<llvm::Instruction, false, false, void>, false, true>, bool&) + 373
    frame #8: 0x0000000100705efe ldc2`llvm::SelectionDAGISel::SelectAllBasicBlocks(llvm::Function const&) + 8718
    frame #9: 0x0000000100703244 ldc2`llvm::SelectionDAGISel::runOnMachineFunction(llvm::MachineFunction&) + 1748
    frame #10: 0x00000001008bafd6 ldc2`llvm::MachineFunctionPass::runOnFunction(llvm::Function&) + 134
    frame #11: 0x0000000101381802 ldc2`llvm::FPPassManager::runOnFunction(llvm::Function&) + 514
    frame #12: 0x0000000101381a03 ldc2`llvm::FPPassManager::runOnModule(llvm::Module&) + 51
    frame #13: 0x0000000101381e84 ldc2`llvm::legacy::PassManagerImpl::run(llvm::Module&) + 900
    frame #14: 0x00000001002b5e1d ldc2`codegenModule(llvm::TargetMachine&, llvm::Module&, llvm::raw_fd_ostream&, llvm::TargetMachine::CodeGenFileType) + 285
```

thewilsonator:
By resetting `gTargetMachine` and `gIR`. Fixes #2307.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -519,6 +519,11 @@ cl::list<std::string>
                              " list. Use 'ocl-xy0' for OpenCL x.y, and "
                              "'cuda-xy0' for CUDA CC x.y"),
                      cl::value_desc("targets"));
+cl::opt<std::string>
+    dcomputeFilePrefix("mdcompute-file-prefix",
+                       cl::desc("Prefex to prepend to the generated kernel files."),
+                       cl::init("kernels"),
+                       cl::value_desc("prefix"));
 #endif
 
 static cl::extrahelp footer(
--- a/gen/dcompute/target.cpp
+++ b/gen/dcompute/target.cpp
@@ -13,6 +13,7 @@
 #include "ddmd/scope.h"
 #include "driver/linker.h"
 #include "driver/toobj.h"
+#include "driver/cl_options.h"
 #include "gen/dcompute/target.h"
 #include "gen/llvmhelpers.h"
 #include "gen/runtime.h"
@@ -45,7 +46,7 @@ void DComputeTarget::writeModule() {
 
   std::string filename;
   llvm::raw_string_ostream os(filename);
-  os << "kernels_" << short_name << tversion << '_'
+  os << opts::dcomputeFilePrefix << '_' << short_name << tversion << '_'
      << (global.params.is64bit ? 64 : 32) << '.' << binSuffix;
 
   const char *path = FileName::combine(global.params.objdir, os.str().c_str());
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
       env: LLVM_VERSION=3.7.1 OPTS="-DTEST_COVERAGE=ON"
     - os: osx
       d: ldc-beta
-      env: LLVM_VERSION=6.0.0-1 OPTS="-DBUILD_SHARED_LIBS=OFF" LLVM_SPIRV_AVAILABLE=ON
+      env: LLVM_VERSION=6.0.0-2 OPTS="-DBUILD_SHARED_LIBS=OFF" LLVM_SPIRV_AVAILABLE=ON
     - os: osx
       d: ldc
       env: LLVM_VERSION=4.0.0 OPTS="-DBUILD_SHARED_LIBS=ON"
@@ -27,7 +27,7 @@ matrix:
 
 cache:
   directories:
-    - llvm-spirv-6.0.0-1
+    - llvm-spirv-6.0.0-2
     - llvm-4.0.1
     - llvm-4.0.0
     - llvm-3.9.1
@@ -54,7 +54,7 @@ before_install:
     fi;
     if [ -z "$(ls -A llvm-$LLVM_VERSION)" ]; then
       if [ -n "${LLVM_SPIRV_AVAILABLE}" ]; then
-        wget -O llvm-spirv-$LLVM_VERSION.tar.bz2 https://github.com/thewilsonator/llvm/releases/download/pre-intrinsics/LLVM-6.0.0git-7375e8c-Darwin.tar.bz2;
+        wget -O llvm-spirv-$LLVM_VERSION.tar.bz2 https://github.com/thewilsonator/llvm/releases/download/pre-intrinsics/LLVM-6.0.0svn-Darwin.tar.bz2;
         mkdir llvm-spirv-$LLVM_VERSION;
         tar -xjf llvm-spirv-$LLVM_VERSION.tar.bz2 --strip 1 -C llvm-spirv-$LLVM_VERSION;
       else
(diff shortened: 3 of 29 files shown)


thewilsonator:
So would I, but I haven't the time at the moment. Also I want to get a minimal example working soon so that other people can start working on the D side of things.

I suppose we could write it out to disk (or memory) and then read it back and then put it through the backend from LDC, breaking whatever cross module references were causing the issue in the first place.

I'll have a play around and see what I can get to work and try removing the dependance on `llc`.

thewilsonator:
Found the problem, `gTargetMachine` was left as the NVPTX target for host code. The `DComputeCodeGenManager` now restores `gIR` and  `gTargetMachine` in its destructor.
Will include a test soon.

thewilsonator:
There is a race condition in the tests due to clashing file names, I will fix it tomorrow.

driver/cl_options.cpp
```diff
@@ -519,6 +519,11 @@ cl::list<std::string>
                              " list. Use 'ocl-xy0' for OpenCL x.y, and "
                              "'cuda-xy0' for CUDA CC x.y"),
                      cl::value_desc("targets"));
+cl::opt<std::string>
+    dcomputeFilePrefix("mdcompute-file-prefix",
+                       cl::desc("Prefex to prepend to the generated kernel files."),
```
kinke: Pref**i**x. Please give us a couple of days for potential reviews instead of merging eagerly yourself.

joakim-noah:
Did this break ldc compilation?  Had to revert it locally on linux/x64.

gen/dcompute/target.cpp
```diff
@@ -45,7 +46,7 @@ void DComputeTarget::writeModule() {
 
   std::string filename;
   llvm::raw_string_ostream os(filename);
-  os << "kernels_" << short_name << tversion << '_'
+  os << opts::dcomputeFilePrefix << '_' << short_name << tversion << '_'
```
joakim-noah: This is the line that doesn't compile.

thewilsonator: That is strange. Works for me, all the CIs passed (except circle which fails to get the latest LLVM). What error do you get?

thewilsonator: Are you using GCC, perhaps? I think all of the CIs use the clang that comes with the LLVM that LDC is built against. I'm also using clang.

kinke: As shown by the logs, the *nix CI boxes all use gcc except for Travis OSX, which uses the system/XCode clang.

joakim-noah: Tried both compilers now, hits with both clang++ 4.0.1 and g++ 7.1.1 on Arch linux/x64.  Here's the clang++ error, pretty much the same with g++:
```
[25%] Building CXX object CMakeFiles/LDCShared.dir/gen/dcompute/targetCUDA.cpp.o
/home/joakim/ldc/gen/dcompute/target.cpp:49:15: error: no member named
              'dcomputeFilePrefix' in namespace 'opts'
os << opts::dcomputeFilePrefix << '_' << short_name << tver...
     ~~~~~~^    
1 error generated.                                                   
make[3]: *** [CMakeFiles/LDCShared.dir/build.make:602:
CMakeFiles/LDCShared.dir/gen/dcompute/target.cpp.o] Error 1
```

kinke: The cmdline options is guarded by `#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX`, this here apparently isn't.

kinke: #2320

Issue #2303, opened 2017-09-01
Use zip instead of tar in ldc-build-runtime.d


andre2007:
Is it possible to use the github release zip archive instead of the tar archive in ldc-build-runtime.d?
Phobos has a module for extracting zip archives. The dependency on an external application (tar) would be superfluous and it would be much easier for windows user. The console parameter ldcSrcDir is not self explaining for beginners.

JohanEngelen:
The github zip archive does not contain submodules, so we have to use our selfmade tar archive. But we could also upload a zip archive if that makes things easier.

JohanEngelen:
@andre2007 Do you have an idea for an improved documentation tetx for ldcSrcDir?

andre2007:
Providing a zip file would make the use of ldc-build-runtime.d straight forward on windows too, as you do not need additional applications (tar) or think about an additional command parameter.

For your other questions, I can tell you what I was thinking while trying to start the ldc-build-runtime.d
(I read the WIKI before).

- As described the tool does not work out of the box on windows
- I did a quick search on google whether there is an easy way to install tar on windows (self contained executable does not exist, only a larger package, but it was only a very quick look)
- There is a command parameter ldcSrcDir: Do I have to specify some folder of my ldc installation...?
- After reading the source code of ldc-build-runtime.d, I think I have to download from github releases the right source tar, but there are at least two tars, (the one created by github and the one created by your tools)

andre2007:
There is a tar application available here "http://gnuwin32.sourceforge.net/packages/gtar.htm" but it is not usuable:

 ["tar", "-xzf", "ldc-src.tar.gz", "--strip", "1", "-C", "ldc-src"]
tar: unrecognized option `--strip'
`tar --help' zeigt weitere Informationen.
Error: command failed with status 2

kinke:
[FWIW, `tar.exe` is part of git for Windows, I'm using that one. It's not standalone though, it requires some MSYS DLLs.]

joakim-noah:
I'd like to have a dub file for druntime and Phobos, which could be useful upstream too, but that means reproducing the relevant portions of the current CMake configuration in dub.  This was the easiest way to get going, but we'll get it on dub one day.  @kinke, why not provide a zip file and just use that on Windows with `ldc-build-runtime`?  Seems like using tar is a pain there.

kinke:
[...]

PR #2314, opened 2017-09-06
ldc-build-runtime: Get rid of tar, let Phobos extract ZIP source archive

kinke:
Based on https://github.com/ldc-developers/ldc/issues/2303#issuecomment-327170328.


kinke:
Tested on Windows with https://github.com/ldc-developers/ldc-scripts/commit/4580950700f28e0c6d293b3f885e730e43f82620.

kinke:
Tested manually on Linux too.

PR #2315, opened 2017-09-06
Adapt to LLVM 5.0 debuginfo changes & misc. fixes, mainly for Win64

kinke:
A ~~subset~~ rework of #2294.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -113,22 +113,22 @@ ldc::DIScope ldc::DIBuilder::GetCurrentScope() {
   return fn->diLexicalBlocks.top();
 }
 
-void ldc::DIBuilder::Declare(const Loc &loc, llvm::Value *var,
+// Sets the memory address for a debuginfo variable.
+void ldc::DIBuilder::Declare(const Loc &loc, llvm::Value *storage,
                              ldc::DILocalVariable divar,
                              ldc::DIExpression diexpr) {
   unsigned charnum = (loc.linnum ? loc.charnum : 0);
   auto debugLoc = llvm::DebugLoc::get(loc.linnum, charnum, GetCurrentScope());
-  DBuilder.insertDeclare(var, divar, diexpr, debugLoc, IR->scopebb());
+  DBuilder.insertDeclare(storage, divar, diexpr, debugLoc, IR->scopebb());
 }
 
-// Use this to tag debuginfo onto a non-alloca LLVM value, e.g. a function
-// parameter.
-void ldc::DIBuilder::DbgValue(const Loc &loc, llvm::Value *var,
+// Sets the (current) value for a debuginfo variable.
+void ldc::DIBuilder::SetValue(const Loc &loc, llvm::Value *value,
                               ldc::DILocalVariable divar,
                               ldc::DIExpression diexpr) {
   unsigned charnum = (loc.linnum ? loc.charnum : 0);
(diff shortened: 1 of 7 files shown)


gen/dibuilder.cpp
```diff
@@ -1040,12 +1055,23 @@ void ldc::DIBuilder::EmitLocalVariable(llvm::Value *ll, VarDeclaration *vd,
...
-    auto T = DtoType(type);
-    TD = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type, TD,
-                                      getTypeAllocSize(T) * 8, // size (bits)
-                                      DtoAlignment(type) * 8); // align (bits)
+    // Note: createReferenceType expects the size to be the size of a pointer,
+    // not the size of the type the reference refers to.
+    TD = DBuilder.createReferenceType(
+        llvm::dwarf::DW_TAG_reference_type, TD,
+        gDataLayout->getPointerSizeInBits(), // size (bits)
+        DtoAlignment(type) * 8);             // align (bits)
```
JohanEngelen: I am now thinking that the alignment also needs to be that of the pointer, not of the pointee.

kinke: Yeah, I found that pretty weird.

kinke:
At least with LLVM 5.0, the additional deref for GEPs is clearly wrong, e.g.:
```
arg0 = <error reading variable arg0 (Cannot access memory at address 0x1)>
enc_n = <error reading variable enc_n (Cannot access memory at address 0x7d)>
```

kinke:
Hmm nice, without the additional deref for the GEPs, the debuginfos are now also available on Win64 (VS debugger) with LLVM 5.0. Not just for the outer function as with the alloca+expression approach, but also for the inner one. I.e., analog to Linux x64 with gdb. Nested refs are not displayed (on Linux: `<optimized out>`).

kinke:
The CI results seem to suggest the GEP deref was needed for LLVM < 5 but isn't anymore...

kinke:
Alright, nested refs now also showing up properly in outer & inner functions, on both Linux and Win64 with LLVM 5.0.

kinke:
Debuginfos for nested vars on Win64 with LLVM < 5.0 seem impossible. Their values are garbage when using the GEP directly, and unavailable with GEP+deref, for LLVM 4.0.1.
On Linux, the additional GEP deref is required for LLVM 4.0.1, otherwise the values are garbage.

kinke:
I'd like to see this in 1.4 final; mainly for people building LDC themselves with LLVM 5.0.

kalev:
Yes please. Fedora is switching to llvm 5.0 soon and would be nice to have an ldc release that builds with it.

gen/dibuilder.cpp
```diff
@@ -113,12 +113,26 @@ ldc::DIScope ldc::DIBuilder::GetCurrentScope() {
   return fn->diLexicalBlocks.top();
 }
 
-void ldc::DIBuilder::Declare(const Loc &loc, llvm::Value *var,
+// Use this to set the memory address for a debuginfo variable.
```
klickverbot: Nitpick: `Use this to set…` is a weirdly verbose comment style, just write `Sets…` instead.

gen/dibuilder.cpp
klickverbot: So is this correct as is, or does it need to be changed to the pointer type?

PR #2316, opened 2017-09-07
Travis: Add LLVM 5.0

JohanEngelen:
LLVM 5.0 was just released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ matrix:
   include:
     - os: linux
       d: ldc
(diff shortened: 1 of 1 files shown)


kinke:
There's still no prebuilt archive for Debian x64, they gotta be kidding.

JohanEngelen:
Ubuntu x64 is there, just the filename is following a different pattern :/

JohanEngelen:
Perhaps the LLVM binaries are a strange build. The `-dwarf-version` option is not supposed to be in the LLVM libs that we link with...?

kinke:
You're most likely building with LLD, right? LLD 5.0+ needs #2148.

JohanEngelen:
Ah, yes it's building with LLD.

PR #2317, opened 2017-09-09
Fix compilation warnings with ZLib

JohanEngelen:


JohanEngelen:
warnings are of this kind: `../runtime/phobos/etc/c/zlib/gzread.c:35:15: warning: implicit declaration of function 'read' is invalid in C99 [-Wimplicit-function-declaration]`

runtime/CMakeLists.txt
```diff
@@ -308,6 +310,11 @@ if("${TARGET_SYSTEM}" MATCHES "MSVC")
     # warning C4996: zlib uses 'deprecated' POSIX names
     append("/wd4131 /wd4206 /wd4996" CMAKE_C_FLAGS_RELEASE)
 endif()
+CHECK_INCLUDE_FILE(unistd.h HAVE_UNISTD_H)
+if (HAVE_UNISTD_H)
+    # Needed for zlib
+    append("-DHAVE_UNISTD_H" CMAKE_C_FLAGS_RELEASE)
```
dkgroot: Would it not be beter to use CMAKE_C_FLAGS instead of CMAKE_C_FLAGS_RELEASE ? While unittesting i had some issues compiling the zlib functions when using CMAKE_C_FLAGS_RELEASE.

JohanEngelen: Ah, indeed I've seen the warnings pop up again with ldc-build-runtime building the unittesters; probably that fixes that too.

PR #2320, opened 2017-09-09
Fix build for LLVM without DCompute targets

kinke:


gen/dcompute/target.cpp
```diff
@@ -7,6 +7,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX
```
klickverbot: Shouldn't these `.cpp` files just be excluded from the build completely?

kinke: I don't want to waste more time on this. The mocks for the manager make sense to me, and I'd rather have the disabling here than buried in the CMake mess. E.g., when excluded, they wouldn't be part of the Visual Studio project in my case, as I don't have DCompute targets enabled for my dev LLVM...

klickverbot: Agreed on the time wasting – it shouldn't be your job to clean this up in the first place.

joakim-noah:
Confirm that this fixes ldc compilation for me on both Android/ARM and linux/x64.  I don't understand how the CI missed this, @thewilsonator, are we building ldc with some special config on there?

thewilsonator:
I'm not sure either: DCompute is built when LLVM is built with either the NVPTX backend and/or the SPIRV backend is enabled, everything passed except circle which fails to get the latest LLVM. I was (apparently wrongly) under the assumption that only the travis and circle LLVM's had all targets enabled, and so the appveyor and semaphore CI's would catch anything related to missing targets.

I am sorry for the breakage.

kinke:
> I am sorry for the breakage.

No worries, can happen to anyone, especially if strangely all CI LLVMs have DCompute enabled but no release package has. Should we enable nvptx/CUDA for the releases at least?
Also note that I've created the LDC-LLVM `ldc-release_50` branch with our 3 commits on top of vanilla LLVM 5.0.0. So if you'd like to include SPIR-V for LDC 1.5, it would be a good time to work on it. :)

thewilsonator:
NVTPX yes. 
SPIRV has a bunch of fixes for LLVM master and I don't really feel like back porting them. Also SPIRV is untested and I don't currently have SPIR(V) capable hardware to test it, whereas I've almost (I hope) got a runnable test for PTX on my PC's ancient graphics card (compute capability 2.1). SPIRV also doesn't have proper windows support on account of the Itanium mangling used in the backend.

PR #2321, opened 2017-09-10
Update druntime to fix dlang issue 17821

JohanEngelen:
See https://github.com/ldc-developers/druntime/pull/102

https://issues.dlang.org/show_bug.cgi?id=17821


JohanEngelen:
Yeah, I am assuming that `val = newval` semantics are desired (with any kind of conversion, potentially user-defined).

kinke:
I'd only expect a matching `T` rhs (with implicit conversion) to work, but that's probably just me. ;)

PR #2324, opened 2017-09-11
Add druntime standalone tests

kinke:


kinke:
Failing on OSX:
* druntime-test-exceptions

Failing on Win32/64:
* druntime-test-allocations
* druntime-test-coverage (apparently just a backslash instead of fwd slash in output)
* druntime-test-cycles
* druntime-test-exceptions (different failures for Win32 and Win64)

kinke:
Finally got round to finishing this.

PR #2332, opened 2017-09-16
Move ldc executable sources from top level directory to 'sources'

Hardcode84:
With minimal changes in cmake for now.
This allow to make different cmake setup for runtime and compiler itself (e.g. do find_package(LLVM) with different sets of libs for runtime and compiler) eventually.
And this is more logical directory structure in general from my opinion (group all ldc executable related files in single directory).

But it is not too late to stop me :)


JohanEngelen:
Can you start a discussion on our forum about this, and your reasons for this big change?

> This allow to make different cmake setup for runtime and compiler itself (e.g. do find_package(LLVM) with different sets of libs for runtime and compiler) eventually.

The runtime can already be built with a different cmake setup. It's what ldc-build-runtime makes use of. I don't understand the example you gave (find_package(LLVM)): the LLVM libs are not used to build the runtime libs.

Hardcode84:
@JohanEngelen 
https://forum.dlang.org/post/dprcbslfwmmmlxvugxmk@forum.dlang.org

kinke:
Could you please list the LLVM libs your JIT runtime library depends on, a `readelf -d` dump or something? I'd just like to know how many libs we are talking about.

kinke:
What I'm getting is that I'd hope something manual like this would work if the number of libs is manageable:
```
if("${TARGET_SYSTEM}" MATCHES "MSVC")
    set(ldc-jit-rt-link_flags /LIBPATH:${LLVM_LIBRARY_DIRS} ….lib …)
else()
    set(ldc-jit-rt-link_flags -L${LLVM_LIBRARY_DIRS} -l… …)
endif()
```
I think it's not that ugly and worth the effort if it allows skipping the LLVM detection for the runtime libs, relying on the top-level CMake script to provide `LLVM_LIBRARY_DIRS` (or manually if run via `ldc-build-runtime` - optionally of course, skipping building the JIT runtime lib if it isn't available).

PR #2334, opened 2017-09-18
[ltsmaster] Backport PR #2315.

JohanEngelen:
The change is to fix debuginfo build errors with Phobos and LLVM5.0. This PR is mostly copy-pasting from master and fixing compile errors, so not much checking whether things make sense or not for LTS.

### DIFF
--- a/gen/nested.cpp
+++ b/gen/nested.cpp
@@ -58,14 +58,13 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
 
   // Check whether we can access the needed frame
   FuncDeclaration *fd = irfunc->decl;
-  while (fd != vdparent) {
-    if (fd->isStatic()) {
-      error(loc, "function %s cannot access frame of function %s",
-            irfunc->decl->toPrettyChars(), vdparent->toPrettyChars());
-      return new DVarValue(astype, llvm::UndefValue::get(DtoPtrToType(astype)));
-    }
+  while (fd && fd != vdparent) {
     fd = getParentFunc(fd, false);
-    assert(fd);
+  }
+  if (!fd) {
+    error(loc, "function %s cannot access frame of function %s",
+          irfunc->decl->toPrettyChars(), vdparent->toPrettyChars());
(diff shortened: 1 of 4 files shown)


JohanEngelen:
@redstar As you're busy with lts master, is this PR still valuable? Or should we just drop debuginfo support in lts (because we only use it as bootstrapper)? I reckon debuginfo may help with fixing bugs but...

redstar:
Yes, I consider this PR still valueable. One of my next tasks. I like to have ltsmaster in a good, clean state.

redstar:
@JohanEngelen I do not get the test failure (`running/closure.d`) on ARM. Could you please rebase on ltsmaster?

JohanEngelen:
@redstar done

gen/nested.cpp
```diff
@@ -150,27 +137,42 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
...
-  val = DtoGEPi(val, 0, idx, vd->toChars());
+  LLValue *gep = DtoGEPi(val, 0, idx, vd->toChars());
+  val = gep;
   IF_LOG {
     Logger::cout() << "Addr: " << *val << '\n';
     Logger::cout() << "of type: " << *val->getType() << '\n';
   }
-  if (byref || (vd->isParameter() && getIrParameter(vd)->arg &&
-                getIrParameter(vd)->arg->byref)) {
+  const bool isRefOrOut = vd->isRef() || vd->isOut();
```
redstar: Evaluation of `isRefOrOut` and `vd->isParameter...` gives different results and as consequence a missing load, leading to the failure in the test suite. I think a simple solution is to add 

    const bool isRefParameter = vd->isParameter() && getIrParameter(vd)->arg && getIrParameter(vd)->arg->byref;

and change the condition below to 

    } else if (byref || isRefOrOut || isRefParameter) {

This makes sure that no load is missed.

JohanEngelen: @redstar you can push to this branch (afaik), can you add your fix?

redstar:
I merged your commit.

PR #2345, opened 2017-09-24
Extend search logic for finding compiler RT libs. (ASan and libFuzzer)

JohanEngelen:
This should help find the compilerRT libs when LDC is installed in the system's usual dirs. With this change, the ASan and libFuzzer libs do not need to be copied as part of building LDC.

With this PR, the path `LDC_LIB_DIR/clang/6.0.0/lib/linux/libclang_rt.asan-x86_64.a` is also searched.
To be honest, I don't know whether this actually works in practice. It all depends where clang would be installed. But I think it doesn't do harm, and should be clear how the logic can be improved by someone who does know.

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -12,6 +12,7 @@
 #include "driver/cl_options_instrumentation.h"
 #include "driver/cl_options_sanitizers.h"
 #include "driver/exe_path.h"
+#include "driver/ldc-version.h"
 #include "driver/tool.h"
 #include "gen/irstate.h"
 #include "gen/logger.h"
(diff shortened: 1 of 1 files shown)


JohanEngelen:
@WebDrake Do you know where compiler-rt would be installed when clang is installed with snap or other package managers?

WebDrake:
@JohanEngelen there are typically two different possibilities. For stuff that's custom built for the package, it'll be installed as if `--prefix=SNAP_ROOT` where `SNAP_ROOT` is `/snap/[name]/[revision]`.

If it's stuff that's installed into the snap from pre-existing deb packages (e.g. to include a necessary dependency) it'll typically be in the “normal” install location but copy-pasted into snap root (so e.g. you might have a `usr/lib` in there.

I'm hoping to get on to the new RT features this week, so I'll let you know when I have more specific info for this particular case. Hope this helps in the meanwhile!

JohanEngelen:
On Arch Linux, the path is `/usr/lib/clang/5.0.1/lib/linux/`

kinke:
Another, more flexible option (wrt. all the different distros) would be introducing a new command-line option à la `-compiler-rt-libdir`, optionally specified by the user or in ldc2.conf. Prebuilt packages and custom builds would use the copied libs in the main libs directory by default, and distro maintainers not shipping with copies could specify their compiler-rt libs directory in the config file. And the defaultlibs command-line help category would have an additional entry. ;)

I'd like to include the builtins lib from compiler-rt, which would be used for each link operation (i.e., not explicitly enabled like sanitizers/fuzzer), but only forwarded to the linker if the lib is found, so that depends on a robust way of looking up the lib.

JohanEngelen:
> It could also be useful for cross-linking scenarios, to make LDC pick up the correct libs dir for the target platform.

That's a good reason for a cmdline flag.

Still, I think the auto-detection of this PR is worth keeping.
But I still don't know why this PR broke CI.

driver/linker-gcc.cpp
```diff
@@ -183,40 +184,83 @@ void ArgsBuilder::addLTOLinkFlags() {
...
+std::string getFullClangCompilerRTLibPath(const llvm::Triple &triple,
+                                          llvm::StringRef name,
+                                          bool sharedLibrary = false) {
+  llvm::StringRef OSName =
+      triple.isOSDarwin()
+          ? "darwin"
+          : triple.isOSFreeBSD() ? "freebsd" : triple.getOSName();
+
+  auto relPath = llvm::Twine("clang/") + getLLVMVersionWithoutSuffix() +
+                 "/lib/" + OSName + "/" + name;
```
kinke: This looks suspicious; I think I remember similar issues when trying to use Twines like that (using a Twine lvalue for readability). From the LLVM docs:

> A Twine is not intended for use directly and should not be stored, its implementation relies on the ability to store pointers to temporary stack objects which may be deallocated at the end of a statement.

JohanEngelen: Ahhh yes indeed. I think one should only ever _pass a Twine to a function_ and not do what I am doing here. Thanks!
Will fix and try again.

klickverbot: > I think one should only ever pass a Twine to a function

Indeed!

JohanEngelen:
OK, I finally found the bug. Indeed involving Twine, but in a tricky way.
```c++
auto relPath = std::string("clang/") + ldc::llvm_version + "/lib/" + OSName + "/" + name;
```
That is still broken.

The reason is that because of the `+ OSName`, the temporary becomes an `llvm::Twine`. (`std::string + llvm::StringRef --> llvm::Twine`)  So the `auto` should not be there.

Issue #2346, opened 2017-09-24
Bad codegen with struct alignment and `offsetof`


JohanEngelen:
~~Although the exact alignment requirements are unclear~~
The exact alignment requirements are clear since 2.075 (see forum post: http://forum.dlang.org/post/vzvxkabifknrzffycjao@forum.dlang.org), the following code should not crash but does.

```d
struct UInt {
align(1):
    uint a;
}

struct Bug {
    ubyte one;
    UInt two;
}

align(1)
struct Align1UInt {
align(1):
    uint a;
}

struct BugAlign1 {
    ubyte one;
    Align1UInt two;
}

version(CRASH_DURING_COMPILATION) {
    struct A {
        Bug a;
        BugAlign1 b;
    }
}

void foobug(Bug* b) {
    b.two.a = 42;

    void* ptr = cast(void*)b;
    ptr += Bug.two.offsetof;
    assert( ptr == &b.two );
    assert( (cast(UInt*)ptr).a == 42 );
}

void fooalign1bug(BugAlign1 *b) {
    b.two.a = 42;

    void* ptr = cast(void*)b;
    ptr += BugAlign1.two.offsetof;
    assert( ptr == cast(void*)&b.two );
    assert( (cast(UInt*)ptr).a == 42 );
}

void main() {
    Bug b;
    foobug(&b);

    BugAlign1 b2;
    fooalign1bug(&b2);
}
```

DMD runs the code fine.

JohanEngelen:
We create the LLVM type wrong for `Bug` and `BugAlign1` (they should be the same btw).

We almost output things correctly, I seems.
Slightly more complex case:
```d
struct UInt {
align(2):
    uint a;
}

struct Bug {
    ubyte one;
    UInt two;
}
```
We output 
```llvm
%agrsize.UInt = type { i32 }
%agrsize.Bug = type { i8, [1 x i8], %agrsize.UInt }
```
(note that we already add the padding)
but I think it should be 
```llvm
%agrsize.UInt = type { i32 }
%agrsize.Bug = type <{ i8, [1 x i8], %agrsize.UInt }>
```
(angled brackets to indicate packed type, and that we ourselves take care of the padding)

kinke:
I have a fix I think, PR is coming up shortly.

JohanEngelen:
Fixed by #2347

PR #2347, opened 2017-09-24
Fix packed struct layout regression (issue #2346)

kinke:
I probably introduced this regression lately. Back then, I didn't account for container aggregates, which require that the LL struct is packed if the overall D aggregate alignment is lower than a field's natural alignment.

### DIFF
--- a/ir/irtypeaggr.cpp
+++ b/ir/irtypeaggr.cpp
@@ -203,22 +203,44 @@ IrTypeAggr::IrTypeAggr(AggregateDeclaration *ad)
       aggr(ad) {}
 
 bool IrTypeAggr::isPacked(AggregateDeclaration *ad) {
-  const auto aggregateSize = (ad->sizeok == SIZEOKdone ? ad->structsize : ~0u);
+  // If the aggregate's size is unknown, any field with type alignment > 1 will
+  // make it packed.
+  unsigned aggregateSize = ~0u;
+  unsigned aggregateAlignment = 1;
+  if (ad->sizeok == SIZEOKdone) {
+    aggregateSize = ad->structsize;
+
+    const auto naturalAlignment = ad->alignsize;
+    auto explicitAlignment = STRUCTALIGN_DEFAULT;
+    if (auto sd = ad->isStructDeclaration())
+      explicitAlignment = sd->alignment;
+
+    aggregateAlignment = explicitAlignment == STRUCTALIGN_DEFAULT
+                             ? naturalAlignment
+                             : explicitAlignment;
+  }
+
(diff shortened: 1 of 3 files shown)


tests/codegen/gh2346.d
```diff
@@ -0,0 +1,53 @@
...
+static assert(PackedContainer2.alignof == 1);
+static assert(PackedContainer2.sizeof == 5);
+static assert(PackedContainer2.two.offsetof == 1);
+
+// CHECK-DAG: %gh2346.WeirdContainer = type { i8, [1 x i8], %gh2346.UnalignedUInt, [2 x i8] }
+align(4) struct WeirdContainer {
+    ubyte one;
+    align(2) UnalignedUInt two;
+}
+static assert(WeirdContainer.alignof == 2);
```
kinke: Hmm, that was a bit surprising.

kinke: Trying to fix this upstream, see https://github.com/dlang/dmd/pull/7164.

JohanEngelen: Nice catch.

JohanEngelen: Comment-out the `static assert(...=4)` then?

ir/irtypeaggr.cpp
```diff
@@ -213,10 +217,12 @@ bool IrTypeAggr::isPacked(AggregateDeclaration *ad) {
     // aligned, and LLVM would insert padding.
     const auto naturalFieldAlignment = field->type->alignsize();
     const auto mask = naturalFieldAlignment - 1;
-
-    // If the aggregate's size is unknown, any field with natural alignment > 1
-    // will make it packed.
-    if ((aggregateSize & mask) != 0 || (field->offset & mask) != 0) {
+    // Additionally, the aggregate's overall alignment must not be lower than
+    // a field's natural one, otherwise LLVM would use a higher overall LL
+    // struct alignment, causing additional padding in aggregates containing
```
JohanEngelen: This is related to
```
align(2) struct S {
   align(4) int i;
}
```
?
(perhaps add a test for that too?)

tests/codegen/gh2346.d
kinke: I'll just wait for upstream to be merged and will then embed it here.

JohanEngelen: nice

ir/irtypeaggr.cpp
kinke: Hmm you're right, explicit alignment needs to be taken into consideration here as well.

kinke:
Upstream fix (merged in the meantime) incorporated.

PR #2349, opened 2017-09-25
Adapt embedded llvm-ar to LLVM 5.0 final

kinke:
It was based on early LLVM 5.0, and there's been one notable change in the meantime, see https://reviews.llvm.org/D33575. I think @joakim-noah mentioned some issues wrt. object names in static libs. **Edit**: And @JohanEngelen something wrt. object names and LTO?

### DIFF
--- a/driver/archiver.cpp
+++ b/driver/archiver.cpp
@@ -50,7 +50,7 @@ bool Symtab = true;
 bool Deterministic = true;
 bool Thin = false;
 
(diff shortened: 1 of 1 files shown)


joakim-noah:
This patch gets rid of the paths and fixes the issue I had with llvm 5.0.

JohanEngelen:
LTO on OSX has troubles with same-name objects in archives indeed. But it's a linker issue (quite sure), perhaps with XCode 9, the problem is gone. Can't test right now though :/

PR #2353, opened 2017-09-29
Win32: Drop extra leading underscore for D symbol mangling

kinke:
Compatible with DMD.

### DIFF
--- a/gen/mangling.cpp
+++ b/gen/mangling.cpp
@@ -97,8 +97,8 @@ std::string hashSymbolName(llvm::StringRef name, Dsymbol *symb) {
 }
 }
 
-std::string getMangledName(FuncDeclaration *fdecl, LINK link) {
-  std::string mangledName(mangleExact(fdecl));
+std::string DtoMangledName(FuncDeclaration *fdecl, LINK link) {
+  std::string mangledName = mangleExact(fdecl);
 
   // Hash the name if necessary
   if (((link == LINKd) || (link == LINKdefault)) &&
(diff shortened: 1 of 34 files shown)


gen/mangling.cpp
```diff
@@ -97,8 +97,8 @@ std::string hashSymbolName(llvm::StringRef name, Dsymbol *symb) {
 }
 }
 
-std::string getMangledName(FuncDeclaration *fdecl, LINK link) {
-  std::string mangledName(mangleExact(fdecl));
+std::string DtoMangledName(FuncDeclaration *fdecl, LINK link) {
```
JohanEngelen: please don't change these function names to the awful "Dto*" pattern.

gen/moduleinfo.cpp
```diff
@@ -56,11 +57,11 @@ llvm::Function *buildForwarderFunction(
   const auto fnTy =
       LLFunctionType::get(LLType::getVoidTy(gIR->context()), {}, false);
 
-  std::string const symbolName = gABI->mangleFunctionForLLVM(name, LINKd);
-  assert(gIR->module.getFunction(symbolName) == NULL);
+  const auto llMangle = DtoMangledFuncName(name, LINKd);
```
JohanEngelen: you are changing a lot of variable names, where the new name is worse than before. How is `llMangle` better than `symbolName` ? If you want to indicate that it is mangled, then change it to `mangledSymbolName`.

ir/irclass.cpp
```diff
@@ -124,19 +130,15 @@ LLGlobalVariable *IrAggr::getInterfaceArraySymbol() {
   LLType *InterfaceTy = DtoType(Type::typeinfoclass->fields[3]->type->nextOf());
 
   // create Interface[N]
-  LLArrayType *array_type = llvm::ArrayType::get(InterfaceTy, n);
+  const auto llMangle = DtoMangledInterfaceInfosSymbolName(cd);
```
JohanEngelen: `mangledName` --> `llMangle` ?

gen/moduleinfo.cpp
kinke: All purely subjective. There are 3 mangles - base mangle, LL mangle, final mangle.

gen/mangling.cpp
kinke: I find it absolutely appropriate, as it translates from D layer to LLVM layer.

gen/moduleinfo.cpp
JohanEngelen: Of course subjective. But you can do a lot better here.
All these abbreviations are bad for legibility. Just write down in the code what you are doing: e.g. `irMangledSymbolName`.

gen/mangling.cpp
JohanEngelen: `getMangledNameForIR`, `getIRMangledName`, something like that is much clearer than "D to".

kinke: Renamed.

Issue #1970, opened 2017-01-16
Can't use rt_options


yosupo06:
System: OSX(10.11.6)
Architecture: x86_64
LDC: 1.1.0-beta6, 0.17.2
LLVM: 3.8.0

https://dlang.org/spec/garbage.html#gc_config

I try to use this URL's rt_options.

Source code:

```
extern(C) __gshared string[] rt_options = ["gcopt=profile:1"];

import std.stdio;
int main(string[] args) {
	writeln("Hello.");
	return 0;
}

```

This code print profile with dmd(I checked v2.072.2 and v2.070.2), but don't print profile with ldc2.

rainers:
Works for me with x64, but not x86.

rainers:
Works with -m32, too, if you change `rt_options` to `_rt_options`.

The trouble is that the `cPrefix` in `core.demangle` is `_` while LDC adds duplicate underscores `__` for Win32. Why is that?

yosupo06:
I try to `_rt_options`, and work with OSX(with or without -m32).

kinke:
The extra OSX underscore is https://github.com/ldc-developers/ldc/issues/114. I guess druntime's `core.demangle` is to be fixed then.
Wrt. 32-bit Windows, I thought Rainer would know. ;)

kinke:
Fixed in https://github.com/ldc-developers/ldc/pull/2354.

PR #2354, opened 2017-09-30
druntime: Fix mangling issue #1970 for rt.config on Win32 and OSX

kinke:
~~LDC's~~ `pragma(mangle, …)` allows specifying a 'base' mangle for the symbol's linkage. So by using `extern(C) pragma(mangle, …)`, a base C mangle can be specified, without having to take care of a platform-specific prefix.

### DIFF
--- a/tests/linking/rt_options.d
+++ b/tests/linking/rt_options.d
@@ -1,0 +1,9 @@
+// RUN: %ldc -run %s
+
+extern(C) __gshared string[] rt_options = [ "key=value" ];
+
+void main()
+{
+    import rt.config;
+    assert(rt_configOption("key") == "value");
+}


kinke:
Upstream PR: https://github.com/dlang/druntime/pull/1925

JohanEngelen:
If you apply the upstream druntime PR exactly, without `version(LDC)`, we'd get correct git-merge behavior without manual intervention, right?

kinke:
I should have checked DMD behavior earlier, now it's already in our druntime's `ldc` branch. But yeah, as long as our version then looks identical to upstream, the upstream patch to be merged should later be discarded ('changes already applied`) afaik.

PR #2359, opened 2017-10-05
DMD-style inline asm: Add `s` suffix for fild etc. with 16-bit memory operand

kinke:
Needed for Win64 with 80-bit reals (i.e., MinGW) here (Win64-specific):
https://github.com/ldc-developers/phobos/blob/882449f7916031667e09fe7a203d305a60898641/std/math.d#L3691

### DIFF
--- a/gen/asm-x86.h
+++ b/gen/asm-x86.h
@@ -2535,7 +2535,7 @@ struct AsmProcessor {
     case FPInt_Types:
       switch (ptrtype) {
       case Short_Ptr:
-        type_suffix = "s";
+        type_suffix = 's';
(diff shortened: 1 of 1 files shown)


kinke:
I guess Phobos is wrong there, but at least it uncovered this issue - inline asm test coverage isn't really high. Figuring out which ops exactly are affected by this and verifying that the `s` suffix is the proper one for all of those isn't something I'm looking forward to; this was the only spot where no suffix was deemed necessary, so it's smelly enough for me. ;)

kinke:
Any concerns?

PR #2360, opened 2017-10-05
cmake: use -of= instead of -of

wirx6:
Building on mingw-w64 fails without it(shouldn't break other builds).


kinke:
Thanks; we use `-of=` exclusively for master's runtime CMake script.

PR #2363, opened 2017-10-08
[2.076] Debuginfos: Use fully qualified names

kinke:


kinke:
This, in combination with switching from `Object` (no explicit fields) to `Exception` in `runnable/testpdb.d`, gets the first assertions to work; it doesn't find any debuginfos for function `test15432()` though (using its `.mangleof`). We emit the human-readable name as DI function name and its mangle as DI function linkage name; maybe just a bug in LLVM 5; I haven't digged deeper - also because `llvm-pdbutil dump -symbols testpdb.pdb` crashes due to a failing assertion.

rainers:
I use https://github.com/Microsoft/microsoft-pdb/tree/master/cvdump to dump debug information in object or PDB files.

As it seems, LDC doesn't write a S_PUB32 for the function, but only a LF_FUNC_ID with the prettyfied name. That doesn't seem to make a big difference to the debugger, though.
I'm not sure how portable that specific test15432() is anyway, as it tests for specific dmd like generated code bytes, so it could be disabled for LDC.

PR #2364, opened 2017-10-11
Provide explicit template instantiation definitions for FPTypeProperties<T> static members

JohanEngelen:
Fixes compile warnings with XCode 9.

`gen/target.cpp:59:18: warning: instantiation of variable 'Target::FPTypeProperties<long double>::nan' required here, but no definition is available [-Wundefined-var-template]`

See https://stackoverflow.com/questions/43439862/c-template-instantiation-of-variable-required-here-but-no-definition-is-ava

No need to upstream because the warning is caused by the LDC-specific modification in the same file.

### DIFF
--- a/ddmd/target.h
+++ b/ddmd/target.h
@@ -83,4 +83,22 @@ struct Target
     static LINK systemLinkage();
 };
 
+#if IN_LLVM
+// Provide explicit template instantiation definitions for FPTypeProperties<T>
+// static members. See:
+// https://stackoverflow.com/questions/43439862/c-template-instantiation-of-variable-required-here-but-no-definition-is-ava
+template <typename T> real_t Target::FPTypeProperties<T>::max;
+template <typename T> real_t Target::FPTypeProperties<T>::min_normal;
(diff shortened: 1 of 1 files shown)


JohanEngelen:
oops, clang formatted.

kinke:
It's defined in D: https://github.com/ldc-developers/ldc/blob/master/ddmd/target.d#L67 - if it wasn't, there'd be a linker error.

JohanEngelen:
Yes, this is just to kill compiler warnings.

JohanEngelen:
good to merge?
(the warning flood is bad)

kinke:
Hmm, so how does this guarantee the (properly initialized) D definitions get used for float/double? Linker command-line order?

JohanEngelen:
The linker merges these C++ symbols with the D symbols (as long as the C++ source is in-sync with the D source). I assume that because the C++ source doesn't specify initializers, the D initializers are used.

kinke:
[I got rid of this in #2374.]

Issue #2357, opened 2017-10-04
Crash via "Assertion `decl->ir->irGlobal != NULL' failed"


ximion:
We found this one in Debian/Ubuntu while building Sambamba[1]
```
ldc2: /build/ldc-1.4.0/ir/irvar.cpp:44: IrGlobal* getIrGlobal(VarDeclaration*, bool): Assertion `decl->ir->irGlobal != NULL' failed.
/usr/lib/x86_64-linux-gnu/libLLVM-5.0.so.1(_ZN4llvm3sys15PrintStackTraceERNS_11raw_ostreamE+0x1a)[0x7f5d42064cea]
/usr/lib/x86_64-linux-gnu/libLLVM-5.0.so.1(_ZN4llvm3sys17RunSignalHandlersEv+0x56)[0x7f5d42062fd6]
```
```
LDC - the LLVM D compiler (1.4.0):
  based on DMD v2.074.1 and LLVM 5.0.0
  built with LDC - the LLVM D compiler (0.17.5)
  Default target: x86_64-pc-linux-gnu
  Host CPU: haswell
  http://dlang.org - http://wiki.dlang.org/LDC
```
The issue appears to affect all architectures and did not happen with LDC 1.2.

Here is a slightly more minimized testcase, with no additional dependencies other than LDC: [ldc_assert_crash_test.tar.gz](https://github.com/ldc-developers/ldc/files/1354469/ldc_assert_crash_test.tar.gz) (could likely be reduced even further).

Cc @LocutusOfBorg

[1]: https://github.com/lomereiter/sambamba

kinke:
Note that I mentioned and somewhat analyzed this already in https://github.com/ldc-developers/ldc/issues/2336. It's not clear whether that has to do with the general template culling issue or whether it's something different.

ximion:
Ah, nice - so this likely is a duplicate then.

kinke:
The crash should be fixed by #2366. I hope sambamba now builds fine on Debian. The fix is in v1.5.0-beta1 which will be released in a couple of hours.

LocutusOfBorg:
thanks!

ximion:
This is indeed fixed now, thank you! :-)

Issue #2336, opened 2017-09-20
Too aggressive duplicate code culling


kinke:
As we know since #2168 at least, there's too much code culling going on. There, it was part of a performance problem due to less inlining opportunities. Weka apparently need to weaken the culling to get around missing symbols. Even non-unittest Phobos needs to be compiled all-at-once to prevent missing symbols since 2.074.

And now it appears to cause a [crash](https://github.com/ldc-developers/ldc/issues/2300#issuecomment-330947562) for another code base due to a required TypeInfo being wrongly discarded as speculative, once again only when compiling that module separately. It compiles fine with DMD 2.074.0 though.

klickverbot:
> And now it appears to cause a crash for another code base due to a required TypeInfo being wrongly discarded as speculative, once again only when compiling that module separately. It compiles fine with DMD 2.074.0 though.

I don't think that mechanism can directly lead to crashes.

kinke:
```
ldc2: /home/martin/ldc/ir/irvar.cpp:44: IrGlobal* getIrGlobal(VarDeclaration*, bool): Assertion `decl->ir->irGlobal != NULL' failed.
#0 0x000000000189223a llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/home/martin/build-ldc/bin/ldc2+0x189223a)
#1 0x00000000018900ce llvm::sys::RunSignalHandlers() (/home/martin/build-ldc/bin/ldc2+0x18900ce)
#2 0x0000000001890242 SignalHandler(int) (/home/martin/build-ldc/bin/ldc2+0x1890242)
#3 0x00007fd8e6aae390 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x11390)
#4 0x00007fd8e5a49428 gsignal /build/glibc-bfm8X4/glibc-2.23/signal/../sysdeps/unix/sysv/linux/raise.c:54:0
#5 0x00007fd8e5a4b02a abort /build/glibc-bfm8X4/glibc-2.23/stdlib/abort.c:91:0
#6 0x00007fd8e5a41bd7 __assert_fail_base /build/glibc-bfm8X4/glibc-2.23/assert/assert.c:92:0
#7 0x00007fd8e5a41c82 (/lib/x86_64-linux-gnu/libc.so.6+0x2dc82)
#8 0x00000000006d4d6f (/home/martin/build-ldc/bin/ldc2+0x6d4d6f)
[...]

PR #2365, opened 2017-10-13
-betterC

kinke:


gen/runtime.cpp
```diff
@@ -356,6 +371,17 @@ static void buildRuntimeModule() {
...
+  // C assert function:
+  // OSX:  void __assert_rtn(const char *func, const char *file, unsigned line,
+  //                         const char *msg)
+  // else: void [_]_assert(const char *msg, const char *file, unsigned line)
+  createFwdDecl(
+      LINKc, Type::tvoid, {getCAssertFunctionName()},
+      global.params.targetTriple->isOSDarwin()
+          ? llvm::ArrayRef<Type *>({voidPtrTy, voidPtrTy, uintTy, voidPtrTy})
+          : llvm::ArrayRef<Type *>({voidPtrTy, voidPtrTy, uintTy}),
+      {}, Attr_Cold_NoReturn);
```
kinke: I guess adding `nounwind` would be appropriate?

kinke:
Wrt. the TypeInfos to be omitted for `-betterC`, I stumbled upon [DtoTypeInfoOf()](https://github.com/ldc-developers/ldc/blob/merge-2.076/gen/llvmhelpers.cpp#L1208) again (recently for [this issue](https://github.com/ldc-developers/ldc/issues/2336#issuecomment-330984390) as well), which fails horribly if the TypeInfo isn't emitted (LDC assertion or segfault; happens for DCompute, if the type is deemed speculative or if we would simply skip its emission for `-betterC` there).
So to handle the many code parts relying on `DtoTypeInfoOf()` gracefully, I think we should forward-declare the IR global at least, so that linker errors will result if the TypeInfo is needed but unavailable.

While thinking about this, I seemed to recall reading on the forum that GDC handles things differently, only emitting the TypeInfos 'on-demand', so that people don't necessarily need to resort to `-betterC` for less bloated binaries. AFAICT, we currently emit a struct's TypeInfo incl. its helper functions [exclusively] in its (instantiating) module(s) AFAICT, which may account for bloat (pulling in that object with all of its dependencies during linking).
I like the idea of emitting a struct's TypeInfo and its helper functions into each module which references it.

I haven't looked at the ClassInfos yet.

kinke:
I'm nowhere near the end of this rabbit hole, but I gave up on skipping the TypeInfo emission in the declaring module for now, as redundant TypeInfos in other objects still depend on the struct's init symbol, which is only emitted in the declaring module etc. That's the case for `-betterC` as well btw; DMD emits it too.

I guess there was a historic reason for building extra 'typeid(...)' LL types for the TypeInfos, similar as the custom types for the init symbols which I got rid of not too long ago, but I found none that justifies them any longer, only complicating stuff and uglifying the IR due to more overall types and pointer bitcasts.

Issue #2357 is fixed by the 5th commit - at least the module can be compiled without issues; not sure if the TypeInfo for that class is really defined in another object so that linking will succeed with Debian's separate-compilation-setup.

kinke:
Oh please don't tell me the reason for the 'typeid(...)' types were opaque structs. I hoped to get away by skipping their TypeInfo definition, but `runnable/testtypeid` then fails. LDC 1.4 emits a weird thing, basically a TypeInfo (vtable + monitor only), but with TypeInfo_Struct's vtable (but none of its 10/12 extra fields). DMD defines something too.

PR #2366, opened 2017-10-15
Backport fix for issue #2357

kinke:
[Derived from [this 2.076 commit](https://github.com/ldc-developers/ldc/pull/2365/commits/b3cfd7afc4c079f95593b6263316f0b074cf2734).]

Improve robustness for TypeInfos of speculative types by only eliding their TypeInfo definition, not the declaration of the LL global altogether.

`DtoTypeInfoOf()` expects the LL global to be created and otherwise fails with an assertion or segfault (e.g., issue #2357). So now only linker errors should result in case the TypeInfo definition is missing.

Also normalize the calls to `DtoTypeInfoOf()` and revise the following pointer bitcasts, as the LL type of forward-declared TypeInfo globals may be opaque.

### DIFF
--- a/gen/declarations.cpp
+++ b/gen/declarations.cpp
@@ -89,13 +89,16 @@ public:
       }
 
       // Emit TypeInfo.
-      DtoTypeInfoOf(decl->type);
+      DtoTypeInfoOf(decl->type, /*base=*/false);
 
-      // Define __InterfaceZ.
+      // Declare __InterfaceZ.
       IrAggr *ir = getIrAggr(decl);
       llvm::GlobalVariable *interfaceZ = ir->getClassInfoSymbol();
-      interfaceZ->setInitializer(ir->getClassInfoInit());
-      setLinkage(decl, interfaceZ);
+      // Only define if not speculative.
+      if (!isSpeculativeType(decl->type)) {
+        interfaceZ->setInitializer(ir->getClassInfoInit());
+        setLinkage(decl, interfaceZ);
+      }
     }
   }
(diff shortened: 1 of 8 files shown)


kinke:
Tested manually, works.

gen/declarations.cpp
```diff
@@ -134,12 +137,11 @@ class CodegenVisitor : public Visitor {
       IrAggr *ir = getIrAggr(decl);
       auto &initZ = ir->getInitSymbol();
       auto initGlobal = llvm::cast<LLGlobalVariable>(initZ);
-      assert(initGlobal);
-      setLinkage(decl, initGlobal);
```
kinke: `llvm::cast()` already asserts the types match. `setLinkage()` right before or after `IRState::setGlobalVarInitializer()` is fully equivalent, but the latter is consistent with the other normal `setLinkage()` calls after setting a LL global's initializer.

gen/toconstelem.cpp
```diff
@@ -68,11 +68,8 @@ class ToConstElemVisitor : public Visitor {
     }
 
     if (TypeInfoDeclaration *ti = e->var->isTypeInfoDeclaration()) {
-      LLType *vartype = DtoType(e->type);
       result = DtoTypeInfoOf(ti->tinfo, false);
-      if (result->getType() != getPtrToType(vartype)) {
```
kinke: This was a harmless bug. All TypeInfos are classes, so `DtoType()` already returns a LL pointer.

Issue #2341, opened 2017-09-24
druntime issues on Alpine Linux (Musl C runtime)


kirbyfan64:
Ok, so I've been trying to compile LDC on Alpine Linux, and it's been...interesting...

So, first off: because #1664 was only merged into the "modern" LDC tree, it's *impossible* to bootstrap LDC 0.17.5. I had to manually apply the patch onto the LDC source, and then I could bootstrap.

However, on Alpine, the `backtrace` functions are in [libexecinfo](http://www.freshports.org/devel/libexecinfo), not libc:

```
/usr/local/lib/libdruntime-ldc.a(dwarf.o): In function `_D2rt9backtrace5dwarf23traceHandlerOpApplyImplFAxPvMDFKmKxAaZiZi':
rt/backtrace/dwarf.d:(.text._D2rt9backtrace5dwarf23traceHandlerOpApplyImplFAxPvMDFKmKxAaZiZi[_D2rt9backtrace5dwarf23traceHandlerOpApplyImplFAxPvMDFKmKxAaZiZi]+0x2a): undefined reference to `backtrace_symbols'
/usr/local/lib/libdruntime-ldc.a(runtime.o): In function `_D4core7runtime19defaultTraceHandlerFPvZC6object9Throwable9TraceInfo':
core/runtime.d:(.text._D4core7runtime19defaultTraceHandlerFPvZC6object9Throwable9TraceInfo[_D4core7runtime19defaultTraceHandlerFPvZC6object9Throwable9TraceInfo]+0x6b): undefined reference to `backtrace'
/usr/local/lib/libdruntime-ldc.a(runtime.o): In function `_D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPS4core3sys5posix6signal9siginfo_tPvZv':
core/runtime.d:(.text._D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPS4core3sys5posix6signal9siginfo_tPvZv[_D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPS4core3sys5posix6signal9siginfo_tPvZv]+0x23): undefined reference to `backtrace'
[...]

PR #2373, opened 2017-10-20
Add predefined version CRuntime_Musl

kinke:
Wrt. issue #2341.

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -426,10 +426,12 @@ void ArgsBuilder::addUserSwitches() {
 void ArgsBuilder::addDefaultLibs() {
   bool addSoname = false;
 
-  switch (global.params.targetTriple->getOS()) {
+  const auto &triple = *global.params.targetTriple;
+
+  switch (triple.getOS()) {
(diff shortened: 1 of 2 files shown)


kinke:
The question is whether `CRuntime_Glibc` should be defined too, to minimize the required druntime patches, as it seems to basically work out of the box already.

joakim-noah:
Ideally we'd just define `CRuntime_Musl` and patch druntime to only need that, as I've done with Bionic before, but obviously this pull will get them up and running much quicker, as you say.  [I've opened a forum thread on microservices](http://forum.dlang.org/thread/twtbrappuarwhyzlchrb@forum.dlang.org), hopefully we can get some feedback and help on this issue.

kinke:
I added an unrelated patch due to the refactoring taken place in the first commit. It replaces the fatal error for an unknown target OS (with the MSP430 exception) with a warning, to allow for bare-metal targets.

kinke:
There's a first `CRuntime_Musl` occurrence in 2.078 (`core.sys.posix.poll`), so this should be good to go now.

Issue #2381, opened 2017-10-23
Issues with Visual Studio 2017 v15.4


kinke:
After the TLS breakage in VS v15.3 and the archiver trouble with Xcode 9.0.0, it's Microsoft's turn again. The unittest testrunners don't find ~~the~~ a lot of ModuleInfos anymore apparently (see AppVeyor). Argh, right in the middle of the v1.5 release.

kinke:
I suspect the MS linker's `/WHOLEARCHIVE` isn't pulling in all objects from the unittest libs anymore.

kinke:
Fixed by https://github.com/ldc-developers/ldc/commit/5f40c6b213814c646724bbbd8716a74554edde9b.

PR #2383, opened 2017-10-23
MSVC: Emit ModuleInfo references into .minfo section

kinke:
Instead of the legacy linked list.


kinke:
Doesn't help with #2381, but the ugly global ctor in the IR has been bothering me for a while now, I just never got round to do something about it until now.

Issue #2388, opened 2017-10-26
ICE on compiling Objective-C code in multiple files


jacob-carlborg:
The following code fails to compile:

```
$ touch foo.d
$ cat main.d
```

```d
void alloc()
{
    NSObject o;
    o.alloc();
}

extern (Objective-C):

interface NSObject
{
    NSObject alloc() @selector("alloc");
}
```

Compiling the above code gives the following error:

```
$ ldc2 --version
LDC - the LLVM D compiler (1.5.0-beta1):
  based on DMD v2.075.1 and LLVM 5.0.0
  built with LDC - the LLVM D compiler (1.5.0-beta1)
  Default target: x86_64-apple-darwin16.7.0
  Host CPU: haswell

$ ldc2 foo.d main.d
module flag identifiers must be unique (or of 'require' type)
!"Objective-C Version"
module flag identifiers must be unique (or of 'require' type)
!"Objective-C Image Info Version"
module flag identifiers must be unique (or of 'require' type)
!"Objective-C Image Info Section"
module flag identifiers must be unique (or of 'require' type)
!"Objective-C Garbage Collection"
LLVM ERROR: Broken module found, compilation aborted!
```

There has to be at least two files for this error to occur. It might be related to this issue in DMD [1].

[1] https://issues.dlang.org/show_bug.cgi?id=17936

kinke:
Thanks for testing. The are multiple LDC-specific issues here; in short, the current buggy Objective-C support only works when compiling a single D module per LDC invokation.

kinke:
@jacob-carlborg: It would be perfect if you could do some more Objective-C testing with #2389.

jacob-carlborg:
I tried to compile LDC but I get an error when I run `cmake`:

```
llvm-config: unknown component name: ipa
CMake Error at cmake/Modules/FindLLVM.cmake:140 (string):
  string sub-command REGEX, mode MATCHALL needs at least 5 arguments total to
  command.
Call Stack (most recent call first):
  cmake/Modules/FindLLVM.cmake:174 (llvm_set_libs)
  CMakeLists.txt:19 (find_package)
```

I've tried both LLVM 4.0 and 5.0. Pre-compiled binaries from http://releases.llvm.org/ and compiled myself.

kinke:
[...]

Issue #2387, opened 2017-10-26
ICE on Objective-C  method taking `bool`


jacob-carlborg:
The following code compiles an works using DMD:

```d
extern (Objective-C) interface NSView
{
    void setWantsLayer(bool value) @selector("setWantsLayer:");
}
```

Compiling the same code using LDC gives the following error:

```
Wrong types for attribute: signext zeroext
void (%mainMod.NSView*, i8*, i1)* @_D7mainMod6NSView13setWantsLayerMYbZv
LLVM ERROR: Broken module found, compilation aborted!
```

Replacing `bool` with `int` solves the problem.

kinke:
Thanks for testing, fixed by #2387.

PR #2389, opened 2017-10-26
Objective-C fixes

kinke:


tests/compilable/objc_gh2387.d
```diff
@@ -0,0 +1,9 @@
+// REQUIRES: Darwin
```
JohanEngelen: could you test this on other platforms with `-mtriple=x86_64-apple-macosx`?
(with a `// REQUIRES: target_X86`)

kinke: Why? I'd like to have it tested for the other Objective-C supporting targets too.

JohanEngelen: Such that the test runs on most of our platforms.
We can of course also add `-mtriple` for other targets.
(adding support for more targets while testing is easy, just build LLVM with more targets enabled, but it's not possible to add a host platform)

kinke: We had that same discussion a while back for another test. I see no point in testing on all platforms that we generate something successfully for a single target, if that target is covered by CI anyway. I see much more value in testing all affected native targets. If I run the testsuite on experimental iOS, I don't care whether it works for x64 OSX, but expect a green Objective-C test to mean that it works on the host.

JohanEngelen: CI is not the only testing we do, and for ARM we do not have CI hosts. So if you want that tested, you need to add explicit `-mtriple` in the tests.
All platforms should be testing as many things as possible. If I can have a test fail on my local machine, it is much more efficient to develop (and results in more overall test coverage because of different settings).

kinke: As stated earlier, same argument as a while ago, I don't and will never share your view on this.

> for ARM we do not have CI hosts

Well I know that of course, but then it's just more important for the tests to cover the host if one's developing on that one. I'm talking about the guy bringing LDC/D to a new platform. 99% of the other tests do test the host after all. He'd potentially only later find out about Objective-C failures if the testsuite used some hardcoded target triple(s). Testing the x64 codegen and that he can cross-compile to x64 (most likely he won't anyway due to disabled x86 LLVM backend) seems of little to no value to me.
So your stance would increase test coverage minimally (basically more LLVM configs, host shouldn't really matter), but due to hardcoded target triples automatically excludes future targets, hurting test coverage much worse (especially when adding a new platform!) and that I find unacceptable.

jacob-carlborg:
This fixes both #2388 and #2387 for me and my application builds and runs fine using LDC 👍.

kinke:
Perfect, thank you very much for testing.

PR #2393, opened 2017-10-27
Inline asm: Add `s` suffix for fild etc. with 16-bit memory operand

wirx6:
Backport https://github.com/ldc-developers/ldc/pull/2359


kinke:
It's already in ltsmaster.

wirx6:
I see it only in master.

PR #2400, opened 2017-11-01
Clean up LDC-specifics in druntime/Phobos

kinke:


kinke:
`rt.sections_elf_shared` could definitely use a cleanup on its own; the big diff wrt. upstream due to included Darwin support is mostly not guarded by `version(LDC)` checks.

PR #2401, opened 2017-11-04
Merge 2.077.0 front-end and stdlibs

kinke:


kinke:
Already green here on my Win64 box. No 64-bit `real` regressions this time! :)

kinke:
Hmm it just seems as if `~` (in `llvm~5.0`)is expanded to `/Users/dpn`.

kinke:
OS X should be green too according to Travis, as is Travis Linux (I already amended the LLVM 3.7 fix). The memory requirements may have gone up; Circle is killed when compiling the `std.regex.internal.tests` unittests and Semaphore is running out of RAM disk space. :/

kinke:
Semaphore now has a pre-installed gcc 4.9, that seems to suffice to work around the memory issues there. Circle seems to be fine with the workaround, as is hopefully Travis.

So on the CI side, what remains are a couple of release-only 32-bit unittest failures on both Windows (1 failure, `rt.switch_`) and Linux (additionally `core.sys.posix.stdio` and `core.sys.linux.tipc`) as well as a Circle-specific failure for the `debuginfo/classtypes_gdb.d` lit test with its gdb v7.12.

kinke:
32-bit x86 failures should be fixed.

I can reproduce the gdb issue on Ubuntu 16.04 with gdb 7.11 too. It doesn't find the D TLS global `classtypes_gdb.gvar` anymore, the mangled one `_D14classtypes_gdb4gvarCQw2uv` is there instead. I guess more recent gdb's support D demangling (but obviously the pre-2.077 mangling scheme; the previous mangle for that symbol was `_D14classtypes_gdb4gvarC14classtypes_gdb2uv`) while older gdb versions probably rely on the [display] names we emit in the debuginfo and therefore still pass that test:

```
!1 = distinct !DIGlobalVariable(name: "gvar", linkageName: "_D14classtypes_gdb4gvarCQw2uv", scope: !2, file: !3, line: 19, type: !6, isLocal: false, isDefinition: true)
```

PR #2407, opened 2017-11-12
Disassemble resulting jit module and output asm to debug handler

Hardcode84:
Debug asm output was disabled previously because`addPassesToEmitFile` asm output was different from actual jit asm. This is attempt to disassemble jit binary instead.

### DIFF
--- a/runtime/jit-rt/cpp-so/disassembler.cpp
+++ b/runtime/jit-rt/cpp-so/disassembler.cpp
@@ -199,6 +199,19 @@ public:
     // Nothing
   }
 };
+
+void processRelocations(SymTable &symTable,
+                        const llvm::object::ObjectFile &object,
+                        const llvm::object::SectionRef &sec) {
+  for (const auto &reloc : sec.relocations()) {
+    const auto symIt = reloc.getSymbol();
+    if (object.symbol_end() != symIt) {
+      const auto sym = *symIt;
+      symTable.addExternalSymbolRel(reloc.getOffset(),
+                                    llvm::cantFail(sym.getName()));
+    }
+  }
+}
 }
 
 void disassemble(const llvm::TargetMachine &tm,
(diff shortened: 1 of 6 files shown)


Hardcode84:
Not very well tested yet

Hardcode84:
Still disabled for llvm < 5.0 because too much ifdef clutter required but otherwise look good.

Ping @kinke

JohanEngelen:
You should add a test that tests that there is actual asm in the outputted text.

Hardcode84:
It is interesting qustion, how to properly test this. Asm output is different between platforms and llvm versions. For now I only checking asm output callback is being called.

JohanEngelen:
> Asm output is different between platforms and llvm versions.

You can test some part that is common across all platforms and versions. 
See e.g. `codegen/asm_output.d`.

Hardcode84:
Added simple check function and global var symbols present in asm (and found a bug of course).

Hardcode84:
Ping @kinke @JohanEngelen

runtime/jit-rt/cpp-so/disassembler.cpp
```diff
@@ -0,0 +1,319 @@
...
+    auto sym = name.empty() ? context.createTempSymbol("", false)
+                            : context.getOrCreateSymbol(name);
+    assert(nullptr != sym);
+    labelsPos.insert({pos, sym});
+    labelsTargets.insert({target, sym});
+  }
+
+  llvm::MCSymbol *getPosLabel(uint64_t pos) const {
+    auto it = labelsPos.find(pos);
+    if (labelsPos.end() != it) {
```
JohanEngelen: You have many (all?) comparisons in if-statements reversed. I find this harder to read, as it is contrary to convention. Any particular reason for it? Otherwise, please make it match with the rest of the codebase.

Hardcode84: I am used to this defensive style, when errors like `if(nullptr = smth)` will cause compilation error. But I can fix this.

kinke: I prefer the usual style too but also think that content is way more important than style, especially as long as it is consistent per file. So if you don't mind changing the order, fine, but it's definitely not a merge requirement if you have better use for your time. ;)

Issue #2384, opened 2017-10-23
Enhancement: Add '--resetOnly' to ldc-build-runtime


jondegenhardt:
It would be convenient to be able to run `ldc-build-runtime` so that it only ran the `--reset` logic, without actually building. I'm finding this would be useful for trying to create makefiles that use `ldc-build-only`. Essentially, a `make clean` operation. I'm finding I want to reset once for series of builds, but not on every build.

kinke:
If you want to implement it yourself, here's the file: https://github.com/ldc-developers/ldc/blob/master/runtime/ldc-build-runtime.d.in

jondegenhardt:
Thanks. The code looks straightforward. I'll need know how/where ".in" file is getting processed. I'm sure I'll be able to figure it out by looking around the code, if there something you can point me to that'd be useful. Also, if there is automated testing  that should be updated.

kinke:
The ".in" file is processed by CMake, replacing the `@VARIABLE@` occurrences. Those are used in string literals, so the file should compile as-is after renaming to .d, and those vars aren't required for your purpose here (used to figure out the source download URL after resetting the dir). The executable is built on CI, that's the only test. ;)

PR #2408, opened 2017-11-12
ldc-build-runtime: Add --resetOnly command. Catch getopt exceptions.

jondegenhardt:
PR makes two changes to `ldc-build-runtime`:

- Adds a `--resetOnly` option (Issue #2384). This does the `--reset` operation, but doesn't do the build.
- Catches exceptions thrown by `getopt`. This provides nicer error message formatting to the user.

On the latter item: Typical example is passing a command line argument that doesn't exist, like `ldc-build-runtime --version`. The argument doesn't exist, triggering an exception. Because the exception isn't caught, the user see a stack trace. This change displays the error text only, not the stack trace.

### DIFF
--- a/runtime/ldc-build-runtime.d.in
+++ b/runtime/ldc-build-runtime.d.in
@@ -9,6 +9,7 @@ struct Config {
     string ldcExecutable;
     string buildDir;
     bool resetBuildDir;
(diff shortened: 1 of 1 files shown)


kinke:
Thx Jon.

PR #2417, opened 2017-11-19
Android: Use proper C assert function signature

kinke:
Pinging @joakim-noah.


joakim-noah:
Thanks for adding, would have taken me longer to do it myself.

Issue #2415, opened 2017-11-18
ICE on chained ref-returning opIndex op-assign


Temtaime:
```

struct Matrix(uint _M)
{
	enum
	{
		M = _M,
		isVector = M == 1
	}

	union
	{
		float[] flat ;
		float[]A;
	}

	static if(isVector)
		ref opIndex(size_t i) { return flat[i]; }
	else
		ref opIndex(size_t ) 		{
			return *cast(Matrix!(1) *)A;
		}

}

alias Matrix4 = Matrix!(4);

void draw()
{
	Matrix4 mat;

	auto k = mat ;
	k[1][2] += 0.;
}
```

ldc2 crashes
dmd works
OS: windows

kinke:
Thanks, that's an interesting bug. Apparently the lhs matrix element being incremented loses its lvalue-ness after casting from float to double, which makes absolute sense to me. The front-end rewrites `k[1][2] += 0.`, with `k` being a matrix of floats, to `((ref Matrix!(float, 1u) __dop2 = k.opIndex(1LU);) , cast(double)__dop2.opIndex(2LU)) += 0.000000 @ float`, which seems clearly wrong.

As workaround, you can use `k[1][2] += 0.0f;` instead.

Temtaime:
A little more usable version

```
struct Matrix(uint M, uint N = M)
{
	union
	{
		float[M * N] flat;
		float[N][M] A;
	}

	static if(N == 1)
	{
		ref opIndex(size_t i) { return flat[i]; }
	}
	else
	{
		ref opIndex(size_t i) { return *cast(Matrix!(M, 1) *)A[i].ptr; }
	}
}

void main()
{
	Matrix!4 mat;
	mat[1][2] += 0.;
}
```

kinke:
Wrt. the original issue, LDC handles that binassign quirk in the AST wrt. lhs lvalue casting, but it doesn't handle that inside a comma expression. After a quick glance, neither does DMD (they also [skip over casts](https://github.com/dlang/dmd/blob/master/src/ddmd/e2ir.d#L2131-L2159), but don't handle comma expressions in a special way afaict).

Wrt. the arrayop warnings, those shouldn't be reported, so yes, please post a new issue with testcase.

kinke:
DMD seems to handle the comma expressions in `toAddressElem()`; we don't have that option as our `DValue` results of `toElem()` calls don't feature nested DValues as DMD's `elem` structs.

[...]

PR #2418, opened 2017-11-19
Fix issue #2415

kinke:


JohanEngelen:
This introduced regression #2471.
New fix being worked on: #2479

Issue #2422, opened 2017-11-21
ldc2 exits silently when using class from foreign module in associative array literal


FeepingCreature:
Testcase:

bar.d:

    class Empty { }

foo.d:

    import bar;
    void main() { auto foo = ["": new Empty]; }

Command: `ldc2 foo.d`
Expected result: linker error
Actual result: silent exit with error code 1.

What's happening here is that ldc attempts to *speculatively* generate the associative array literal as a constant expression, but fails because "new Empty" cannot be constant-evaluated because we do not pass bar.d on the commandline. As a result, global.errors is nonzero in `IrTypeClass::get()` (ir/irtypeclass.cpp:102), even though no error has been output. (The silent error can be seen by passing `-verrors-spec`.) I'm not sure why that error check is in there at all.

See also #2423

thewilsonator:
Fixed by #2423

PR #2423, opened 2017-11-21
fix issue 2422 - remove overzealous error check

FeepingCreature:
global.errors may be nonzero due to speculative error at this point!
ie. when trying to speculatively make a constant class expression

### DIFF
--- a/ir/irtypeclass.cpp
+++ b/ir/irtypeclass.cpp
@@ -99,10 +99,6 @@ IrTypeClass *IrTypeClass::get(ClassDeclaration *cd) {
     builder.addTailPadding(cd->structsize);
   }
 
-  if (global.errors) {
-    fatal();
-  }
-
   // set struct body and copy GEP indices
   isaStruct(t->type)->setBody(builder.defaultTypes(), t->packed);
   t->varGEPIndices = builder.varGEPIndices();
--- a/tests/compilable/gh2422.d
+++ b/tests/compilable/gh2422.d
@@ -1,0 +1,5 @@
+// RUN: %ldc -c -I%S/inputs %s
+
+import gh2422a;
+
+void main() { auto aa = ["": new Empty]; }
--- a/tests/compilable/inputs/gh2422a.d
+++ b/tests/compilable/inputs/gh2422a.d
@@ -1,0 +1,1 @@
+class Empty { }


thewilsonator:
Please add the test case from #2422

thewilsonator:
Also did you mean to PR this from your master branch?

FeepingCreature:
I don't know how to do it properly. Yes? Was that wrong?

thewilsonator:
That test is designed to fail, but to actually produce an error message yes?
In which case you `// RUN:` command should start with `not` to make programs that fail succeed and vice versa. 
You should also include a `// CHECK:` to ensure that the expected error occurs.

thewilsonator:
Also I'm pretty sure there is a way to check if the code is being speculated, in which case you should probably leave the check in place and add an additional constraint on in to be not in speculative.

FeepingCreature:
The test should not produce an error; it should run normally and output an .o. (It doesn't.) I just said the error thing in the bug report to clarify that the issue does not happen at all if you pass both .d files on the command line.

I don't know how to check if we're being speculated.

FeepingCreature:
Update: this seems to have been added [nine years ago](https://github.com/ldc-developers/ldc/blob/909c6dae184dd96aac7dceb6d31032c7e99f84aa/gen/classes.cpp) to check for errors occurring while adding interface vtables to a class. I don't think that process can even error anymore? I don't see any code in there that can.

thewilsonator:
Fair enough, will merge once green. Semaphore failed with an IO error (OOM?), probably not due to this.

FeepingCreature:
A single Travis CI build seems to have timed out. Semaphore fails in unrelated tests. I don't know what I should do to fix this.

kinke:
[Semaphore just went out of memory (they use a RAM disk), this happens from time to time unfortunately.]

JohanEngelen:
> global.errors may be nonzero due to speculative error at this point!
> ie. when trying to speculatively make a constant class expression

So then the correct fix is to gag the errors during speculative parts, right? The current fix hides real errors too.

FeepingCreature:
As I understand it, real errors will still error out later on.

JohanEngelen:
I agree that it is strange that there is a check so deep. But how come it is triggered here, and not later? Perhaps it is because the error is triggered while still inside a speculative context? (such that the error count is restored back to zero at the end of speculative call chain)

FeepingCreature:
Yes, that is what happens, as far as I can tell. Hence the removal. I'm not aware of a way to condition on errors being gagged that's being used elsewhere in the code.

FeepingCreature:
Ping...? I don't know how to restart the Travis build.

thewilsonator:
There is only a single timeout, I'll just merge it. For future reference pushing a new commit or close/reopen the issue will restart the CI jobs.

Issue #2425, opened 2017-11-22
betterC tried check array bounds


deviator:
Hello. I try build this simple example:
```d
import core.stdc.stdio;
import std.algorithm : min;

extern (C) void main()
{
    char[256] buf;
    buf[] = '\0';

    auto str = "hello world";
    auto ln = min(buf.length, str.length);
    buf[0..ln] = str[0..ln];
    printf("%s\n", buf.ptr);
}
```
`ldc2 -betterC bettercarray2.d`

And I get error: 
```
bettercarray2.o: In function `main':
bettercarray2.d:(.text.main[main]+0xd0): undefined reference to `_d_array_slice_copy'
collect2: error: ld returned 1 exit status
Error: /usr/bin/gcc failed with status: 1
```

On dmd 2.076.1 I have [same problem](https://forum.dlang.org/post/qcsgtxdiewfeiwckagqj@forum.dlang.org), but it solves by `-noboundscheck` flag. 
If I try `-boundscheck=off` on ldc2 I have no effect.

ldc2 version is [`1.6.0-beta1`](https://github.com/ldc-developers/ldc/releases/tag/v1.6.0-beta1) (prebuild for linux x86_64)

kinke:
As the name of the function suggests, this isn't directly related to array bounds checks, but with `buf[0..ln] = str[0..ln]` being mapped to druntime function `_d_array_slice_copy()` (which probably also checks that both slices feature the same length). We'll have to check what DMD does in this case; it's most likely unrelated to `-betterC`.

kinke:
See https://github.com/ldc-developers/druntime/blob/ldc/src/ldc/arrayinit.d#L142. According to the comment, it's only used for enabled assertions (also checking that the slices don't overlap); so an additional `-release` makes it work: https://godbolt.org/g/joZVzt

deviator:
> We'll have to check what DMD does in this case; it's most likely unrelated to -betterC.

It related to `-betterC` if it try use runtime when code builds with `-betterC` flag. `-betterC` must don't use runtime and all calls must be replaced with inline code or something else (c-runtime equivalent for example). But it's behavior isn't related to ldc, it's must be implement in dmd frontend...

This issue about difference between behavior of ldc and dmd.

kinke:
> It related to -betterC if it try use runtime when code builds with -betterC flag.

This divergence between LDC and DMD has absolutely nothing to do with `-betterC`.

[...]

PR #2426, opened 2017-11-22
Enforce SimplifyDRuntimeCalls pass for -betterC

kinke:
To reduce druntime dependencies. Fixes issue #2425.


kinke:
Argh, depends on optimized IR (identical length values).

PR #2430, opened 2017-11-30
Merge 2.077.1 front-end and stdlibs

kinke:


joakim-noah:
Wow, master finally caught up to current dmd, like seeing the dog finally catch the car. 🚗 🐕

PR #2433, opened 2017-12-02
Extend LDC_inline_ir pragma

Hardcode84:
Add extended version of  `LDC_inline_ir` pragma which allow to add arbitrary code before and after function.

### DIFF
--- a/gen/inlineir.cpp
+++ b/gen/inlineir.cpp
@@ -42,14 +42,69 @@ void copyFnAttributes(llvm::Function *wannabe, llvm::Function *idol) {
   auto fnAttrSet = attrSet.getFnAttributes();
   wannabe->addAttributes(LLAttributeSet::FunctionIndex, fnAttrSet);
 }
-} // anonymous namespace
 
-namespace {
 std::string exprToString(StringExp *strexp) {
   assert(strexp != nullptr);
   assert(strexp->sz == 1);
   return std::string(strexp->toPtr(), strexp->numberOfCodeUnits());
 }
+} // anonymous namespace
+
+void DtoCheckInlineIRPragma(Identifier *ident, Dsymbol *s) {
+  assert(ident != nullptr);
+  assert(s != nullptr);
+  if (TemplateDeclaration *td = s->isTemplateDeclaration()) {
+    Dsymbol *member = td->onemember;
+    if (!member) {
+      error(s->loc, "the `%s` pragma template must have exactly one member",
+            ident->toChars());
+      fatal();
+    }
+    FuncDeclaration *fun = member->isFuncDeclaration();
+    if (!fun) {
+      error(
+          s->loc,
+          "the `%s` pragma template's member must be a function declaration",
+          ident->toChars());
+      fatal();
+    }
+    // The magic inlineIR template is one of
+    // pragma(LDC_inline_ir)
+    //   R inlineIR(string code, R, P...)(P);
+    // pragma(LDC_inline_ir)
+    //   R inlineIREx(string prefix, string code, string suffix, R, P...)(P);
+
+    TemplateParameters &params = *td->parameters;
+    bool valid_params =
+        (params.dim == 3 && params[1]->isTemplateTypeParameter() &&
+         params[2]->isTemplateTupleParameter()) ||
+        (params.dim == 5 && params[3]->isTemplateTypeParameter() &&
+         params[4]->isTemplateTupleParameter());
+
+    if (valid_params) {
+      TemplateValueParameter *p0 = params[0]->isTemplateValueParameter();
+      valid_params = valid_params && p0 && p0->valType == Type::tstring;
+      if (params.dim == 5) {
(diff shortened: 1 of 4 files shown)


Hardcode84:
Test https://github.com/ldc-developers/dmd-testsuite/pull/27

gen/inlineir.cpp
```diff
@@ -65,33 +74,55 @@ DValue *DtoInlineIRExpr(Loc &loc, FuncDeclaration *fdecl,
...
     Objects &objs = tinst->tdtypes;
-    assert(objs.dim == 3);
-
-    Expression *a0 = isExpression(objs[0]);
-    assert(a0);
-    StringExp *strexp = a0->toStringExp();
-    assert(strexp);
-    assert(strexp->sz == 1);
-    std::string code(strexp->toPtr(), strexp->numberOfCodeUnits());
+    assert(objs.dim == 3 || objs.dim == 5);
```
thewilsonator: Perhaps add a comment like:

The magic inlineIR template is one of 
```
pragma(LDC_inline_ir)
      R inlineIR(string code, R, P...)(P);
pragma(LDC_inline_ir)
      R inlineIREx(string prefix, string code, string suffix, R, P...)(P);
```
And create a struct like https://github.com/ldc-developers/ldc/blob/master/gen/dcompute/druntime.h#L28 to hold the data and validity checking.

JohanEngelen:
> Also clang-formatted changed files.

Thanks, but please submit such things as separate PRs. Such PRs can be pulled quickly.
Having unrelated changes in a PR makes reviewing hard, creates bad git history making reverting harder, etc...

About the changes in particular: which version of clang-format are you using? In our clang-format config file, we explicitly disable sorting of include files. Latest clang-format version is 5; some of the other formatting decisions are also different than clang-format 5 on my machine.

thewilsonator:
Is it possible to express a generic non temporal load/store with this?

Hardcode84:
Removed unrelated formatting changes.

> About the changes in particular: which version of clang-format are you using? In our clang-format config file, we explicitly disable sorting of include files. Latest clang-format version is 5; some of the other formatting decisions are also different than clang-format 5 on my machine.

I had explicit `-sort-includes` in my script and clang-format built from trunk couple months ago.

Hardcode84:
> Is it possible to express a generic non temporal load/store with this?

I think so, but where to put it as it's not related to simd directly?

thewilsonator:
I meant more along the lines of that it is possible to do, than actually do it.

gen/inlineir.cpp
Hardcode84: I don't think creating additional struct worth it. Added comment and `DtoCheckInlineIRPragma` function to keep all related check in one file.

thewilsonator: I suppose this is only used in one place. But I still think that having all that parameter extraction logic adds a lot of noise.

Hardcode84:
Ping

thewilsonator:
I'm fine with this. @JohanEngelen ?

gen/inlineir.cpp
```diff
@@ -42,8 +42,71 @@ void copyFnAttributes(llvm::Function *wannabe, llvm::Function *idol) {
...
+    //   R inlineIR(string code, R, P...)(P);
+    // pragma(LDC_inline_ir)
+    //   R inlineIREx(string prefix, string code, string suffix, R, P...)(P);
+
+    TemplateParameters &params = *td->parameters;
+    bool valid_params =
+        (params.dim == 3 && params[1]->isTemplateTypeParameter() &&
+         params[2]->isTemplateTupleParameter()) ||
+        (params.dim == 5 && params[3]->isTemplateTypeParameter() &&
+         params[4]->isTemplateTupleParameter());
```
JohanEngelen: in case `params.dim == 5`, the 1st and 2nd parameter types are not checked here, but later. Is there a way to write these checks a little nicer?
Perhaps first check that dim is either 3 or 5, and then index with e.g. `[params.dim-1]`.

gen/inlineir.cpp
```diff
@@ -63,35 +126,61 @@ DValue *DtoInlineIRExpr(Loc &loc, FuncDeclaration *fdecl,
...
+      Expression *a1 = isExpression(objs[1]);
+      assert(a1);
+      StringExp *strexp = a1->toStringExp();
+      Expression *a2 = isExpression(objs[2]);
+      assert(a2);
+      StringExp *suffexp = a2->toStringExp();
+      prefix = exprToString(prefexp);
+      code = exprToString(strexp);
+      suffix = exprToString(suffexp);
+
```
JohanEngelen: enter?

JohanEngelen:
I'd prefer an actual IR test, instead of the current runtime test.

Hardcode84:
> I'd prefer an actual IR test, instead of the current runtime test.

Will do in next few days.

Hardcode84:
This took more than "few days", sorry about that.
Added simple IR test but I am not very familiar with FileCheck tool.

JohanEngelen:
LGTM

PR #2434, opened 2017-12-02
[WIP] some experiments with i1 vec

Hardcode84:
Current status:
```
struct __VecBitBool
{
    bool b;
    alias b this;
}

alias V = Vector!(__VecBitBool[4]);

V foo(V v1) // Compiles
{
    v1[2] =true;
    return v1;
}
```

```
V foo() // Crash in front end
{
    V v1;
    v1[2] =true;
    return v1;
}
```

```
// Assertion failed: getTypeStoreSize(DtoType(type)) <= getTypeStoreSize(pointeeTyp
// e) && "LValue type mismatch, encountered type too small.", file D:\LDC\ldc\gen\l
// lvmhelpers.cpp, line 1865
V foo()
{
    V v1 = void;
    v1[2] =true;
    return v1;
}
```


thewilsonator:
Awesome! This will be useful for some of the odd function signatures  in the SPIR-V intrinsics. 
You should add tests for what works.

JohanEngelen:
> You should add tests for what works.

You should add tests. (period) Especially for what doesn't work during development because clearly it is needed!!! ;-)

PR #2440, opened 2017-12-06
Fix C++ header virtual function overrides

JohanEngelen:
A number of overrides were actually not overriding (instead hiding the virtual functions).
Fixed upstream in https://github.com/dlang/dmd/pull/7402 and https://github.com/dlang/dmd/pull/7401

Applying early because it can cause hard to find bugs and leads (thankfully) to a compiler warning flood.


kinke:
Thx Johan, the warning flood on OSX was really ugly (no warnings with MSVC...).

PR #2457, opened 2017-12-18
Add the ldc.attributes.assumeUsed attribute corresponding to "attribute((used))” in GNU C

JohanEngelen:
The attribute adds the symbol to `llvm.used`.

My use case: LTO is nicely aggressive in removing symbols, but some symbols at Weka are not supposed to be removed.

### DIFF
--- a/ddmd/id.d
+++ b/ddmd/id.d
@@ -442,6 +442,7 @@ immutable Msgtable[] msgtable =
     { "udaLLVMFastMathFlag", "llvmFastMathFlag" },
     { "udaSection", "section" },
     { "udaTarget", "target" },
+    { "udaAssumeUsed", "_assumeUsed" },
     { "udaWeak", "_weak" },
     { "udaCompute", "compute" },
     { "udaKernel", "_kernel" },
--- a/ddmd/id.h
+++ b/ddmd/id.h
@@ -58,6 +58,7 @@ public:
     static Identifier *udaSection;
     static Identifier *udaOptStrategy;
     static Identifier *udaTarget;
+    static Identifier *udaAssumeUsed;
     static Identifier *udaWeak;
     static Identifier *udaAllocSize;
     static Identifier *udaLLVMAttr;
--- a/gen/uda.cpp
+++ b/gen/uda.cpp
@@ -1,5 +1,6 @@
 #include "gen/uda.h"
 
+#include "gen/irstate.h"
 #include "gen/llvm.h"
 #include "gen/llvmhelpers.h"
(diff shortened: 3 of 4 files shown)


JohanEngelen:
I think with `export`, whole-program optimization can still strip the symbol? It does not preserve the symbol with LTO at the moment.

JohanEngelen:
(Note to self, but it is related to this PR.  Putting a symbol in a C-identifier section should also work: https://reviews.llvm.org/D35639 . But that's what Weka already does and it appears to not work with LLVM 5.0. Further investigation needed. )

kinke:
Currently yes, but I think David meant making `export` put it into llvm.used. As long as there's no use case for private symbols to be retained this way, that should be fine. [On Windows, the applied `dllexport` attribute seems to be enough to retain the symbol, so `export` would already work there.]

JohanEngelen:
Why would we add `export` symbols to `llvm.used`? What's the upside?
Note that adding things to `llvm.used` is a pessimization (performance degradation, binary size increase, etc.) This is about always retaining a symbol (so also after final static linking into an executable) that would otherwise be stripped. It's very advanced use, and strange to add that meaning to the `export` keyword. (you would not be able to strip any exported symbol out of a library then...)

JohanEngelen:
(`export` is about linkage and visibility; `used` is about retaining a symbol or not. They are related, but not the same and shouldn't be mixed. E.g. I might want to not strip a _private_ symbol because a piece of runtime-generated code might use it.)

JohanEngelen:
(Ah indeed, bad example. Another attempt: you can put non-exported symbols in a special section and use section_begin/end to access them, treating the section as a static array of those symbols.)

I'll rename it to `@assumeUsed`.  (renaming is annoying because of submodule involved... let's keep it at that?)

JohanEngelen:
> Btw: with this addition, LDC1.6.0 can build weka's very large program with -flto=thin !

And passes their (standard) testing! ^_^ Gratz on awesome compiler :-)

kinke:
Nice. So I think it'd be worth being included in 1.7 final, maybe after renaming the commit titles here and in druntime as well. - I'm not at my place and won't be for a week, so I'd be grateful if somebody else did the 1.7.0 tagging and src.tar.gz fixup so that we can release it still in 2017.

JohanEngelen:
No rush for this, so I'll keep it out of 1.7.0.
I created the release branches, is there any bugfix still out that needs inclusion?  https://github.com/ldc-developers/ldc/issues/2469

PR #2462, opened 2017-12-21
CircleCI test

kinke:


kinke:
`Bad credentials` => worked fine when uploading a file from my machine with exactly the same token in an environment variable. There, the MIME type didn't make a difference (`application/octet-stream`, as apparently always used by `github-release`, and `application/gzip`).

=> I guess we'll just have to live with fixing the .tar.gz src archive manually for the time being.

Issue #2391, opened 2017-10-27
undefined reference to `__multi3'


9il:

kinke:
Similar thing as for your issue #2262 - LLVM lowers the 128-bit multiplication to a `__multi3()` C function call. Google found an implementation in LLVM's compiler-rt libary: https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/multi3.c

Issue #818, opened 2014-12-27
Ship compiler-rt or audit for implicit calls to it


dnadlinger:
LLVM emits function calls for all sorts of operations which it can't efficiently generate target architecture code for. Some of the functions exist only in [compiler-rt](http://compiler-rt.llvm.org/), and not in libgcc. One example for this is wide overflow-checked integer arithmetic, see #813.

We need to either ship compiler-rt or make sure that we don't accidentally emit references to those symbols on accident.

kinke:
Our prebuilt packages now ship with the used compiler-rt libs; the builtins lib isn't linked by default though (only on Windows).

PR #2464, opened 2017-12-24
Remove ldc.profile.getData(string) to remove profile-rt Phobos dependency

JohanEngelen:
I think the extra functionality added into profile-rt (compared with LLVM's profile-rt) is of very little value, and perhaps not worth making our own version of compiler-rt's profile-rt. We could kill all that, and just use LLVM's profile-rt, but for now, I just want to remove the Phobos dependency in non-templated code.
Wit this PR only the templated code in ldc.profile depends on Phobos, so it's possible to link with profile-rt without requiring Phobos.

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -329,8 +329,6 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
   }
 
   // Link with profile-rt library when generating an instrumented binary.
-  // profile-rt uses Phobos (MD5 hashing) and therefore must be passed on the
-  // commandline before Phobos.
   if (global.params.genInstrProf) {
 #if LDC_LLVM_VER >= 308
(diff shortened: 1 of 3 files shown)


kinke:
> perhaps not worth making our own version of compiler-rt's profile-rt. We could kill all that, and just use LLVM's profile-rt

Yep I'd be very much in favor of depending on LLVM's compiler-rt. With LLVM 6.0 and its integration of libFuzzer as another compiler-rt module IIRC, we should be able to embed it as git submodule in LDC-LLVM and thus simplify its build script. Integrating more libs like XRay (#2465) and the builtins lib (#818, #2391) in addition to profile-rt, libFuzzer and sanitizers should be straight forward then.

JohanEngelen:
OK, so let this be the first step to removing our special profile-rt lib.

Issue #1873, opened 2016-11-09
[Enhancement] Add XRay support


JohanEngelen:
See https://reviews.llvm.org/D26386
Also: http://research.google.com/pubs/pub45287.html

It looks like it is a relatively small effort to add this to LDC, and it can be nicely compartmentalized. Adding the xray module from compiler-rt can be done exactly the same as the profile-rt runtime lib (`/runtime/profile-rt`).

Note that this is still actively developed in LLVM.

JohanEngelen:
I am actually against adding xray similarly to profile-rt by copying code. It'd be much better if we'd find a way to incorporate compiler-rt libraries without copying more code from compiler-rt into our repository.

JohanEngelen:
> It'd be much better if we'd find a way to incorporate compiler-rt libraries without copying more code from compiler-rt into our repository.

Which we now know how to do: look at how asan and fuzzer are included

JohanEngelen:
(we have this now!)

PR #2465, opened 2017-12-24
Add XRay support

JohanEngelen:
Submitting this as PR, so that people can see progress towards implementing #1873 .

It's a WIP, but I think it may already work on Linux (didn't test). Darwin support is on its way in compiler-rt/LLVM :)

Some refactoring is still needed now that the number of different instrumentation options is growing.


JohanEngelen:
https://llvm.org/docs/XRay.html

JohanEngelen:
XRay is only fully supported on Linux for now. Should I disable this on unsupported platforms? Or give a warning or error?

driver/linker-gcc.cpp
```diff
@@ -268,6 +270,34 @@ void ArgsBuilder::addFuzzLinkFlags() {
...
+    getFullCompilerRTLibPath("libclang_rt.xray"),
+  };
+
+  bool linkerLinux = global.params.targetTriple->isOSLinux();
+  if (!linkerLinux)
+    warning(Loc(), "XRay is not (fully) supported on non-Linux target OS.");
+
+  for (const auto &filepath : searchPaths) {
+    if (linkerLinux)
+      addLdFlag("--whole-archive");
```
kinke: In case this would be needed on OSX too, there's `-force_load` as used [here](https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L780).

JohanEngelen: Thanks, I've fixed it.
We'll have to see how LLVM will deal with XRay on OSX; it's currently a WIP and not supported.

JohanEngelen:
Any concerns before I merge this?

JohanEngelen:
More documentation here: https://llvm.org/docs/XRayExample.html

PR #2468, opened 2017-12-27
Link the druntime and Phobos test-runners with LDC.

JohanEngelen:
This is needed for example when building with -fsanitize=address. Linking with LDC is needed in that case such that the correct ASan runtime library is linked in (and rpath is set correctly for it). It is similar for `-flto`.

It's a bit of a hack, but works on my machine. Let's see what the CI systems say.

My testcase:
```
ldc-build-runtime --dFlags='-fsanitize=address;-fsanitize-blacklist=/Users/johan/ldc/ldc2-1.7.0-beta1-osx-x86_64/asan_blacklist.txt' BUILD_SHARED_LIBS=OFF --testrunners
```
 Adding `-fsanitize=address` to the `--linkerFlags` doesn't work, because then the wrong version of address sanitizer runtime is used (the one provided by my system's clang is different from the ldc2-1.7.0 compiler)

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -445,8 +445,10 @@ endmacro()
 # using LDC as archiver. LDC handles (cross-)archiving internally via LLVM
 # and supports LTO objects.
 set(CMAKE_D_CREATE_STATIC_LIBRARY "${LDC_EXE_FULL} -lib -of=<TARGET> <OBJECTS>")
+set(CMAKE_D_LINK_EXECUTABLE "${LDC_EXE_FULL} -of=<TARGET> <LINK_FLAGS> <OBJECTS>")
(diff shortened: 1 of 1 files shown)


kinke:
This is a bit problematic. Firstly, we don't want any default druntime/Phobos lib to be linked in for the test runners. Secondly, cross-linking would now require setting up the LDC flags accordingly too (while still having to set up the CMake linker for the shared runtime libs).

JohanEngelen:
yeah, I think I'll give up on this.  It is easier to just manually specify the correct link flags for now I think. For ASan, manually specifying the correct ASan lib is not too bad.
(for LTO it doesn't matter as the library is not doing any codegen, just bundling of the bitcode files)

PR #2474, opened 2017-12-28
Add IR-based PGO

JohanEngelen:
First, some refactoring such that the code doesn't assume just one kind of PGO.

### DIFF
--- a/driver/cl_options_instrumentation.h
+++ b/driver/cl_options_instrumentation.h
@@ -21,15 +21,34 @@
 namespace opts {
 namespace cl = llvm::cl;
 
-// PGO options
-extern cl::opt<std::string> genfileInstrProf;
-extern cl::opt<std::string> usefileInstrProf;
-
 extern cl::opt<bool> instrumentFunctions;
 
 /// This initializes the instrumentation options, and checks the validity of the
 /// commandline flags. It should be called only once.
 void initializeInstrumentationOptionsFromCmdline();
+
+enum PGOKind {
+  PGO_None,
+  PGO_ASTBasedInstr,
+  PGO_ASTBasedUse,
+  PGO_IRBasedInstr,
+  PGO_IRBasedUse,
+};
+extern PGOKind pgoMode;
+inline bool isInstrumentingForPGO() {
+  return pgoMode == PGO_ASTBasedInstr || pgoMode == PGO_IRBasedInstr;
+}
+inline bool isUsingPGOProfile() {
+  return pgoMode == PGO_ASTBasedUse || pgoMode == PGO_IRBasedUse;
+}
+inline bool isInstrumentingForASTBasedPGO() {
+  return pgoMode == PGO_ASTBasedInstr;
+}
+inline bool isUsingASTBasedPGOProfile() { return pgoMode == PGO_ASTBasedUse; }
+inline bool isInstrumentingForIRBasedPGO() {
+  return pgoMode == PGO_IRBasedInstr;
+}
+inline bool isUsingIRBasedPGOProfile() { return pgoMode == PGO_IRBasedUse; }
 
 } // namespace opts
 #endif // LDC_DRIVER_CL_OPTIONS_INSTRUMENTATION_H
--- a/driver/cl_options_instrumentation.h
+++ b/driver/cl_options_instrumentation.h
@@ -24,7 +24,8 @@ namespace cl = llvm::cl;
 extern cl::opt<bool> instrumentFunctions;
 
 /// This initializes the instrumentation options, and checks the validity of the
-/// commandline flags. It should be called only once.
+/// commandline flags. targetTriple should be initialized before calling this.
+/// It should be called only once.
 void initializeInstrumentationOptionsFromCmdline();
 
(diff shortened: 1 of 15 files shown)


driver/cl_options_instrumentation.h
```diff
@@ -21,16 +21,32 @@
...
 
+enum PGOKind {
+  PGO_None,
+  PGO_FrontendBasedInstr,
+  PGO_FrontendBasedUse,
+  PGO_IRBasedInstr,
+  PGO_IRBasedUse,
+};
+extern PGOKind pgoMode;
+inline bool isDoingPGOInstr() {
```
klickverbot: I don't have a strong opinion on this, but this reads a bit odd – maybe `isPGOInstrumenting`, `instrumentingForPGO`, or something like that?

JohanEngelen: and then the other `isUsingPGOProfile` ?

JohanEngelen:
I used "frontend" because I think that's what is usually used in LLVM circles. In LLVM code, it's called "ClangPGO" (frontend based) and "LLVMPGO" (IR based).

JohanEngelen:
I'm renaming it to "AST Based", I like that better indeed.
Got IR-based to work aswell, very easy :)
(and it seems that using a sample-based PGO profile is also very easy now!)

JohanEngelen:
@jondegenhardt Done! :-)   Important difference between AST- (old) and IR-based PGO in practical use is that the other commandline flags must be the same with IR PGO. This means that you have to specify `-O3` when _instrumenting_ with IR PGO as well.  (you'll see "hash mismatch" warnings without it during `-fprofile-use`, meaning that the profile cannot be used and you won't get the optimizations you want)

jondegenhardt:
Very nice! I won't have a chance to try it until after new year, but I'll give it shot then.

JohanEngelen:
Strange, anybody know why the file is not created? The CI systems error with:
```
773: $ "/Users/travis/build/ldc-developers/ldc/bin/ldc2" "-O3" "-fprofile-generate=/Users/travis/build/ldc-developers/ldc/tests/PGO/Output/irbased_indirect_calls.d.tmp.profraw" "-run" "/Users/travis/build/ldc-developers/ldc/tests/PGO/irbased_indirect_calls.d"

773: $ "/Users/travis/build/ldc-developers/ldc/bin/ldc-profdata" "merge" "/Users/travis/build/ldc-developers/ldc/tests/PGO/Output/irbased_indirect_calls.d.tmp.profraw" "-o" "/Users/travis/build/ldc-developers/ldc/tests/PGO/Output/irbased_indirect_calls.d.tmp.profdata"

773: # command stderr:

773: error: /Users/travis/build/ldc-developers/ldc/tests/PGO/Output/irbased_indirect_calls.d.tmp.profraw: No such file or directory
```
But the lit-test works fine on my machine.

JohanEngelen:
The failure on Windows is due to some bug in LLVM. I've a testcase and will submit it to their bugtracker.
So for now, I'll have to disable this on Windows...

JohanEngelen:
Shall I fully disable it on Windows, or only mark the test as XFail ?  (only generating the profile for a Windows target is broken; profile usage still works)
I can also give a nice error message when a Windows target is chosen for profile generation?  (right now, LLVM just outputs "Error:")

JohanEngelen:
(needless to say, I'd like to get this in quick, so other folks can start testing)

kinke:
Thanks for doing this. Will it eventually supersede the AST-based one? [I would be looking forward to the resulting code cleanup.]

Wrt. Windows, XFAIL sounds good to me, as does a proper error message for the time being.

JohanEngelen:
> Will it eventually supersede the AST-based one?

Maybe. The AST-based code makes a few PGO items easier and facilitates prototyping things (like the virtual call promotion I worked on). But probably most importantly, the AST-based instrumentation is the same as what's needed for code coverage. (https://clang.llvm.org/docs/SourceBasedCodeCoverage.html)
I don't have plans right now on taking the AST-based PGO further though, so I think people should move towards the IR-based PGO (Clang's AST-based PGO also stopped progressing, and all PGO effort is on IR-based PGO afaict).

>  [I would be looking forward to the resulting code cleanup.]

"cleanup" :'( 


I'll add the error for Windows targets. I've already sent a question about the error to LLVM folks, I think it's just a bug somewhere (I _can_ get it to (appear to) work with llvm's cmdline tools).

kinke:
> "cleanup" :'(

What I actually meant was code reduction, the less code to maintain, the better ;) - even more so if scattered around the code base like this one.

driver/main.cpp
```diff
@@ -1033,6 +1032,8 @@ int cppmain(int argc, char **argv) {
     global.lib_ext = "a";
   }
 
+  opts::initializeInstrumentationOptionsFromCmdline();
```
JohanEngelen: i had to place the call here, after the triple has been initialized. The option sanity-checking needs to know the target triple, and I didn't want to separate the sanity checking from the initialization code.

JohanEngelen:
Any concerns before I pull this?

PR #2475, opened 2017-12-28
Fix unistd header inclusion for all builds.

JohanEngelen:
This should fix remaining issues with unistd.h.
@dkgroot


dkgroot:
@JohanEngelen I think this fix should be ok. Travis CI only failed on std.date and timedout.

PR #2477, opened 2017-12-30
Add DMD-style function trace profiling.

JohanEngelen:
There are big issues with DMD-style profiling, and I would never recommend people to use it.
But because it is easy to support, I added it.

Depends on https://github.com/dlang/druntime/pull/2024

I couldn't find how to enable the druntime testing of `-profile`...

Adding GC profiling is also easy: we simply have to rename calls to e.g. `newclass` to `newclassTrace` (and add extra arguments).


JohanEngelen:
todo: fix the memcmp

JohanEngelen:
https://github.com/dlang/druntime/pull/2024 was merged upstream

JohanEngelen:
- fixed the memcmp.
- enabled -profile testing in dmdtestsuite
- rebased the druntime changes upon master druntime.

JohanEngelen:
(off topic: while working on this, I noticed that there are a bunch of druntime extern(C) functions that do not start with double underscores, but really should.)

PR #2478, opened 2017-12-31
Add possibility to set custom LLVM_LIBRARY_DIRS variable from outside

ThomasMader:
This PR tries to solve the problem on NixOS that LLVM_LIBRARY_DIRS can not be correctly determined via llvm-config --libdir.

See https://github.com/NixOS/nixpkgs/issues/33127 for more details.

### DIFF
--- a/cmake/Modules/FindLLVM.cmake
+++ b/cmake/Modules/FindLLVM.cmake
@@ -27,6 +27,9 @@
 # We also want an user-specified LLVM_ROOT_DIR to take precedence over the
 # system default locations such as /usr/local/bin. Executing find_program()
 # multiples times is the approach recommended in the docs.
+#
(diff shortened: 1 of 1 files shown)


JohanEngelen:
I don't mind this too much, and we should get a fix for the problem into the 1.7.0 release.

Because ` llvm-config --cmakedir` is the better fix, I'd rather have (with if-else thing because it's only available from LLVM 4). That way, we prevent people from depending on `LLVM_LIBRARY_DIRS` for other reasons and we cannot remove it without a fight.

@ThomasMader  Could you take the effort to do the `--cmakedir` solution? Thanks.

JohanEngelen:
Alternative fix is here: https://github.com/ldc-developers/ldc/pull/2482

ThomasMader:
Closing this as https://github.com/ldc-developers/ldc/pull/2482 is the better solution.

Issue #2471, opened 2017-12-27
ldc-1.7.0-beta1 (regression): Internal compiler error with implicit cast


rickrankin:
I'm getting an internal compiler error with ldc-1.7.0-beta1 on code that works fine with dmd. When I compile the sample below, I get 

```
bugs\ldc-iterator.d(44): Error: Internal Compiler Error: Invalid struct cast from `Iterator!string` to `char`
```

but as I said, the code works as-is with dmd. If I changed line 44 to

```d
char ch = cast(immutable(char)) it++;
```

then the code compiles and works as expected with ldc, also. This is the first time I've tried to compile with ldc, but this code has been working with dmd for a long time. It seems like ldc is ignoring the opCast operator _unless_ the cast is explicit, but that's just a guess based on the error message and some ways I've tried to alter the definition of opCast to get ldc to accept it. So far, the only way I've been able to get it to work with ldc is by using an explicit cast.

```d
module example;

struct Iterator(T : T[])
{
  alias ThisType = Iterator!(T[]);
  alias ItemType = T;

  this(T[] container, size_t index = 0)
  {
    container_ = container;
    index_ = index;
  }

  U opCast(U)() const
    if (is(U == ItemType))
  {
    return this ? container_[index_] : ItemType.init;
  }

  ref ThisType opUnary(string op)()
    if (op == "++" || op == "--")
  {
    mixin(op ~ "index_;");
    return this;
  }

  @property ItemType value() const        { return container_[index_]; }

  alias value this;

  private T[] container_  = null;
  private size_t index_   = 0;
}

void main()
{
  import std.stdio : writefln;

  string test = "abcde";
  auto it = Iterator!string(test);

  writefln("ItemType = %s", typeid(Iterator!string.ItemType));
  writefln("it = %c", it);
  char ch = it++;
  writefln("ch = %c, it = %c", ch, it);
}
```

kinke:
It has something to do with the postfix increment (and is not specific to Windows). `char ch = it; it++;` compiles, as does `char ch = ++it;` (the latter with different result obviously). I don't have my machine at my disposal, so I can't look into more details yet.

JohanEngelen:
LDC 1.6.0 on godbolt works, but 1.7 doens't. So it's a 1.7 regression.

kinke:
Most likely introduced by #2418; AST diff:

1.6:
```
[...]

PR #2479, opened 2018-01-01
Try to fix issue #2471

kinke:


gen/toir.cpp
```diff
@@ -600,6 +600,9 @@ class ToElemVisitor : public Visitor {
   template <BinOpFunc binOpFunc, bool useLValTypeForBinOp>
   static DValue *binAssign(BinAssignExp *e) {
     Expression *lvalExp = skipOverCasts(e->e1);
+    while (lvalExp->op == TOKcomma) {
+      lvalExp = skipOverCasts(static_cast<CommaExp *>(lvalExp)->e2);
+    }
     DValue *lhsLVal = toElem(lvalExp);
```
kinke: Ah nope this won't work, we need to evaluate the CommaExp and cannot descend into it. We could try getting rid of all nested casts of CommaExp right-hand-sides before `toElem()`enting a CommaExp.

kinke:
Developing via GitHub online editor is a bit painful but this should work now and be a nicer fix for #2415.

JohanEngelen:
> Developing via GitHub online editor is a bit painful

>:(  indeed. Nice that you are working on it. Sorry I can't chip in at the moment.
CI says it's indeed fixed.

PR #2482, opened 2018-01-04
CMake: use llvm-config to determine LLVM's cmake directory, if possible.

JohanEngelen:
This should fix https://github.com/NixOS/nixpkgs/issues/33127

@ThomasMader can you verify that this fixes the issue for NixOS?  (LLVM >= 4.0)


JohanEngelen:
This is an alternative to https://github.com/ldc-developers/ldc/pull/2478

ThomasMader:
I haven't done a full package build because I would need a release for that but reviewed the changes and started a cmake build via _cmake .._ inside _build_ directory with your changes and with your changes reverted and can say that all is good. :+1:

Issue #2480, opened 2018-01-03
Don't eliminate frame pointer at `-O0`


JohanEngelen:
For previous discussion on this topic: #2096

I think we should not eliminate the frame pointer at `-O0`. It is not just our own stack traces that are (no longer) affected. ASan also does not show correct/complete stack traces with frame pointer elimination. 

Clang does not even eliminate the frame pointer at `-O3` on macOS. We don't have to mimic that, but at `-O0` I feel we really should keep the fp.

klickverbot:
I agree, we shouldn't do frame-pointer elimination at -O0. (Is this possibly a consequence of a LLVM API change at some point?) Just to recap, in #2096 I was merely arguing for a better backtrace mechanism, rather than against disabling FPE.

JohanEngelen:
Related: https://github.com/ldc-developers/ldc/issues/2222

kinke:
Fixed by #2483.

PR #2483, opened 2018-01-04
Don't eliminate frame pointer by default at -O0 (issue #2480)

kinke:


tests/codegen/attr_targetoptions.d
```diff
@@ -1,26 +1,26 @@
 // Tests that our TargetMachine options are added as function attributes
 
 // RUN: %ldc -c -output-ll -of=%t.ll %s && FileCheck %s --check-prefix=DEFAULT < %t.ll
-// RUN: %ldc -c -output-ll -of=%t.ll %s -disable-fp-elim && FileCheck %s --check-prefix=FRAMEPTR < %t.ll
+// RUN: %ldc -c -output-ll -of=%t.ll %s -O2 && FileCheck %s --check-prefix=OPTIMIZED < %t.ll
```
JohanEngelen: we should test the combination of `-O2` and `-disable-fp-elim` aswell.

gen/functions.cpp
```diff
@@ -460,8 +460,11 @@ void applyTargetMachineAttributes(llvm::Function &func,
   func.addFnAttr("no-nans-fp-math", TO.NoNaNsFPMath ? "true" : "false");
 
   // Frame pointer elimination
-  func.addFnAttr("no-frame-pointer-elim",
-                 opts::disableFPElim() ? "true" : "false");
+  const llvm::cl::boolOrDefault disableFPElimEnum = opts::disableFPElim();
+  const bool disableFPElim =
+      disableFPElimEnum == llvm::cl::BOU_TRUE ||
+      (disableFPElimEnum == llvm::cl::BOU_UNSET && !isOptimizationEnabled());
```
JohanEngelen: Let's move this logic into `bool opts::disableFPElim()` and have `opts::disableFPElim()` be the definitive answer to whether FPElim is disabled or not. So then the only code change would be in the implementation of `bool opts::disableFPElim()`.

kinke: I did that first but changed it to this, in order not to have any custom LDC-specific logic in this module, whose scope is to only expose the required functionality (by LDC and LLD) of the special LLVM header.

tests/codegen/attr_targetoptions.d
kinke: Should => could :], as we could check at least 3 other options by now too (`-enable-{unsafe-fp-math,no-infs-fp-math,no-nans-fp-math`), but I thought that'd be an overkill.

kinke:
A dmd-testsuite test recently added by @rainers (https://github.com/dlang/dmd/commit/2db369ec86772aa802eedd417559c2d78c956f1a#diff-dcdd7a556bb06c2fe688ec531d0ae4c7) now fails on 32-bit Linux x86: `[0x0000003a]  Special opcode 5: advance Address by 0 to 0x14 and Line by 0 to 3`

tests/codegen/attr_targetoptions.d
JohanEngelen: Yes, those should be tested too, but that's another topic.

gen/functions.cpp
JohanEngelen: OK, so then it's better to move it to the general cmdline options file. Right now, the actual "do we disable FP elim" is determined here deep down, instead of in the options code.

kinke:
Reworked.

Issue #2024, opened 2017-03-06
Android and ARM regressions in master 1.2


joakim-noah:
I recently tried building master both natively and as a linux/x64 cross-compiler and ran into these issues with the tests.  I will look into these more and submit patches for the phobos bugs upstream, would like some help with the LDC ones.

LDC issues
- `std.digest.murmurhash` segfaults with a `Bus error` on [its last test](https://github.com/ldc-developers/phobos/blob/4866f6c79dcb36e6e2959b8bd8a5125f52ef0ad4/std/digest/murmurhash.d#L750), when it tries to compute the `unalignedHash` with any llvm optimizations.  This is reproducible going back to ltsmaster: all of 0.17.3, 1.1.1, and 1.2 (each built against llvm 3.9.1) pass this test without optimizations but segfault with `-01` or higher optimizations.  This module was added with 1.2 so obviously it wasn't tested with those older D versions before, but it seems to have uncovered some ARM codegen issue.
- `std.random` segfaults with this error when compiled with `-O2` or higher:
`Fatal error in EH code: _Unwind_RaiseException failed with reason code: 9
Aborted`
A gdb backtrace shows that it fails in [the massive last test block](https://github.com/ldc-developers/phobos/blob/77fd9145e96ca096c026f18fa6833b62bc704192/std/random.d#L2741).  ldc 0.17.3 can't build the 1.2 version of this module's tests because of a static assert, while 1.1.1 has no problem with it, even with all optimizations.  This looks like the ARM codegen errors we were getting before, @smolt?

Phobos issues
- [Vlad added a new linux-only test that seems to expect too much precision from `std.file` on Android](https://github.com/dlang/phobos/pull/4089/files#diff-52580fb75b304ba7b04a6b178fe6cdf4R1230), it passes if I change the last digit from a 7 to 0.
- [tsbockman added some tests to `std.json`](https://github.com/dlang/phobos/pull/4345/files#diff-7441f65d4929cc6c45a3094b58973fcdR1739) that needed to be [relaxed for platforms with 64-bit reals](https://github.com/ldc-developers/phobos/commit/0d14f4ed441b4e70997e930d23ab47f26f7483d7#diff-7441f65d4929cc6c45a3094b58973fcdR1704).  Adding `Android` to those two version blocks gets the tests to pass again.
[...]

PR #2486, opened 2018-01-06
Merge 2.078.1 front-end and stdlibs

kinke:


kinke:
The ICE when compiling LDC on Windows is bad; it's a front-end bug for MSVC targets which has been (wrongly) worked around only since 2.078.0 itself AFAICT, i.e., all current host compilers on Windows will crash. If we're lucky, an older version (ltsmaster) doesn't feature that particular MSVC-specific check.

The bug is here: https://github.com/ldc-developers/ldc/blob/13d5cd657241b3db32507d4b08a121a15b436e5d/ddmd/func.d#L664-L674
`FuncDeclaration.parent` is a TemplateMixin (`dmd.transitivevisitor.ParseVisitMethods` mixed into `dmd.visitor.SemanticTimeTransitiveVisitor`) while the code assumes it's a valid ClassDeclaration (and since 2.078.0 wrongly aborts the check if it isn't).

rainers:
The crash happens since dmd 2.071, see https://issues.dlang.org/show_bug.cgi?id=18093.

If my tests were correct, the workaround to build with pre dmd 2.078 versions could be to add extern (C++)  to the overloads in the template mixins.

kinke:
Ah thanks, so it's known already.

LDC 1.0.0 (2.070) manages to compile the new front-end, but the LDC then immediately fails when compiling Phobos (`visit()` function not hit, most likely due to another bug wrt. overload resolution or vtable...).
I don't see how an additional `extern(C++)` in the mixin could solve this; I tried slapping an `extern(C++):` on top of the > 1k lines mixin, to no avail.

rainers:
Adding extern(C++) worked in the reduced test case in the bug report (no idea why), but doesn't help in dmd.
Instead, it seems to be ok to make all non-override functions private: the check of the base class vtable is skipped in that case:
```
 src/dmd/transitivevisitor.d | 28 ++++++++++++++--------------
 1 file changed, 14 insertions(+), 14 deletions(-)

diff --git a/src/dmd/transitivevisitor.d b/src/dmd/transitivevisitor.d
index 19ee198..8f8fba4 100644
--- a/src/dmd/transitivevisitor.d
+++ b/src/dmd/transitivevisitor.d
@@ -56,7 +56,7 @@ package mixin template ParseVisitMethods(AST)
         }
     }
 
-    void visitVarDecl(AST.VarDeclaration v)
+    private void visitVarDecl(AST.VarDeclaration v)
     {
         //printf("Visiting VarDeclaration\n");
         if (v.type)
@@ -168,7 +168,7 @@ package mixin template ParseVisitMethods(AST)
             s.elsebody.accept(this);
     }
 
-    void visitArgs(AST.Expressions* expressions, AST.Expression basis = null)
+    private void visitArgs(AST.Expressions* expressions, AST.Expression basis = null)
     {
         if (!expressions || !expressions.dim)
             return;
@@ -298,7 +298,7 @@ package mixin template ParseVisitMethods(AST)
             imp.accept(this);
     }
 
-    void visit(AST.Catch c)
+    private void visit(AST.Catch c)
     {
         //printf("Visiting Catch\n");
         if (c.type)
@@ -310,7 +310,7 @@ package mixin template ParseVisitMethods(AST)
 //   Type Nodes
 //============================================================
 
-    void visitType(AST.Type t)
+    private void visitType(AST.Type t)
     {
         //printf("Visiting Type\n");
         if (!t)
@@ -324,7 +324,7 @@ package mixin template ParseVisitMethods(AST)
             t.accept(this);
     }
 
-    void visitFunctionType(AST.TypeFunction t, AST.TemplateDeclaration td)
+    private void visitFunctionType(AST.TypeFunction t, AST.TemplateDeclaration td)
     {
         if (t.next)
             visitType(t.next);
@@ -336,7 +336,7 @@ package mixin template ParseVisitMethods(AST)
         visitParameters(t.parameters);
     }
 
-    void visitParameters(AST.Parameters* parameters)
+    private void visitParameters(AST.Parameters* parameters)
     {
         if (parameters)
         {
@@ -405,7 +405,7 @@ package mixin template ParseVisitMethods(AST)
         visitFunctionType(cast(AST.TypeFunction)t.next, null);
     }
 
-    void visitTypeQualified(AST.TypeQualified t)
+    private void visitTypeQualified(AST.TypeQualified t)
     {
         //printf("Visiting TypeQualified\n");
         foreach (id; t.idents)
@@ -481,7 +481,7 @@ package mixin template ParseVisitMethods(AST)
 
 //      Declarations
 //=========================================================
-    void visitAttribDeclaration(AST.AttribDeclaration d)
+    private void visitAttribDeclaration(AST.AttribDeclaration d)
     {
         if (d.decl)
             foreach (de; *d.decl)
@@ -570,7 +570,7 @@ package mixin template ParseVisitMethods(AST)
         visitAttribDeclaration(cast(AST.AttribDeclaration)d);
     }
 
-    void visitFuncBody(AST.FuncDeclaration f)
+    private void visitFuncBody(AST.FuncDeclaration f)
     {
         //printf("Visiting funcBody\n");
         if (!f.fbody)
@@ -582,7 +582,7 @@ package mixin template ParseVisitMethods(AST)
         f.fbody.accept(this);
     }
 
-    void visitBaseClasses(AST.ClassDeclaration d)
+    private void visitBaseClasses(AST.ClassDeclaration d)
     {
         //printf("Visiting ClassDeclaration\n");
         if (!d || !d.baseclasses.dim)
@@ -591,7 +591,7 @@ package mixin template ParseVisitMethods(AST)
             visitType(b.type);
     }
 
-    bool visitEponymousMember(AST.TemplateDeclaration d)
+    private bool visitEponymousMember(AST.TemplateDeclaration d)
     {
         //printf("Visiting EponymousMember\n");
         if (!d.members || d.members.dim != 1)
@@ -647,7 +647,7 @@ package mixin template ParseVisitMethods(AST)
         return false;
     }
 
-    void visitTemplateParameters(AST.TemplateParameters* parameters)
+    private void visitTemplateParameters(AST.TemplateParameters* parameters)
     {
         if (!parameters || !parameters.dim)
             return;
@@ -669,7 +669,7 @@ package mixin template ParseVisitMethods(AST)
             s.accept(this);
     }
 
-    void visitObject(RootObject oarg)
+    private void visitObject(RootObject oarg)
     {
         if (auto t = AST.isType(oarg))
         {
@@ -687,7 +687,7 @@ package mixin template ParseVisitMethods(AST)
         }
     }
 
-    void visitTiargs(AST.TemplateInstance ti)
+    private void visitTiargs(AST.TemplateInstance ti)
     {
         //printf("Visiting tiargs\n");
         if (!ti.tiargs)
```

kinke:
Thanks, I'll give that a shot too. I'm currently stumbling from one problem to the next; DMD 2.078.0 doesn't link out of the box in an MSVC 2017 environment, I thought that was one of the news? `Error: can't run 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\\bin\link.exe', check PATH` (cmd [here](https://ci.appveyor.com/project/kinke/ldc/build/1.0.4869/job/gtatmindwwcns34h#L338)).

rainers:
> I'm currently stumbling from one problem to the next; DMD 2.078.0 doesn't link out of the box in an MSVC 2017 environment, I thought that was one of the news?

Unfortunately a corresponding installer PR didn't make it into the release, so you have to nuke most settings in sc.ini. For auto detection the Environment64 section should look this:
```
[Environment64]
LIB=%@P%\..\lib64
DFLAGS=%DFLAGS% -L/OPT:NOICF
```

rainers:
The LINKCMD settings is propably the only harmful setting in the zip file version. You could filter that out with `sed -i /LINKCMD/d sc.ini`.

kinke:
Thanks; tweaking the ini locally worked, but I forgot about the cmdline parsing issue (`dmd ... "-LC:\Program Files (x86)\Microsoft Visual Studio\2017\Community\DIA SDK\lib\amd64\diaguids.lib"` => `LINK : fatal error LNK1181: cannot open input file 'C:\Program.obj'`) rendering DMD unusable for compiling LDC with an LLVM incl. LLD...

Making the previously new virtual functions private works indeed, i.e., gets rid of the segfault, but the `visit()` issue persists (so that wasn't 2.070-specific - making sense, as ltsmaster works as host compiler for Travis). Maybe an MSVC-specific divergence between D and C++ headers...

kinke:
The issue seems to boil down to vtable layout. Excerpt of stack trace with debug build of LDC:
```
ldc2.exe!Visitor::visit(Statement * __formal) Line 306	C++	Symbols loaded.
ldc2.exe!Visitor::visit(ForStatement * s) Line 406	C++	Symbols loaded.
ldc2.exe!dmd.dimport.Import.accept() Line 311	D	Symbols loaded.
```

`Import.accept(Visitor v)` doesn't forward to `Visitor.visit(Import)`, but to `Visitor.visit(ForStatement)`, both miles apart in `dmd/visitor.h`.

kinke:
Oh this appears to be fixed by fixing `visitor.h` and letting class `Visitor` derive from a `ParseTimeVisitor` base class, as in D code. Visual C++ probably uses a reverse layout...

kinke:
Green on Linux. :)

kinke:
Windows now green too; most likely just a fundamental OSX-specific issue wrt. shared druntime/Phobos remaining. No touchups for druntime/Phobos necessary this time! 👍

kinke:
Can anyone on OSX check the issue with shared libs?

JohanEngelen:
I have had a number of things in the pipeline for a long time so I want to spend my time on that first in the next two weeks or so, sorry.

joakim-noah:
Tried this pull out on Android/ARM and ran the stdlib tests, have 6 modules that segfault, all related to optimization codegen as they only fail at `-O1` or higher.  All but one are related to regex, as they segfault in the Thompson regex engine, only `std.net.curl` ends up in libcrypto unrelated to regex.

First time I've had issues on Android/ARM since late last summer, #2024, will investigate further.

kinke:
@klickverbot: Can you be bothered? ;) - By the looks of it, each shared unittest fails, so possibly even a hello world due to some fundamental issue during druntime initialization. So building this branch, a hello world with `-link-defaultlib-shared -link-defaultlib-debug` and running it with a debugger might be enough to get an idea of what's happening. I had a second look at the `rt.sections_elf_shared` diff, but didn't find a likely cause. As there's no output in the CI logs, I can't do much myself.

kinke:
[This somehow doesn't suffer from the new dmd-testsuite timeouts for CircleCI Linux and the master branch. What seems to be hanging there is 32-bit `runnable/pi.d`, but that only started about a week ago or so. So merging this soon would fix the Circle issue, which is required for automated builds.]

joakim-noah:
@kinke, [do you have shell access rights for the Travis Mac CI](https://docs.travis-ci.com/user/running-build-in-debug-mode/)?  That could be a way for you to get into a Mac.

kinke:
I was a bit reluctant at first but got SSH set up with Circle, works like a charm. So the OSX shared lib issue should be fixed.

2.078.1 came with a new test, `runnable/test13742.sh`. It tests emission of coverage files; 2 modules in a static lib, which is then linked with a third module, all compiled with `-cov`. We don't emit the coverage file for the first module; it only contains a template. In `emitCoverageLinecountInc()`, there's this comment:

```C++
// Only emit coverage increment for locations in the source of the current module
// (for example, 'inlined' methods from other source files should be skipped).
if (!global.params.cov || !loc.linnum || !loc.filename || !m->d_cover_data ||
    strcmp(m->srcfile->name->toChars(), loc.filename) != 0) {
  return;
}
```

So this means that a template which isn't emitted in its declaring module won't have any coverage, although it may be instantiated and emitted in other modules, and seems like the obvious reason for the test failure. DMD apparently now decorates the coverage data symbol names with the module ID and made them global, so we should probably do the same.

kinke:
@joakim-noah: Any insights? I'm about to merge this to restore Circle builds.

joakim-noah:
Been looking into ARM, believe it's a Phobos regression related to some std.regex refactoring Dmitry did recently. I will submit a pull once I track it down, likely upstream, please go ahead and merge this pull whenever you're ready.

joakim-noah:
Spent several hours in a debugger trying to track down the ARM regression in `std.regex`, [one of the recurring segfaults happens here](https://github.com/dlang/phobos/blob/41c229647d994917f31dd1e06abedde5a829ae87/std/regex/internal/thompson.d#L354), because `t` is misaligned by the time it gets there.  That suggests earlier memory corruption, which doesn't happen without any llvm optimizations, ie only at `-O1` or higher.  Also, if I revert Dmitry's `std.regex` redesign, dlang/phobos#5722, the tests start working again at any optimization level, as that's just where `std.regex` was with ldc 1.7 anyway.

@DmitryOlshansky, the main big change in `std.regex.internal.thompson` is that `ThompsonMatcher` goes from a struct to a class, not sure if that affects the code generation somehow.  Before I spend several more hours tracking down the source of the memory corruption, I thought I'd run it by you and Martin to see if you guys have an inkling of where that pull might be throwing off ARM.  Let me know if you have an idea.

JohanEngelen:
@joakim-noah Do you have a dustmite session running in the meanwhile to reduce the problem case?

kinke:
Well looking at the PR is much more pleasant, and I strongly suspect the manual memory management via `GenericFactory` to be the culprit:

```D
   override Matcher!Char dup(Matcher!Char engine, in Char[] input) const @trusted
    {
        immutable size = EngineType!Char.initialMemory(engine.pattern) + classSize;
        auto memory = enforce(malloc(size), "malloc failed")[0 .. size];
        scope(failure) free(memory.ptr);
        auto copy = construct(engine.pattern, input, memory);
        GC.addRange(memory.ptr, classSize);
        engine.dupTo(copy, memory[classSize .. size]); // <-- memory.ptr + classSize may likely be unaligned
        assert(copy.refCount == 1);
        return copy;
    }
```

I didn't dig deep, but it looks like two instances are stored consecutively in memory, exactly without padding bytes, so that the 2nd payload is likely to be misaligned.

joakim-noah:
@JohanEngelen, never tried dustmite, will it narrow down Phobos regressions?

@kinke, anyway, dustmite doesn't matter now, because you got it.  Adding that padding fixes the problems across all six modules, which were all related to this `std.regex` change.  I focused too closely on the Thompson engine where the segfaults were all happening, and not on the rest of that giant patch.  However, even if I had looked at the whole pull as you did, I wouldn't have narrowed it down to that memory allocation issue just by looking at the patch.

Go ahead and commit your `std.regex` fix and we'll be back to only a single stdlib test failing on ARM again: the EH inlining codegen bug remaining at #2024.

JohanEngelen:
> never tried dustmite, will it narrow down Phobos regressions?

Sure! https://github.com/CyberShadow/DustMite/wiki#minimizing-the-standard-library

kinke:
Thx for verifying, Joakim. I pushed the hotfix directly to master; the upstream PR is https://github.com/dlang/phobos/pull/6185.

joakim-noah:
Maybe time to release a 1.8 beta now?

PR #2490, opened 2018-01-11
LLVM7

thewilsonator:
See also https://github.com/ldc-developers/druntime/pull/112


kinke:
Can't wait to have all the profile-rt stuff removed from LDC. ;)

thewilsonator:
Me too but I can't think of any other way to keep it in synch and compatible with the various LLVM versions we support.

kinke:
See https://github.com/ldc-developers/ldc/pull/2464. All we really need is another compiler-rt lib matching the LLVM version LDC was built with.

runtime/profile-rt/profile-rt-70/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
+
```
JohanEngelen: this file isn't needed, please remove

JohanEngelen: (and check whether there is anything import new that we should have in our `DefineBuildProfileRT.cmake`)

klickverbot: Do we have a wiki page or something describing am the requisite steps for upgrading LLVM?

JohanEngelen: copy-paste, that's it.

JohanEngelen: (the cmake changes are rarely needed, every now and then I check them)

thewilsonator: >this file isn't needed, please remove

Done.

JohanEngelen:
@thewilsonator Merge when you're happy with it.

PR #2491, opened 2018-01-11
Fix up command-line help for FlagParser options

kinke:
Old:

```
-enable-asserts         - (*) Enable assertions
  =enable-asserts       -   (*) Enable assertions
```

New:

```
-enable-asserts         - (*) Enable assertions
```


JohanEngelen:
nice, lgtm

(btw, the LLVM_OVERRIDE we should be able to replace with `override`. I guess it's a C++03/C++11 transition macro)

PR #2492, opened 2018-01-11
@dynamicCompileConst optimization test

Hardcode84:


JohanEngelen:
More tests is always good :)

(If you don't mind, please flip the comparisons, `4 == foo()` is just awkward to read)

Hardcode84:
Fixed

Issue #2450, opened 2017-12-11
Strict left-to-right evaluation of function arguments


JohanEngelen:
Testcase:
```d
void test240a(int a, int b, int c)
{
    assert(a == 0);
    assert(b == 0);
    assert(c == 1);
}

void main()
{
    int a = 0;
    test240a(a, a++, a++);
    assert(a == 2);
}
```

See https://bugzilla.gdcproject.org/show_bug.cgi?id=240 .
The test should be added upstream, because apparently there isn't any testing of that.
Spec: https://dlang.org/spec/expression.html#order-of-evaluation

kinke:
Oh man, this probably isn't going to be a trivial fix. Our 'evaluation' is `toElem()`, which returns a `DLValue` for lvalues like `a` as first arg. So if the function doesn't take the arg by ref, we only load from the `DLValue` in a second pass right before generating the IR call, i.e., too late, after all args' `toElem()` side effects. So for this to work, the load needs to happen right after `toElem()` for primitive types, and for structs and static arrays, we sadly need a new alloca + memcpy. I guess detecting the cases where we can skip that copy isn't trivial (well, the last arg should be safe), so more work for LLVM to optimize.

joakim-noah:
The spec notes that dmd doesn't even follow this, [there have been forum threads discussing this inconsistency going back a decade](http://forum.dlang.org/thread/f0pvsn$1pju$1@digitalmars.com).  Not saying it isn't worth fixing, but I actually like that it discourages much confusing code and it should be fixed upstream for it to be taken seriously.

klickverbot:
> Thinking about it, IrFuncTy::putParam() should probably be called right after toElem()enting an arg. It applies the TargetABI rewrite if there is one (and some of those already perform a copy) and otherwise performs the load (but doesn't copy not-rewritten structs & static arrays passed by value yet).

That sounds about right – IIRC this was very much the intention during the last overhaul, but it seems like we forgot about the lvalue -> rvalue conversion.

> Another issue here is that all structs and static arrays are represented as DLValues, even if the argument expression is a D rvalue, and we surely don't want an extra copy for those.

I'm not sure I understand your point here – we don't want an _extra_ copy, but surely there is already one copy made per argument for by-value semantics?

kinke:
> > Another issue here is that all structs and static arrays are represented as DLValues, even if the argument expression is a D rvalue, and we surely don't want an extra copy for those.
[...]

PR #2502, opened 2018-01-13
Fix issue #2450 (function call arguments evaluation/load order)

kinke:


kinke:
DMD 2.078.0 passes the test, even 2.068.2 does. LDC 1.0 already fails.

kinke:
Any concerns? I consider this a really serious bug, surely one of the worst lately...

JohanEngelen:
Yeah indeed.
Merge it when you're satisfied with it.

Issue #2458, opened 2017-12-19
llvm_expect doesn't work with CTFE


n8sh:
https://run.dlang.io/gist/eeefb62c315e0f25ed25ab09749e93db?compiler=ldc
```d
import std.stdio;
import ldc.intrinsics: llvm_expect;

uint notzero()(uint x)
{
    if (llvm_expect(x == 0, false))
        return 1;
    return x;
}
void main(string[] args)
{
    immutable x = notzero(1);
    enum y = notzero(1);
    writefln("x = %d, y = %d", x, y);
}
```
fails with:
>onlineapp.d(6): Error: llvm_expect cannot be interpreted at compile time, because it has no available source code
>onlineapp.d(13):        called from here: notzero(1u)

kinke:
Amazing what people expect from a compiler. :) - In case you need a workaround for the time being, there's `if (__ctfe)` as used [here](https://github.com/dlang/phobos/blob/master/std/math.d#L2107-L2110).

klickverbot:
Hm – maybe there should just be a method body that returns the first argument, which would be ignored at runtime?

kinke:
Fixed by #2506.

PR #2506, opened 2018-01-14
Support llvm_expect() intrinsic during CTFE

kinke:
Fixes issue #2458.

### DIFF
--- a/ddmd/builtin.d
+++ b/ddmd/builtin.d
@@ -456,6 +456,16 @@ extern (C++) Expression eval_ctpop(Loc loc, FuncDeclaration fd, Expressions *arg
     return new IntegerExp(loc, cnt, type);
 }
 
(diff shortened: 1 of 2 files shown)


ddmd/builtin.d
```diff
@@ -456,6 +456,16 @@ extern (C++) Expression eval_ctpop(Loc loc, FuncDeclaration fd, Expressions *arg
     return new IntegerExp(loc, cnt, type);
 }
 
+extern (C++) Expression eval_expect(Loc loc, FuncDeclaration fd, Expressions *arguments)
```
JohanEngelen: out of interest, any reason this is `extern(C++)` besides uniformity with surrounding code?

kinke: Yep, there's a list of pointers to these functions, see https://github.com/ldc-developers/ldc/blob/master/ddmd/builtin.d#L44.

PR #2508, opened 2018-01-15
DMD issue 17899

thewilsonator:
This is an LDC for the in progress dmd fix https://github.com/dlang/dmd/pull/7704

But because I don't know dmd's code base very well I'm fixing it here first because I want it to fix https://github.com/libmir/dcompute/issues/32 and ldc has better diagnostics.

There is something very strange indeed going on with the codegen:

```
10B4C4500 10B4C43F0
40 10B4C4500
bar2
2 3
core.exception.AssertError@./runnable/test17899.d(28): Assertion failure
```

### DIFF
--- a/gen/toconstelem.cpp
+++ b/gen/toconstelem.cpp
@@ -11,6 +11,7 @@
 #include "gen/binops.h"
 #include "gen/classes.h"
 #include "gen/complex.h"
+#include "gen/functions.h"
 #include "gen/irstate.h"
 #include "gen/llvm.h"
 #include "gen/llvmhelpers.h"
@@ -458,23 +459,21 @@ public:
       fd->vthis = nullptr;
     }
 
-    if (fd->tok != TOKfunction) {
-      assert(fd->tok == TOKdelegate || fd->tok == TOKreserved);
-      e->error("non-constant nested delegate literal expression `%s`",
-               e->toChars());
-      if (!global.gag) {
-        fatal();
-      }
-      result = llvm::UndefValue::get(DtoType(e->type));
-    } else {
-      // We need to actually codegen the function here, as literals are not
-      // added
-      // to the module member list.
-      Declaration_codegen(fd, p);
-      assert(DtoCallee(fd));
+    // We need to actually codegen the function here, as literals are not
+    // added
+    // to the module member list.
+    Declaration_codegen(fd, p);
 
-      result = DtoCallee(fd);
+    result = DtoCallee(fd);
+    if (fd->tok == TOKdelegate)
+    {
+      // If the literal was a delegate construct an initialiser with a null
+      // context pointer.
+      auto *i8null = LLConstant::getNullValue(
+                        LLType::getInt8PtrTy(gIR->context()));
+      result = LLConstantStruct::getAnon(gIR->context(), { i8null, result });
     }
+      assert(result);
   }
 
   //////////////////////////////////////////////////////////////////////////////
--- a/gen/toconstelem.cpp
+++ b/gen/toconstelem.cpp
@@ -475,6 +475,9 @@ public:
       // FIXME: this won't work for a module scope AssocArray with delegates
       // as keys or values.
       if (global.gag) {
+        // Issue an error so that ToElemVisitor::visit(AssocArrayLiteralExp *e)
+        // uses runtime initialisation of the AA.
+        e->error("dummy error"); 
         result = llvm::UndefValue::get(DtoType(e->type));
       }
       else
--- a/gen/toconstelem.cpp
+++ b/gen/toconstelem.cpp
@@ -467,13 +467,26 @@ public:
     result = DtoCallee(fd);
     if (fd->tok == TOKdelegate)
     {
-      // If the literal was a delegate construct an initialiser with a null
-      // context pointer.
-      auto *i8null = LLConstant::getNullValue(
-                        LLType::getInt8PtrTy(gIR->context()));
-      result = LLConstantStruct::getAnon(gIR->context(), { i8null, result });
+      // AssocArrayLiteralExp::toElem determines whether it can allocate
+      // the needed arrays statically by just invoking toConstElem on its
+      // key/value expressions with error gagging. see
+      // ToConstElemVisitor::visit(Expression *e)
+
+      // FIXME: this won't work for a module scope AssocArray with delegates
+      // as keys or values.
+      if (global.gag) {
+        result = llvm::UndefValue::get(DtoType(e->type));
+      }
+      else
+      {
+        // If the literal was a delegate construct an initialiser with a null
+        // context pointer for delegates declared at module scope.
+        auto *i8null = LLConstant::getNullValue(
+                          LLType::getInt8PtrTy(gIR->context()));
+        result = LLConstantStruct::getAnon(gIR->context(), { i8null, result });
+      }
     }
-      assert(result);
+    assert(result);
   }
 
   //////////////////////////////////////////////////////////////////////////////


thewilsonator:
Stupid question: why is there still a conflict?

JohanEngelen:
don't know, but you should rebase instead of merging-in master

thewilsonator:
Hmm, that didn't appear to do what I had hoped it would.

thewilsonator:
I seem to have two sets of commits.

thewilsonator:
blasted it and started again. Hope that fixes it.

thewilsonator:
`runnable/funclit.d` fails on:
```
void test7650()
{
    //...
    int n = 10;
    int[int delegate(int)] aa3 = [x=>n+x:1, x=>n+x*2:2];
    foreach (k, v; aa3) {
        if (v == 1) assert(k(10) == 20); //Here
        if (v == 2) assert(k(10) == 30); //Here
    }
    printf("%d\n",__LINE__);
    int delegate(int)[int] aa4 = [1:x=>n+x, 2:x=>n+x*2];
    assert(aa4[1](10) == 20); //Here
    assert(aa4[2](10) == 30); // Here
}
```

@JohanEngelen Any insight as to why?

thewilsonator:
Looking at the callers of `toConstElem` it looks like it can only come from:
[here](https://github.com/ldc-developers/ldc/blob/438ff5873e1aa768333a41dcd6d682c15ad088e7/gen/toir.cpp#L315),[here](https://github.com/ldc-developers/ldc/blob/438ff5873e1aa768333a41dcd6d682c15ad088e7/gen/arrays.cpp#L575), or [here](https://github.com/ldc-developers/ldc/blob/907dae171a43ffbbfa295b858437414c136f611a/gen/llvmhelpers.cpp#L1175)

thewilsonator:
Hmm, the degenerate case of `toConstElem` has an interesting comment:
```
void visit(Expression *e) override {
    e->error("expression `%s` is not a constant", e->toChars());
    if (!global.gag) {
      fatal();
    }

    // Do not return null here, as AssocArrayLiteralExp::toElem determines
    // whether it can allocate the needed arrays statically by just invoking
    // toConstElem on its key/value expressions, and handling the null value
    // consequently would require error-prone adaptions in all other code.
    result = llvm::UndefValue::get(DtoType(e->type));
}
```

kinke:
> Obviously the context pointer is null.

It obviously isn't, as all of these delegates capture `n` from the outer function.

thewilsonator:
It is, I checked the generated code. It shouldn't be, but thats because `AssocArrayLiteralExp` does its codegen conditionally on `result == undef`, see my above comment.

kinke:
So the difficulty should be establishing whether the lambda requires the context. Only if it doesn't, your change to allow `toConstElem()`enting a delegate is valid, otherwise it still needs to produce an error.

thewilsonator:
Yeah that fix doesn't work properly, I'm working on something less flakey.

thewilsonator:
Ah I get what you mean now.

```
class ToElemVisitor : public Visitor {
  //...
  void visit(AssocArrayLiteralExp *e) override {
  //...
        unsigned errors = global.startGagging();
        LLConstant *ekeyConst = toConstElem(ekey, p);
        LLConstant *evalConst = toConstElem(eval, p);
        if (global.endGagging(errors)) {
          goto LruntimeInit;
        }
       //...
    }
}
```

kinke:
Oh I hoped it'd be clear that this is definitely unmergeable, it's a huge hack.

thewilsonator:
Oh, sorry. I agree that it is a hack, but then so is the entire way that `ToElemVisitor::visit(AssocArrayLiteralExp *e)` functions. I was just going by https://github.com/ldc-developers/ldc/pull/2571#issuecomment-364776245 that #2571 was a good way to fix this.

kinke:
There are multiple issues with this. Firstly, the lambda is now always emitted, even if the code is triggered in a tentative context. And secondly, you now emit a constant null context without checking if the context is really unused, assuming that it is unused if errors aren't gagged.

So I'd expect the following assertion to fail now:
```D
void main() {
    int n = 123;
    auto dg = (int x) => n + x; // clearly accessing its context
    static assert(dg.ptr !is null);
}
```

thewilsonator:
```
test.d(4): Error: variable dg cannot be read at compile time
test.d(4):        while evaluating: static assert(dg.ptr !is null)
```

thewilsonator:
Also:
```
import std.stdio;
void main() {
    int n = 123;
    auto dg = (int x) => n + x; // clearly accessing its context
    writefln("%s, %s",dg.ptr, dg.funcptr);
}
```
e.g. `1061A6000, 105C26D00`

thewilsonator:
The only reason why the AA stuff originally failed was precisely because it relied on the code emitting an error while trying to force evaluation of the dg at compile time. Nothing else relies on that (old) codegen property of delegates.

thewilsonator:
I'd rather improve the way that AA literals are chosen to be statically initialised, e.g. to check if the const elems are `undef` instead. #2572

`fd->vthis != nullptr` is always false for delegates, isn't it?

(do we a have a `isUdef`?)

kinke:
That doesn't look like an improvement to me, sorry.

`FuncDeclaration.isNested()` would seem suitable to determine if there's a context; unfortunately, `FuncLiteralDeclaration` overrides it so that a delegate literal is classified as nested as long as `!isThis()` (can that ever be false?); i.e., it doesn't check whether there's really any context, but rather indicates whether the function takes a nested context argument. I guess it's a similar story for `vthis` but haven't looked it up.

So if there's no appropriate flag in the frontend yet and the other compilers are supposed to allow context-less static delegates too, then something like setting a FuncDeclaration flag when declaring an outer variable (or when calling another nested function, passing through the nested context) in semantic will probably be needed. [I anticipated this difficulty, that's why I suggested going to the IR layer and checking the IR parameter uses, which would be trivial if the function is always emitted beforehand as in this patch.]
The flag can then be cleanly handled in `toConstElem()`.

Until then, please revert this, 1.8.0-beta1 would be pretty ready from my side (+ LLVM plugins support and maybe dynamic-compile fixes).

kinke:
[If you only care about module-level initializers, you could also restrict the support to delegate literals with a module parent.]

thewilsonator:
I would really like to have _a_ solution for this, because I can't merge libmir/dcompute#39 then to fix libmir/dcompute#32 which is a rather severe bug caused by  https://issues.dlang.org/show_bug.cgi?id=18412 Particularly because having DCompute usable with a released version of LDC.

>FuncDeclaration.isNested() would seem suitable to determine if there's a context; unfortunately, FuncLiteralDeclaration overrides it so that a delegate literal is classified as nested as long as !isThis() (can that ever be false?); i.e., it doesn't check whether there's really any context, but rather indicates whether the function takes a nested context argument. I guess it's a similar story for vthis but haven't looked it up.

Indeed.

> So if there's no appropriate flag in the frontend yet and the other compilers are supposed to allow context-less static delegates

There was no front end flag added in the dmd PR, and I fail to see why one is needed. The _only_ reason this same approach (is the const elem a delegate? if so set context to null) is a problem because LDC relies on `toConstElem` failing for AA initialisation. I'll alter #2572 to runtime init any AA that has delegates as keys of values.

kinke:
> (is the const elem a delegate? if so set context to null)

That's the problem. The question isn't if it's a delegate that makes it to toConstElem(), but is 'does the delegate access its context? if not, then toConstElem() is valid, and emit a static one'. As David said, having an additional check here is nice to catch AST quirks or bugs in our code, and reduces the inter-call dependencies. And there'd be no need to account for AA literals - if all delegates are contextless, a static initialization of the AA is fine and consistent with a module-level delegate initializer.

I'm not going to beg you to revert it. Once 2 LDC devs ask you to, it should be clear that something needs to be rectified. If you're looking for a more sane quick'n dirty solution for your DCompute problem, just check the delegate literal's parent as I suggested earlier, I can live with that.

thewilsonator:
> Once 2 LDC devs ask you to, it should be clear that something needs to be rectified.

Hence why I wanted to improve it.

> If you're looking for a more sane quick'n dirty solution for your DCompute problem, just check the delegate literal's parent as I suggested earlier, I can live with that.

You mean the `FuncExp`s `FuncLiteralDeclaration` fd's `.parent` (`e->fd->parent`) which will be a module iff the expression is a module scope? Will do.

kinke:
You'll probably have to do a `fd.toParent2()` to skip over template (mixin) parents.

> I'll alter #2572 to runtime init any AA that has delegates as keys of values.

Just for clarification, the runtime init isn't specific to delegates:
```D
void foo(int arg) {
    int[int] staticallyInitializedLiteral = [ 123: 666 ]; // toConstElem() succeeding for all keys and values
    int[int] dynamicallyInitializedLiteral = [ 123: arg ];
}
```

thewilsonator:
> You'll probably have to do a fd.toParent2() to skip over template (mixin) parents.

Done.

> Just for clarification, the runtime init isn't specific to delegates:

Noted. Does this still apply with the current #2572 ? i.e. do I still need to do that given the change in error emission.

Issue #2515, opened 2018-01-17
Missed optimization for scope allocated classes


rracariu:
The given program compiled with ldc 1.7.0 and the switches `-O3 -release -betterC`

```d
    class Foo
    {
        bool status;

        this(bool status)
        {
            this.status	= status;
        }
    }

    void main()
    {
        auto f = new Foo(true);

        import core.stdc.stdio;

        printf("%d", f.status);
    }
```

produces this optimized code:

```asm
_D7example3Foo6__ctorMFbZCQyQs:
  mov byte ptr [rdi + 16], sil
  mov rax, rdi
  ret

_Dmain:
  push rax
  lea rdi, [rip + .L.str.1]
  mov esi, 1
  xor eax, eax
  call printf@PLT
  xor eax, eax
  pop rcx
  ret
```

Change  `auto f = new Foo(true);` to `scope f = new Foo(true);` and the following less optimal asm is produced:
```asm
_D7example3Foo6__ctorMFbZCQyQs:
  mov byte ptr [rdi + 16], sil
  mov rax, rdi
  ret

_Dmain:
  sub rsp, 24
  mov rax, qword ptr [rip + example.Foo.__vtbl@GOTPCREL]
  mov qword ptr [rsp], rax
  mov qword ptr [rsp + 8], 0
  mov byte ptr [rsp + 16], 1
  lea rdi, [rip + .L.str.1]
  mov esi, 1
  xor eax, eax
  call printf@PLT
  mov rdi, rsp
  call _d_callfinalizer@PLT
  xor eax, eax
  add rsp, 24
  ret
```

kinke:
Wrt. finalization, while the front-end could check whether there are any destructors for the type, it also involves deleting the object's monitor, whose existence probably can't be determined easily.

rracariu:
OK, this is surprising, because one would expect that `scope` will be the optimized variant.

I still don't understand why `scope` doesn't follow through the same code gen path as gc2stack?

I'm not familiar with the fronted code but I assume that you can check for the simplest case if the class is only inheriting from Object then verify if it has a dtor, if not, then skip the finalizer call altogether.

rracariu:
I saw your commits. I'll let you do your magic 👍

klickverbot:
If our gc2stack implementation is still valid (it's been around for some time), there is indeed no reason why we shouldn't be able to teach it to recognize `_d_callfinalizer` as well.

klickverbot:
[...]

PR #2516, opened 2018-01-17
Don't finalize scope objects without dtors and monitor

kinke:
The dtors can be checked at compile-time; insert a runtime check for the monitor before finalizing the stack-allocated class object via druntime call.

See issue #2515.

### DIFF
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -21,6 +21,7 @@
 #include "gen/llvmhelpers.h"
 #include "gen/logger.h"
 #include "gen/nested.h"
+#include "gen/optimizer.h"
 #include "gen/rttibuilder.h"
 #include "gen/runtime.h"
 #include "gen/structs.h"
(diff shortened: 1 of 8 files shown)


kinke:
Damn, further testing reveals that one cannot rely on the variable type being the real dynamic type, so the dtors cannot be checked at compile-time.

kinke:
That'd be a bit costly, we'd need to look up the TypeInfo (another indirection) and check its hasDtors flag. Doing that before _d_callfinalizer(), which performs these checks again, may lead to a slowdown instead of an improvement in non-trivial cases, i.e., where the TypeInfo isn't available in the same LL module. Checking the monitor alone seems worth it, but I'd rather not do the dtors check at runtime too.
IIRC, scope objects need to be initialized with their declaration; if they cannot be reassigned (don't know yet), we could exploit that and store the dynamic type in the front-end to still perform the dtors check at compile-time.

kinke:
Ah that would also work.

```D
        scope f = new Foo(true);
        import core.stdc.stdio;
        printf("%d", f.status);
        f = new Foo(false);
```

=> `Error: cannot rebind scope variables`

So saving the dynamic type as part of the declaration should be safe and avoid the vptr check.

kinke:
Should be working now.

tests/codegen/gh2515.d
```diff
@@ -0,0 +1,48 @@
...
+{
+    int val = 123;
+    void foo() { val *= 3; }
+    void bar() { synchronized(this) val *= 2; }
+}
+
+class Foo : Base
+{
+    ~this() {}
+}
```
JohanEngelen: I feel more safe if you'd also include a test with a class that does not have an explicit dtor, but one of its members does have a dtor (or one of _its_ members...).
E.g.
```d
class FooFoo {
   Foo f;
}
class FooFooFoo {  // <-- test with this class
   FooFoo f;
}

```

gen/classes.cpp
```diff
@@ -193,6 +193,42 @@ void DtoFinalizeClass(Loc &loc, LLValue *inst) {
...
+      break;
+    }
+  }
+
+  if (hasDtor) {
+    DtoFinalizeClass(loc, inst);
+    return;
+  }
+
+  // no dtors => only finalize (via druntime call) if monitor is set,
```
JohanEngelen: perhaps only do this for `-O1` builds? (or `-O2` or...)

kinke: The load-compare-and-branch combo is tiny so I thought it'd be worth including it for unoptimized builds as well, but I don't have a strong opinion on this.

tests/codegen/gh2515.d
kinke: [Class references as members aren't destructed, a struct with dtor would be needed.]

JohanEngelen: (yep, sorry, too much in C++ land atm)

kinke: I added a test for a class with implicit dtor.

gen/classes.cpp
JohanEngelen: For a non-optimized build, perhaps one may expect there to be a call to class finalization.
However, checking clang's behavior, ctor and dtor are not called (I suppose because they are not generated) in non-optimized build.
I tend towards not enabling this for `-O0`, for cleanliness and to keep the IR/asm as readable as possible. 
(there's also `-Os` and `-Oz` btw... argh :))

JohanEngelen: Looks this code is not reached for `extern(C++)` classes. Add a test to make sure?

(is there a bug with `scope` `extern(C++)` classes?  https://godbolt.org/g/iNbePi )

kinke: Yeah the implicit delete() expression isn't applied to C++ classes, there's a related comment in the front-end.

kinke: > to keep the IR/asm as readable as possible

Yep that's right, so now skipped for `-O0`. I'm fine with having this in `-Os` and `-Oz` as it really is tiny, should really pay off in the vast majority of cases (no monitor) by eliminating the call & typeinfo chain lookup etc., and its likely foldability.

Test for C++ class added, which actually tests the assertion in `DtoFinalizeScopeClass()`.

kinke:
Good to go?

JohanEngelen:
> Yeah the implicit delete() expression isn't emitted for C++ classes, there's a related comment in the front-end.

Wow, that's very surprising.
This LGTM.

In more common cases, where LLVM cannot reason about the monitor any more because the object was passed by reference (as it always is) to an opaque function (i.e. any cross-module calls without LTO), also in those common cases there is still a perf boost because the monitor is almost never used.

PR #2524, opened 2018-01-20
DCompute CUDA: Use the datalayout from the targetmachine created by LLVM

JohanEngelen:
Fixes CUDA dcompute compilation with LLVM 6.0.
Why do we have to set the datalayout to something explicit? This breaks every now and then when LLVM changes things, making our datalayout invalid. If there is no good reason for, then we should use LLVM's datalayout. It currently is a maintenance problem.

@thewilsonator The dcompute compilation is still a major hack. Let's work towards making it less so.
For example, LDC's code is referencing gDataLayout throughout the code base, which is invalid when generating dcompute code. We should remove the `gDataLayout` global var, and just use `gTargetMachine->getDataLayout` (or `gIR->module.getDataLayout()` when possible).


thewilsonator:
LGTM.

The SPIR-V one need to be set because there is no Target machine yet. The CUDA one should be generated from the TargetMachine.

Issue #2520, opened 2018-01-20
Bounds checking fails to compile in 16-bit mode (MSP430)


luismarques:
The following code, when compiled with bounds checking on, fails to compile for `-mtriple=msp430 `:

```
void test()
{
    int[1] array;
    int k = 0;
    auto value = array[k];
}
```

```
Assertion failed: (getOperand(0)->getType() == getOperand(1)->getType() && "Both operands to ICmp instruction are not of the same type!"), function AssertOK, file llvm/include/llvm/IR/Instructions.h, line 1113.
0  ldc2                     0x0000000108cc4f8c llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 60
1  ldc2                     0x0000000108cc5559 PrintStackTraceSignalHandler(void*) + 25
2  ldc2                     0x0000000108cc0ee9 llvm::sys::RunSignalHandlers() + 425
3  ldc2                     0x0000000108cc58d2 SignalHandler(int) + 354
4  libsystem_platform.dylib 0x00007fff4fffcf5a _sigtramp + 26
5  libsystem_platform.dylib 0x00007fe07ed13948 _sigtramp + 785476104
6  libsystem_c.dylib        0x00007fff4fe27312 abort + 127
7  libsystem_c.dylib        0x00007fff4fdef368 basename_r + 0
8  ldc2                     0x000000010615ca13 llvm::ICmpInst::AssertOK() + 243
9  ldc2                     0x000000010615c90e llvm::ICmpInst::ICmpInst(llvm::CmpInst::Predicate, llvm::Value*, llvm::Value*, llvm::Twine const&) + 142
10 ldc2                     0x000000010615c873 llvm::ICmpInst::ICmpInst(llvm::CmpInst::Predicate, llvm::Value*, llvm::Value*, llvm::Twine const&) + 51
11 ldc2                     0x000000010615352d llvm::IRBuilder<llvm::ConstantFolder, llvm::IRBuilderDefaultInserter>::CreateICmp(llvm::CmpInst::Predicate, llvm::Value*, llvm::Value*, llvm::Twine const&) + 221
12 ldc2                     0x00000001060845eb DtoIndexBoundsCheck(Loc&, DValue*, DValue*) + 347
13 ldc2                     0x00000001061f7d93 ToElemVisitor::visit(IndexExp*) + 1571
14 ldc2                     0x00000001061eda26 toElem(Expression*) + 70
[...]

PR #2530, opened 2018-01-21
Fix bounds checking for 16-bit targets (Issue #2520)

luismarques:
Fixes Issue #2520 

This undoes the automatic zero extension of the index value used in the array bounds checking (back to `size_t`, which will not be a 32-bit for 16-bit targets). That fixes an internal LDC error that occurs when compiling, with bounds checking on, code like this:

```
void test()
{
    int[1] array;
    ushort k = 0;
    auto value = array[k];
}
```

### DIFF
--- a/tests/compilable/bounds_16_bit.d
+++ b/tests/compilable/bounds_16_bit.d
@@ -1,0 +1,20 @@
+// REQUIRES: target_MSP430
+
+// RUN: %ldc -mtriple=msp430 -c -output-ll %s
+
+void test()
+{
+    int[1] array;
+
+    ushort i = 0;
+    auto value = array[i];
+
+    short j = 0;
+    value = array[j];
+
+    uint k = 0;
+    value = array[k];
+
+    int l = 0;
+    value = array[l];
+}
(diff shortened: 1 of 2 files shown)


JohanEngelen:
Needs a testcase.

luismarques:
Now with a test.

tests/compilable/bounds_16_bit.d
```diff
@@ -0,0 +1,20 @@
+// REQUIRES: target_MSP430
+
+// RUN: %ldc -mtriple=msp430 -c -output-ll %s
```
JohanEngelen: Nit: add an explicit `-of %t`, such that the output is created in the expected location.

luismarques: @JohanEngelen What's different about this test that requires `-of %t`, while the others on the same directory don't? Note that I added `-output-ll` not because I want to keep the IR output but because just `-c` doesn't compile (no integrated MSP430 assembler) and `-o-` doesn't trigger the bug.

JohanEngelen: Nothing is different, all those tests write their output in an "OK" dir. If you look at the tests dir structure in your build dir, you'll see ./tests/compilable/Output  which is empty, but tests/compilable has all the output files. Nevermind, it's ok.

kinke:
I think it makes sense to investigate where the zero-extension from i16 to i32 comes from, as this may just be a symptom of a more general issue. Studying the `-vv` output, starting from the back, should give insight.

Issue #2537, opened 2018-01-26
Broken module error due to invalid LLVM integer arithmetic


s-ludwig:
Running `dub test` on [`std_data_json`](https://github.com/s-ludwig/std_data_json) yields the following error:
```
Integer arithmetic operators only work with integral types!
  %6 = add i8* %4, %5, !dbg !5446
LLVM ERROR: Broken function found, compilation aborted!
```

Build logs: https://travis-ci.org/s-ludwig/std_data_json/builds/333671316

kinke:
Thx for the report.

Failing command line:
```
ldc2 -c -unittest -Istd_data_json-master\source
    -I...\dub\packages\taggedalgebraic-0.10.9\taggedalgebraic\source
    std_data_json-master\source\stdx\data\json\value.d
```

Examplary offending IR with pointer arithmetic:
```LL
define weak_odr i8* @_D15taggedalgebraic__T12performOpRawTS4stdx4data4json5value9JSONValue12PayloadUnionVEQDf6OpKindi1VAyaa1_2bTnTQcZQDmFNaNbNiNfKQsQuZQx(i8* %_param_1_arg, i8** dereferenceable(8) %value) #0 comdat {
  %_param_1 = alloca i8*, align 8                 ; [#uses = 2, size/byte = 8]
  store i8* %_param_1_arg, i8** %_param_1
  %1 = load i8*, i8** %_param_1                   ; [#uses = 1]
  %2 = load i8*, i8** %value                      ; [#uses = 1]
  %3 = add i8* %1, %2                             ; [#uses = 1]
  ret i8* %3
}
```

kinke:
The issue seems to boil down to
```D
void main()
{
    typeof(null) a, b;
    auto x = a + b;
}
```

apparently being valid code, but LDC trying to emit an addition of the two dummy pointers.

kinke:
Okay, DMD swallows this too %)
```D
void main()
{
    typeof(null) a;
    int[string] dict;
    auto y = dict + a;
}
```

kinke:
Fixed by #2540.

PR #2540, opened 2018-01-27
Account for AST binop quirks

kinke:
This fixes issue #2537.

### DIFF
--- a/tests/codegen/gh2537.d
+++ b/tests/codegen/gh2537.d
@@ -1,0 +1,25 @@
+// RUN: %ldc -run %s
+
+void main()
+{
+    int[string] aa = [ "one": 123 ];
+    typeof(null) nul;
+
+    auto sum = nul + nul;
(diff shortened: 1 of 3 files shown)


tests/codegen/gh2537.d
```diff
@@ -0,0 +1,25 @@
...
+    int[string] aa = [ "one": 123 ];
+    typeof(null) nul;
+
+    auto sum = nul + nul;
+    auto diff = nul - nul;
+
+    assert(aa + nul == aa);
+    assert(nul + aa == aa);
+    assert(aa - nul == aa);
+    assert(nul - aa == aa);
```
kinke: If compiled with DMD 2.078, the binary segfaults here, as DMD really emits `0 - ptr`.

So this should most likely all be fixed in the front-end.

kinke:
DMD issue: https://issues.dlang.org/show_bug.cgi?id=18317

PR #2544, opened 2018-01-28
Support compiler-rt libs on Windows and get rid of our profile-rt libs

kinke:


CMakeLists.txt
```diff
@@ -739,12 +739,7 @@ endif()
 # Locate ASan and other runtime libraries, and copy them to our lib folder
 # Location is LLVM_LIBRARY_DIRS/clang/<version>/lib/<OS>/ , for example LLVM_LIBRARY_DIRS/clang/4.0.0/lib/darwin/
 #
-if(APPLE)
-    set(LDC_INSTALL_LLVM_RUNTIME_LIBS_DEFAULT ON)
-else()
-    set(LDC_INSTALL_LLVM_RUNTIME_LIBS_DEFAULT OFF)
-endif()
-set(LDC_INSTALL_LLVM_RUNTIME_LIBS ${LDC_INSTALL_LLVM_RUNTIME_LIBS_DEFAULT} CACHE BOOL "Copy/install runtime libraries (ASan, libFuzzer) from LLVM/Clang into LDC lib dir when available.")
+set(LDC_INSTALL_LLVM_RUNTIME_LIBS ON CACHE BOOL "Copy/install runtime libraries (ASan, libFuzzer) from LLVM/Clang into LDC lib dir when available.")
```
JohanEngelen: The reason this was off per default on linux, is to make life a little easier for packagers. Let's keep that.

CMakeLists.txt
```diff
@@ -759,24 +754,41 @@ function(copy_compilerrt_lib llvm_lib_name ldc_lib_name fixup_dylib)
...
 endfunction()
+include(CheckTypeSize)
+check_type_size(void* ptr_size)
 if (LDC_INSTALL_LLVM_RUNTIME_LIBS)
     # Locate LLVM sanitizer runtime libraries, and copy them to our lib folder
     # Note: libFuzzer is part of compiler-rt version >= 6.0, but was part of LLVM =< 5.0
 
     if(APPLE)
         copy_compilerrt_lib("darwin/libclang_rt.asan_osx_dynamic.dylib" "libldc_rt.asan_osx_dynamic.dylib" TRUE)
+        copy_compilerrt_lib("darwin/libclang_rt.profile_osx.a"          "libldc_rt.profile_osx.a"          FALSE)
```
JohanEngelen: the builtins lib is called `libclang_rt.osx.a` on macOS.

JohanEngelen:
@kinke We should do this in steps. The PR has now become impossible to review.

kinke:
Not if you're reviewing the commits separately. It's all trivial changes. The relevant changes in the last commit, except for the removals, are in the top 2 files listed for that commit.

kinke:
What's still left to do is add the builtins lib on Posix too and probably move the `profile.di` header to druntime.

CMakeLists.txt
kinke: Reverted to only defaulting to ON on OSX and Windows.

CMakeLists.txt
kinke: Thx for verifying, added.

JohanEngelen:
The PR is doing so many things at the same time that it'll be hard to revert, hard to figure out which change caused what, etc. All against incremental development policy.

kinke:
I can always split off the removal in a separate PR, right now I just want to see some CI results. And they indicate that `profile.di` isn't available for the intermediate compiler, as it's only copied to the install dir. Moving it to druntime would solve this.

JohanEngelen:
> And they indicate that profile.di isn't available for the intermediate compiler

That's because you removed setting `PROFILERT_DIR`, which breaks `ldc2_phobos.conf.in`.

kinke:
Ah thx, I removed the offending line.

kinke:
Now `tests/sanitizers/fuzz_basic.d` has just failed for Circle.

JohanEngelen:
> Now tests/sanitizers/fuzz_basic.d has just failed for Circle.

Hm, so the program appears to be crashing on _something_, but it's another crash than the one intended... https://github.com/ldc-developers/ldc/pull/2546

PR #2545, opened 2018-01-30
[trivial] Add Lit version output to runlit.py.

JohanEngelen:


JohanEngelen:
Hm, I just saw that we already print the lit version elsewhere on our CI systems. I still think this has merit, but, what do you think?

PR #2546, opened 2018-01-30
Temporarily make fuzz_basic test more verbose to track down spurious failure

JohanEngelen:
When the test fails, it shows the libfuzzer output, so we can see crash details.


JohanEngelen:
Next time CI fails because of fuzz_basic.d, let me know.

PR #2549, opened 2018-01-31
Fix building with LLVM 7.0: update druntime.

JohanEngelen:
See: https://github.com/ldc-developers/druntime/pull/116


kinke:
We should probably introduce an ABI-like overload `llvm_memcpy(dst, dstAlign, src, srcAlign, size, volatile)` instead of the new one here, and forward to that one with 2 identical alignments. But we'll have to handle it manually then.

Issue #2548, opened 2018-01-31
Inline asm silently ignores opcodes db, ds, di, dl, df, dd, de


n8sh:
Trying to figure out why a function worked when compiled with DMD but not LDC I ran across this:

https://github.com/ldc-developers/ldc/blob/68d3c64318cc341c188d3bcf32bb497c4b5a1552/gen/asm-x86.h#L3888-L3890

At minimum this should be a compile-time error like with unknown opcodes.

klickverbot:
Indeed – somewhat surreal that you are the first person to encounter this, as that piece of code is ancient.

n8sh:
I suspect other people who encountered this were unlucky and there weren't signs that their code wasn't doing what they thought, like [this poor fellow from December 2016](http://www.0eval.com/dds-stuff/275).

JohanEngelen:
Pretty bad! PR: https://github.com/ldc-developers/ldc/pull/2550 .

klickverbot:
Fixed by #2550.

PR #2550, opened 2018-01-31
Error on data definition directives inside inline asm.

JohanEngelen:
See: https://github.com/ldc-developers/ldc/issues/2548


JohanEngelen:
breaks on this testcase: https://github.com/dlang/dmd/pull/3483

PR #2551, opened 2018-01-31
[trivial] LLVM7.0: update code to use new API for creating MemCpy intrinsic call.

JohanEngelen:
http://lists.llvm.org/pipermail/llvm-dev/2018-January/120542.html


kinke:
We may want to pass separate alignments at some point, but this will do fine for now.

PR #2553, opened 2018-02-03
README: Small tweaks/clarifications/plain text beautifications

klickverbot:
Removed the d1/ltsmaster branch links, as they are not very useful
(what would a user do on the GitHub page?) and bloat the file when
viewed in a terminal/editor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
 LDC – the LLVM-based D Compiler
 ===============================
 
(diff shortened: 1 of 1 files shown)


kinke:
Thx, lgtm.

> they are not very useful (what would a user do on the GitHub page?)

My thinking was that it may be useful to immediately see infos like `This branch is 4378 commits behind master.` and `Latest commit 5a9336b Mar 3, 2013`.

Issue #2547, opened 2018-01-31
Enable loading LLVM plugins


thewilsonator:
e.g. for AFLfuzz and `easy::jit`'s jit.

kinke:
Enabled by #2554.

PR #2554, opened 2018-02-04
Add LLVM-pass plugin support to LDC. Commandline flag: `-plugin=...`.

JohanEngelen:
This adds functionality to load plugins such as the AFL llvm-mode plugin: https://github.com/mirrorer/afl/blob/master/llvm_mode/afl-llvm-pass.so.cc

Adds a simple test that demonstrates how to create a plugin.
Test only enabled on Darwin and Linux.


JohanEngelen:
resolves https://github.com/ldc-developers/ldc/issues/2547

JohanEngelen:
Interesting: on Linux this needs `ldc2` to be built with `-Wl,--export-dynamic`.

JohanEngelen:
(the `./` in front of the `.so` name is needed on Linux because of this: https://stackoverflow.com/questions/17715774/c-linux-dlopen-cant-find-so-library )

JohanEngelen:
Green :)
Comments @klickverbot ?

kinke:
> on Linux this needs ldc2 to be built with -Wl,--export-dynamic.

I just built this PR and current master with LLVM 5.0.0 (Release + asserts), Release config, on Linux x64. The executable size increases by 20% (50 MB => 60 MB).

JohanEngelen:
> The executable size increases by 20% (50 MB => 60 MB).

Yeah, I was afraid it'd be a lot :/   Any alternatives? (I'm having a hard time figuring out what Clang does to provide the symbols for the plugin)

kinke:
I don't know, but simply making the plugin support opt-in via CMake sounds good enough to me, so that ARM builds etc. can remain small if need be.

JohanEngelen:
I'll add a (default off) CMake option for it. But I'm sure it'll mean that very few people will be able to use plugins then... :/
LLVM/Clang has similar issues it looks like. When LLVM is built/used as shared lib, things are OK though.
https://bugs.llvm.org/show_bug.cgi?id=18407
https://bugs.llvm.org/show_bug.cgi?id=13144
https://bugs.llvm.org/show_bug.cgi?id=19333

CMakeLists.txt
```diff
@@ -570,6 +570,7 @@ else()
 endif()
 set(LDC_ENABLE_PLUGINS ${LDC_ENABLE_PLUGINS_DEFAULT} CACHE BOOL "Build LDC with plugin support (increases binary size)")
 if(LDC_ENABLE_PLUGINS)
+    message(STATUS "Building LDC with plugin support")
```
thewilsonator: If a default of 'off' means that few people will try it, perhaps you should add a status "Building LDC without plugin support"?

kinke:
> But I'm sure it'll mean that very few people will be able to use plugins then... :/

I'd be fine with it defaulting to ON on non-Windows. Alternatively, enabling it explicitly for the Circle build is another way to significantly increase the feature availability (and additionally makes sure it is tested on Linux too, not just OSX as of now).

JohanEngelen:
Plugins are default on on Linux+Apple systems. Can be turned off with `-D LDC_ENABLE_PLUGINS=OFF`.
Good to squash+merge?

kinke:
Lgtm.

PR #2561, opened 2018-02-07
WIP: Dynamic pgo

Hardcode84:
Just some experiments

### DIFF
--- a/gen/dynamiccompile.cpp
+++ b/gen/dynamiccompile.cpp
@@ -207,33 +207,24 @@ void fixRtModule(llvm::Module &newModule,
     });
   }
 
-  int objectsFixed = 0;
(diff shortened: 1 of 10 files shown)


Hardcode84:
Works on toy examples but asserts "Critical edge is not split" in `PGOInstrumentationGenLegacyPass` on bigger project.

@klickverbot No, I only have few toy projects (old software rasterizer https://github.com/Hardcode84/ship3d is biggest)

Hardcode84:
For some reason `PGOInstrumentationGenLegacyPass`  doesn't like code with exception, marking everything `nothrow` helps.

Tested on my old software rasterizer, instrumentation makes code like 20% slower, and using this instrumentation data makes it like 3% faster.

PR #2562, opened 2018-02-07
Remove checking of fuzz testcase characters

JohanEngelen:
..., because they are not output when the reproducing testcase is too long (> 256 chars).

This should fix the spurious failures https://github.com/ldc-developers/ldc/pull/2556#issuecomment-363888571


kinke:
Heh that was quick, thx!

PR #2566, opened 2018-02-08
Solaris: fix assert for betterC build.

nykytenko:


gen/llvmhelpers.cpp
```diff
@@ -311,6 +311,14 @@ void DtoCAssert(Module *M, Loc &loc, LLValue *msg) {
     args.push_back(file);
     args.push_back(line);
     args.push_back(msg);
+  } else if (global.params.targetTriple->isOSSolaris()) {
+    const auto irFunc = gIR->func();
+    const auto funcName =
+        irFunc && irFunc->decl ? irFunc->decl->toPrettyChars() : "";
```
JohanEngelen: can you add parens here, to avoid relying on operator precedence rules. (i know it is a copy from a few lines above)

JohanEngelen:
LGTM (with the nit). A lit test would be nice for betterC assert function name (do we have one already?).

kinke:
A [DMD test](https://github.com/ldc-developers/dmd-testsuite/blob/ldc/runnable/betterc.d) makes sure a betterC program with assertions can be linked.

gen/llvmhelpers.cpp
nykytenko: Done.

PR #2567, opened 2018-02-08
ARM: Disable NEON if not supported

redstar:
On ARMv7 LLVM assumes that NEON instructions are supported. On many
devices this is not true. This commit detects this situation and
adds `-mattr=-neon` to the LDC configuration file.


JohanEngelen:
is this the buildbot illegal instruction issue?

JohanEngelen:
lgtm

redstar:
Yes. This fixes the build for ltsmaster.
For master branch it does not work. The phobos built does not use the ldc_phobos.conf - the build uses explicitly `-conf=`.

redstar:
@klickverbot Reading the current `ARM.td` file I say no:

    def ARMv7a    : Architecture<"armv7-a",   "ARMv7a",   [HasV7Ops,
                                                           FeatureNEON,
                                                           FeatureDB,
                                                           FeatureDSP,
                                                           FeatureAClass]>;

PR #2571, opened 2018-02-11
Don't waste an alloca for the nested context argument

kinke:
And so get rid of all loads too; just use the untouched pointer ~~argument~~ parameter directly.


kinke:
With this change, we could check whether the function accesses its nested context on the IR level by checking whether there are any uses of the IR parameter, e.g., for #2508.
Previously, there was always exactly one use, the store to memory, independent of whether the context was used or not.

thewilsonator:
Looks good. Feel free to push to #2508 since I'm not quite sure how to utilise this.

PR #2573, opened 2018-02-13
llvm 70 jit fixes

Hardcode84:
Also, I am going to drop jit support for pre 5.0 llvm to reduce ifdef clutter.

### DIFF
--- a/runtime/jit-rt/cpp-so/compile.cpp
+++ b/runtime/jit-rt/cpp-so/compile.cpp
@@ -186,17 +186,15 @@ public:
             llvm::sys::getHostCPUName(), getHostAttrs())),
         dataLayout(targetmachine->createDataLayout()),
 #if LDC_LLVM_VER >= 700
-        execSession(stringPool),
-        resolver(createResolver()),
-        objectLayer(
-          execSession,
-          [](llvm::orc::VModuleKey) {
-            return std::make_shared<llvm::SectionMemoryManager>();
-          },
-          [this](llvm::orc::VModuleKey) { return resolver; }),
+        execSession(stringPool), resolver(createResolver()),
+        objectLayer(execSession,
+                    [](llvm::orc::VModuleKey) {
+                      return std::make_shared<llvm::SectionMemoryManager>();
+                    },
+                    [this](llvm::orc::VModuleKey) { return resolver; }),
 #else
         objectLayer(
-          []() { return std::make_shared<llvm::SectionMemoryManager>(); }),
+            []() { return std::make_shared<llvm::SectionMemoryManager>(); }),
 #endif
         listenerlayer(objectLayer, ModuleListener(*targetmachine)),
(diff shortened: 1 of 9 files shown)


runtime/jit-rt/cpp-so/compile.cpp
```diff
@@ -198,21 +239,30 @@ class MyJIT {
...
+#if LDC_LLVM_VER >= 700
+    auto k = execSession.allocateVModule();
+    auto result = compileLayer.addModule(k, std::move(module));
+    if (result) {
+      execSession.releaseVModule(k);
+      return true;
+    }
+    moduleHandle = k;
+#elif LDC_LLVM_VER >= 500
+    auto Resolver = createResolver();
```
JohanEngelen: A bit dangerous to create a local variable with only a capitalization difference with a class member variable (`Resolver` vs. `resolver`) ?

JohanEngelen:
> Also, I am going to drop jit support for pre 5.0 llvm to reduce ifdef clutter.

Sounds very good. We release with LLVM 5.0, so it'll be available for many.

runtime/jit-rt/cpp-so/compile.cpp
```diff
@@ -187,9 +199,38 @@ class MyJIT {
             llvm::Triple(llvm::sys::getProcessTriple()), llvm::StringRef(),
             llvm::sys::getHostCPUName(), getHostAttrs())),
         dataLayout(targetmachine->createDataLayout()),
-#if LDC_LLVM_VER >= 500
+#if LDC_LLVM_VER >= 700
+        execSession(stringPool),
```
JohanEngelen: it's a pity that so much code is put in the initializer list of the ctor...

JohanEngelen:
(I don't have the time to review all of it, I trust it's OK ;-))

runtime/jit-rt/cpp-so/compile.cpp
```diff
@@ -483,8 +511,8 @@ __declspec(dllexport)
 #else
 __attribute__ ((visibility ("default")))
 #endif
-void JIT_API_ENTRYPOINT(const void *modlist_head,
-                        const Context *context, size_t contextSize) {
+    void JIT_API_ENTRYPOINT(const void *modlist_head, const Context *context,
```
JohanEngelen: clang-format bug?

runtime/jit-rt/cpp/compile.cpp
```diff
@@ -31,7 +31,7 @@ const void *dynamiccompile_modules_head = nullptr;
 #ifdef _WIN32
 __declspec(dllimport)
 #endif
-extern void JIT_API_ENTRYPOINT(const void *modlist_head,
+    extern void JIT_API_ENTRYPOINT(const void *modlist_head,
```
JohanEngelen: clang-format bug?

PR #2578, opened 2018-02-16
fix LLVM7.0 build

JohanEngelen:


runtime/jit-rt/cpp-so/compile.cpp
```diff
@@ -188,10 +188,11 @@ class MyJIT {
...
         objectLayer(execSession,
-                    [](llvm::orc::VModuleKey) {
-                      return std::make_shared<llvm::SectionMemoryManager>();
-                    },
-                    [this](llvm::orc::VModuleKey) { return resolver; }),
+                    [this](llvm::orc::VModuleKey) {
+                      return llvm::orc::RTDyldObjectLinkingLayer::Resources{
+                          std::make_shared<llvm::SectionMemoryManager>(),
+                          resolver};
+                    }),
```
JohanEngelen: @Hardcode84 Does this make sense?

JohanEngelen:
@Hardcode84 If you agree with the JIT code change, please squash+merge, thanks.

Issue #2442, opened 2017-12-09
macOS binary releases should set shared stdlib install_name to relative path


klickverbot:
For the shared druntime/Phobos libraries to be used from a client program, the user currently needs to go through dynamic linker search path gymnastics or manually run `install_name_tool` after extracting a release tarball.

One way of doing this would be to set `-install_name` to `@rpath/{druntime,phobos2}-ldc-shared.dylib` on linking (or after the fact using `install_name_tool`), and telling users to set `-rpath` accordingly.

(Am I missing a more elegant way of doing this?)

klickverbot:
See #2448 – perhaps just using the plain library name would be a better default? (@timotheecour?)

timotheecour:
@klickverbot @rpath/{druntime,phobos2}-ldc-shared.dylib + telling users to set -rpath accordingly seems indeed like the preferred way to go.

Also relevant is how cmake does it:

```
cmake --help-policy CMP0042
CMP0042
-------

``MACOSX_RPATH`` is enabled by default.

CMake 2.8.12 and newer has support for using ``@rpath`` in a target's install
name.  This was enabled by setting the target property
``MACOSX_RPATH``.  The ``@rpath`` in an install name is a more
flexible and powerful mechanism than ``@executable_path`` or ``@loader_path``
for locating shared libraries.

CMake 3.0 and later prefer this property to be ON by default.  Projects
wanting ``@rpath`` in a target's install name may remove any setting of
the ``INSTALL_NAME_DIR`` and ``CMAKE_INSTALL_NAME_DIR``
variables.

This policy was introduced in CMake version 3.0.  CMake version
3.10.0 warns when the policy is not set and uses OLD behavior.  Use
the cmake_policy command to set it to OLD or NEW explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.

```

kinke:
Should be fixed by #2581.

PR #2581, opened 2018-02-17
OSX: Set install_name of merged dylibs (issue #2442)

kinke:
If I understand the CMake docs correctly, CMake >= 3 defaults to setting the install_name appropriately when installing a shared library target (e.g., for MULTILIB=OFF).

Once again, the 32/64-bit merging step for OSX libraries prevents this automatism, as the merged libs are custom targets, so do it manually in form of a post-build command.


kinke:
@klickverbot: Good to go?

Issue #2585, opened 2018-02-19
1.8.0-beta1, OS X: Uncaught exceptions using [Thin|Full]LTO+PGO; FullLTO alone


jondegenhardt:
This is another variant of #2208 (feel free to close as a duplicate, but I wanted to highlight new info).

One of my tools, `tsv-filter`, has uncaught exceptions when built with FullLTO (with or without PGO) or ThinLTO+PGO. The latter case is new in 1.8.0-beta1. In prior versions (1.3.0 - 1.7.0) this failure occurs with FullLTO, but not ThinLTO. This is material for my `tsv-filter` tool because it means that LTO & PGO cannot be used together in 1.8.0-beta1.

As background - The `tsv-filter` tool has a fair number of delegates (about 45) invoking code that can throw exceptions, plus a top-level exception catcher. Exceptions are triggered when checking command line args. Unit tests provide an invalid input for each. Approximately the first 30 delegates throw exceptions that are caught successfully, the but exceptions thrown by the remaining delegates are not caught.

There is more info in the printed output in 1.8.3-beta1 that may be helpful tracking down the cause. The executable itself exits printing:
```
Illegal instruction: 4
```
This is printed to the console, independent of stdout or stderr. The message to stderr follows a form like:
```
uncaught exception
object.Exception@/Users/jdegenhardt/devtools/dlang/ldc2-1.8.0-beta1-osx-x86_64/bin/../import/std/getopt.d(880): Missing value for argument --ff-lt.
----------------
0   tsv-filter                          0x0000000102750156 object.Throwable.TraceInfo core.runtime.defaultTraceHandler(void*) + 246
```

The errors are all similar, differing primarily in text describing cause. In particular, all refer to the same `object.Thowable.TraceInfo core.runtime.defaultTraceHandler(void*)`. There was no reference to in output of previous LDC versions. Perhaps clue here.

I'll take a shot at a reduced test case as well.

JohanEngelen:
for our info, this is with LTO'd Phobos+druntime, right?

jondegenhardt:
> for our info, this is with LTO'd Phobos+druntime, right?

Yes. Sorry, that should have been in the report.

jondegenhardt:
A reduced case:
[getopt_uncaught.d.txt](https://github.com/ldc-developers/ldc/files/1738295/getopt_uncaught.d.txt)

[...]

PR #2587, opened 2018-02-20
WIP: Upgrade to D v2.079.1

kinke:


kinke:
Status update: LDC + druntime/Phobos building fine on Win64 and Linux x64, most lit tests already green.

kinke:
Status update on Win64: all unittests + druntime standalone tests green; 2 failing lit-tests (PGO/hash.d and codegen/betterC_typeinfo.d).
There's a bunch of dmd-testsuite failures, primarily due to missing support for new `-i` and `-Xi` switches as well as a betterC TypeInfo regression (erroring out whenever coming across a struct in betterC mode). The new upstream minimal/pay-as-you-go druntime support will also need to be checked.

=> looking good; I expect similar results on Linux x64, but getting CI to work is a bit tedious (apparently more Phobos unittest module compilations requiring significantly more than 1 GB of memory; 32-bit std.file unittests crashing for Semaphore, possibly due to their RAM disk...).
OSX builds quickly die during CMake setup due to duplicate symbols; I haven't looked further into that.

kinke:
Missing switches implemented. Most remaining dmd-testsuite failures are due to different error messages for the many new CLI tests.

kinke:
Down to 2 dmd-testsuite failures on Win64, `{runnable,compilable}/minimal.d`, all other tests green on that platform. We'll have to check whether we can delay some object.d checks or narrow down the requirements to support a minimal druntime.

On Linux, the 32-bit std.file unittests with enabled optimizations fail, related to EH abort()s apparently, in addition to the 2 dmd-testsuite failures above.

On macOS, the duplicate vtable symbols issue blocks all CI tests:
```
duplicate symbol __D2gc4impl5protoQo7ProtoGC11__interface2gc11gcinterface2GC6Thn16_6__vtblZ in:
    /Users/distiller/bootstrap/lib/libdruntime-ldc.a(proxy.o)
    /Users/distiller/bootstrap/lib/libdruntime-ldc.a(gc.o)
ld: 1 duplicate symbol for architecture x86_64
```

kinke:
The unwinding issue for the 32-bit Linux std.file unittests seems bizarre. It can be reproduced via `bin/ldc2 -main -unittest -O -g ../ldc/runtime/phobos/std/file.d -m32 && ./file`; it needs the release libs and of course doesn't fail if the unittest block is extracted into a separate file. A call to `_Unwind_Resume()` results in an abort; i.e., smelling very much like the ARM EH issue.

I turned on `debug = EH_personality;` in `rt.dwarfeh`, lots of output, same abort. Additionally adding a printf() call in `_d_eh_resume_unwind()` (before calling _Unwind_Resume) made it consistently work, as long as providing an arg to be formatted.

So I checked the optimized assembly for `void _d_eh_resume_unwind(void* ptr) { return _Unwind_Resume(cast(_Unwind_Exception*) ptr); }` in druntime. The X86 backend somehow manages to screw up
```LL
define x86_stdcallcc void @void onlineapp._d_eh_resume_unwind(void*)(i8* inreg %ptr_arg) local_unnamed_addr #0 comdat {
  tail call void @_Unwind_Resume(i8* %ptr_arg) #0
  ret void
}
```
to
```asm
	.cfi_startproc
	pushl	%ebx
.Lcfi0:
	.cfi_def_cfa_offset 8
	subl	$8, %esp
.Lcfi1:
	.cfi_def_cfa_offset 16
.Lcfi2:
	.cfi_offset %ebx, -8
	calll	.L0$pb
.Lcfi3:
	.cfi_adjust_cfa_offset 4
.L0$pb:
	popl	%ebx
.Lcfi4:
	.cfi_adjust_cfa_offset -4
	movl	%eax, (%esp)
.Ltmp0:
	addl	$_GLOBAL_OFFSET_TABLE_+(.Ltmp0-.L0$pb), %ebx
	calll	_Unwind_Resume@PLT
	addl	$8, %esp
	popl	%ebx
	retl
.Lfunc_end0:
	.size	void onlineapp._d_eh_resume_unwind(void*), .Lfunc_end0-void onlineapp._d_eh_resume_unwind(void*)
	.cfi_endproc
```

See https://run.dlang.io/is/qPiiCR.

Defining it as
```D
void _d_eh_resume_unwind(void* ptr)
{
    asm { naked; jmp _Unwind_Resume; }
}
```
seems to make it work (with static druntime, shared druntime complains about invalid relocation...).

This [comment](https://github.com/ldc-developers/druntime/blob/756fa7160dfb5bd1acc0d0fc6091be6f67a17a01/src/ldc/eh_asm.S#L33-L40) seems interesting:
> Note: Current codegen of D catch landing pads are incompatible with GCC provided _Unwind_Resume because the LLVM inliner can create landing pads that advertise to catch more exceptions than are  handled, falling into _d_eh_resume_unwind to find the real handler.  _Unwind_Resume ignores the saved IP and resets it to the original  call site in this frame, but we need the callsite of _d_eh_resume_unwind to find the next landing pad.  Workaround is to capture it, passing to _d_arm_eabi_end_cleanup as second arg.

kinke:
A `printf("bla");` (optimized to a puts call) before calling _Unwind_Resume() wouldn't do, but a `printf("bla %p", ptr)` did. When I diffed the 2 assemblies, one thing that stood out was that the single-arg one subtracted 8 from ESP (like the original code above without printf), while the (working) two-args version subtracted 32. The alignment of the function is 16 bytes...

kinke:
Right, I see now.

I switched to calling `_Unwind_Resume()` (as runtime function) directly, without `_d_eh_resume_unwind()` trampoline, that seems to work fine.

kinke:
> Should _Unwind_Resume be noreturn, by the way?

I added cold & noreturn in #2642.

---

Wrt. the remaining OSX issue, the issue is that a global in [gc.proxy](https://github.com/ldc-developers/druntime/blob/ldc-merge-2.079/src/gc/proxy.d#L33) is initialized with a concrete class reference (`gc.impl.proto.gc.ProtoGC`) since 2.079.
The static init leads to the interface vtables and thunks being emitted into the `proxy` object file too, not just the `gc` object. As the class is not a template, the vtable symbols gets normal non-template linkage; everywhere except on OSX, it gets a comdat though, so the duplicate symbols are unified by the linker later.

So statically initializing globals with class references seems rarely used, otherwise we'd probably have heard of this OSX bug before.

kinke:
Ah, feels good:
```
LDC - the LLVM D compiler (1.9.0git-025c69c):
  based on DMD v2.079.1 and LLVM 6.0.0
```

So there's one more remaining issue, `runnable/testconst.d` segfaulting for Travis Linux with `BUILD_SHARED_LIBS=ON`.

kinke:
As we don't have SSH access for Travis (at least I don't), I quickly modified the Semaphore config to build with `BUILD_SHARED_LIBS=ON` and planned on interactively debugging it, but it [works](https://semaphoreci.com/ldc-developers/ldc/branches/merge-2-079/builds/38). It's using (LDC-)LLVM 4.0.1; Travis fails with vanilla LLVM 4.0.1 (and 3.9.1, these are the 2 Travis Linux jobs with shared libs only; the OSX shared-libs only job with LLVM 4.0.0 works too). I'm going to experimentally set `BUILD_SHARED_LIBS=ON` for all Travis jobs.

kinke:
So Travis Linux fails in general, not depending on LLVM version. It's based on Ubuntu 14.04, just like Semaphore. Awaiting Circle results (Ubuntu 14.04 too).

kinke:
Working fine on Circle Linux too. So I guess it's somehow related to the Travis environment; I propose to publish 1.9-beta1 as-is and check if people run into any issues.

joakim-noah:
Sounds good.

kinke:
Using `BUILD_SHARED_LIBS=ON` for Circle Mac too revealed another issue; exceptions in `d_do_test` aren't caught (`_Unwind_RaiseException()` returning with `_URC_END_OF_STACK`, as if there weren't any catch clauses in the call stack - there's at least one explicit one in that tester plus the implicit root one in druntime), leading to many dmd-testsuite failues. The symptoms are the same as for #2585, so it may not be a new issue.

call stack from Circle SSH, just in case:
```
  * frame #0: 0x00007fffa34d0d42 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fffa35be457 libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x00007fffa3436420 libsystem_c.dylib`abort + 129
    frame #3: 0x00000001008b4c83 libdruntime-ldc-shared.dylib`_d_throw_exception + 227
    frame #4: 0x00000001002bafb5 libphobos2-ldc-shared.dylib`_D3std9exception__T7enforceZ__TQmTbZQrFNaNfbLAxaAyamZb + 261
    frame #5: 0x0000000100005a35 d_do_test`_D9d_do_test7executeFKS3std5stdio4FileAyabQeZQh(f=0x00007fff5fbfe380, command=(length = 46, ptr = 0x0000000100d02140), expectpass=true, result_path=(length = 26, ptr = 0x0000000100d00080)) at d_do_test.d:406
    frame #6: 0x000000010000a67a d_do_test`_D9d_do_test7tryMainFAAyaZ15testCombinationMFbQymQBbZEQCaQBtFQBoZ6Result(autoCompileImports=false, argSet=(length = 0, ptr = 0x00000001000851c8), permuteIndex=0, permutedArgs=(length = 0, ptr = 0x0000000000000000)) at d_do_test.d:789
    frame #7: 0x000000010000891f d_do_test`_D9d_do_test7tryMainFAAyaZi(args=string[] @ 0x00007fff5fbfef88) at d_do_test.d:848
    frame #8: 0x00000001000071ad d_do_test`_Dmain(args=string[] @ 0x00007fff5fbfeff0) at d_do_test.d:517
    frame #9: 0x00000001008b4640 libdruntime-ldc-shared.dylib`_D2rt6dmain211_d_run_mainUiPPaPUAAaZiZ6runAllMFZv + 112
    frame #10: 0x00000001008b44a1 libdruntime-ldc-shared.dylib`_d_run_main + 433
    frame #11: 0x00000001000815a5 d_do_test`main(argc=4, argv=0x00007fff5fbff1d8) at __entrypoint.d:8
    frame #12: 0x00007fffa33a2235 libdyld.dylib`start + 1
    frame #13: 0x00007fffa33a2235 libdyld.dylib`start + 1
```

joakim-noah:
I'm seeing a strange error when compiling for Android/ARM, which isn't reproducible when building for Linux/x64. Here's the command and error:
```
./bin/ldc2 -c --output-o -conf= -w -mcpu=cortex-a8 -mtriple=armv7-none-linux-android -O3 -release -unittest -d-version=StdUnittest -I../runtime/druntime/src -I../runtime/phobos -of=timezone.o ../runtime/phobos/std/datetime/timezone.d

std/datetime/timezone.d(2845): Error: undefined identifier `DateTimeException`
```
You should be able to reproduce with that command, as long as your llvm has the ARM backend. No idea why this frontend error would vary by arch.

kinke:
> std/datetime/timezone.d(2845)

Notice the relative path - it's not `../runtime/phobos/std/datetime/timezone.d`. Try running the compilation command from another working directory.

joakim-noah:
No, that's not the problem.  I took the generated command spit out by ninja and modified it so you could run it from a build directory easily, ie without my absolute paths for druntime and Phobos. Our config builds that file from the Phobos directory, hence that error I cut and pasted, but this command should spit out the same error but with the relative path, at least it did when I just tried it.

kinke:
This looks like an (upstream?) regression for Android, as that line 2845 is in a `version(Android)` block. In other places, e.g., [here](https://github.com/ldc-developers/phobos/blob/ldc-merge-2.079/std/datetime/timezone.d#L2734), the symbol is imported before using it.

joakim-noah:
Yes, you're right, it's obviously a regression upstream from scoping the imports and not testing with Android. I'm so used to there being almost no Android differences in Phobos that I didn't even look at that file, but assumed it was a dmd frontend issue once it worked fine for linux.

Sorry about the noise, adding the import easily fixes it.

kinke:
Glad you noticed so we can fix it before tagging beta1.

joakim-noah:
Alright, with the simple fix for that regression that I merged to this branch at ldc-developers/phobos@eb79925, all the stdlib unit tests pass for Android/ARM, except for the single long-standing ARM EH inlining issue. Only two more tests fail from the dmd testsuite after updating from ldc 1.8 to this 1.9 branch, likely nothing substantial but haven't looked into them.

I also tried building the stdlib tests for Android/AArch64, pretty much the same results as 1.8, with only two different floating-point-related failures in `std.complex`.

kinke:
> Using BUILD_SHARED_LIBS=ON for Circle Mac too revealed another issue

... which is gone with #2659, see #2660.

PR #2590, opened 2018-02-21
Differentiate C++ catch clauses from D ones with __cpp_type_info_ptr

Syniurge:
The following code currently segfaults:

```d
extern (C++) void throwStdException();

void main() {
    import std.stdio : writeln;

    try {
        throwStdException();
    } catch (Throwable t) {
        writeln("Caught a Throwable");
    }
}
```

```cpp
#include <exception>
void throwStdException() { throw std::exception(); }
```

Because the personality routine in [`druntime/rt/dwarfeh.d`](https://github.com/ldc-developers/druntime/blob/ec48d54d331519f87beb03d82d5c4934eef3f03a/src/rt/dwarfeh.d#L1240) makes the wrong assumption that when it encounters a C++ exception, the catch clauses in the action table that it gets to examine are raw `std::type_info` pointers, which isn't the case if it's a `Throwable` (or derived) catch.

To make it possible for the personality routine to differentiate the two kinds of catch clauses, DMD wraps `std::type_info` pointers inside `__cpp_type_info_ptr` class instances, and this is what this PR implements for LDC.

-----

Associated ldc/druntime PR: https://github.com/ldc-developers/druntime/pull/119 

BTW should the druntime submodule be updated in the ldc commit too?

### DIFF
--- a/gen/trycatchfinally.cpp
+++ b/gen/trycatchfinally.cpp
@@ -14,10 +14,13 @@
 #include "gen/classes.h"
 #include "gen/funcgenstate.h"
 #include "gen/llvmhelpers.h"
+#include "gen/mangling.h"
 #include "gen/ms-cxx-helper.h"
+#include "gen/rttibuilder.h"
 #include "gen/runtime.h"
 #include "gen/tollvm.h"
 #include "ir/irfunction.h"
+#include "ir/irtypeclass.h"
 
 ////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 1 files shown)


kinke:
Looking good, thanks a lot, I remember having trouble to understand what DMD was doing there.

> BTW should the druntime submodule be updated in the ldc commit too?

Only possible if the commit is in the LDC repo, i.e., already in the main `ldc` branch or on its own branch - the latter is what I just did, so please include the submodule update for proper CI testing.

We'll also have to add your testcase in dmd-testsuite, too bad this quite important aspect of DMD's implementation isn't tested at all.

Syniurge:
> We'll also have to add your testcase to dmd-testsuite, too bad this quite important aspect of DMD's implementation isn't tested at all. Edit: Like here, throwing C++ function already available.

Added a `try...catch(Throwable)` block inside that test: https://github.com/ldc-developers/dmd-testsuite/pull/33

gen/trycatchfinally.cpp
```diff
@@ -159,9 +161,27 @@ void TryCatchScope::emitCatchBodies(IRState &irs, llvm::Value *ehPtrSlot) {
...
+      auto cpp_ti = getOrCreateGlobal(
           p.cd->loc, irs.module, getVoidPtrType(), /*isConstant=*/true,
           LLGlobalValue::ExternalLinkage, /*init=*/nullptr, name);
+
+      // Wrap std::type_info pointers inside a __cpp_type_info_ptr class instance so that
+      // the personality routine may differentiate C++ catch clauses from D ones.
+      OutBuffer mangleBuf;
+      mangleBuf.writestring("_D");
+      mangleToBuffer(p.cd, &mangleBuf);
+      mangleBuf.printf("%d%s", 18, "_cpp_type_info_ptr");
```
kinke: We have the concept of 'IR mangles', to give the TargetABI a chance to tweak the mangled name. So please use `getIRMangledVarName(mangleBuf.peekString(), LINKd)` as name for the `getOrCreateGlobal()` call below. While at it, please also include the dmd-testsuite update.

Syniurge:
Ready for m-.. oh you already did, thanks!

PR #2593, opened 2018-02-22
[Trivial] Fix isnan/isinf based on changes in dragonfly math.h

dkgroot:


dkgroot:
@joakim-noah Need to figure out what changed with isnan / isinf. Seems to also be the cause for the posix/math.d issue in the other PR.
I need to either use all the platform/os specific version of the math functions, or use the llvm builtin ones. Mixing them leads to problems with rounding during asserts.

dkgroot:
[DFly SemaphoreCI](https://semaphoreci.com/dkgroot/ldc_dragonfly_ci)

dkgroot:
@joakim-noah This trivial change is still open (would be nice if it could be included. There was a change in the DragonFly library a while back, causing this 'regression'

dkgroot:
Thanks !

PR #2594, opened 2018-02-23
[ltsmaster] Fix backport of PR #2315.

redstar:
A changed condition leads to a missing load.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -62,7 +62,9 @@ Module *ldc::DIBuilder::getDefinedModule(Dsymbol *s) {
 ////////////////////////////////////////////////////////////////////////////////
 
 ldc::DIBuilder::DIBuilder(IRState *const IR)
-    : IR(IR), DBuilder(IR->module), CUNode(nullptr) {}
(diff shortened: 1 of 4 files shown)


gen/dibuilder.cpp
```diff
@@ -75,27 +77,50 @@ ldc::DIScope ldc::DIBuilder::GetCurrentScope() {
...
+  instr->setDebugLoc(debugLoc);
+#else // if LLVM >= 3.7
+  DBuilder.insertDeclare(storage, divar, diexpr, debugLoc, IR->scopebb());
+#endif
+}
+
+// Sets the (current) value for a debuginfo variable.
+void ldc::DIBuilder::SetValue(const Loc &loc, llvm::Value *value,
+                              ldc::DILocalVariable divar
+#if LDC_LLVM_VER >= 306
```
thewilsonator: LLVM 3.6 support was dropped in master. Do we still support it in ltsmaster?

redstar: Yes, we still support LLVM 3.6 in ltsmaster.

PR #2595, opened 2018-02-23
[ltsmaster] Update Travis configuration.

redstar:
Update to LLVM 5.0.1 and add LLVM 3.6.2 (which is still supported
on ltsmaster).


JohanEngelen:
@redstar Not sure whether it is worthwhile to still support LLVM 3.6.x

PR #2601, opened 2018-02-27
Sanitizers: add blacklisting of source files

JohanEngelen:


JohanEngelen:
(I'm working on a blacklist file for druntime and phobos. At first it is blacklisting pretty much everything. Having this per-file functionality will be quite useful.)

PR #2603, opened 2018-03-01
[ltsmaster] Enable LLVM 3.5 build again.

redstar:
We claim in the release notes that we support LLVM 3.5. Let's
check this.


redstar:
This was just a test.

PR #2604, opened 2018-03-02
Pull in Musl's assert function for betterC mode.

joakim-noah:
With this pull, only two tests fail from the dmd testsuite on Alpine/x64.  [This one asserts because of some stack trace issue](https://github.com/ldc-developers/dmd-testsuite/blob/ldc/runnable/test17559.d#L48), and [this 32-bit test fails out with a different error than expected](https://github.com/ldc-developers/dmd-testsuite/blob/ldc/fail_compilation/fail80_m32.d) because [this druntime module has not been ported to `X86` yet](https://github.com/ldc-developers/druntime/blob/ldc/src/core/sys/posix/sys/types.d#L642). Neither is a big deal, so unless @yshui wants to handle them, this patch should be enough for the 1.8 release.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -827,10 +827,8 @@ void registerPredefinedTargetVersions() {
     if (triple.getEnvironment() == llvm::Triple::Android) {
       VersionCondition::addPredefinedGlobalIdent("Android");
       VersionCondition::addPredefinedGlobalIdent("CRuntime_Bionic");
-#if LDC_LLVM_VER >= 309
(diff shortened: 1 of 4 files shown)


JohanEngelen:
LLVM <= 3.8 doesn't have `Triple::isMusl`. So you can add a helper function `bool isMusl(Triple &t)` and `#ifdef` inside, to avoid having `#ifdef` everwhere. 
LGTM otherwise.

joakim-noah:
@kinke, good to merge, don't think those CI failures matter?

PR #2606, opened 2018-03-05
Fixup ABI for uClibc

rracariu:
The default ABI determined in the `createTargetMachine` for the `uclibc` environment and `Linux` `ARM` is `EABI`, uClibc uses `GNUEABI`.

Without the correct ABI, exceptions, array ops and probably others where not working.

Don't know if this is the best way to do it,  but this made druntime test pass with the triple `armv6-hardfloat-linux-uclibc` on actual hardware.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -165,6 +165,9 @@ void printVersionStdout() {
 
 namespace {
 
+// True when target triple has an uClibc environment
+bool isUClibc = false;
+
 // Helper function to handle -d-debug=* and -d-version=*
 void processVersions(std::vector<std::string> &list, const char *type,
                      unsigned &globalLevel, Strings *&globalIDs) {
@@ -642,6 +645,19 @@ static void registerMipsABI() {
   }
 }
 
+// Check if triple environment name starts with "uclibc" and change it to "gnu"
+void fixupUClibcEnv()
+{
+  llvm::Triple triple(mTargetTriple);
+  if (triple.getEnvironmentName().find("uclibc") != 0)
+    return;
+  std::string envName = triple.getEnvironmentName();
+  envName.replace(0, 6, "gnu");
+  triple.setEnvironmentName(envName);
+  mTargetTriple = triple.normalize();
+  isUClibc = true;
+}
+
 /// Register the float ABI.
 /// Also defines D_HardFloat or D_SoftFloat depending if FPU should be used
 void registerPredefinedFloatABI(const char *soft, const char *hard,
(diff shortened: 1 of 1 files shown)


kinke:
What's the default triple of your LLVM?

rracariu:
@klickverbot I got confused, the same name is used in LLVM, and my grep foo missed the `targetmachine.cpp` one. I'll look on how to move it there.

@kinke my default triple is `x86_64-pc-linux-gnu`

rracariu:
Tried to move that quick fix-up in `createTargetMachine` function. The problem is that I can't perform the same trick of switching back the environment name. 

1. the target machine triple is const
2. even after const_cast the EH is broken (the program halts when an exception is thrown).

The environment name is used somewhere to configure the right ABI and EH model

I also tried to set the ABI and EH model in `TargetOptions` directly, before constructing the target machine, but still haven't got the EH working.  I applied the same `ABI` type selected for `gnueabi` environment. Strangely the 'ExceptionModel' for `gnueabi` is 'None'...

Setting the `ExceptionModel` to `llvm::ExceptionHandling::DwarfCFI` resulted in a crash in `_d_throw_exception` as `_Unwind_RaiseException` returned with some random exit code.

I'm in pickle here, suggestions are welcome, thanks!

kinke:
I asked for the default triple as I hoped you'd have a native LLVM, to figure out what the triple should look like. According to the LLVM docs, there are various environments for musl (Musl, MuslEABI, MuslEABIHF in the `EnvironmentType` enum). Due to unfortunately lacking official support for uClibc, I guess we should expect similar triples (`armv6-linux-uclibceabihf`?) and will have to map that to the appropriate environment for `createTargetMachine()` (by replacing the `uclibc` substring in the environment string with `gnu`?) but then restore the custom environment string in order to be able to detect uClibc later on. Or the other way round, detect `uclibc` in the original environment string, save it somewhere inside LDC and then use the according GNU environment for an 'official' triple.

rracariu:
This is what I'm doing in this patch, overloading the `gnueabi` environment and then switching back to `uclibc` for the global triple.

The issue raised by @klickverbot  was that the location is not the right one, suggesting to move it in `targetmachine.cpp`. I couldn't make it work there, due to the way the global triple is loaded and the fact that changing the target machine triple doesn't work...

kinke:
The problem I have with the current patch is that it only works with an exact `uclibc` environment. It should support `uclibc` as a substring instead, so that different variants can be used (`uclibc[eabi[hf]]`). I guess the current logic already handles the separate float-ABI cmdline option etc. correctly, so that mapping `uclibc` to `gnu` beforehand should do.

I don't know what LLVM's reasons for not supporting uClibc officially are; if it's because there are only negligible diffs to GNU, we may as well just stash away the uclibc flag (only required for the predefined `CRuntime_UClibc`, right?) and then use the fixed-up (GNU) triple.

JohanEngelen:
@rracariu Did you try to getting support for this into LLVM?  I can't find indications of someone already pursuing that.

rracariu:
@kinke  I changed the approach as suggested by you as it is indeed more flexible and simpler.

@JohanEngelen I haven't considered moving it to LLVM, it looks to be a big undertaking for relative low chance of inclusion as there is major similarity between the two.

Thanks for the input!

kinke:
Thanks!

Issue #2356, opened 2017-10-03
ldc2 fails to run simple main example on ppc64el


LocutusOfBorg:
Debian Buster:
```
$ ldc2  -v
binary    /usr/bin/ldc2
version   1.3.0 (DMD v2.073.2, LLVM 4.0.1)
config    /etc/ldc2.conf (powerpc64le-unknown-linux-gnu)
OVERVIEW: LDC - the LLVM D compiler

ldc 1.3.0-2 is fine:
echo "void main() {}" > foo.d
ldc2 foo.d
./foo
echo $?
```

Debian sid, ldc 1.4
```
$ ldc2 -v
binary    /usr/bin/ldc2
version   1.4.0 (DMD v2.074.1, LLVM 5.0.0)
config    /etc/ldc2.conf (powerpc64le-unknown-linux-gnu)
OVERVIEW: LDC - the LLVM D compiler
```
```
$ echo "void main() {}" > foo.d
$ ldc2 foo.d
$ ./foo
Segmentation fault
$ echo $?
139
$ gdb foo
GNU gdb (Debian 7.12-6) 7.12.0.20161007-git
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "powerpc64le-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from foo...(no debugging symbols found)...done.
(gdb) r 
Starting program: /home/locutusofborg/foo 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/powerpc64le-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00003fffb799c3a0 in rt.util.container.array.Array!(void[]).Array.opSlice() inout () from /usr/lib/powerpc64le-linux-gnu/libdruntime-ldc.so.74
(gdb) bt
[...]

PR #2608, opened 2018-03-09
Add support for LLD 6.0.0 & switch CI to LDC-LLVM 6.0.0

kinke:


kinke:
@thewilsonator: Shall we include SPIR-V in LDC-LLVM 6.0.0? Our [ldc-release_60](https://github.com/ldc-developers/llvm/tree/ldc-release_60) branch should be ready [incl. compiler-rt as git submodule with automatic build inclusion].

kinke:
@joakim-noah: It'd be great if you could check our Android compatibility with LLVM 6.0.0 via the source artifact available [here](https://circleci.com/gh/ldc-developers/llvm/122#artifacts/containers/0) (I had to resolve a few conflicts while cherry-picking).

kinke:
Incomplete list of 6.0.0 regressions:

- [ ] Linux x86_64: failing lit-test `codegen/exception_stack_trace.d`
- [ ] Win32: issue #2257 preventing defaultlib compilation
- [ ] Win64: failing lit-tests `debuginfo/nested_cdb.d` and `sanitizers/fsanitize_coverage.d`

joakim-noah:
No problem compiling LDC against the pre-built llvm you linked, but it doesn't link the final `ldc2` binary. Will investigate further.

kinke:
> but it doesn't link the final ldc2 binary

Same issue on OSX apparently; I guess that's related to the order of the new LLD libs...

kinke:
The Win32 inline asm issue appears to be triggered by [core.cpuid](https://github.com/ldc-developers/druntime/blob/ldc/src/core/cpuid.d#L514-L521), with enabled optimizations only, so probably has to be considered as bug.

Failing `debuginfo/nested_cdb.d` on Win64 (most likely Win32 too) also appears to be a CodeView debuginfo regression (CodeView support is supposed to be greatly improved by v6); once again related to the GEPs we use for nested variables, I can't believe that's bothering us again. :(

kinke:
Oh @joakim-noah: I can't use the prebuilt LLVMs in my Ubuntu 16.04 VM either. They're built on Ubuntu 14.04 with gcc 4.9, and IIRC there have been breaking ABI changes with newer gcc versions.

joakim-noah:
OK, I've kicked off a build of the latest `llvm-69a834b2.src` for Android/ARM, though I had to stop it from generating the build target for the tests with `-DLLVM_INCLUDE_TESTS=OFF` because of some CMake errors otherwise.

kinke:
[Ah yes sry I forgot to mention that `COMPILER_RT_INCLUDE_TESTS=OFF` is required when building LDC-LLVM to get rid of the clang dependency.]

kinke:
The exception stack trace issue on Linux x86_64 is due to the output of `ldc2 -link-defaultlib-debug -run ../ldc/tests/codegen/exception_stack_trace.d`:
```
object.Exception@../ldc/tests/codegen/exception_stack_trace.d(6): lala
----------------
runtime.d:779 [0x425dd8]
runtime.d:1035 [0x425718]
dmain2.d:288 [0x40b6d0]
deh.d:24 [0x42a64e]
dwarfeh.d:316 [0x40c1c2]
```
regressing to
```
object.Exception@../ldc/tests/codegen/exception_stack_trace.d(6): lala
----------------
runtime.d:2734 [0x4187c8]
```

Only with debug defaultlibs that is; the release libs show no change:
```
??:? [0x414f6e]
??:? [0x4180aa]
??:? [0x406e4c]
??:? [0x40224d]
??:? [0x402258]
??:? [0x40226c]
??:? [0x406bdf]
??:? [0x406aeb]
??:? [0x4023c4]
??:? __libc_start_main [0x3227b82f]
??:? [0x4020c8]
```

joakim-noah:
Alright, finally got `ldc2` to build on linux/x64 and cross-compile for Android/ARM once I built llvm from source.  I had to turn `LLVM_INCLUDE_TESTS` back on and disable `COMPILER_RT_INCLUDE_TESTS` instead like you said, or the generated `llvm-config` doesn't work properly, spitting out some weird errors about missing libgtest.a when run with `--libs`.

All the same stdlib tests pass with your tweaked llvm 6.0 and ldc master, with only that long-standing EH test failing.

kinke:
LLVM 6.0.0 appears to default to a newer Dwarf version (at least for Linux), which the stacktrace generator in druntime doesn't handle yet. This then lead to the discovery of a [bug](https://github.com/dlang/druntime/pull/2151#discussion_r176944607) in that druntime code.

kinke:
Should be ready now, except for LDC-LLVM 6.0 finalization (@thewilsonator: SPIR-V?). I promoted the 2 MSVC regressions to separate issues now.

joakim-noah:
How's the built-in lld working these days? I'm thinking of trying it out for Android and switching to it if it passes all the tests. I could remove the binutils/clang dependency for the Termux Android package too.

kinke:
Pretty sure it wouldn't cause a lot of issues, but the huge command-line issue remains the [same](https://github.com/ldc-developers/ldc/pull/2203#issuecomment-339167131).
With LLD 6, there's a new WebAssembly driver which could prove useful sometime.

joakim-noah:
Yeah, gathering all those flags is a pain for an arbitrary distro but we could probably do it for more standard setups like Win/Mac or the Android NDK and Termux app. I'll try it out.

kinke:
[It's working fine on Windows, there's no cmdline hell as on Posix; it's all in environment variables, which are set up by calling an MSVC batch file (still better than having to perform a dummy link with `gcc -v` and parsing the output). `-link-internally` works on Windows out of the box.]

kinke:
How urgent is the LLVM 6 support? I hope we'll be able to provide a LDC 1.9 beta1 in about 2 weeks and the final in ~4.

kalev:
I was hoping to be able to switch Fedora 28's LDC to LLVM 6 in time for the Fedora 28 final freeze, which is on April 15. Right now it's stuck with LLVM 4 due to various ppc64el bugs in LLVM 5 (https://github.com/ldc-developers/ldc/issues/2356) and now Fedora LLVM people want to remove LLVM 4 from Fedora 28 (https://src.fedoraproject.org/rpms/ldc/pull-request/1).

kalev:
Ah, awesome, let me try that. Thanks @kinke!

kinke:
[Oh and you could also achieve the latter by simply adding a new default switch `-dwarf-version=3` in the ldc2.conf file.]

PR #2617, opened 2018-03-14
[ltsmaster] Use doubledouble as real type on PPC.

redstar:
The PPC platform uses the special doubledouble type for real numbers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,10 +169,6 @@ if( CMAKE_COMPILER_IS_GNUCXX
     AND CMAKE_C_COMPILER_VERSION VERSION_LESS "4.6.0" )
     append("-mminimal-toc" DMD_CXXFLAGS LDC_CXXFLAGS)
 endif()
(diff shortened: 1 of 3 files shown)


kinke:
Do we really want that? I thought it isn't implemented in hardware and so no match for `real` semantics.

redstar:
We should support it because it is standard on ppc/ppc64/ppc64le platforms. Otherwise we force the user to recompile libraries using `long double`. This includes current LLVM for example.
Well, there was some hardware support in the past (but it is droppped now). It is not an floating emulation, just a special representation using two doubles. I would stretch the definition of hardware support to include this.

kalev:
Looks like ppc64le has new 128-bit IEEE long double ABI in the works. https://fedoraproject.org/wiki/Changes/PPC64LE_Float128_Transition

kinke:
@ximion: Ah sorry, it's just a PR, this one. Is Debian also switching to this quadruple-precision ABI/has it already switched?

PR #2623, opened 2018-03-24
Add Nix/NixOS install commands to the install instructions

wilzbach:
See also: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/ldc/default.nix

CC @ThomasMader


wilzbach:
BTW there's also this nice LDC Docker package which might be worth mentioning:

https://hub.docker.com/r/dlanguage/ldc/

thewilsonator:
Thanks!

PR #2624, opened 2018-03-24
Phobos: Boost some transcendental math functions for float/double

kinke:
Based on upstream PR https://github.com/dlang/phobos/pull/6272.

### DIFF
--- a/dmd/builtin.d
+++ b/dmd/builtin.d
@@ -602,8 +602,10 @@ else
     add_builtin("_D3std4math5expm1FNaNbNiNeeZe", &eval_unimp);
     add_builtin("_D3std4math4exp21FNaNbNiNeeZe", &eval_unimp);
     // @safe @nogc pure nothrow double function(double)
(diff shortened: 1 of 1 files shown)


kinke:
lit-test `tests/compilable/ctfe_math.d` fails because the newly introduced, statically unrolled `poly()` version isn't CTFE-able (I guess due to passing a static immutable static array by ref).

kinke:
That additional `poly()` version doesn't affect the optimized LDC timings for the transcendental math functions, i.e., being in the same module, the existing slice-version is apparently inlined and unrolled as it should.
So in theory I could just add that additional version for DMD; the proper way to fix the failure is probably to wait for https://github.com/dlang/dmd/pull/8071#discussion_r176926607 and then add the new float/double versions as CTFE builtins too.

kinke:
A (slightly updated) version of this is incorporated in branch `ldc-merge-2.082`, as the upstream pull was merged just in time.

PR #2625, opened 2018-03-25
[CMake] Future-proof detection of Clang compiler on macOS [NFC]

JohanEngelen:
See:
- https://cmake.org/cmake/help/v3.0/policy/CMP0025.html#policy:CMP0025
- https://stackoverflow.com/a/10055571/3215806


kinke:
There are 4 more occurrences.

JohanEngelen:
Wow, how did I miss them.

Issue #2613, opened 2018-03-13
Wrong arguments for variadic function


belka-ew:
The code is a bit long, it depends on `undead`'s socket streams, but I'll point out where the problem comes from.

ServerTest.d:
```d
import core.thread;
import std.conv;
import std.socket;
import std.stdio;
import undead.socketstream;
import undead.stream;

void write2out(OutputStream output)
{
    output.writef("%s %s\r\n", "HTTP/1.1", "OK");
    output.writeString("\r\n");
    output.flush;
}

void main()
{
    Socket server = new TcpSocket;
    server.setOption(SocketOptionLevel.SOCKET, SocketOption.REUSEADDR, true);
    server.bind(new InternetAddress("localhost", InternetAddress.PORT_ANY));

    InternetAddress address = new InternetAddress("localhost", (cast(std.socket.InternetAddress) server.localAddress).port);

    server.listen(32);

    Socket client = new TcpSocket;

    void run() /** Start */
    {
        while (true)
        {
            try
            {
                auto socket = server.accept;

                scope (exit)
                {
                    socket.shutdown(SocketShutdown.BOTH);
                    socket.close;
                }

                auto output = new SocketStream(socket, FileMode.Out);

                write2out(output);
            }
            catch (SocketAcceptException)
            {
                continue;
            }
            return;
        }
    }

    new Thread(&run).start;
    Thread.sleep(100.msecs);
    /** Start */

    client.connect(address);

    auto output = new SocketStream(client);

    // when
    output.writeString("POST / HTTP/1.1\r\nContent-Length: 9\r\n\r\n<PACKET/>");
    output.flush;

    auto input = new SocketStream(client);
    char[50] line;
    size_t i = 0;
    char c;

    while (c != '\r')
    {

        input.read(c);

        line[i] = c;
        ++i;
    }
    writeln(line[0 .. i]);
}
```
then:
```shell
git clone https://github.com/dlang/undeaD.git
cd undeaD
git checkout 14540c73f8116fe0dd6e98fdedc03e3af2678c74

dub build --compiler=../ldc2-1.8.0-linux-x86_64/bin/ldc2
cd ..
./ldc2-1.8.0-linux-x86_64/bin/ldc2 -I undeaD/src undeaD/bin/libundead.a ServerTest.d
./ServerTest
```
Output:
> HTTP/1.1OK

The same procedure with dmd will result in (expected):
> HTTP/1.1 OK

[...]

PR #2630, opened 2018-03-27
Make interface thunks forward variadic args

kinke:
This fixes issue #2613.

### DIFF
--- a/ir/irclass.cpp
+++ b/ir/irclass.cpp
@@ -415,7 +415,8 @@ llvm::GlobalVariable *IrAggr::getInterfaceVtbl(BaseClass *b, bool new_instance,
       // call the real vtbl function.
       llvm::CallInst *call = gIR->ir->CreateCall(callee, args);
       call->setCallingConv(irFunc->getCallingConv());
-      call->setTailCallKind(llvm::CallInst::TCK_Tail);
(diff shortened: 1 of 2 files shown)


ir/irclass.cpp
```diff
@@ -415,7 +415,8 @@ llvm::GlobalVariable *IrAggr::getInterfaceVtbl(BaseClass *b, bool new_instance,
       // call the real vtbl function.
       llvm::CallInst *call = gIR->ir->CreateCall(callee, args);
       call->setCallingConv(irFunc->getCallingConv());
-      call->setTailCallKind(llvm::CallInst::TCK_Tail);
+      call->setTailCallKind(thunk->isVarArg() ? llvm::CallInst::TCK_MustTail
+                                              : llvm::CallInst::TCK_Tail);
```
JohanEngelen: is there a reason why this shouldn't always be a tail call?

kinke: I got Phobos compile errors (std.regex) when always using must-tail, so I switched to this without further investigation.

JohanEngelen: Hmm ok. LGTM otherwise. Perhaps we can investigate at some later moment

JohanEngelen: Looks like `musttail` is causing problems on PowerPC: https://github.com/ldc-developers/ldc/issues/2686

PR #2632, opened 2018-03-30
Lit tests that use DMD-style asm should explicitly target a supporting target (X86)

JohanEngelen:
This should fix these tests for e.g. arm platforms, see failure on armv7 platform : http://buildbot.ldc-developers.org/builders/armv7neon_builder/builds/289/steps/testlit/logs/stdio


thewilsonator:
LGTM.

PR #2633, opened 2018-03-30
Clarify that error messages relate only to DMD-style asm.

JohanEngelen:
Inspired by the initially surprising error message reported on ARM:
`......./semantic/asm_datadefinition_diag.d(10): Error: inline asm is not supported for the "armv7a" architecture`

Shall I enhance the first error message with a hint towards `ldc.llvmasm.__asm`?
`"DMD-style inline asm is not supported for the \"%s\" architecture. Instead, use ldc.llvmasm.__asm."

### DIFF
--- a/gen/asmstmt.cpp
+++ b/gen/asmstmt.cpp
@@ -109,13 +109,13 @@ Statement *asmSemantic(AsmStatement *s, Scope *sc) {
   if (!(t.getArch() == llvm::Triple::x86 ||
         t.getArch() == llvm::Triple::x86_64)) {
     s->error(
-        "DMD-style inline asm is not supported for the \"%s\" architecture",
+        "the `asm` statement is not supported for the \"%s\" architecture, use `ldc.llvmasm.__asm` instead",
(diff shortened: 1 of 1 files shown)


thewilsonator:
Yes, If i were wanting to use assembly on e.g. ARM and I got that message I would be very confused if I didn't know what DMD-style asm was and why I couldn't use it.

JohanEngelen:
Perhaps "DMD-style" is also not the correct wording, because what is meant is "the inline assembly `asm` statement according to the D language spec".  Ideas about how to write that concisely?

thewilsonator:
perhaps just use `ldc.llvmasm.__asm` for assembly on target X?

JohanEngelen:
`"'asm' statement is not supported for the "%s" architecture. Instead, use 'ldc.llvmasm.__asm'."`

thewilsonator:
Sounds good.

kinke:
[The line info is/should be attached, so it should be pretty clear what the fuss is about.] Use backticks for code in error msges.

PR #2637, opened 2018-03-30
Fix array_equals_memcmp_dyn test.

JohanEngelen:
The frontend now does most of the dyn-dyn comparison optimization (turning it into a druntime template function call).

Caused a failure on PPC: http://buildbot.ldc-developers.org/builders/powerosl_builder/builds/1080/steps/check/logs/stdio


JohanEngelen:
note to self: test lit test with `i686-pc-windows-msvc`

PR #2640, opened 2018-03-31
Ship with LTO default libs on Linux & macOS

kinke:
This allows additional bitcode libs to be generated simultaneously with the normal druntime/Phobos libs (i.e., single compilation) by including something like `-DBUILD_LTO_LIBS=ON -DD_FLAGS="-w;-flto=thin"` in the CMake command line (additional D_FLAGS entry here to include the ThinLTO module summaries in the bitcode files; not needed for FullLTO).

This then allows linking against the LTO default libs via `ldc2 -flto=<thin|full> ... -defaultlib=phobos2-ldc-lto,druntime-ldc-lto`; unfortunately, I couldn't test this yet, as my LLVM 6 Linux build has a 0-byte LLVMgold.so...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -169,9 +169,9 @@ commonSteps: &commonSteps
           # TODO: enable for OSX too; linker currently crashing with:
           #       Assertion failed: (WasErased && "Expected to drop a reference")
           if [ "$CI_OS" = "linux" ]; then
-            $LDC_INSTALL_DIR/bin/ldc2 hello.d -of=hello_thin -flto=thin -defaultlib=phobos2-ldc-bc,druntime-ldc-bc
+            $LDC_INSTALL_DIR/bin/ldc2 hello.d -of=hello_thin -flto=thin -defaultlib=phobos2-ldc-lto,druntime-ldc-lto
             ./hello_thin
-            $LDC_INSTALL_DIR/bin/ldc2 hello.d -of=hello_full -flto=full -defaultlib=phobos2-ldc-bc,druntime-ldc-bc
+            $LDC_INSTALL_DIR/bin/ldc2 hello.d -of=hello_full -flto=full -defaultlib=phobos2-ldc-lto,druntime-ldc-lto
             ./hello_full
           fi
     - run:
@@ -257,7 +257,7 @@ jobs:
       - CI_OS: linux
       - LLVM_VERSION: 6.0.0
       - HOST_LDC_VERSION: 1.8.0
-      - EXTRA_CMAKE_FLAGS: "-DMULTILIB=ON -DBUILD_BC_LIBS=ON -DD_FLAGS='-w;-flto=thin' -DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++ -DLDC_INSTALL_LTOPLUGIN=ON -DLDC_INSTALL_LLVM_RUNTIME_LIBS=ON"
+      - EXTRA_CMAKE_FLAGS: "-DMULTILIB=ON -DBUILD_LTO_LIBS=ON -DD_FLAGS='-w;-flto=thin' -DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++ -DLDC_INSTALL_LTOPLUGIN=ON -DLDC_INSTALL_LLVM_RUNTIME_LIBS=ON"
       - DUB_VERSION: v1.7.2
   build-osx:
     <<: *commonSteps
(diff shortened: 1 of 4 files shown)


kinke:
Tested it now; it's principally working, but the C and asm parts of druntime/Phobos are missing in the bitcode libs. We don't have these object files explicitly, CMake takes care of compiling them, so that's blocking this LTO-libs-for-free scheme.

kinke:
Working now. A Phobos-hello-world can be linked against the normal libs as well as with both thin & full LTO against the ThinLTO bitcode libs (incl. C/ASM parts in machine code):

```D
void main()
{
    import std.stdio;
    writefln("Hello world, %d bits", cast(int) size_t.sizeof * 8);
}
```

```
bin/ldc2 -O -release ../hello.d -of=hello_normal -linker=gold
  => 833 KB, 1.6 secs
bin/ldc2 -O -release ../hello.d -of=hello_full -flto=full -defaultlib=phobos2-ldc-bc,druntime-ldc-bc -linker=gold
  => 691 KB, 6.9 secs
bin/ldc2 -O -release ../hello.d -of=hello_thin -flto=thin -defaultlib=phobos2-ldc-bc,druntime-ldc-bc -linker=gold
  => 690 KB, 6.2 secs
```

kinke:
I'd prefer an `-lto` suffix instead of `-bc`; I don't think there's much room for backwards compatibility concerns (considering that BUILD_BC_LIBS=ON was broken for default BUILD_SHARED_LIBS=BOTH etc.).

.circleci/config.yml
```diff
@@ -258,7 +270,7 @@ jobs:
       - LLVM_VERSION: 5.0.1
       - HOST_LDC_VERSION: 1.6.0
       - BOOTSTRAP_CMAKE_FLAGS: "-DCMAKE_CXX_FLAGS='-stdlib=libc++' -DCMAKE_EXE_LINKER_FLAGS=-lc++"
-      - EXTRA_CMAKE_FLAGS: "-DMULTILIB=ON -DCMAKE_CXX_FLAGS='-stdlib=libc++' -DCMAKE_EXE_LINKER_FLAGS=-lc++"
+      - EXTRA_CMAKE_FLAGS: "-DMULTILIB=ON -DBUILD_BC_LIBS=ON -DD_FLAGS='-w;-flto=thin' -DCMAKE_CXX_FLAGS='-stdlib=libc++' -DCMAKE_EXE_LINKER_FLAGS=-lc++"
```
JohanEngelen: Adding `-flto` to `D_FLAGS` also builds the normal libs with that option, right? But then how do the tests pass: how does linking succeed without specifying the LLVMgold plugin on the linker line for normal tests without LTO ?

JohanEngelen:
This is pretty neat.

> I'd prefer an -lto suffix instead of -bc (and BUILD_LTO_LIBS over BUILD_BC_LIBS)

Me too.

.circleci/config.yml
kinke: > Adding -flto to D_FLAGS also builds the normal libs with that option, right?

In combination with `BUILD_BC_LIBS=ON`, this yields a command line `ldc2 -c -output-o -output-bc -flto=thin ...`; producing both the normal object file in machine code and the LTO bitcode file (`-flto=thin` only affecting the module summaries in the bitcode files in this case).

> how does linking succeed without specifying the LLVMgold plugin

We use LDC to firstly compile the modules, one object (and bitcode file for BUILD_BC_LIBS) per druntime/Phobos module. We then archive them to static libs via separate LDC invokations - normal libs containing machine code only, and bitcode/LTO libs containing bitcode for the D parts and machine code the C/ASM parts. The shared libs are obviously only generated for the normal machine-code objects.

So specifying `-flto=...` for linking is only required when choosing the bitcode/LTO libs explicitly via `-defaultlib=...`, which is only tested by the CircleCI integration test so far (Linux only; linker crashing on OSX for both thin and full LTO tests).

kinke: This scheme allows the user to choose LTO for his code only via `-flto=...` and LTO incl. default libs via `-flto=... -defaultlib=...`.

JohanEngelen: > producing both the normal object file in machine code and the LTO bitcode file

OK got it.

kinke:
Btw the CircleCI package sizes are increased by about 10%.

kinke:
Oh wow, the CircleCI macOS LTO integration tests now work with LLVM 6. :)

JohanEngelen:
> Oh wow, the CircleCI macOS LTO integration tests now work with LLVM 6. :)

There was an XCode update recently, perhaps related?

kinke:
I don't think so, we're using the XCode 9.0 CircleCI image.

PR #2641, opened 2018-04-01
[2.079] Support minimal runtime

kinke:
Thereby deferring requirements wrt. object.d declarations (Object, Throwable, TypeInfo, TypeInfo_Class, TypeInfo_Struct...) and paving the way for a minimal druntime.

This also greatly reduces the `-vv` output for tiny code samples, e.g., reduced test cases.

### DIFF
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -177,12 +177,6 @@ void emitLLVMUsedArray(IRState &irs) {
 namespace ldc {
 CodeGenerator::CodeGenerator(llvm::LLVMContext &context, bool singleObj)
     : context_(context), moduleCount_(0), singleObj_(singleObj), ir_(nullptr) {
-  if (!ClassDeclaration::object) {
-    error(Loc(), "declaration for class `Object` not found; druntime not "
-                 "configured properly");
-    fatal();
-  }
-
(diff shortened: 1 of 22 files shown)


kinke:
Wrt. runtime module, I wasn't sure whether it was just exploited as stringmap or whether caching the functions there brings some performance. Right now, the lazy 'declarers' may declare more than one function (same type, multiple names), so switching this exact scheme to forward-declaring directly into the desired module may yield a few superfluous declarations.

kinke:
Yep, the loc would be nice to have of course. Given the elision of Module-/TypeInfos if there's no declaration, the chances of a missing TypeInfo should be pretty much restricted to a runtime function call requiring one of the few TypeInfo base types. As such, the existing loc param of the `getRuntimeFunction()` should do fine, it'll just have to be forwarded to the LazyType error msg.

---

merge-2.079 green on Win64 now with this. :)

kinke:
Well actually it'd probably be useful to do the runtime function lookup as early as possible. In various places, we first construct the args for the runtime function call (incl. TypeInfos) and afterwards do `getRuntimeFunction()` immediately before the call; the loc will be missed in thes cases.

kinke:
Expecting Win32 to succeed too now and so AppVeyor to become green.

PR #2642, opened 2018-04-02
[2.079] Call _Unwind_Resume() directly (except for ARM EABI)

kinke:
This fixes the 32-bit std.file unittests on Linux with enabled optimizations.

The remaining LDC-specific ARM EABI wrapper (`_d_eh_resume_unwind()`) is implemented in assembly (ldc/eh_asm.S), preserving registers apparently.


kinke:
There's this note in the SysV ABI docs wrt. `_Unwind_Resume()`:
> This is the only routine in the unwind library which is expected to be
> called directly by generated code: it will be called at the end of a
> landing pad in a "landing-pad" model.

PR #2647, opened 2018-04-07
[2.079] Only emit interface vtables in the declaring module

kinke:


kinke:
Fixes the druntime compilation errors on OSX.

On 64-bit OSX, the C++ mangling of 64-bit integers still needs to be figured out; the 5th commit was required to get rid of linking errors for the `all` target, but there are still related failures for dmd-testsuite (`runnable/{cppa,externmangle}.d`). All other 2.079 tests green on OSX (except for the minimal runtime ones fixed by #2641).

kinke:
OSX mangling should be fixed now, incl. self-compilability of 2.079. The breaking change affects `size_t` (and `ptrdiff_t`) too, as it's defined as alias to `u{int,long}` in object.d; this means that a 64-bit D `size_t` is now mangled as C++ `uint64_t` (`unsigned long long`), which on 64-bit OSX doesn't correspond to C++ `size_t` anymore (apparently a `unsigned long`).
Luckily, we apparently only use very few cross-language calls with `size_t` args, so adapting the type in the C++ headers wasn't hard, but there are many more C++ function declarations which would actually need to be changed from `size_t` to `d_size_t` to fix all potential linking errors.

Additionally, it looks as if a C++ `size_t` cannot be represented in D on 64-bit OSX at all anymore. One would think that `core.stdc.config.c_ulong` would fit the bill, but I think it doesn't: https://github.com/ldc-developers/druntime/blob/ldc-merge-2.079/src/core/stdc/config.d#L67-L71

kinke:
I'm not the only one finding the mangling change at least problematic:
https://github.com/dlang/dmd/pull/7523
https://github.com/dlang/dmd/pull/7416

kinke:
Absolutely. Even when fixing `core.stdc.config` by providing `cpp_[u]long` and `cpp_{size,ptrdiff}_t`, I think having a D size_t naturally C++-mangled as the corresponding C++ size_t is the most important thing for easy-peasy D <--> C++ interop, as a size_t should be much more frequently used than explicit 64-bit integer types.

kinke:
The test confirms the regression on 64-bit macOS. It also shows that 32-bit `size_t` C++ mangling has already been broken for 32-bit macOS (D (size_t being an alias for uint): `unsigned int` vs. C++ `unsigned long` apparently).

JohanEngelen:
We are already ABI incompatible with DMD and GDC, iirc, so I'm fine with fixing the mangling ourselves if upstream dfrontend doesn't want to.

kinke:
Walter doesn't seem to acknowledge the `size_t` interop problem and is going for a `core.stdc.config.cpp_size_t` alias (at least getting rid of the separate implicit conversion issue of the magic structs) in https://github.com/dlang/druntime/pull/2160.

I reverted the mangling and added aliases `c(pp)_(u)longlong` and `c(pp)_(u)int64_t`. 32-bit macOS aside, this makes sure D/C++ `size_t` still match on all tested platforms. For (the in my experience way less often used) C++ `(u)int64_t`, one needs to resort to `c(pp)_(u)int64_t` for portable code (it's D `(u)long` everywhere except for 64-bit macOS). Any opinions @klickverbot @redstar?

[I'll split off the mangling things as separate PR. I just abused this PR here as the interface vtable emission fix is a prerequisite for running the CI tests on macOS.]

kinke:
> So DMD's core.stdc.config.cpp_size_t alias does correctly mangle to C++'s size_t on all platforms?

Yep it will.

> I thought it had become completely impossible to write a D function signature that would mangle to a C++ function with size_t

That's the status quo for DMD 2.079.0 on 64-bit macOS (unless you roll your own magic `__c_ulong` struct).

Issue #2646, opened 2018-04-06
Allow the user to overwrite a global `-link-defaultlib-shared` flag


wilzbach:
Now that OS packers are starting to add the new `-link-defaultlib-shared` (see https://github.com/ldc-developers/ldc/pull/2443) to their global `ldc.conf` file, this option breaks release build scripts somehow.
What do you think about:

1) Adding a CLI flag to negate this, e.g. `-disable-link-defaultlib-shared`? (the last seen option would take precedence)
2) Let `-static` overwrite the global `-link-defaultlib-shared` preference? At the moment one just gets an error:

> Error: Can't use -link-defaultlib-shared and -static together

wilzbach:
> -link-defaultlib-shared=false is working as expected, last seen option takes precedence as usual.

Great! Just missing documentation then?

```
> ldc -help
...
Linking options:

  -L=<linkerflag>                         - Pass <linkerflag> to the linker
  -Xcc=<ccflag>                           - Pass <ccflag> to GCC/Clang for linking
  -defaultlib=<lib1,lib2,...>             - Default libraries to link with (overrides previous)
  -disable-linker-strip-dead              - Do not try to remove unused symbols during linking
  -link-defaultlib-debug                  - Link with debug versions of default libraries
  -link-defaultlib-shared                 - Link with shared versions of default libraries
  -linker=<lld-link|lld|gold|bfd|...>     - Linker to use
  -mscrtlib=<libcmt[d]|msvcrt[d]>         - MS C runtime library to link with
  -static                                 - Create a statically linked binary, including all system dependencies
```

Maybe the boolean flags could be marked with e.g. `-link-defaultlib-shared=<true,false>` somehow?
Not sure what the standard on this is - DMD would do this
```
-link-defaultlib-shared
-link-defaultlib-shared=<on,off>
```

(e.g.`-color` -> https://dlang.org/dmd#switch-color)

BTW is there any reason why `ldc` without args and `ldc -help` print different pages?

kinke:
> Maybe the boolean flags could be marked with e.g. -link-defaultlib-shared=<true,false> somehow?

It's just a default LLVM Boolean commandline option with default help formatting, which doesn't mark the Boolean options in a special way. `-static` is a bool too, as most other options are, so `-static[=<true|TRUE|True|1|false|FALSE|False|0>]` are all valid cmdline args.

> BTW is there any reason why ldc without args and ldc -help print different pages?

[...]

PR #2649, opened 2018-04-10
Enforce static default libs with cmdline option -static

kinke:
Wrt. issue #2646.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -997,7 +997,7 @@ int cppmain(int argc, char **argv) {
   }
 
   if (files.dim == 0) {
-    cl::PrintHelpMessage();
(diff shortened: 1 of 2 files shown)


JohanEngelen:
No testcase?

kinke:
It's not [trivially] testable for BUILD_SHARED_LIBS=ON configs, as used for Travis.

JohanEngelen:
Can't we do a `not %ldc -v -static -link-defaultlib-shared  %s | FileCheck %s` and check the `-v` output?  (with requires: Linux)
(add a reference to a missing symbol in the D source such that the linking will always error)

kinke:
Oh I didn't think about enforcing a linking failure. - Test amended.

PR #2650, opened 2018-04-11
[2.079] WIP: 64-bit macOS (u)long mangling

kinke:
Previous discussion in https://github.com/ldc-developers/ldc/pull/2647.


kinke:
I propose the following: we keep the old mangling scheme for LDC 1.9 and introduce the change when it's done properly upstream, presumably in the next version.

The migration from magic structs to special enums (`enum __c_long : long { dummy }`; still no `__c_longlong` support), solving the implicit conversion problem, is [already](https://github.com/dlang/dmd/pull/8152) in dmd master, incl. tests. The druntime changes are [still pending](https://github.com/dlang/druntime/pull/2160) (still lacking `c(pp)_(u)longlong`).

kinke:
Merging in order to upgrade to 2.079.1; we can always revert to something else as this PR covers both options.

PR #2651, opened 2018-04-12
[2.079] MSVC: Support exporting naked functions

### LINKED ISSUE #2648: Naked functions not being exported on Windows
Creating DLLs on Windows with functions that use the naked attribute results in export being ignored.
Example code:
```D
import core.runtime;
import core.sys.windows.windows;
import core.sys.windows.dll;

extern(C) export void TestFunc(){
	asm{
		naked;
		mov EAX, 42;
	}
}

extern (Windows)
BOOL DllMain( HINSTANCE hInstance, ULONG ulReason, LPVOID pvReserved) {
	switch (ulReason) {
		default: break;
	
		case DLL_PROCESS_ATTACH:
			dll_process_attach( hInstance, true );
			TestFunc();
		break;

		case DLL_PROCESS_DETACH: 
			dll_process_detach( hInstance, true );
			break;
	}
	return true;
}
```

kinke:
This fixes #2648.

### DIFF
--- a/gen/naked.cpp
+++ b/gen/naked.cpp
@@ -247,6 +247,15 @@ void DtoDefineNakedFunction(FuncDeclaration *fd) {
   gIR->module.appendModuleInlineAsm(asmstr.str());
   asmstr.str("");
 
+  if (global.params.targetTriple->isWindowsMSVCEnvironment() &&
+      fd->isExport()) {
+    // Embed a linker switch telling the MS linker to export the naked function.
+    // This mimics the effect of the dllexport attribute for regular functions.
+    const auto linkerSwitch = std::string("/EXPORT:") + mangle;
+    auto Value = llvm::MDString::get(gIR->context(), linkerSwitch);
+    gIR->LinkerMetadataArgs.push_back(llvm::MDNode::get(gIR->context(), Value));
+  }
+
   gIR->funcGenStates.pop_back();
 }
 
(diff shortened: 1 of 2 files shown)


gen/naked.cpp
```diff
@@ -249,6 +249,15 @@ void DtoDefineNakedFunction(FuncDeclaration *fd) {
   gIR->module.appendModuleInlineAsm(asmstr.str());
   asmstr.str("");
 
+  if (global.params.targetTriple->isWindowsMSVCEnvironment() &&
+      fd->isExport()) {
+    // Embed a linker switch telling the MS linker to export the naked function.
+    // This mimicks the effect of the dllexport attribute for regular functions.
+    const auto linkerSwitch = std::string("/EXPORT:") + mangle;
+    auto Value = llvm::MDString::get(gIR->context(), linkerSwitch);
```
JohanEngelen: If you want, this looks like the linkerSwitch alloc can be optimized with a Twine.

JohanEngelen:
Cool. LGTM.
(the x86 test failure is trivial)

gen/naked.cpp
kinke: That was actually my first version (`get(gIR->context(), llvm::Twine("/EXPORT") + mangle)`), but it takes a `llvm::StringRef`.

kinke:
With 2.079, the test would pass on Win32. I'll probably just wait for 2.079 to be merged instead of adapting this twice.

PR #2655, opened 2018-04-15
Try to fix Travis OSX regression wrt. pip installation

kinke:


kinke:
Erm, seems like the Travis guys fixed it in the meantime.

PR #2665, opened 2018-04-24
Upgrade to D v2.080.0

kinke:


kinke:
This may be one of the smoothest transitions ever - it's apparently pretty much fully working already.

kinke:
Btw, what's the next LDC version? 1.10, 2.0, 2.080? ;)

kinke:
Should be green, except for a failing std.random unittest assertion on Win64 with enabled optimizations.

joakim-noah:
I like 2.080. I see no reason not to sync with dmd, less confusing for a casual user.

kinke:
I like 2.080 too; it should make things clearer for everyone.

---

The Win64 issue may be an LLVM 6 optimizer problem. It fails [here](https://github.com/ldc-developers/phobos/blob/ldc-merge-2.080/std/random.d#L1382). I added some printfs before the assertion, and the values are fine; the relevant IR before optimization looks fine as well; with optimizations, it's fully inlined into the `std.random` module unittest function.

While checking the IR, I noticed an ugly inefficiency wrt. array equality comparisons - when comparing a `uint[4]` against an int array literal, the literal is allocated by the GC and the comparison performed by `_ArrayEq()`. If the array literal's element type matches (uint), the literal is allocated on the stack and the comparison performed via a `memcmp()`. It's a front-end thing. See https://run.dlang.io/is/xh0rTI.
[When using a matching uint array literal, the test succeeds on Win64.]

joakim-noah:
Tried this branch out natively on Android/ARM, handful of `std.math` tests fail because they try to use `ieeeFlags` on soft-float and `setControlState` needs to be annotated as `pure`. Otherwise, stdlib tests are good, will upstream a pull for `std.math`. Dmd testsuite also looking good, pretty much same results as 1.9.

kinke:
Hmm, but your Android build is for ARMv7+ and so should be using SoftFP, right? Shouldn't the IEEE flags be usable then?

I looked up the definition of `version(D_SoftFloat)`
> The target hardware does not have a floating point unit

LDC sets it for SoftFP too and so seems to be wrong in this regard.

joakim-noah:
Yeah, I was planning on trying it out, just left it as `ARM_SoftFloat` all this time because that's how it was set and I didn't care how it was done. Regardless, those tests will need to be elided for actual soft-float platforms, but I'll try using the hard-float path on Android/ARM and see how it does.

PR #2668, opened 2018-04-26
Fix building with LLVM trunk

JohanEngelen:


driver/cl_options-llvm.cpp
```diff
@@ -1,6 +1,6 @@
 //===-- cl_options-llvm.cpp -----------------------------------------------===//
 //
-//                         LDC � the LLVM D compiler
+//                         LDC - the LLVM D compiler
```
JohanEngelen: this is a char encoding change, to make clang-format work.

kinke: Different char, not just different encoding, right? We use the longer dash everywhere else, so I'm wondering why it's necessary.

JohanEngelen: yes indeed. Replaced it with the usual dash now.

PR #2678, opened 2018-05-04
Predefine version D_HardFloat instead of D_SoftFloat for -float-abi=softfp

kinke:
Matching the `D_HardFloat` [semantics](https://dlang.org/spec/version.html#predefined-versions):
> The target hardware has a floating point unit

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -618,21 +618,22 @@ void fixupUClibcEnv()
 /// Also defines D_HardFloat or D_SoftFloat depending if FPU should be used
 void registerPredefinedFloatABI(const char *soft, const char *hard,
                                 const char *softfp = nullptr) {
-  // Use target floating point unit instead of s/w float routines
-  // FIXME: This is a semantic change!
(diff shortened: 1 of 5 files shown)


JohanEngelen:
Tests are missing.

joakim-noah:
Ran the stdlib unit tests and the dmd testsuite natively on Android/ARM with this pull, pretty much all the same tests passed, except no problems in `std.math` anymore. That's the only Phobos module that actually uses these predefined versions and it was using the FPU fine before, but these versions were just set to `SoftFloat`. Didn't really matter as long as most of the tests weren't exercising `IeeeFlags`, which has now changed with the tests Jack added.

If you want to test for it, the ARM backend sets the CPU to `generic` by default, which will set the soft-float path, but `-mcpu=cortex-a8` will invoke soft-fp with the FPU. If you want to add a test for this pull, you could append that flag and check to see if the predefined version is set, though I don't know if there's some way for tests to check for that, other than grepping the log output (which is what I did to make sure this pull was working).

kinke:
Thanks for testing. I added a test which I hope doesn't depend on `target_ARM` (via `-o-`).

PR #2681, opened 2018-05-05
RestrictAll UDA

mabau:
- added @restrictAll UDA for functions that marks all pointer parameters
  in LLVM as noalias
- analog to restrict keyword in C

needs changes in druntime that come with this  [pull request]( https://github.com/ldc-developers/druntime/pull/137)

### DIFF
--- a/tests/codegen/attr_restrict.d
+++ b/tests/codegen/attr_restrict.d
@@ -1,0 +1,45 @@
+// Test ldc.attributes.restrict
+
+// RUN: %ldc -O3 -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+import ldc.attributes;
+
+
+extern (C) int testReturnConstant( int * p1, int * p2) @restrictAll
+// CHECK-LABEL: define{{.*}}@{{.*}}testReturnConstant
+// CHECK: noalias
+// CHECK: ret i32 42
+{
+    *p1 = 42;
+    *p2 = 2;
+    return *p1; //has to return 42 when p1 does not alias p2
+}
+
+extern (C) int testWithoutRestrict( int * p1, int * p2)
+// CHECK-LABEL: define{{.*}}@{{.*}}testWithoutRestrict
+// CHECK-NOT: noalias
+// CHECK: ret i32 %
+{
+    *p1 = 42;
+    *p2 = 2;
+    return *p1; //here the constant return is not possible
+}
+
+extern (C) void testDoubleApplication( int * p1, int * p2) @(restrictAll, restrictAll)
+// CHECK-LABEL: define{{.*}}@{{.*}}testDoubleApplication
+// CHECK: noalias
+{
+    *p1 = 1;
+    *p2 = 2;
+}
+
+@restrictAll
+{
+    // CHECK-LABEL: define{{.*}}@{{.*}}testNoPointers
+    // CHECK-NOT: noalias
+    // CHECK: ret
+    extern (C) int testNoPointers(int a, int b) @restrictAll
+    {
+        return a + b;
+    }
+}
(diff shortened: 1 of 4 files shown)


kinke:
Nice, thanks. We'll probably have some bikeshedding about the UDA name (`@no[Pointer]Aliases` etc.). It also doesn't work for slices, as most if not all ABIs pass those as IR structs, and I think that it'd be especially useful for slices.

mabau:
Great :) I have no particular preference regarding the UDA name.. what do you prefer? `@noPointerAliases` Or `@noAlias`?

tests/codegen/attr_restrict.d
```diff
@@ -0,0 +1,45 @@
+// Test ldc.attributes.restrict
+
+// RUN: %ldc -O3 -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+import ldc.attributes;
+
+
+extern (C) int testReturnConstant( int * p1, int * p2) @restrictAll
+// CHECK-LABEL: define{{.*}}@{{.*}}testReturnConstant
+// CHECK: noalias
```
JohanEngelen: you can do a `CHECK-SAME` to check that noalias is on the same line as the previous `CHECK`

kinke:
Let's move forward; I assume this is mostly interesting for a few hot key functions, so not supporting slices directly can always be worked around via a nested function taking pointers.
The current exact semantics are presumably difficult for people not dealing with low-level stuff and/or not used to LLVM IR. I think we can use an experimental/unfinished `@noAliases` UDA (with final name) for the time being and revise the semantics/support slices at a later stage. Other opinions?

JohanEngelen:
Definitely move forward. There is no adverse effect that it doesn't work with slices.
From the alternatives proposed, I like `@noPointerAliases` best. (although that doesn't _quite_ match the semantics of LLVM's `noalias`)

The lit tests need to be improved to not match on `noalias` on different lines.

kinke:
Wrt. lit tests, I'd like one function to move from `int*` to `ref int[N]`.

JohanEngelen:
> Wrt. lit tests, I'd like one function to move from int* to ref int[N].

Good point. Also, classes.

JohanEngelen:
note that UDAs can now be put on parameters, so we can implement an `@restrict` that is more C-like instead of `restrictAll`?

9il:
Awesome feature! Could you please make it compatible with ndslice?
The simplest example is:
```d
struct Iterator
{
    double* _ptr;
}

struct Vector
{
    size_t _length;
    Iterator _iterator;
}

auto foo(@restrict Vector a, @restrict Vector b)
{
}

// and

auto foo(ref @restrict Vector a, ref @restrict Vector b)
{
}
```

The second one is better for types with automatic reference counting.

9il:
> What would be the exact semantics of `@restrict` on members?

If the argument is a struct/class (optionally with ref qualifier) and with @restrict attribute, then `noalias` attribute should be added for and only for fields that are marked with `@restrict` keyword, recursively. If a member is a pointer or dynamic array, then @restrict attribute adds `noalias` LLVM attribute.

kinke:
I don't think such a scheme is feasible or even needed.

> auto foo(ref @restrict Vector a, ref @restrict Vector b)

By mapping `@restrict` directly to the LLVM noalias parameter attribute, all addresses based on a/b are assumed to be in disjoint a/b spaces while executing the function, incl. members and their members, i.e., `&a`, `&a.length`, `a.iterator._ptr`. It's already as transitive as it gets.

Then we'd just need to be able to bind rvalue args to ref params.

kinke:
`@restrict` parameter UDA has been added in #3043. Thx anyway, @mabau!

Issue #2687, opened 2018-05-06
ARMv7: dynamiccompile failures


JohanEngelen:
@Hardcode84 Could you have a look at the ARMv7 buildbot Lit-test failures related to dynamicCompile? Thanks!
http://buildbot.ldc-developers.org/builders/armv7_builder/builds/3169/steps/testlit/logs/stdio
http://buildbot.ldc-developers.org/builders/armv7_builder/builds/3169

Hardcode84:
@JohanEngelen 

I will add more logs to `asm_output` test to see actual output and probably just mark exception test as XFAIL.

Also, what LLVM version is there?

JohanEngelen:
@Hardcode84 You can check the LLVM version in the "5. check" step. It's LLVM 5.0.

Hardcode84:
@JohanEngelen thanks.

Do we have any other ARM systems where these tests are run (and pass)?

I am going to just write `XFAIL: target_ARM`.

JohanEngelen:
> Do we have any other ARM systems where these tests are run (and pass)?

Not that I know of. :/

Hardcode84:
lit is green now

http://buildbot.ldc-developers.org/builders/armv7_builder/builds/3189

JohanEngelen:
fixed by #2723

PR #2688, opened 2018-05-08
#2687: ARMv7: dynamiccompile failures

Hardcode84:
* Disable jit exceptions test on ARM (I don't have hardware, nor time to debug it right now)
* Dump actual asm output to stdout in `asm_output` test (lit will only display it if test fail, what I can see now from test failure, it should be at least partially correct)

@JohanEngelen

### DIFF
--- a/tests/dynamiccompile/throw.d
+++ b/tests/dynamiccompile/throw.d
@@ -4,7 +4,7 @@
 // XFAIL: Windows
 //
 // Also, some issue on Arm
-// XFAIL: target_ARM
+// XFAIL: host_ARM
 // RUN: %ldc -enable-dynamic-compile -run %s
 
 import std.exception;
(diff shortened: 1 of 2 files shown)


tests/dynamiccompile/throw.d
```diff
@@ -2,6 +2,9 @@
 // exceptions is broken on windows
 // win64 issue https://bugs.llvm.org//show_bug.cgi?id=24233
 // XFAIL: Windows
+//
+// Also, some issue on Arm
+// XFAIL: target_ARM
```
kinke: `target_ARM` means 'my LLVM features an ARM backend'. AFAICT we don't predefine the host architecture for lit. You could disable the test via `version(ARM) { void main() {} } else:` for now.

Hardcode84:
Update: used new `host` lit test feature

JohanEngelen:
LGTM
(you want to keep the stdout output?)

Hardcode84:
> you want to keep the stdout output?

lit print it only when test fails, so I think it's ok.

PR #2689, opened 2018-05-09
[Trivial / ltsmaster] Allow shared library on freebsd/dragonflybsd

dkgroot:
Allow building shared library on freebsd/dragonflybsd (ltsmaster) using:
```
cmake BUILD_SHARED_LIB=ON -DLLVM_CONFIG=[path-to-llvm]/llvm-config ..
```


dkgroot:
Looking at the CI results, none of them seem to be related to the PR, as far as i can tell.

PR #2692, opened 2018-05-09
Allow shared library on dragonflybsd

dkgroot:
Allow shared library on dragonflybsd


runtime/CMakeLists.txt
```diff
@@ -492,6 +492,10 @@ function(set_common_library_properties target name output_dir c_flags ld_flags i
         set_target_properties(${target} PROPERTIES
             POSITION_INDEPENDENT_CODE ON)
     endif()
+    if("${TARGET_SYSTEM}" MATCHES "FreeBSD|DragonFly")
+        set_target_properties(${target} PROPERTIES
+            POSITION_INDEPENDENT_CODE ON)
+    endif()
```
kinke: Is this necessary? LDC doesn't explicitly default to PIC for BSD [but LLVM might].

dkgroot: It's not 100% necessary, but i thought in current day and age might be the prefered default. Not sure.

kinke: The default would be letting CMake choose based on static/shared library, and not enforcing PIC for the C parts of static default libs too as what this is doing.

PR #2693, opened 2018-05-10
CMake: Fix bash-completion issues on Linux

kinke:
1) Allow the user to specify a `BASH_COMPLETION_COMPLETIONSDIR` variable.
2) If not specified explicitly, only use the autodetected dir if the accompanying bash-completion prefix matches `CMAKE_INSTALL_PREFIX`. The autodetected dir seems to always be the system one, so prevent local installs from needing root permissions, by installing the bash-completion files locally into `<CMAKE_INSTALL_PREFIX>/etc/bash_completion.d`.

For reference, here's my `/usr/share/cmake/bash-completion/bash-completion-config.cmake`:
```
set (BASH_COMPLETION_VERSION "2.8")

set (BASH_COMPLETION_PREFIX "/usr")
set (BASH_COMPLETION_COMPATDIR "/etc/bash_completion.d")
set (BASH_COMPLETION_COMPLETIONSDIR "/usr/share/bash-completion/completions")
set (BASH_COMPLETION_HELPERSDIR "/usr/share/bash-completion/helpers")

set (BASH_COMPLETION_FOUND "TRUE")
```


kinke:
The check could be relaxed to checking whether `CMAKE_INSTALL_PREFIX` *begins* with `BASH_COMPLETION_PREFIX($|/)`; that would include default `/usr/local`.

dkgroot:
Why would you confine this check for bash_completion only to the linux platform. Other OS'es can also have bash installed and also support bash_completion. For example FreeBSD / DragonFly (And logically the other BSD's) do happily support both bash and bash_completion.

kinke:
> Why would you confine this check for bash_completion only to the linux platform.

I was wondering about that too shortly, but apparently nobody missed them so far. This PR is about the repeatedly reported installation issues that little and probably quite outdated file causes though.

PR #2694, opened 2018-05-10
Fix inline asm label naming issue in overloaded functions

### LINKED ISSUE #2667: Inline asm label naming issue
ZILtoid1991/CPUblit@fa802a4 doesn't compile with LDC.

The cause is in `src/CPUbilt/composing.d`, which fails with errors like:

```
<inline asm>:6:2: error: unsupported relocation of undefined symbol 'L7CPUblit9composing11blitter8bit_eightpixel'
        jl L7CPUblit9composing11blitter8bit_eightpixel
```

I haven't had a proper look at this yet, but from a quick glance at the `-vv` output, it seems like we are emitting the label – properly! – as `L_D7CPUblit9composing11blitter8bitFNiPhPhPhmZv_eightpixel`, but try to reference it without the full mangle.

This only occurs if there is more than one overload of the function in question.

kinke:
Fixes #2667.

### DIFF
--- a/gen/asm-x86.h
+++ b/gen/asm-x86.h
@@ -3032,7 +3032,7 @@ struct AsmProcessor {
               stmt->error("dollar labels are not supported");
               asmcode->dollarLabel = 1;
             } else if (e->op == TOKdsymbol) {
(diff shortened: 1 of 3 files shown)


JohanEngelen:
lgtm

PR #2695, opened 2018-05-10
Fix internal archiver issue for SemaphoreCI

kinke:
It seems to be hitting non-ENOENT when trying to open a non-existent target library, although the error code's message is `No such file or directory`. - Prefer an `exists()` check instead of relying on an exact error code.


kinke:
This enables upgrading SemaphoreCI's LLVM to 6.0.0 (was still at 4.0.1).

Issue #2278, opened 2017-08-16
ThinLTO lit tests fail with bfd linker


Fatmice:
I built and install llvm 4.0.1 with
`cmake -DCMAKE_INSTALL_PREFIX=../llvm-test -DLLVM_TARGETS_TO_BUILD="X86;AArch64;ARM;PowerPC" -DLLVM_BINUTILS_INCDIR=/usr/include -DCMAKE_BUILD_TYPE=Release -DLLVM_LIBDIR_SUFFIX=64 ../llvm`
Then built and test ldc origin/master with
`cmake -DCMAKE_INSTALL_PREFIX=../ldc-test -DINCLUDE_PATH=import -DCMAKE_BUILD_TYPE=Release -DLIB_SUFFIX=64 -DLDC_INSTALL_LTOPLUGIN=ON -DLDC_INSTALL_LLVM_RUNTIME_LIBS=ON -DLLVM_ROOT_DIR=../llvm-test -DD_COMPILER=../install/ldc-0.17.4-bootstrap/bin/ldmd2 ../ldc`

All tests passed except for two of the lit tests.
linking/thinlto_modulecdtors.d (134 of 153)
linking/thinlto_modulecdtors_2.d (144 of 153)
Here's [complete log](https://gist.githubusercontent.com/Fatmice/ef8dd6750b1a9cb98d03f193f5d96bb1/raw/4d5d1dad97bdc9f7e9a5675d04fa57c8f05cb28f/LastTest.log)

The relevant portions
```
FAIL: LDC :: linking/thinlto_modulecdtors.d (134 of 153)
******************** TEST 'LDC :: linking/thinlto_modulecdtors.d' FAILED ********************
Script:
--
/home/quyen/D/ldc-build/bin/ldc2 -flto=thin -O3 -run /home/quyen/D/ldc/tests/linking/thinlto_modulecdtors.d | FileCheck /home/quyen/D/ldc/tests/linking/thinlto_modulecdtors.d
--
Exit Code: 2

Command Output (stdout):
--
$ "/home/quyen/D/ldc-build/bin/ldc2" "-flto=thin" "-O3" "-run" "/home/quyen/D/ldc/tests/linking/thinlto_modulecdtors.d"
# command stderr:
Error: /tmp/thinlto_modulecdtors-93a9e0a-1f491d failed with status: -2
Error: message: Illegal instruction (core dumped)
Error: program received signal 2 (Interrupt)

error: command failed with exit status: 2
[...]

PR #2696, opened 2018-05-11
SemaphoreCI: Add bash scripts and move to Docker & Ubuntu 18.04

kinke:


kinke:
Well it's not too bad when distributing the now serially compiled defaultlib unittests among both jobs (debug/release); overall wall-time is around 36 minutes.

kinke:
The backtrace issue was a stupid regression depending on length of path to executable.

The ThinLTO failures are #2278 (not requiring `-O` in this case). On Ubuntu 18.04, the bfd linker is used regardless of what the `/usr/bin/ld` symlink points at; `update-alternatives --install /usr/bin/ld ld /usr/bin/ld.gold 99` has no effect when linking via `gcc`. `-linker=gold` works.

PR #2700, opened 2018-05-16
druntime: Fix core.stdc.stdint.(u)int64_t on 64-bit macOS etc.

kinke:
* Use it instead of newly introduced, LDC-specific `core.stdc.config.c(pp)_(u)int64_t`.
* Cherry-pick Walter's upstream commit replacing the magic structs in `core.stdc.config` by magic enums (supporting implicit conversion from integer literals) [as well as adding `cpp_{size,ptrdiff}_t`].
* Add a clarifying comment wrt. DMD/LDC mangling difference on 64-bit OS X and use `version(LDC)` guards.
* Adapt dmd-testsuite accordingly.


JohanEngelen:
Related: https://github.com/ldc-developers/ldc/pull/2650

I find this mangling situation very unfortunate :(  Sorry if asking the same question again, I forgot the answer. Is it now that from 2.079 onwards it is necessary to use different D types for C++ interop with DMD and LDC? Is there any way we can prevent that? (either convince DMD to do what we think is sane, or change LDC to do what DMD thinks is sane?)

kinke:
A short summary:

* No different D types need to be used, now that there's proper `cpp_(u)long`, `cpp_{size,ptrdiff}_t` and `(u)int64_t`.
* With DMD >= 2.079, you can use native D `(u)long` to represent C++ `(unsigned) long long` and `(u)int64_t`, although that's a bad idea as it's not portable (e.g., won't work on Linux x86_64).
* With LDC, it continues to represent C++ `(unsigned) long` and `{size,ptrdiff}_t`, maintaining backwards compatibility and, if you don't care about 32-bit OSX, not requiring an extra import and the ugly `cpp_` prefixes for your C++ declarations in D making ample use of `size_t` (due to perfect match on all other platforms for both `{size,ptrdiff}_t`).
* I tried to convince Walter multiple times already, e.g., just yesterday in https://github.com/dlang/dmd/pull/8249.

kinke:
The above summary is all 64-bit macOS specific, except for the parts where I mentioned 32-bit macOS and 'other platforms'. So 'it = D (u)long on 64-bit macOS and thus D {size,ptrdiff}_t on 64-bit macOS`.

kinke:
[On Windows, D `(u)long` is C++ `(unsigned) long long`. But so is C++ `size_t`, ergo no issues.]

PR #2706, opened 2018-05-20
Win64: Fix extern(C++) ABI wrt. passing non-POD structs by value

kinke:
Return types have been checked for POD-ness already, but parameters haven't.

### DIFF
--- a/gen/abi-win64.cpp
+++ b/gen/abi-win64.cpp
@@ -42,15 +42,46 @@ private:
            && (t->ty == Tfloat80 || t->ty == Timaginary80);
   }
 
-  // Returns true if the D type is passed byval (the callee getting a pointer
-  // to a dedicated hidden copy).
-  bool isPassedWithByvalSemantics(Type *t) const {
-    return
-        // * aggregates which can NOT be rewritten as integers
-        //   (size > 8 bytes or not a power of 2)
-        (isAggregate(t) && !canRewriteAsInt(t)) ||
-        // * 80-bit real and ireal
-        isX87(t);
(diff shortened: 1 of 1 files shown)


kinke:
The dmd-testsuite tests work for Win32 and Win64, so both ABI and destruction rules (MSVC++ lets the callee destruct the param, just like D) match across D and C++, which is pretty nice. Maybe there's a slight chance that 32-bit POSIX passes the tests too, I haven't checked yet.

There's still at least one remaining special case not covered yet, and that's MVC++ apparently treating `struct S { ~this(); }` as POD (any ctor will make it a POD, AFAIK another MSVC special case, but a dtor alone doesn't), while the D frontend doesn't.

kinke:
> Maybe there's a slight chance that 32-bit POSIX passes the tests too

Nope, 32-bit Linux fails with a segfault; like for x86_64, a ref is passed, and the caller destructs it.

kinke:
Erm, turns out this only fixes the case where there's a deleted copy ctor and a move ctor, where 64-bit MSVC++ really passes a ref. With copy ctor, the argument is passed by value in a register, but the return value is still returned via hidden sret (!!), so the existing code and the inconsistent POD-ness check for return/param types is actually correct. A deleted C++ copy ctor apparently enforces by-ref passing though...

kinke:
I reworked this, especially the [tests](https://github.com/ldc-developers/dmd-testsuite/commit/8dad7ca490fbffe8b925c8f84793cd8c728c9b61). They don't handle all permutations and corner cases of course, but the resulting fix should make byval-interop with C++ on Windows/MSVC work in most cases now.

kinke:
[Slightly reworked once more.]

PR #2711, opened 2018-05-22
Define D_AVX and D_AVX2 if the target supports them.

JohanEngelen:


jromang:
Hello, I'm following this pull request because I asked the question in this thread : https://forum.dlang.org/thread/mqgmjoscgztnxudnjfjk@forum.dlang.org

Should I test the changes locally so that this pull request can be accepted ? I've never built LDC, but I can try to if this helps.

kinke:
@jromang: Thanks, but such trivial stuff doesn't need your help. ;)

jromang:
Great, thanks !

PR #2714, opened 2018-05-25
Win64: Use LLVM's vector calling convention for extern(D)

kinke:
I.e., pass and return vectors in registers. With the default C calling convention (and the Win64 TargetABI not touching any vectors), vectors are returned in XMM0, but inefficiently passed as ref to hidden copy (automatically by LLVM and conforming to the regular Win64 platform ABI).

Microsoft's vector calling convention, introduced opt-in with VS 2013, additionally puts HFAs and HVAs (Homogeneous Float/Vector Aggregates) into registers, see https://blogs.msdn.microsoft.com/vcblog/2013/07/11/introducing-vector-calling-convention/.

### DIFF
--- a/gen/abi-win64.cpp
+++ b/gen/abi-win64.cpp
@@ -88,6 +88,26 @@ public:
   Win64TargetABI()
       : isMSVC(global.params.targetTriple->isWindowsMSVCEnvironment()) {}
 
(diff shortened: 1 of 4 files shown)


JohanEngelen:
Nice. lgtm

kinke:
[Now excluding extern(D) variadics, as druntime's `va_arg()` doesn't know about different calling conventions.]

Issue #1935, opened 2016-12-19
ABI error wrt. COM interfaces returning structs


kinke:
DMD bug: https://issues.dlang.org/show_bug.cgi?id=16987

From http://forum.dlang.org/post/tiyttpodenppvlfxixcw@forum.dlang.org, relevant wrt. interfacing with DirectX libraries:

COM classes/interfaces are apparently subject to a separate ABI. Unfortunately, googling it hasn't turned up any official (and not even some inofficial) documentation so far. Based on the first few tests on Win64, integers are returned in RAX, floats (and I guess doubles too) in XMM0, and structs (incl. 2x int32 and 2x float) via hidden sret pointer, with `this` pointer in RCX (1st arg) and `sret` in RDX (2nd arg). Compared to the normal Win64 C++ ABI it just seems more conservative by always returning structs via hidden pointer.

kinke:
Fixed by #2720.

PR #2720, opened 2018-06-02
MSVC: Port Rainer's upstream extern(C++) method ABI fixes

kinke:
MSVC++ methods apparently return *all* structs via sret, even small PODs, see https://issues.dlang.org/show_bug.cgi?id=18928 and #1935.

The problem is that the ABI code operates with `TypeFunction`, which contains no information about enclosing aggregate nor whether it's a static function declaration (a class member function, behaving like other regular functions).

~~The very hacky attempt here is adding a new bool `ismethod` to TypeFunction, setting it at parse-time to whether it's declared in a non-static scope and later checking if it's additionally an aggregate member during type-semantic (at which point the `static` storage class appears to be masked out), setting it back to false if it isn't. But I'm not familiar with this front-end code at all, so may/hopefully have missed a simpler/more robust solution.~~

### DIFF
--- a/dmd/semantic3.d
+++ b/dmd/semantic3.d
@@ -602,7 +602,7 @@ private extern(C++) final class Semantic3Visitor : Visitor
                     }
                 }
 
-                if (!funcdecl.inferRetType && !Target.isReturnOnStack(f))
+                if (!funcdecl.inferRetType && !Target.isReturnOnStack(f, funcdecl.needThis()))
                     funcdecl.nrvo_can = 0;
 
                 bool inferRef = (f.isref && (funcdecl.storage_class & STC.auto_));
@@ -656,7 +656,7 @@ private extern(C++) final class Semantic3Visitor : Visitor
                     if (funcdecl.storage_class & STC.auto_)
                         funcdecl.storage_class &= ~STC.auto_;
                 }
-                if (!Target.isReturnOnStack(f))
+                if (!Target.isReturnOnStack(f, funcdecl.needThis()))
                     funcdecl.nrvo_can = 0;
 
                 if (funcdecl.fbody.isErrorStatement())
--- a/dmd/target.d
+++ b/dmd/target.d
@@ -21,6 +21,7 @@ import dmd.dmodule;
 import dmd.dstruct;
 import dmd.dsymbol;
 import dmd.expression;
+import dmd.func;
 import dmd.globals;
 import dmd.id;
 import dmd.identifier;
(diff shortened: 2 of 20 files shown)


TurkeyMan:
Oh wow, awesome. I have a bunch of patches in DMD already that we're using, is it possible to port this patch to DMD? We can't use LDC for the time being.

kinke:
> We can't use LDC for the time being.

Your loss. ;) - I hope that's just because you don't want to port your existing patches to LDC.

> is it possible to port this patch to DMD?

Sure, but not by me, as I don't know and don't plan on getting involved with DMD's backend.

TurkeyMan:
I have already exceeded my hacking time allocation. When are you likely to sync to DMD master next?
If this doesn't get ported to DMD, there's a high chance that Walter will deploy his own solution, and these will be in conflict... is that fine?

rainers:
I think this approach is too hacky. For example, types are merged according their mangling, but `ismethod` is not part of that. So you either loose it at some point or even modify a different function type.

Here's my attempt to fix this in dmd: https://github.com/dlang/dmd/pull/8330

kinke:
Seen it, thx Rainer; relying on a FuncDeclaration would have been my first choice as well, but it's more work for LDC; no other ABI seems to treat methods differently than normal functions, so it doesn't fit the current architecture.

kinke:
Okay porting was actually straight-forward and seems to work fine, thx Rainer!

kinke:
@rainers: Sorry to bother you; AppVeyor have upgraded their VS 2017 to v15.7.3, and that apparently broke `msvc_eh_terminate()` again (luckily not for default `libcmt` though). 32-bit AppVeyor uses VS 2015, so I don't know about that status. If you find some time... ;)
[I was wondering why I haven't seen that update yet; it's apparently not under 'tools & extensions' anymore since v15.6, but now 'check for updates' in the VS help menu...].

rainers:
> If you find some time... ;)

I might be able to have a look tonight.

rainers:
I didn't yet succeed in building and testing LDC myself: when using LLVM master as I did so far, it always just builds an LDC that produces .o files instead of .obj files (the default target seems wrong). Anything new how to configure LLVM?

With the prebuilt LLVM-6.0.0 libs, I got quite a bit further, but had to fight with a number of things (including my laptop overheating with the workload and shutting off multiple times).

In the end, I was unable to reproduce issues in eh.d/eh2.d with VS 15.7.1 and VS 15.7.3 (which seems to not have changed the libs). I also installed the spectre-libs and passed them explicitely when building eh/eh2, but no problems.

kinke:
> builds an LDC that produces .o files instead of .obj files (the default target seems wrong)

Wtf? I haven't had any issues like that for years. Nothing new wrt. LLVM CMake config either IIRC.

> including my laptop overheating with the workload

Wow :D - building LDC itself with prebuilt LLVM should only take a couple of minutes. `ninja -j<a few less>` may help.

> I was unable to reproduce issues in eh.d/eh2.d with VS 15.7.1 and VS 15.7.3

Thx a lot for checking. AppVeyor's previous image was at 15.7.1 without issues.
What's failing for CI is
```
ldc2 -mscrtlib=msvcrt  -link-defaultlib-debug -run ...\dmd-testsuite\runnable\eh.d
```

Are you sure the Spectre-mitigation-DLLs are used at runtime? Do the filenames feature some sort of suffix?

rainers:
I tried the static libs and they worked. I also tried DLLs, but failed to noticed that I still had an older vcruntime140.dll in the runnable folder. I see it crashing now with any DLL with version 14.14.

PR #2738, opened 2018-06-09
Avoid LLVM intrinsic wrappers, prefer aliases

kinke:
As `pragma(inline, true)` is currently ignored without explicitly enabled cross-module-inlining, LTO with LTO-default-libs was previously required to get rid of the druntime/Phobos calls and enable further optimizations.

### DIFF
--- a/dmd/builtin.d
+++ b/dmd/builtin.d
@@ -337,6 +337,22 @@ extern (C++) Expression eval_llvmsqrt(Loc loc, FuncDeclaration fd, Expressions *
     return new RealExp(loc, CTFloat.sqrt(arg0.toReal()), type);
 }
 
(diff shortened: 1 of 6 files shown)


kinke:
Hmm, there's apparently the need to have the function addresses taken... https://issues.dlang.org/show_bug.cgi?id=4541.

kinke:
> I'm not convinced this – or rather, the std.math part of this – is the correct approach.

Me neither wrt. Phobos.

> the fix is not very time-critical

No worries, I didn't want to squeeze this into 1.10. ;)

kinke:
[LTO is still not really an option on Windows; the default libs don't compile with it.]

kinke:
I removed the `std.math` adaptations; I'll probably follow-up with some more `pragma(inline, true)` for the time being.

PR #2740, opened 2018-06-10
Set proper calling convention and attributes to jit thunk call

Hardcode84:
For some reason, this broke things only on win32.


JohanEngelen:
lgtm

PR #2747, opened 2018-06-11
@dynamicCompileEmit attribute

Hardcode84:
Will be useful for `bind`

druntime part: https://github.com/ldc-developers/druntime/pull/144


Hardcode84:
Updated

PR #2752, opened 2018-06-18
Upgrade to D v2.081.0

kinke:
[Very early stage, Phobos doesn't compile yet, apparently due to NRVO and special ref variable - at least on Win64; no problem for Linux x64 apparently.]


kinke:
I included upstream's/Rainer's new `dmd/root/longdouble.d`, although it's not used yet (this accounts for almost 900 new lines). It's a clean (now all in D + C++ header, no more inline asm) x87 `real_t` implementation for compile-time reals on MSVC hosts (MSVC only supports double-precision reals) and I'll probably make that work soon, primarily for consistent compile-time reals on x86[_64] hosts and no cross-compilation issues from Windows -> Linux etc.

kinke:
Semaphore CI (Ubuntu 18.04) is doing extremely well without self-compilation cycle, only `runnable/minimal2.d` and `std.numeric` (debug only) failing.

kinke:
Only `runnable/minimal2.d` remaining for Semaphore, and that checks whether classes with only `static shared` data members and/or static function members don't require `object.d`, which seems of little value to me.
lit-tests and unittests on Win64 working fine on my box as well; I haven't checked dmd-testsuite yet.

Wrt. crashing self-compiled LDC, I guess that's due to the numerous `extern(C++)` changes (vtable?), and apparently something not covered by dmd-testsuite yet. :/

kinke:
With `{runnable,compilable}/minimal2.d` now disabled, all tests on my Win64 box are green now too.

kinke:
Merging stable and fixing the resulting regressions turned out to be almost as much work as merging beta1. ;)

Unfortunately it didn't fix the self-compilation issues. The vtables seem to be totally off if debugging on Windows is to be trusted; an indirect call to `isTemplateDeclaration()` ending up in `toStringExp()` of an unrelated class IIRC etc.

kinke:
Nope, thx for the hint. I only checked stable, those all sadly made it into master.

kinke:
Oh, I see we have an LDC-specific non-virtual (but `extern(D)`) dtor for class `Dsymbol`, with no corresponding C++ declaration, this may be promising.

kinke:
A new Windows-only debuginfo test (`runnable/testpdb.d`, `test18984()`) fails, checking whether there's a DI local associated with an NRVO/sret variable - we don't declare any DI variable, independent of target.

OSX also fails due to a new, OSX-specific debuginfo test (`runnable/debug_info.d`), checking whether the executable contains line infos in a specified segment and section of the executable. According to [StackOverflow](https://stackoverflow.com/questions/10044697/where-how-does-apples-gcc-store-dwarf-inside-an-executable), the debuginfos are stripped by the linker (from the final executable, but present in the object file), and my quick SSH checks seem to confirm this. I guess they are somehow kept for DMD (non-standard segment?) and disabling the test for LDC is probably safe.
Another issue on OSX is `runnable/cppa.d` failing for 2 out of 4 runs (`Invalid UTF-8 sequence` for 32-bit release and segfault on 64-bit debug, working for 32-bit debug and 64-bit release...).

The rest should work.

kinke:
Ah, finally! http://llvm.org/docs/SourceLevelDebugging.html#llvm-dbg-addr - I hope this will finally allow us to pass an arbitrary IR value to be used as address for some local-like variable (outer variables, sret vars, params passed indirectly by val); new since v7 I guess, I didn't find anything in my LLVM 6 headers. The previous restriction (mainly just working with local allocas and byval params) will hopefully be a thing of the past.

kinke:
Hmm, the `runnable/cppa.d` issue on 64-bit macOS without optimizations is strange; it segfaults at the end of its `main()`, where it's storing the (unused) printf() return value to stack at `-0x3c(%rbp)`, but RBP is null...

kinke:
Looks like a buffer overflow in yet another earlier test - `C.sizeof` vs. `__traits(classInstanceSize, C)` as buffer size for an object to be emplaced.

kinke:
For OSX' `runnable/debug_info.d`, we'd need to set Mach-O section flags according to https://github.com/dlang/dmd/commit/2bf7d0db29416eacbb01a91e6502140e354ee0ef. We'd gain file/line infos in backtraces (for frames in the executable) on OSX then.

JohanEngelen:
> For OSX' runnable/debug_info.d, we'd need to set Mach-O section flags according to dlang/dmd@2bf7d0d. We'd gain file/line infos in backtraces (for frames in the executable compiled with -g) on OSX then.

That's a nice find. (clang runs `dsymutil` afterwards. There are nice links to rationale here: https://stackoverflow.com/a/32299029/3215806)

kinke:
Thx for the link. So if I understand this correctly, running `dsymutil` will create an additional .dSYM file containing the Dwarf data of all object files. druntime's `rt.backtrace` would have to read & parse that file instead of a section of the executable itself. [This seems very similar to the CodeView .pdb files from MS.]

I quickly looked into how to set Mach-O section flags in LLVM but didn't find anything useful. I guess the .dSYM method is cleaner and what Mac debuggers expect, so I'm inclined to disabling the test for now and relying on a future `rt.backtrace` support for .dSYM files.

jacob-carlborg:
> Thx for the link. So if I understand this correctly, running dsymutil will create an additional .dSYM file containing the Dwarf data of all object files. druntime's rt.backtrace would have to read & parse that file instead of a section of the executable itself. [This seems very similar to the CodeView .pdb files from MS.]

That's correct. But these files are more used to ship debug symbols with a released application to the users than used during development. During development the linker inserts references to the object files, the debugger will read the references, find the object files on disk and finally read the debug info from the object files. That's why the debugger works with a D compiler even though no `.dSYM` is generated. This is also easily verified by compiling an executable, remove the object files and then debug the program. The debugger will fail to find the debug symbols.

> I quickly looked into how to set Mach-O section flags in LLVM but didn't find anything useful. I guess the .dSYM method is cleaner and what Mac debuggers expect, so I'm inclined to disabling the test for now and relying on a future rt.backtrace support for .dSYM files.

I implemented the current approach because not using `.dSYM` files is one file less to worry about for the user (developer). This also required less changes in druntime because for Linux and FreeBSD it was looking in a section of the current executable. The debugger will read the debug info correctly with this change as well. It seems like the debugger is looking for specific sections while the linker is looking for attributes on the section.

BTW, please ping me in the future for anything macOS related, like this. I might be able to help and explaining, this might have saved you guys some time instead of having to figuring out this again after I already figured out how it works.

jacob-carlborg:
> I quickly looked into how to set Mach-O section flags in LLVM but didn't find anything useful

I think it would be set in a similar way like this [1]. In that example `literal_pointers` is a section type and `no_dead_strip` is a section attribute (flag).

[1] https://github.com/ldc-developers/ldc/blob/8f4027fa0019adcf916d5178dd919d51b3f9121f/gen/objcgen.cpp#L83

JohanEngelen:
@jacob-carlborg Could you try and see whether you can make it work for LDC?

We can then add a flag that allows users to switch between the two possibilities. Having to run dsymutil is a little annoying (I ran into it when doing ASan work), and not many people may know about it (clang runs it for the user) so I'm in favor of having the default be the same as DMD. But I think it's good to be able to switch to dsymutil-based workflow.

jacob-carlborg:
@JohanEngelen any hints to where this need to be changed/set in LDC?

kinke:
> BTW, please ping me in the future for anything macOS related, like this. I might be able to help and explaining

Will do, but you already did help a lot via the very detailed commit message, thanks!

> The debugger will read the debug info correctly with this change as well.

I was thinking about other code analyzers too, profilers etc., but I'm not familiar with Mac.

> any hints to where this need to be changed/set in LDC?

That's the main issue. While `setSection()` works for symbols themselves, the debuginfo stuff is generated 'implicitly' by LLVM AFAICT, and I have no idea if it's possible to override those section flags. Embedding linker command-line args works (at least for the MS linker), but I didn't find any useful `ld64` args either.

> Having to run dsymutil is a little annoying

Manually, of course; I meant doing it like clang and invoking it automatically after linking; but thinking about it, that would defeat the original idea behind these separate debuginfos (faster linking during iterative development).

jacob-carlborg:
> I was thinking about other code analyzers too, profilers etc., but I'm not familiar with Mac.

True. But I'm not sure which tools would read the debug info.

jacob-carlborg:
> Could you try and see whether you can make it work for LDC?

I've done some digging in LLVM and this is what I came up with: the section attribute is set in in the `DwarfLineSection` field in `MCObjectFileInfo::initMachOMCObjectFileInfo` here [1]. For macOS 64bit there's a subclass `X86_64MachoTargetObjectFile` that is used, with the following inheritance chain (most derived to the right):

```
MCObjectFileInfo -> TargetLoweringObjectFile -> TargetLoweringObjectFileMachO -> X86_64MachoTargetObjectFile
```

A `Target` contains a `TargetMachine`. A `TargetMachine` contains a `MCObjectFileInfo`. For macOS 64bit there's a `X86TargetMachine` target machine, with the following inheritance chain (most derived to the right):

```
TargetMachine -> LLVMTargetMachine -> X86TargetMachine
```

The `DwarfLineSection` field is protected, so in theory it would be possible to subclass `X86_64MachoTargetObjectFile` to override the default value of `DwarfLineSection`. This would require subclassing `X86TargetMachine`, which unfortunately is not possible because it's `final` 😞.

[1] https://github.com/llvm-mirror/llvm/blob/e8508360bc0544c086058911cc8673d98f35aa4c/lib/MC/MCObjectFileInfo.cpp#L234-L236

kinke:
@jacob-carlborg: Thanks a lot for checking. There is one viable option - we have our own LLVM [fork](https://github.com/ldc-developers/llvm) (mainly used for the prebuilt release packages; the `-link-internally` feature depends on it too for example, or, more precisely, the LLD fork ;)). As 6.0.1 is about to be released soon, I'd have upgraded anyway, so we could include a tiny patch for Mach-O debug sections while at it.

[We could also enable more LLVM targets for that build. Linux distros seem to default to including all targets.]

jacob-carlborg:
> Thanks a lot for checking. There is one viable option - we have our own LLVM fork (mainly used for CI and the prebuilt release packages; the -link-internally feature depends on it too for example, or, more precisely, the LLD fork ;)). As 6.0.1 is about to be released soon, I'd have upgraded anyway, so we could include a tiny patch for Mach-O debug sections while at it.

I was thinking of that as well, not sure how much customization you're willing to make. As far as I can see there are two ways to patch this: either change the attribute directly or remove the `final` specifier for `X86TargetMachine` to allow creating a subclass of `X86TargetMachine`. I think the second option is more flexible and would be able to support a command line switch that Johan suggested.

jacob-carlborg:
A third option could be, not sure if that's possible to do for LDC, is to duplicate the data in `__debug_line` in a new custom section without the debug attribute. druntime would read the custom section while all other standard tools should not be affected.

kinke:
> or remove the final specifier for X86TargetMachine to allow creating a subclass of X86TargetMachine

Making the flags configurable at runtime would be nice, but is surely more complicated (we need to remain fully compatible with vanilla LLVM).

> duplicate the data in __debug_line in a new custom section without the debug attribute

I guess this would be doable via some late extra LLVM pass, but having the (potentially large) debug data twice in the object files seems undesirable.

jacob-carlborg:
> we need to remain fully compatible with vanilla LLVM

I thought you weren't, that's why you're using a fork?

kinke:
Nope, the fork is just used for more advanced features not working with vanilla LLVM (`-link-internally`, which requires a one-liner patch for LLD I haven't gotten round to submitting yet, and emulated TLS for Android, which I know nothing about). It also contains compiler-rt (e.g., sanitizer libs) directly in its source tree and prebuilt binaries (=> CI).
Semaphore and Travis (except for 1 job) use vanilla LLVM.

kinke:
Oh I just realized LLVM 6.0.1 has already been tagged (annoyingly, no tags in git mirrors...). I merged it into our fork and included a Mach-O section hack: https://github.com/ldc-developers/llvm/commit/7798278e8d7c2738ea7187498f6db7dc9896004d

kinke:
I also found the culprit for sporadic std.experimental.allocator unittest failures on Win64 (and OSX too IIRC, not sure about Linux): https://github.com/dlang/phobos/pull/6622

kinke:
Oh, that `stable` made it to the final 2.081.0 upstream shortly afterwards.

PR #2754, opened 2018-06-24
Don't count the "skipping definition" warning as a real warning.

JohanEngelen:
The warning currently stops compilation with `-w`, but because this is not a real warning that the user can fix, compilation should continue. I've also marked the warning as "LDC INTERNAL" for clarity.

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -968,13 +968,13 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   llvm::Function *const func = irFunc->getLLVMFunc();
 
   if (!func->empty()) {
-    // Don't count this warning as a real warning (to not stop compilation with `-w`)
-    auto savedWarningCount = global.params.warnings;
+    // XXX: HACK Don't count this warning as a real warning (to not stop compilation with `-w`)
+    auto savedWarningCount = global.warnings;
     warning(fd->loc,
             "LDC INTERNAL: skipping definition of function `%s` due to "
             "previous definition for the same mangled name: %s",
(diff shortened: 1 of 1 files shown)


kinke:
> because this is not a real warning that the user can fix

Where are you hitting this? I thought multiple *definitions* would be easy to avoid except maybe for extern(C++) definitions in templates with instantiations in multiple modules (although they arguably shouldn't be in a template then if they don't depend on any template param).

With 2.081, aggregates (not just classes) now store their 'linkage', so we'll probably be able to handle multiple C++ *declarations* (sharing a common IR aggregate type) gracefully soon.

JohanEngelen:
There is a change in behavior between 1.8 and 1.9, where the compiler now (correctly!) errors out on warnings during codegen. I've never spent the time to dustmite the complex cases where this warning is hit, perhaps now I should...   (this of course involves multiple modules and templates and separate compilation)

Regardless, what do you think about this code?
```d
void foo() {}
void foo() {}
```
This is accepted by DMD, also with `-w`. LDC 1.8 "warns", LDC 1.9 really warns and errors with `-w`.
I find that superquestionable, but needs discussion on language spec.

kinke:
> what do you think about this code

Horrible. :)
So it does compile. There's at least an error when trying to call it though (`Error: '...' called with argument types '(...)' matches both: ...`).
Codegen-wise, DMD seems to concatenate the function bodies: https://run.dlang.io/is/RFzEJF

JohanEngelen:
I've started minimizing the test case. It's mixin template mixing in stuff, apparently the template is instantiated twice.

PR #2757, opened 2018-06-27
[macOS] CMake: Work around ranlib potentially corrupting static libs

kinke:
Also cross-checked against LLVM 6.0.0 (no relevant changes though).


JohanEngelen:
perhaps you wanted to target `master` instead?

lgtm

kinke:
Nope, this is done as part of remaining 2.081 OSX troubleshooting (dynamic-compile integration tests fail for Circle, once again due to a ranlib-corrupted static lib in the install folder when using the integrated archiver...).

PR #2758, opened 2018-06-27
Make LDC custom passes available to jit, add API for jit compiler options

Hardcode84:
Somewhat hacky, directly adding some files from `gen/passes` to jit cmake project.


Hardcode84:
And question about licenses, runtime use Boost license but passes part of compiler and use BSD.

JohanEngelen:
Licenses: if you copy stuff with a different license, you have to include that license aswell. You should mark which parts have which license.  Most of LDC source has the BSD license indeed (like LLVM).

Hardcode84:
license fix, please check.

LICENSE
```diff
@@ -24,7 +24,8 @@ Libraries (lib/ and import/ in binary packages):
    Software License. See the individual source files for author information.
 
  - The ldc-jit-rt library is distributed under the terms of the 
-   Boost Software license (but additionally makes use of LLVM code).
+   Boost Software license (but additionally makes use of LLVM code and
+   code from LDC compiler).
```
JohanEngelen: @Hardcode84  Here you need to state that the library is distributed under which licenses: Boost, LLVM's BSD license, and LDC's BSD license.

JohanEngelen: It would certainly help if you can mark in the source file which parts are under which license. That way, it is also easier to remove the dependency on that particular piece of code+license in the future (if need be).

Hardcode84:
Made these passes optional for jit, but I am still not sure what to do with licenses

Hardcode84:
Updated license

JohanEngelen:
Sorry for being unresponsive, I'll edit the license to what I think it should be.

Hardcode84:
thanks

Hardcode84:
rebased

runtime/jit-rt/cpp-so/optimizer.cpp
```diff
@@ -4,6 +4,8 @@
 //
 // This file is distributed under the Boost Software License. See the LICENSE
 // file for details.
+// Uses some parts from gen/optimizer.cpp which is under the BSD-style LDC
```
JohanEngelen: It can be nice to explicitly mark which parts (e.g. put them between `/// LDC LICENSE START/END`).

joakim-noah:
There seems to be [some interest in your work on the forum](https://forum.dlang.org/post/dvuigmebrqebbbjmkwrz@forum.dlang.org), mind answering their questions?

Hardcode84:
Suddenly I decided to revive this PR.

Are everyone still ok with license changes.

Also I had to copypaste `DtoSize_t` and `DtoConstSize_t` to `SimplifyDRuntimeCalls.cpp` to avoid dependency on rest of the compiler.

Hardcode84:
@kinke @JohanEngelen I'm going to merge this

kinke:
Sure, go ahead. - I'd like to release a first v1.18 beta soon; is there anything else besides presumably #3154 you want to see in there?

Hardcode84:
> 
> 
> Sure, go ahead. - I'd like to release a first v1.18 beta soon; is there anything else besides presumably #3154 you want to see in there?

Only #3154

Issue #2153, opened 2017-06-05
[tracker] AArch64 progress


klickverbot:
I'm currently on-off working on `aarch64-unknown-linux-gnu` and have a large part of the test suite passing already, including `std.math` (IEEE 754 quad-precision real support). The idea is to get an `ltsmaster` compiler that passes the whole test suite first.

Open items:
 - [ ] `core.stdc.stdarg` support
 - [ ] C++ ABI tests

Upstream PRs to integrate into `ltsmaster`:
 - [ ] dlang/dmd#6857
 - [ ] dlang/dmd#6858

---

@redstar: How is your `cent` work progressing?

redstar:
@klickverbot I still have some trouble with the 128bit emulation code.

klickverbot:
@redstar: What sort of trouble in particular?

egrimley:
Link to Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=791976

joakim-noah:
I've been trying out Android/AArch64, now that I have access to an Android/AArch64 tablet, with an updated version of dlang/druntime#1267 for `AArch64`.  One oddity is that llvm doesn't invoke `__tls_get_addr` on linux/AArch64, which was one of the main reasons native TLS didn't work on Android/ARM, as the Android dynamic linker doesn't provide that function.

As a result, I wonder if native TLS can be made to work on Android/AArch64, though when I tried reverting my llvm patch and other ldc/druntime modifications for Walter's brand of emulated TLS, the resulting test binary would segfault when the GC called some thread-local variable.  I didn't look into it further and switched to llvm's emulated TLS instead, which is what clang uses on Android/AArch64, but I'll investigate using native TLS more later.

Using llvm's emulated TLS and cross-compiling the druntime tests, I had to disable these `real`-related tests in `core.internal.{convert,hash}`:
```
diff --git a/src/core/internal/convert.d b/src/core/internal/convert.d
index eeb5580c..a31cbb3e 100644
--- a/src/core/internal/convert.d
+++ b/src/core/internal/convert.d
@@ -348,13 +348,13 @@ version(unittest)
         //testNumberConvert!("-double.nan"); //BUG @@@3632@@@
         testNumberConvert!("double.nan");
 
-        testNumberConvert!("-real.infinity");
+        /*testNumberConvert!("-real.infinity");
         testNumberConvert!("real.infinity");
         testNumberConvert!("-0.0L");
[...]

PR #2760, opened 2018-06-29
Circle & AppVeyor: Upgrade to LDC-LLVM 6.0.1

kinke:
Incl. `__debug_line` Mach-O section hack.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -263,7 +263,7 @@ jobs:
       - image: ubuntu:14.04
     environment:
       - CI_OS: linux
-      - LLVM_VERSION: 6.0.0
+      - LLVM_VERSION: 6.0.1
       - HOST_LDC_VERSION: 1.10.0
       - EXTRA_CMAKE_FLAGS: "-DMULTILIB=ON -DBUILD_LTO_LIBS=ON -DD_FLAGS='-w;-flto=thin' -DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++ -DLDC_INSTALL_LTOPLUGIN=ON -DLDC_INSTALL_LLVM_RUNTIME_LIBS=ON"
       - DUB_VERSION: v1.9.0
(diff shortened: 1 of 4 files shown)


joakim-noah:
I'd like to get [this llvm patch for Android/AArch64](https://github.com/ldc-developers/ldc/issues/2153#issuecomment-397947656) into our 6.0.1 llvm release, but there's a slight chance it might require tweaking, as it could be causing the remaining GC issues listed there. I'll submit a pull for our llvm once I get that worked out, just letting you know so you don't release before I get that in.

kinke:
The Mach-O section hack seems to work fine:
```
3062:  ... runnable/debug_info.d          -g -O ()!!! [DISABLED for LDC]
3062:  !!! runnable/debug_info.d          DISABLED but PASSES!
```

[The test cannot be enabled, as it would still fail for non-LDC-LLVM, e.g., Travis.]

> just letting you know so you don't release before I get that in.

Thanks; I won't wait forever though. ;)

We should also discuss whether we want to include more targets, at least for the x86[_64] builds. With MIPS being usable now AFAIK, I tend to just enabling all targets. I'll check the effect on the LDC executable file size.

kinke:
On Win64, `ldc2.exe` (`RelWithDebInfo` CMake config, enabled LLVM/LDC assertions) increases by 24% (57 MB vs. 46 MB) when enabling all targets, which is more than I thought.

joakim-noah:
Better to just include the targets for which we have some support, which is less than half of those listed.

kinke:
I think so too; MIPS alone adds less than 2 MB.

> targets for which we have some support

That's where `-betterC` can come in handy; cross-compiling a hello-world with C printf etc. might just work for more exotic targets and so potentially enourage someone to attempt a port. That's why I preferred the `all` variant first, but I expected an increase of 10-15%.

joakim-noah:
Forgot about betterC, is it as bad for the release build with assertions turned off?

JohanEngelen:
Note that `all` doesn't include WebAssembly
(needs `-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly`).

kinke:
> is it as bad for the release build with assertions turned off?

I'll just guess so, not too interested in rebuilding LLVM twice to find out. ;)

> Note that all doesn't include WebAssembly

Yep, there's a bunch of excluded experimental targets, incl. RISCV, but good point, there appears to be interest. So shall we just add Mips and WebAssembly for 6.0.1?

JohanEngelen:
> there appears to be interest

I think this is a good measure to decide on. If we've seen some interest/work on a certain target, then let's include it to ease things.

kinke:
Alright, that'd translate to `-DLLVM_TARGETS_TO_BUILD=AArch64;ARM;Mips;MSP430;NVPTX;PowerPC;X86 -DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=RISCV;WebAssembly` then. There's also code for SPARC and SystemZ in LDC, but I guess the benefit of a x86_64 cross compiler for those is pretty low.

kinke:
Mach-O section hack now opt-out via (hidden) `-preserve-dwarf-line-section=false`; option only available if built with supporting LLVM.

PR #2764, opened 2018-07-05
Hotfix

kinke:


kinke:
Currently building with qemu for armhf. The newer qemu version shipping with Ubuntu 18.04 is apparently finally able to use multiple host CPU cores; building LLVM 6.0.1 took less than 12 hours, that's a huge improvement. :]

PR #2765, opened 2018-07-06
Add some support for classes without TypeInfos

kinke:
For `-betterC` and/or a minimal `object.d` without TypeInfo:

* Skip TypeInfo emission for classes and interfaces.
* Emit null as first vtable member.

Makes dmd-testsuite's `{compilable,runnable}/minimal2.d` work.

### DIFF
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -65,11 +65,6 @@ void DtoResolveClass(ClassDeclaration *cd) {
     getIrField(vd, true);
   }
 
-  // emit the interfaceInfosZ symbol if necessary
-  if (cd->vtblInterfaces && cd->vtblInterfaces->dim > 0) {
-    irAggr->getInterfaceArraySymbol(); // initializer is applied when it's built
-  }
-
   // interface only emit typeinfo and classinfo
   if (cd->isInterfaceDeclaration()) {
     irAggr->initializeInterface();
--- a/gen/declarations.cpp
+++ b/gen/declarations.cpp
@@ -89,14 +89,13 @@ public:
       }
 
       // Emit TypeInfo.
-      DtoTypeInfoOf(decl->type, /*base=*/false);
-
-      // Declare __InterfaceZ.
-      IrAggr *ir = getIrAggr(decl);
-      llvm::GlobalVariable *interfaceZ = ir->getClassInfoSymbol();
-      // Only define if not speculative.
-      if (!isSpeculativeType(decl->type)) {
-        defineGlobal(interfaceZ, ir->getClassInfoInit(), decl);
+      if (global.params.useTypeInfo && Type::dtypeinfo) {
+        IrAggr *ir = getIrAggr(decl);
+        llvm::GlobalVariable *interfaceZ = ir->getClassInfoSymbol();
+        // Only define if not speculative.
+        if (!isSpeculativeType(decl->type)) {
+          defineGlobal(interfaceZ, ir->getClassInfoInit(), decl);
+        }
       }
     }
   }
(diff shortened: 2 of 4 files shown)


kinke:
Pinging @JinShil.

gen/classes.cpp
```diff
@@ -65,11 +65,6 @@ void DtoResolveClass(ClassDeclaration *cd) {
     getIrField(vd, true);
   }
 
-  // emit the interfaceInfosZ symbol if necessary
-  if (cd->vtblInterfaces && cd->vtblInterfaces->dim > 0) {
-    irAggr->getInterfaceArraySymbol(); // initializer is applied when it's built
-  }
```
kinke: Ignore the comment, this only declared it. It should be declared lazily on-demand anyway.

JinShil:
cc @ibuclaw

I never got this working for `-betterC` in DMD.  I ran into a problem due to `-betterC`'s janky implementation of importing `object.d` at compile-time, but then pulling the rug out from under it at link-time:  https://github.com/dlang/dmd/pull/8324

kinke:
> I never got this working for -betterC in DMD.

When making runnable/minimal2.d's main() `extern(C)` (may be an LDC quirk) and compiling with `ldc2 -betterC minimal2.d`, I get 4 undefined symbols, the virtual base methods from `Object` in druntime's `object.o`, required for the vtables (LDC defines those in the module declaring the class (and only there), regardless of whether they are used or not). So a custom `object.d` is still required to get it to link.

JinShil:
Sorry if this turned out to be a PITA.  My intention was that if `TypeInfo` and `Object` weren't defined, class instances would not be supported, and *nothing* from `Object` would even need to be emitted:  See dsymbolsem in https://github.com/dlang/dmd/pull/8204/files  It seemed to work right for DMD.  I expected the compiler to determine what it needed to emit from what was defined in `object.d`.  It appears the compiler is still redundantly hardcoding things that are already defined in druntime, so when druntime isn't actually there, it's out of sync.  

My intention in all of this is for druntime to be *less special* and treated just like any other library:  If a symbol is not declared, but is used, the compiler emits and error just like it would for any other library.  If a symbol is not declared, but also not used, it would merrily carry on.  It was [an idea I proposed almost 4 years ago](https://forum.dlang.org/post/psssnzurlzeqeneagora@forum.dlang.org).  When @WalterBright implemented -betterC, I saw the chance to make it work.

kinke:
I think I understand what you're trying to achieve, and see value in it.

All I'm saying is that you still need a custom `object.d` with either no `Object` (so that your new check kicks in, only implicitly deriving top-level classes from `Object` if there is such a class in `object.d`) or some `Object` base class (and compile/link the object.d file if it has any virtual methods). Apart from that, `-betterC` alone seems to suffice to use classes to some extent already with this PR.
I guess there's no way around a custom `object.d` here, as I would still like to be able to specify a custom `Object` for `-betterC` code.

My intention is to also enable usage of class instances with a minimal runtime, e.g., I don't see why virtual function calls shouldn't work (excl. interfaces) with manually (or `scope`?) allocated instances. E.g., this works (no druntime dependency) with an empty `object.d`:
```D
class A
{
    bool foo() const { return false; }
}

class B : A
{
    override bool foo() const { return true; }
}

__gshared immutable A a = new A();
__gshared immutable B b = new B();

extern(C) int main()
{
    assert(!a.foo());
    assert(b.foo());
    return 0;
}
```

JohanEngelen:
@kinke Perhaps it's easiest to create a `tests/baremetal/` directory with lit configuration that automatically specifies a modified `ldc.conf` that does not include the std libs nor object.d etc. ? I think it would ease adding testcases

kinke:
@JohanEngelen: Good idea, I implemented something similar in the WebAssembly-PR, https://github.com/ldc-developers/ldc/pull/2766/commits/e46cf560360fa76a488dfa2e890e785180433ad0. I'll add the code above as further test here after that one is merged.

PR #2768, opened 2018-07-10
Modify baremetal/classes.d test such that it doesn't need `-betterC`.

JohanEngelen:
@kinke A simple change to the test, because it's cool that we can do this in normal D mode.


kinke:
Fine by me. I took the liberty of simplifying it a bit.

PR #2773, opened 2018-07-11
Add @naked UDA

kinke:


kinke:
Motivation: being able to replace the DMD-style naked asm functions with LLVM asm + @naked UDA, in order to get LTO working with druntime/Phobos on Windows.

gen/uda.cpp
```diff
@@ -13,6 +13,14 @@
 #include "id.h"
 
 #include "llvm/ADT/StringExtras.h"
+#include "llvm/ADT/StringSwitch.h"
+
+namespace llvm {
+// Auto-generate:
+// Attribute::AttrKind getAttrKindFromName(StringRef AttrName) { ... }
```
JohanEngelen: That's a nice find.

kinke: Yeah, took me a bit of time. ;) - Let's see which LLVM versions support this.

JohanEngelen:
LGTM. Would be nice to add a link to the LLVM langref `naked` to the documentation.

kinke:
There's no direct URL to the `naked` spec, it's just one of the many function attributes. I copied the 2 sentences there as druntime comment.

kinke:
I didn't find a similar function in LLVM 3.7 and 3.8, so this kind of `llvmAttr("EnumAttributeName")` may be LLVM 3.9+ only (and so cannot be used in druntime/Phobos until we lift the min LLVM version :/).

gen/uda.cpp
```diff
@@ -13,6 +13,20 @@
 #include "id.h"
 
 #include "llvm/ADT/StringExtras.h"
+#include "llvm/ADT/StringSwitch.h"
```
ZombineDev: This include looks unnecessary.

kinke: It isn't. ;)

JohanEngelen:
> There's no direct URL to the naked spec, it's just one of the many function attributes. I copied the 2 sentences there as druntime comment.

I more meant that'd it be nice to say explicitly that the UDA is applying the LLVM `naked` attribute.

gen/uda.cpp
```diff
@@ -204,6 +218,18 @@ void applyAttrLLVMAttr(StructLiteralExp *sle, llvm::Function *func) {
...
+#if LDC_LLVM_VER >= 309
+    const auto kind = llvm::getAttrKindFromName(key);
+    if (kind != llvm::Attribute::None) {
+      func->addFnAttr(kind);
+      return;
+    }
+#else
+    // no getAttrKindFromName(); just detect `naked` for now
+    if (key == "naked") {
+      func->addFnAttr(llvm::Attribute::Naked)
```
JohanEngelen: forgot `return` here. Better to use an if-else chain to prevent such errors?

kinke: Wow, even managed to forget the semicolon. ;)

kinke:
It may need to do more than just applying the LLVM attribute. We shouldn't allocate storage for implicit and explicit parameters (our 'prologue'); I had to resort to declaring a parameter-less impl-function and overriding its mangle in order to fake a naked function taking 2 params.

kinke:
LDC prologue/epilogue now disabled too, which seems to work fine for Win64. On Linux x64 however, the 2 params of the testcase are still dumped to memory; the generated IR is *exactly* the same as the Win64 one, incl. function attributes.

kinke:
Alright, same thing for Linux happening [with clang](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(fontScale:1.0749542399999998,j:1,lang:c%2B%2B,source:'//+Type+your+code+here,+or+load+an+example.%0A__attribute__((naked))+void+sum(+int+a,+int+b)+%7B%0A++++asm(%22leal+(%25rdi,%25rsi),+%25eax%22)%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:clang600,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'1',trim:'1'),lang:c%2B%2B,libs:!(),options:'-emit-s',source:1),l:'5',n:'0',o:'x86-64+clang+6.0.0+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4). `-O` can be used as ugly workaround.

The updated version doesn't require a return statement anymore (and so allows for naked functions without return instruction), see the extended test.

Issue #628, opened 2014-06-11
BUILD_BC_LIBS doesn't work on Windows


Trass3r:
The llvm-ar commandline in the VS project file looks perfectly fine but when running it one slash is lost for one of the arguments:

```
llvm-ar.exe: ..ldcsrc/objdir/runtime/src/rt/typeinfoti_ptr-debug.bc: The system cannot find the file specified.
```

PR #2774, opened 2018-07-12
Avoid naked DMD-style asm functions & enable LTO on Win64

kinke:


kinke:
Just checking CI status for now. This is more work than I expected; `ldc.eh_msvc` and `std.internal.digest.sha_SSSE3` are still missing (and non-trivial). There's some more DMD-naked, but for non-Windows only.

kinke:
> I'm somewhat wary of the extra maintenance overhead, but I guess it's acceptable, as LTO is an important use case.

Yeah, not too happy about the duplicated code in `core.thread` either. Also not sure if Rainer likes the AT&T syntax in `ldc.eh_msvc`. ;)

> Idea: What about a global version that can be specified on the command line to use the upstream asm, so we can easily test whether we can remove the workaround in the future?

Hmm, we could also just make an existing LTO lit-test with naked DMD asm xfail on Win64. Enabling the LTO lit-tests would be the next step anyway (+ integration test for AppVeyor, maybe even bundling `lld-link.exe` in the release package).

> we might reimplement naked asm to lower to a naked function rather than module-level asm.

I was thinking about that too, may be a good idea in general (inline-able etc.).

kinke:
Alright, LTO lit tests enabled for Windows, integration test added with LTO default libs and lld-link.exe bundled with x64 and multilib packages. Parameter-less impl-function workaround not needed anymore with updated @naked semantics.

kinke:
[This fixes #628.]

PR #2778, opened 2018-07-16
Fix overzealous check for multiple main() functions

kinke:


thewilsonator:
Add a test case for this?

gen/functions.cpp
```diff
@@ -594,9 +594,9 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
...
 
   // main
-  if (isMainFunction(fdecl)) {
-    // Detect multiple main functions, which is disallowed. DMD checks this
-    // in the glue code, so we need to do it here as well.
+  if (isMainFunction(fdecl) && fdecl->fbody) {
+    // Detect multiple main function definitions, which is disallowed.
+    // DMD checks this in the glue code, so we need to do it here as well.
     if (gIR->mainFunc) {
       error(fdecl->loc, "only one `main` function allowed");
```
ZombineDev: ```"only one `main` function definition is allowed"```

kinke: Hardly worth mentioning explicitly IMO, DMD [doesn't either](https://github.com/dlang/dmd/blob/e4f19f3ea8a22bc8bb9cf13ff491d08f8dbc775d/src/dmd/glue.d#L1384).

Issue #2144, opened 2017-06-02
WebAssembly support


dhasenan:
I'm looking into adding WebAssembly support for ldc2. Should be pretty straightforward -- add sections for it in the four-ish places in `driver/` and in `gen/abi.cpp`.

Just putting this bug up for posterity.

thewilsonator:
I find that the lit tests for the backends in question are a good place to start when it comes to adding support for new backends.

You'll want to (at a minimum)
* Add a version specifier and OS in `driver/main.cpp:registerPredefinedTargetVersions`. you'll probably have to change the file extension code to be `.js` when wasm is the target.
* add abi support (look at the backend lit tests for what it should look like) in `gen/abi.cpp` and add your own files as `gen/abi-wasm.h` and `gen/abi-wasm.cpp`

and then whatever else the in the lit tests are obviously required. Good luck!

brochington:
@dhasenan I'm very exited about the possibility of using D with WebAssembly. Any updates?

dhasenan:
No updates at the moment, sorry. I'll be a bit time-constrained in the next few days but hope to make some progress in the coming week.

cbleser:
I am looking forward to seeing the first release.
Because I am mostly doing an internal project in the company I am working in I switch from C++ to D in 2008 when it was D1.
So most of the code I write is in D, C and Javascript.
It is really interesting if we can get webassember backend of LDC.
When it is ready I would like to test it!

SrMordred:
Waiting for update on this too :)

ZombineDev:
I opened a PRs for [dmd](https://github.com/dlang/dmd/pull/7202) and [dlang.org](https://github.com/dlang/dlang.org/pull/1905) to add the Emscripten and WASM32 (WebAssembly 32-bit) reserved version identifiers.

kinke:
Rudimentary support has landed in https://github.com/ldc-developers/ldc/pull/2766. It's still using the primitive `UnknownTargetABI` and may need a custom one.

The biggest hurdle is probably some sort of light-weight runtime to be built upon, but that's not really WebAssembly specific and should be of interest for other bare-metal usages too.

Closing this.

PR #2779, opened 2018-07-16
Replace predefined versions WebAssembly{32,64} by general WebAssembly

kinke:
As that (+ `D_LP64` for wasm64) is what DMD does, according to Petar (https://github.com/ldc-developers/ldc/issues/2144#issuecomment-335181401).


ZombineDev:
I really hope I can find more free time to play with wasm these days :P I'm glad see you guys also interested in this!

PR #2781, opened 2018-07-17
Fix corrupt prefix in integrated LLD's console output

kinke:


driver/linker-msvc.cpp
```diff
@@ -205,7 +205,7 @@ int linkObjToBinaryMSVC(llvm::StringRef outputPath,
 #if LDC_WITH_LLD
   if (useInternalLLDForLinking()) {
     const auto fullArgs =
-        getFullArgs("lld-link.exe", args, global.params.verbose);
+        getFullArgs("lld-link", args, global.params.verbose);
```
kinke: E.g., here. The implicitly generated `std::string` got destructed right after the call, so `fullArgs[0]` was dangling. It's used as prefix in:
```
lld-link: warning: ignoring unknown argument: -wqaoirjwoj
lld-link: error: entry point must be defined
```

driver/tool.cpp
```diff
@@ -139,14 +139,14 @@ void createDirectoryForFileOrFail(llvm::StringRef fileName) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-std::vector<const char *> getFullArgs(const std::string &tool,
+std::vector<const char *> getFullArgs(const char *tool,
```
JohanEngelen: I think I'd rather see this return a `vector<string>` and take a non-ref (!) `string` as argument, so that corruption cannot happen. Performance is not important here.

kinke: IIRC, this only exists because LLVM expects a vector of `const char*` (and now with LLVM 7, a vector of `llvm::StringRef` if I remember a recent PR of yours right).

PR #2785, opened 2018-07-19
Phobos: Tiny tweaks to make std.{algorithm,range} baremetal-friendly

kinke:
The first Phobos commit is pending upstream, https://github.com/dlang/phobos/pull/6639.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -227,6 +227,7 @@ after_test:
             (cat etc\ldc2.conf).replace('%%ldcbinarypath%%/../lib', '%%ldcbinarypath%%/../lib64') | Set-Content etc\ldc2.conf
             $conf32 = cat ..\ldc2-x86\etc\ldc2.conf -Raw
             $conf32 = "`r`n""i[3-6]86-.*-windows-msvc"":" + $conf32.Substring($conf32.IndexOf("`r`ndefault:") + 10)
+            $conf32 = $conf32.Substring(0, $conf32.IndexOf("`r`n};`r`n") + 6)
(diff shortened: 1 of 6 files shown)


wilzbach:
> A nitpick: I think it'd be nice to keep the explicit version(Posix) branch in std.ascii,

Can't we do an `version (WebAssembly)` branch there? I wouldn't mind accepting this at upstream.

kinke:
I'm only aware of 3 line endings, the Windows one (`\r\n`), the Mac one before OSX (`\r`) and the Posix one (`\n`). Notepad++ only knows about these 3 too. The old Mac one is dead/unsupported, so just having a special case for Windows makes the most sense to me. A WebAssembly-only special case is undesirable for all other bare-metal targets, each needing its own `\n` branch then...

kinke:
Our multilib `ldc2.conf` for Posix with its single section combining 32 and 64 bit targets (lib dirs, rpath dirs, `--no-warn-search-mismatch`) is a PITA here. `-betterC` isn't enough; wasm lld craps out with unsupported `--no-warn-search-mismatch`, so that one would additionally need `-conf= -I<druntime dir> -I<phobos dir>`.

I tend towards splitting up the single section into default (64-bit) and i686 sections, as done for Windows.

kinke:
I went with an extra wasm section in the config file instead. Even with #2790, the main problem with separate sections is figuring out a triple pattern for the 32-bit arch, which is trivial for x86 but likely a problem for exotic platforms. A wasm section is cleaner anyway, and enables convenience stuff like default `-link-internally`.

Issue #2784, opened 2018-07-19
Fails to build as Xcode project


jacob-carlborg:
If cmake is used to generate an Xcode project the `ldc-jit-rt-so` target will fail to compile. The reason is because it contains a C file, `ldc-jit-rt-so-CMakeForceLinker.c`. When Xcode builds this file it will invoke `clang`. For some reason the target it configured with the `-std=c++11`, which is not a legal flag when invoking the C compiler. Xcode has something call `Other C Flags` and `Other C++ Flags`, both of these configurations have the same value, this is where the `-std=c++11` flag is specified.

I don't know if this is a problem with cmake in general or something specific to LDC.

kinke:
This might have been fixed with v1.11, so please check.

jacob-carlborg:
Yes, sorry, this is working now.

jacob-carlborg:
Yes, I somehow thought it was already closed.

PR #2786, opened 2018-07-21
CMake: Fix issue #2784 (build as Xcode project)

kinke:
Purely speculative - use CXX instead of C as linker language for the jit-rt DLL.


kinke:
@jacob-carlborg: Please check. :)

PR #2795, opened 2018-07-31
Fix IR PGO for Windows (upgrade to LDC-LLVM 6.0.1-3)

kinke:
See https://github.com/ldc-developers/llvm/commit/4db8831367f2879f46c245ab046b8205567c164a.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -272,7 +272,7 @@ jobs:
       - image: ubuntu:14.04
     environment:
       - CI_OS: linux
-      - LLVM_VERSION: 6.0.1-2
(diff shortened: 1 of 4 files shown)


JohanEngelen:
Was the patch upstreamed to LLVM already?

kinke:
I never upstreamed anything to LLVM; I hate svn ;) (oh, didn't they switch to git recently?).

Upstreaming the LLD patches (incl. refactoring them to a single common function for all drivers) would also be very helpful, so that `-link-internally` works for distro builds too (now especially useful for WebAssembly).

PR #2812, opened 2018-08-11
Omit CPU type from man page

bmwiedemann:
When a reproducible build is wanted,
we should not add the build CPU type to man pages like ldmd2.1

See https://reproducible-builds.org/ for why this is matters.

The specification for SOURCE_DATE_EPOCH is at
https://reproducible-builds.org/specs/source-date-epoch/
and everyone doing reproducible builds, sets it.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -112,7 +112,9 @@ void printVersion(llvm::raw_ostream &OS) {
 #endif
   OS << "  Default target: " << llvm::sys::getDefaultTargetTriple() << "\n";
   std::string CPU = llvm::sys::getHostCPUName();
-  if (CPU == "generic") {
+  if (CPU == "generic" || getenv("SOURCE_DATE_EPOCH")) {
+    // SOURCE_DATE_EPOCH indicates that a reproducible build is wanted
+    // so we omit build system CPU type from man page
     CPU = "(unknown)";
   }
   OS << "  Host CPU: " << CPU << "\n";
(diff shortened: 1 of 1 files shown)


kinke:
Lgtm, any objections?

JohanEngelen:
This to me looks strange. The output of `ldc2 --version` should not need this special case. While generating the man page, just `sed` the line out. (I guess the problem is that `ldc2 --help` is used for generating the man page?)
If you still want to continue with this, at the very least it needs an explaining comment of what is going on here.

bmwiedemann:
`ldc2 --help` does not capture the CPU type. But `ldmd2 --help` does. And technically, your own machine's CPU type is not part of the ldc version or help and most other compilers do not include it there, so a third option would be to avoid the special case and just drop it.

bmwiedemann:
some context from the ldc2 man-page:
```
Options  marked with (*) also have a -disable-FOO variant with inverted
       meaning.

              based on DMD v2.080.1 and LLVM 6.0.1 built with LDC - the LLVM D
              compiler  (0.17.5) Default target: x86_64-unknown-linux-gnu Host
              CPU: bdver1 http://dlang.org - http://wiki.dlang.org/LDC

              Registered Targets:

       aarch64
              - AArch64 (little endian)

              aarch64_be
```

bmwiedemann:
@JohanEngelen added an explaining comment in the code
my tests with the patch look good.

bmwiedemann:
Can this be merged now?

kinke:
Yeah, I guess the chance of some other distro/packager benefitting from this is higher than someone using `SOURCE_DATE_EPOCH` by accident *and* caring about the host CPU output.

Issue #2150, opened 2017-06-04
AArch64: Structs should not be passed with byval attribute


klickverbot:
Need to look into what is going on in detail, but the `fopencookie` test in `core.sys.linux.stdio` fail due to an ABI incompatibility. We emit

```llvm
declare %core.stdc.stdio._IO_FILE* @fopencookie(i8*, i8*, %core.sys.linux.stdio.cookie_io_functions_t* byval align 8)
```

but Clang emits

```llvm
declare %struct._IO_FILE* @fopencookie(i8*, i8*, %struct._IO_cookie_io_functions_t*)
```

It really is the extra `byval`, `cookie_io_functions_t` is mapped fine.

klickverbot:
Yep, seems like Clang just never emits byval for AArch64 and instead lowers it manually where required. Patch forthcoming (along with all the other changes to make the test suite compile on ARM64, including all the std.math stuff.

klickverbot:
```diff
diff --git a/gen/abi-aarch64.cpp b/gen/abi-aarch64.cpp
index d94cd7c6..f20f7e2f 100644
--- a/gen/abi-aarch64.cpp
+++ b/gen/abi-aarch64.cpp
@@ -12,15 +12,24 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "gen/abi.h"
-#include "gen/abi-generic.h"
 #include "gen/abi-aarch64.h"
+#include "gen/abi-generic.h"
+#include "gen/abi.h"
 
 struct AArch64TargetABI : TargetABI {
+private:
+  ExplicitByvalRewrite byvalRewrite;
   HFAToArray hfaToArray;
   CompositeToArray64 compositeToArray64;
   IntegerRewrite integerRewrite;
 
+  bool passByValExplicit(Type *t) {
+    t = t->toBasetype();
+    return t->ty == Tsarray ||
+           (t->ty == Tstruct && t->size() > 16 && !isHFA((TypeStruct *)t));
+  }
+
+public:
   bool returnInArg(TypeFunction *tf) override {
     if (tf->isref) {
       return false;
@@ -31,33 +40,21 @@ struct AArch64TargetABI : TargetABI {
     if (!isPOD(rt))
       return true;
 
-    return passByVal(rt);
+    return passByValExplicit(rt);
   }
 
-  bool passByVal(Type *t) override {
-    t = t->toBasetype();
-    return t->ty == Tsarray ||
[...]

PR #2813, opened 2018-08-11
AArch64: Fix C++ mangling of va_list & apply ltsmaster's IndirectByvalRewrite fix

kinke:
C++ va_list is mangled as `std::__va_list` struct. According to clang IR, it's special in the sense of not ever being passed by value.

Let's do the same and pass a pointer while at the same time mangling the function as taking the arg by value.

### DIFF
--- a/gen/abi-aarch64.cpp
+++ b/gen/abi-aarch64.cpp
@@ -35,10 +35,24 @@
  * little bit of compiler magic in the following implementations.
  */
 struct AArch64TargetABI : TargetABI {
+private:
+  IndirectByvalRewrite byvalRewrite;
   HFAToArray hfaToArray;
   CompositeToArray64 compositeToArray64;
   IntegerRewrite integerRewrite;
 
+  bool isVaList(Type *t) {
+    return t->ty == Tstruct && strcmp(t->toPrettyChars(true),
+                                      "ldc.internal.vararg.std.__va_list") == 0;
+  }
+
+  bool passIndirectlyByValue(Type *t) {
+    t = t->toBasetype();
+    return t->ty == Tsarray || (t->ty == Tstruct && t->size() > 16 &&
+                                !isHFA(static_cast<TypeStruct *>(t)));
+  }
+
+public:
   bool returnInArg(TypeFunction *tf, bool) override {
     if (tf->isref) {
       return false;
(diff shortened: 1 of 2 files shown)


kinke:
There's an ABI fix wrt. ExplicitByvalRewrite in ltsmaster which isn't in master yet...

kinke:
But a va_copy limitation which isn't in master... argh.

joakim-noah:
This pull gets ldc master to be able to bootstrap itself on Android/AArch64, fixing the link error I mentioned in #2811 yesterday. You may want to add the stdlib/dmd tests and that self-bootstrap to Shippable in this pull to make sure it doesn't regress.

gen/abi-aarch64.cpp
```diff
@@ -32,115 +63,61 @@ struct AArch64TargetABI : TargetABI {
...
+      byvalRewrite.applyTo(arg);
+    }
+    // Rewrite HFAs only because union HFAs are turned into IR types that are
+    // non-HFA and messes up register selection
+    else if (t->ty == Tstruct &&
+             isHFA(static_cast<TypeStruct *>(t), &arg.ltype)) {
+      hfaToArray.applyTo(arg, arg.ltype);
+    } else {
+      if (isReturnVal) {
+        integerRewrite.applyTo(arg);
```
kinke: ltsmaster doesn't use the `IntegerRewrite` at all; it uses `CompositeToArray64` for rewriting returned structs (those not returned via sret). That casts the struct's address to `i64*` and loads from it, so if the struct is smaller than 8 bytes, this might be an issue (for args too, not just the return value).

kinke:
The `IndirectByvalRewrite` fix resolves #2150.

kinke:
Yay, now really fully buildable with latest ltsmaster.

joakim-noah:
I can confirm that ldc master still bootstraps itself and passes the same tests on linux and Android/AArch64 with this pull. The segfault in the tests for `core.sys.linux.stdio` that I mentioned in #2153 yesterday also goes away with David's fix that you pulled in now.

kinke:
Not sure what to make of the
```
warning: Error disabling address space randomization: Operation not permitted
During startup program terminated with signal SIGSEGV, Segmentation fault.
```

stderr output for various DMD tests (apparently not gdb related as the first Google results would suggest).

joakim-noah:
Probably related to the unoptimized codegen issues I mentioned in #2153: I just tried it and can reproduce many segfaults and hangs with `dmd-testsuite-debug` on Kai's linux/AArch64 VPS. I suggest disabling building the debug versions until we figure that out.

joakim-noah:
Looking good, once I get that druntime commit in, pretty much all the druntime tests should pass too.

When do you plan to tag the 1.11 release and start building the compiler binaries? I'd like to get your suggested method of initializing TLS data on Android in for 1.11, ie replacing the delimiter symbols with parsing the ELF headers, but I'll have to get the path of the ELF file differently for shared libraries, which is how Android GUI apps (apks) are distributed, versus command-line binaries, which is easier and seems to be working in my experiments so far.

kinke:
> When do you plan to tag the 1.11 release and start building the compiler binaries?

This weekend (basically just waiting for this to be merged). 1.11 has been ready from my side for weeks, and now additionally got somewhat improved for AArch64 at the very last moment. I guess we'll have a 1.12 beta in a few weeks, so if you don't manage with the Android TLS stuff in those few days, that's fine too IMO.

joakim-noah:
No, a couple days is plenty of time, should be able to get that in.

joakim-noah:
Looking into the Android TLS setup, but thought of a different way to do it than just calling functions from `rt.backtrace.elf`, which will take some time to investigate. Feel free to do the 1.11 release whenever you like, no need to wait on me.

PR #2815, opened 2018-08-13
Merge v2.081.2

kinke:


kinke:
@joakim-noah: Please advance the submodule hashes here when cherry-picking druntime/Phobos stuff. For testing, but also visibility; where do the Phobos changes come from, and have they been reviewed?

kinke:
Ah, cherry-pick of (merged) https://github.com/dlang/phobos/pull/6633.

joakim-noah:
I usually don't bother updating the submodule hashes, because I figure somebody else will for their own commits eventually and I'm in no hurry. My stdlib changes are almost never tested by the ldc CI anyway, since they're for Android, ARM and AArch64. I only cherry-pick commits that've been approved upstream and tested locally, so shouldn't be a problem. If they're ldc-specific, I submit a pull here.

kinke:
> I usually don't bother updating the submodule hashes, because I figure somebody else will for their own commits eventually

I know, 'cause it's almost always me. ;) Please do bother, as changes may get temporarily lost until someone else trying to branch off `ldc` (as opposed to just the latest submodule hash) eventually notices that things are out of sync. It's also helpful when generating the change log.

joakim-noah:
Sure, will do.

PR #2817, opened 2018-08-16
druntime: pull in two AArch64 commits

joakim-noah:
Test them on Shippable to make sure the test runners all build now. Will make this change to the same line in #2813 instead once that's in, but just testing these commits this way for now. Note that the FreeBSD commit has been merged upstream, while the `core.internal.convert` one has been submitted upstream but not approved yet.


joakim-noah:
OK, the druntime test runners now build, let's see what happens when they're run.

kinke:
Doesn't affect the main platforms, except for the additional field in the (private) `Float` struct, so lgtm.

joakim-noah:
You can go ahead and remove the `true` by default when building the test runners in #2813.

PR #2818, opened 2018-08-17
Upgrade frontend & libs to v2.082.0-beta.2

kinke:


kinke:
1.
Cannot be built with current ltsmaster, as the `core.stdc.stdarg.va_*` functions aren't `nothrow`. ltsmaster should probably be adapted if that's really the only blocker.

2.
Cannot build itself on Linux:
```
/root/project/dmd/root/array.d(34): Error: Function type does not match previously declared function with the same mangled name: `_ZN5ArrayImEC1Em`
```
[That's `Array<unsigned long>::Array(unsigned long)`.] Doesn't happen with the v1.10 host compiler, and neither on macOS or Windows.

3.
The `std.exception` failure isn't a real issue (relative `__FILE__` path and unittest runner apparently expected to operate in the src dir as working dir).

=> Almost fully working already.

joakim-noah:
Any plans to put out a linux/AArch64 build of 1.11? One popular D project would like to test their code in a Debian AArch64 Qemu chroot by using such a build, abraunegg/onedrive#95. Maybe you could have Shippable build and upload one, just like it looks like you're doing for the other releases, ie generating them from the CI. If not, I can put one together manually from Kai's VPS, let me know.

kinke:
> Maybe you could have Shippable build and upload one, just like it looks like you're doing for the other releases, ie generating them from the CI.

I was hoping someone else (you?) would eventually do that; I can't do everything. I'd also like to get rid of the manual armhf build and replace it by the AArch64 one.

joakim-noah:
OK, I only mentioned you doing it because you already did it for the other releases with the CI: I figured it'd be a simple copy-n-paste with some small tweaks for you, whereas I'd need to look into it to see what to tweak. I'll put up a manually-built release this time.

For next time, do you want me to look into having the CI do it?

kinke:
Thx, that'd be nice. It should be mostly copy & paste from the CircleCI configs (modulo predefined environment variables), but as experience has shown, it can take a high number of test runs and persistence until it fully works. To make sure it's on par with the other packages (feature-wise), we'll also need a Shippable config + deployment for our LLVM fork. We'll also need to make sure the Shippable jobs are triggered by tags (even if the same commit has already been built), something CircleCI for example doesn't do by default. There's also no prebuilt AArch64 package for [github-release](https://github.com/aktau/github-release); maybe Shippable offers direct GitHub assets uploading functionality.

joakim-noah:
OK, I'll look into getting all that working for 1.12.

PR #2825, opened 2018-08-20
[ltsmaster] Backport LLVM 7 adaptations

kinke:
Tested with current LLVM branch `release_70` on Ubuntu x86_64. Builds fine; default libs too except for an assertion when compiling druntime's `ldc.arrayinit` (not sure if that's related or simply untested by CI):
```
ldc2: /home/martin/llvm/include/llvm/Support/Casting.h:255: typename llvm::cast_retty<X, Y*>::ret_type llvm::cast(Y*) [with X = llvm::Function; Y = llvm::Constant; typename llvm::cast_retty<X, Y*>::ret_type = llvm::Function*]: Assertion `isa<X>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
#8 0x00005597f2cb2c5a (bin/ldc2+0x215fc5a)
#9 0x00005597f2cba9b7 llvm::IRBuilderBase::CreateMemCpy(llvm::Value*, unsigned int, llvm::Value*, unsigned int, llvm::Value*, bool, llvm::MDNode*, llvm::MDNode*, llvm::MDNode*, llvm::MDNode*) (bin/ldc2+0x21679b7)
#10 0x00005597f13939b4 DtoMemCpy(llvm::Value*, llvm::Value*, llvm::Value*, unsigned int) /home/martin/ldc/gen/tollvm.cpp:369:0
#11 0x00005597f13ea30d DtoInitClass(TypeClass*, llvm::Value*) /home/martin/ldc/gen/classes.cpp:182:0
#12 0x00005597f13ea45f DtoNewClass(Loc&, TypeClass*, NewExp*) /home/martin/ldc/gen/classes.cpp:112:0
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,11 +83,26 @@ commonSteps: &commonSteps
     - run:
         name: Run DMD testsuite
         when: always
-        command: cd ../ninja-ldc && DMD_TESTSUITE_MAKE_ARGS=-j3 ctest -V -R dmd-testsuite
+        command: |
+          cd ../ninja-ldc
+          if [ "$CI_OS" = "linux" ]; then
+            # runnable/{sdtor,variadic} fail, see https://github.com/ldc-developers/ldc/pull/2825#issuecomment-415527889
+            rm ../project/tests/d2/dmd-testsuite/runnable/{sdtor,variadic}.d
+          fi
+          DMD_TESTSUITE_MAKE_ARGS=-j3 ctest -V -R dmd-testsuite
+          cd ..
     - run:
         name: Run stdlib unittests
         when: always
-        command: cd ../ninja-ldc && ctest -j3 --output-on-failure -E "dmd-testsuite|lit-tests"
+        command: |
+          cd ../ninja-ldc
+          if [ "$CI_OS" = "osx" ]; then
+            # 32-bit std.process unittests fail; exclude for now
+            ctest -j3 --output-on-failure -E "std\.process.*-32|dmd-testsuite|lit-tests"
+          else
+            ctest -j3 --output-on-failure -E "dmd-testsuite|lit-tests"
+          fi
+          cd ..
     - run:
         name: Install LDC
         command: |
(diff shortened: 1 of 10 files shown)


kinke:
The druntime compile error is due to the changed LLVM memcpy intrinsic; `ldc.arrayinit` is the only defaultlib module using it. I'll just switch to C memcpy to keep the changes minimal; `ldc.intrinsics` doesn't fully support LLVM 4.0+ anyway (`INTRINSICS_FROM_...` only defined for LLVM <= 3.9). As long as it manages to bootstrap LDC and DMD...

kinke:
Bootstrapping master with the same early LLVM 7 works; master then successfully compiles its default libs. After a little ldmd master fix for LLVM 7, it also passes dmd-testsuite and all but 2 defaultlib tests (`std.math`, `assert(pow(-1.0L,  maxOdd) == -1.0L);` failing, maybe an LLVM 7 regression; and `core.internal.convert`, possibly due to slight differences wrt. `real` literal parsing between master and ltsmaster). So LLVM 7 looking pretty good on Linux x64.

We could use this as 0.17.6; it's ready for LLVM 7 and 2.082. I may get round to adding proper CircleCI incl. automated release+src package generation (we need a prebuilt Linux package for the Travis 0.17.x-LDC job for 2.082).

kinke:
I don't have the energy to look into/fix these ltsmaster CI failures, so just a summary for future reference:

- OSX: 32-bit `std.process` unittests failing (debug + release). Untested by the single other OSX CI job (Travis), which only tests 64-bit.
- Linux: 2 dmd-testsuite failures (for all 4 configs, i.e. 32/64-bit debug/release).
  1. `runnable/sdtor`: `dmd2/ctfeexpr.c:374: UnionExp copyLiteral(Expression*): Assertion `ue.exp()->op == TOKarrayliteral' failed.` No other existing CI uses an LLVM with enabled assertions, so somewhat understandable that it's only showing now. OSX passes though - there's no OS-specific `version`ing; only dependency is `core.vararg`.
  2. `runnable/variadic`: Not sure why Circle is the only one failing this; I don't think it has anything to do with enabled assertions:
```
Error: integer constant expression expected instead of <cant>
Error: integer constant expression expected instead of <cant>
Error: integer constant expression expected instead of <cant>
Error: integer constant expression expected instead of <cant>
Error: integer constant expression expected instead of <cant>
runnable/variadic.d(84): Error: template instance variadic.Foo3!(int, int, int) error instantiating
```

joakim-noah:
Good to see 0.17.6 released. @redstar, if you plan on keeping the AArch64 buildbot going now that we have Shippable, you may want to update [the host ldc 0.17 on there to this release](http://buildbot.ldc-developers.org/builders/aarch64_builder/builds/276/steps/cmake/logs/stdio). We'll see a lot more green on there then. :)

PR #2826, opened 2018-08-21
Debug info improvements (for GDB): printing global and imported symbols, non-member and member function calls

Syniurge:
These commits bring LDC closer to GDC's DWARF output and Clang's IR output (regarding static variables and `this` parameters), and enables access to global/imported symbols and function calls (previously only non-member function calls worked, but only with the mangled name):

https://github.com/ldc-developers/ldc/blob/3d9aebc6677474261cc086dc8ee0bc154ce78a65/tests/debuginfo/print_gdb.d

Until recently I wrongly assumed that GDB's D support was incomplete and that calling functions with `print` or `call` wasn't implemented, then I tried GDC and it worked perfectly, @ibuclaw did an awesome job.

I don't know how this PR affects PDB/CodeView output. Last year symbol names got fully qualified: https://github.com/ldc-developers/ldc/pull/2362/commits/6e7249c5744c768bab9332505d6f6d7963ce3cb9 but now symbols get attached to the proper `DIScope` entry (while previously they were attached to the compile unit), so it might still work as expected. Could a Windows user test the PR?

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -917,7 +917,7 @@ ldc::DISubprogram ldc::DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
     name = fd->toPrettyChars(true);
   } else {
     scope = GetSymbolScope(fd);
-    name = fd->toChars();
+    name = fd->isMain() ? fd->toPrettyChars(true) : fd->toChars();
   }
 
   const auto linkageName = irFunc->getLLVMFuncName();
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -914,6 +914,23 @@ ldc::DISubprogram ldc::DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
 
   ldc::DIFile file = CreateFile(fd);
 
+  // A special case is `auto foo() { struct S{}; S s; return s; }`
+  // The return type is a nested struct, so for this particular chicken-and-egg case
+  // we need to create a temporary subprogram.
+  irFunc->diSubprogram = DBuilder.createTempFunctionFwdDecl(
+      GetSymbolScope(fd),                 // context
+      fd->toChars(),                      // name
+      irFunc->getLLVMFuncName(),          // linkage name
+      file,                               // file
+      fd->loc.linnum,                     // line no
+      nullptr,                            // type
+      fd->protection.kind == Prot::private_, // is local to unit
+      true,                               // isdefinition
+      fd->loc.linnum,                     // FIXME: scope line
+      DIFlags::FlagPrototyped,            // Flags
+      isOptimizationEnabled()             // isOptimized
+      );
+
   // Create subroutine type
   ldc::DISubroutineType DIFnType = mustEmitFullDebugInfo() ?
       CreateFunctionType(static_cast<TypeFunction *>(fd->type)) :
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -82,19 +82,6 @@ bool ldc::DIBuilder::mustEmitLocationsDebugInfo() {
 
 ////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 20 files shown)


Syniurge:
There's a still a big issue with function calls which is that GDB isn't aware of DMD/LDC's calling convention (GDC stuck with the C calling convention hence doesn't have this problem) so the order of explicit arguments is reversed.
I added it with 1b1db863482ab84d5ce7de996ffefd767ce0ce25 and in https://github.com/Syniurge/binutils-gdb/commit/257406ad7fd02a68597f1638d124c12eec36621e:

https://github.com/Syniurge/binutils-gdb/tree/dlang-callconv-and-mangling-update

Also GDB is hasn't been updated yet to the new mangling scheme so it cannot find functions using aggregate/enum types which all have a mangling different from before, I'm working on this.

Is this okay to add a calling convention like this? (It could be reverted anyway if it's not accepted by GDB devs or if another solution is found for example.)

kinke:
I wasn't aware of GDC not reversing the args; I'd prefer not reversing them as well, but that requires adapting druntime and Phobos (inline asm & `__xopCmp` IIRC).

rainers:
> Also GDB hasn't been updated yet to the new mangling scheme so it cannot find functions using aggregate/enum types which all have a mangling different from before, I'm working on it.

https://github.com/ibuclaw/gdb/pull/8 is just waiting to be merged.

Syniurge:
> ibuclaw/gdb#8 is just waiting to be merged.

Nice! I merged your PR into my GDB fork and now every function call appear to work.

kinke:
Oh, basically green now except for Windows cdb debugger test failures; I'll have a look this weekend.

kinke:
@Syniurge: If you don't mind, I'll add 2 commits; one adapting the cdb tests and a little refactoring one. You keep force-pushing, so we'll need to sync. ;)

kinke:
Regarding the dmd-testsuite failures on Windows, most are due to the MS linker erroring out with:
```
fatal error LNK1318: Unexpected PDB error; OK (0) ''
```

Those are triggered by `-g -enable-inlining`. We also run into LLVM assertions sometimes with this combination, a long-standing issue.

kinke:
Seems related to nested functions, i.e., functions in a DI scope of another function. Probably not too surprising that it isn't supported by CodeView. Maybe we could use a DI namespace (as direct child of the DI module) as scope.

Syniurge:
> If you don't mind, I'll add 2 commits; one adapting the cdb tests and a little refactoring one. You keep force-pushing, so we'll need to sync. ;)

Okay I won't force push anymore.
Got a busy week, I'm back at it, looking into the Windows failures with inlining.

kinke:
I tested this with LLVM 7; it then fails to link the Phobos debug unittest runner, which worked with 6.0.1:
```
phobos2-ldc-unittest-debug.lib(utf.obj) : fatal error LNK1318: Unexpected PDB error; OK (0) ''
```
This may be valuable as there's no inlining and apparently a single involved module, `std.utf`.

kinke:
I've rebased onto master (force-push) and added a CodeView workaround.

gen/dibuilder.cpp
```diff
@@ -956,8 +957,18 @@ ldc::DISubprogram ldc::DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
   assert(GetCU() &&
          "Compilation unit missing or corrupted in DIBuilder::EmitSubProgram");
 
-  const auto scope = GetSymbolScope(fd);
-  const auto name = fd->toChars();
+  // FIXME: work around apparent LLVM CodeView bug wrt. nested functions
+  ldc::DIScope scope;
+  const char *name;
+  if (isTargetMSVC && fd->toParent2()->isFuncDeclaration()) {
```
kinke: `fd->isNested()` isn't enough, static functions are problematic too.

kinke:
One more force-push, rebasing + getting rid of LLVM specifics < 3.9.

PR #2831, opened 2018-08-25
ShippableCI extensions

kinke:


shippable.yml
```diff
@@ -1,13 +1,18 @@
 language: c
 
 compiler:
-  - clang
+  - gcc
```
joakim-noah: Why switch?

kinke: Our LLVM is built with gcc (8.1) too (even fails to link with clang 6.0). Previously, I had to use clang here, as the official prebuilt LLVM apparently was compiled with clang.

We also assume gcc on Linux for misc. tests; there are ABI differences (wrt. empty structs etc.).

joakim-noah: OK, I use gcc 5.4 on linux/AArch64 and clang 6.0 for Android/AArch64, building llvm and ldc with the same compiler and haven't had issues, guess the problem here is using one for llvm and the other for ldc. The empty struct issue was worked around upstream, but there are still one or two ABI differences, like you said.

shippable.yml
```diff
@@ -79,6 +92,72 @@ build:
     - DMD_TESTSUITE_MAKE_ARGS='-j16 -k' ctest -V -R dmd-testsuite -E "-debug$" || true
     # Run druntime/Phobos unittests (non-debug only for now)
     - ctest -j16 --output-on-failure -E "dmd-testsuite|ldc2-unittest|lit-tests|-debug(-shared)?$" || true
+    # Install LDC
+    - ninja install
+    - cd ..
+    - perl -pi -e s?$LDC_INSTALL_DIR/?%%ldcbinarypath%%/../?g $LDC_INSTALL_DIR/etc/ldc2.conf
+    - cp LICENSE $LDC_INSTALL_DIR
+    - git clone https://github.com/ldc-developers/ldc-scripts.git
+    - cp ldc-scripts/ldc2-packaging/pkgfiles/README $LDC_INSTALL_DIR
```
joakim-noah: Why not just get that file? `curl -L -o $LDC_INSTALL_DIR/README https://raw.githubusercontent.com/ldc-developers/ldc-scripts/master/ldc2-packaging/pkgfiles/README`

Same for the clones below, can just get the release tarballs.

shippable.yml
```diff
@@ -79,6 +92,72 @@ build:
...
+    # Pack installation dir
+    - |
+      if [ "$IS_GIT_TAG" = "false" ]; then
+        artifactBasename="ldc2-${COMMIT:0:8}-linux-aarch64-$(date "+%Y%m%d")"
+      else
+        artifactBasename="ldc2-${GIT_TAG_NAME:1}-linux-aarch64"
+      fi
+      export LDC_ARTIFACT=$artifactBasename.tar.xz
+      mv $LDC_INSTALL_DIR $artifactBasename
+      tar -cf - $artifactBasename | 7za a $LDC_ARTIFACT -si -txz -mx9
```
joakim-noah: Why not just `tar -cJf`, you want to compress it more?

kinke: I want it to make use of multiple threads. The previous simple tar command timed out IIRC (for LLVM, > 10 mins), this here takes way less.

shippable.yml
kinke: I could, but I don't really see the point. The CI services probably use some efficient caching mechanism; cloning ldc-scripts takes 0 secs, dub 2 secs, dlang-tools 1 sec.

joakim-noah: I don't see anything in their docs about cacheing all git clones, but the commands don't show the normal output, so maybe.

kinke:
> I don't know what's going on with the Makefile change

That's a workaround for the dlang-tools Makefile, which [imports](https://github.com/dlang/tools/blob/master/posix.mak#L15) that `osmodel.mak` file (even shallow-clones the whole DMD repo if `DMD_DIR` isn't set appropriately, i.e., for CircleCI!). I used to patch dlang-tools for the manual armhf build; this is a more elegant way of handling it (and generally prepares dmd-testsuite for platforms without `-m{32,64}` (C++ compiler) support, like armhf and aarch64).

> haven't verified github upload API

I did some local tests; it should work. To find out, I'll just merge (the other option being reopening a new PR from a branch on the remote repo). It's currently necessary to check the log output, as a failing upload (e.g., artifact with same name already exists) doesn't make curl return non-zero...

kinke:
With some delay (these rather frequent Shippable hangs/time-outs seem random; sometimes it hangs while packing the installation dir, sometimes when compiling dub (single-threaded!)...), we finally have the first AArch64 CI artifact: https://github.com/ldc-developers/ldc/releases/download/CI/ldc2-03a5e91b-linux-aarch64-20180828.tar.xz

joakim-noah:
Great! I'll try it out on Kai's VPS.

joakim-noah:
Working well, using dub to build Dscanner. Thanks for putting this together, would've taken me longer.

kinke:
You're welcome, thx for testing.

PR #2837, opened 2018-08-30
Upgrade to v2.082.0-rc.1

kinke:


kinke:
I don't know how that managed to slip through, but the `std.typecons` unittests fail to compile for rc1. See https://github.com/dlang/phobos/pull/6674#discussion_r214159110.

PR #2839, opened 2018-09-01
Travis OSX: Downgrade to xcode v9.2

kinke:
With August, they changed the default OSX image from xcode v8.3 to v9.4.1, and that's apparently when the druntime-test-shared failure was introduced for Travis OSX:

```
Testing finalize
/Users/travis/build/ldc-developers/ldc/runtime/druntime-test-shared/finalize
Illegal instruction: 4
```

CircleCI OSX is using xcode v9.2 without issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,11 @@ matrix:
       d: dmd
       env: LLVM_VERSION=3.7.1 OPTS="-DTEST_COVERAGE=ON"
     - os: osx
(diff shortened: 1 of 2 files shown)


kinke:
Mhm, 9.3 fails too, 9.2 works.

kinke:
The 9.2 image is based on OS X 10.12, 9.3+ on 10.13, that might play a role too.

PR #2840, opened 2018-09-01
Make LDC_INSTALL_LLVM_RUNTIME_LIBS more flexible

kinke:


kinke:
@JohanEngelen: Are you okay with the streamlined `ldc_rt.*` library names?

beta1 would be ready from my side after this; giving us time for LLVM 7 support + upgrade in the beta period.

JohanEngelen:
> Are you okay with the streamlined ldc_rt.* library names

Why is this needed? The old names clearly indicated for which system they are built. Don't think it's wise to change the name to something very different from clang.

kinke:
It's not needed, but makes stuff simpler IMO. User X complaining about libs not being found on an exotic platform, easy solution: create symlinks with platform-independent names (as all of our other libs) in one of the config's lib-dirs. No need to wait for an LDC patch getting OS/arch right.

PR #2845, opened 2018-09-04
Shippable: Fix GitHub release name for tagged builds

kinke:
[Needs to be on the official repo for secure env variable...]


kinke:
Alright, post-adding the AArch64 package worked. ;) [Its asan lib is huge, 32 megs, larger than the LTO plugin...]

PR #2850, opened 2018-09-17
Upgrade LLVM to v7.0.0

kinke:


joakim-noah:
I'm reworking the emulated TLS patches for LLVM 7, so don't bother integrating those two commits to our tweaked LLVM repo. I will submit a pull once I know the new approach is working well, by all the same tests passing, likely next week.

kinke:
[The two Android TLS commits have already been merged to a single one and are on the according branch.]

kinke:
> When do you want to release our tweaked LLVM 7?

Asap (and when green obviously), as always. ;) If you manage by next week or so, that's fine, otherwise we can always release a 7.0.0-2 later.

> Do you want to switch the final ldc 1.2 release to it?

I've at least been playing with that idea. But that also depends on when we're green again and on the users' feedback with a second beta using LLVM 7. I was hoping for potential improvements wrt. musttail etc., besides general performance improvements; another hopefully big deal is the new debuginfo intrinsic, where we should be able to associate an arbitrary IR pointer as lvalue for some variable (mitigating the various GEP issues).

kinke:
Windows:
* linker error for `runnable/testmodule.d`: `unresolved external symbol __D7dstress3run17unicode_06_哪里11__moduleRefZ`
  The LLVM 7.0 asm doesn't include a linker directive (`/INCLUDE:...__moduleRefZ`) anymore (=> an `llvm.used` entry in IR), and so the symbol is wrongly stripped out. As that's no problem for all other ModuleInfos, it's almost certainly related to Unicode.
* AppVeyor has timed out every time so far IIRC. We might have to skip the build-itself bootstrapping step or switch to an LLVM with disabled assertions. :/ So there might be more issues, I haven't run the tests locally yet.

OSX: all green

Linux:
* Strange segfaults with `-link-internally`, at least for WebAssembly (lit-tests `baremetal/wasm*.d`). Also reproducible in my Ubuntu 18.04 VM; it crashes while cleaning up the LLD stuff (`freeArena()` in LLD wasm driver). Strange because it's working fine on Windows and OSX. I hope it's not related to order of linker command-line args or something obscure like that.
* `core.internal.convert` unittest failures while bit-wise comparing compile- and run-time `real.max` IIRC, not depending on `-O`. Cannot reproduce locally (with v1.11 as host compiler).
* `std.math.pow(real, real)` unittest failures with enabled optimizations; reproducible in the VM:
```D
immutable real maxOdd = pow(2.0L, real.mant_dig) - 1.0L;
assert(pow(-1.0L,  maxOdd) == -1.0L);       // => -nan
assert(pow(-1.0L, -maxOdd) == -1.0L);       // => -nan
assert(pow(-1.0L, -maxOdd + 1.0L) == 1.0L); // => -nan
assert(pow(-1.0L, maxOdd - 1.0L) == 1.0L);  // => -nan
```

kinke:
> The LLVM 7.0 asm doesn't include a linker directive (/INCLUDE:...__moduleRefZ) anymore (=> an llvm.used entry in IR), and so the symbol is wrongly stripped out. As that's no problem for all other ModuleInfos, it's almost certainly related to Unicode.

I got that mixed up, it's the other way around - these `/INCLUDE` directives are new in the LLVM 7 asm. The embedded cmdline switch with Unicode chars may either not be supported by the MS linker (latest version) or be wrongly encoded; anyway, it's trying to retain that nonexisting ModuleInfo ref. `-link-internally` works, i.e., LLD 7.

kinke:
So Windows is fine now too.

On Linux, the `pow()` issue apparently boils down to a limited lossless convertibility of 80-bit `real` (with 64-bit mantissa) to `ulong` - it seems to hit a brick at 18446743523953737728 = 0xffffff80_00000000; casting 2.0L^64-1.0L and 2.0L^64-2.0L to ulong yields those values, starting at `-O1`.

ZombineDev:
Could that be related to this change in LLVM 7:

> Optimization of floating-point casts is improved. This may cause surprising results for code that is relying on the undefined behavior of overflowing casts. [..]

https://releases.llvm.org/7.0.0/docs/ReleaseNotes.html

joakim-noah:
Hmm, I'm hitting an issue when compiling for ARM with LLVM 7 that the Shippable linux/AArch64 CI doesn't seem to run across. Compiling LDC master against a mostly stock LLVM 7, with only the single emulated TLS patch from our tweaked LLVM applied, I get the following when cross-compiling the tests for `std.digest.murmurhash` from linux/x64 to any ARM platform I've tried, shown here for linux/AArch64:
```
./bin/ldc2 -c -mtriple=aarch64-none-linux-gnu -unittest ../runtime/phobos/std/digest/murmurhash.d
isvolatile argument of memory intrinsics must be a constant int
call void @llvm.memcpy.p0i8.p0i8.i64(i8* nocapture writeonly %12, i8* nocapture readonly %13, i64 %14, i1 %16) #3
```
This is reproducible when compiling for 32-bit Android/ARM also. However, looking at the Shippable log above, it seems to compile that module's tests just fine.

Can you reproduce?

kinke:
@joakim-noah: Looks like you're not using the correct druntime with the ldc.intrinsics fix (2nd commit here). This used to be a Shippable problem and was fixed by that commit.

@ZombineDev: Possibly (in form of an LLVM regression, nothing overflowing for these casts). What's really strange though is that it's only happening on Linux, not on OSX.

joakim-noah:
OK, that fixes it, didn't know you'd already come across it on Shippable. I've submitted ldc-developers/llvm#3, reworking the Android emulated TLS patch.

kinke:
Wrt. long-double-to-ulong cast issue, clang 7 fails too, at least on Linux, starting with `-O1` and unaffected by new `-fno-strict-float-cast-overflow`:

```C++
#include <cmath>
#include <cstdio>

unsigned long long foo(long double x) { return static_cast<unsigned long long>(x); }

int main()
{
    const auto d1 = powl(2.0, 64) - 1;
    const auto r1 = foo(d1);
    printf("d1 = %La\nr1 = 0x%llx\n", d1, r1);

    const long double d2 = 0xffffff8000000001L;
    const auto r2 = foo(d2);
    printf("d2 = %La\nr2 = 0x%llx\n", d2, r2);

    const long double d3 = 0xffffff7fffffffffL;
    const auto r3 = foo(d3);
    printf("d3 = %La\nr3 = 0x%llx\n", d3, r3);
}
```
=>

```
d1 = 0xf.fffffffffffffffp+60
r1 = 0xffffff8000000000
d2 = 0xf.fffff8000000001p+60
r2 = 0xffffff8000000000
d3 = 0xf.fffff7fffffffffp+60
r3 = 0xffffff0000000000
```

With clang v6.0.1 and `-O3`, everything works:
```
d1 = 0xf.fffffffffffffffp+60
r1 = 0xffffffffffffffff
d2 = 0xf.fffff8000000001p+60
r2 = 0xffffff8000000001
d3 = 0xf.fffff7fffffffffp+60
r3 = 0xffffff7fffffffff
```

[I've used https://wandbox.org/ for online clang testing.]

kinke:
This might be https://bugs.llvm.org/show_bug.cgi?id=38289.

kinke:
So what shall we do for v1.12? Assuming the LLVM bug isn't fixed soon (already 2 months old and not considered a release blocker), I only see one option if we want to use LLVM 7 - raising the CPU requirements to SSSE3 for x86[_64] (min: Intel Core; AMD Bulldozer (introduced ~ 7 years ago)) via default flag `-mattr=+ssse3`.

I guess the `core.internal.convert` issue is related to this bug as well (and only manifests if LDC itself is built with an LLVM-7-LDC). The strange LLD cleanup issue on Linux can probably be hacked around by skipping the cleanup (the process is going to terminate very shortly afterwards anyway).

joakim-noah:
I can apply my reworked Android patch to the LLVM 6 branch too, if you decide to release with it. Just let me know.

kinke:
That wouldn't really be required, right? Does it more than add support for x64? [I need to build all 4 Windows packages manually due to AppVeyor slowness, so I'm a bit reluctant wrt. a 6.0.0-4.]

JohanEngelen:
> I only see one option if we want to use LLVM 7 - raising the CPU requirements to SSSE3 for x86[_64] (min: Intel Core; AMD Bulldozer (introduced ~7 years ago)) via default switch `-mattr=+ssse3`.

Yeah :/  I think that's acceptable but not very nice. We need to keep track of that LLVM bug then.

kinke:
I feel the same - unfortunate, but acceptable.

Both unittest failures on Linux are gone as expected. The wasm LLD segfault cannot be worked around by not explicitly freeing the data (commenting out `freeArena()`) - but by passing param `CanExitEarly = true` on the LDC side.

The lit pip package was apparently upgraded to v0.7 (= LLVM 7?) a few days ago, and that broke one fuzzer lit test, in which we use the `cat` command, which is (now?) treated as lit builtin. I had a quick look; the installed package is apparently missing an accompanying helper script (`builtin_commands/cat.py`)...

kinke:
I've made it so that we default to `-mattr=+ssse3` for optimized x86[_64] code with LLVM 7. The user can disable it explicitly (and recompile the default libs with it).

`CanExitEarly = true` for LLD does have a few undesirable side effects (`-run` not running the executable, no object files cleanup...), but it's only used 1) on Linux and 2) for wasm (not runnable anyway).

joakim-noah:
The new Android LLVM patch pretty much just adds x64 support, but I'd like to put out an LDC 1.12 Termux package for Android/x64, [which is currently disabled](https://github.com/joakim-noah/termux-packages/blob/beta/packages/ldc-beta/build.sh#L16), as I'd like to get LDC on x64 Chromebooks using Termux. But the patch doesn't need to go in our branch, can always have that packaging script apply it, so no big deal.

joakim-noah:
BTW, have you tried pinging someone on the LLVM IRC about that bug? I've done that before to get attention on a bug, particularly if you think it's important enough to get fixed for 7.0.1.

kinke:
> have you tried pinging someone on the LLVM IRC about that bug?

Nope, but I added a comment to the LLVM bug. I don't have an IRC client (and haven't used IRC in a decade or so) and wouldn't know who to ping anyway.

joakim-noah:
You don't need a client, there's a [web interface](https://webchat.oftc.net), channel's named #llvm. Doubt you need to find the right person yourself, somebody in there will likely help.

joakim-noah:
Heh, Android/x64 uses a 128-bit `long double`, apparently matching AArch64, just like they use 64 bits on Android/x86, likely to match 32-bit ARM. Tried setting that in ldc, but then Phobos has issues because it assumes `real` isn't that wide everywhere. I'll see if it takes much to get Phobos compiling for x64 with a 128-bit `real`.

kinke:
Wow, performance for `real` should be horrible then...

joakim-noah:
Looks like the codegen just puts out SSE, but I haven't closely checked if it's doing the 128-bit math right.

kinke:
> Looks like the codegen just puts out SSE, but I haven't closely checked if it's doing the 128-bit math right.

It seems to use the SSE registers for the `fp128` IR type, but it's all done in software - e.g., a subtraction becomes a `__addtf3()` call. https://godbolt.org/z/_6ciK1

kinke:
For Phobos `std.math` support, make sure the `*_X87` versions aren't defined, and `INLINE_YL2X` neither (all depending on x87 `real`). A 3rd `real` type for x86 wasn't expected that soon. ;)

joakim-noah:
Yeah, hit a couple of those already, but surprisingly didn't have to patch much at all to get the Phobos test runner built with the 128-bit `real`, just a couple places in `std.math`. Now I have to go through the remaining test failures and see what's going on.

kinke:
That was more of an ad-hoc decision, as the timing seemed right, but yeah, I forgot about the explicit-emulated-TLS setting we talked about a week ago (and there's still no PR here).

kinke:
Oh and something unrelated I noticed yesterday - while Shippable seems to time out way too often, that appears to be an issue between Shippable and their actual AArch64 provider (Packet? don't recall the name), e.g., the uploads make it through although the logs end at some point with a time-out.

joakim-noah:
I was planning on submitting the small patch disabling the built-in emulated TLS as part of a larger pull with the Android/x86-x64 patches and getting rid of the section delimiters for emulated TLS. Like I said, no big deal for Android, just would be nice to have a day or two notice before releases.

kinke:
Maybe try adding the vendor explicitly - `aarch64-unknown-linux-android`.

Possibly related: the default triple for Windows changed from `...-windows-msvc` to `...-win32`, so I had to override the default triple on Windows too, otherwise it would later complain about TLS not being supported for this target etc. I don't think we've ever had that much trouble with a new LLVM version...

joakim-noah:
Yes, I should have said before that when I tried explicitly passing the triple to ldc, the config line of ldc's verbose output said `aarch64--linux-android`, ie it added a blank vendor and made it a quadruple. However, when it was passed with no vendor altogether as the default triple to configure LLVM 7 before building it and no triple was passed to ldc, it just spit out the same string in the config line, ie `aarch64-linux-android`, and had problems, such as not setting the D predefined versions for `linux`, `Android`, etc.

It looks like they changed the way quadruples are parsed in LLVM 7 to make it less forgiving of omitting the vendor, don't see how this hasn't caused problems for others already.

PR #2851, opened 2018-09-19
Fix hidden overloaded virtual function 'RootObject::print' warning (NFC)

JohanEngelen:
Upstream, the print method has been removed already, so this is a temporary fix for LDC 1.12 only.


thewilsonator:
This was removed in upstream.

Issue #2854, opened 2018-09-21
Tests involving inline x86 pass with DMD and fail with LDC


pineapplemachine:
I am the author of mach.d and I've been trying to get tests to pass with LDC. I have tests failing for several modules when compiling using LDC, and the common denominator appears to be the inline x86 implementations of some trig functions.

These tests pass with DMD for at least OSX 10.9 64-bit, Ubuntu 64-bit, Win7 64-bit, and Win7 32-bit targets. I have been testing with LDC 1.11.0.

https://github.com/pineapplemachine/mach.d/blob/master/mach/math/trig/tangent.d#L49
https://github.com/pineapplemachine/mach.d/blob/master/mach/math/trig/sincos.d#L75

The `sincos` function, for example, returns `{sin: 0, cos: 0}` [here](https://github.com/pineapplemachine/mach.d/blob/master/mach/math/trig/sincos.d#L31) and the `tan` function returns `0` [here](https://github.com/pineapplemachine/mach.d/blob/master/mach/math/trig/tangent.d#L29).

You can see Travis CI logs for DMD and LDC here:

DMD (passing): https://travis-ci.org/pineapplemachine/mach.d/jobs/431511305
LDC (failing): https://travis-ci.org/pineapplemachine/mach.d/jobs/431511307

You can reproduce locally by cloning [the mach repository](https://github.com/pineapplemachine/mach.d), navigating to the root directory, and running `dub test`.

kinke:
> the tan function returns 0 here.

The x87 asm implementation in Phobos has the same problem: https://github.com/ldc-developers/phobos/blob/ldc/std/math.d#L987

I think it may have to do with accessing variables (in this case, the param). The Win64 code doesn't access the variable, but the register directly, so if that works, we'd know where the problem is, for `tan()` at least. Unlike (LDC's) Phobos, your code also assumes that `real` is an 80-bit x87 value for Windows MSVC targets, which isn't the case for LDC (=> 64-bit double, just like MSVC's `long double`).

pineapplemachine:
Speaking of which, how would I even do this using `__asm`?

kinke:
I'd try with naked DMD-style asm, that seems the easiest variant.

The following naked version of your tan function works on non-Windows x86_64 with DMD and LDC; I hope there's a better solution for returning `real.nan` than having to call a helper function:
```D
real retnan() { return real.nan; }

real tanx86(in real value)
{
    asm pure nothrow @nogc
    {
[...]

PR #2855, opened 2018-09-22
Phobos: Enable x87 asm versions for std.math.{tan,expi}

kinke:
Following #2854.


pineapplemachine:
I haven't got an inline asm tangent or sincos implementation to work with LDC yet. However, I did have some apparent success with this particular part of the problem by declaring `immutable nan = real.nan` for DMD and `static immutable nan = real.nan` for LDC (each failed to compile for the other) and then writing `fld nan; ret;` inside the `asm` block instead of `return real.nan;` outside of it.

kinke:
> static immutable nan = real.nan

That should have the same problem when compiled as PIC (`-relocation-model=pic`), the global won't be accessible directly.

kinke:
Revived this with a viable workaround for PIC (pass `real.nan` as extra default parameter); the x87 asm version is more than 3x faster than the generic 'software' one on my Ivy Bridge, tested on Linux x64 with `-O` and:
```D
void main()
{
    enum MAXITER = 1 << 20;

    import std.datetime.stopwatch;
    import std.math;
    import std.stdio;

    real sum = 0;

    auto sw = StopWatch(AutoStart.yes);
    for(uint i; i < MAXITER; i++)
        sum += tan(real(i));
    sw.stop();

    writeln(sw.peek.total!"msecs", "ms\t", sum);
}
```

kinke:
The x87 asm for `std.math.expi()` is faster by ~5x (`expi(real(i) / MAXITER)`) to ~12x (`expi(real(i))`) on my box.

PR #2863, opened 2018-09-30
Android: Prevent LLVM 7+ defaulting to incompatible emutls

kinke:


kinke:
Pinging @joakim-noah; completely untested.

driver/targetmachine.cpp
```diff
@@ -528,6 +528,15 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
     targetOptions.DataSections = true;
   }
 
+#if LDC_LLVM_VER >= 700
+  // On Android, we depend on a custom TLS emulation scheme implemented in our
+  // LLVM fork. LLVM 7+ enables regular emutls by default; prevent that.
+  if (triple.getEnvironment() == llvm::Triple::Android) {
+    targetOptions.EmulatedTLS = 0;
```
joakim-noah: Change these to `false` and `true` instead of numbers and this is exactly my patch.

joakim-noah:
Yes, `-emulated-tls=false` also works, I've been building the native Termux packages with it.

PR #2864, opened 2018-09-30
Implement SysV x64 ABI properly

kinke:
The upstream implementation (`dmd/argtypes.d`) is... incomplete, to put it nicely; there's a number of DMD issues about it. I felt kinda bad when dropping our previous implementation (a few years ago), which was definitely more correct. This is an attempt to revive it, reimplemented from scratch using [this spec](https://www.uclibc.org/docs/psABI-x86_64.pdf), and tries to stay close to the algorithm laid out in the spec.

### DIFF
--- a/gen/abi-x86-64.cpp
+++ b/gen/abi-x86-64.cpp
@@ -59,24 +59,38 @@ LLType *getAbiType(Type *ty) {
   }
 
   // Okay, we may need to transform. Figure out a canonical type:
+  llvm::SmallVector<Type *, 2> argTypes;
 
-  TypeTuple *argTypes = Target::toArgTypes(ty);
-  if (!argTypes || argTypes->arguments->empty()) {
-    return nullptr; // don't rewrite
+  // try to reuse cached arg{1,2}type of StructDeclarations
+  if (ty->ty == Tstruct) {
+    const auto sd = static_cast<TypeStruct *>(ty)->sym;
+    if (sd && sd->sizeok == SIZEOKdone) {
+      if (!sd->arg1type) {
+        return nullptr; // don't rewrite
+      }
+      argTypes.push_back(sd->arg1type);
+      if (sd->arg2type) {
+        argTypes.push_back(sd->arg2type);
+      }
+    }
+  }
+
+  if (argTypes.empty()) {
+    TypeTuple *tuple = Target::toArgTypes(ty);
+    if (!tuple || tuple->arguments->empty()) {
+      return nullptr; // don't rewrite
+    }
+    for (auto param : *tuple->arguments) {
+      argTypes.push_back(param->type);
+    }
   }
 
   LLType *abiTy = nullptr;
-  if (argTypes->arguments->size() == 1) {
-    abiTy = DtoType((*argTypes->arguments->begin())->type);
-    // don't rewrite to a single bit (assertions in tollvm.cpp), choose a byte
-    // instead
-    abiTy = i1ToI8(abiTy);
+  if (argTypes.size() == 1) {
+    abiTy = DtoType(argTypes[0]);
   } else {
-    llvm::SmallVector<LLType *, 4> parts;
-    for (auto param : *argTypes->arguments) {
-      parts.push_back(DtoType(param->type));
-    }
-    abiTy = LLStructType::get(gIR->context(), parts);
+    abiTy = LLStructType::get(gIR->context(),
+                              {DtoType(argTypes[0]), DtoType(argTypes[1])});
   }
 
   return abiTy;
--- a/gen/abi-x86-64.cpp
+++ b/gen/abi-x86-64.cpp
@@ -49,7 +49,6 @@
 #include <utility>
 
 namespace {
-namespace dmd_abi {
 // Structs, static arrays and cfloats may be rewritten to exploit registers.
 // This function returns the rewritten type, or null if no transformation is
 // needed.
@@ -73,19 +72,9 @@ LLType *getAbiType(Type *ty) {
     // instead
     abiTy = i1ToI8(abiTy);
   } else {
-    std::vector<LLType *> parts;
+    llvm::SmallVector<LLType *, 4> parts;
     for (auto param : *argTypes->arguments) {
-      LLType *partType = DtoType(param->type);
-      // round up the DMD argtype for an eightbyte of a struct to a
-      // corresponding 64-bit type
-      // this makes sure that 64 bits of the chosen register are used and thus
-      // makes sure all potential padding bytes of a struct are copied
-      if (partType->isIntegerTy()) {
-        partType = LLType::getInt64Ty(gIR->context());
-      } else if (partType->isFloatTy()) {
-        partType = LLType::getDoubleTy(gIR->context());
-      }
-      parts.push_back(partType);
+      parts.push_back(DtoType(param->type));
     }
     abiTy = LLStructType::get(gIR->context(), parts);
   }
@@ -101,9 +90,6 @@ bool passByVal(Type *ty) {
 
   return argTypes->arguments->empty(); // empty => cannot be passed in registers
 }
-} // namespace dmd_abi
-
-LLType *getAbiType(Type *ty) { return dmd_abi::getAbiType(ty->toBasetype()); }
 
 struct RegCount {
   char int_regs, sse_regs;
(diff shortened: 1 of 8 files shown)


ZombineDev:
Can this upstreamed... pls?

kinke:
That's the goal mid/long-term; I don't know if DMD's backend has to be fixed too.

kinke:
This should fix DMD issues [5570](https://issues.dlang.org/show_bug.cgi?id=5570), [9239](https://issues.dlang.org/show_bug.cgi?id=9239), [13957](https://issues.dlang.org/show_bug.cgi?id=13957) and [14711](https://issues.dlang.org/show_bug.cgi?id=14711) after a quick search, possibly more.

kinke:
[Note to self: only consider types > 16 bytes if target CPU supports AVX.]

kinke:
Regarding the DMD issues mentioned above:
* 14711 is apparently a DMD codegen issue, `toArgTypes()` was already correct (there's even a test for it).
* 13957 applied to LDC as well and is fixed by this (=> now tested in dmd-testsuite).
* 9239 appears to be fixed (for DMD too - I haven't checked the old `toArgTypes()`, just asm).
* 5570 isn't specific and (now) seems to be an umbrella issue (with references to DMD ABI issues on Win64, although it's actually a Linux-specific issue).

gen/abi-x86-64.cpp
```diff
@@ -57,24 +57,38 @@ LLType *getAbiType(Type *ty) {
...
+  if (argTypes.size() == 1) {
+    abiTy = DtoType(argTypes[0]);
   } else {
-    llvm::SmallVector<LLType *, 4> parts;
-    for (auto param : *argTypes->arguments) {
-      parts.push_back(DtoType(param->type));
-    }
-    abiTy = LLStructType::get(gIR->context(), parts);
+    LLType *fieldTypes[] = {DtoType(argTypes[0]), DtoType(argTypes[1])};
+    abiTy = LLStructType::get(gIR->context(), fieldTypes);
```
kinke: This chooses the `bool` overload for `fieldTypes` instead of an `llvm::ArrayRef`... brrr.

kinke:
[Ready for review.]

SSoulaimane:
I was just starting to work on this issue on dmd when I was pointed to this. This is definitely good progress.
> I don't know if DMD's backend has to be fixed too.
slightly.
> AVX512 isn't part of the spec (not sure if there's a newer spec version).
there is in fact the 1.0 spec which covers AVX2 and also accepts structs up 64 bytes to be put in registers [REPO](https://github.com/hjl-tools/x86-psABI) [PDF](https://raw.githubusercontent.com/wiki/hjl-tools/x86-psABI/x86-64-psABI-1.0.pdf)
it seems gcc does implement this newer spec. it checks  if struct size <= 64 rather than <= 32 [gcc/config/i386](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/i386.c)

SSoulaimane:
Why skip windows 64 entirely. I don't how llvm backend works but windows' 64 bit ABI albeit different than SysV' still supports register passing for structs <= 64bits.
> [SPEC](https://docs.microsoft.com/en-us/cpp/build/parameter-passing?view=vs-2017) ...Integer values are passed (in order left to right) in RCX, RDX, R8, and R9...Structs/unions of size 8, 16, 32, or 64 bits and __m64 are passed as if they were integers...

kinke:
> there is in fact the 1.0 spec which covers AVX2

Thx for checking.

> Why skip windows 64 entirely.

Our Win64 ABI implementation doesn't use `toArgTypes()` and is already fully compatible with C++, simple as it is. AFAICT, it only affects
* a TypeInfo_Struct's `m_arg{1,2}` fields & `argTypes()` method, and
* the very special `is(T ArgTypesTuple == __argTypes)` expression (not sure whether that's actually documented).

Considering that they've been SysV-based all this time (and similar garbage for non-x86 architectures), using `null` instead seems like an improvement.
AFAIK, the argtypes being exposed in TypeInfo_Struct is just because it's needed in the varargs code in druntime for POSIX x86_64.

> How about this potential segfault on win64

Would have been caught by CI obviously; patching that line is part of this PR. ;)

SSoulaimane:
Alright I'm convinced about the null return for Win64 targets. The way It looks like the glue and backend for dmd go in the same spirit. there seems to be a check for win64 every where the arg1type (which is dumped from the result of toArgTypes) is used. which smells a bit. ideally it should all be in one place (preferably inside toArgTypes) but that's a different issue.  
The only thing that seems off is that this behavior (null return) does not conform with what the documentation of toArgTygpes() says about return values. Since you renamed the new toArgtypes to a more platform specific name. then maybe it should assert(0) if the target is not SysV 64bit compatible. This behavior is easier to detect if dmd decides to add other targets like arm for example.

kinke:
> The only thing that seems off is that this behavior (null return) does not conform with what the documentation of toArgTygpes() says about return values.

Well, while the actual implementations say that `null` represents `void`, `Target.toArgTypes()` doesn't say anything about null (as the frontend only calls it for structs, which are never 0 bytes long). And the target-specific logic should clearly be in that function.

> Since you renamed the new toArgtypes to a more platform specific name. then maybe it should assert(0) if the target is not SysV 64bit compatible.

I'd rather keep it from accessing `global.params.*` to validate its call itself (plus we don't have the full LDC target info on the D side); the function name is pretty clear, so IMO it's the caller's responsibility to only call it if appropriate.

SSoulaimane:
Ok. but on target.d that
```
        if (global.params.is64bit && global.params.isWindows)
            return null;
        return .toArgTypes(t); 
```
code still makes no sense to me seince ldc calls toArgTypes_sysv_x64 directly.

kinke:
This is just the DMD-specific part from the upstream PR.

SSoulaimane:
you can do:
```
        version (IN_LLVM)
            return null;
        else
            return .toArgTypes(t);
````
?

kinke:
LDC's `Target.toArgTypes()` is implemented in C++, and has to be, due to required target info. That's why it's only declared in D.

SSoulaimane:
Yes that what I mean. but isn't that the purpose of the ```private bool isDMDx64Target();```?
Yes it makes more sense discussing this in the upstream PR.

SSoulaimane:
Just as a side note on the 1.0 spec. I'm not sure if it's worth spending more time on implementing the 1.0 spec (although the changes are trivial, it's really up to you) for the following reason: from my examination of the code generated by both gcc (trunk) and clang (trunk) (I used compiler explorer website) it looks like:
(1) they both follow the same rules.
(2) their interpretation is not literally correct (or at least that's how it seems to me) specificaly on the subrule (c) of the rule (5) for the classification of aggregates which says:
> 5. Then a post merger clean up is done:
> ...
>  (c) If the size of the aggregate exceeds two eightbytes and the first eight-byte isn’t SSE or any other eightbyte isn’t SSEUP, the whole argument is passed in memory.

My observation is:
(A) What I understand form the spec is that if a struct has 4 doubles which makes it 4 eightbytes  long then it should be passed in registers. both gcc and clang push it on the stack.
(B) Another interesting observation is that if a struct has 2 double or 4 floats meaning that it is 2 eightbytes long, it is passed by registers but it is never put in a one single AVX register even with the -mavx and -mavx2 and all the -mavx512* flags on. instead, the first half is on xmmN and the other half on xmm(N+1) a struct never gets in ymmN nor in zmmN. However this is not true for SSE types. if you turn on the -mavx switch then AVX types do get stored in AVX registers, for example __m256 will go to ymmN. but for __m512 the -mavx2 is not enough apparently, but instead you need one of the -mavx512* switches then it would go into one of the zmmN regiters.

To be frank I was expecting a similar behavior as this is an natural decision to keep ABI compatibility with pre AVX(2) CPUs. I don't know how LDC does with this but this definitely something to I'll keep in mind for DMD.
If you want to test for yourself here is the test bench I used [LINK](https://godbolt.org/z/InW2rm)

kinke:
> What I understand form the spec is that if a struct has 4 doubles which makes it 4 eightbytes long then it should be passed in registers. both gcc and clang push it on the stack.

Nope, that's correct behavior. I initially thought the same - 'hey, very nice, fully exploiting the register widths' - but no, every double field gets its own vector register. The only way a 256-bit struct is passed in an AVX register is if it wraps a single 256-bit vector. That's the only way `SSEUp` makes it to the final classification for an eightbyte; otherwise, every eightbyte with a single double or two floats starts a new `SSE` and is thus passed in its own register.

SSoulaimane:
Yep you're right that's the correct behavior. I was mislead by that statement.

PR #2866, opened 2018-10-01
Fix target_traits.d test: Pentium is a 32bit CPU

JohanEngelen:
(need to fix commit msg when pulling)


JohanEngelen:
(this fix was needed for LLVM 8 btw)

kinke:
I thought so, after seeing your druntime branch. ;)

Issue #2865, opened 2018-09-30
Subtracting a nonzero integer from a function pointer casted to a 64-bit integer causes a segmentation fault


tuxcrafting:
LDC2 version: 1.11.0  
LLVM version: 6.0.1

Example code:
```d
void main() {
    long n = cast(long)&f - 10;
}
void f() {}
```

This happens no matter the signature of the function, and only when casting to a 64-bit integer type, signed or unsigned.

Stack trace:
```
LLVMSymbolizer: error reading file: No such file or directory
#0 0x00007fde1cb4d7cb llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/usr/lib/libLLVM-6.0.so+0x8517cb)
#1 0x00007fde1cb4b4be llvm::sys::RunSignalHandlers() (/usr/lib/libLLVM-6.0.so+0x84f4be)
#2 0x00007fde1cb4b72b (/usr/lib/libLLVM-6.0.so+0x84f72b)
#3 0x00007fde1c2de3c0 __restore_rt (/usr/lib/libpthread.so.0+0x123c0)
#4 0x00007fde1c9d743a llvm::ConstantInt::destroyConstantImpl() (/usr/lib/libLLVM-6.0.so+0x6db43a)
#5 0x0000556af1304aed (ldc2+0x663aed)
#6 0x0000556af12fbc4f (ldc2+0x65ac4f)
#7 0x0000556af1301a56 (ldc2+0x660a56)
#8 0x0000556af12fbc4f (ldc2+0x65ac4f)
#9 0x0000556af12b5c31 (ldc2+0x614c31)
#10 0x0000556af12b925b (ldc2+0x61825b)
#11 0x0000556af1305d41 (ldc2+0x664d41)
#12 0x0000556af12fd14f (ldc2+0x65c14f)
#13 0x0000556af12e7fa5 (ldc2+0x646fa5)
#14 0x0000556af12e9a14 (ldc2+0x648a14)
#15 0x0000556af12e77f4 (ldc2+0x6467f4)
#16 0x0000556af12abb18 (ldc2+0x60ab18)
#17 0x0000556af12a18b4 (ldc2+0x6008b4)
#18 0x0000556af12c2b9e (ldc2+0x621b9e)
#19 0x0000556af13787a7 (ldc2+0x6d77a7)
#20 0x0000556af134c58a (ldc2+0x6ab58a)
#21 0x0000556af1251933 (ldc2+0x5b0933)
[...]

PR #2867, opened 2018-10-02
Fix issue #2865

kinke:
Functions, labels etc. aren't sized; respect that when coming across a `SymOffExp` (and checking whether we can elide a pointer cast to `i8*` for the GEP).


JohanEngelen:
Perhaps we should also test the generated LLVM IR to verify that the value is calculated correctly?

kinke:
My plan for v1.12 final: this + v2.081.1 when it is released.

PR #2870, opened 2018-10-05
Simplify DSO registration with druntime

kinke:
* The explicit reference to the ModuleInfo should be obsolete, it's added to the `llvm.used` array.
* Only keep a single pair of `ldc.dso_{c,d}tor` per DSO (pointing to a common `ldc.register_dso()` function) instead of a pair per linked-in D object file. This simplifies `ldc.register_dso()` and allows to get rid of the explicit ctor/dtor functions and the `ldc.dso_initialized` global.

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -524,8 +524,10 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
 void ArgsBuilder::addLinker() {
   if (!opts::linker.empty()) {
     args.push_back("-fuse-ld=" + opts::linker);
-  } else if (global.params.isLinux && opts::isUsingThinLTO()) {
-    // default to ld.gold for ThinLTO on Linux due to ld.bfd issues (see #2278)
+  } else if (global.params.isLinux) {
+    // Default to ld.gold on Linux due to ld.bfd issues with ThinLTO (see #2278)
+    // and older bfd versions stripping llvm.used symbols (e.g., ModuleInfo
+    // refs) with --gc-sections (see #2870).
     args.push_back("-fuse-ld=gold");
   }
 }
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -237,39 +237,21 @@ llvm::Function *buildGetTLSAnchor() {
 /// {c, d}tors to invoke _d_dso_registry.
 ///
 /// Pseudocode:
-/// void ldc.register_dso(bool isShutdown, void* minfoUsedPointer) {
-///   if (dsoInitialized == isShutdown) {
-///     dsoInitialized = !isShutdown;
-///     auto record = {1, dsoSlot, minfoBeg, minfoEnd[, getTlsAnchor],
-///       minfoUsedPointer};
-///     _d_dso_registry(cast(CompilerDSOData*)&record);
-///   }
+/// void ldc.register_dso() {
+///   auto record = {1, dsoSlot, minfoBeg, minfoEnd[, getTlsAnchor]};
+///   _d_dso_registry(cast(CompilerDSOData*)&record);
 /// }
 ///
 /// On Darwin platforms, the record contains an extra pointer to a function
 /// which returns the address of a TLS global.
-llvm::Function *buildRegisterDSO(RegistryStyle style,
-                                 llvm::Value *dsoInitialized,
-                                 llvm::Value *dsoSlot, llvm::Value *minfoBeg,
-                                 llvm::Value *minfoEnd) {
-  llvm::Type *argTypes[] = {llvm::Type::getInt1Ty(gIR->context()),
-                            llvm::Type::getInt8PtrTy(gIR->context())};
-  const auto fnType = llvm::FunctionType::get(
-      llvm::Type::getVoidTy(gIR->context()), argTypes, false);
+llvm::Function *buildRegisterDSO(RegistryStyle style, llvm::Value *dsoSlot,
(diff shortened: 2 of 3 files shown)


kinke:
Number of global {c,d}tors for a Phobos hello-world with LDC v1.11: 143
With this: 2

=> Eliding 141 useless ctor/dtor calls on startup/shutdown.

kinke:
Hmm, Travis' ld v2.24 apparently strips the ModuleInfos; CircleCI (also based on Ubuntu 14.04) works with gold.

Mac is looking bad too.

kinke:
It's working with Ubuntu 18.04's ld v2.30 at least, otherwise I wouldn't have opened this PR, so older versions of ld not respecting `llvm.used` may likely be a bug.

Do you know if ld64 is doing something similar implicitly? AFAIK we only use `--gc-sections` for Linux targets.

kinke:
Btw how would you guys feel about dropping LLVM support for < 3.9? IIRC, that's the minimum version required for MSVC and LTO, and there are 61 occurrences of `#if LDC_LLVM_VER >= 309` in our C++ code.

kinke:
AFAICT, the problem for macOS is that the global {c,d}tors aren't folded to a single pair. From SSH session:

```
$ size -x -l -m hello
...
Segment __DATA: 0x49000 (vmaddr 0x100235000 fileoff 2314240)
	...
	Section __mod_init_func: 0x480 (addr 0x10023bb60 offset 2341728) // ctors
	Section __mod_term_func: 0x480 (addr 0x10023bfe0 offset 2342880) // dtors
	...
	Section .minfo: 0x480 (addr 0x10027b468 offset 2602088) // ModuleInfos
```

So there are 0x480 / 8 = 144 ModuleInfos, and unfortunately the same number of ctors/dtors, while there should be a single one for those. As each `_d_dso_registry()` call toggles between registering and unregistering the DSO, it should be 72x registered and 72x unregistered at startup...

[The symptom is quite a different one, an LDC segfault in `StringTable.update()` when registering the predefined versions - if LDC is built with itself.]

kinke:
Further findings: the 'workaround' leads to the same symptom on Linux, apparently because the manual {c,d}tor refs are stripped - makes sense. So I added them to `llvm.used` => ending up with *all* ctors again. So I put them into a Comdat - and that finally works again. But Mach-O of course doesn't support them, so I have no idea how to make this simpler scheme work on Mac.

kinke:
Emitting the refs manually works for macOS, they are folded correctly. Without putting them into `llvm.used`, they get stripped when using `-O` (I don't know where that comes from). Luckily, Comdats appear not to be required to prevent duplicate refs.

JohanEngelen:
> Btw how would you guys feel about dropping LLVM support for < 3.9?

Fine with me. LLVM 3.9 was released 2 years ago. (it's perhaps nice that we support so many LLVM versions, but it is a burden and I don't know how many people actually make use of that; clang only supports one LLVM version ;-) )

JohanEngelen:
We should write down clearer why the default linker is now `ld.gold` and should very clearly write down that using another linker may break the binary (which is what I read here). Do I understand it correctly that this PR doesn't just set the default to ld.gold, it actually makes ld.gold a requirement? `ld.gold` is not always available, so in that case people cannot use LDC anymore?
The Ubuntu Trusty package `gcc-aarch64-linux-gnu` for cross compiling to AArch64 only contains `ld.bfd` 2.24. It's confusing to see the error message `collect2: fatal error: cannot find 'ld'` because of the now explicit `fuse-ld=gold` passed to gcc. The fix `-linker=bfd` is not obvious. Worse: does it break binaries?

kinke:
> Worse: does it break binaries?

It does under following circumstances OTOH:
- working around missing gold with `-linker=bfd`
- bfd is an apparently buggy older one, stripping the ModuleInfos
- ModuleInfos are required, e.g., to invoke module ctors

> It's confusing to see the error message collect2: fatal error: cannot find 'ld' because of the now explicit fuse-ld=gold passed to gcc.

[What makes it worse is that the error msg doesn't say `cannot find ld.gold`, which would make it easier to track down.]

---

To me, the advantages of this PR are clear. If older bfd really ignores `llvm.used` (whatever that lowers down to in asm), then it's not just an issue for ModuleInfos, but for the `@assumeUsed` UDA too. If we still have to support it, I'd prefer looking into the bfd issue and find a workaround for it, instead of going the quick'n dirty route of (temporarily) reverting this PR.

kinke:
> The Ubuntu Trusty package gcc-aarch64-linux-gnu for cross compiling to AArch64 only contains ld.bfd 2.24.

gold is included in the `binutils-aarch64-linux-gnu` packages for xenial (16.04) and bionic (18.04).

kinke:
Debian also includes gold in its oldest supported `binutils-aarch64-linux-gnu` package (Jessie; binutils-2.25).

kinke:
Trusty's `binutils-arm-linux-gnueabi` does include gold as well (as does the native armhf package).

It's missing in AArch64's native `binutils` package too though. It does look like an isolated special case to me though.

JohanEngelen:
It is very nice that you did the digging to get data about availability of ld.gold, thanks.
I still think that it's better to change the wording to something like "due to a bug in the `bfd` linker, `ld.gold` is a now requirement on Linux, which can be worked around using `-linker` ".
I do read the release notes (often several versions back)! I find them more useful for reference (when experiencing troubles after updating the compiler) than for news flash bits. (it's a compiler after all)

PR #2875, opened 2018-10-14
druntime: Test core.elf refactoring

kinke:
Also testing whether all CI services support testing submodule PRs from non-official repos. **Edit**: Okay, CircleCI doesn't support it, so nope, definitely not an option.


kinke:
Btw, from what I gathered, file I/O seems to be mandatory when looking up ELF sections. That's costly, so I gave up the original plan of replacing the magic linker symbols for the `.minfo` section.

joakim-noah:
How costly can it be? If it's only done once when initializing druntime, it can't be that bad, can it?

kinke:
I haven't measured it, but opening the file for each D binary plus (quite) a bunch of mmap and munmap calls, all syscalls, and string comparisons (for each section, until we finally find the matching one) etc. doesn't seem very nice.

kinke:
What is it again that you need for Android TLS? The `.tdata` section range, right? Does the linker not support the magic `__{start,stop}_<sectionName>` symbols?

joakim-noah:
The `.tdata` range and the `.tbss` size are needed for emulated TLS. `ld.bfd` puts out start and stop delimiters for the `.bss` and `__minfo` sections, that's about it.

joakim-noah:
Yeah, I thought of changing the section name in LLVM when responding now too, I'll see how bad it'd be to patch.

kinke:
`__minfo` is a custom section of ours, so the magic symbols should be generically supported. If they are missing for default `.t{bss,data}`, the simplest variant seems to rename them in LLVM.

kinke:
Now upstream: https://github.com/dlang/druntime/pull/2330

kinke:
An updated version of this has landed in branch `ldc-merge-2.091`.

PR #2883, opened 2018-10-25
[ltsmaster] Emit constructors as .init_array instead of .ctors when appropriate

myfreeweb:
AArch64 only supports `.init_array`, for example. The reason this wasn't discovered earlier is that BFD and gold linkers automatically convert `.ctors` to `.init_array` as a performance optimization.

Based on clang code: https://github.com/llvm-mirror/clang/blob/5897428cd24e2deefbcc6f6744c0d7d233aa6747/lib/Driver/ToolChains/Gnu.cpp#L2598-L2620

https://github.com/ldc-developers/druntime/pull/146

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -587,6 +587,23 @@ llvm::TargetMachine *createTargetMachine(
     break;
   }
 
+  // Taken from clang's lib/Driver/ToolChains/Gnu.cpp
+  if (triple.getArch() == llvm::Triple::aarch64 ||
+      triple.getArch() == llvm::Triple::aarch64_be ||
+      (triple.getOS() == llvm::Triple::FreeBSD &&
+       triple.getOSMajorVersion() >= 12) ||
+      triple.getOS() == llvm::Triple::NaCl ||
+      (triple.getVendor() == llvm::Triple::MipsTechnologies &&
+       !triple.hasEnvironment()) ||
+      triple.getOS() == llvm::Triple::Solaris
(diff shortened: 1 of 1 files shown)


kinke:
Thx; too bad that ugly piece of code isn't in LLVM itself.

myfreeweb:
> it apparently defaults to .init_array on Linux

LLVM itself? hm, why does clang then have a check for *the GCC installation* on Linux:

```c++
(getTriple().getOS() == llvm::Triple::Linux &&
       ((!GCCInstallation.isValid() || !V.isOlderThan(4, 7, 0)) ||
getTriple().isAndroid()))
```

kinke:
> LLVM itself?

Yep, as can be seen by compiling with `-output-s -mtriple={x86_64,aarch64}-linux-gnu`; the object file sections are `.{init,fini}_array`, i.e., not converted by the linker.

> why does clang then

Don't know, don't care. ;)

JohanEngelen:
This looks OK to me, besides a missing test case. No need to test all the different possible triples, but a test with one triple and checking that setting that targetOption flag actually does the right thing would be good.
Any other concerns @kinke ?

kinke:
Nope, no concerns from my side. We must not forget to cherry-pick it to master afterwards.

JohanEngelen:
@myfreeweb Can you fix the build errors in the CI? (just use #ifdef to remove the checking of triples unknown to older LLVMs)

myfreeweb:
Now passes on travis. Semaphore seems dead

PR #2886, opened 2018-10-27
Windows: Make LDC package fully self-sufficient (get rid of MSVC dependency)

kinke:
When experimenting with the MinGW-w64 based libs, there was a problem (well, one of many!) - an undefined reference to function `_(_)chkstk`, emitted by LLVM for stack probing (e.g., after an `alloca`) and not provided by MinGW-w64.

Initially, I used the symbol exported by ntdll.dll; normal user apps aren't supposed to depend on that system DLL though, and it also led to duplicate symbols for 32-bit code.

After some digging, I found that LLVM's builtins compiler-rt library contains 2 implementations (1 MS-compatible, 1 for MinGW/Cygwin, both with different names than the MS one). Our library didn't contain it though, so I digged deeper.

I found the culprit in the builtins CMake script; when using an MSVC toolchain, the builtins lib on Windows only contains very few C files and none of the various ASM ones, because they use AT&T syntax.

So I decided to try building LLVM & compiler-rt with clang instead. That works fine, but I had to fix the CMake script, as the ASM files are excluded for MS-*compatible* toolchains, incl. clang. So I forked
compiler-rt to fix this and some more annoyances (CMake fixes only, no code changes).

When compiling LLVM with clang-cl (`cl.exe`-compatible wrapper), the LLVM C++ compile flags include some clang-specific ones, so building LDC with MSVC didn't work. While that can be fixed at some point (stripping the clang-specific ones), I decided to build LDC (and the C/C++/ASM files in our default libs) with clang too. That required some adaptations, as it started out with > 5k warnings (and crashes in druntime). ;)

So this commit switches to clang for compiling C/C++/ASM, but still uses Microsoft's linker.

---

Some of the required adaptations are already in master (https://github.com/ldc-developers/ldc/commit/7b7e0e927c0a9adfb90ea3bcb11eb5bb9ac0af86, https://github.com/ldc-developers/ldc/commit/c0d269d5d9ed9586f114fceb1fe52e5f2239e02b, https://github.com/ldc-developers/druntime/commit/b2c58c7fc4a14e7b83a46713eb9b2109ec3ec68a, https://github.com/ldc-developers/druntime/commit/9250b1cffe3557f04b40263cf0a49271061835c2, https://github.com/ldc-developers/druntime/commit/4afc0547278e4dae1941101f5d39f76374dce144). I appended them to the merge-2.083 PR in order to make sure everything still works with the regular MSVC toolchain.

The compiler-rt changes will be pushed as soon as this works.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -168,6 +168,20 @@ after_test:
         }
         # Now rename the installation dir to test portability.
         ren "$ldcInstallDir" ldc2-install
+  # Include MinGW-w64-based libs
+  - ps: |
+        cd c:\projects
+        $ldcInstallDir = 'c:\projects\ldc2-install'
+        mkdir mingw-w64-libs
+        cd mingw-w64-libs
+        appveyor DownloadFile "https://github.com/ldc-developers/mingw-w64-libs/releases/download/v6.0.0-rc.1/mingw-w64-libs-v6.0.0-rc.1.7z" -FileName mingw-w64-libs.7z
+        7z x mingw-w64-libs.7z > $null
+        If ($Env:APPVEYOR_JOB_ARCH -eq 'x64') {
+          cp -r lib64 "$ldcInstallDir\lib\mingw"
+        } Else {
+          cp -r lib32 "$ldcInstallDir\lib\mingw"
+        }
+        cd ..
   # Hello world integration test with LTO (x64 only)
   - ps: |
         If ($Env:APPVEYOR_JOB_ARCH -eq 'x64') {
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -87,7 +87,7 @@ install:
           #echo 'Using LLVM with enabled assertions'
           #$assertsSuffix = '-withAsserts'
         }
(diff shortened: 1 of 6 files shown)


kinke:
AppVeyor retriggered due to new prebuilt Win32 LLVM package (see https://github.com/ldc-developers/llvm/commits/mingw-w64).

The MSVC detection logic still needs to be adapted in order to fall back to the MinGW-w64 libs (and those libs be bundled in the AppVeyor script); a `std.stdio` hello-world & `runnable/eh[2].d` now work on both Win32/64 and with both LLD and MS linker.

kinke:
Alright, this should be finally complete now. `ldc2 -run hello.d` working fine in a naked cmd.exe.

thewilsonator:
Nice!

kinke:
Still not blazingly fast, but the lit-tests runtime went down from ~105 secs to ~45 secs for AppVeyor, i.e., saving one minute and decreasing the chance of timeouts. The clang-builds seem to be somewhat faster too, about ~3 minutes for the full job; I expected a speed up, but hoped it'd be greater. ;)

PR #2892, opened 2018-10-31
Pull in druntime fixes for Android

joakim-noah:
First two druntime commits submitted for review upstream at dlang/druntime#2325 and the third is LDC-specific. This should get Android building again for the upcoming 1.13 beta, can always revise it later if needed.


kinke:
You'd save me some extra work if you managed to get those druntime PRs be merged in time into `stable` (generally, not related to this specific PR). I don't see any connection to the TLS range, so by keeping those things separate, the upstream PR should have been merged already (as mentioned, ideally rebased onto stable).
While at it, you could also change the `char[16]` to `byte[16]`.

And please prefer branches on your repo instead of the official one, it's just extra work for CI (2x Travis and especially AppVeyor).

kinke:
Btw, I take it you still haven't gotten round to testing the TLS section renaming?

joakim-noah:
>You'd save me some extra work if you managed to get those druntime PRs be merged in time into stable

You mean work reviewing this PR? Until recently, I used to get them merged upstream then cherrypick them into the ldc fork, can go back to that after this PR.

>While at it, you could also change the char[16] to byte[16].

I want to handle that more generally, as I found other cases where it hits too.

>And please prefer branches on your repo instead of the official one, it's just extra work for CI (2x Travis and especially AppVeyor).

I don't know what you mean, any pull isn't tested the same? Anyway, I haven't forked ldc on github.

>Btw, I take it you still haven't gotten round to testing the TLS section renaming?

Didn't bother with it because I still need to pass the `.data` and `.bss` sections to the GC, which means hunting for encircling delimiters that work across all linkers (`__bss_end__` is only put out by `ld.bfd`, not `gold` or `lld`), or simply reading the ELF headers.

I'm doing the latter, currently testing it with shared libraries.

joakim-noah:
>Nope, I mean merge conflicts

OK, I will target upstream stable from now on.

>Some CI services don't run per-commit, but per branch and PR, i.e., twice if you use an official branch.

Didn't know there was a difference, will fork ldc then.

>The normal rt/sections_elf_shared.d does that by simply checking the binary's segments (not sections)

Yeah, I know, I was planning on raising the issue of whether that's too much when submitting my commit upstream to read the ELF headers on Android. We can hash that out there.

>In the long run, we should probably try to get rid of the Android-specific sections file and incorporate remaining extra logic in sections_elf_shared.d.

I agree, especially since Android should support loading multiple D shared libraries too. However, I like having this simpler one around, if nothing else as an example for porters who can't use the much larger `rt.sections_elf_shared`.

This pull's ready to go, I'll change the way I submit pulls from now on.

joakim-noah:
Finally tried renaming the TLS sections in LLVM to `__tdata`/`__tbss`, but ran into an issue where the linker started putting out separate sections for every TLS symbol, ie `__tdata._D2rt4util3utf18__unittest_L296_C1FZ2s1Aya` in the final executable. Presumably these were coalesced by the `ld.bfd` linker for the recognized `.tdata`/`.tbss` sections before, but we'd have to do that ourselves if we change the section names, either with linker scripts or changing how ldc emits those sections.

I don't think it's worth the hassle, easier to just read the ELF headers.

kinke:
> > The normal rt/sections_elf_shared.d does that by simply checking the binary's segments (not sections)
>
> Yeah, I know, I was planning on raising the issue of whether that's too much when submitting my commit upstream to read the ELF headers on Android. We can hash that out there.

Oh, you mean there may be more sections in those segments? [I have no idea about that.]

> Presumably these were coalesced by the ld.bfd linker for the recognized .tdata/.tbss sections before, but we'd have to do that ourselves if we change the section names, either with linker scripts or changing how ldc emits those sections.

Yep, that seems likely, as the object files contain separate sections. If you want to continue with manual sections reading, that's fine, another alternative would be hacking LLD (to just emit the magic symbols without section renaming) and defaulting to `-link-internally` for Android.

kinke:
Okay, so the magic symbols aren't emitted because of the leading `.` in the section name, see https://github.com/ldc-developers/lld/blob/ldc-release_70/ELF/Writer.cpp#L1745-L1757. So by minimally patching LLD, we could get it to emit magic `__{start,stop}` symbols for `.t{data,bss}` too. Overall, this is surely simpler and faster than parsing the sections at runtime in druntime. But `-link-internally` would be required, not sure if that'd be acceptable.

joakim-noah:
>Oh, you mean there may be more sections in those segments? [I have no idea about that.]

Yes, but thinking about it some more, not sure it matters, as it's globally shared data and doesn't need to be copied to every thread like thread-local data. I'll check and report my results in that upstream pull.

> If you want to continue with manual sections reading, that's fine, another alternative would be hacking LLD (to just emit the magic symbols without section renaming) and defaulting to -link-internally for Android.

I want to get the code to read ELF headers in and remove [the `ld.bfd` and other emulated TLS requirements](https://wiki.dlang.org/Build_D_for_Android#Changes_for_Android), then see about refining it later. I've long wanted to remove the `clang`/`ld` package requirement for the native ldc on Android and switch it to `-link-internally` by default: I'll start there and work back to the cross-compiler.

PR #2894, opened 2018-11-02
Add option '-fvisibility=<hidden|default>' able to hide symbols not marked as 'export' on non-Windows targets. Resolves #2431

### LINKED ISSUE #2431: Stripping all symbols except `export`ed functions
For binary size reason, we use list of symbols to export in final builds.

```
ldc2 -exported_symbols_list list-of-symbols.lst -dead_strip <rest-of-cmdline>
```
These two flags together allow to reduce object code dramatically since all symbols are public by default on POSIX else (about 3x reduction IIRC).

Now that LDC supports the `export` attribute, would it be possible to avoid to write such a list of symbols, and remove the need for `-exported_symbols_list`? The problem is that it's error prone and the information is already there through `export`.

(edit: I'm not sure if `export` does anything on POSIX though)

MrSmith33:
#2431

### DIFF
--- a/dmd/globals.d
+++ b/dmd/globals.d
@@ -262,8 +262,9 @@ struct Param
         bool fullyQualifiedObjectFiles;
         bool cleanupObjectFiles;
 
-        // Export only symbols marked as 'export' on linux and OSX
-        bool export_only_symbols_marked_export;
+        // Export only symbols marked as 'export' on non-Windows targets
+        // Doesn't affect Windows, since it handles exports via setDLLStorageClass
+        bool exportOnlySymbolsMarkedExport;
 
         // Profile-guided optimization:
         const(char)* datafileInstrProf; // Either the input or output file for PGO data
--- a/dmd/globals.d
+++ b/dmd/globals.d
@@ -262,6 +262,9 @@ struct Param
         bool fullyQualifiedObjectFiles;
         bool cleanupObjectFiles;
 
+        // Export only symbols marked as 'export' on linux and OSX
+        bool export_only_symbols_marked_export;
+
         // Profile-guided optimization:
         const(char)* datafileInstrProf; // Either the input or output file for PGO data
 
--- a/dmd/globals.d
+++ b/dmd/globals.d
@@ -262,10 +262,6 @@ struct Param
         bool fullyQualifiedObjectFiles;
         bool cleanupObjectFiles;
 
-        // Export only symbols marked as 'export' on non-Windows targets
-        // Doesn't affect Windows, since it handles exports via setDLLStorageClass
-        bool exportOnlySymbolsMarkedExport;
-
         // Profile-guided optimization:
         const(char)* datafileInstrProf; // Either the input or output file for PGO data
 
--- a/dmd/globals.h
+++ b/dmd/globals.h
@@ -240,10 +240,6 @@ struct Param
     bool fullyQualifiedObjectFiles;
     bool cleanupObjectFiles;
 
-    // Export only symbols marked as 'export' on non-Windows targets
-    // Doesn't affect Windows, since it handles exports via setDLLStorageClass
-    bool exportOnlySymbolsMarkedExport;
-
     // Profile-guided optimization:
     const char *datafileInstrProf; // Either the input or output file for PGO data
 
--- a/dmd/globals.h
+++ b/dmd/globals.h
@@ -240,6 +240,9 @@ struct Param
     bool fullyQualifiedObjectFiles;
     bool cleanupObjectFiles;
 
+    // Export only symbols marked as 'export' on linux and OSX
+    bool export_only_symbols_marked_export;
+
     // Profile-guided optimization:
     const char *datafileInstrProf; // Either the input or output file for PGO data
 
--- a/dmd/globals.h
+++ b/dmd/globals.h
@@ -240,8 +240,9 @@ struct Param
     bool fullyQualifiedObjectFiles;
     bool cleanupObjectFiles;
 
-    // Export only symbols marked as 'export' on linux and OSX
-    bool export_only_symbols_marked_export;
+    // Export only symbols marked as 'export' on non-Windows targets
+    // Doesn't affect Windows, since it handles exports via setDLLStorageClass
+    bool exportOnlySymbolsMarkedExport;
 
     // Profile-guided optimization:
     const char *datafileInstrProf; // Either the input or output file for PGO data
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -83,8 +83,8 @@ static cl::opt<bool, true>
 
 static cl::opt<bool, true> exportOnlyMarkedExport(
         "export-marked-symbols", cl::ZeroOrMore,
-        cl::desc("Export only symbols marked as 'export' on linux and OSX"),
-        cl::location(global.params.export_only_symbols_marked_export));
+        cl::desc("Export only symbols marked as 'export' on non-Windows targets"),
+        cl::location(global.params.exportOnlySymbolsMarkedExport));
 
 static cl::opt<bool, true> verbose("v", cl::desc("Verbose"), cl::ZeroOrMore,
                                    cl::location(global.params.verbose));
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -81,14 +81,12 @@ static cl::opt<bool, true>
     createSharedLib("shared", cl::desc("Create shared library (DLL)"),
                     cl::ZeroOrMore, cl::location(global.params.dll));
 
-cl::opt<bool> defaultToHiddenVisibility(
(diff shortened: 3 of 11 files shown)


dmd/globals.d
```diff
@@ -262,6 +262,9 @@ struct Param
         bool fullyQualifiedObjectFiles;
         bool cleanupObjectFiles;
 
+        // Export only symbols marked as 'export' on linux and OSX
```
thewilsonator: Also add a note that this is the default behaviour on windows

dmd/globals.h
```diff
@@ -240,6 +240,9 @@ struct Param
     bool fullyQualifiedObjectFiles;
     bool cleanupObjectFiles;
 
+    // Export only symbols marked as 'export' on linux and OSX
```
thewilsonator: ditto

driver/cl_options.cpp
```diff
@@ -81,6 +81,11 @@ static cl::opt<bool, true>
     createSharedLib("shared", cl::desc("Create shared library (DLL)"),
                     cl::ZeroOrMore, cl::location(global.params.dll));
 
+static cl::opt<bool, true> exportOnlyMarkedExport(
+        "export-marked-symbols", cl::ZeroOrMore,
+        cl::desc("Export only symbols marked as 'export' on linux and OSX"),
```
thewilsonator: Ditto

driver/main.cpp
```diff
@@ -558,6 +558,11 @@ void parseCommandLine(int argc, char **argv, Strings &sourceFiles,
 
   global.params.hdrStripPlainFunctions = !opts::hdrKeepAllBodies;
   global.params.disableRedZone = opts::disableRedZone();
+
+  // On windows this is already a default behavior
```
thewilsonator: Nit `already the default behavior`

dmd/globals.d
```diff
@@ -262,6 +262,9 @@ struct Param
...
+        bool export_only_symbols_marked_export;
```
JohanEngelen: please use lowerCamelCase such that it matches our coding style.

driver/main.cpp
```diff
@@ -558,6 +558,11 @@ void parseCommandLine(int argc, char **argv, Strings &sourceFiles,
...
+  if (global.params.isWindows) {
```
JohanEngelen: This may bite us in the future. Better to have the option mean what the behaviour will look like to the user. If Windows already does this by default, we should not internally change this option to `false` (which would mean the opposite would happen!), but instead change the way this option is used to achieve that behaviour.

gen/declarations.cpp
```diff
@@ -308,6 +308,12 @@ class CodegenVisitor : public Visitor {
           gvar->setDLLStorageClass(LLGlobalValue::DLLExportStorageClass);
         }
 
+        // Hide non-exported symbols
+        if (global.params.export_only_symbols_marked_export &&
+          !decl->isExport()) {
+          gvar->setVisibility(LLGlobalValue::HiddenVisibility);
```
JohanEngelen: If this is causing trouble on windows, then the check for windows should be here, instead of setting the option to `false` for Windows.

dmd/globals.d
kinke: I'd rather not have the flag in the `Param` struct at all; it's a pure codegen flag and doesn't affect the frontend.

thewilsonator: And if it is there it should be in `version(IN_LLVM)`.

driver/main.cpp
kinke: +1. I'd call it internally something like `defaultToHiddenVisibility`; the gcc/clang option is `-fvisibility=hidden`, so our cmdline option could be named similarly.

gen/declarations.cpp
kinke: We use the DLL storage classes for Windows; I guess the visibility doesn't really matter.

kinke:
This basic approach is more or less a port of the current Windows approach, i.e., allowing selective exports of functions and globals.

It doesn't handle any of the special symbols (TypeInfos, vtables, init symbols...); only exporting those for aggregates marked with `export` would be a follow-up we shouldn't forget.

driver/cl_options.cpp
kinke: The description should say that it only affects non-Windows targets (as it's enforced for Windows, not just the default, and doesn't just apply to Linux and Mac).

MrSmith33:
* What is the correct directory for this test?
* Is it ok to use external tools for testing? (namely `nm` utility to dump exported symbols)
`// RUN: ldc2 %s.d -betterC -shared -export-marked-symbols && nm lib%s.so`

kinke:
* `tests/codegen`. There's an `export.d` already, testing the Windows dll{ex,im}port stuff (testing IR, not binary code).
* `nm` is okay, as long as it works for CI for both Linux and Mac. Use `XFAIL: Windows` as Windows workaround.
* Pls move the flag out of `Params` and expose it in `driver/cl_options.h`.

tests/codegen/export_marked_symbols_thin_lto.d
```diff
@@ -0,0 +1,13 @@
+// Test if -export-marked-symbols works with thin LTO
+
+// RUN: ldc2 %S/inputs/export_marked_symbols_thin_lto_lib.d -c -export-marked-symbols -of=%t1.o
```
JohanEngelen: This will need a `// REQUIRES: LTO` and `-flto=thin`

tests/codegen/export_marked_symbols1.d
```diff
@@ -0,0 +1,25 @@
+// Test if passing -export-marked-symbols hides all unexported symbols
+
+// RUN: ldc2 %s -betterC -shared -export-marked-symbols -of=lib%t.so
+// RUN: nm lib%t.so | FileCheck %s
+
+// UNSUPPORTED: Windows
```
JohanEngelen: Nit: can you move this to the top? It's how we do it in other test files (requirements first), and it took me a while to see that you disabled this test on Windows. thanks!  (also in the other test file)

kinke:
Thanks for figuring out the `UNSUPPORTED` on your own, I forgot we had it. ;)

For Mac, `nm` apparently needs `-g` (which should work on Linux too) to only display the public symbols.

tests/codegen/export_marked_symbols2.d
```diff
@@ -0,0 +1,26 @@
...
+
+// RUN: ldc2 %s -betterC -shared -of=lib%t.so
+// RUN: nm lib%t.so | FileCheck %s
+
+// CHECK: test__exportedFunDef
+// CHECK: test__exportedVarDef
+// CHECK: test__nonExportedFunDef
+// CHECK: test__nonExportedVarDef
+// CHECK-NOT: test__nonExportedFunDecl
+// CHECK-NOT: test__nonExportedVarDecl
```
kinke: The problem here is that you're only checking that the 2 latter symbols don't exist in the `nm` output *after* `test__nonExportedVarDef`. I don't think replacing `CHECK` with `CHECK-DAG` helps in that case; see https://llvm.org/docs/CommandGuide/FileCheck.html.
But apparently `nm` sorts alphabetically by default, so the list here should too.

kinke:
Thx dude, lgtm now. There's only the bikeshedding about the name of the cmdline option - I'm still in favor of gcc/clang-compatible `-fvisibility=<hidden|default>` and internally calling the flag `defaultToHiddenVisibility`. Interested in hearing other opinions.

JohanEngelen:
terminology-wise it is smartest to use what others know about. LLVM and C seem to call the attribute "hidden visibility", so that'd have my vote too.  (Also for buildsystems, the more LDC looks like a C compiler, the better)

MrSmith33:
Sounds good. Would `flto` and `useDeprecated` option implementations be used as a guide?

kinke:
Something like:
```C++
cl::opt<bool> defaultToHiddenVisibility(
    "fvisibility", cl::ZeroOrMore, cl::desc("Default visibility of symbols (not relevant for Windows)"),
    clEnumValues(
        clEnumValN(false, "default", "Default: export all symbols"),
        clEnumValN(true, "hidden", "Hidden: only export symbols marked with 'export'")
    ));
```

MrSmith33:
Somehow clEnumValues doesn't work for opt<bool>

kinke:
[I added a fixup commit, `opt<ubyte>` does the trick.]

MrSmith33:
@kinke What is the plan on this PR?

kinke:
It'll be part of v1.13, don't worry. ;)

p0nce:
Very happy about this outcome :)

patrickkh7788:
@p0nce 
I try use -fvisibility=hidden with non-shared binary, it seems not working (test on mac-osx)?  

I am not sure it should work for non-shared binary, but if so we can made binary more small.

PR #2895, opened 2018-11-03
WIP: Add DI namespaces & use them for templates and C++ namespaces

kinke:
And make sure the template instance names include the template parameters, a regression after #2826.

Also prevent `TemplatedStruct!(...).TemplatedStruct` names, just use `TemplatedStruct!(...)` instead.


kinke:
No tests yet (just manual ones in Visual Studio, looking good so far); just to let you know that the regression is being worked on and will be fixed in v1.13 final. ;)

kinke:
Test added (cdb only, gdb too would be nice), and of course some refactoring, as DIBuilder really needed it badly.

kinke:
I tested this manually with gdb 8.1 (Ubuntu 18.04). Unlike VS/cdb, it uses the mangled name for functions, so there's no change wrt. displayed function names.
Btw, types nested in a function don't show up as nested, e.g., VS' `scopes_cdb::test::NestedStruct` shows up as `scopes_cdb.NestedStruct` (no change from master). This PR does fix missing template instance params for gdb typenames (`whatis <var>`) too though: `scopes_cdb.TemplatedNestedStruct` => `scopes_cdb.TemplatedNestedStruct!(S, 3)`.

PR #2896, opened 2018-11-03
Cherry-pick: Emit constructors as .init_array instead of .ctors when appropriate (#2883)

JohanEngelen:
AArch64 only supports .init_array, for example. The reason this wasn't discovered
earlier is that BFD and gold linkers automatically convert .ctors to .init_array
as a performance optimization.

Cherry-pick from issue #2883

### DIFF
--- a/tests/codegen/ctor_initarray_gh2883.d
+++ b/tests/codegen/ctor_initarray_gh2883.d
@@ -1,0 +1,11 @@
+// REQUIRES: target_AArch64
+
+// RUN: %ldc -mtriple=aarch64-unknown-linux -output-s -of=%t.s %s
+// RUN: FileCheck %s < %t.s
+
+// CHECK-NOT: .ctors
+// CHECK-NOT: .dtors
+// CHECK: .section  .init_array
+// CHECK: .section  .fini_array
+
+// No code needed to generate asm for a module.
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -469,6 +469,23 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
     break;
   }
 
+  // Taken from clang's lib/Driver/ToolChains/Gnu.cpp
+  if (triple.getArch() == llvm::Triple::aarch64 ||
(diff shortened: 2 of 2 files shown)


tests/codegen/ctor_initarray_gh2883.d
```diff
@@ -0,0 +1,11 @@
+// REQUIRES: target_AArch64
+
+// RUN: %ldc -mtriple=aarch64-unknown-linux -output-s -of=%t.s %s
```
kinke: This already works for that triple (default for Linux). It's also already working for `aarch64-freebsd`. @myfreeweb: What's the triple for which the default wasn't right?

myfreeweb: It was aarch64-freebsd. I'm not sure how it could be "already working". Definitely wasn't already working on ltsmaster, at least.

JohanEngelen: different LLVM versions?
I tested with explicitly setting `targetOptions.UseInitArray = false;` for AArch64 and then the test fails. Perhaps now it is already set to `true` by LLVM?

kinke: Ah, I guess it's because ltsmaster doesn't use [this](https://github.com/ldc-developers/ldc/blob/master/driver/cl_options-llvm.cpp#L56), i.e., doesn't let LLVM choose sane initial `TargetOptions`. So I guess we don't need this for master.

kinke: Yep: https://github.com/ldc-developers/llvm/blob/ldc-release_70/include/llvm/CodeGen/CommandFlags.inc#L287
So master requires an explicit `-use-ctors` cmdline option to use these (old?) names.

JohanEngelen:
> Ah, I guess it's because ltsmaster doesn't use [this](https://github.com/ldc-developers/ldc/blob/master/driver/cl_options-llvm.cpp#L56), i.e., doesn't let LLVM choose sane initial `TargetOptions`. So I guess we don't need this for master.

So I'll change this PR to only include the testcase.

PR #2907, opened 2018-11-10
Improve core.bitop.bsr(uint) implementation.

JohanEngelen:
This improves x86 codegen from
```
mov eax, edi
bsr rax, rax
```
to
```
bsr eax, edi
```


kinke:
Same thing for `bsf` - I guess the LLVM intrinsic used not to be available for `i32` for 64-bit targets, otherwise enforcing `size_t` makes no sense.

JohanEngelen:
> Same thing for `bsf` 

Doh! of course :)  fixed that one too.

PR #2908, opened 2018-11-11
Debuginfo: Fix Win64 CodeView emission for some by-value params

kinke:
Don't declare some special params as DI references, but as DI locals instead.

Also enable some Win32 checks in the args_cdb test, as they are working now with LLVM 7.


kinke:
@thewilsonator: You seem to have removed your SPIR-V LLVM download => our Travis script needs an update.

thewilsonator:
I'm going to completely redo the SPIRV stuff to use https://github.com/KhronosGroup/SPIRV-LLVM-Translator soon.

Issue #2628, opened 2018-03-26
LLVM 6: CodeView debuginfo regression for outer variables in nested functions


kinke:
From https://github.com/ldc-developers/ldc/pull/2608#issuecomment-372067505:
> Failing debuginfo/nested_cdb.d on Win64 (most likely Win32 too) also appears to be a CodeView debuginfo regression (CodeView support is supposed to be greatly improved by v6); once again related to the GEPs we use for nested variables (inner functions only, i.e., where the GEP stems from the bitcast context pointer argument; no problem in the outer function, where it's a GEP from the frame alloca).

The issue is a wrong number of indirections IIRC, so the values are garbage.

Revert https://github.com/ldc-developers/ldc/pull/2608/commits/f1712b2c39edbb22faec1cfbe8d0e3a1c619206e when it's fixed.

kinke:
Fixed in #2909.

PR #2909, opened 2018-11-11
Debuginfo: Fix nested variables (primarily for CodeView)

kinke:
We've been going back and forth between GEPs and complex DWARF expressions based on the context pointer; this goes back to the latter, as that fixes simple nested variables for CodeView with LLVM >= 6.

I guess it also helps for debuginfos of nested vars with enabled optimizations.


gen/dibuilder.h
```diff
@@ -206,7 +206,7 @@ class DIBuilder {
 
     uint64_t offset =
         gDataLayout->getStructLayout(type)->getElementOffset(index);
-    addr.push_back(llvm::dwarf::DW_OP_plus);
+    addr.push_back(llvm::dwarf::DW_OP_plus_uconst);
```
kinke: LLVM's CodeView code only supports this one (and `DW_OP_constu` for plus *and* minus); e.g., no support for `DW_OP_deref` yet.

kinke:
[This fixes #2628.]

rainers:
Does it work for CodeView? I failed to find a way that allows more than one indirection when evaluating local variables.
I recently added structure debug info definitions to dmd for allocated closures (accessed via `__closptr`) and captured context pointer (renamed from `this` to `__capture`). Mago has now support to add these to the local variables and access them in the watch window.

kinke:
It works for the test case, i.e., if the nested variables are from a single parent frame - with both normal VS (`-gc`) and mago (`-g`). If there are multiple outer frames, it won't work, as that needs a deref.

kinke:
> I failed to find a way that allows more than one indirection when evaluating local variables

Yep, I recently came across this apparent limitation too, in LLVM code, see https://github.com/ldc-developers/ldc/pull/2908/files#diff-3fbb5b92ceb5bf1469f5f3b79ecc1ef9R66. It's the reason LLVM emits some by-value params > 64 bits as DI references (depending on parameter index), to trick the debugger into doing the 2nd dereferencing. This works for VS (which shows `LargeValue %`), but doesn't with mago.

Btw, another thing that recently broke with mago are base classes, broken since using proper DI scopes. VS still works.

kinke:
[I removed the IR test, as I don't find it to be of particular value, even less so with these changes. Exactly the same code is much more thoroughly tested in practice with the equivalent cdb and gdb tests. There's still much room for extensions, e.g., multiple outer frames, nested functions in an aggregate method (`vthis`)...]

rainers:
> Btw, another thing that recently broke with mago are base classes, broken since using proper DI scopes. VS still works.

Seems broken because the base class uses `::` in the name, e.g. `test::Base`, even without `-gc`.

kinke:
> It only works because the context pointer is enregistered, so an appropriate record can be emitted. Is this enforced somehow? It seems it is not: if I add a function call into nested RCX is destroyed and the nested variables don't show anymore ("optimized away" with VS).

Ah, interesting. The context pointer should usually be in a register for Win64 (not sure about `vthis` etc., haven't looked into that at all yet). Not too long ago, I got rid of dumping it to stack on function entry and loading it every time it is needed, as it seemed horribly inefficient and just more work for the optimizer; I didn't have debugging in mind at all.

> Seems broken because the base class uses :: in the name, e.g. test::Base, even without -gc.

I guess I'll need to investigate the CodeView output then, to check whether that `::` is really emitted into the names (if so, then by LLVM), or whether that's just the 'implicit' concatenation of CodeView scopes.

rainers:
> Ah, interesting. The context pointer should usually be in a register for Win64

It's passed in RCX, but it doesn't stay there once you call another function as the first argument is again in RCX.

> I guess I'll need to investigate the CodeView output then, to check whether that `::` is really emitted into the names (if so, then by LLVM), or whether that's just the 'implicit' concatenation of CodeView scopes.

It's definitely in the object file, as shown by these type records:
```
0x1011 : Length = 46, Leaf = 0x1505 LF_STRUCTURE
	# members = 0,  field list type 0x0000, FORWARD REF, 
	Derivation list type 0x0000, VT shape type 0x0000
	Size = 0, class name = object::Object, unique name = C6Object

0x1012 : Length = 26, Leaf = 0x1203 LF_FIELDLIST
	list[0] = LF_BCLASS, public, type = 0x1011, offset = 0
	list[1] = LF_MEMBER, public, type = T_INT4(0074), offset = 16
		member name = 'b'

0x1013 : Length = 58, Leaf = 0x1505 LF_STRUCTURE
	# members = 1,  field list type 0x1012, 
	Derivation list type 0x0000, VT shape type 0x0000
	Size = 24, class name = nested_cdb::Base, unique name = C10nested_cdb4Base
```

kinke:
A bit ugly, but seems to do the job: https://github.com/ldc-developers/llvm/commit/dc866bbe2548629da17abbfe180bbaa186ff254b

PR #2910, opened 2018-11-12
druntime: Check fiber migration for Linux/AArch64

kinke:
The `core.thread` release unittests are apparently hanging and leading to the ShippableCI timeouts; check if this helps it (the `CheckFiberMigration` hint is shown for the `core.thread-shared` release unittests).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -640,27 +640,57 @@ else()
     endif()
 endif()
 
-function(build_d_executable output_exe compiler_args linker_args compile_deps link_deps)
-    # Compile all D modules to a single object.
-    set(object_file ${output_exe}${CMAKE_CXX_OUTPUT_EXTENSION})
+function(build_d_executable output_exe d_src_files compiler_args linker_args extra_compile_deps link_deps)
     set(dflags "${D_COMPILER_FLAGS} ${DDMD_DFLAGS}")
     if(UNIX)
       separate_arguments(dflags UNIX_COMMAND "${dflags}")
     else()
       separate_arguments(dflags WINDOWS_COMMAND "${dflags}")
     endif()
-    add_custom_command(
-        OUTPUT ${object_file}
-        COMMAND ${D_COMPILER} -c ${dflags} -of${object_file} ${compiler_args}
-        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-        DEPENDS ${compile_deps}
-    )
+
+    set(object_files)
+    if(NOT COMPILE_D_MODULES_SEPARATELY)
+        # Compile all D modules to a single object (single-threaded).
+        set(object_file ${output_exe}${CMAKE_CXX_OUTPUT_EXTENSION})
+        add_custom_command(
+            OUTPUT ${object_file}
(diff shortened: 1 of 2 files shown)


kinke:
Right, this time it was `core.thread-shared` hanging. This may be an AArch64-specific regression with newer druntime, as I don't recall near-systematic Shippable timeouts for `core.thread` when setting it up.

As this ~98% timeout rate is a serious issue for the automated AArch64 packages, I tend to exclude the `core.thread` tests from the unittests for now.

kinke:
Okay, back to 'normal' 1h timeouts again - the Shippable/Packet performance seems to have dropped considerably. I could replace building ltsmaster with a download of a prebuilt AArch64 package, but that would only save nearly 2 minutes. I could also NOT build the debug druntime/Phobos unittests (untested), but I'd prefer keeping them as compilation error checks.

I guess the CMake script could use an option to build the compiler D parts separately, for systems with low single-core performance but many cores.

joakim-noah:
Nice, will be using separate compilation of ldc on Android/AArch64 too.

I think you mean `CRuntime_Glibc`, not `linux`. The latter would affect other libcs like `Musl` or `uClibc` also.

joakim-noah:
Just tried the D separate compilation commit for ldc and it shaved about 23% off my build time, cuz I can actually use all these cores on my smartphone now. :smile:

kinke:
> I think you mean CRuntime_Glibc, not linux, for the druntime change. The latter would affect other libcs like Musl or uClibc also.

I was totally unsure about that. I also have no idea what's controlling this fiber migration, haven't been interested enough to have a look at it yet.

> Just tried the D separate compilation commit for ldc and it shaved about 23% off my build time

It's just a hack for now. For Shippable, `COMPILE_D_MODULES_SEPARATELY=ON` seems to reduce the `ldc2` build time by a very rough 3 minutes, and `COMPILE_ALL_D_FILES_AT_ONCE=OFF` for (non-unittest) druntime/Phobos an additional very rough 2 minutes - bringing down the overall time from 8-9 to about 3 minutes.

I only use `COMPILE_D_MODULES_SEPARATELY=ON` for the first compilation (bootstrapping the 2nd one), as the resulting binary should be somewhat slower (no implicit cross-D-module optimizations).

`COMPILE_D_MODULES_SEPARATELY=ON` is also helpful for incremental builds, e.g., when hacking the frontend and repeatedly recompiling.

joakim-noah:
> I also have no idea what's controlling this fiber migration, haven't been interested enough to have a look at it yet.

It may be a false positive at this point, as that warning hits me on linux/glibc/x64 also these days, which it didn't use to when he first added it. Or maybe it's a regression and the warning is right, never looked much into how that works, just know it won't work for emulated TLS on Android.

> also helpful for incremental builds, e.g., when hacking the frontend and repeatedly recompiling.

Yep, needed it when tweaking the frontend for an upcoming pull for Android/x64, as I kept waiting for the build to finish when iterating.

kinke:
`CheckFiberMigration` didn't fix `core.thread`; I moved a somewhat cleaner implementation of the new CMake option and the Shippable tweaks to #2914.

PR #2912, opened 2018-11-15
Upgrade to LDC-LLVM 7.0.0-2 and adapt cdb + dmd-testsuite tests

kinke:


rainers:
> I hate breaking bad news, but MSVC v15.9.0 broke the EH code again; for x64, it's just `-mscrtlib=msvcrt` of the 4 variants

https://github.com/ldc-developers/druntime/pull/148

PR #2916, opened 2018-11-17
Optionally enable GC for front-end

kinke:
This isn't looking too bad, although druntime cannot be compiled yet:

Phobos all-at-once compilation, Win64:
```
GC: used bytes = 1707268112
GC: used bytes = 570342720 [after collection]
```

`std.string` unittests:
```
GC: used bytes = 1651332656
GC: used bytes = 361209920
```

`std.algorithm.iteration` unittests:
```
GC: used bytes = 1496288448
GC: used bytes = 644807264
```

`std.regex.internal.tests` unittests:
```
GC: used bytes = 2079689120
GC: used bytes = 292884624
```

It doesn't reduce the peak memory requirements (well, maybe a little bit if LLVM allocates significant amounts during codegen), but reduces the time that memory is claimed, which should especially pay off with `-O`, as much more time will be spent in the LLVM optimizer. This might enable the usage of more CPU cores when building on systems with limited memory, as it reduces the chance of overlapping LDC processes with fat memory requirements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,6 +126,7 @@ script:
     # The -lowmem tests don't work with an ltsmaster host compiler.
     if [[ "$(${DC} --version | head -n 1)" == *0.17.* ]]; then
       rm tests/d2/dmd-testsuite/runnable/{testptrref,xtest46}_gc.d;
+      rm tests/d2/dmd-testsuite/fail_compilation/mixin_gc.d;
     fi
   - DMD_TESTSUITE_MAKE_ARGS=-j3 ctest -V -R "dmd-testsuite"
   # Run defaultlib unittests & druntime stand-alone tests.
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,11 +117,15 @@ script:
   # Run LIT testsuite.
   - ctest -V -R "lit-tests"
   # Run DMD testsuite.
-  # Optimized runnable/testconst.d crashes (in _aaInX) with shared libs on Linux -
-  # only for Travis apparently.
-  -
+  - |
+    # Optimized runnable/testconst.d crashes (in _aaInX) with shared libs on Linux -
+    # only for Travis apparently.
     if [ "${TRAVIS_OS_NAME}" = "linux" ] && [[ "${OPTS}" == *-DBUILD_SHARED_LIBS?ON* ]]; then
       rm tests/d2/dmd-testsuite/runnable/testconst.d;
+    fi
+    # The -lowmem tests don't work with an ltsmaster host compiler.
+    if [[ "$(${DC} --version | head -n 1)" == *0.17.* ]]; then
+      rm tests/d2/dmd-testsuite/runnable/{testptrref,xtest46}_gc.d;
     fi
   - DMD_TESTSUITE_MAKE_ARGS=-j3 ctest -V -R "dmd-testsuite"
   # Run defaultlib unittests & druntime stand-alone tests.
--- a/dmd/mars.d
+++ b/dmd/mars.d
@@ -465,6 +465,12 @@ else
         return result;
     }
 
+    if (params.mixinFile)
+    {
+        params.mixinOut = cast(OutBuffer*)calloc(1, OutBuffer.sizeof);
+        atexit(&flushMixins); // see comment for flushMixins
+    }
+    scope(exit) flushMixins();
     global.path = buildPath(params.imppath);
     global.filePath = buildPath(params.fileImppath);
 
@@ -1475,8 +1481,8 @@ private void setTargetCPU(ref Param params)
 /**************************************
  * we want to write the mixin expansion file also on error, but there
  * are too many ways to terminate dmd (e.g. fatal() which calls exit(EXIT_FAILURE)),
- * so we cant use scope(exit) ...
- * so we do it with atexit(&flushMixins);
+ * so we can't rely on scope(exit) ... in tryMain() actually being executed
+ * so we add atexit(&flushMixins); for those fatal exits (with the GC still valid)
  */
 extern(C) void flushMixins()
 {
@@ -1488,6 +1494,8 @@ extern(C) void flushMixins()
     OutBuffer* ob = global.params.mixinOut;
     f.setbuffer(cast(void*)ob.data, ob.offset);
     f.write();
+
+    global.params.mixinOut = null;
 }
 
 version (IN_LLVM)
@@ -1929,11 +1937,7 @@ bool parseCommandLine(const ref Strings arguments, const size_t argc, ref Param 
             auto tmp = p + 6 + 1;
             if (!tmp[0])
                 goto Lnoarg;
-            // The following are usedin atexit, so we can't rely on main's argv...
             params.mixinFile = mem.xstrdup(tmp);
-            // ... or the GC's memory being valid.
-            params.mixinOut = cast(OutBuffer*)calloc(1, OutBuffer.sizeof);
-            atexit(&flushMixins);
         }
         else if (arg == "-g") // https://dlang.org/dmd.html#switch-g
             params.symdebug = 1;
--- a/dmd/mars.d
+++ b/dmd/mars.d
@@ -802,6 +802,22 @@ version (IN_LLVM) {} else
     }
 version (IN_LLVM)
 {
+    import core.memory : GC;
+
+    static if (__traits(compiles, GC.stats))
+    {
+        if (global.params.verbose)
+        {
+            static int toMB(ulong size) { return cast(int) (size / 1048576.0 + 0.5); }
+
+            const stats = GC.stats;
+            const used = toMB(stats.usedSize);
+            const free = toMB(stats.freeSize);
+            const total = toMB(stats.usedSize + stats.freeSize);
+            message("GC stats  %dM used, %dM free, %dM total", used, free, total);
+        }
+    }
+
     codegenModules(modules);
 }
 else
--- a/dmd/root/filename.d
+++ b/dmd/root/filename.d
@@ -710,16 +710,16 @@ nothrow:
                     // exists and name is *really* a "child" of path
                     if (exists(cname) && strncmp(cpath, cname, strlen(cpath)) == 0)
                     {
-                        .free(cast(void*)cpath);
+                        mem.xfree(cast(void*)cpath);
                         const(char)* p = mem.xstrdup(cname);
-                        .free(cast(void*)cname);
+                        mem.xfree(cast(void*)cname);
                         return p;
                     }
                 cont:
                     if (cpath)
-                        .free(cast(void*)cpath);
+                        mem.xfree(cast(void*)cpath);
                     if (cname)
-                        .free(cast(void*)cname);
+                        mem.xfree(cast(void*)cname);
                 }
             }
             return null;
--- a/dmd/root/rmem.d
+++ b/dmd/root/rmem.d
@@ -12,236 +12,288 @@
 
 module dmd.root.rmem;
 
+import core.stdc.stdlib;
+import core.stdc.stdio;
 import core.stdc.string;
+
+version = GC;
 
 version (GC)
 {
     import core.memory : GC;
 
-    extern (C++) struct Mem
-    {
-        static char* xstrdup(const(char)* p) nothrow
-        {
-            return p[0 .. strlen(p) + 1].dup.ptr;
-        }
-
-        static void xfree(void* p) nothrow
-        {
-            return GC.free(p);
-        }
-
-        static void* xmalloc(size_t n) nothrow
-        {
-            return GC.malloc(n);
-        }
-
-        static void* xcalloc(size_t size, size_t n) nothrow
-        {
-            return GC.calloc(size * n);
-        }
-
-        static void* xrealloc(void* p, size_t size) nothrow
-        {
-            return GC.realloc(p, size);
-        }
-
-        static void error() nothrow
-        {
-            import core.stdc.stdlib : exit, EXIT_FAILURE;
-            import core.stdc.stdio : printf;
-
-            printf("Error: out of memory\n");
-            exit(EXIT_FAILURE);
-        }
-    }
-
-    extern (C) void* allocmemory(size_t m_size) nothrow
-    {
-        return GC.malloc(m_size);
-    }
-
-    extern (C++) const __gshared Mem mem;
+    enum isGCAvailable = true;
 }
 else
+    enum isGCAvailable = false;
+
+extern (C++) struct Mem
 {
-    import core.stdc.stdlib;
-    import core.stdc.stdio;
-
-    extern (C++) struct Mem
+    static char* xstrdup(const(char)* s) nothrow
     {
-        static char* xstrdup(const(char)* s) nothrow
-        {
-            if (s)
-            {
-                auto p = .strdup(s);
-                if (p)
-                    return p;
-                error();
-            }
+        if (!s)
             return null;
-        }
 
-        static void xfree(void* p) nothrow
+        version (GC)
+            if (isGCEnabled)
+                return s[0 .. strlen(s) + 1].dup.ptr;
+
+        auto p = .strdup(s);
+        if (!p)
+            error();
+        return p;
+    }
+
+    static void xfree(void* p) nothrow
+    {
+        if (!p)
+            return;
+
+        version (GC)
+            if (isGCEnabled)
+                return GC.free(p);
+
+        .free(p);
+    }
+
+    static void* xmalloc(size_t size) nothrow
+    {
+        if (!size)
+            return null;
+
+        version (GC)
+            if (isGCEnabled)
+                return GC.malloc(size);
+
+        auto p = .malloc(size);
+        if (!p)
+            error();
+        return p;
+    }
+
+    static void* xcalloc(size_t size, size_t n) nothrow
+    {
+        const totalSize = size * n;
+        if (!totalSize)
+            return null;
+
+        version (GC)
+            if (isGCEnabled)
+                return GC.calloc(totalSize);
+
+        auto p = .calloc(size, n);
+        if (!p)
+            error();
+        return p;
+    }
+
+    static void* xrealloc(void* p, size_t size) nothrow
+    {
+        version (GC)
+            if (isGCEnabled)
+                return GC.realloc(p, size);
+
+        if (!size)
         {
             if (p)
                 .free(p);
+            return null;
         }
 
-        static void* xmalloc(size_t size) nothrow
+        if (!p)
         {
-            if (!size)
-                return null;
-
-            auto p = .malloc(size);
+            p = .malloc(size);
             if (!p)
                 error();
             return p;
         }
 
-        static void* xcalloc(size_t size, size_t n) nothrow
-        {
-            if (!size || !n)
-                return null;
-
-            auto p = .calloc(size, n);
-            if (!p)
-                error();
-            return p;
-        }
-
-        static void* xrealloc(void* p, size_t size) nothrow
-        {
-            if (!size)
-            {
-                if (p)
-                    .free(p);
-                return null;
-            }
-
-            if (!p)
-            {
-                p = .malloc(size);
-                if (!p)
-                    error();
-                return p;
-            }
-
-            p = .realloc(p, size);
-            if (!p)
-                error();
-            return p;
-        }
-
-        static void error() nothrow
-        {
-            printf("Error: out of memory\n");
-            exit(EXIT_FAILURE);
-        }
+        p = .realloc(p, size);
+        if (!p)
+            error();
+        return p;
     }
 
-    extern (C++) const __gshared Mem mem;
-
-    enum CHUNK_SIZE = (256 * 4096 - 64);
-
-    __gshared size_t heapleft = 0;
-    __gshared void* heapp;
-
-    extern (C) void* allocmemory(size_t m_size) nothrow
+    static void error() nothrow
     {
-        // 16 byte alignment is better (and sometimes needed) for doubles
-        m_size = (m_size + 15) & ~15;
-
-        // The layout of the code is selected so the most common case is straight through
-        if (m_size <= heapleft)
-        {
-        L1:
-            heapleft -= m_size;
-            auto p = heapp;
-            heapp = cast(void*)(cast(char*)heapp + m_size);
-            return p;
-        }
-
-        if (m_size > CHUNK_SIZE)
-        {
-            auto p = malloc(m_size);
-            if (p)
-            {
-                return p;
-            }
-            printf("Error: out of memory\n");
-            exit(EXIT_FAILURE);
-        }
-
-        heapleft = CHUNK_SIZE;
-        heapp = malloc(CHUNK_SIZE);
-        if (!heapp)
-        {
-            printf("Error: out of memory\n");
-            exit(EXIT_FAILURE);
-        }
-        goto L1;
+        printf("Error: out of memory\n");
+        exit(EXIT_FAILURE);
     }
 
-    version (DigitalMars)
+    version (GC)
     {
-        enum OVERRIDE_MEMALLOC = true;
-    }
-    else version (LDC)
-    {
-        // Memory allocation functions gained weak linkage when the @weak attribute was introduced.
-        import ldc.attributes;
-        enum OVERRIDE_MEMALLOC = is(typeof(ldc.attributes.weak));
-    }
-    else
-    {
-        enum OVERRIDE_MEMALLOC = false;
-    }
+        __gshared bool isGCEnabled = true;
 
-    static if (OVERRIDE_MEMALLOC)
-    {
-        extern (C) void* _d_allocmemory(size_t m_size) nothrow
+        static void disableGC()
         {
-            return allocmemory(m_size);
+            isGCEnabled = false;
         }
 
-        extern (C) Object _d_newclass(const ClassInfo ci) nothrow
+        static void addRange(const(void)* p, size_t size) nothrow
         {
-            auto p = allocmemory(ci.initializer.length);
-            p[0 .. ci.initializer.length] = cast(void[])ci.initializer[];
-            return cast(Object)p;
+            if (isGCEnabled)
+                GC.addRange(p, size);
         }
 
-        version (LDC)
+        static void removeRange(const(void)* p) nothrow
         {
-            extern (C) Object _d_allocclass(const ClassInfo ci) nothrow
-            {
-                return cast(Object)allocmemory(ci.initializer.length);
-            }
-        }
-
-        extern (C) void* _d_newitemT(TypeInfo ti) nothrow
-        {
-            auto p = allocmemory(ti.tsize);
-            (cast(ubyte*)p)[0 .. ti.initializer.length] = 0;
-            return p;
-        }
-
-        extern (C) void* _d_newitemiT(TypeInfo ti) nothrow
-        {
-            auto p = allocmemory(ti.tsize);
-            p[0 .. ti.initializer.length] = ti.initializer[];
-            return p;
-        }
-
-        // TypeInfo.initializer for compilers older than 2.070
-        static if(!__traits(hasMember, TypeInfo, "initializer"))
-        private const(void[]) initializer(T : TypeInfo)(const T t)
-        nothrow pure @safe @nogc
-        {
-            return t.init;
+            if (isGCEnabled)
+                GC.removeRange(p);
         }
     }
 }
+
+extern (C++) const __gshared Mem mem;
+
+enum CHUNK_SIZE = (256 * 4096 - 64);
+
+__gshared size_t heapleft = 0;
+__gshared void* heapp;
+
+extern (C) void* allocmemory(size_t m_size) nothrow
+{
+    // 16 byte alignment is better (and sometimes needed) for doubles
+    m_size = (m_size + 15) & ~15;
+
+    // The layout of the code is selected so the most common case is straight through
+    if (m_size <= heapleft)
+    {
+    L1:
+        heapleft -= m_size;
+        auto p = heapp;
+        heapp = cast(void*)(cast(char*)heapp + m_size);
+        return p;
+    }
+
+    if (m_size > CHUNK_SIZE)
+    {
+        auto p = malloc(m_size);
+        if (p)
+        {
+            return p;
+        }
+        printf("Error: out of memory\n");
+        exit(EXIT_FAILURE);
+    }
+
+    heapleft = CHUNK_SIZE;
+    heapp = malloc(CHUNK_SIZE);
+    if (!heapp)
+    {
+        printf("Error: out of memory\n");
+        exit(EXIT_FAILURE);
+    }
+    goto L1;
+}
+
+version (DigitalMars)
+{
+    enum OVERRIDE_MEMALLOC = true;
+}
+else version (LDC)
+{
+    // Memory allocation functions gained weak linkage when the @weak attribute was introduced.
+    import ldc.attributes;
+    enum OVERRIDE_MEMALLOC = is(typeof(ldc.attributes.weak));
+}
+else
+{
+    enum OVERRIDE_MEMALLOC = false;
+}
+
+static if (OVERRIDE_MEMALLOC)
+{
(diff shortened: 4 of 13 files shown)


kinke:
On Win64, I can now compile all default libs (incl. unittests) without crashing. I don't see the memory drop though in the VS tools; adding an explicit `GC.minimize()` makes no difference...

kinke:
I've also tried NOT disabling the GC at startup; that seems to work too, at least for 2 Phobos module unittests, and does reduce the memory footprint (`std.string`: from ~1.9 GB to ~1.1).

kinke:
`GC.minimize()` is supposed to do exactly that (according to the docs). Maybe the fragmentation is so high, in combination with the page size, so that there isn't a relevant number of unused pages to be freed.

kinke:
Yeah I think it's the fragmentation, as in a few cases, `GC.minimize()` seems to have a (pretty small) effect:
```
> bin\ldc2 -c -unittest ..\ldc\runtime\phobos\std\regex\internal\tests.d
GC: 1983M used, 76M free, 2059M total
GC: 280M used, 1331M free, 1611M total
```

rainers:
GC.minimize only returns memory to the OS if a complete pool (usually 4MB or larger IIRC) is unused, so very unlikely for "small" object pools. Might happen for "large" allocations (>2k) eventually.

rainers:
Did you also measure impact on compile times? Last time I tried that it was some 10-20% IIRC.

My actual target was to compile phobos unittests all at once, but that still used more memory than my system was able to provide (16GB) and crashed at some point.

kinke:
That's what I thought. Is the pool/page size configurable via cmdline switch?

I haven't checked the compile times yet; the single collect + minimize appears pretty cheap though, that's why I wanted to focus on that. Enabling the GC from the start is surely a different matter, but probably not that costly (relatively) with `-O`.

I also tried compiling the Phobos unittests all at once a while back, and IIRC, it was something like 16 GB (my system has 32), and was also very slow compared to parallel separate compilation with my 4 cores.

rainers:
> That's what I thought. Is the pool/page size configurable via cmdline switch?

You can tweak it with `--DRT-gcopt`, but not below 1MB:
```
>dmd --DRT-gcopt=help
GC options are specified as whitespace separated assignments:
    disable:0|1    - start disabled (0)
    profile:0|1|2  - enable profiling with summary when terminating program (0)
    gc:conservative|manual - select gc implementation (default = conservative)

    initReserve:N  - initial memory to reserve in MB (0)
    minPoolSize:N  - initial and minimum pool size in MB (1)
    maxPoolSize:N  - maximum pool size in MB (64)
    incPoolSize:N  - pool size increment MB (3)
    heapSizeFactor:N - targeted heap size to used memory ratio (2)
```

kinke:
Ah yeah nice. LDC doesn't use the D main args, but the druntime args, i.e., these DRT options aren't filtered out and so unusable with LDC yet.

Still, I tried embedding the option as global:
```
extern(C) __gshared string[] rt_options = [ "gcopt=maxPoolSize:1"];
```

That has quite an effect for the regex unittests, reducing the total size from 1.6 GB with default settings to 0.8:
```
GC: 1984M used, 1M free, 1985M total
GC: 279M used, 534M free, 813M total
```

kinke:
Oh wow. I've just tried additionally enabling the GC selectively during `ctfeInterpret()`:
```diff
--- a/dmd/dinterpret.d
+++ b/dmd/dinterpret.d
@@ -54,6 +54,8 @@ import dmd.visitor;
  */
 public Expression ctfeInterpret(Expression e)
 {
+    import core.memory;
+
     if (e.op == TOK.error)
         return e;
     assert(e.type); // https://issues.dlang.org/show_bug.cgi?id=14642
@@ -61,6 +63,8 @@ public Expression ctfeInterpret(Expression e)
     if (e.type.ty == Terror)
         return new ErrorExp();

+    GC.enable();
+
     // This code is outside a function, but still needs to be compiled
     // (there are compiler-generated temporary variables such as __dollar).
     // However, this will only be run once and can then be discarded.
@@ -75,6 +79,8 @@ public Expression ctfeInterpret(Expression e)
     if (CTFEExp.isCantExp(result))
         result = new ErrorExp();

+    GC.disable();
+
     return result;
 }
```

Results, without `gcopt=maxPoolSize:1`:
```
> bin\ldc2 -c -unittest ..\ldc\runtime\phobos\std\regex\internal\tests.d
GC: 491M used, 41M free, 532M total
GC: 281M used, 251M free, 532M total

>bin\ldc2 -c -unittest ..\ldc\runtime\phobos\std\string.d
GC: 995M used, 40M free, 1035M total
GC: 345M used, 690M free, 1035M total
```

Additionally with `gcopt=maxPoolSize:1` (seems rather costly in comparison):
```
bin\ldc2 -c -unittest ..\ldc\runtime\phobos\std\regex\internal\tests.d
GC: 491M used, 1M free, 492M total
GC: 281M used, 210M free, 491M total

C:\LDC\ninja-ldc>bin\ldc2 -c -unittest ..\ldc\runtime\phobos\std\string.d
GC: 631M used, 133M free, 764M total
GC: 346M used, 400M free, 746M total
```

This is a cut-down of the regex unittests front-end memory requirements by 75%.

kinke:
Runtime-wise, selectively enabling the GC makes almost no difference compared to having it enabled all the time, while enabling it from the start yields a 10% lower total size for the `std.string` unittests.
After some rudimentary checks, enabling it from the start adds very roughly 10-25% to overall runtime without `-O`.
`gcopt=maxPoolSize:4` makes the regex tests compilation twice as long (with enabled GC), compared to the default setting (!).

rainers:
> This is a cut-down of the regex unittests front-end (OS) memory requirements by 75% (2 GB -> 0.5 GB).

Sounds promising.

> Runtime-wise, selectively enabling the GC makes almost no difference compared to having it enabled all the time, while enabling it from the start yields a 10% lower total size for the `std.string` unittests.

I guess a collection that has been avoided while the GC is disabled will always be done with the next allocation during CTFE.

> After some rudimentary checks, enabling it from the start adds very roughly 10-25% to overall runtime without `-O`.

That's consistent with what I remember.

> `gcopt=maxPoolSize:4` makes the regex tests compilation run twice as long (with enabled GC), compared to the default setting (!).

That's probably because it will cause more collections whenever the next pool is exhausted. At that size each collection will take considerable time. You can show some stats with gcopt `profile:1`.

kinke:
Damn - just enabling `version = GC` in `dmd.root.rmem` (while keeping the GC disabled, ) doesn't just cost runtime performance, but also a non-negligible memory overhead. I've compared the runtimes and memory requirements for `ninja -j1 phobos2-ldc-debug` (using `/usr/bin/time -v`):

`LDC_WITH_GC=OFF`:
```
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:36.54
	Maximum resident set size (kbytes): 1407776
```

`LDC_WITH_GC=ON`:
```
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:40.35
	Maximum resident set size (kbytes): 1890952
```

`LDC_WITH_GC=ON -lowmem`:
```
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:44.84
	Maximum resident set size (kbytes): 1284764
```

So while the savings with `-lowmem` look quite nice compared to not using the new switch, it hardly pays off in this case compared to current master (23% runtime overhead for -9% memory requirements).

rainers:
I guess a command line switch would have to select the allocation method GC/bump-pointer at runtime in rmem.d. Memory overhead is expected to be about 25% for the GC as it allocates in blocks that are a power of 2 (up to 8k).

Building phobos without unittests is probably not a good candidate for freeing memory as most data is still referenced in the AST. There have to be more intermediate results caused by CTFE, for example.

kinke:
> Building phobos without unittests is probably not a good candidate for freeing memory as most data is still referenced in the AST.

Well, for Win64 the actually referenced data after a final collection was 544 MB, see the first post.

kinke:
> I guess a command line switch would have to select the allocation method GC/bump-pointer at runtime in rmem.d.

The difficulty is that the non-GC version also overrides the allocation functions (`_d_allocmemory` etc.) at link-time. No idea how to optionally still call the original druntime functions.

> Memory overhead is expected to be about 25% for the GC as it allocates in blocks that are a power of 2 (up to 8k).

In the Phobos all-at-once case it's 34%.

kinke:
What's apparently eating lots of memory is CTFE and templates. IIRC, building the `ldc-build-runtime` tool (300 lines, but importing Phobos templates) requires about twice as much memory as the full DMD frontend all-at-once.

kinke:
> as it allocates in blocks that are a power of 2 (up to 8k).

You mean for each instance? So if we have `extern(C++) class C { size_t[4] blub; }`, the size is padded from 8 (vptr) + 4x8 = 40 bytes to 64? And so from 72 bytes to 128, 136 to 256 etc.?

rainers:
> Well, for Win64 the actually referenced data after a final collection was 544 MB (one third of the allocated GC memory), see the first post.

That's probbly due to the `heapSizeFactor` with a default of 2.

> What's apparently eating lots of memory is CTFE and templates.

I agree for CTFE, but are there a lot of temporary allocations for templates? I suspect all instances end up attached to some module to be sent to code generation.

> You mean for each instance? So if we have `extern(C++) class C { size_t[4] blub; }`, the size is padded from 8 (vptr) + 4x8 = 40 bytes to 64? And so from 72 bytes to 128, 136 to 256 etc.?

Yes.

kinke:
> > Well, for Win64 the actually referenced data after a final collection was 544 MB (one third of the allocated GC memory), see the first post.
>
> That's probbly due to the heapSizeFactor with a default of 2.

May well be. A bloat factor of 2 seems to be a very rough average for a normal compilation run, with some extremes like `std.regex.internal.tests` (bloat factor > 7), e.g., see https://ci.appveyor.com/project/kinke/ldc/builds/20376954/job/nh7cfphk3pfm5w69?fullLog=true. That's why I tested with Phobos all-at-once as more realistic test, which should have shown some improvements with its factor of 3.

```
[124/1217] Generating bin/ldc2.exe.obj (=> all D parts all-at-once)
GC: used bytes = 586820864
GC: used bytes = 290709920
...
[7/1217] Generating bin/ldc-build-runtime.exe.obj
GC: used bytes = 901887232
GC: used bytes = 274935664
```

> > You mean for each instance?
>
> Yes.

Oh wow, that's *significant* waste. Would it be hard to get rid of this?

kinke:
> Memory overhead is expected to be about 25% for the GC

I got interested. With
```diff
diff --git a/dmd/mars.d b/dmd/mars.d
index e1ebcb367..fd9a455e6 100644
--- a/dmd/mars.d
+++ b/dmd/mars.d
@@ -73,6 +73,9 @@ version (IN_LLVM)
     int linkObjToBinary();
     void deleteExeFile();
     int runProgram();
+
+    extern __gshared ulong[32] wastages;
+    extern __gshared ulong totalAlloc;
 }
 else
 {
@@ -860,6 +863,19 @@ version (IN_LLVM)

     printGCStats();

+    static int toMB(ulong size) { return cast(int) (size / 1048576.0 + 0.5); }
+
+    printf(".: Wastages:\n");
+    ulong totalWaste = 0;
+    foreach (i, e; wastages)
+    {
+        if (e == 0)
+            continue;
+        totalWaste += e;
+        printf("  bin size %d: %dM\n", cast(int) (1 << i), toMB(e));
+    }
+    printf("  total: %dM (%d%% of %dM)\n", toMB(totalWaste), cast(int) (100 * (totalWaste / cast(double) totalAlloc) + 0.5), toMB(totalAlloc));
+
     codegenModules(modules);
 }
 else
diff --git a/dmd/root/rmem.d b/dmd/root/rmem.d
index ebd5ac651..ab9f957b2 100644
--- a/dmd/root/rmem.d
+++ b/dmd/root/rmem.d
@@ -172,8 +172,27 @@ else
     __gshared size_t heapleft = 0;
     __gshared void* heapp;

+    extern(C++) __gshared ulong totalAlloc;
+    extern(C++) __gshared ulong[32] wastages;
+
     extern (C) void* allocmemory(size_t m_size) nothrow
     {
+        totalAlloc += m_size;
+        import std.math : nextPow2;
+        const binSize = nextPow2(m_size - 1);
+        const wastedSize = binSize - m_size;
+        if (wastedSize != 0)
+        {
+            foreach (i, ref e; wastages)
+            {
+                if (binSize >> i == 1)
+                {
+                    e += wastedSize;
+                    break;
+                }
+            }
+        }
+
```

druntime all-at-once:
```
.: Wastages:
  bin size 64: 8M
  bin size 128: 11M
  bin size 256: 18M
  bin size 512: 21M
  bin size 1024: 8M
  total: 66M (47% of 140M)
```

Phobos all-at-once:
```
.: Wastages:
  bin size 32: 0M
  bin size 64: 35M
  bin size 128: 39M
  bin size 256: 101M
  bin size 512: 103M
  bin size 1024: 28M
  total: 306M (42% of 727M)
```

regex unittests:
```
.: Wastages:
  bin size 32: 0M
  bin size 64: 39M
  bin size 128: 23M
  bin size 256: 49M
  bin size 512: 51M
  bin size 1024: 16M
  total: 178M (39% of 453M)
```

`std.string` unittests:
```
.: Wastages:
  bin size 32: 0M
  bin size 64: 34M
  bin size 128: 33M
  bin size 256: 85M
  bin size 512: 92M
  bin size 1024: 27M
  total: 271M (41% of 665M)
```

kinke:
```diff
diff --git a/dmd/mars.d b/dmd/mars.d
index e1ebcb367..2a4c900d3 100644
--- a/dmd/mars.d
+++ b/dmd/mars.d
@@ -73,6 +73,9 @@ version (IN_LLVM)
     int linkObjToBinary();
     void deleteExeFile();
     int runProgram();
+
+    extern __gshared ulong[32] totalAlloc;
+    extern __gshared ulong[32] wastages;
 }
 else
 {
@@ -860,6 +863,22 @@ version (IN_LLVM)

     printGCStats();

+    static int toMB(ulong size) { return cast(int) (size / 1048576.0 + 0.5); }
+
+    printf(".: Wastages:\n");
+    ulong totalSize = 0;
+    ulong totalWaste = 0;
+    foreach (i, e; wastages)
+    {
+        const totalBinAlloc = totalAlloc[i];
+        if (e == 0 && totalBinAlloc == 0)
+            continue;
+        totalSize += totalBinAlloc;
+        totalWaste += e;
+        printf("  bin size %d: %dM (%d%% of %dM)\n", cast(int) (1 << i), toMB(e), cast(int) (100 * (e / cast(double) totalBinAlloc) + 0.5), toMB(totalBinAlloc));
+    }
+    printf("  total: %dM (%d%% of %dM)\n", toMB(totalWaste), cast(int) (100 * (totalWaste / cast(double) totalSize) + 0.5), toMB(totalSize));
+
     codegenModules(modules);
 }
 else
diff --git a/dmd/root/rmem.d b/dmd/root/rmem.d
index ebd5ac651..eeb77b0e3 100644
--- a/dmd/root/rmem.d
+++ b/dmd/root/rmem.d
@@ -172,8 +172,24 @@ else
     __gshared size_t heapleft = 0;
     __gshared void* heapp;

+    extern(C++) __gshared ulong[32] totalAlloc;
+    extern(C++) __gshared ulong[32] wastages;
+
     extern (C) void* allocmemory(size_t m_size) nothrow
     {
+        import std.math : nextPow2;
+        const binSize = nextPow2(m_size - 1);
+        const wastedSize = binSize - m_size;
+        foreach (i, ref e; wastages)
+        {
+            if (binSize >> i == 1)
+            {
+                totalAlloc[i] += m_size;
+                e += wastedSize;
+                break;
+            }
+        }
+
         // 16 byte alignment is better (and sometimes needed) for doubles
         m_size = (m_size + 15) & ~15;

```

druntime all-at-once:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 1M)
  bin size 32: 0M (0% of 10M)
  bin size 64: 8M (39% of 19M)
  bin size 128: 11M (79% of 14M)
  bin size 256: 18M (37% of 49M)
  bin size 512: 21M (63% of 33M)
  bin size 1024: 8M (59% of 14M)
  total: 66M (47% of 140M)
```

Phobos all-at-once:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 4M)
  bin size 32: 0M (0% of 44M)
  bin size 64: 35M (35% of 101M)
  bin size 128: 39M (71% of 55M)
  bin size 256: 101M (35% of 293M)
  bin size 512: 103M (57% of 179M)
  bin size 1024: 28M (56% of 50M)
  total: 306M (42% of 727M)
```

regex unittests:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 2M)
  bin size 32: 0M (0% of 36M)
  bin size 64: 39M (32% of 122M)
  bin size 128: 23M (66% of 34M)
  bin size 256: 49M (35% of 139M)
  bin size 512: 51M (57% of 90M)
  bin size 1024: 16M (56% of 29M)
  total: 178M (39% of 453M)
```

`std.string` unittests:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 4M)
  bin size 32: 0M (0% of 40M)
  bin size 64: 34M (33% of 105M)
  bin size 128: 33M (65% of 50M)
  bin size 256: 85M (34% of 247M)
  bin size 512: 92M (54% of 171M)
  bin size 1024: 27M (56% of 48M)
  total: 271M (41% of 665M)
```

rainers:
>  bin size 128: 48M (70% of 70M)

These look suspicious ;-)

kinke:
> bin size 128: 48M (70% of 70M)

Why suspicious? Sum of actual sizes for that bin = (rounded) 70M, (rounded) 48M would be padding waste, and that's (rounded, but based on actual bytes, not Megabytes) 70% of the sum of actual sizes.

rainers:
Padding cannot be larger than 50%, the next smaller bin would be used instead.

kinke:
The padding can almost be 100%, relative to the *actual* size. 72 bytes `m_size` => bin size 128 bytes => (128 - 72) / 72 = 78%.

rainers:
Ah, ok, that's relative to the requested size, not the allocated size.

kinke:
Yep, these percentages represent the increase of the actually requested size due to padding.

rainers:
By that measure, the expected ratio for random size distribution would be 33%. With smaller size a bit more probable, your measured 40% seem consistent.

kinke:
As someone not familiar with GC designs, 40% seems pretty high to me. So using the GC (over a bump-ptr scheme) is only worth it if more than ~30% (size-wise, 70% * 1.4 padding increase = 98%) of all memory allocations are temporary, otherwise it increases the memory footprint besides costing runtime performance.

kinke:
The latest experiment wrt. bypassing the GC for the x{malloc,...} functions isn't the best of both worlds, but the worst of both. ;) For the `std.string` unittests, `815M used, 28M free, 843M total` (815 MB referenced data vs. 345 MB before, while less data is allocated via the GC); `Maximum resident set size` for Phobos all-at-once is minimally less than the earlier one with `LDC_WITH_GC=ON` *without* `-lowmem` (i.e., no collections at all).
Are the finalizers not called at all during a collection, or can circular references be the cause for this?

kinke:
> Comparison with bump-ptr is a bit unfair, though. C's malloc also has it overhead. I just tried it on Win32/Win64: both seem to add 16 bytes per alloc.

Yeah but I'm not talking about a trivial malloc() for every single alloc, but a minimally more advanced version like the non-version-(GC) one, allocating blocks of at least ~1 MB. 2000 (for 2 GB) x 24 bytes is just 48 KB.

kinke:
> That's still bump-pointer-allocation

Yes of course, but I meant the paging, which gets rid of the 16 bytes overhead per allocation.

> These 10% are almost already wasted by the 16 byte alignment.

I meant 10% over non-`version (GC)` as baseline, which already includes this alignment padding.
I'm not sure if it would pay off to try to shrink the alignment requirements for appropriate types; I wanted to investigate the GC memory alignments at some point, e.g., if the memory is guaranteed to be aligned to 64 bytes if the class contains a struct with 64-bytes alignment, and optionally pass the required alignment as additional param to `_d_allocmemory`. As long as the size isn't smaller than the alignment (unlikely), this should be almost fine with the current GC, as the bin sizes are powers of 2, and so the offset of each bin relative to the page start should be fine; the remaining problem I assume is that the page alignment itself isn't guaranteed to match. - But that's an entirely different topic.

> non-power-of-2 bins also waste memory due to page alignment, though

Assuming the page size is sufficiently high (page = pool of min size 1 MB, or is a pool a set of pages?), that should hardly matter for the vast majority of small allocations (max bin size above was 1 KB). Bin sizes of [1, 2, 4,] 8, 12, 16, 24, 32, 48, 64, 96, 128, 192, 256, i.e., adding a further bin size inbetween each power of 2, might be worth a try. I may get round to simulating that with the diff above to check the waste with such a scheme.

kinke:
Yep, that would already improve the situation considerably, reducing the overhead from ~40% to ~15%:

```diff
diff --git a/dmd/mars.d b/dmd/mars.d
index a9f988513..d39aee630 100644
--- a/dmd/mars.d
+++ b/dmd/mars.d
@@ -841,6 +841,21 @@ version (IN_LLVM) {} else
     }
 version (IN_LLVM)
 {
+    static int toMB(ulong size) { return cast(int) (size / 1048576.0 + 0.5); }
+
+    printf(".: Wastages:\n");
+    ulong totalActualSize = 0;
+    ulong totalPadding = 0;
+    foreach (bin; bins)
+    {
+        if (bin.sumOfActualSizes == 0)
+            continue;
+        totalActualSize += bin.sumOfActualSizes;
+        totalPadding += bin.sumOfPadding;
+        printf("  bin size %d: %dM (%d%% of %dM)\n", cast(int) bin.binSize, toMB(bin.sumOfPadding), cast(int) (100 * (bin.sumOfPadding / cast(double) bin.sumOfActualSizes) + 0.5), toMB(bin.sumOfActualSizes));
+    }
+    printf("  total: %dM (%d%% of %dM)\n", toMB(totalPadding), cast(int) (100 * (totalPadding / cast(double) totalActualSize) + 0.5), toMB(totalActualSize));
+
     version (LDC_WITH_GC)
     {
         import core.memory : GC;
@@ -849,8 +864,6 @@ version (IN_LLVM)
         {
             if (global.params.verbose)
             {
-                static int toMB(ulong size) { return cast(int) (size / 1048576.0 + 0.5); }
-
                 const stats = GC.stats;
                 const used = toMB(stats.usedSize);
                 const free = toMB(stats.freeSize);
diff --git a/dmd/root/rmem.d b/dmd/root/rmem.d
index ebd5ac651..c90e04771 100644
--- a/dmd/root/rmem.d
+++ b/dmd/root/rmem.d
@@ -172,8 +172,35 @@ else
     __gshared size_t heapleft = 0;
     __gshared void* heapp;

+    struct Bin { ulong binSize; ulong sumOfActualSizes; ulong sumOfPadding; }
+    __gshared Bin[32] bins;
+    shared static this()
+    {
+        foreach (i, ref bin; bins)
+        {
+            const exponent = i / 2;
+            bin.binSize = (1L << exponent);
+            if ((i & 1) == 1 && bin.binSize >= 32)
+                bin.binSize = bin.binSize * 3 / 2;
+            // else: just a duplicate of the previous bin, won't be used
+        }
+    }
+
     extern (C) void* allocmemory(size_t m_size) nothrow
     {
+        import std.math : nextPow2;
+        const binSize = nextPow2(m_size - 1);
+        const wastedSize = binSize - m_size;
+        foreach (ref bin; bins)
+        {
+            if (m_size <= bin.binSize)
+            {
+                bin.sumOfActualSizes += m_size;
+                bin.sumOfPadding += (bin.binSize - m_size);
+                break;
+            }
+        }
+
         // 16 byte alignment is better (and sometimes needed) for doubles
         m_size = (m_size + 15) & ~15;

```

druntime all-at-once:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 1M)
  bin size 32: 0M (0% of 10M)
  bin size 48: 1M (9% of 16M)
  bin size 64: 0M (12% of 3M)
  bin size 96: 5M (38% of 13M)
  bin size 128: 0M (12% of 1M)
  bin size 192: 2M (6% of 37M)
  bin size 256: 3M (23% of 12M)
  bin size 384: 7M (22% of 33M)
  bin size 768: 3M (19% of 14M)
  total: 22M (16% of 140M)
```

Phobos all-at-once:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 4M)
  bin size 32: 0M (0% of 44M)
  bin size 48: 5M (6% of 76M)
  bin size 64: 4M (14% of 25M)
  bin size 96: 16M (32% of 51M)
  bin size 128: 0M (9% of 4M)
  bin size 192: 13M (8% of 170M)
  bin size 256: 27M (22% of 123M)
  bin size 384: 32M (18% of 179M)
  bin size 768: 9M (17% of 50M)
  total: 106M (15% of 727M)
```

regex unittests:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 2M)
  bin size 32: 0M (0% of 36M)
  bin size 48: 2M (2% of 99M)
  bin size 64: 3M (13% of 23M)
  bin size 96: 9M (31% of 30M)
  bin size 128: 0M (6% of 4M)
  bin size 192: 6M (8% of 79M)
  bin size 256: 14M (23% of 60M)
  bin size 384: 16M (18% of 90M)
  bin size 768: 5M (17% of 29M)
  total: 56M (12% of 453M)
```

`std.string` unittests:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 4M)
  bin size 32: 0M (0% of 40M)
  bin size 48: 2M (3% of 86M)
  bin size 64: 2M (12% of 19M)
  bin size 96: 13M (26% of 48M)
  bin size 128: 0M (9% of 2M)
  bin size 192: 11M (9% of 128M)
  bin size 256: 27M (23% of 118M)
  bin size 384: 26M (15% of 171M)
  bin size 768: 8M (17% of 48M)
  total: 90M (14% of 665M)
```

`std.typecons` unittests:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 6M)
  bin size 32: 0M (0% of 66M)
  bin size 48: 6M (8% of 80M)
  bin size 64: 4M (12% of 29M)
  bin size 96: 19M (29% of 67M)
  bin size 128: 0M (13% of 3M)
  bin size 192: 19M (7% of 275M)
  bin size 256: 41M (22% of 185M)
  bin size 384: 48M (17% of 279M)
  bin size 768: 11M (17% of 67M)
  total: 149M (14% of 1057M)
```

`std.range.package` unittests:
```
.: Wastages:
  bin size 8: 0M (0% of 0M)
  bin size 16: 0M (0% of 7M)
  bin size 32: 0M (0% of 70M)
  bin size 48: 6M (7% of 88M)
  bin size 64: 4M (13% of 34M)
  bin size 96: 23M (29% of 79M)
  bin size 128: 0M (11% of 4M)
  bin size 192: 22M (8% of 276M)
  bin size 256: 40M (22% of 184M)
  bin size 384: 49M (18% of 271M)
  bin size 768: 13M (17% of 75M)
  total: 159M (15% of 1087M)
```

joakim-noah:
>And If I had a Ryzen with 16 logical cores, I'd prefer not having to have something like 32 GB of memory to make use of all of them when compiling the unittests. I guess it's similar for Joakim's phone with 8 cores.

Yep, I cannot use all 8 cores when building Phobos tests, been using `ninja -j5` but even that craps out with the last couple LDC releases. `std.string` and the other modules you build above eat up all available memory on my 6GB RAM smartphone and Android kills the Termux app once the GUI framerate starts hiccuping.

Would love for the GC to forestall this, [as I noted on the forum](https://forum.dlang.org/thread/acxedxzzesxkyomrsmjo@forum.dlang.org).

kinke:
I got rid of the CMake option and now switch at runtime; it's now either running with an enabled GC (`-lowmem`) or keeping on running with a disabled GC + bump-pointer allocation scheme (faster, less memory overhead). The additional runtime overhead for the 2 modes isn't really measurable.

rainers:
Looks good. It seems the default GC options strike the best balance between speed and memory size. I wonder how much the enforced collection before termination costs (https://github.com/dlang/druntime/blob/master/src/gc/proxy.d#L94). As it is unnecessary for 99% of all programs I think it should be hidden behind a GC option.

kinke:
Yeah the defaults seem reasonable; playing with a slightly lower `heapSizeFactor` (something like 1.67 / 1.75) might pay off in some cases.

I guess the one more collection at shutdown probably doesn't really matter here; the finalization cost is probably very low for the front-end (`Array` dtors, nothing else coming to mind immediately). Hiding it may be a good idea; I'm wondering what that 'popular demand' for reenabling it was. ;)

kinke:
@rainers: Do you think it'd be worth it trying to upstream the 2 modes at runtime (& `-lowmem` switch or whatever), getting rid of compile-time `version (GC)`?

rainers:
> Do you think it'd be worth it trying to upstream the 2 modes at runtime (& `-lowmem` switch or whatever), getting rid of compile-time `version (GC)`?

I'd approve, but it might get some opposition from nano-cycle-counting people (that seem fine with calling the virtual function ti.initializer multiple times). Maybe it can still be versioned, but integrated with the other code, i.e. `version(GC) if(!isGCDisabled) {}`. Will it still work with GC and OVERRIDE_MEMALLOC=0?

2 Observations:
- xstrdup now crashes for s==null
- `!isGCDisabled` is a double negative used a lot, I'd prefer `isGCEnabled`

kinke:
> but it might get some opposition from nano-cycle-counting people

That's what I was afraid of ;) - but yeah, keeping `version (GC)` with different semantics shouldn't make much of a difference.

> Will it still work with GC and OVERRIDE_MEMALLOC=0?

I think it should; `OVERRIDE_MEMALLOC` should only matter for a disabled GC, decreasing runtime and memory requirements.

> xstrdup now crashes for s==null

Yep; I took the `version (GC)` code as baseline, which doesn't check for null, but only later figured that that code isn't used at the moment. I'll change it to the safe variant, although apparently not really required.

> !isGCDisabled is a double negative used a lot, I'd prefer isGCEnabled

Makes sense.

kinke:
Rebased (2.085 druntime features the GC improvements wrt. less memory overhead) and resynced with the upstream PR, incl. `-lowmem` tests for all targets and `--DRT-*` CLI support.

kinke:
@rainers: Wow, I'm now getting much better timings for `hsf=1.5 mps=1`. Timings on Win64 with this PR:

- `ldc2 -unittest -c ..\ldc\runtime\phobos\std\regex\internal\tests.d`: 8.99 secs
- `ldc2 -unittest -c ..\ldc\runtime\phobos\std\regex\internal\tests.d -lowmem "--DRT-gcopt=heapSizeFactor:1.5 maxPoolSize:1"`: 17.44 secs (+94%)

My previous test (see above) was on Linux (and using an `rt_options` global), where it ran 7x longer than with bump-ptr (!). The memory savings are unchanged (~1,700 MB -> 560 MB).

Can this huge speed-up (3.5x) be attributed to your 2.085 GC improvements?

kinke:
For those regex tests, I also see a consistent ~3% runtime improvement when switching from bump-ptr to a simple disabled GC without cleanup (`-lowmem "--DRT-gcopt=disable:1 cleanup:none"`), at the cost of ~6% higher memory requirements.

rainers:
The only explicit performance improvements was for sweeping large allocations (with maybe a slight effect on allocation, too), but I doubt there will be many large allocations (>2kB) by dmd.

> For those regex tests, I also see a consistent ~3% runtime improvement when switching from bump-ptr to a simple disabled GC without cleanup (`-lowmem "--DRT-gcopt=disable:1 cleanup:none"`)

Not sure why it can be faster, but it's similar to my observation with compiling phobos not showing any difference within the accuracy of measurements. Maybe it's because the GC malloc/free functions are faster than C malloc/free for Mem.xmalloc/xfree.

Issue #1692, opened 2016-08-15
Revise class alignments


kinke:
We probably don't correctly support them at the moment. This goes from making sure the GC memory is appropriately aligned, using the correct alignment when allocating a class on the stack (in `DtoNewClass()` and `GarbageCollect2Stack` etc.), verifying `DtoAlignment()` and in particular, inserting appropriate padding inbetween a class header (vtable for `extern(C++)`, vtable + monitor for `extern(D)`) and the class' actual payload (for which alignment matters), incl. not forgetting about that when accessing the payload with a GEP and a hardcoded constant.

kinke:
Also to be considered: padding inbetween base class payload and derived class one, e.g., 32-bytes-aligned base class and 64-bytes-aligned derived one, and vice-versa.

kinke:
After #2919, the only remaining issue should be the alignment of GC-allocated memory and instances promoted to stack by `GarbageCollect2Stack`.

PR #2917, opened 2018-11-17
Various tweaks for Android, plus an Alpine fix

joakim-noah:
Android/x64 mostly works, but has some issues because of their unusual choice of having `long double` be a `__float128`, likely to match AArch64.

I'm getting this up early to get feedback. There's also two issues with Phobos I'll look into: [`std.math.copysign` won't compile with `llvm_copysign` on both x86 and x64](https://github.com/joakim-noah/termux-packages/blob/beta/packages/ldc-beta/ldc-phobos-x64.patch#L53) and many of the `std.variant` tests segfault on x64.

Otherwise, only a few tests fail, mostly the same ones as on AArch64 because of a couple trailing bits or missing IEEE Quadruple support:
```
diff --git a/std/complex.d b/std/complex.d
index eef57f864..8a90a98ce 100644
--- a/std/complex.d
+++ b/std/complex.d
@@ -939,7 +939,7 @@ deprecated
     }
     else
     {
-        assert(z1.re == z2.re && z1.im == z2.im);
+        //assert(z1.re == z2.re && z1.im == z2.im);
     }
 }
 
diff --git a/std/math.d b/std/math.d
index 6742ae105..3e5897b1e 100644
--- a/std/math.d
+++ b/std/math.d
@@ -4211,7 +4211,7 @@ real log(real x) @safe pure nothrow @nogc
 ///
 @safe pure nothrow @nogc unittest
 {
-    assert(log(E) == 1);
+    //assert(log(E) == 1);
 }
 
 /**************************************
diff --git a/std/numeric.d b/std/numeric.d
index 40f80440e..2c91ef0c1 100644
--- a/std/numeric.d
+++ b/std/numeric.d
@@ -610,7 +610,7 @@ public:
     }
 }
 
-@safe unittest
+version(none) @safe unittest
 {
     import std.meta;
     alias FPTypes =
@@ -651,7 +651,7 @@ public:
     // @system due to to!string(CustomFloat)
     import std.conv;
     CustomFloat!(5, 10) y = CustomFloat!(5, 10)(0.125);
-    assert(y.to!string == "0.125");
+    //assert(y.to!string == "0.125");
 }
 
 /**
```

### DIFF
--- a/dmd/cppmangle.d
+++ b/dmd/cppmangle.d
@@ -1196,7 +1196,9 @@ extern(C++):
             case Tint128:                c = 'n';       break;
             case Tuns128:                c = 'o';       break;
             case Tfloat64:               c = 'd';       break;
-            case Tfloat80:               c = 'e';       break;
+            case Tfloat80:
+                                         c = 'e';
+                   version (IN_LLVM) goto default; else break;
             case Tbool:                  c = 'b';       break;
             case Tchar:                  c = 'c';       break;
             case Twchar:                 c = 't';       break;  // unsigned short (perhaps use 'Ds' ?
--- a/dmd/cppmangle.d
+++ b/dmd/cppmangle.d
@@ -1196,9 +1196,15 @@ extern(C++):
             case Tint128:                c = 'n';       break;
             case Tuns128:                c = 'o';       break;
             case Tfloat64:               c = 'd';       break;
-            case Tfloat80:
-                                         c = 'e';
-                   version (IN_LLVM) goto default; else break;
+version (IN_LLVM)
+{
+            // there are special cases for D `real`, handled via Target.cppTypeMangle() in the default case
+            case Tfloat80:               goto default;
+}
+else
+{
+            case Tfloat80:               c = 'e';       break;
+}
             case Tbool:                  c = 'b';       break;
             case Tchar:                  c = 'c';       break;
             case Twchar:                 c = 't';       break;  // unsigned short (perhaps use 'Ds' ?
--- a/dmd/target.d
+++ b/dmd/target.d
@@ -109,11 +109,10 @@ struct Target
 
     // implemented in gen/target.cpp:
     static void _init();
-    // Type sizes and support.
     static uint alignsize(Type type);
     static uint fieldalign(Type type);
     static uint critsecsize();
-    static Type va_listType();  // get type of va_list
+    static Type va_listType();
     static int isVectorTypeSupported(int sz, Type type);
     static bool isVectorOpSupported(Type type, TOK op, Type t2 = null);
 
@@ -134,7 +133,6 @@ struct Target
     }
 
     static const(char)* cppTypeMangle(Type t);
-    static Expression getTargetInfo(const(char)* name, const ref Loc loc);
   }
   else // !IN_LLVM
   {
@@ -501,6 +499,19 @@ struct Target
         else
             static assert(0, "fix this");
     }
+
+    /**
+     * Gets vendor-specific type mangling for C++ ABI.
+     * Params:
+     *      t = type to inspect
+     * Returns:
+     *      string if type is mangled specially on target
+     *      null if unhandled
+     */
+    extern (C++) static const(char)* cppTypeMangle(Type t)
+    {
+        return null;
+    }
   } // !IN_LLVM
 
     /**
@@ -536,6 +547,17 @@ struct Target
         return global.params.isWindows ? LINK.windows : LINK.c;
     }
 
+  version (IN_LLVM)
+  {
+    extern (C++):
+
+    static TypeTuple toArgTypes(Type t);
+    static bool isReturnOnStack(TypeFunction tf, bool needsThis);
+    // unused: static ulong parameterSize(const ref Loc loc, Type t);
+    static Expression getTargetInfo(const(char)* name, const ref Loc loc);
+  }
+  else // !IN_LLVM
+  {
     /**
      * Describes how an argument type is passed to a function on target.
      * Params:
@@ -545,19 +567,12 @@ struct Target
      *      empty tuple if type is always passed on the stack
      *      null if the type is a `void` or argtypes aren't supported by the target
      */
-  version (IN_LLVM)
-  {
-    extern (C++) static TypeTuple toArgTypes(Type t);
-  }
-  else
-  {
     extern (C++) static TypeTuple toArgTypes(Type t)
     {
         if (global.params.is64bit && global.params.isWindows)
             return null;
         return .toArgTypes(t);
     }
-  }
 
     /**
      * Determine return style of function - whether in registers or
@@ -568,12 +583,6 @@ struct Target
      * Returns:
      *   true if return value from function is on the stack
      */
-  version (IN_LLVM)
-  {
-    extern (C++) static bool isReturnOnStack(TypeFunction tf, bool needsThis);
-  }
-  else
-  {
     extern (C++) static bool isReturnOnStack(TypeFunction tf, bool needsThis)
     {
         if (tf.isref)
@@ -726,10 +735,7 @@ struct Target
             return false;
         }
     }
-  } // !IN_LLVM
 
-  version (IN_LLVM) {} else
-  {
     /***
      * Determine the size a value of type `t` will be when it
      * is passed on the function parameter stack.
--- a/dmd/target.d
+++ b/dmd/target.d
@@ -133,6 +133,7 @@ struct Target
             return cppTypeInfoMangleItanium(cd);
     }
 
+    static const(char)* cppTypeMangle(Type t);
     static Expression getTargetInfo(const(char)* name, const ref Loc loc);
   }
   else // !IN_LLVM
@@ -501,19 +502,6 @@ struct Target
             static assert(0, "fix this");
     }
   } // !IN_LLVM
-
-    /**
-     * Gets vendor-specific type mangling for C++ ABI.
-     * Params:
-     *      t = type to inspect
-     * Returns:
-     *      string if type is mangled specially on target
-     *      null if unhandled
-     */
-    extern (C++) static const(char)* cppTypeMangle(Type t)
-    {
-        return null;
-    }
 
     /**
      * Get the type that will really be used for passing the given argument
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -524,7 +524,9 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
 void ArgsBuilder::addLinker() {
   if (!opts::linker.empty()) {
     args.push_back("-fuse-ld=" + opts::linker);
-  } else if (global.params.isLinux) {
+  } else if (global.params.isLinux &&
+             global.params.targetTriple->getEnvironment() !=
+                 llvm::Triple::Android) {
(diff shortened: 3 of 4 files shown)


dmd/compiler.d
```diff
@@ -107,7 +107,7 @@ else
 }
         Identifier id = Id.entrypoint;
         auto m = new Module("__entrypoint.d", id, 0, 0);
-        scope p = new Parser!ASTCodegen(m, cmaincode, false);
+        auto p = new Parser!ASTCodegen(m, cmaincode, false);
```
joakim-noah: The reason for all these changes from stack-allocating the parser to GC allocation is that the GC aligns to 128-bit by default, whereas stack allocation aligns to 64-bit.

The problem is that when [`CTFloat.parse` is called in `Lexer.inreal`](https://github.com/ldc-developers/ldc/blob/master/dmd/lexer.d#L2178), it segfaults when trying to assign to `t.floatvalue`, because it expects that union in the token to be 128-bit aligned and uses a `MOVAPS` instruction. If the parser is 128-bit aligned, then so is the token and the union and everything works. I tried putting an `align(16)` on the `Parser` or `Lexer`, but it makes no difference. Maybe getting `scope` to use those alignments is the real fix.

I checked why this doesn't hit on linux/x64, where the stack-allocated parser isn't 128-bit aligned either, and it's because LLVM generates a `FSTPT` when setting `t.floatvalue` there, which doesn't care about alignment.

dmd/cppmangle.d
```diff
@@ -1196,7 +1196,7 @@ extern(C++):
             case Tint128:                c = 'n';       break;
             case Tuns128:                c = 'o';       break;
             case Tfloat64:               c = 'd';       break;
-            case Tfloat80:               c = 'e';       break;
+            case Tfloat80:               c = 'g';       break;
```
joakim-noah: Along with using `__float128` comes its own mangle, as can be seen in the comment block above, but only for Android/x64, llvm-mirror/clang@309b490. I guess I'll have to add a `global.params.isAndroid` and change this for the Android/x64 target.

dmd/dmangle.d
```diff
@@ -913,7 +913,7 @@ public:
         {
             enum BUFFER_LEN = 36;
             char[BUFFER_LEN] buffer;
-            const n = CTFloat.sprint(buffer.ptr, 'A', value);
+            const n = CTFloat.sprint(buffer.ptr, 'g', value);
```
joakim-noah: `sprint`/`printf` from Bionic/x64 segfault for certain `long double` values passed to them with the hex format, at least in the Anbox Android/x64 environment I'm testing in, termux/termux-packages#3058. Maybe they gave up on implementing hex formatting in Bionic, as they simply closed this issue, android-ndk/ndk#437, and the Termux issue shows the garbled output on AArch64 too.

At least it doesn't segfault on AArch64, so the tests still pass, whereas a few tests in `std.conv` and `std.algorithm.iteration` get ldc to segfault in this `sprintf` when it's passed some floating-point numbers, such as `1.414`. Simply using decimal formatting works around the Android/x64 bug.

dmd/compiler.d
kinke: So the C++ code assumes a 16-bytes alignment for 128-bit `long double`, which seems to make perfect sense. Judging by your comment and a superficial quick glance, the issue seems to be that the `Token` (or its nested union's) alignment on the D side doesn't match (as the Tokens seem to be allocated in D IIRC). And I would guess that that's due to the used host compiler's `real.alignof` not returning 16, which should be trivial to check. You can also just extract the `Token` declaration and check its alignment (and the union's offset) with the host compiler.

kinke: I played around with run.dlang.io, and it looks as if `Token` is 16-bytes aligned, and the union's offset 64 bytes: https://run.dlang.io/is/WnzZVI

If so, then these Tokens must be somehow wrongly allocated (in IR, we obviously use `align 16` for locals on the stack).

kinke: After some more tests, I think it's `scope`; the `Parser` instance allocated on the stack only has an 8-bytes alignment (corresponding to the class *reference* alignment).

JohanEngelen: The misalignment of `scope` class objects is a pretty big bug imo! Nice find.

kinke: [related: https://github.com/ldc-developers/ldc/issues/1692]

joakim-noah: OK, given the long-standing issue noting this misalignment, how should we deal with this? I can move this and the `sprintf` segfault workarounds to local patches in the Termux repo, as they only affect native Android/x64 builds, not cross-compiling. Don't want to clutter LDC up with hacks for such a niche platform as Android/x64.

kinke: Seems like a good reason to finally fix it, at least partially. As making sure `scope` allocations use the actual class alignment seems to be enough in this case, that'd be a good start.

dmd/cppmangle.d
```diff
@@ -1196,7 +1199,10 @@ extern(C++):
             case Tint128:                c = 'n';       break;
             case Tuns128:                c = 'o';       break;
             case Tfloat64:               c = 'd';       break;
-            case Tfloat80:               c = 'e';       break;
+            case Tfloat80:
+                // LDC: `long double` on Android/x64 is __float128 and mangled as `g`.
+                c = global.params.isAndroid && isArchx86_64() ? 'g' : 'e';
```
joakim-noah: Apparently Kai added this before but Iain removed it, dlang/dmd#6887, because GDC probably doesn't reuse `Tfloat80` for 128-bit floats, so this doesn't hit and goes to the `Target.cppTypeMangle` in the default case below instead.

kinke: That is probably preferable; the C++ mangling of `Tfloat80` representing D `real` / C++ `long double` is target-specific with this now too (and most likely would be for proper PowerPC support too).

joakim-noah: Updated, let me know if this new commit is what you had in mind, passes the tests.

kinke:
I'd prefer `Target::cppTypeMangle()` to be implemented in our `gen/target.cpp`, where all or most of our `Target` specifics reside. It probably doesn't pay off to forward to a new `TargetABI` method yet for this one special case. You have full access to the target triple on the C++ side and can get rid of most of the front-end adaptations (`isAndroid` etc., which all need to be guarded by `version (IN_LLVM)` btw, so pls don't modify upstream comments etc.).

joakim-noah:
Alright, changed the mangling in the file you wanted.

Ignore the Phobos commits, that's not going to be in the final version of this pull. I simply figured that if this pull is going to kick off a Shippable build, might as well take advantage and check some AArch64 stuff too.

kinke:
Thx; I re-added the (versioned-out) upstream `Target.cppTypeMangle()`, slightly modified the `cppmangle.d` diff and reorganized a few things in `gen/target.cpp`/`dmd/target.d` for improved consistency.

kinke:
So this is still working as intended with `Target.cppTypeMangle()`, right? Did you check `const` and indirections (`ref real`, `real*`) too? Then feel free to merge after removing the Phobos advancement.

joakim-noah:
Haven't checked on Android/x64 yet with your changes or those additional cases, also want to get the std.variant/math issues ironed out. Need to look into those.

joakim-noah:
Alright, looked into the `std.variant` failures more, [two](https://github.com/ldc-developers/phobos/blob/fd5facfe0456d8f7ae16af8415f1ea01606c917f/std/variant.d#L1317) [tests](https://github.com/ldc-developers/phobos/blob/fd5facfe0456d8f7ae16af8415f1ea01606c917f/std/variant.d#L1697) trip because of alignment issues when trying to access 128-bit reals. That first one trips because it tries comparing the struct members stored in `MyVariant` one by one, and segfaults when trying to load the 128-bit real at `S.d` with a `MOVAPS` instruction, which requires 128-bit alignment but the real is only 64-bit aligned. Checking the alignment and offset by inserting a `pragma(msg, "V alignment is ", v.alignof, " and offset of struct is ", v.store.offsetof);` gives 8LU for both.

OK, that's expected given the union it's using, so I tried sticking an `align(16)` on the union in `VariantN` just to see if it worked around the issue. After that, all the tests pass if I compile that module alone, but not if I build the full `phobos2-test-runner`. I start getting other tests segfaulting then, such as [this block](https://github.com/ldc-developers/phobos/blob/fd5facfe0456d8f7ae16af8415f1ea01606c917f/std/variant.d#L39) and [this one](https://github.com/ldc-developers/phobos/blob/fd5facfe0456d8f7ae16af8415f1ea01606c917f/std/variant.d#L1199), presumably other alignment issues for the codegen.

Not unexpected given the union and all the casts I see scattered around, but not sure the best way to fix this. I guess `std.variant` happens to work with 64-bit alignment and AArch64 128-bit reals that don't require 128-bit alignment, but not on Android/x64.

joakim-noah:
Regarding the last `std.math` issue, ldc itself errors out in the backend when trying to use [`llvm_copysign` in `std.math.copysign`](https://github.com/ldc-developers/phobos/blob/ldc/std/math.d#L7277) on Android/x86 or x64:
```
LLVM ERROR: Cannot select: 0x7f56d69fac98: ch = store<(store 8 into %stack.1), trunc to f64> 0x7f56d2dd3758, 0x7f56d637b6e8, FrameIndex:i64<1>, undef:i64
0x7f56d637b6e8: f128,ch = CopyFromReg 0x7f56d2dd3758, Register:f128 %0
    0x7f56d0ff8270: f128 = Register %0             
0x7f56d637b270: i64 = FrameIndex<1>                                                               
0x7f56d571d270: i64 = undef
In function: _D3std4math5asinhFNaNbNiNfdZd
```
It works if I use the straight D bit-casting version just below or replace `llvm_copysign` with `core.stdc.math.copysignl` instead, which I believe is all the frontend is trying to do. I can dump the IR when using `llvm_copysign` just fine with `--output-ll`, implying the problem is with the IR we're sending to LLVM.

@kinke, you know this stuff best, what do you suggest?

joakim-noah:
>It clearly needs to use at least the max alignment of AllowedTypes.

[Like this](https://github.com/dlang/phobos/pull/6790/files#diff-7a8a8126c2ec19a23ab8601c061d4018)? All `std.variant` tests pass with that patch on Android/x64, and seem to be working most everywhere else too, only one third-party static assert in dyaml.

joakim-noah:
Pulled in the upstream Phobos commits too, will check those Android/x64 ABI cases you mentioned and this should be ready to go.

Once LLVM 7.0.1 is out, release 1.13 this weekend?

kinke:
Yeah, that'd be the current plan.

joakim-noah:
Alright, looked over your commit and checked those other Android/x64 mangles: `const` does nothing and `ref`/`*` mangle to `Rg` and `Pg`, just like clang does. Tests all pass, I'm done, so merge when ready.

kinke:
The mangling adaptation introduced a regression wrt. backreferences (see https://forum.dlang.org/thread/dsrjmpnavxqsjqmlvotf@forum.dlang.org):
```D
extern(C++) real _modulo(real x, real y);
pragma(msg, _modulo.mangleof);
```

`_Z7_moduloee` for LDC < 1.13 and DMD 2.083, `_Z7_moduloeS_` for LDC 1.13.

JohanEngelen:
--> https://github.com/ldc-developers/ldc/pull/2953

PR #2919, opened 2018-11-19
Fix alignment of scope-allocated class instances

kinke:


kinke:
I noticed that an `alignof(N) class ...` seems to be totally ignored; i.e., it has no effect on `Type::alignment()` and neither on `ClassDeclaration::alignsize()`.

I checked the paddings too (and the tests cover them as well); they seem fine already, both inbetween header and payload (and for both D and C++ classes, the latter missing the monitor field) as well as inbetween base payload and derived payload.

kinke:
Ah nice, removing `-release` for the frontend if using an LLVM with enabled assertions already pays off - `struct S { align(16) int a; }` is quite exotic, as it's 16-bytes large, but only the first quarter is actually used, leaving the 2nd 64-bit chunk totally to padding, and the SysV ABI classification doesn't handle that (and neither do the ABI docs).

clang passes it as a single `i32` (i.e., in just one GP register), ignoring the padding altogether. With 32-bytes alignment, clang passes it on the stack, so we don't have to consider types > 16 bytes (except for (wrapped) vectors) as still potentially being passed in registers due to lots of trailing padding.

gen/classes.cpp
```diff
@@ -83,7 +83,10 @@ DValue *DtoNewClass(Loc &loc, TypeClass *tc, NewExp *newexp) {
   // allocate
   LLValue *mem;
   if (newexp->onstack) {
-    mem = DtoRawAlloca(DtoType(tc)->getContainedType(0), DtoAlignment(tc),
+    unsigned alignment = tc->sym->alignsize;
+    if (alignment == STRUCTALIGN_DEFAULT)
+      alignment = 0;
```
JohanEngelen: probably also good to add a testcase to the `align_class.d` file for this branch.

kinke: [I'm not sure when that is the case, maybe for an empty class without any fields. `STRUCTALIGN_DEFAULT` is `~0`, so never a power of 2, so I'd guess LLVM immediately asserts for such garbage.]

kinke: I checked; it's always 8-bytes alignment (x64) for empty D/C++ classes, and classes with a single naturally aligned field.

kinke:
Btw, clang also doesn't zero out the unpassed padding: https://godbolt.org/z/J-Bb2m

joakim-noah:
Thanks, I rebuilt LDC 1.13 using this patch and can confirm it fixes the issue on Android/x64.

PR #2921, opened 2018-11-21
fix compilation with recent LLVM

thewilsonator:
https://reviews.llvm.org/D54597


gen/dibuilder.cpp
```diff
@@ -1009,17 +1013,35 @@ DISubprogram DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
     // The return type is a nested struct, so for this particular
     // chicken-and-egg case we need to create a temporary subprogram.
     irFunc->diSubprogram = DBuilder.createTempFunctionFwdDecl(
-        scope, name, linkageName, file, lineNo, /*ty=*/nullptr, isLocalToUnit,
-        isDefinition, scopeLine, flags, isOptimized);
+        scope, name, linkageName, file, lineNo, /*ty=*/nullptr,
+#if LDC_LLVM_VER <= 700
```
kinke: Pls use `< 800`, as this won't work for LLVM 7.0.1.

gen/dibuilder.cpp
```diff
@@ -1009,17 +1013,35 @@ DISubprogram DIBuilder::EmitSubProgram(FuncDeclaration *fd) {
...
+                                    isLocalToUnit, isDefinition,
+#endif
+                                    scopeLine, flags,
+#if LDC_LLVM_VER >= 800
+                                    dispFlags
+#else
+                                    isOptimized
+#endif
+        );
```
kinke: As we have 3 `DBuilder.createFunction()` occurrences to be adapated, I'd refactor this into a `DIBuilder::createFunction()` helper. Just a suggestion, no requirement. ;)

gen/dibuilder.cpp
thewilsonator: done

gen/dibuilder.cpp
thewilsonator: done

PR #2715, opened 2018-05-26
Some ABI refactoring, incl. efficiency improvements for IndirectByvalRewrite

kinke:
Unoptimized old IR for the included lit-test:
```LL
define void @_D22indirect_byval_rewrite3barFZv() #0 comdat {
  %bigLValue = alloca %indirect_byval_rewrite.Big, align 8 ; [#uses = 2, size/byte = 16]
  %withPostblitLValue = alloca %indirect_byval_rewrite.WithPostblit, align 8 ; [#uses = 2, size/byte = 16]
  %.ExplicitByvalRewrite_dump = alloca %indirect_byval_rewrite.Big, align 16 ; [#uses = 2, size/byte = 16]
  %__copytmp3 = alloca %indirect_byval_rewrite.WithPostblit, align 8 ; [#uses = 3, size/byte = 16]
  %.ExplicitByvalRewrite_dump1 = alloca %indirect_byval_rewrite.WithPostblit, align 16 ; [#uses = 2, size/byte = 16]
  %.sret_tmp = alloca %indirect_byval_rewrite.Big, align 8 ; [#uses = 2, size/byte = 16]
  %.ExplicitByvalRewrite_dump2 = alloca %indirect_byval_rewrite.Big, align 16 ; [#uses = 2, size/byte = 16]
  %.structliteral = alloca %indirect_byval_rewrite.WithPostblit, align 8 ; [#uses = 2, size/byte = 16]
  %.ExplicitByvalRewrite_dump3 = alloca %indirect_byval_rewrite.WithPostblit, align 16 ; [#uses = 2, size/byte = 16]
  %1 = bitcast %indirect_byval_rewrite.Big* %bigLValue to i8* ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %1, i8 0, i64 16, i32 1, i1 false)
  %2 = bitcast %indirect_byval_rewrite.WithPostblit* %withPostblitLValue to i8* ; [#uses = 1]
  call void @llvm.memset.p0i8.i64(i8* %2, i8 0, i64 16, i32 1, i1 false)
  %3 = bitcast %indirect_byval_rewrite.Big* %.ExplicitByvalRewrite_dump to i8* ; [#uses = 1]
  %4 = bitcast %indirect_byval_rewrite.Big* %bigLValue to i8* ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %3, i8* %4, i64 16, i32 8, i1 false)
[...]

PR #2927, opened 2018-11-27
Assigning T.init should only create a temporary when needed.

JohanEngelen:
PR #2715 introduced a regression where assigning `T.init` creates a temporary when that's not needed in unoptimized builds. Although technically not a problem, it becomes a problem when `T.init` is very large, and the user runs out of stackspace. Weka has hit this regression in unoptimized builds; Weka has a workaround function that people should use instead of writing `a = a.init`, but we'd rather not have the compiler do these silly things :-)

The commit that introduced the regression did that to fix another bug, see https://github.com/ldc-developers/ldc/pull/2715/commits/c01dfc3db85c3df3dbbb980f76541f5337347a48
By language spec `T.init` is an rvalue (or at least by compiler behavior, cannot take address of `T.init` for example), so we have to be careful to treat it that way (e.g. when function calls are involved) which is what #2715 does.

This PR adds a codegen optimization to attempt in-place construction for assignments of struct ~~and array literals~~ `.init`. This solves the regression.  (DMD seems to do the same in unoptimized builds)

Note: it is unfortunate that this adds more noise to an already somewhat hard-to-grasp `toir.cpp`...


tests/codegen/assign_struct_init_without_stack.d
```diff
@@ -0,0 +1,82 @@
...
+    globalStruct = FloatStruct([1, 0, 0, 0, 0, 0, 0, 0, 0, 42]);
+    // There should be only one memcpy.
+    // CHECK: memcpy
+    // CHECK-NEXT: ret void
+}
+
+// CHECK-LABEL: _D32assign_struct_init_without_stack3gggFZv
+void ggg() {
+    globalStruct = opaque(globalStruct);
+    // There should be one memcpy from a temporary (sret return).
```
kinke: This probably fails for Win64 - the ABI mandates the caller makes the copy and passes a ref, so there'll be another memcpy.

kinke:
Did you check that it's safe to `inPlaceConstruct()` into an existing variable? Wrt. that it's correctly destructed before resetting it.

tests/codegen/assign_struct_init_without_stack.d
```diff
@@ -0,0 +1,82 @@
+// Tests that even in non-optimized builds, member = member.init immediately does a memcpy,
+// instead of going through a temporary stack allocated variable.
+// Exception: when member has an opAssign with auto ref parameter, then the .init symbol
+// apparently counts as an rvalue so a local temporary has to be passed.
```
kinke: `T.init` is an rvalue. I guess the simple reason is that the assignment in the AST isn't an `AssignExp` anymore, but a `CallExp` (`globalOpAssignStruct.opAssign(OpAssignStruct.init)`).

tests/codegen/assign_struct_init_without_stack.d
kinke: The expected copy would be elided if you remove your conditions wrt. struct/array literal; an sret call as rhs is supported by `toInPlaceConstruction()` and makes perfect sense here as well.

kinke: The only real special case I can think of would be that the result variable in an out contract of `opaque(ref FloatStruct)` might alias the `globalStruct` arg passed by ref. I don't see that as an issue though.

tests/codegen/assign_struct_init_without_stack.d
JohanEngelen: Yep, that's the comment in the toir code.

JohanEngelen:
This is tricky, e.g. in `runnable/structlit.d`:
```d
void test8738()
{
    int[3] a = [1, 2, 3];

    struct S { int a, b, c; }
    S s = S(1, 2, 3);

    a = [4, a[0], 6]; // needs temporary
    s = S(4, s.a, 6); // needs temporary

    assert(a == [4, 1, 6]);
    assert(s == S(4, 1, 6));
}
```

Let's move in steps: I'll change this PR to only work for `.init` where there can be no complications like this example above.

kinke:
Yep, sounds good. I don't think the other tricky potential optimizations aren't worth it, deferring those to LLVM is fine IMO.

tests/codegen/assign_struct_init_without_stack.d
```diff
@@ -0,0 +1,115 @@
+// Tests that even in non-optimized builds, member = member.init immediately does a memcpy,
+// instead of going through a temporary stack allocated variable.
+// Exception: when member has an opAssign with auto ref parameter a local temporary
+// has to be passed because `.init` is an rvalue,
```
kinke: I still find the last sentence a bit misleading. It has nothing to do with `auto ref`, it's just a regular rvalue arg for a function call, and we don't really have struct/static array rvalues. They are lvalues, always, except when being rewritten to fit into register(s) during function passing.
Aliasing the init symbol as we did earlier is a problem if we assume that an rvalue expression in memory can safely be reused as lvalue without extra copy, which is what #2715 implemented to optimize `IndirectByvalRewrite` (and which we'll use a lot if the postMove DIP is ever to be implemented). If `T.init` for a struct was an lvalue expression, we wouldn't have this problem, but that would probably require it to be of `immutable T` type and not be consistent with `T.init` for primitive types etc.

tests/codegen/assign_struct_init_without_stack.d
```diff
@@ -0,0 +1,115 @@
...
+    // Future work: test optimized codegen (at -O0)
+}
+
+// CHECK-LABEL: define{{.*}} @{{.*}}_D32assign_struct_init_without_stack3gggFZv
+void ggg() {
+    globalStruct = opaque(globalStruct);
+    // There should be one memcpy from a temporary (sret return).
+    // CHECK: alloca %assign_struct_init_without_stack.FloatStruct
+    // CHECK: call
+    // CHECK: call void @llvm.memcpy
```
kinke: So for Win64, the first `call` above doesn't match the `opaque()` call, but the first memcpy, so this works now - but so it doesn't really check that there's a single temporary, and it obviously cannot as long as it should work for Win64 too, so this test is kind of awkward.
What you can do though is change the signature of `opaque()`, either using no args at all or some primitive type.

JohanEngelen: this was testing two things at once, split the test into two. Explicitly tests for sret_tmp now too

PR #2931, opened 2018-12-07
Add __ir to ldc.llvmasm, to replace ldc.simd.inlineIR.

JohanEngelen:
Inline IR is more generally applicable than SIMD, so it is out-of-place in `ldc.simd`. Because it is so similar in functionality to `__asm` and LLVM IR is called LLVM assembly, I think `__ir` fits better and should be in `ldc.llvmasm`. I've also added `*_trusted` variants similar to `__asm_trusted`.
For backwards compatibility, `ldc.simd.inlineIR` is still there (forwards to `ldc.llvmasm.__ir_trusted`.

### DIFF
--- a/tests/codegen/inlineIR_math.d
+++ b/tests/codegen/inlineIR_math.d
@@ -1,4 +1,4 @@
-// Tests inlineIR + math optimizations
+// Tests inline IR + math optimizations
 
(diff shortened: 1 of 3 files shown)


kinke:
Nice, I found it badly misplaced as well.

PR #2933, opened 2018-12-08
Merge v2.083.1

kinke:


kinke:
[LLVM v7.0.1 is to be released these days, so I'll wait a bit anyway.]

thewilsonator:
Its still marked WIP, is it good to go?

joakim-noah:
No, let me modify that Phobos pull a bit: Iain points out gdc doesn't need one of those tweaks and I may want to modify `std.variant` too, before the 1.13 release.

PR #2938, opened 2018-12-13
Always use PIC on FreeBSD, just like on Linux

myfreeweb:
This fixes errors like

```
[…]
/usr/bin/ld: error: can't create dynamic relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol: _D4core5cpuid6_ssse3yb in reado
nly segment; recompile object files with -fPIC                                                                               
>>> defined in runtime/objects-shared/core/cpuid.o
>>> referenced by cpuid.d                                                                                                    
>>>               runtime/objects-shared/core/cpuid.o:(_D4core5cpuid5ssse3FNaNbNdNiNeZb)                                     

/usr/bin/ld: error: can't create dynamic relocation R_AARCH64_LDST8_ABS_LO12_NC against symbol: _D4core5cpuid6_ssse3yb in read
only segment; recompile object files with -fPIC
>>> defined in runtime/objects-shared/core/cpuid.o
>>> referenced by cpuid.d
>>>               runtime/objects-shared/core/cpuid.o:(_D4core5cpuid5ssse3FNaNbNdNiNeZb)                                     
                                                                                                                             
/usr/bin/ld: error: can't create dynamic relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol: _D4core5cpuid6_sse41yb in reado
nly segment; recompile object files with -fPIC                                                                               
>>> defined in runtime/objects-shared/core/cpuid.o                                                                           
>>> referenced by cpuid.d
>>>               runtime/objects-shared/core/cpuid.o:(_D4core5cpuid5sse41FNaNbNdNiNeZb)                                     

/usr/bin/ld: error: can't create dynamic relocation R_AARCH64_LDST8_ABS_LO12_NC against symbol: _D4core5cpuid6_sse41yb in read
only segment; recompile object files with -fPIC
>>> defined in runtime/objects-shared/core/cpuid.o
>>> referenced by cpuid.d
>>>               runtime/objects-shared/core/cpuid.o:(_D4core5cpuid5sse41FNaNbNdNiNeZb)
```

and allows ASLR on [HardenedBSD](https://hardenedbsd.org) (/ "coming soon" ASR on upstream FreeBSD) to do its thing

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -423,6 +423,10 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
       // these OSes.
       // On Android, PIC is default as well.
       relocModel = llvm::Reloc::PIC_;
+    } else if (triple.isOSFreeBSD()) {
+      // We default to PIC code to avoid linking issues on FreeBSD, especially
+      // on aarch64.
+      relocModel = llvm::Reloc::PIC_;
     } else {
       // ARM for other than Darwin or Android defaults to static
       switch (triple.getArch()) {


kinke:
This shouldn't happen when linking shared druntime. The D modules are/should be explicitly compiled with `-relocation-model=pic`: https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L596

myfreeweb:
This was coming from the [variants macro](https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L610) I think. (But adding `-relocation-model=pic` there resulted in complaining about this option used two times.)

myfreeweb:
hm. I actually ended up with these after screwing with the CMake files. The original errors are in curl

```
/usr/bin/ld: error: can't create dynamic relocation R_AARCH64_ADD_ABS_LO12_NC against symbol: _D6object__T7destroyTS3std3net4curl4SMTP4ImplZQBiFKQBeZ9__lambda3MFNaNbNiNeZv in readonly segment; recompile object files with -fPIC                         
>>> defined in runtime/objects-debug/etc/c/curl.o
>>> referenced by object.d:530 (/home/greg/src/github.com/ldc-developers/ldc/runtime/druntime/src/object.d:530)              
>>>               runtime/objects-debug/etc/c/curl.o:(_D6object__T7destroyTS3std3net4curl4SMTP4ImplZQBiFKQBeZv)
```

kinke:
This is Phobos now, which should be compiled once as PIC too [without `BUILD_SHARED_LIBS=OFF`; added in `build_runtime_variant()` too]. What's the library failing to link?

kinke:
Argh... `list(APPEND ${phobos2_d_flags} ...)` in line 566 should most likely be `list(APPEND phobos2_d_flags ...)`.

kinke:
Thx for checking. So is defaulting to PIC for FreeBSD safe/needed already, or would that fix suffice for the moment?

Issue #2937, opened 2018-12-12
Crash while building ErupteD sample code on Windows x64


GoaLitiuM:
LDC2 1.13.0-beta2 x64 build for Windows crashes while building this minified [ErupteD](https://github.com/ParticlePeter/ErupteD) sample code:

```D
import erupted;
import erupted.vulkan_lib_loader;
import erupted.platform_extensions;
import core.sys.windows.windows;
mixin Platform_Extensions!USE_PLATFORM_WIN32_KHR;
int main() { return 0; }
```

```
ldc2 -g -m64 -I=erupted\source source\main.d erupted\source\erupted\functions.d erupted\source\erupted\types.d erupted\source\erupted\platform_extensions.d erupted\source\erupted\vulkan_lib_loader.d erupted\source\erupted\dispatch_device.d
invalid scope
!6378 = !DICompositeType(tag: DW_TAG_structure_type, name: "VkMemoryGetWin32HandleInfoKHR", scope: Wrote crash dump file "C:\Users\GOALIT~1\AppData\Local\Temp\ldc2.exe-370919.dmp"
0x00007FF6CE901ADE (0x000000CDF9FFDBE8 0x00007FF6CEB1A36A 0x0000000000000000 0x0000000000000000)
0x00007FF6CE90CAB3 (0x0000000000000000 0x0000000000000000 0x0000000000000000 0x0000000000000000)
0x00007FF6CE906F38 (0x7FFFFFFFFFFFFFFF 0x0000000000000000 0x0000000000000000 0x0000000600000114)
0x00007FF6CE8FF85E (0x00007FF6CFAEA870 0x0000026F83228C88 0x0000000000000000 0x00007FF6CEB1A160)
0x00007FF6CE8FDC47 (0x0000026F8486F758 0x00007FF6CE8EA901 0x0000EF26E504E80F 0x0AFFFFFFFFFFFFF0)
0x00007FF6CE8D24F9 (0x0000026F846C2770 0x0000026F83228A30 0xFFFFFFFFFFFFFFF8 0x0000026F83228B68)
0x00007FF6CE8CF647 (0x0000000000000040 0x0000026F00000002 0xFFFFFFFFFFFFFFF0 0x0000026F83228B68)
0x00007FF6CE8CFCC8 (0x0400010000000000 0x0000026F846C2770 0xFFFFFFFFFFFFFFF8 0x0000026F83228B68)
[...]

PR #2939, opened 2018-12-14
Fix issue #2937 (DI scope for types in static foreach in mixin templates)

kinke:
The test is Windows-only, but tests more than just compilability.


JohanEngelen:
About the test: I much prefer if we'd try to keep the number of specific-platform-only tests to a minimum. But apply your own judgement here for the debuginfo testing.

Issue #2932, opened 2018-12-08
LDC segmentation fault


yshui:
Unintentionally caused a ldc crash: https://d.godbolt.org/z/cLZ3QH

I don't know what to put in the title, sorry.

JohanEngelen:
Testcase:
```d
void main() {
    int[] a, tmp;
    import std.algorithm;
    pragma(msg, typeof(a.map!((a) => {return tmp;})));
}
```
LDC 1.13.0-beta2

joakim-noah:
Doesn't work for every LDC I tried on godbolt or the latest two gdc versions either, only works for DMD.

kinke:
Also happens for `pragma(msg, typeof(a.map!(a => tmp)));`. The reason is that the delegate literal is codegen'd (although actually speculative); resolving nested var `tmp` in there doesn't work, as the outer `main()` has no actually captured locals. - Ideally, codegen would be skipped.

kinke:
`static if (is(typeof(...)))` works fine, it's detected as speculative in that case. The `pragma(mangle, <exp>)` expression isn't.

PR #2940, opened 2018-12-14
Fix issue #2932 (allow speculative nested variables)

kinke:


gen/nested.cpp
```diff
@@ -84,11 +100,11 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
   IF_LOG { Logger::cout() << "Context: " << *ctx << '\n'; }
 
   DtoCreateNestedContextType(vdparent->isFuncDeclaration());
-  assert(isIrLocalCreated(vd));
```
kinke: @klickverbot: I guess it would be safer to do the check after the `DtoCreateNestedContextType()` call?

gen/nested.cpp
```diff
@@ -84,11 +84,27 @@ DValue *DtoNestedVariable(Loc &loc, Type *astype, VarDeclaration *vd,
...
+  // with `-allinst`, see https://github.com/ldc-developers/ldc/issues/2932).
+  // Use an invalid null storage in that case, so that accessing the var at
+  // runtime will cause a segfault.
+  if (irLocal->nestedIndex == -1) {
+    Logger::println(
+        "WARNING: isn't actually nested, using invalid null storage");
+    auto llType = DtoPtrToType(astype);
+    if (isSpecialRefVar(vd))
+      llType = llType->getPointerTo();
+    return makeVarDValue(astype, vd, llvm::ConstantPointerNull::get(llType));
```
kinke: Unsure as to whether `null` or `undefined` makes more sense here.

kinke:
Are we okay with this? We probably all agree that such functions shouldn't be emitted in the first place, but in the meantime, this hopefully allows people to play around with `-allinst` as workaround for linker errors due to template culling (last time I suggested this to someone, he replied that LDC immediately crashed, and I bet due to this).

gen/nested.cpp
klickverbot: If you don't expect it to be used (or only from unused code), I'd use `undef`.

gen/nested.cpp
klickverbot: Sorry, it's been a while – I'd need to look into this myself.

gen/nested.cpp
kinke: Heh yeah, those functions *shouldn't* be called, but we could just as well have a bug somewhere leading to this, where we wouldn't bail out anymore. I now lean more towards `null` for deterministic segfaults; e.g., the `std.random` 2.084 issue (`undef` initial seed) was only noticeable on 32-bit Linux.

JohanEngelen: Using `null` would not guarantee segfaults, but it clearly defines accessing it as UB (unless with `"null-pointer-is-valid"=true`, but that would then result in segfaulting). I don't think that dereferencing `undef` is UB in LLVM IR. So I too think `null` is the better choice here (optimizer can perhaps also use it e.g to remove dead code).

klickverbot: Makes sense

PR #2946, opened 2018-12-18
Upgrade front-end & libs to v2.084.0

kinke:


kinke:
Somebody pls have a look at some of the later commits - especially `DtoCheckPragma` & `ldc.arrayinit` - as they got a bit more invasive than initially expected. I only tackled the latter as preparation for the remaining `runnable/betterc.d` issue, where DMD now inlines `_d_arraycopy` for `-betterC` (in their backend, not in the AST).

Besides `runnable/betterc.d`, which strangely passes on Linux (wtf?), there are many strange 32-bit dmd-testsuite crashes on Linux I couldn't reproduce in my Ubuntu 18.04 VM. The rest should be green if I haven't missed anything.

gen/passes/SimplifyDRuntimeCalls.cpp
```diff
@@ -291,6 +293,9 @@ struct LLVM_LIBRARY_VISIBILITY ArraySliceCopyOpt : public LibCallOptimization {
 
     // Equal length and the pointers definitely don't alias, so it's safe to
     // replace the call with memcpy
+    auto Size = Sz != llvm::MemoryLocation::UnknownSize
```
JohanEngelen: more clear to move this 10 lines up and merge it with the `std::uint64_t Sz = llvm::MemoryLocation::UnknownSize;` bit. (such that you don't need `Sz` at all).

gen/arrays.cpp
```diff
@@ -912,7 +914,7 @@ DSliceValue *DtoCatArrays(Loc &loc, Type *arrayType, Expression *exp1,
   }
 
   auto newArray =
-      gIR->funcGen().callOrInvoke(fn, args, ".appendedArray").getInstruction();
+      gIR->CreateCallOrInvoke(fn, args, ".appendedArray").getInstruction();
```
JohanEngelen: the replacement of `funcGen().callOrInvoke` with `CreateCallOrInvoke` will no longer put attributes (such as `readonly`, `nocapture`) on the call site. Is that OK because the attributes are still visible on the runtime function declarations?

JohanEngelen:
Is there anything in particular you are worried about in those commits? I looked at them and, although a lot is going on, it looks OK to me. There is probably a lack of tests for all those pragmas, so can't be sure of no breakage.

gen/arrays.cpp
kinke: Hmm, `CreateCallOrInvoke` is just for convenience and forwards to `funcGen().callOrInvoke`. I added a new overload taking an `ArrayRef` and the optional `isNothrow` arg, which was previously only exposed via `funcGen`, so the functionality is the same. It's just the old `CreateCallOrInvoke` overloads which don't feature the `isNothrow` arg, but all forward to the `funcGen` method.

kinke:
Nothing in particular, it's just the `a lot is going on`. ;)

gen/passes/SimplifyDRuntimeCalls.cpp
kinke: Yep, good idea.

gen/arrays.cpp
JohanEngelen: ah, oops, I thought it was the other way around (`funcGen().callOrInvoke` forwarding to `CreateCallOrInvoke`). Sorry for the noise.

kinke:
The slice copy optimization wasn't really needed either, but I thought it deserves a bit of attention.
```D
int[10] a1 = void;
int[10] a2 = void;
a1[] = a2[];
```

Old IR:
```LL
  %a1 = alloca [10 x i32], align 4                ; [#uses = 2, size/byte = 40]
  %a2 = alloca [10 x i32], align 4                ; [#uses = 1, size/byte = 40]
  %1 = bitcast [10 x i32]* %a2 to i32*            ; [#uses = 1]
  %2 = insertvalue { i64, i32* } { i64 10, i32* undef }, i32* %1, 1 ; [#uses = 2]
  %3 = bitcast [10 x i32]* %a1 to i32*            ; [#uses = 1]
  %4 = bitcast i32* %3 to i8*                     ; [#uses = 1]
  %.ptr = extractvalue { i64, i32* } %2, 1        ; [#uses = 1]
  %5 = bitcast i32* %.ptr to i8*                  ; [#uses = 1]
  %.len = extractvalue { i64, i32* } %2, 0        ; [#uses = 1]
  %6 = mul i64 %.len, 4                           ; [#uses = 1]
  call void @_d_array_slice_copy(i8* nocapture %4, i64 40, i8* nocapture %5, i64 %6) #1
  %7 = bitcast [10 x i32]* %a1 to i32*            ; [#uses = 1]
  %8 = insertvalue { i64, i32* } { i64 10, i32* undef }, i32* %7, 1 ; [#uses = 0]
  ret void
```

New IR:
```
  %a1 = alloca [10 x i32], align 4                ; [#uses = 2, size/byte = 40]
  %a2 = alloca [10 x i32], align 4                ; [#uses = 1, size/byte = 40]
  %1 = bitcast [10 x i32]* %a2 to i32*            ; [#uses = 2]
  %2 = insertvalue { i64, i32* } { i64 10, i32* undef }, i32* %1, 1 ; [#uses = 0]
  %3 = bitcast [10 x i32]* %a1 to i32*            ; [#uses = 1]
  %4 = bitcast i32* %3 to i8*                     ; [#uses = 1]
  %5 = bitcast i32* %1 to i8*                     ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 1 %4, i8* align 1 %5, i64 40, i1 false)
  %6 = bitcast [10 x i32]* %a1 to i32*            ; [#uses = 1]
  %7 = insertvalue { i64, i32* } { i64 10, i32* undef }, i32* %6, 1 ; [#uses = 0]
  ret void
```

This optimization doesn't check for overlap, but neither does the existing optimization for static arrays on both sides (`a1 = a2`).

---

As to the actual `_d_array_slice_copy` -betterC issue, I'm just not convinced inlining in each backend is the way to go, considering the number of still missing similar functions and the hassle of emitting such code (compared to writing 3 lines in D), so I disabled that subtest for now.

kinke:
Btw, the bounds are still checked for
```D
int[10] a1 = void;
int[10] a2 = void;
a1[0..256] = a2[0..256];
```

kinke:
Travis OSX (64-bit only):
```
c++ -m64 -c runnable/extra-files/cpp_stdlib.cpp -o ../../../../ninja-ldc/dmd-testsuite-debug/runnable/cpp_stdlib.cpp.o -std=c++11 -std=c++11
runnable/extra-files/cpp_stdlib.cpp:1:10: fatal error: 'array' file not found
2980: #include <array>
2980:          ^~~~~~~
2980: 1 error generated.
```

Any idea what's missing? Incomplete Xcode installation?!

kinke:
Btw @JohanEngelen, the strange LLVM 7.0 crashes for Travis are apparently back, master failed too...

thewilsonator:
> `-std=c++11 -std=c++11`

One of those should be a `-stdlib=libc++`.

JohanEngelen:
> considering the number of still missing similar functions

How many functions are we talking about?
What is `_d_array_slice_copy` supposed to do with `-betterC` when the slices overlap, lengths disagree, or nullptrs?

kinke:
> How many functions are we talking about?

I don't know exactly, but inlining every non-templated runtime function usable/useful in `-betterC` (`_d_array_assign_{l,r}` etc.) sounds horrifying.

> What is _d_array_slice_copy supposed to do with -betterC when the slices overlap, lengths disagree, or nullptrs?

I guess it depends on the new `-checkaction` switch, but presumably either calling C assert or emitting an unreachable.

JohanEngelen:
We could try forcing cross module inlining them? (i.e. inlining them from D source)

kinke:
Nope, the regular D version throws Errors, and exceptions aren't supported by `-betterC`. Or do you mean having specific betterC versions of these functions in druntime and inlining them? That might work, but anyway, I don't see this as high priority. Walter chose to tackle/inline that specific use case and added a test for it; as an LDC `-betterC` user, I would go ahead and implement my own `_d_array_slice_copy` in a few lines for the time being.

This stuff (compiler support functions) needs to be tackled with a clear strategy IMO, for bare-metal and `-betterC` friendliness.

kinke:
Wrt. 32-bit Linux dmd-testsuite-failures, running a single one of them manually (via SSH directly on CI test machine) works. According to the logs, it seems like there are conflicts wrt. generated temp filenames for tests running in parallel:
```
Test runnable/a17.d failed: ../../../../ninja-ldc/dmd-testsuite-debug_32/17048cdb74d71bd8683b1617baab89f0b04fa1f3132184ee8ffebe912b90308: No such file or directory
Test runnable/argufilem.d failed: ../../../../ninja-ldc/dmd-testsuite-debug_32/17048cdb74d71bd8683b1617baab89f0b04fa1f3132184ee8ffebe912b90308: No such file or directory
Test runnable/bug16146.d failed: ../../../../ninja-ldc/dmd-testsuite-debug_32/17048cdb74d71bd8683b1617baab89f0b04fa1f3132184ee8ffebe912b90308: No such file or directory
Test runnable/cov2.d failed: ../../../../ninja-ldc/dmd-testsuite-debug_32/17048cdb74d71bd8683b1617baab89f0b04fa1f3132184ee8ffebe912b90308: No such file or directory
...
```

I didn't spot relevant changes in `d_do_test.d` or `std.random` though after a quick glance.

kinke:
Argh I think I found the culprit:
```D
private ulong bootstrapSeed() @nogc nothrow
{
    ulong result = void;
    enum ulong m = 0xc6a4_a793_5bd1_e995UL; // MurmurHash2_64A constant.
    void updateResult(ulong x)
    {
        x *= m;
        x = (x ^ (x >>> 47)) * m;
        result = (result ^ x) * m;
    }
    import core.thread : getpid, Thread;
    import core.time : MonoTime;

    updateResult(cast(ulong) cast(void*) Thread.getThis());
    updateResult(cast(ulong) getpid());
    updateResult(cast(ulong) MonoTime.currTime.ticks);
    //printf(".: bootstrap seed [early]: %llx\n", result);
    result = (result ^ (result >>> 47)) * m;
    result = result ^ (result >>> 47);
    //printf(".: bootstrap seed: %llx\n", result);
    return result;
}
```

With `-O`, the result of this is `undef` in IR, due to `result` not being initialized (`= void`)...

JohanEngelen:
Nice find. Small note: the optimizer plays no role. Using an uninitialized value is UB in D. https://dlang.org/spec/declaration.html#void_init

kinke:
Finally green.

n8sh:
Upstreamed the std.random fix.
https://github.com/dlang/phobos/pull/6833

JohanEngelen:
The `core.thread` issue with `-O` in Windows is probably because those tests are wrong. 
```
mixin("asm pure nothrow @nogc { mov "~REG~", 0xFFFFFFFFFFFFFFFF; }");
zeroRegister.call();
mixin("asm pure nothrow @nogc { mov after, "~REG~"; }");
```
Allthough callee `zeroRegister.call()` shouldn't change the register tested (before==after the call), but the caller itself (`testNonvolatileRegister`) is allowed to change the registers in between those 3 lines of code: setting up the call to `zeroRegister.call()` may involve changing those registers.

kinke:
I extracted the test in a separate module and looked at the optimized IR (not the asm) - it was fine, register being set, then the `zeroRegister.call()` call, then register being read again. The inline asm constraints/clobbers were okay too. And it also worked as expected.
So I guess it's due to some additional inlining going on if directly in `core.thread`, where the `Fiber` implementation and the support functions reside. What's strange is that that module wasn't touched at all.

gen/passes/SimplifyDRuntimeCalls.cpp
kinke: [`Sz` is needed for the alias check in the lines inbetween.]

PR #2951, opened 2018-12-22
Support external wasm-ld linker for WebAssembly targets

kinke:
Overridable with `-linker`. See https://forum.dlang.org/post/qruindgwgbtmuccfujnp@forum.dlang.org.

### DIFF
--- a/tests/baremetal/lit.local.cfg
+++ b/tests/baremetal/lit.local.cfg
@@ -1,4 +1,16 @@
+import subprocess
+
 # Define `%baremetal_args` as LDC args making sure
 # * no ldc2.conf file is used (=> no implicit command-line args), and
 # * the empty object.d in ./inputs is imported (=> no TypeInfo, ModuleInfo, Object...).
 config.substitutions.append( ('%baremetal_args', '-conf= -I' + config.test_source_root + '/baremetal/inputs') )
+
+# Add "link_WebAssembly" feature if we can link wasm (-link-internally or wasm-ld in PATH).
+if config.ldc_with_lld:
+    config.available_features.add('link_WebAssembly')
+else:
+    try:
+        if (subprocess.call(["wasm-ld", "--version"]) == 0):
+            config.available_features.add('link_WebAssembly')
+    except OSError:
+        pass
--- a/tests/baremetal/wasm.d
+++ b/tests/baremetal/wasm.d
@@ -1,7 +1,8 @@
 // Compile and link directly to WebAssembly.
 
 // REQUIRES: target_WebAssembly
-// RUN: %ldc -mtriple=wasm32-unknown-unknown-wasm %s %baremetal_args
+// REQUIRES: internal_lld
+// RUN: %ldc -mtriple=wasm32-unknown-unknown-wasm -link-internally %s %baremetal_args
 
 extern(C): // no mangling, no arguments order reversal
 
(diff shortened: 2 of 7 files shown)


kinke:
@JohanEngelen: `link_WebAssembly` lit feature added.

tests/baremetal/wasm.d
```diff
@@ -1,7 +1,8 @@
 // Compile and link directly to WebAssembly.
 
 // REQUIRES: target_WebAssembly
-// RUN: %ldc -mtriple=wasm32-unknown-unknown-wasm %s %baremetal_args
+// REQUIRES: internal_lld
+// RUN: %ldc -mtriple=wasm32-unknown-unknown-wasm -link-internally %s %baremetal_args
```
kinke: This is a bit unfortunate - due to `%baremetal_args` disabling the config file, `-link-internally` isn't used by default, so it needs to be explicit, otherwise an external `wasm-ld` is required.

tests/baremetal/lit.local.cfg
```diff
@@ -1,4 +1,16 @@
+import subprocess
+
 # Define `%baremetal_args` as LDC args making sure
 # * no ldc2.conf file is used (=> no implicit command-line args), and
 # * the empty object.d in ./inputs is imported (=> no TypeInfo, ModuleInfo, Object...).
 config.substitutions.append( ('%baremetal_args', '-conf= -I' + config.test_source_root + '/baremetal/inputs') )
+
+# Add "link_WebAssembly" feature if we can link wasm (-link-internally or wasm-ld in PATH).
+if config.ldc_with_lld:
+    config.available_features.add('link_WebAssembly')
```
JohanEngelen: Are you sure about this? Isn't it more straightforward if `link_WebAssembly` means we can externally link with Wasm?

kinke: [It's not entirely sufficient, as the LLD version must be 6+ IIRC.] I don't have a strong opinion about it, but these 2 conditions allow you to link wasm in the tests, as long as not using `-conf=` (e.g., just `-betterC`, which is what >99% of users will probably do).
The 3 main CI systems (using our LLVM/LLD fork) now have a `wasm-ld` in their PATH, so it doesn't really matter for CI either - but the PATH adaptations aren't required anymore with the last commit.

JohanEngelen: If I build LDC with internal LLD, but do not have wasm-ld installed, will the test with `link_WebAssembly` try the internal LLD? Or would LDC with internal LLD automatically use the internal linker?

kinke: > If I build LDC with internal LLD, but do not have wasm-ld installed, will the test with link_WebAssembly try the internal LLD?

Yes, and succeed, that's the point.

> Or would LDC with internal LLD automatically use the internal linker?

It does automatically, as the wasm section in ldc2.conf specifies `-link-internally` when building with internal LLD. But the other `wasm.d` test is a minimalistic bare-metal test, disabling the conf file, that's why I have to specify `-link-internally` manually (the other option: making that test depend on an external wasm-ld, as it did after the first commit).

JohanEngelen: OK

kinke:
Good to go?

JohanEngelen:
lgtm

PR #2952, opened 2018-12-22
Add internal_lld Lit-test feature.

JohanEngelen:
This prevents test failures when LDC is built without internal LLD.

### DIFF
--- a/tests/baremetal/wasm.d
+++ b/tests/baremetal/wasm.d
@@ -1,7 +1,6 @@
 // Compile and link directly to WebAssembly.
 
 // REQUIRES: target_WebAssembly
-// REQUIRES: internal_lld
 
 // RUN: %ldc -mtriple=wasm32-unknown-unknown-wasm -link-internally %s %baremetal_args
 
--- a/tests/baremetal/wasm.d
+++ b/tests/baremetal/wasm.d
@@ -1,7 +1,6 @@
 // Compile and link directly to WebAssembly.
 
 // REQUIRES: target_WebAssembly
-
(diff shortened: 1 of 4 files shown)


kinke:
https://github.com/ldc-developers/ldc/pull/2951 makes these tests work without LLD integration, as long as `wasm-ld` is installed or an appropriate `-linker=...` is set in ldc2.conf.

kinke:
> The problem is that testing needs to pass with LLVM Wasm backend but without all the other required pieces

That depends on the desired semantics; as stated above, I'd prefer 'does generating wasm work on the test system if the wasm LLVM backend is available?' over 'does generating wasm work if the test system fulfills all requirements'. So that a user knows that he can't generate .wasm as long as he doesn't have a linker, and doesn't assume it's working because it works with an official package and his test run is green. Well, thinking about it, we have quite a bunch of tests which are only run if the test system matches certain requirements, so you're probably right with `link_Webassembly` being required as proper solution.

> In any case, if -link-internally is used, we need a feature to say that such test requires that. (so that's independent of wasm)

Yep, that's fine for the single Windows-only test after #2951.

JohanEngelen:
Can you please add a `link_WebAssembly` feature in another PR? Lit tests have been broken on my machine for quite a while because of this. It's annoying. I don't have the Wasm stuff installed, and tbh wouldn't even know what all required pieces are.
(a good reason for having `target_WebAssembly` without linking capability is the same as what I mentioned in another recent PR: it is good to have as many tests enabled as possible on all platforms, especially developer platforms. That way, you can see tests fail during development. Specifically for wasm, having `target_WebAssembly` enables testing on IR level without always having to link.)

PR #2955, opened 2018-12-23
Enable LLVM 7.0.1 test on Travis

JohanEngelen:


JohanEngelen:
@kinke The LLVM 7 segfaults, were they spurious or always happening?

kinke:
IIRC, always happening.

.travis.yml
```diff
@@ -31,7 +30,7 @@ matrix:
 
 cache:
   directories:
-    - llvm-7.0.0
+    - llvm-7.0.1
```
kinke: Caching 7.0.0 would still pay off for the OSX job (for which there's apparently no 7.0.1 download yet; IIRC, they don't publish OSX download for minor releases).

JohanEngelen: good catch

JohanEngelen:
> IIRC, always happening.

OK. Build succeeded twice now, so let's merge and see.

PR #2965, opened 2019-01-15
Fix XRay with LLVM 7 and enable testing (fix typo).

JohanEngelen:


JohanEngelen:
XRay needed functionality from C++ stdlib, see: https://bugs.llvm.org/show_bug.cgi?id=32274
This was fixed since LLVM 7.
On Darwin, apparently it is not needed to explicitly link with C++ stdlib (on my machine, Travis, and CircleCI).

kinke:
The lit-tests are fine and take the usual time; the AppVeyor speed is reduced unfortunately so both jobs timed out after 1h. They seem to have multiple clouds with significant performance differences, and the slow one (e.g., lit-tests taking over 60 secs, while the faster ones take ~45 secs and around ~45 min for a full job) seems to be used more frequently lately (cheaper?).

JohanEngelen:
I actually got XRay to work on Linux, but a bunch of additional stuff was needed. Included all in this PR.

driver/linker-gcc.cpp
```diff
@@ -61,6 +61,8 @@ class ArgsBuilder {
   virtual void addCppStdlibLinkFlags(const llvm::Triple &triple);
   virtual void addProfileRuntimeLinkFlags(const llvm::Triple &triple);
   virtual void addXRayLinkFlags(const llvm::Triple &triple);
+  virtual bool addWholeRTLinkFlags(llvm::StringRef baseName,
+                                   const llvm::Triple &triple);
```
kinke: I find the name a bit misleading (whole RT); something like `addWholeCompilerRTArchive` or so might be better. As it currently doesn't enforce the whole archive on Mac, is that as intended (then it should probably be XRay specific and named appropriately) or should it be `-force_load bla.a` (ld64 args) there?

JohanEngelen: I changed the name.
It's what clang does (no whole-archive needed on darwin).

PR #2970, opened 2019-01-19
Travis: Disable Linux LLVM 7.0.1 job (C++ compiler ABI mismatch?)

kinke:
This only *appeared* to work, as long as the official prebuilt 7.0.1 archive wasn't available for download (I guess). The 'successful' [job](https://travis-ci.org/ldc-developers/ldc/jobs/471620842) in #2955 used a preinstalled LLVM 5.0.0 (`/usr/local/clang-5.0.0`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,10 @@ sudo: false
 
 matrix:
   include:
-    - os: linux
-      d: ldc
-      env: LLVM_VERSION=7.0.1 OPTS="-DBUILD_SHARED_LIBS=ON"
+    # FIXME: strange crashes, possibly due to C++ compiler mismatch
+    #- os: linux
+    #  d: ldc
+    #  env: LLVM_VERSION=7.0.1 OPTS="-DBUILD_SHARED_LIBS=ON"
     - os: linux
       d: ldc-beta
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Instead of disabling, shall we mark it as "allowed failure" ?

kinke:
That'd be just a waste of resources IMO, as long as the C++ compiler doesn't change.

JohanEngelen:
What C++ compiler is used to build the LLVM release binaries? AFAICT gcc-4.9 is used for LLVM's nightly builds on Ubuntu Trusty, which is also what we use.

kinke:
We're using their official builds, not the nightlies. I guess it's clang, and clang v7 introduced some ABI incompatibility they already fixed in master IIRC; that's why they have been talking about releasing a 7.1 (7.0.1 + that breaking ABI fix).

JohanEngelen:
... I mentioned the nightlies because it may be the same compiler for their releases. I've asked the provider of the ubuntu binaries about what compiler was used.

bug report: https://bugs.llvm.org/show_bug.cgi?id=39427

JohanEngelen:
Disappointed that you merged this without opening an issue about re-enabling it.

kinke:
It's been disabled for a while until you enabled it. Instead of feelin g disappointed, just open an issue yourself if you find it worth it, as I clearly don't.

PR #2971, opened 2019-01-19
Support templated LLVM intrinsics with vector arguments

### LINKED ISSUE #2962: ICE: intrinsic with vector arguments
Testcase that segfaults:
```d
import core.simd;

pragma(LDC_intrinsic, "llvm.fshl.i#")
    T llvm_fshl(T)(T a, T b, T shift)
        if (__traits(isIntegral, T));

int4 rotateLeftIntrinsic(int4 x, int4 bits) {
    return llvm_fshl(x, x, bits);
}
```
https://d.godbolt.org/z/XUvSX9

Apparently vector types pass the `isIntegral` test. At the least, we shouldn't crash on this.

kinke:
Fixes #2962.

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -1302,32 +1302,38 @@ static char *DtoOverloadedIntrinsicName(TemplateInstance *ti,
   assert(ti->tdtypes.dim == 1);
   Type *T = static_cast<Type *>(ti->tdtypes.data[0]);
 
-  char prefix = T->isreal() ? 'f' : T->isintegral() ? 'i' : 0;
-  if (!prefix) {
+  char prefix;
+  if (T->isfloating() && !T->iscomplex()) {
+    prefix = 'f';
+  } else if (T->isintegral()) {
+    prefix = 'i';
+  } else {
     ti->error("has invalid template parameter for intrinsic: `%s`",
               T->toChars());
     fatal(); // or LLVM asserts
   }
 
-  llvm::Type *dtype(DtoType(T));
-  char tmp[21]; // probably excessive, but covers a uint64_t
-  sprintf(tmp, "%lu",
-          static_cast<unsigned long>(gDataLayout->getTypeSizeInBits(dtype)));
+  std::string name = td->intrinsicName;
 
-  // replace # in name with bitsize
-  std::string name(td->intrinsicName);
+  // replace `{f,i}#` by `{f,i}<bitsize>` (int: `i32`) or
+  // `v<vector length>{f,i}<vector element bitsize>` (float4: `v4f32`)
+  llvm::Type *dtype = DtoType(T);
+  std::string replacement;
+  if (dtype->isPPC_FP128Ty()) { // special case
+    replacement = "ppcf128";
+  } else if (dtype->isVectorTy()) {
+    llvm::raw_string_ostream stream(replacement);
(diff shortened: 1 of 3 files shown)


gen/tollvm.cpp
```diff
@@ -187,17 +187,11 @@ LLType *DtoMemType(Type *t) { return i1ToI8(voidToI8(DtoType(t))); }
 LLPointerType *DtoPtrToType(Type *t) { return DtoMemType(t)->getPointerTo(); }
 
 LLType *voidToI8(LLType *t) {
-  if (t == LLType::getVoidTy(gIR->context())) {
-    return LLType::getInt8Ty(gIR->context());
-  }
-  return t;
+  return t->isVoidTy() ? LLType::getInt8Ty(t->getContext()) : t;
```
kinke: `gIR` wasn't set yet, hence the crash.

gen/llvmhelpers.cpp
```diff
@@ -1309,25 +1309,27 @@ static char *DtoOverloadedIntrinsicName(TemplateInstance *ti,
...
-          static_cast<unsigned long>(gDataLayout->getTypeSizeInBits(dtype)));
-
-  // replace # in name with bitsize
-  std::string name(td->intrinsicName);
+  std::string name = td->intrinsicName;
+
+  // replace "{f,i}#" by "{f,i}<bitsize>" or "v<vector length>{f,i}<vector
+  // element bitsize>"
+  llvm::Type *dtype = DtoType(T);
+  char replacement[24]; // probably excessive, but covers a uint64_t
```
JohanEngelen: I don't understand the comment.

Safer to use `llvm::SmallString` and `llvm::formatv` for the replacement string.

JohanEngelen: besides this, lgtm

kinke: What isn't understandable, the `{f,i}` bash-style set?

kinke: Argh, `llvm::formatv()` not available for 3.9.

JohanEngelen:
The tests needs to do linking aswell, such that it would fail if we'd get the intrinsic name wrong as mentioned here: https://github.com/ldc-developers/druntime/pull/151#issuecomment-455828670

kinke:
[LLVM 7+ errors out if the name doesn't match the signature, that's how I got aware of that missing piece.]

I'll look for another intrinsic and link.

kinke:
Slightly reworked, incl. fix for FP vectors (`TypeVector` doesn't forward `isreal()` to its element type, `isfloating()` does).

gen/llvmhelpers.cpp
```diff
@@ -1302,32 +1302,39 @@ static char *DtoOverloadedIntrinsicName(TemplateInstance *ti,
...
+    llvm::raw_string_ostream stream(replacement);
+    stream << 'v' << dtype->getVectorNumElements() << prefix
+        << gDataLayout->getTypeSizeInBits(dtype->getVectorElementType());
+    stream.flush();
+  } else {
+    replacement = prefix + std::to_string(gDataLayout->getTypeSizeInBits(dtype));
+  }
 
-  std::string needle("#");
+  const std::string needle = "#";
```
Hardcode84: string::find support const char* overload I think

PR #2972, opened 2019-01-19
Experiment with Semaphore CI 2.0

kinke:


kinke:
Errors during git checkout - may need a repo owner to set it up (they require separate v2 accounts...). - Documentation is very scarce, but they now feature a non-dockerized Ubuntu 18.04 image with 8 GB of usable memory (previously: dockerized 18.04 with ~2.4 GB), so that building single-threadedly and splitting the run in 2 jobs shouldn't be required anymore.

kinke:
Okay, a simple `checkout || true` works around the shallow clone 'error'.

The available memory seems to be 4 GB, not 8 GB. The full run (native Ubuntu 18.04 image, using its LLVM 6.0.0, gcc 7 etc., multilib) now succeeds and took about 42 minutes.

Can you guys display the CI log without being logged in?

kinke:
Oh wow, that's a serious problem. Another one is that they are supposed to *prepare* a free service for open-source projects; right now we have a free 20$ credit per month, which translates to about one 42-minutes run per day.

kinke:
A full multilib run with their octacore-machines takes under 20 minutes. I expected a somewhat better scaling; `runnable/xtest46.d` with its many (serialized) permutations seems to be a main culprit, delaying the `runnable` block of the 4 dmd-testsuite runs (as it is usually started pretty late too).

kinke:
This isn't going anywhere; free credit was halved in the meantime, and the only available machine for the free plan has 2 cores.

PR #2974, opened 2019-01-20
Fix C++ mangling of `real` for MSVC targets

kinke:
Cherry-pick from #2953 (Posix mangling fix already in 2.084 upstream => master):

Mangle as native Visual C++ long double. DMD has to special-case, as it's using 80-bit extended precision.


JohanEngelen:
The posix fix is not in master btw, so the amended #2953 is still needed.

kinke:
Of course it is, otherwise the lit test would have failed.

PR #2976, opened 2019-01-20
@weak on definitions should not result in extern_weak linkage on declarations

JohanEngelen:
This is useful for optionally linking with libraries. (if not linked with the function, the function address is set to `null` by the linker).

I don't know whether it works on Windows. On macOS, it's somewhat problematic because it requires telling the linker to accept such undefined references. With ELF, it should work fine.
The difference between platforms make this less useful, but it is not much trouble to support the usecase.

### DIFF
--- a/tests/codegen/attr_weak_external.d
+++ b/tests/codegen/attr_weak_external.d
@@ -1,0 +1,12 @@
+// Test that an imported @weak function does not result in an extern_weak reference.
+
+// RUN: %ldc -c -I%S -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+import inputs.attr_weak_external_input: weak_definition_seven;
+
+void foo()
+{
+    auto a = &weak_definition_seven;
+}
+
+// CHECK-NOT: extern_weak
--- a/tests/codegen/attr_weak_lto.d
+++ b/tests/codegen/attr_weak_lto.d
@@ -1,0 +1,22 @@
+// Test importing a @weak function with LTO
+
+// REQUIRES: LTO
+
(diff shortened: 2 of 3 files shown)


kinke:
I doubt it works for MSVC; AFAIK, there are (undocumented) weak aliases (use symbol as fallback for another symbol name, if that one would otherwise be undefined), but that's it.

kinke:
I guess you don't want to squeeze this into 1.14 beta1, right? Then I'd release it tonight (still need to extend the #2971 tests, as they don't crash with older LDC anymore).

Wrt. to the Win32 LTO failure here, are you sure the declaration linkage is always overridden for definitions?

JohanEngelen:
There is no need to rush this.

kinke:
As it's `_d_newarrayU`, I wouldn't expect there to be regular declarations, just declarations copied from the special runtime module (and thus independent from attribs used in druntime); that's why my guess was that the definition still has the `extern_weak` linkage.

kinke:
It's imported in `rt.aaA` and declared manually in `object` and `std.array` (without LDC-specific `weak` UDA)...

JohanEngelen:
Looking at the results of the extra LTO test, I think we should just not do this on Windows.
It works for ELF and apparently also for Mach-O, so we enable it only for those object formats.

JohanEngelen:
I've completely reworked this. The original idea was wrong. `@weak` on functions should not mean that the symbol gains `extern_weak` linkage because it will mean that importing an `@weak` function from another package would automatically make it an optional function link.
Instead, we already have `pragma(LDC_extern_weak)` and so I just have to make that work for function declarations aswell.
The only thing left in this PR is additional testing of `@weak`.

PR #2986, opened 2019-01-26
ltsmaster: Get rid of __inline_isnan (fixes #2982)

### LINKED ISSUE #2982: Build of ldc 0.17.6 fails on OSX because of missing __inline_isnanl
When building ldc 0.17.6 on Darwin with Nix I get the following error:

[  5%] Building CXX object CMakeFiles/LDCShared.dir/dmd2/root/port.c.o
/tmp/nix-build-ldcBuild-0.17.6.drv-3/ldc-0.17.6-src/dmd2/root/port.c:791:12: error: use of undeclared identifier '__inline_isnanl'
    return __inline_isnanl(r);

port.c had always this problem also when building dmd till it was removed in favor of port.d.
I always replaced __inline_isnanl with __inline_isnan to make it work. Would be nice if that could be fixed upstream.
I would have made a PR but I don't know if that is the proper fix.

ThomasMader:


kinke:
Newer LDC requires a C++11 compiler anway, so that none of this crap should be necessary anymore when compiling ltsmaster now; `std::isnan()` should do. - Anyway, your force-push now gets rid of the BSD special cases, and that's no good.

ThomasMader:
Why is C++11 not required on BSD?

kinke:
I shouldn't have mentioned the C++11 thing, it was just a side-comment fed by C++ frustration. ;) - Wrt. `ltsmaster`, the less changes the better, especially wrt. untested BSD, so using the C++11 function for Mac now too looks good to me.

Issue #2820, opened 2018-08-18
Segmentation fault after last line of main 


Sandman83:
compiled a project with 
dub build --compiler=ldc2 --build=release-debug

Program runs, end ends with a segmentation fault (Segmentation fault: 11) AFTER the last line of my program. 
Tried to debug this with lldb and got the stack, written at the end of this issue. 

Environment: 
System: MacOs 10.13.6
Architecture: 64bit
LLVM: llvm 6.0.1
LDC: LDC - the LLVM D compiler (1.10.0):
  based on DMD v2.080.1 and LLVM 5.0.1
both installed with homebrew. 

I was not able to reproduce this segmentation fault with following environments
Environment: 
System: MacOs 10.13.6
Architecture: 64bit
DMD64 D Compiler v2.081.2

Environment: 
System: Linux parallels-vm 4.15.0-32-generic, <span>#</span>35-Ubuntu
Architecture: x86_64
DMD64 D Compiler v2.081.1

Environment: 
System: Linux parallels-vm 4.15.0-32-generic <span>#</span>35-Ubuntu
Architecture: x86_64
DMD64 D Compiler v2.081.1
LDC - the LLVM D compiler (1.8.0):
  based on DMD v2.078.3 and LLVM 5.0.1

Stack trace: 

121 was compiled with optimization - stepping may behave oddly; variables may not be available.
Process 52870 stopped
* thread <span>#</span>2, stop reason = EXC_BAD_ACCESS (code=1, address=0x18)
[...]

PR #2998, opened 2019-02-07
CI: Set up Azure Pipelines, drop AppVeyor, rededicate Circle

kinke:


kinke:
@ZombineDev: For Windows, the performance seems to very roughly match AppVeyor - 2 (logical?) cores, 3.5 GB RAM. But the max configurable time-out seems to be 6h, which should definitely eliminate all time-out issues. :)

kinke:
Overall, I'm quite happy with it so far. It's not perfect, e.g., there's no ssh support, and a step's `displayName` comes after its script (which can be quite long). I also like the GUI. For Windows, all that's still needed is adding a multilib job (yes, one can define a job with dependencies on other jobs - so that it's run after both x64 and x86 jobs have succeeded, in order to create a merged multilib package) and uploading the packages to GitHub (incl. how to figure out if the build is a PR/a commit on the master branch/a tagged commit). Then we don't need AppVeyor anymore. Our LLVM fork can/should be migrated too, as AppVeyor has lately always been too slow to build the LLVM packages in 1h (so that I had to build the 4 packages manually, getting on my nerves).

We could also replace CircleCI (Linux & macOS) with this and use the Azure Pipelines for all x86/x64 packages (and Shippable for the AArch64 one). We'd only use 5 jobs per commit then (Windows x86 & x64 & tiny multilib, Linux multilib, macOS x64), so that 2 commits can be tested in parallel (10 free concurrent jobs).

kinke:
Current status:

* Package upload to GitHub is still to be tested, as is the package generation for tagged builds (LLVM with disabled assertions, package filename, package root dir name).
* Windows: should be fine, incl. additional tests compared to AppVeyor (dynamic compile integration test, multilib smoke tests). Untagged builds use LLVM with enabled assertions again.
* macOS:
  * v10.13 (Circle: 10.12) and default Xcode 9.4.1 (multiple Xcode versions preinstalled, incl. v10; Circle: v9.2). The issue worked around in #2839 for Travis thus strikes again, `druntime-test-shared` fails. @JohanEngelen, @dnadlinger: someone of you Mac guys please consider debugging, as I guess your system would match that setup. Maybe related to https://forum.dlang.org/thread/q0l0t6$2mnn$1@digitalmars.com and/or #2820.
  * **x64 only**, no multilib package with fat libs anymore.
  * FIXME: lit isn't found.
  * Should we raise `MACOSX_DEPLOYMENT_TARGET` from 10.8 to 10.9 while at it (incl. our LLVM fork), to get rid of `clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of OS X 10.9 [-Wdeprecated]`?
* Linux:
  * Ubuntu 16.04 (Circle: 14.04) and default gcc 5.4.0 (Circle: backported 6.5.0). The LLVM prebuilt by Circle doesn't work for linking, apparently due to C++ ABI mismatch. I'll thus have to switch our LLVM fork to Azure first; probably with LLVM 8.0.

kinke:
I rededicated CircleCI: no self-compilation & packaging anymore, also combining bits from Semaphore and Travis.

* Linux
  * Ubuntu 18.04: using distro gcc (7.3) + LLVM (6.0) + compiler-rt libs, multilib, druntime sanitizers support
  * Ubuntu rolling (latest, currently 18.10): using distro gcc (8.2) + LLVM (7.0) + compiler-rt libs, multilib, shared druntime/Phobos libs only (=> running lit-tests & DMD testsuite with shared libs, previously only tested by Travis)
* macOS: using our prebuilt LLVM 7.0.1 with enabled assertions, x64 only
  * static+shared libs
  * shared libs only

kinke:
Alright, looks like we're pretty much set for LLVM 8 (from yesterday), incl. Khronos SPIRV-LLVM-Translator as git submodule. It just needed some final adaptations, but apparently also introduced a rather strange `std.math.cmp()` regression on Win32 without optimizations for 64-bit NaNs with custom payload...

JohanEngelen:
nice @kinke

kinke:
Appreciated. Whoever has an MS account (AFAICT, mandatory) is free to join the Azure ldc-developers organization, we'll figure out how. ;) - It's not absolutely crucial, as anyone can create their own Azure project and automatically use the YAML files, just in case I wouldn't be around anymore.

PR #3005, opened 2019-02-25
Upgrade to LLVM 8.0

kinke:


kinke:
The final (originally scheduled for February 27th) *may* take another week, but there might be another rc5 as well - so I'm going to release v1.15-beta1 with rc4.

PR #3013, opened 2019-03-06
Fix FPelim from LLVM8

thewilsonator:
It was apparently removed in LLVM8 not LLVM9

see https://github.com/llvm-mirror/llvm/blob/release_80/include/llvm/CodeGen/CommandFlags.inc


kinke:
Already fixed, among other stuff, in #3005.

Issue #2552, opened 2018-02-01
Don't ignore pragma(inline, true) without explicit -enable-cross-module-inlining


kinke:
I almost forgot about this regression (disabled in #1737) but have just come across it again. It's particularly bad for Phobos callees, e.g., tiny std.math or LLVM intrinsic wrappers which have been annotated with this pragma for a reason.

kinke:
Restored in #3014.

PR #3014, opened 2019-03-06
Re-enable cross-module-inlining for `pragma(inline, true)`

kinke:
At -O0 already, tackling #2552.

`pragma(inline, true);` *statements* (now?) seem to require semantic analysis, so I adapted LDC-specifics in druntime/Phobos to decorate the function declaration directly - `tests/codegen/inlining_stdlib.d` failed without adaptation (to inline non-templated `core.bitop.bsf()`).


kinke:
I'll include this in v1.15-beta1 - let's see if there are any remaining issues for the few `pragma(inline, true)` functions.

PR #3017, opened 2019-03-10
Tentative Azure CI fixes

kinke:


kinke:
The `isPreRelease` flag works (tested in LLVM repo), whether tags now trigger will have to be seen.

PR #3022, opened 2019-03-12
dmd-testsuite: Parallelize building of tools

kinke:


kinke:
This saves very roughly 40 seconds (about -8% of full dmd-testsuite run) for Shippable, as intended/hoped for.

thewilsonator:
Nice.

Issue #3028, opened 2019-03-13
Crashing exe when passing big vectors as function parameters (win64)


Hardcode84:
`LDC - the LLVM D compiler (1.15.0-beta1):`
`ldc2 main.d -O0 -g`
```
import core.simd;

alias Vec = Vector!(float[32]);
//alias Vec = Vector!(float[4]); works

auto foo(V)(V v1, V v2)
{
    return v1 + v2;
}

void main()
{
    Vec v1;
    Vec v2;
    Vec v3 = foo(v1, v2);
    import std.stdio;
    writeln(v3);
}
```

Disasm:
```
main!main.foo!(__vector(float[32])):
00007ff6`351c1700 55              push    rbp
00007ff6`351c1701 4881ecf0010000  sub     rsp,1F0h
00007ff6`351c1708 488dac2480000000 lea     rbp,[rsp+80h]
00007ff6`351c1710 440f29bd60010000 movaps  xmmword ptr [rbp+160h],xmm15
00007ff6`351c1718 440f29b550010000 movaps  xmmword ptr [rbp+150h],xmm14
00007ff6`351c1720 440f29ad40010000 movaps  xmmword ptr [rbp+140h],xmm13
00007ff6`351c1728 440f29a530010000 movaps  xmmword ptr [rbp+130h],xmm12
00007ff6`351c1730 440f299d20010000 movaps  xmmword ptr [rbp+120h],xmm11
00007ff6`351c1738 440f299510010000 movaps  xmmword ptr [rbp+110h],xmm10
00007ff6`351c1740 440f298d00010000 movaps  xmmword ptr [rbp+100h],xmm9
00007ff6`351c1748 440f2985f0000000 movaps  xmmword ptr [rbp+0F0h],xmm8
00007ff6`351c1750 0f29bde0000000  movaps  xmmword ptr [rbp+0E0h],xmm7
00007ff6`351c1757 0f29b5d0000000  movaps  xmmword ptr [rbp+0D0h],xmm6
00007ff6`351c175e 4883e480        and     rsp,0FFFFFFFFFFFFFF80h
00007ff6`351c1762 488b8538020000  mov     rax,qword ptr [rbp+238h] <-- load some garbage in rax
00007ff6`351c1769 0f2800          movaps  xmm0,xmmword ptr [rax] ds:00000000`00000000=???????????????????????????????? <-- try use it as pointer
00007ff6`351c176c 488b8528020000  mov     rax,qword ptr [rbp+228h]
[...]

PR #3029, opened 2019-03-16
passing big value cause issues on win64 (#3028)

Hardcode84:
Ignore non power of 2 sizes for now, they are causing asserts inside compiler in debug builds.


kinke:
I expected something way simpler, directly in `passPointeroHiddenCopy()`, just checking if it's a vector types and larger than, say, 512 bits. I.e., no special case for the 1st arg, no dependencies of later args on their predecessors (which just seem like pretty arbitrary restrictions imposed by the current LLVM implementation), and especially no difference wrt. passing and returning such a vector type.

Hardcode84:
Well, what else we can do? First parameter is special for some reason it can handle any size and I actually getting broken module if I try to use `byvalRewrite` on it. And you can hit 64 byte limit for vector args pretty easy by code like `foo(someParam, Vector!(float[8]),Vector!(float[8]),Vector!(float[8]))`.

Anyway, I locally workarounded it by manually passing vectors by `ref`, so this fix not critical for me.

kinke:
> Well, what else we can do?

What I suggested? ;)

> First parameter is special for some reason, it can handle any size

Must be an LLVM issue. It makes absolutely no sense to me.

> And you can hit 64 byte limit for vector args

Definitely smells like another LLVM issue, á la 'hey, Win64 ABI only uses 4 regs, and hey, we only assume max 16-bytes vectors per register, so crap out after 4x16=64 bytes'.

> I locally workarounded it by manually passing vectors by ref

Yes, I would have assumed that copying 1 KB of data for a single `float64` passed by value doesn't exactly help performance.

kinke:
[Btw, we only use the `__vectorcall` ABI for `extern(D)` and the regular Win64 ABI for `extern(C++)` etc., where all vectors are inefficiently passed as pointers to hidden copies.]

Issue #1751, opened 2016-09-10
Revive code coverage tracking


klickverbot:
I've just removed the Coveralls badge from our README, since it hasn't been updated in months and was showing a wrong total.

We should get something up and running again, either based on Coveralls or an alternative like Codecov.io.

PR #3031, opened 2019-03-16
Travis: Don't build with coverage instrumentation

kinke:
That job is frequently timing out. The implicit `-O0` for LDC's C++ parts (excl. LLVM) certainly don't help, and the coverage data hasn't been used for years (issue #1751).


kinke:
Once the king of open-source CI, nowadays degrading to a liability. Sad.

PR #3034, opened 2019-03-19
Add lib files when linking for a static library

Temtaime:
LDC currently silently ignores all passed lib files when linking for a static library, but dmd doesn't.
Current proposal makes LDC match DMD's behavior and link the libraries together.

### DIFF
--- a/driver/archiver.cpp
+++ b/driver/archiver.cpp
@@ -325,7 +325,7 @@ int createStaticLibrary() {
   for (auto objfile : global.params.objfiles) {
     args.push_back(objfile);
   }
-  
+
   // user libs
   for (auto libfile : global.params.libfiles) {
     args.push_back(libfile);
--- a/driver/archiver.cpp
+++ b/driver/archiver.cpp
@@ -325,6 +325,11 @@ int createStaticLibrary() {
   for (auto objfile : global.params.objfiles) {
     args.push_back(objfile);
   }
+  
+  // user libs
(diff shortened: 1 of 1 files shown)


Temtaime:
test13774.sh is failing because it uses internal linker, but llvm-lib currently doesn't seem to support linking multiple lib files together.
Before this patch the test incorrectly passed because it was silently ignoring lib files and didn't pass them to the linker.
Is there a way to disable llvm-lib for this test only and use msvc's link ? Or maybe disable the test itself (not a good option) ?

kinke:
Yeah, these few lines have shown that the test is inadequate and that even LLVM 8.0's `llvm-lib` still doesn't allow static libs merging (no linking involved here).
As you may have seen, we always default to LLVM's archiver, for various reasons (cross-archiving, LTO/bitcode support, independency from system archiver...); the archiver can be set to an external one via `-archiver=lib.exe`. We'd have to use that one for Windows targets only (`OS` env variable starting with `win`, checkable in the bash script).

kinke:
Hmm, looks like Posix `ar` doesn't support simple merging (https://stackoverflow.com/questions/3821916/how-to-merge-two-ar-static-libraries-into-one).

Temtaime:
Thanks very much for taking the time for this!
Kind regards.

kinke:
You're welcome, thank *you* for spotting and fixing the issue.

PR #3037, opened 2019-03-22
Don't emit object.RTInfo(Impl) instantiations in dcompute modules

kinke:
This is not a perfect solution, as other modules might be relying on an emission (template culling...), but should work in most cases (and otherwise probably yields a linker error).

It'd have been a shame to ship with a totally broken dcompute for the first official LDC packages able to emit OpenCL (v1.15), so I reluctantly took matters in my own hands.


kinke:
@rainers: Is the second global [here](https://github.com/ldc-developers/druntime/blob/3a42b95d449ed2704c8a1d8027c9388fe3e15c8a/src/object.d#L4220) really required?

PR #3042, opened 2019-03-29
Propagate well-known length of newly allocated arrays

### LINKED ISSUE #3041: Suboptimal code: Array literal length doesnt propagated as constant (I suppose?)
```
alias F = void function();

void bar(F[] funcs)
{
    foreach (f; funcs)
    {
        if(f is null)
        {
            return;
        }
        f();
    }
}

void baz1(F a)
{
    auto f = [a,null,a,a];
    bar(f);
}

void baz2(F a)
{
    auto f = [a,null,a,a];
    f = f[0..4];
    bar(f);
}
```

```
void example.baz1(void function()*):
        push    rbx
        sub     rsp, 32
        mov     qword ptr [rsp], rdi
        mov     qword ptr [rsp + 8], 0
        mov     qword ptr [rsp + 16], rdi
        mov     qword ptr [rsp + 24], rdi
        test    rdi, rdi
        je      .LBB1_4
        mov     ebx, 1
.LBB1_2:
        call    rdi
        cmp     rbx, 3
        ja      .LBB1_4
        mov     rdi, qword ptr [rsp + 8*rbx]
        add     rbx, 1
        test    rdi, rdi
        jne     .LBB1_2
.LBB1_4:
        add     rsp, 32
        pop     rbx
        ret

void example.baz2(void function()*):
        test    rdi, rdi
        je      .LBB2_1
        jmp     rdi
.LBB2_1:
        ret
```
`-O3 -release -boundscheck=off`

https://godbolt.org/z/MNI5o3

kinke:
For better optimizability, as the length information is lost when using the slice returned by the druntime hooks directly. Resolves #3041.

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -682,10 +682,13 @@ DSliceValue *DtoNewDynArray(Loc &loc, Type *arrayType, DValue *dim,
   IF_LOG Logger::println("DtoNewDynArray : %s", arrayType->toChars());
   LOG_SCOPE;
 
-  // get runtime function
   Type *eltType = arrayType->toBasetype()->nextOf();
-  bool zeroInit = eltType->isZeroInit();
 
+  if (eltType->size() == 0)
+    return DtoNullValue(arrayType, loc)->isSlice();
+
+  // get runtime function
+  bool zeroInit = eltType->isZeroInit();
   const char *fnname = defaultInit
                            ? (zeroInit ? "_d_newarrayT" : "_d_newarrayiT")
                            : "_d_newarrayU";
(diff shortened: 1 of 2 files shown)


gen/arrays.cpp
```diff
@@ -682,10 +682,13 @@ DSliceValue *DtoNewDynArray(Loc &loc, Type *arrayType, DValue *dim,
   IF_LOG Logger::println("DtoNewDynArray : %s", arrayType->toChars());
   LOG_SCOPE;
 
-  // get runtime function
   Type *eltType = arrayType->toBasetype()->nextOf();
-  bool zeroInit = eltType->isZeroInit();
 
+  if (eltType->size() == 0)
+    return DtoNullValue(arrayType, loc)->isSlice();
```
kinke: That's the only special case where the length returned by the druntime hooks does not match the input length.

kinke:
Pinging @Hardcode84; your example is optimized as expected with this.

kinke:
[There would be more opportunities like this (`_d_arraysetlength*` etc.), but their impact is probably smaller, and these hooks may be replaced by templates pretty soon...]

PR #3045, opened 2019-04-02
Only default to SSSE3 (with -O) for LLVM 7.x

kinke:
Looks like that regression has been fixed for LLVM 8.


kalev:
Nice, thanks! Linux distros that are supposed to support older HW are going to love this.

PR #3046, opened 2019-04-02
druntime: Don't exclude core.stdcpp.* any longer

kinke:


kinke:
That was a pure experiment. No idea how something like [that](https://github.com/ldc-developers/druntime/blob/072e7ac6854fd000164b78030dd8ddd7d5573430/src/core/stdcpp/typeinfo.d#L111-L112) is supposed to work, emitting the vtbl on the D side cannot work.

PR #3051, opened 2019-04-05
Jit: better error handling and handle correctly __chkstk function

Hardcode84:
Compiler need to build a list of functions jit can call back to static code. This is done via analyzing jitted IR but some special functions require special handling as they are not present in IR but reference to them can still be generated.


kinke:
[Please squash the 2 last commits before merging.]

Hardcode84:
I will probably just use 'squash and merge' button

Issue #3050, opened 2019-04-05
[mips][vibe.d] Interface call memory corruption


rracariu:
When calling the following https://github.com/vibe-d/vibe.d/blob/03b85894ff8e88f3f92001c0750f809aa3ccb1f8/http/vibe/http/session.d#L93

The `set` function will get corrupted data for the parameters and eventually segfault.
If a cast to the concrete type is added like `(cast(MemorySessionStore) m_store).set(m_id, key, serialize(value));` the call will succeed.

Casting `m_store` to `void*` will get you `$1 = (ubyte *) 0x77ab0618`
But the `this` for the concrete type is `0x77ab0610`, observe :
```#0  vibe.http.session.MemorySessionStore.set() (this=**0x77ab0610**, id=..., name=..., value=...) at session.d:244
#1  0x0077a788 in vibe.http.session.Session() (this=0x77ab0618, key=..., value=...) at session.d:94
#2  0x0068ca84 in vibe.http.server.HTTPServerResponse.startSession() (this=0xb11670, path=..., options=vibe::http::server::httpOnly) at server.d:1591
#3  0x005bf8a0 in httpTest.test() (req=0xb110b0, res=0xb11670) at /httpTest.d:51
#4  0x00610d0c in std.functional() (this=<optimized out>, a0=0xb110b0, a1=0xb11670)
    at /dlang/ldc-1.15.0-beta2/bin/../import/std/functional.d-mixin-1430:1441
#5  0x00611814 in vibe.http.router.URLRouter() (req=0xb110b0, res=0xb11670) at .dub/packages/vibe-d-0.8.5/vibe-d/http/vibe/http/router.d:274
#6  0x006785bc in vibe.http.router.URLRouter.handleRequest() (ridx=5, values=...) at router.d:218
#7  0x0067c1d8 in vibe.http.router() (this=0x18, text=..., del=...) at router.d:674
#8  0x00678260 in vibe.http.router() (this=0x18, text=..., del=...) at router.d:607
#9  0x00677ed8 in vibe.http.router.URLRouter.handleRequest() (this=0x77ab2b00, req=0xb110b0, res=0xb11670) at router.d:211
#10 0x00678b20 in _DThn8_4vibe4http6router9URLRouter13handleRequestMFNfCQBvQBt6server17HTTPServerRequestCQDcQDaQBh18HTTPServerResponseZv () at router.d:194
[...]

PR #3061, opened 2019-04-19
Fix MIPS ABI for large aggregates when passed by value

rracariu:
https://github.com/ldc-developers/ldc/issues/3050


kinke:
Thx; looks okay, although it could be made simpler. The usage of `IntegerRewrite` should be purely optional; it's used for Win64 because it's needed to match the MS specs (pass struct of power-of-2-size <= 64 bits in GP register). The magic C++ structs aren't used anymore IIRC (replaced by magic enums).

You should IMO add a clarifying comment somewhere, referencing the issue and explaining why `byval` isn't used (and that the `IndirectByvalRewrite` is just a hack to make the interface calls work, and that the ABI is nowhere near C-compatible).

Wrt. test, adding an explicit one certainly wouldn't hurt. It might be already covered by existing tests (if so, probably in dmd-testsuite), but I guess you don't know either. ;)

rracariu:
OK, thanks!

I simplified the code as suggested and added the comment.
As for the test, this one looks to be doing it ldc\tests\codegen\indirect_byval_rewrite.d.

kinke:
That test only makes sure `IndirectByvalRewrite` itself isn't too dumb/inefficient (in-place construction etc.). We need a test specifically for #3050, making sure that interface calls work with non-tiny aggregates.

---

Some aggregates are now not rewritten (those rewritable to integers), which is probably less than ideal (e.g., a struct with 8 bools is probably passed as 8 separate bools). I'll rework the logic sometime; it'd be nice if you could try to reduce the #3050 testcase to the bare minimum and verify that it still reproducibly fails on MIPS.

rracariu:
> That test only makes sure `IndirectByvalRewrite` itself isn't too dumb/inefficient (in-place construction etc.). We need a test specifically for #3050, making sure that interface calls work with non-tiny aggregates.
> 
OK, I'll try to simplify the code.

> Some aggregates are now not rewritten (those rewritable to integers), which is probably less than ideal (e.g., a struct with 8 bools is probably passed as 8 separate bools). I'll rework the logic etc. sometime; it'd be nice if you could try to reduce the #3050 testcase to the bare minimum and verify that it still reproducibly fails on MIPS (before this PR obviously).

If I remove that test stuff will break, I think it might be related to non-power-of-2 aggregates.

kinke:
> If I remove that test stuff will break, I think it might be related to non-power-of-2 aggregates.

So just using `bool passPointerToHiddenCopy(Type *t, bool) const { return isAggregate(t); }` does not work? If so, can you check whether a somewhat relaxed `TY ty = t->toBasetype()->ty; return ty == Tstruct || ty == Tsarray;` does?

PR #3062, opened 2019-04-22
Upgrade front-end & libs to v2.086.0

kinke:


kinke:
lit-test `codegen/union.d` fails due to a frontend regression wrt. union initializers: https://issues.dlang.org/show_bug.cgi?id=19822

kinke:
Looking pretty good. Remaining issues:

* union initializer regression: https://issues.dlang.org/show_bug.cgi?id=19822
* runnable/test19731.d failure: https://github.com/dlang/dmd/pull/9467#pullrequestreview-231422371
* ltsmaster doesn't work as host compiler anymore: `dmd/init.d(100): Error: dmd.ast_node.Visitor at dmd/ast_node.d(15) conflicts with dmd.visitor.Visitor at dmd/visitor.d(28)` etc.

John-Colvin:
Is there an ETA on this? No hurry, but ideally I would like to be able to plan around a rough time when a beta will be available.

kinke:
The union regression was fixed upstream.
The aggregate invariants + `auto` function issue needs analysis (just LDC affected by horrible AST inconsistency or DMD too?) but is no blocker IMO (so I might just revert the relevant DMD commit for the time being).

kinke:
The `runnable/test19731.d` issue seems to be the following: dlang/dmd#9467 may replace an already sema3'd member function by a 'clone' (and redoing semantic, so that the new function contains the invariant calls). Cloning is performed via `FuncDeclaration.syntaxCopy()`, which leads to a syntaxCopy of the function body, and so to a syntaxCopy of a `ThisExp` for the testcase. There's no `syntaxCopy` override in `ThisExp`, so the `VarDeclaration ThisExp.var` member is NOT adjusted, and will point to a `VarDeclaration` in the old, replaced function. There seem to be more Expressions with similar `syntaxCopy()` behavior, i.e., not expecting to be syntaxCopied into another function.

Issue #3023, opened 2019-03-12
Every export in WebAssembly is removed in ldc 1.15.0-beta1


skoppe:
In ldc 1.14 and before, any function marked with `extern(C) export` was exported in the wasm binary.

With ldc 1.15-beta1 pretty much everything is stripped. The only way I can export anything is to pass `--export=funcname` to the linker.

Ironically in ldc 1.14 the hard part was to remove exports (since it was exporting way too much, including *init_Z), but in ldc 1.15-beta1 all that stuff is removed (yay!), but also the things I want to export.

skoppe:
Here is the relevant section from lld

> When building an executable, only the entry point and symbols flagged as WASM_SYMBOL_EXPORTED are exported by default. In LLVM the WASM_SYMBOL_EXPORTED flag is applied to any symbol in the llvm.used list which corresponds to __attribute__((used)) in C/C++ sources.
> 
> In addition, symbols can be exported via the linker command line using --export.
> 
> Finally, just like with native ELF linker the --export-dynamic flag can be used to export symbol in the executable which are marked as visibility=default.

If I add `--export-dynamic`to the linker I get the same behaviour as with ldc 1.14

But, is there a way I can add the `__attribute__((used))` to my function?

kinke:
Sure: https://github.com/ldc-developers/druntime/blob/3a42b95d449ed2704c8a1d8027c9388fe3e15c8a/src/ldc/attributes.d#L73

skoppe:
Nice. I see in the llir that it is correct:

`@llvm.used = appending global [1 x i8*] [i8* bitcast (i8* (i32)* @allocString to i8*)], section "llvm.metadata" ; [#uses = 0]`

except it still doesn't get exported.

I am not sure what version of llvm ldc uses. It could be that the change I am waiting for (https://reviews.llvm.org/D57869) is still in the pipeline.

kinke:
> I am not sure what version of llvm ldc uses.

8.0.0-rc4 as mentioned in the release notes. That change is very fresh, so very likely not to be part of LLVM 8 (which was branched off trunk mid-January).

skoppe:
Ok. I am going to use `--export-dynamic` until that change is released. Thanks for the quick replies.

PR #3072, opened 2019-05-11
WebAssembly: Use --export-dynamic when linking with LLD 8+

kinke:
This fixes issue #3023 and restores the behavior of previous LLD versions (of exporting/not stripping all symbols with default/public visibility).

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -260,7 +260,11 @@ elseif(${BUILD_SHARED_LIBS} STREQUAL "OFF")
 endif()
 
 if(LDC_WITH_LLD)
-    set(WASM_DEFAULT_LDC_SWITCHES "\n        \"-link-internally\",")
+    set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-link-internally\",")
+endif()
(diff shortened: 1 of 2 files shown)


runtime/CMakeLists.txt
```diff
@@ -260,7 +260,11 @@ elseif(${BUILD_SHARED_LIBS} STREQUAL "OFF")
 endif()
 
 if(LDC_WITH_LLD)
-    set(WASM_DEFAULT_LDC_SWITCHES "\n        \"-link-internally\",")
+    set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-link-internally\",")
+endif()
+# LLD 8+ requires (new) `--export-dynamic` for WebAssembly (https://github.com/ldc-developers/ldc/issues/3023).
+if(NOT (LDC_LLVM_VER LESS 800))
+    set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-L--export-dynamic\",")
```
kinke: Without `-link-internally`, we assume the external wasm linker to support `--export-dynamic` if compiling against LLVM 8+. This should work fine for distros by default, as their system-wide `wasm-ld` will match that version.

The user still has the option to remove it from `ldc2.conf` (LLD < 8 does not support `--export-dynamic`).

PR #3076, opened 2019-05-19
Merge upstream stable (0b606cad20)

kinke:


kinke:
`syntaxCopy` causing troubles again (`ThisExp` already needed an override to pass the 2.086 testsuite). While DMD is now using the LDC contract invocation scheme (https://github.com/dlang/dmd/pull/9782, passing the params explicitly), the slight adaptations thereof (convert all non-lazy params to `ref` ones, not just `out` ones, and some additional syntaxCopies) cause trouble.

Before the hack (last commit), dmd-testsuite's `runnable/{ldc_,test}contracts.d` would fail.

**ldc_contracts** failed to compile (`extern(C)` linkage lost somewhere - `TypeFunction.syntaxCopy()` looks okay though):
```
function `ldc_contracts.Test.foo.__require(ref void function(void*) bar)` is not callable using argument types `(extern (C) void function(void*))`
cannot pass argument `bar` of type `extern (C) void function(void*)` to parameter `ref void function(void*) bar`
```
Skipping a few syntaxCopies with the hack, it compiles and runs fine.

**testcontracts** hit an assertion at runtime, but compiled fine. With the hack, it doesn't compile anymore (delegate arg -> lazy param):
```
function Class.f4.__require(ref int i, lazy int j) is not callable using argument types (int, int delegate() pure @nogc @safe)
cannot pass argument __dgliteral3 of type int delegate() pure @nogc @safe to parameter lazy int j
```
It works for DMD though. I quickly skimmed over our frontend diff in files I deemed relevant but found nothing. I tried disabling https://github.com/ldc-developers/ldc/blob/5e5ce199f493091e6aea39f7f19f609dcf683da2/dmd/func.d#L2446-L2451 and https://github.com/ldc-developers/ldc/blob/5e5ce199f493091e6aea39f7f19f609dcf683da2/dmd/semantic3.d#L415-L436, to no avail.

kinke:
Getting this green turned out about as much work as merging 2.086.0 ;) - much action on DMD's stable branch lately.

The `rt.dwarfeh` 'fix' is just a hack which happened to work out after debugging the issue a bit. See https://github.com/dlang/druntime/pull/2602#discussion_r287568173 for some context.

PR #3082, opened 2019-05-31
Fix C assert calls for uClibc

### LINKED ISSUE #3078: uClibc and -checkaction=C
uClibc's `assert()` implementation uses a 4-parameter `__assert()` function, like the `__assert_fail()` function in other libcs.

Currently ldc2 doesn't recognise uClibc and falls back to generating code for a 3-parameter `__assert()`.  I don't see an easy fix because LLVM's target triples don't seem to support uClibc.

kinke:
Resolves #3078.


JohanEngelen:
lgtm

Issue #3083, opened 2019-05-31
VS debugger throws Error: Could not find 'runtime\objects\rt\dmain2.obj'


rainers:
Not sure when it started, but with recent versions of VS2017 or VS2019 I get this message in the debugger when the callstack of a simple "hello world" app is shown:

Error: Could not find 'runtime\objects\rt\dmain2.obj' in 'c:\D\ldc2-1.15.0-windows-multilib\lib64\druntime-ldc.lib'. WindowsApp2.exe was built with /DEBUG:FASTLINK which requires object files for debugging.

Maybe "fastlink" shouldn't be used for releases?

kinke:
That has got got to be a changed default setting for the MS linker, apparently depending on whether invoked from within the IDE. LDC itself only adds `/DEBUG` (with `-g[c]`). From [MSDN](https://docs.microsoft.com/en-us/cpp/build/reference/debug-generate-debug-info?view=vs-2019):

> When you specify /DEBUG with no additional options, the linker defaults to /DEBUG:FULL for command line and makefile builds, for release builds in the Visual Studio IDE, and for both debug and release builds in Visual Studio 2015 and earlier versions. Beginning in Visual Studio 2017, the build system in the IDE defaults to /DEBUG:FASTLINK when you specify the /DEBUG option for debug builds. Other defaults are unchanged to maintain backward compatibility.

rainers:
Indeed, the message disappears with /DEBUG:FULL. But AFAICT the release library doesn't contain any debug info. I also don't see similar messages with dmd.
BTW: there is an initial MessageBox, with further messages to the output window, but operation continues.

Trying again, the message appears when stepping into template instances of writeln. The first one is about sections_win64, so unlikely to be related to the actual callstack. I suspect there is something unexpected about the library, maybe that VS doesn't like the paths in the object file names. Is it generated by LLD instead of the MS lib tool?

kinke:
> Is it generated by LLD instead of the MS lib tool?

Sort-of (LLD doesn't archive AFAIK), it's a slightly modified copy of LLVM's `llvm-lib.exe` tool embedded in LDC.

I suppose everything's good when linking in a command-line (VS command prompt and naked) and so really just an issue if invoked within VS/VisualD?

rainers:
Yeah, probably whenever linked with /DEBUG:FASTLINK which is the default for VC projects.

kinke:
[...]

PR #3085, opened 2019-05-31
MSVC: Work around /DEBUG:FASTLINK incompatibility of release default …

kinke:
…libs archived with llvm-lib. Works around #3083.


rainers:
I'm afraid this is the wrong fix, as the linker invocation is not done by LDC. I'd rather expect the released libraries to be created by the standard lib.exe. Is that feasable? It probably won't work for the LTO libraries...

PR #3087, opened 2019-06-08
Support building with gdmd host compiler

MoritzMaxeiner:
Depends on [this](https://github.com/D-Programming-GDC/gdmd/pull/6) gdmd pull request.

### DIFF
--- a/cmake/Modules/ExtractDMDSystemLinker.cmake
+++ b/cmake/Modules/ExtractDMDSystemLinker.cmake
@@ -1,10 +1,8 @@
 # Determines the system linker program and default command line arguments
-# used by a DMD-compatible D compiler to link executables.
+# used by a (CLI-wise, DMD-compatible) D compiler to link executables.
 #
 # The following variables are read:
 #  - D_COMPILER: The D compiler command (i.e., executable) to use.
-#  - D_COMPILER_DMD_COMPAT: Must be TRUE, only compilers with a DMD-compatible
-#    command line interface are supported.
 #
 # The following variables are set:
 #  - D_LINKER_COMMAND: The system linker command used (gcc, clang, …)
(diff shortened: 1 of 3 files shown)


cmake/Modules/ExtractDMDSystemLinker.cmake
```diff
@@ -34,6 +34,8 @@ execute_process(
     RESULT_VARIABLE result_code
     OUTPUT_VARIABLE stdout
     OUTPUT_STRIP_TRAILING_WHITESPACE
+    ERROR_VARIABLE stdout
+    ERROR_STRIP_TRAILING_WHITESPACE
```
kinke: Thanks! - Merging stderr output into the `stdout` variable breaks the linker line detection if the linker writes to stderr, as shown by Shippable CI for AArch64.

MoritzMaxeiner: Okay, that's concerning, but that's what CI is for. gdmd requires parsing `stderr`, so I'll take this into account.

MoritzMaxeiner: Should be solved in the new commit?

MoritzMaxeiner:
@kinke semaphore ci job was killed while executing. Should I force push again to trigger a new job?

kinke:
No worries, that happens quite frequently lately, unfortunately.

MoritzMaxeiner:
The required [gdmd PR](https://github.com/D-Programming-GDC/gdmd/pull/6) has been merged.

kinke:
I'd like to kill the detection of `ldc2` and `gdc`, `D_COMPILER_DMD_COMPAT` etc., as it is currently in vain and will most likely remain unsupported in the foreseeable future. Any objections before I get to it?

cmake/Modules/ExtractDMDSystemLinker.cmake
kinke: Only as long as the linker used by gdc doesn't produce any warnings I guess, but oh well. ;)

MoritzMaxeiner:
@kinke I'm sorry, I don't think I understand what exactly you're asking. Do you want to close this PR without merging?

kinke:
Ah sry, that wasn't directed at you, but at other LDC devs.

JohanEngelen:
> I'd like to kill the detection of `ldc2` and `gdc`, `D_COMPILER_DMD_COMPAT` etc., as it is currently in vain and will most likely remain unsupported in the foreseeable future. Any objections before I get to it?

I think that's ok. It'd be nice to change the error `No D compiler found` to something like `No D compiler found (dmd, ldmd2, gdmd)`, so it is more clear that it is looking for DMD-compatible compiler commandline interface (but writing "DMD compatible" won't make sense to most people).

PR #3089, opened 2019-06-10
add __traits(getTargetInfo, "dcomputeTargets")

thewilsonator:
and __traits(getTargetInfo, "dcomputeFilePrefix”. Also fixes a segfault
when passing `-mdcompute-targets=whatever` and no compute modules were
passed to ldc. The test doubles for that fix as well.

https://github.com/dlang/dmd/pull/10015 is the DMD pr for the addition of TupleExp::create

### DIFF
--- a/gen/target.cpp
+++ b/gen/target.cpp
@@ -269,5 +269,21 @@ Expression *Target::getTargetInfo(const char *name_, const Loc &loc) {
   if (name == "cppStd")
     return createIntegerExp(static_cast<unsigned>(global.params.cplusplus));
 
+#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX
+  if (name == "dcomputeTargets") {
+    Expressions* exps = new Expressions();
+    for (auto &targ : opts::dcomputeTargets) {
+        exps->push(createStringExp(mem.xstrdup(targ.c_str())));
+    }
+      return TupleExp::create(loc, exps);
+
+  }
+
+  if (name == "dcomputeFilePrefix") {
+    return createStringExp(
+                mem.xstrdup(opts::dcomputeFilePrefix.c_str()));
+  }
+#endif
+    
   return nullptr;
 }
--- a/tests/semantic/target_traits_dcompute.d
+++ b/tests/semantic/target_traits_dcompute.d
@@ -1,0 +1,8 @@
+// Tests LDC-specific target __traits
+
+//RUN: %ldc -c %s -mdcompute-targets=cuda-350 -mdcompute-file-prefix=testing
+// REQUIRES: target_NVPTX
+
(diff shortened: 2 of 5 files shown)


tests/semantic/target_traits_dcompute.d
```diff
@@ -0,0 +1,8 @@
+// Tests LDC-specific target __traits
+
+//RUN: %ldc -c %s -mdcompute-targets=cuda-350 -mdcompute-file-prefix=testing
+// REQUIRES: target_NVPTX
+
+// CHECK-NOT: Error:
```
JohanEngelen: you don't run filecheck, so this line can be removed? (test will already fail if the LDC invoke fails)

tests/semantic/target_traits_dcompute.d
```diff
@@ -0,0 +1,8 @@
+// Tests LDC-specific target __traits
+
+//RUN: %ldc -c %s -mdcompute-targets=cuda-350 -mdcompute-file-prefix=testing
```
JohanEngelen: nit: put a space between `//` and `RUN`

tests/semantic/target_traits_dcompute.d
thewilsonator: Whoops, I think that was left over from some debugging, I'll fix it tomorrow.

gen/target.cpp
```diff
@@ -269,5 +269,21 @@ Expression *Target::getTargetInfo(const char *name_, const Loc &loc) {
   if (name == "cppStd")
     return createIntegerExp(static_cast<unsigned>(global.params.cplusplus));
 
+#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX
+  if (name == "dcomputeTargets") {
+    Expressions* exps = new Expressions();
+    for (auto &targ : opts::dcomputeTargets) {
+        exps->push(createStringExp(mem.xstrdup(targ.c_str())));
```
JohanEngelen: nit, are the strdups needed here? the string options are constant and available for whole duration of execution? (`targ` is not a temporary).  I'm OK with erring on the safe side here though, I guess we don't expect the user to CTFE this 1million times. It does add a memleak though.

thewilsonator: I wary of creating potentially stale pointers to `std::string`s managed by LLVM.

> It does add a memleak though.

True, and I just realised that so does the `new` three lines up! Forgot this is C++ lol.

> I'm OK with erring on the safe side here though, I guess we don't expect the user to CTFE this 1million times

that would be a problem if they called the trait a million times. I think `enum = __trait(dcomputeTargets)` caches the initialising expression, no? In which case `-lowmem` would be able to fix it.

thewilsonator:
Bummer accidentally issued this from master.

PR #3097, opened 2019-06-25
[ltsmaster] Enable testing with LLVM 7

redstar:
The idea here is to use xenial release and the clang compiler
which comes with LLVM.


kinke:
> The following tests FAILED:
> 22 - core.internal.convert (Failed)
> 232 - std.math (Failed)
> 352 - core.internal.convert-debug (Failed)

IIRC, that is https://github.com/ldc-developers/ldc/blob/c81a0c94e535c3d670a4fab7aca9b1de22c3778c/driver/targetmachine.cpp#L405-L411

redstar:
Great! Thanks for the hint!

redstar:
Travis is green now.

PR #3105, opened 2019-07-07
Do not require gold plugin when linking with LLD

myfreeweb:
LLD supports plugin-style flags, but the plugin itself might not exist. (`getLTOGoldPluginPath` might fail)

(on FreeBSD, it exists in an LLVM-version-specific place like `/usr/local/llvm80/lib/LLVMgold.so`. that should probably be added as well, the plugin is just not required for LLD. LLD is the primary linker here, `/usr/bin/ld` is LLD on aarch64 and amd64.)

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -69,9 +69,10 @@ private:
   void addDefaultPlatformLibs();
   virtual void addTargetFlags();
 
-  void addLTOGoldPluginFlags();
+  void addLTOGoldPluginFlags(bool requirePlugin);
   void addDarwinLTOFlags();
   void addLTOLinkFlags();
+  bool isLldDefaultLinker();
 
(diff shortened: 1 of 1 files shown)


myfreeweb:
hm, I wonder if lld on macos expects Darwin flags

kinke:
I guess calling `addLTOGoldPluginFlags()` makes no sense for LLD in general, i.e., adding the other `-plugin-*` options.

myfreeweb:
I said "LLD supports plugin-style flags" for a reason :)

https://github.com/llvm-mirror/lld/blob/beb1e40a02f8be0cb2c71d551d86f84d495d41e3/ELF/Options.td#L505-L521 ← it's real!

JohanEngelen:
LGTM. Can we add a test for this some way, or are we happy with not testing it?

PR #3106, opened 2019-07-07
 Enable gc-sections on FreeBSD when not using old ld.bfd linker

myfreeweb:
(Requires #3105 for `isLldDefaultLinker`)

The ancient ld.bfd doesn't like this:

```
.dub/obj/dtest.o: In function `ldc.register_dso':
app.d:(.text.ldc.register_dso+0x1c): undefined reference to `__stop___minfo'
app.d:(.text.ldc.register_dso+0x28): undefined reference to `__start___minfo'
/usr/bin/ld.bfd: .dub/build/application-release-posix.freebsd-x86_64-ldc_2086-54657107D68790E5698D25167B20AC6A/dtest: hidden symbol `__start___minfo' isn't defined
/usr/bin/ld.bfd: final link failed: Nonrepresentable section on output
```

(miiight be the same bug that's mentioned in the context of PGO on Linux)

but other linkers are of course fine, so let's just enable it when we know BFD is not used. (Eventually it won't be used by default at all: as of FreeBSD 12, LLD is default on amd64 and aarch64 already.)

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -470,11 +470,12 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
   }
 
   // Right now, we only support linker-level dead code elimination on Linux
-  // using the GNU toolchain (based on ld's --gc-sections flag). The Apple ld
-  // on OS X supports a similar flag (-dead_strip) that doesn't require
-  // emitting the symbols into different sections. The MinGW ld doesn't seem
-  // to support --gc-sections at all, and FreeBSD needs more investigation.
+  // and FreeBSD using GNU or LLD linkers (based on the --gc-sections flag).
+  // The Apple ld on OS X supports a similar flag (-dead_strip) that doesn't
+  // require emitting the symbols into different sections. The MinGW ld doesn't
+  // seem to support --gc-sections at all.
   if (!noLinkerStripDead && (triple.getOS() == llvm::Triple::Linux ||
+                             triple.getOS() == llvm::Triple::FreeBSD ||
                              triple.getOS() == llvm::Triple::Win32)) {
     targetOptions.FunctionSections = true;
     targetOptions.DataSections = true;
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -549,7 +549,11 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
       addLdFlag("-rpath", rpath);
   }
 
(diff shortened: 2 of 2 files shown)


driver/targetmachine.cpp
```diff
@@ -470,11 +470,12 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
...
-  // on OS X supports a similar flag (-dead_strip) that doesn't require
-  // emitting the symbols into different sections. The MinGW ld doesn't seem
-  // to support --gc-sections at all, and FreeBSD needs more investigation.
+  // and FreeBSD using GNU or LLD linkers (based on the --gc-sections flag).
+  // The Apple ld on OS X supports a similar flag (-dead_strip) that doesn't
+  // require emitting the symbols into different sections. The MinGW ld doesn't
+  // seem to support --gc-sections at all.
   if (!noLinkerStripDead && (triple.getOS() == llvm::Triple::Linux ||
+                             triple.getOS() == llvm::Triple::FreeBSD ||
                              triple.getOS() == llvm::Triple::Win32)) {
```
kinke: @ all: Is there a good reason for this OS-specific white-list wrt. separate sections by default? Can't we just make it a blacklist and exclude OSX, or could that break things down on more exotic BSDs etc.? I'm just thinking that it's easier to add a `-L--gc-sections` manually on those platforms (linker cmdline only!), instead of having to figure out that `-data-sections -function-sections` is required too in each cmdline, in order to reduce the binary size...

myfreeweb: According to the comment, you'd also have to blacklist MinGW.

OpenBSD has adopted LLD too (in 6.5 for amd64 and i386, on aarch64 probably since the beginning because that Last GPLv2 BFD Version predates aarch64, on other architectures not yet), DFly and NetBSD seem to still have That as the default linker.

IIRC That BFD Bug only happens when actually doing `gc-sections`, so always emitting `function-sections` on these OSes should be okay.

kinke: > wrt. separate sections by default

That was inaccurate; this block here *overrides* the LLVM defaults or user-provided cmdline options.

> so always emitting function-sections on these OSes should be okay

Thx. LLVM also defaults to separate sections for WebAssembly.
So I think enforcing separate sections with (default) `-disable-linker-strip-dead=false` for all targets except OSX should be fine. If the linker somehow doesn't like it, the user can still control it in the cmdline.

Issue #3118, opened 2019-07-26
warning about critical section size in wasm compile


skoppe:
since ldc 1.16.0 I get this `Warning: Assuming critical section size = 40 bytes` when compiling for WebAssembly.

Dub treats warnings as errors by default, causing builds to fails unless you add allowWarnings (which I don't really want to).

I speculate that in a44c78fb88fdd9559d0f90fb6847b920a0cfd9ef the critical section check is move earlier in compilation, causing it to be evaluated, whereas before it didn't in wasm builds.

kinke:
> I speculate [...]

You're absolutely right. We should return 0 for wasm in `getCriticalSectionSize()`, which then leads to a compilation error as soon as it is really used in `Target.critsecsize()`.

kinke:
Fixed in #3119, thx again.

PR #3119, opened 2019-07-26
return 0 in getCriticalSectionSize for wasm

skoppe:
https://github.com/ldc-developers/ldc/issues/3118

### DIFF
--- a/gen/target.cpp
+++ b/gen/target.cpp
@@ -78,6 +78,9 @@ unsigned getCriticalSectionSize(const Param &params) {
     break;
   }
 
+  if (arch == llvm::Triple::wasm32 || arch == llvm::Triple::wasm64)
+    return 0;
+
 #ifndef _MSC_VER
   unsigned hostSize = sizeof(pthread_mutex_t);
   warning(Loc(), "Assuming critical section size = %u bytes", hostSize);
--- a/tests/baremetal/wasm.d
+++ b/tests/baremetal/wasm.d
@@ -2,7 +2,7 @@
 
 // REQUIRES: target_WebAssembly
 // REQUIRES: internal_lld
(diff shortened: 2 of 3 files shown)


gen/target.cpp
```diff
@@ -78,6 +78,9 @@ unsigned getCriticalSectionSize(const Param &params) {
     break;
   }
 
+  if (arch == llvm::Triple::wasm32 || arch == llvm::Triple::wasm64)
+    return 0;
+
```
kinke: Thx! - Please get rid of the changes in the other files (part of the front-end [except for the init in main.cpp], where LDC-specifics are guarded by `version (IN_LLVM)` to simplify future front-end merges - and DMD won't feature wasm support anytime soon, if ever).

dmd/target.d
```diff
@@ -321,6 +321,10 @@ else // !IN_LLVM
             // sizeof(pthread_mutex_t) for Solaris.
             return 24;
         }
+        else if (params.isWasm)
+        {
+            return 0;
+        }
```
kinke: [This is the upstream implementation of `getCriticalSectionSize()` and excluded from LDC, see line 123. `Target` is a front-end class which heavily depends on the back-end, that's why LDC only uses a few functions from upstream.]

gen/target.cpp
skoppe: you'll still need the changed to `dmd/globals.h` right? else where does `global.params.isWasm` point to in main.cpp:init?

kinke: I meant `main.cpp` being the only other file outside the front-end, but the change in there is superfluous, we just need the 3 lines above - and ideally a test, maybe adding a `-w` here? https://github.com/ldc-developers/ldc/blob/d04e7081511a5d14546ab8eec6f3b66c13d62ec5/tests/baremetal/wasm.d#L5

skoppe: done and done

PR #3129, opened 2019-08-10
-fvisibility=hidden: Hide init symbols, TypeInfos and vtables too if the associated aggregate isn't exported

### LINKED ISSUE #3122: Webassembly and --export-dynamic causes too much bloat
First of all, I am sorry, since I was probably the reason `--export-dynamic` got put in by default, but it is causing too much bloat.

It will actually export all symbols, including things which aren't even used. In the case of the todo-mvc example in spasm, as a consequence the binary grows from 62Kb to 4Mb.

I know I can remove the switch from `ldc2.conf`, but this doesn't scale well to end users.

Again, I am sorry for not noticing this earlier. Next beta's I will definitely check. Probably will setup my CI to do the tests for me.

I would prefer if we can remove `--export-dynamic` again. Then if people want to export functions they need to annotate the function with `extern(C) export @assumeUsed` (when https://reviews.llvm.org/D57869 is released), or alternatively they can drop the `@assumeUsed` attribute and add `--export=funcname` to the linker invocation.

kinke:
Fixes #3122.

### DIFF
--- a/gen/declarations.cpp
+++ b/gen/declarations.cpp
@@ -146,7 +146,7 @@ public:
       auto &initZ = ir->getInitSymbol();
       auto initGlobal = llvm::cast<LLGlobalVariable>(initZ);
       initZ = irs->setGlobalVarInitializer(initGlobal, ir->getDefaultInit());
-      setLinkage(decl, initGlobal);
+      setLinkageAndVisibility(decl, initGlobal);
 
       // emit typeinfo
       if (!ir->suppressTypeInfo()) {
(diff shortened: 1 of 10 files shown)


kinke:
Pinging @skoppe.

skoppe:
Nice! Gonna try out right away. Any chance this is going to be part of 1.17?

kinke:
Quite likely, in case it works for you/spasm.

skoppe:
Works like a charm!

kinke:
Nice, thx for testing.

PR #3130, opened 2019-08-10
Don't cast an uninitialized void __vector

baziotis:
I'm not really sure about this. We certainly don't want to try to cast when there's a `void` as element type (so, cast from `ubyte` to `void`). But I don't know if it's a good solution.

_Edit_: The doc seems to agree: [Void Arrays](https://dlang.org/spec/arrays.html#void_arrays).
Which is that `void[]` are mostly fancy `ubyte[]`. Although the 4th point mentions some GC stuff that I don't know how they apply to LDC. The one thing I don't know and I'll try to find is where the `ubyte` comes from in `VectorExp`.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -2612,7 +2612,12 @@ public:
     } else {
       Logger::println("normal (splat) expression");
       DValue *val = toElem(e->e1);
-      LLValue *llval = DtoRVal(DtoCast(e->loc, val, type->elementType()));
+      LLValue *llval;
+      if (type->elementType()->ty != Tvoid) {
+        llval = DtoRVal(DtoCast(e->loc, val, type->elementType()));
+      } else {
+        llval = DtoRVal(val);
+      }
       for (unsigned int i = 0; i < e->dim; ++i) {
         DtoStore(llval, DtoGEPi(vector, 0, i));
       }
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -2612,12 +2612,13 @@ public:
     } else {
       Logger::println("normal (splat) expression");
       DValue *val = toElem(e->e1);
-      LLValue *llval;
-      if (type->elementType()->ty != Tvoid) {
-        llval = DtoRVal(DtoCast(e->loc, val, type->elementType()));
-      } else {
-        llval = DtoRVal(val);
-      }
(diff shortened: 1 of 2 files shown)


kinke:
I'd leave the cast in, but sanitize the `elementType()` before (`void -> ubyte`). And not just for a splat expression (where the vector is initialized by a scalar; here, an implicit `ubyte(0)` added by the frontend for default initialization). but also the previous branch (array literal). The cast is there for safety, just in case the frontend doesn't provide a matching type (the generated AST isn't totally stable, so some fault-tolerance on our side is nice to have).

baziotis:
That seems like the dumbest way to do the conversion because I guess it is. I'm uploading mostly the idea, if you can help me to do it better.
To be a little more detailed on my tries and assumptions so that you can correct me for the future:
1) It took me a while to realize that the fields in classes in `mtype.h` are `public`. I like that, but what is the reasoning? I guess C++ / D interop is involved.
2) Constructors of `mtype.h` are `extern(D)` (in `mtype.d`). Is that intended? It seems that you can't construct an object.
3) I tried to convert the type more sanely. But obviously, converting is not as easy as taking the `elementType()`. So, I followed the `Type*` inheritance tree, and changed the fields appropriately to result in the converted type. How do we do that better?
4) What is the `ctype` field? It has no doc. I guess it is for caching types, so I had to de-cache it to rebuild the type again.

_Note_: I didn't do it for array literals so that I can do it better after reviewing that.

kinke:
What I meant was:
```C++
Type *elementType = v->elementType();
if (elementType->ty == Tvoid)
  elementType = Type::tuns8; // static Type*
```

baziotis:
Ok, simple enough. But to use such logic, shouldn't `Type::tuns8` have been cached / built already somehow ? The reasoning here is it's been built as part of `val = toElem(e->e1)` ?

JohanEngelen:
@baziotis Missing a testcase :) See https://wiki.dlang.org/LDC_Lit-based_testsuite

baziotis:
I guess you mean I didnt add one not that it failed one. My bad. Should one be added for every big fix?

baziotis:
Superseded by: https://github.com/ldc-developers/ldc/pull/3139

Issue #2344, opened 2017-09-24
`if(false)` elimination at `-O0`


JohanEngelen:
Should we do dead code elimination on `if (false)` at `-O0`?
GDC and DMD do so.

```d
void foo() {
	if (false) {
		int a = 1;	
	}
}
```
gives (LDC 1.4.0)
```asm
void example.foo():
        xor     eax, eax
        mov     cl, al
        test    cl, 1
        jne     .LBB0_1
        jmp     .LBB0_2
.LBB0_1:
        mov     dword ptr [rsp - 4], 1
.LBB0_2:
        ret
```

kinke:
Good catch; clang does it too, so I think we should.

JohanEngelen:
Similar work can be done for `if(true){}else{}`, `while(false)`.

Note that this is more complicated than it may appear.
```d
void foo(int a) {
    int temp;
    if (a) goto lab1;
    if (false) {
        lab1:
        temp = 42;
    }
}
```

baziotis:
Probably we can now close this issue :)

JohanEngelen:
yep, implemented by #3134

PR #3133, opened 2019-08-14
[WIP] Eliminate branch when the condition is false

baziotis:
Partially solving (i.e. only false conditions): https://github.com/ldc-developers/ldc/issues/2344
It does _not_ work! You can even see that 2 test cases I added fail. Now that I saw the issue again, I should add a test case for labels.
I'm uploading just for opinions. Essentially the problems are:
* Doesn't test for loops. Probably every statement (if, for, while - probably there are others I can't think now) should be handled separately.
* The `false_cond_with_else.d` should be easy to pass. Fortunately, the front-end breaks an `else if (cond)` to `else { if (cond) }` so if there is an else body, that should be handled.
* The label thing is the one I'm not sure about. Check all the statements in the body for a label statement ? I don't know.. I'll come back to it.

### DIFF
--- a/gen/statements.cpp
+++ b/gen/statements.cpp
@@ -314,53 +314,60 @@ public:
         return;
       }
     }
-
     DValue *cond_e = toElemDtor(stmt->condition);
     LLValue *cond_val = DtoRVal(cond_e);
-
-    llvm::BasicBlock *ifbb = irs->insertBB("if");
-    llvm::BasicBlock *endbb = irs->insertBBAfter(ifbb, "endif");
-    llvm::BasicBlock *elsebb =
-        stmt->elsebody ? irs->insertBBAfter(ifbb, "else") : endbb;
-
-    if (cond_val->getType() != LLType::getInt1Ty(irs->context())) {
-      IF_LOG Logger::cout() << "if conditional: " << *cond_val << '\n';
-      cond_val = DtoRVal(DtoCast(stmt->loc, cond_e, Type::tbool));
+    LLConstant* const_val = llvm::dyn_cast<LLConstant>(cond_val);
+    bool dead_code_br = const_val && const_val->isZeroValue();
+    IF_LOG {
+      if (dead_code_br) {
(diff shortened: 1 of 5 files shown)


JohanEngelen:
Some comments:
- focus on `if(0)` in this PR, do all the other things later
- I would omit the constant propagation testing in the tests here
- for labels, indeed I think the only thing you can do is search all statements for labels... This is painful. Perhaps you can first work on adding that functionality to the semantic analysis in the frontend. It would probably be cheap to store a list all labels in a function during semantic analysis because you are already iterating over all statements. Discuss it with the frontend people, I think it may make sense for other functionality aswell to have this information more readily available.

JohanEngelen:
Also: try and find out how DMD, GDC, (and Clang)  tackle this problem.

baziotis:
> Some comments:
> 
> * focus on `if(0)` in this PR, do all the other things later
> * I would omit the constant propagation testing in the tests here
> * for labels, indeed I think the only thing you can do is search all statements for labels... This is painful. Perhaps you can first work on adding that functionality to the semantic analysis in the frontend. It would probably be cheap to store a list all labels in a function during semantic analysis because you are already iterating over all statements. Discuss it with the frontend people, I think it may make sense for other functionality aswell to have this information more readily available.

Great thanks! I think it's better to close this PR and open a new more specific one. I'll ask on slack if we have any info for labels and how difficult would be for me to add it.

baziotis:
> Also: try and find out how DMD, GDC, (and Clang) tackle this problem.

I believe this is a very good advice. Unfortunately, and these are only things I've heard not that I have worked with any of the two: the DMD back-end is one of the most obscure (i.e. pretty much only Walter understands it). The GDC front-end too (i.e. only Iain understands it). I'll have a look at them, but I dare to say that maybe the most approachable is Clang.

baziotis:
Continues here: https://github.com/ldc-developers/ldc/pull/3134

PR #3134, opened 2019-08-15
Disable dead code when an if/else if condition is constant

baziotis:
A more specific PR, result of a previous: https://github.com/ldc-developers/ldc/pull/3133
Issue: https://github.com/ldc-developers/ldc/issues/2344

Although @JohanEngelen you proposed that I focus on `if (0)` on this PR, when I wrote this, `if (1)` had exactly the same logic and seemed like it fit quite nicely. I can remove it though. 

-- Doesn't take into consideration labels.

The comments in the previous PR are worth a look. The tl;dr is:
- A discussion about labels.
- @JohanEngelen proposed that I take a look at how other compilers (DMD, GDC and Clang) handle this.

Regarding the last one, I had no hope with DMD. I did spend some time and I could not even target where an if statement (or such) is handled.

Lastly, I don't quite get `insertBB` and `irs->DBuilder`. I know what a basic block is and that the DBuilder has to do with debug info. But I don't understand exactly how they are used, when etc.

Edit: I should probably add a `delete cond_e` as it is done in e.g. (WhileStatement](https://github.com/ldc-developers/ldc/blob/master/gen/statements.cpp#L410). Probably was missed by accident here.


baziotis:
Just saw GDC. I have some problems to compile from source and verify the path but 99% it's the following.
So, an `IfStatement` is handled here: [toir.cpp :: visit(IfStatement)](https://github.com/gcc-mirror/gcc/blob/master/gcc/d/toir.cc#L563)
You can see it is quite simple. Generating IR from GCC though (i.e. `fdump-tree-original`) has eliminated the code.
And that certainly happens with the call to `build_vcondition`. Which if you follow it, it eventually goes here [fold_build3_loc](https://github.com/gcc-mirror/gcc/blob/master/gcc/fold-const.c#L12411), which goes here [fold_ternary_loc](https://github.com/gcc-mirror/gcc/blob/master/gcc/fold-const.c#L11355).
Of course it's quite a complicated it path, I'm not even close to understanding it fully, but a couple important points:
a) It's GCC transformations that do the work, not GDC.
b) At some point, this function calls [contains_label_p](https://github.com/gcc-mirror/gcc/blob/master/gcc/fold-const.c#L11343) which checks if a sub-tree contains any label. Which seems to follow what I assumed in a previous comment that we seem to care only for whether a block contains any label and the labels per se. Which is good as it is I think quite easy info to have from the front-end.
Btw, note that this function seems to rather walk the tree to find if it has any labels.

baziotis:
Just saw Clang as well. I did not expect to have so clear code. And it's pretty similar to LDC to.
I was trying for about an hour to install clang without re-installing LLVM with no success. So, this is too without running.
An `IfStmt` is emitted here: [CodeGenFunction::EmitIfStmt](https://github.com/llvm-mirror/clang/blob/master/lib/CodeGen/CGStmt.cpp#L621). I'm glad that Clang seems to do it similarly. It's clever this `Executed - Skipped` usage. Although I see an `incrementProfileCounter` there which I don't do when I eliminate so that might be wrong.

Anyway, here there is again a call to [ContainsLabel](https://github.com/llvm-mirror/clang/blob/master/lib/CodeGen/CodeGenFunction.cpp#L1253].
Which is a very simple function. Simple to implement in LDC too it seems. And it walks the tree.

baziotis:
@kinke , @JohanEngelen what do you think? I'm pinging you as you may haven't seen it because it's marked as WIP and also fails.
The tl;dr is:
- It fails because it does not handle labels.
- To handle labels, we need to know if a block has any label inside. This info can be provided either by DMD or we have to walk the tree.
- The 2 comments above describe what GDC and Clang do. For GDC, GCC does the job while on Clang, Clang. Both walk the tree every time. For DMD, I could not find what it does.

I have discussed this on Slack and I don't see us getting this info, so I guess the only option is to walk the tree.
You can see that Clang has a very simple implementation of `ContainsLabel` because of `->children` method. The closest I know to this in DMD is `flatten` but this is not generic enough (e.g. `IfStatement` does not implement it). So, I guess the only option is to implement our own.

kinke:
Thx for the analysis. I don't find this `-O0` optimization very useful, that's why I didn't post earlier. I'm not sure what the goal is - reducing compile times for countless runtime-ifs with static condition (alternative: `static if`), reducing `-O0` code size, ...?

If the compilation speed isn't the primary concern, we could make our lives easy and leverage some LLVM optimization pass to get rid of these superfluous branches and basic blocks. At `-O1`, one of the first passes is SimplifyCFG, which should do this, but it might also optimize some more which could be unexpected for debugging.

Checking for labels in the dead branch during codegen shouldn't be very hard; there are `Visitor` base classes in DMD which can be used to walk the subtree.

baziotis:
> Checking for labels in the dead branch during codegen shouldn't be very hard; there are `Visitor` base classes in DMD which can be used to walk the subtree.

Ok, I was not aware. Yes the implication with "implement our own" is that it can't be as simple as Clang's. There has to be some sort of visitor.

> Thx for the analysis. I don't find this `-O0` optimization very useful, that's why I haven't posted earlier. I'm not sure what the goal is - reducing compile times for countless runtime-ifs with static condition (alternative: `static if`), reducing `-O0` code size, ...?

TBH, I don't find it very useful either. I mean, it's good to have it, but if the expense is to have a whole visitor for it, probably not worth it. Even more in D, where as you said, there is `static if`. I guess the probabilities of having something like that are very low, except for very deep constant propagation where the fact that the condition is constant is not obvious to the programmer.

Here are though some maybe more subtle things:
a) Clang [mentions](https://github.com/llvm-mirror/clang/blob/master/lib/CodeGen/CGStmt.cpp#L644):
> This avoids emitting dead code and simplifies the CFG substantially.

One does not care to simplify the CFG in `-O0`. Which makes me think that Clang has added it to reduce compilation times but in _optimized_ builds.

b) Using a simple grep, I had seen `ContainsLabel` been used in a couple of other places. If we have more use cases, that could make it worth it (I can't think of any other).

> If the compilation speed isn't the primary concern, we could make our lives easy and leverage some LLVM optimization pass to get rid of these superfluous branches and basic blocks. At `-O1`, one of the first passes is SimplifyCFG, which should do this, but it might also optimize some more which could be unexpected for debugging.

I think that this is a bad option. Both for debugging but also, correct me if I'm wrong, LDC would like to be faster.

JohanEngelen:
I believe this helps in compilation speed for large if statements. The benefit is probably not so big, but the cost isn't either. (you should only traverse the tree when the condition is known at compile-time)

Some extra pointers:
- for traversing the AST, you can use `RecursiveVisitor` and `RecursiveWalker`. Upon visiting a `LabelStatement`, you can stop recursion by setting `stop` to `true`.
- `emitCounterIncrement` is our version of Clang's `incrementProfileCounter` (looks like Clang chose a better function name than I did). It should only be executed for the `if(true)` branch. This triggers me: you should add tests for correct profile counting aswell as correct coverage (`-cov`) counting.

baziotis:
> you should only traverse the tree when the condition is known at compile-time
Yes of course.

> * for traversing the AST, you can use `RecursiveVisitor` and `RecursiveWalker`. Upon visiting a `LabelStatement`, you can stop recursion by setting `stop` to `true`.

Thanks I planned to implement my own visitor. I'm comfortable with the solution, it's the DMD (?) visitors that I'm not sure about. I'll search about them.

> * `emitCounterIncrement` is our version of Clang's `incrementProfileCounter` (looks like Clang chose a better function name than I did). It should only be executed for the `if(true)` branch. This triggers me: you should add tests for correct profile counting aswell as correct coverage (`-cov`) counting.

Clang seems to have both [emitCounterIncrement](https://clang.llvm.org/doxygen/classclang_1_1CodeGen_1_1CodeGenPGO.html#aaee6bde2d81c36ff2d3b6f34ed339921), [incrementProfileCounter](https://clang.llvm.org/doxygen/CodeGenFunction_8h_source.html#l01342).

I'm not at all familiar with PGO. It seems that it tracks branches (or basic blocks..) that are taken, so it makes sense that we want it only for `if (true)`.
For the tests that you referenced, are there example tests?

JohanEngelen:
I meant that you should create your own visitor subclassed from `RecursiveVisitor` and then use `RecursiveWalker`.

- coverage test: `dmd-testsuite/runnable/cov2.d`

- profile counting tests: `tests/PGO/...`  This will probably be hard to understand...

baziotis:
> * coverage test: `dmd-testsuite/runnable/cov2.d`

I had some problems last time I tried to run dmd testsuite. I'll see what I can do, but till then, in this
case, isn't coverage handled by the call to `emitCoverageLinecountInc`. This is above the test for elimination, which means it shouldn't change the current behavior.
Now, as I said I'm not familiar with PGO, but this seems wrong. i.e. maybe this should not be called if we have `if (false)` ?

> * profile counting tests: `tests/PGO/...`  This will probably be hard to understand...
It's weird yes. But I think it starts to make sense. For a simple code like this:
`if (...) { ... }` or `if (...) { } else if (...) { ... }` it seems that for every basic block that a _taken_
branch targets (i.e. in the first example, only the `if:`, not the `else`) there is a pattern:
- get PGO counter from an array
- add 1 to it (i.e. mark taken)
- save back
and the array seems to have as many counters as basic blocks.
Oh, and there's a counter for entering the function as well.

Is it kind of what I described? And the PGO tests should try to verify that every block has this pattern?
Although, the tests themselves are quite complicated, I don't yet understand them.

gen/statements.cpp
```diff
@@ -46,6 +47,32 @@ void CompoundAsmStatement_toIR(CompoundAsmStatement *stmt, IRState *p);
...
+/// to the depth-first recursion.
+struct ContainsLabel : public StoppableVisitor {
+  bool found_label;
+  RecursiveVisitor rv;
+
+  ContainsLabel() : found_label(false) {}
+
+  void visit(LabelStatement *stmt) override {
+    found_label = 1;
+    stop = 1;
```
JohanEngelen: nit: `true`

gen/statements.cpp
```diff
@@ -285,6 +312,14 @@ class ToIRVisitor : public Visitor {
     }
   }
 
+  bool containsLabel(Statement *stmt) {
+    if (!stmt) return false;
```
JohanEngelen: For whitespace: get `clang-format`, it's awesome :)

tests/codegen/false_cond_labels.d
```diff
@@ -0,0 +1,31 @@
...
+    if (0)
+    {
+L1:
+        a = 1;
+    }
+    else
+    {
+        goto L1;
+    }
+}
```
JohanEngelen: also add a more complex case where there is a label somewhere deep in a nested construct of `if`, `while`, `for`, `foreach`, `switch`, to test that you aren't forgetting any recursive iteration.

gen/toir.cpp
```diff
@@ -2596,6 +2596,11 @@ class ToElemVisitor : public Visitor {
 
     LLValue *vector = DtoAlloca(e->to);
 
+    Type *elementType = type->elementType();
+    // `void` array literals and vectors are `ubyte` under the hood.
+    if (elementType->ty == Tvoid)
+      elementType = Type::tuns8;
+
```
JohanEngelen: these changes shouldn't be in this PR.

JohanEngelen:
For coverage: although the code changes may clearly show no bugginess, it's good to test that 
```
   if(false)  // <--- this should still show a line count increase even though completely elided.
   {

   }
```

For PGO: indeed, there is only a counter for the if-branch, and not for the else-branch. (assuming we know the execution count before the if-statement, else_count = before_if - if_branch_count. I wrote the tests and also have to study them a bit ;-)  I think things will make more sense if you study the IR output of the PGO tests a little.

gen/statements.cpp
```diff
@@ -46,6 +47,32 @@ void CompoundAsmStatement_toIR(CompoundAsmStatement *stmt, IRState *p);
...
 //////////////////////////////////////////////////////////////////////////////
 
+/// Used to check if a stmt contains any label.
+/// It is a StoppableVisitor that stops when a label is found.
+/// It's to be passed in a RecursiveWalker. The RecursiveWalker
+/// calls this visitor which in turn calls a RecursiveVisitor to
+/// to the depth-first recursion.
+struct ContainsLabel : public StoppableVisitor {
+  bool found_label;
+  RecursiveVisitor rv;
```
JohanEngelen: sorry about the misinformation about `RecursiveVisitor`, you don't need it. You just pass this stoppablevisitor to recursive walker. Currently your `rv.visit(stmt);` recurses but doesn't do anything useful.

gen/statements.cpp
baziotis: Too much C..

gen/statements.cpp
baziotis: Oh yes, thanks for reminding

gen/toir.cpp
baziotis: Oh sorry, yes my bad. It was somehow merged with other branch.

gen/statements.cpp
baziotis: Oh thanks! The original version was only with `StoppableVisitor`. And then I was trying for half an hour to put `RecursiveVisitor` somewhere. I tried multiple inheritance, weird casts because I'm not the most OO guy so I said if Johan mentioned it, it must be relevant. And so I put it like this which at least was not harmful. :P

baziotis: The `found_label` is also useless. We can use `stop` but let me add a method.

tests/codegen/false_cond_labels.d
baziotis: Just double-checking to see if we're on the same page: Clang tests whether we've seen a case statement. If we have not seen a switch previously, then it signals that we have to visit the code. Now, AFAIK, both in C++ and D, a `case` statement can't appear without a `switch`, so I guess this check is there only so that if such a scenario exists, to actually visit the code and output the error.
In D, DMD does this job, that's why I did not write it.

baziotis:
> For coverage: although the code changes may clearly show no bugginess, it's good to test that
> 
> ```
>    if(false)  // <--- this should still show a line count increase even though completely elided.
>    {
> 
>    }
> ```

I get the idea but not how to do the test. [cov2](https://github.com/ldc-developers/dmd-testsuite/blob/ldc/runnable/cov2.d) or [dmd_coverDestPath doc](https://dlang.org/phobos/core_runtime.html#.dmd_coverDestPath) don't help either.
> For PGO: indeed, there is only a counter for the if-branch, and not for the else-branch. (assuming we know the execution count before the if-statement, else_count = before_if - if_branch_count.

Smart!
> I wrote the tests and also have to study them a bit ;-) I think things will make more sense if you study the IR output of the PGO tests a little.

They start to make sense yes. I'll write some tests tomorrow and you can review the way forward. :)
2 points:
a) Currently, when we have code `if (true) { }` we handle 2 PGO counters (one for the function, the other for the if), as you said above. Do we actually want that? I'm asking because if we are in such a case, we don't generate a branch, so the second counter will always be equal with the first.
b) Most of `PGO/*` tests are unsupported for me. Can I do something so that I can have the tools to run them?

Ah, unrelated but just cross my mind:
In this:
```
if (false) {
	L1:
	int a = 1;
} else {
	goto L1;
}
```
Currently we generate the `false` branch as we should but we also generate a branch instruction.
This can be avoided of course. Meaning, we should just generate the `if` code without branch, then the `else`. Now that I'm thinking of that, I also have to answer the question whether the order matters. A quick thought is not but I'll come back to it when I implement it.

tests/codegen/false_cond_labels.d
JohanEngelen: Good catch! Code like this is allowed in C and D:
```
void foo();

void g(int x) {
    final switch (x) {
    case 1:
        if (false) { // cannot elide!!
    case 2:
            foo();
        }
    }
}
```
so there can be a `case` inside an `if` that you may be on the verge of eliding because of `cond==false`. So indeed: if you come across a `case` that does not belong to a `switch` inside the if-branch, then that `case` counts as a label. (`switch` is simply a more polished `goto` scheme)

```
void foo();

void g(int x) {
    final switch (x) {
    case 1:
        if (false) { // can elide
            final switch(x) {
                case 2:
            }
            foo();
        }
    }
    case 2:
        if (false) { // canNOT elide
            final switch(x) {
                case 2:
            }
    case 4:
            foo();
        }
    }

}
```

JohanEngelen: For reference: read about Duff's device.

JohanEngelen:
> > For coverage: although the code changes may clearly show no bugginess, it's good to test that
> > ```
> >    if(false)  // <--- this should still show a line count increase even though completely elided.
> >    {
> > 
> >    }
> > ```
> 
> I get the idea but not how to do the test. [cov2](https://github.com/ldc-developers/dmd-testsuite/blob/ldc/runnable/cov2.d) or [dmd_coverDestPath doc](https://dlang.org/phobos/core_runtime.html#.dmd_coverDestPath) don't help either.

Note the `// POST_SCRIPT: runnable/extra-files/coverage-postscript.sh` in `cov2.d`.

> > For PGO: indeed, there is only a counter for the if-branch, and not for the else-branch. (assuming we know the execution count before the if-statement, else_count = before_if - if_branch_count.
> 
> Smart!
> 
> > I wrote the tests and also have to study them a bit ;-) I think things will make more sense if you study the IR output of the PGO tests a little.
> 
> They start to make sense yes. I'll write some tests tomorrow and you can review the way forward. :)
> 2 points:
> a) Currently, when we have code `if (true) { }` we handle 2 PGO counters (one for the function, the other for the if), as you said above. Do we actually want that? I'm asking because in such a case, the second counter will always be equal to the first.

Yeah that's fine. People and tooling may query the counters and expect there to be the counter with the correct count.

> b) Most of `PGO/*` tests are unsupported for me. Can I do something?

What platform are you on? When building LLVM, you also need to build LLVM's compiler-rt for the functionality. (should be part of LDC's LLVM repo)

> Ah, unrelated but just crossed my mind:
> In this:
> 
> ```
> if (false) {
> 	L1:
> 	int a = 1;
> } else {
> 	goto L1;
> }
> ```
> 
> Currently we generate the `false` branch as we should but we also generate a branch instruction.
> This can be avoided of course. Meaning, we should just generate the `if` code without branch, then the `else`. 

I would leave this optimization out, it's `-O0` after all ;) Basically treat this as a normal case where you cannot elide.

gen/statements.cpp
```diff
@@ -314,10 +342,32 @@ class ToIRVisitor : public Visitor {
         return;
       }
     }
-
     DValue *cond_e = toElemDtor(stmt->condition);
     LLValue *cond_val = DtoRVal(cond_e);
-
+    LLConstant *const_val = llvm::dyn_cast<LLConstant>(cond_val);
+    if (const_val) {
```
JohanEngelen: nit: define the const_val variable inside the if condition to reduce its scope

gen/statements.cpp
```diff
@@ -314,10 +342,32 @@ class ToIRVisitor : public Visitor {
...
+      Statement *executed = stmt->ifbody;
+      Statement *skipped = stmt->elsebody;
+      bool incrementPC = false;
```
JohanEngelen: i would either remove this boolean (you can recheck const_val->isZeroValue), or would name it according to how it is set: something like `ifBranchExecution`, initialize it to true and set it to false in the swap branch.

tests/codegen/const_cond.d
```diff
@@ -0,0 +1,80 @@
...
+    {
+        int a = 1;
+    }
+    else
+    {
+        int b = 2;
+    }
+}
+
+extern(C) void bar()
```
JohanEngelen: add check-label for each function, so there is no accidental matching to stuff of other functions.

tests/codegen/const_cond.d
```diff
@@ -0,0 +1,80 @@
...
+{
+    // CHECK: bar()
+    // CHECK-NOT: %a = alloca i32, align 4
```
JohanEngelen: check-not is tricky: first `check` is matched, and then in-between `check` the `check-not` are matched.

```
CHECK: 1
CHECK-NOT: 2
CHECK: 3
```
would let this pass:
```
1
3
2
```
so if we happen to codegen `%a` and `%b` in opposite order, then the test passes where it should fail.
Good idea to test the variable allocation btw! That's an additional benefit of this PR: it possible elides large variable allocations on the stack.

tests/codegen/const_cond_labels.d
```diff
@@ -0,0 +1,75 @@
...
+            L2:
+            // CHECK: store i32 10, i32* %c
+                c = 10;
+            }
+        default: assert(0);
+        }
+    }
+    else
+    {
+        goto L2;
```
JohanEngelen: also test that the complex if-branch is indeed elided without this `goto` ?

tests/codegen/false_cond_labels.d
baziotis: Ok, I got it, thanks! I hadn't considered such case. I had forgotten about things like Duff's device.

tests/codegen/const_cond.d
baziotis: Thanks, I did not know `CHECK-LABEL`.

tests/codegen/const_cond.d
baziotis: Thanks :) Oh, ok, I missed this info about `CHECK-NOT`. Understood!

baziotis:
> 
> Note the `// POST_SCRIPT: runnable/extra-files/coverage-postscript.sh` in `cov2.d`.

Thanks, I'll study it tomorrow.
> Yeah that's fine. People and tooling may query the counters and expect there to be the counter with the correct count.

I see, alright.

> 
> What platform are you on? When building LLVM, you also need to build LLVM's compiler-rt for the functionality. (should be part of LDC's LLVM repo)

I'm on ubuntu linux. When I ran CMake, I used `  -DCOMPILER_RT_INCLUDE_TESTS=OFF` as pointed in the Wiki. Maybe this is the problem.

> I would leave this optimization out, it's `-O0` after all ;) Basically treat this as a normal case where you cannot elide.

Indeed. Ok, as we're doing right now.

kinke:
> I used -DCOMPILER_RT_INCLUDE_TESTS=OFF as pointed in the Wiki. Maybe this is the problem.

I'm sure it isn't.

baziotis:
Yes my bad. You have pointed me to that again. It's for runtime tests in LLVM like asserts.

tests/codegen/const_cond.d
baziotis: I think now it tests what we care about without the previous problem.

gen/statements.cpp
```diff
@@ -314,10 +363,30 @@ class ToIRVisitor : public Visitor {
         return;
       }
     }
-
     DValue *cond_e = toElemDtor(stmt->condition);
     LLValue *cond_val = DtoRVal(cond_e);
-
+    // Is it constant?
+    if (llvm::dyn_cast<LLConstant>(cond_val) != NULL) {
```
baziotis: This incurs a double-call on the cast. I haven't checked how expensive it is, I guess not so much. The alternative to both constrain const_val's scope and have a single call is to do the declaration-initialization on the condition which looks bad.

tests/codegen/const_cond_labels.d
```diff
@@ -0,0 +1,154 @@
...
+        // Can elide
+        if (false) {
+            final switch (b) {
+            case 2:
+                // CHECK-NOT: store i32 2, i32* %c
+                c = 2;   
+            }
+    // Test that `switch` in higher or equal nesting level
+    // with a `case` does not impact the handling of `case`s.
+    case 1:
```
baziotis: This fails because of the last commit, although I'm not sure why it doesn't even compile (an assert fires on LLVM).

gen/statements.cpp
```diff
@@ -46,6 +47,47 @@ void CompoundAsmStatement_toIR(CompoundAsmStatement *stmt, IRState *p);
...
+  bool foundLabel(void) { return stop; }
+
+  void visit(Statement *stmt) override {}
+
+  void visit(LabelStatement *stmt) override { stop = true; }
+
+  // NOTE: This is wrong! We have to keep state (as in a recursion)
+  // so that when we come _out_ of a recursion to a switch statement,
+  // `ignore_case_statements` is restored.
+  void visit(SwitchStatement *stmt) override { ignore_case_statements = true; }
```
baziotis: I hope the comment makes it clear why this won't work. `ignore_case_statements` is set but never restored. So, if we have something like:
```
switch(a) {
case 1:
    if (x) {
        switch (b) { }
case 2:
    }
}
```
The `switch (b)` will set `ignore_case_statements` but we will never restore it back to "not ignore" for `case 2`.
The way I think of it, we need recursion or a stack or some notion of scope incrementing and decrementing. The problem is that to do any of those, I have to change `RecursiveWalker`.

baziotis: FWIW, the only somewhat clean solution that I can think of is to create a class that extends `StoppableVisitor` and has a couple more fields that `RecursiveWalker` will set accordingly. In this case, it will be a `inside_switch` field. But, it may be useful for future use.

gen/statements.cpp
JohanEngelen: > do the declaration-initialization on the condition which looks bad.

That's fine. It's a common idiom that is used throughout out codebase. `if (auto constCondVal  = llvm::dyn_cast<LLConstant>(cond_val)) {`  (don't make the `NULL` part explicit).

gen/statements.cpp
JohanEngelen: For now, I think it is best to simply implement your own walker inherited from RecursiveWalker that overrides `visit(SwitchStatement *stmt)` to add scope info.

gen/statements.cpp
baziotis: Yes that's what I meant, good.

gen/statements.cpp
baziotis: There are 2-3 possibilities that I can on that. I'll post the most generic one for review.

gen/recursivevisitor.h
```diff
@@ -476,3 +476,30 @@ class RecursiveWalker : public Visitor {
...
+public:
+  explicit ExtendedRecursiveWalker(ExtendedStoppableVisitor *visitor,
+                                   bool _continueAfterStop = true)
+      : RecursiveWalker(visitor, _continueAfterStop) {}
+
+  void visit(SwitchStatement *stmt) override {
+    ExtendedStoppableVisitor *ev = static_cast<ExtendedStoppableVisitor *>(v);
+    bool save = ev->inside_switch;
+    ev->inside_switch = true;
+    call_visitor(stmt) || recurse(stmt->condition) || recurse(stmt->_body);
```
JohanEngelen: here you could call `RecursiveWalker::visit(SwitchStatement *stmt)` (no virtual call), to avoid code dup and avoid future breakage.

I don't think this is general enough to put it these two classes in this file, rather than in the file where you need them. "Extended" is too vague, maybe something like "SwitchScope". Other than that, looks good.
You can change the bool `inside_switch` flag to a `SwitchStatement *`.

Btw, we don't use underscores in variables names, but camelCase instead.

JohanEngelen: Ah, I guess you chose "extended" so that later others can add more functionality to it. Perhaps not a bad idea. Then indeed it's ok to put it in these files. :)

tests/codegen/const_cond.d
JohanEngelen: Because of the ckec-not trickery, often you'll end up having to write the check-not twice.
In this case:
```
// CHECK-NOT: %a = alloca
// CHECK: %b = alloca
// CHECK-NOT: %a = alloca
```
(note that the `i32` and alignment are extra things that you don't want to test for)

tests/codegen/const_cond_labels.d
```diff
@@ -0,0 +1,154 @@
...
+void case_as_label(int a, int b)
+{
+    int c;
+    final switch (a) {
+    case 1:
+        // Can elide
+        if (false) {
+            final switch (b) {
+            case 2:
+                // CHECK-NOT: store i32 2, i32* %c
```
JohanEngelen: see my note above about check-not. It's needed twice here. Because of that, I think it's easiest to read when all checks are at the start of the function rather than interspersed with code.

tests/codegen/const_cond_labels.d
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+}
+
+// CHECK-LABEL: @case_as_label2
+void case_as_label2(int a, int b)
+{
+    int c;
+    final switch (a) {
+        // Can elide
```
JohanEngelen: it think this one canNOT be elided?

gen/recursivevisitor.h
baziotis: > Ah, I guess you chose "extended" so that later others can add more functionality to it. Perhaps not a bad idea. Then indeed it's ok to put it in these files. :)

Yes :) It's based on what I have said in another comment. Basically, the way I see it there are 2 cases:
a) Create specific visitors for each thing to test, all of which which extend `StoppableVisitor`. That would be e.g.  the `SwitchScope` thing. But then, if one wants to add functionality, they would have to create both a new visitor and a new walker.
b) Create an "extended" visitor and walker, where if we want to add any functionality (i.e. if we're on an if for some reason), we can add the notify act on the walker and the relative field in the visitor. That's sort of what I was going for. Although, I don't like the name either..
> You can change the bool inside_switch flag to a SwitchStatement *.

Yes, better so we can have access to the actual statement.

tests/codegen/const_cond.d
baziotis: Haven't thought about, good idea. Do we need it for this part?

tests/codegen/const_cond_labels.d
baziotis: `CHECK-NOT` is more complicated than it seems initially :P

tests/codegen/const_cond_labels.d
baziotis: Yes yes. Funny that I changed the `CHECK-NOT` to `CHECK` but not the comment..

tests/codegen/const_cond_labels.d
```diff
@@ -0,0 +1,156 @@
...
+                c = 2;   
+            }
+        }
+    case 2:
+        // Can't elide
+        if (false) {
+    case 3:
+            c = 3;
+        }
+    }
```
JohanEngelen: So here you should add something like:
```
        // Can't elide
        if (false) {
    default:
            c = 4;
        }
```

tests/PGO/const_cond.d
```diff
@@ -0,0 +1,40 @@
...
+
+// RUN: %ldc -c -output-ll -fprofile-instr-generate -of=%t.ll %s && FileCheck %s < %t.ll
+
+
+extern(C):
+
+// Test that we have exactly 2 counters for `foo` (one for the function, one for the `if`)
+// and only one counter for `bar` (bar has always false if, should be elided).
+// CHECK: @__{{(llvm_profile_counters|profc)}}_foo ={{.*}} [2 x i64] zeroinitializer
+// CHECK: @__{{(llvm_profile_counters|profc)}}_bar ={{.*}} [1 x i64] zeroinitializer
```
baziotis: I guess that this is a correct check, meaning we should only have one counter for the function. The thing is, it fails. While the checks below in bar succeed (only one counter is generated), in the array 2 counters are generated.
A little help on that would be appreciated. I can't follow where / how this is generated.

tests/codegen/const_cond_labels.d
baziotis: Just by looking at it, it will fail, thanks for spotting it! DMD has `DefaultStatement`, different from `CaseStatement`.

baziotis: Just looked at Clang again and it seems to differentiate between `SwitchCase` and `DefaultStatement` too. But does not check for the default in [ContainsLabel](https://github.com/llvm-mirror/clang/blob/master/lib/CodeGen/CodeGenFunction.cpp#L1253).

dnadlinger: DefaultStmt `isa` SwitchCase in Clang, though.

baziotis: Yes, thanks! I saw [the inheritance tree] https://clang.llvm.org/doxygen/classclang_1_1DefaultStmt.html) and left it there. My guess was that since it is derived and `isa` AFAIK checks the inheritance, that probably made it work. Thanks for verifying though.

tests/PGO/const_cond.d
JohanEngelen: don't you want there to be 2 counters in bar? Function entry + if branch. There shouldn't be a difference with or without known false branch condition.

tests/codegen/const_cond_labels.d
```diff
@@ -131,6 +133,11 @@ void case_as_label(int a, int b)
     case 3:
             c = 3;
         }
+        // Can't elide
+        if (0 && 1) {
```
JohanEngelen: probably better to write `false` here, otherwise the test relies on constant folding logic in the frontend.

tests/PGO/const_cond.d
baziotis: Ah ok. I thought this was only for true conditions because only for true conditions we emit a counter increment. But I guess we want to have a counter although we never reference it.

JohanEngelen: yes indeed, because the user can query these counters, see: https://github.com/ldc-developers/ldc/blob/a9b9097918ae209d0a35cf5dc2c2b342de51fb83/tests/PGO/profile_rt_calls.d#L30-L31

baziotis: Got it. I was late to update this because I had some problems running the tests. But I did a [PR](https://github.com/dlang/dmd/pull/10372) on DMD to add the contains label functionality. We'll see the reception.

baziotis:
I don't understand why circleci and and travis fail on PGO/const_cond.d. I can't reproduce it.
The semaphoreci failure seems totally unrelated.

thewilsonator:
I note that the trans failure occur with over LLVMs.

tests/PGO/const_cond.d
JohanEngelen: tbh, I don't think it will go into DMD. We have similar functionality that is not in DMD either: https://github.com/dlang/dmd/pull/5501

baziotis: Yes... it got an approval but I'm hesitant to put the work until more people agree.

baziotis:
> I note that the trans failure occur with over LLVMs.

Uhm, trans failure being.. ? :P

thewilsonator:
Oh, wow. I think that was meant to be **travis** failures occur with **older** LLVMs

baziotis:
> Oh, wow. I think that was meant to be **travis** failures occur with **older** LLVMs

Haha yes indeed. Does that mean it's my fault? (I don't have older LLVM to test).

JohanEngelen:
haven't looked in detail, but testing with older LLVM also means that FileCheck is older and `CHECK-COUNT` may not work, or work differently...

thewilsonator:
> Does that mean it's my fault?

Possibly (and see also Johan's response above), but they are old LLVMs and this is a new feature. While it would be nice for it to test successfully on older LLVMs, as long as the program generation is still valid.

gen/statements.cpp
```diff
@@ -356,11 +419,11 @@ class ToIRVisitor : public Visitor {
...
     }
 
-    // end the dwarf lexical block
-    irs->DBuilder.EmitBlockEnd();
-
     // rewrite the scope
     irs->scope() = IRScope(endbb);
+
+    // end the dwarf lexical block
+    irs->DBuilder.EmitBlockEnd();
```
thewilsonator: Is there any reason for the position of this to change?

baziotis:
> > Does that mean it's my fault?
> 
> Possibly (and see also Johan's response above), but they are old LLVMs and this is a new feature. While it would be nice for it to test successfully on older LLVMs, as long as the program generation is still valid.

Ok, good, thank you both. I'll try to experiment.

gen/statements.cpp
baziotis: Now, no. Do you want me to change it?

It was left because initially it was
`if (constant) {
  do smth
} else {
  do the normal thing
}
```
but where the smth and the normal thing had a common end, the end of the dwarf block.

**Comment by baziotis** [comment]
How about that.. it worked. Thanks guys. It fails only on the semaphore, but I think it's unrelated because it says execution expired.
Also, maybe we should include that info for `CHECK-COUNT`. Either if there's a way to version it. Or to include it in some doc.

**Comment by JohanEngelen** [comment]
> Also, maybe we should include that info for `CHECK-COUNT`. Either if there's a way to version it. Or to include it in some doc.

I think in the past we have backported some things of newer FileCheck into our own copies of older FileChecks, but looks like we no longer have that and I'd recommend against it unless it is the most awesome feature ever. What you can do is enable your test only for newer LLVM versions (e.g. `REQUIRES: atleast_llvm600`). In this case, just don't use check-count, the count is luckily not one thousand ;-)

**Comment by baziotis** [comment]
> I think in the past we have backported some things of newer FileCheck into our own copies of older FileChecks, but looks like we no longer have that and I'd recommend against it unless it is the most awesome feature ever. What you can do is enable your test only for newer LLVM versions (e.g. `REQUIRES: atleast_llvm600`). In this case, just don't use check-count, the count is luckily not one thousand ;-)

Haha yes I agree. I was thinking of maybe adding a note on [the wiki](https://wiki.dlang.org/LDC_Lit-based_testsuite#FileCheck) too.

**Comment by baziotis** [comment]
Are you happy with [this PR (Adding hasLabel info to DMD)](https://github.com/dlang/dmd/pull/10372) _if_ it gets merged ?

**Comment by baziotis** [comment]
I don't know if the DMD respective PR will ever get merged. It may be better to merge this instead of waiting. AFAIK, the only thing that remains is dmd-testsuite.
I'm sorry but I can't figure out how dmd-testsuite works on LDC. By doing:
`./run.d -u` as in DMD, it says that DMD is not installed and it searches in
`d2/generated/linux/release/32/dmd`.
There's something I'm missing, any help is appreciated.

**Comment by kinke** [comment]
Sorry, I missed your latest post - dmd-testsuite seems fine according to CI. To run it, in the build dir: `        DMD_TESTSUITE_MAKE_ARGS=-j$(nproc) ctest -V -R dmd-testsuite-debug`.

**Comment by baziotis** [comment]
> Sorry, I missed your latest post - dmd-testsuite seems fine according to CI. To run it, in the build dir: ` DMD_TESTSUITE_MAKE_ARGS=-j$(nproc) ctest -V -R dmd-testsuite-debug`.

Oh thanks, I was really struggling with this. If you agree, I'd like to add this in either [LDC contributor's guide](https://wiki.dlang.org/LDC_contributor%27s_guide#Test_suite) or the [dmd-testsuite LDC repo](https://github.com/ldc-developers/dmd-testsuite).

Yes, the suite runs fine but Johan had proposed [here](https://github.com/ldc-developers/ldc/pull/3134#issuecomment-523758730) that I should add coverage test. I'm still not sure what that means (I mean, I know what coverage is but I don't see anything interesting happen ing on either `cov2.d` or `extra-files/coverage-postscript.sh`)

**Comment by JohanEngelen** [comment]
What I meant is that you should add a coverage test that checks that:
```
  if(false) {  // if you elide this piece, it should still increase coverage execution count on this line
  }
```
You can look at `cov2.d` to see how coverage testing is done.

**Comment by JohanEngelen** [comment]
> > Sorry, I missed your latest post - dmd-testsuite seems fine according to CI. To run it, in the build dir: ` DMD_TESTSUITE_MAKE_ARGS=-j$(nproc) ctest -V -R dmd-testsuite-debug`.
> 
> Oh thanks, I was really struggling with this. If you agree, I'd like to add this in either [LDC contributor's guide](https://wiki.dlang.org/LDC_contributor%27s_guide#Test_suite) or the [dmd-testsuite LDC repo](https://github.com/ldc-developers/dmd-testsuite).

Go ahead! You (of course!) do not need approval to add such things to the wiki; it's a wiki! The fact that you struggled with it means that others will struggle too, so it's a valuable addition.

**Comment by JohanEngelen** [comment]
Perhaps you had not noticed this file yet: https://github.com/ldc-developers/dmd-testsuite/blob/a813069f3f095de2929f51ec2ddc695a6bc4d65d/runnable/extra-files/runnable-cov2.lst

**Comment by baziotis** [comment]
> Perhaps you had not noticed this file yet: https://github.com/ldc-developers/dmd-testsuite/blob/a813069f3f095de2929f51ec2ddc695a6bc4d65d/runnable/extra-files/runnable-cov2.lst

Oh well, yes.. :P
Thanks

**Comment by baziotis** [comment]
Probably something went wrong when trying to update the dmd-testsuite. It's kind of complicated with the sub-repositories. Should I do a PR to the dmd-testsuite repo?

**Comment by kinke** [comment]
Something I've never tried but may work as a temporary hack to make CI test submodule changes: edit `.gitmodules` and change the repo URL to yours.

**Comment by baziotis** [comment]
> Something I've never tried but may work as a temporary hack to make CI test submodule changes: edit `.gitmodules` and change the repo URL to yours. You'll probably have to rebase this PR (and your dmd-testsuite branch), as master's dmd-testsuite hash has been advanced in the meantime, hence the merge conflict here.

Ok, rebasing this branch made this not having merge conflicts. Now, I had rebased the dmd-testsuite submodule to the latest, which makes this PR show that there are 2 new commits in it (those are not mine).
Now the problem is I can't somehow submit _my_ changes to the dmd-testsuite. If I leave it as now (the changes are not commited to the submodule), the changes are not visible. If I commit them, then Github says something like I can't find what branch to compare etc.

**Comment by kinke** [comment]
I don't have the time to fully explain how git submodules work, just one note: it's nothing more than a nested repo *reference*, consisting of a repo URL (ldc-developers/dmd-testsuite) and a hash pointing to the desired commit. All development happens in the actual repo; members of ldc-developers can push directly to that repo (mostly creating a new branch for testing and then fast-forwarding the `ldc` branch when merging the associated ldc PR). In your case, you'll have to fork dmd-testsuite and push your changes to that repo. Changing the repo URL in `.gitmodules` might make CI testing work; opening a dmd-testsuite PR only works for some CI services IIRC.

The `ldc` dmd-testsuite branch already contains 2 commits from #3176 (basically due to laziness on my part and dmd-testsuite being touched almost exclusively by myself for a while now); you can skip those by branching your dmd-testsuite branch off of the commit used by current master (`git submodule update` to check out the current hashes). Normally, you'd do something like:
* `cd tests/d2/dmd-testsuite`
* `git checkout -b myBranch`
* `git commit -am "My changes"`
* `git push -u myDMDTestsuiteRemote myBranch`
* `cd ../../..`
* `git add tests/d2/dmd-testsuite` (update submodule hash) 
* `git commit` (commit updated submodule hash)
* `git push`

**Comment by kinke** [comment]
> I haven't forked dmd-testsuite. I assume that the changes should happen in the ldc branch, in the submodule.

This sounds like you were about to have 2 local dmd-testsuite repos, one in the ldc src tree, and another external one for your fork. That's definitely not needed; all you need is fork on GitHub and then, in the dmd-testsuite subdir in ldc's srctree: `git remote add mine <myURL>`.

**Comment by baziotis** [comment]
> I don't have the time to fully explain how git submodules work, just one note: ...

Thanks for the time to explain all those! I'm the worst when it comes to tools. :P
Let me try what you proposed.

**Comment by kinke** [comment]
You forgot the crucial .gitmodule change, that seems to work now (at least for Azure). Your dmd-testsuite branch is still based on current `ldc`, i.e., includes an unrelated cppmangle test change, which will most likely fail.

**Comment by baziotis** [comment]
> You forgot the crucial .gitmodules change, 

Thanks. I actually did the change but of course, because it's me, I forgot to `git add` it.

> that seems to work now (at least for Azure). Your dmd-testsuite branch is still based on current `ldc`, i.e., includes an unrelated cppmangle test change, which will most likely fail.

If I understood correctly, now it should be ok. I rebased back to the commit that the ldc dmd-testsuite submodule is and added my commit on top.

File: gen/recursivevisitor.h [hunk 21]
```diff
@@ -476,3 +476,33 @@ class RecursiveWalker : public Visitor {
   void visit(Initializer *init) override { call_visitor(init); }
   void visit(Dsymbol *init) override {}
 };
+
+/// As a StoppableVisitor but keeps additional informantion.
+class ExtendedStoppableVisitor : public StoppableVisitor {
```
kinke: I don't think there'll soon be a use case for another visitor derived from this ('extended' meaning 'with innermost switch statement'). So I suggest melding the `insideSwitch` field directly into the `ContainsLabel` visitor [and simply inline-initializing it as per C++11 - a `bool` would suffice as well].

gen/recursivevisitor.h
```diff
@@ -476,3 +476,33 @@ class RecursiveWalker : public Visitor {
...
+
+  explicit ExtendedStoppableVisitor()
+      : StoppableVisitor(), insideSwitch(nullptr) {}
+};
+
+/// As the RecursiveWalker, but it provides additional info to the
+/// visitor. For example, when inside a switch statement
+/// it sets `insideSwitch` to the switch statement and restores it back
+/// to nullptr when we're getting out.
+class ExtendedRecursiveWalker : public RecursiveWalker {
```
kinke: Without `ExtendedStoppableVisitor` class, there's no use for this walker in this general .h file, so I suggest moving it to the .cpp and naming it `ContainsLabelWalker`.

gen/statements.cpp
```diff
@@ -46,6 +47,35 @@ void CompoundAsmStatement_toIR(CompoundAsmStatement *stmt, IRState *p);
...
+    if (insideSwitch == nullptr)
+      stop = true;
+  }
+
+  bool foundLabel(void) { return stop; }
+
+  void visit(Expression *exp) override {}
+  void visit(Declaration *decl) override {}
+  void visit(Initializer *init) override {}
+  void visit(Dsymbol *) override {}
```
kinke: Instead of ignoring expressions here, you should ignore them in the walker to prevent superfluous recursion into all expressions.

gen/statements.cpp
```diff
@@ -285,9 +315,21 @@ class ToIRVisitor : public Visitor {
...
+    ExtendedRecursiveWalker walker(&labelChecker, false);
+    stmt->accept(&walker);
+    return labelChecker.foundLabel();
+  }
+
+  //////////////////////////////////////////////////////////////////////////
+
   void visit(IfStatement *stmt) override {
     IF_LOG Logger::println("IfStatement::toIR(): %s", stmt->loc.toChars());
-    LOG_SCOPE;
```
kinke: There's no LOG_SCOPE in the regular non-constant case anymore. I'd find it more logical for the constant case too, i.e., `Constant true/false condition - elide` being indented under `IfStatement::toIR()`, as it clearly belongs to that statement and is no sibling statement.

gen/statements.cpp
kinke: Oh and you'll need `using StoppableVisitor::visit;` (also in the walker: `using RecursiveWalker::visit;`) to suppress hundreds of warnings (see CI output) like
> warning: 'ContainsLabel::visit' hides overloaded virtual functions

gen/statements.cpp
```diff
@@ -314,10 +356,31 @@ class ToIRVisitor : public Visitor {
...
+        IF_LOG Logger::println("Constant true/false condition - elide.");
+        LOG_SCOPE;
+        // True condition, the branch is taken so emit counter increment.
+        if (!const_val->isZeroValue()) {
+          PGO.emitCounterIncrement(stmt);
+        }
+        if (executed)
+          executed->accept(this);
+        // end the dwarf lexical block
+        irs->DBuilder.EmitBlockEnd();
```
kinke: The paired start is missing - `irs->DBuilder.EmitBlockStart(executed->loc);`, apparently before the PGO counter increment to remain compatible with the regular code below.

kinke:
A little trick to quickly locate the dmd-testsuite failures for Azure: click on the dmd-testsuite step to open the log, then press Ctrl+F to open the VS-Code-search, and then search for `failed.` (yes, including the `.`).

gen/statements.cpp
```diff
@@ -46,6 +47,35 @@ void CompoundAsmStatement_toIR(CompoundAsmStatement *stmt, IRState *p);
...
+    if (insideSwitch == nullptr)
+      stop = true;
+  }
+
+  void visit(DefaultStatement *stmt) override {
+    if (insideSwitch == nullptr)
+      stop = true;
+  }
+
+  bool foundLabel(void) { return stop; }
```
kinke: [We don't use this archaic `(void)`.]

gen/recursivevisitor.h
baziotis: For clarification, 'extended' does not mean 'with innermost switch statement'. It is a visitor that keeps a bunch of info that the recursive walker can trigger. _One_ of which is whether we are inside a switch statement. You can check this discussion: https://github.com/ldc-developers/ldc/pull/3134#discussion_r317416317

Because of the same discussion, the `bool` was changed to also keep the statement.

gen/statements.cpp
baziotis: Maybe I misses something. The current isn't what you describe ? For example, I get this:

^ ^ ^ IfStatement::toIR(): test.d(4)
^ ^ ^ IntegerExp::toElem: false @ bool
^ ^ ^ ^ IntegerExp::toConstElem: false @ bool
^ ^ ^ ^ ^ value = i1 false
^ ^ ^ Constant true/false condition - elide.

(Stars being replaced with ^ because of github).

gen/recursivevisitor.h
kinke: This is classical over-engineering IMO - if there's ever a need for another similar visitor/walker, *then* one can refactor the existing stuff and put a base class somewhere. Right now it's just slightly complicating stuff and needlessly spreading an implementation across multiple files, and featuring comments that aren't needed if cleanly putting this in a .cpp section. I don't insist on the change, but I do definitely recommend it.

[There'd be an opportunity for another base class, and that's a `RecursiveStatementWalker`, just descending into statements and ignoring expressions etc. - your walker could use it, and the walker for the `MoreThanXStatements` visitor.]

gen/statements.cpp
kinke: The current output for:
```D
void foo(bool a)
{
    if (a)
        int x = 3;
}
```

```
* * * IfStatement::toIR(): onlineapp.d(2)
* * * * VarExp::toElem: a @ bool
* * * * * DtoSymbolAddress ('a' of type 'bool')
* * * * * * function param
* * * * * * type: bool
* * * * ScopeStatement::toIR(): onlineapp.d(3)
* * * * * ExpStatement::toIR(): onlineapp.d(3)
...
```

https://run.dlang.io/is/fFdB26

gen/statements.cpp
baziotis: My bad. I don't know how DWARF blocks work on LDC (and couldn't understand much). So, this code is more "guessing from context" than thought out.
That block end is supposed to end the block at the start of the function. So, that has to be transformed to the one below ?
```
if (executed) {
  // if (!zeroValue) { PGO.incrCounter }
  // emit block start for 'executed'
  // executed->accept
  // emit block end for 'executed'
}
// emit block end for the block at the start of the function
```

gen/statements.cpp
baziotis: Ah, I get it yes. Btw, sorry, with current I meant this PR's current output, not upstream current output. So the problem is not that `Constant true/false condition - elide`, it's that in this PR, e.g. `VarExp` is not indented. I think that to fix this, one `LOG_SCOPE` has to be added where it is now. And another outside (and after) the block that deals with constant expressions (the `if (LLConstant ...)`.

gen/statements.cpp
kinke: > That block end is supposed to end the block at the start of the function.

Nope, these blocks represent the then/else blocks for debug infos (and fallback LOC etc.). There's no other block to be ended.

> So, that has to be transformed to the one below ?

As I said, according to the code for the non-constant case below, the block starts before the PGO counter increment, although I find that a bit strange. @JohanEngelen should know, he added this.

gen/statements.cpp
baziotis: And what should be used ? I generally use that because of C, where one can have old style and new style functions. Not putting `(void)` means that the function is old style and you can call it with whatever. The fun part is that if you _do_ call it with whatever (in most cases by accident), the compiler doesn't tell anything.

baziotis: Huh, my bad. No old style functions in C++ it seems. Thank god I see some (only some though) amazingly bad C features being taken out.

gen/statements.cpp
kinke: > I think that to fix this, one LOG_SCOPE has to be added where it is now.

Yes.

> And another outside (and after) the block that deals with constant expressions (the if (LLConstant ...).

I wouldn't use another one. This seems perfectly fine to me:
```
IfStatement::toIR(): onlineapp.d(2)
* constant {true/false} condition, skipping {then/else} block without labels
* ScopeStatement::toIR(): onlineapp.d(3)
* * ExpStatement::toIR(): onlineapp.d(3)
```

gen/statements.cpp
kinke: This is luckily C++, so `bool foundLabel() { return stop; }`.

gen/statements.cpp
baziotis: Sorry, there's something I'm missing. Before this PR, the code looked like this:
```
// emit block start
if (stmt->ifbody) {
  // emit block start
  ...
  // emit block end
} <- same for else body without the PGO counter increment.
// emit block end
```
Which made sense, since each block start had a block end. With this PR, there is
```
// emit block start (1)
// do something for constant expr
// .. other statements
// emit block end (1)
```
The `do something for constant expr` returns, so inside it, there should be an `emit block end` for the `block start` numbered (1). That is the role of the block end you pointed (as I was thinking it). If we add another block start, as you proposed, shouldn't we add another block end ?

kinke: Oh sorry, my bad, I totally missed that outer block for the whole if-statement (needed because of `if (auto var = ...)`?) and that you're ending that one. So I'd recommend adding another block for optional PGO counter increment + `executed` block.

Btw: `Statement *temp = executed; executed = skipped; skipped = temp;` is `std::swap(executed, skipped);`.

gen/recursivevisitor.h
baziotis: Yeah, it makes sense. Ok, I'll change it and see how it comes out.

gen/statements.cpp
baziotis: > (needed because of `if (auto var = ...)`?)

If that question is directed to me, I've no idea. :P I guess yes, but it was just there.

> Btw: `Statement *temp = executed; executed = skipped; skipped = temp;` is `std::swap(executed, skipped);`.

Thanks, too much C again.

gen/statements.cpp
```diff
@@ -314,10 +382,33 @@ class ToIRVisitor : public Visitor {
...
+      if (const_val->isZeroValue()) {
+        std::swap(executed, skipped);
+      }
+      if (!containsLabel(skipped)) {
+        IF_LOG Logger::println("Constant true/false condition - elide.");
+        if (executed) {
+          irs->DBuilder.EmitBlockStart(executed->loc);
+        }
+        // True condition, the branch is taken so emit counter increment.
+        if (!const_val->isZeroValue()) {
```
baziotis: Can it ever happen to have `!zeroValue && !executed` ? Basically, this boils down to can ever the `stmt->ifbody` be null ?

kinke: I don't think so, but it's probably better to assume the front-end might spit out a `null`.

baziotis: Good, just double-checking.

baziotis:
> Conflicting files
> tests/d2/dmd-testsuite

Oh really ? What now ? :P

kinke:
Yeah sry about that :] - you'll have to rebase this and your dmd-testsuite branch again. As a bonus, the unrelated Win64 Azure failures will be fixed.

baziotis:
> Yeah sry about that :] - you'll have to rebase this and your dmd-testsuite branch again. As a bonus, the unrelated Win64 Azure failures will be fixed.

Ok ok, thanks. :)

baziotis:
I think that something's wrong with the 2 failed tests.

kinke:
Yeah don't worry about those 2 CI failures for now, they generally fail at the moment, so this PR is virtually green now.

baziotis:
Great!

JohanEngelen:
@baziotis Rather than merging master into your branch, you should rebase the changes onto master. That way history stays much cleaner.

baziotis:
> @baziotis Rather than merging master into your branch, you should rebase the changes onto master. That way history stays much cleaner.

Could you specify where I did the opposite ? I might have missed it, AFAIK, I did what you propose.
Note: The last merge of master into my branch (the branch of this PR) was not done by me.

kinke:
I did, because I also created a corresponding dmd-testsuite branch in the official repo (rebased and both previous commits squashed into one), and that doesn't play with nice with multiple commits in here touching dmd-testsuite. I was going to squash these 30 commits anyway - are there other preferences?

kinke:
@JohanEngelen: Do you prefer it unsquashed then? Otherwise I'll squash-merge later this evening and release the first 1.19 beta.

JohanEngelen:
@kinke sry for late response, I was out traveling. Indeed, with squash merge it doesn't matter

PR #3143, opened 2019-08-24
Upgrade front-end & libs to v2.088.0

kinke:


kinke:
I assume we don't support representing DMD's new `-fPIE` with a combination of existing cmdline options; I played around with `-relocation-model`, but wasn't able to replicate clang's asm with `-fPIE` (https://godbolt.org/z/6Sibvr). clang's cmdline options logic in this regard is a [nightmare](https://github.com/llvm/llvm-project/blob/2b3d49b610bd2a45884115edcb21110bfa325f51/clang/lib/Driver/ToolChains/CommonArgs.cpp#L871-L1082)...

The Win32 unittest failure will be fixed by https://github.com/dlang/phobos/pull/7156.

JohanEngelen:
nice that you included a link to Clang's PIC/PIE logic, oof...  I'm OK with not supporting `-fPIE` for now. Open an issue for someone to workout what the cmdline options should be to get same asm output as clang?

PR #3144, opened 2019-08-28
Fix llvm 10

thewilsonator:
This fixes some compilation errors with LLVM 10. 

Not fixed is the issue with change of CallSite to CallBase that changed with IRBuilder::CreateMemSet

### DIFF
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -33,6 +33,12 @@ namespace llvm {
 }
 #endif
 
+#if LDC_LLVM_VER >= 1000
+using std::make_unique;
+#else
+using llvm::make_unique;
+#endif
+
 namespace {
 
 std::unique_ptr<llvm::ToolOutputFile>
@@ -52,7 +58,7 @@ createAndSetDiagnosticsOutputFile(IRState &irs, llvm::LLVMContext &ctx,
     }
 
     std::error_code EC;
-    diagnosticsOutputFile = llvm::make_unique<llvm::ToolOutputFile>(
+    diagnosticsOutputFile = make_unique<llvm::ToolOutputFile>(
         diagnosticsFilename, EC, llvm::sys::fs::F_None);
     if (EC) {
       irs.dmodule->error("Could not create file %s: %s",
@@ -60,9 +66,11 @@ createAndSetDiagnosticsOutputFile(IRState &irs, llvm::LLVMContext &ctx,
       fatal();
     }
 
+    //FIXME: this has been replaced by setDiagnosticHandler
+#if LDC_LLVM_VER < 1000
     ctx.setDiagnosticsOutputFile(
-        llvm::make_unique<llvm::yaml::Output>(diagnosticsOutputFile->os()));
-
+        make_unique<llvm::yaml::Output>(diagnosticsOutputFile->os()));
+#endif
     // If there is instrumentation data available, also output function hotness
     if (opts::isUsingPGOProfile()) {
 #if LDC_LLVM_VER >= 500
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -29,6 +29,12 @@
 #include "lld/Common/Driver.h"
 #endif
 
+#if LDC_LLVM_VER >= 1000
+using std::make_unique;
+#else
+using llvm::make_unique;
+#endif
+
 //////////////////////////////////////////////////////////////////////////////
 
 static llvm::cl::opt<std::string>
@@ -759,11 +765,11 @@ int linkObjToBinaryGcc(llvm::StringRef outputPath,
 #if LDC_LLVM_VER >= 500
   if (global.params.targetTriple->isOSBinFormatWasm()) {
     tool = getProgram("wasm-ld", &opts::linker);
-    argsBuilder = llvm::make_unique<LdArgsBuilder>();
+    argsBuilder = make_unique<LdArgsBuilder>();
   } else {
 #endif
     tool = getGcc();
-    argsBuilder = llvm::make_unique<ArgsBuilder>();
+    argsBuilder = make_unique<ArgsBuilder>();
 #if LDC_LLVM_VER >= 500
   }
 #endif
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -554,7 +554,7 @@ void loadInstrProfileData(IRState *irs) {
 #if LDC_LLVM_VER >= 1000
         irs->PGOReader->getSummary(/*context sensitive profile=*/ false)
           .getMD(irs->context()),
-        llvm::ProfileSummary::PSK_Instr);
+        llvm::ProfileSummary::PSK_Instr
 #else
         irs->PGOReader->getSummary().getMD(irs->context())
 #endif
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -544,7 +544,7 @@ void loadInstrProfileData(IRState *irs) {
 
     if (!irs->module.getProfileSummary(
 #if LDC_LLVM_VER >= 1000
-    /*non-context sensitive profile*/  false
+    /*context sensitive profile=*/  false
 #endif
                                        )) {
       // Don't reset the summary. There is only one profile data file per LDC
@@ -552,7 +552,7 @@ void loadInstrProfileData(IRState *irs) {
       // set.
       irs->module.setProfileSummary(
 #if LDC_LLVM_VER >= 1000
-        irs->PGOReader->getSummary(/*non-context sensitive profile*/ false)
+        irs->PGOReader->getSummary(/*context sensitive profile=*/ false)
           .getMD(irs->context()),
         llvm::ProfileSummary::PSK_Instr);
 #else
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -542,12 +542,13 @@ void loadInstrProfileData(IRState *irs) {
     }
     irs->PGOReader = std::move(readerOrErr.get());
 
-    if (!irs->module.getProfileSummary()) {
+    if (!irs->module.getProfileSummary(false)) {
       // Don't reset the summary. There is only one profile data file per LDC
       // invocation so the summary must be the same as the one that is already
       // set.
       irs->module.setProfileSummary(
-          irs->PGOReader->getSummary().getMD(irs->context()));
+        irs->PGOReader->getSummary(false).getMD(irs->context()),
+        llvm::ProfileSummary::PSK_Instr);
     }
   }
 }
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -544,7 +544,7 @@ void loadInstrProfileData(IRState *irs) {
 
     if (!irs->module.getProfileSummary(
 #if LDC_LLVM_VER >= 1000
-                                       false
+    /*non-context sensitive profile*/  false
 #endif
                                        )) {
       // Don't reset the summary. There is only one profile data file per LDC
@@ -552,7 +552,8 @@ void loadInstrProfileData(IRState *irs) {
       // set.
       irs->module.setProfileSummary(
 #if LDC_LLVM_VER >= 1000
-        irs->PGOReader->getSummary(false).getMD(irs->context()),
+        irs->PGOReader->getSummary(/*non-context sensitive profile*/ false)
+          .getMD(irs->context()),
         llvm::ProfileSummary::PSK_Instr);
 #else
         irs->PGOReader->getSummary().getMD(irs->context())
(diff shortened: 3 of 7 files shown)


driver/codegenerator.cpp
```diff
@@ -52,17 +58,19 @@ createAndSetDiagnosticsOutputFile(IRState &irs, llvm::LLVMContext &ctx,
...
-    diagnosticsOutputFile = llvm::make_unique<llvm::ToolOutputFile>(
+    diagnosticsOutputFile = make_unique<llvm::ToolOutputFile>(
         diagnosticsFilename, EC, llvm::sys::fs::F_None);
     if (EC) {
       irs.dmodule->error("Could not create file %s: %s",
                          diagnosticsFilename.c_str(), EC.message().c_str());
       fatal();
     }
 
+    //FIXME: this has been replaced by setDiagnosticHandler
```
thewilsonator: N.B. I'm not sure how tot get back the original behaviour.

gen/modules.cpp
```diff
@@ -542,12 +542,13 @@ void loadInstrProfileData(IRState *irs) {
...
 
-    if (!irs->module.getProfileSummary()) {
+    if (!irs->module.getProfileSummary(false)) {
       // Don't reset the summary. There is only one profile data file per LDC
       // invocation so the summary must be the same as the one that is already
       // set.
       irs->module.setProfileSummary(
-          irs->PGOReader->getSummary().getMD(irs->context()));
+        irs->PGOReader->getSummary(false).getMD(irs->context()),
+        llvm::ProfileSummary::PSK_Instr);
```
thewilsonator: We'll see what the CI thinks of this.

thewilsonator:
@Robertorosmaninho if you are trying to build with the latest LLVM you'll need to use this patch. You'll also need to comment out [this line](https://github.com/ldc-developers/ldc/blob/master/gen/passes/GarbageCollect2Stack.cpp#L455) as I haven't managed to work out a solution for this `CallSite` use.

gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -79,11 +79,22 @@ void EmitMemSet(IRBuilder<> &B, Value *Dst, Value *Val, Value *Len,
                 const Analysis &A) {
   Dst = B.CreateBitCast(Dst, PointerType::getUnqual(B.getInt8Ty()));
 
-  CallSite CS =
+#if LDC_LLVM_VER >= 1000
+  auto*
+#else
+  CallSite
+#endif
+    CS =
```
Hardcode84: Why not just `auto` intead of ifdefs?

thewilsonator: because `auto` is never a pointer.

Hardcode84: `auto` can be inferred into pointer in C++. It is advised to be avoided for code style reasons, but in this particular case it will be better than ifdefs IMO.

Hardcode84:
I don't like `make_unique` ifdefs copypaste everywhere but I don't see a good solution here. We can write our `make_unique` wrapper with ifdefs but it will be bigger than actual `make_unique` implementation :)

thewilsonator:
Neither do I (I wanted to put this up sooner so that @Robertorosmaninho can have a functioning LDC to start his work on). 

I could have sworn we used to have an llvm compatibility module somewhere but I couldn't find it, otherwise I'd have put it in there. I think we are going to have to have something like that for the CallSite changes anyway, those look pretty nasty.

gen/modules.cpp
```diff
@@ -542,12 +542,22 @@ void loadInstrProfileData(IRState *irs) {
     }
     irs->PGOReader = std::move(readerOrErr.get());
 
-    if (!irs->module.getProfileSummary()) {
+    if (!irs->module.getProfileSummary(
+#if LDC_LLVM_VER >= 1000
+                                       false
```
JohanEngelen: I verified that this is correct. Add something like  `/*non-context sensitive profile*/`

gen/modules.cpp
JohanEngelen: I too think that this is the correct fix.

driver/linker-gcc.cpp
```diff
@@ -29,6 +29,12 @@
 #include "lld/Common/Driver.h"
 #endif
 
+#if LDC_LLVM_VER >= 1000
+using std::make_unique;
```
JohanEngelen: shouldn't this depend on C++ version? LLVM 10 requires C++14?

gen/optimizer.cpp
```diff
@@ -290,7 +290,9 @@ static void addOptimizationPasses(legacy::PassManagerBase &mpm,
     builder.Inliner = createAlwaysInlinerPass();
 #endif
   }
+#if LDC_LLVM_VER <= 1000
   builder.DisableUnitAtATime = !unitAtATime;
```
JohanEngelen: was this replaced by anything else? The whole cmdline option should be removed aswell. Why the `<=` instead of `<`?

gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -79,11 +79,16 @@ void EmitMemSet(IRBuilder<> &B, Value *Dst, Value *Val, Value *Len,
...
 
-  CallSite CS =
-      B.CreateMemSet(Dst, Val, Len, 1 /*Align*/, false /*isVolatile*/);
+  auto CS = B.CreateMemSet(Dst, Val, Len, 1 /*Align*/, false /*isVolatile*/);
   if (A.CGNode) {
     A.CGNode->addCalledFunction(
-        CS, A.CG->getOrInsertFunction(CS.getCalledFunction()));
+        CS, A.CG->getOrInsertFunction(
+#if LDC_LLVM_VER >= 1000
+        CS.getCalledFunction()
```
JohanEngelen: perhaps add a temp variable for this parameter, to remove some ugliness

gen/pgo_ASTbased.cpp
```diff
@@ -34,6 +34,11 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/MD5.h"
 
+#if LDC_LLVM_VER >= 1000
+using std::make_unique;
```
JohanEngelen: let's move this repeated piece into a helper header file?

driver/codegenerator.cpp
JohanEngelen: do you have a link to the commit that removed `setDiagnosticsOutputFile` ?

JohanEngelen:
@thewilsonator Can you track down which LLVM commit / phabricator removed `removeCallEdgeFor`?

driver/codegenerator.cpp
thewilsonator: No, that was purely speculative as it appears that `setDiagnosticsOutputFile` no longer exists and `setDiagnosticHandler` is the only thing there that could logically replace its functionality.

driver/linker-gcc.cpp
thewilsonator: Possibly, see http://lists.llvm.org/pipermail/llvm-dev/2019-April/131486.html

gen/optimizer.cpp
thewilsonator: > was this replaced by anything else?

The LLVM c-api implementation for the functions to get/set this, just say that it was removed.

> The whole cmdline option should be removed aswell.

+1

> Why the <= instead of <?

Because I'm an idiot.

gen/pgo_ASTbased.cpp
thewilsonator: I swear we used to have an LLMV compatibility file somewhere, did it get removed? Because I can't find it.

gen/passes/GarbageCollect2Stack.cpp
thewilsonator: OK. (I love the fact the in D you don't need to worry about this)

gen/optimizer.cpp
thewilsonator: I've versioned it out, should we just remove it completely?

gen/modules.cpp
thewilsonator: Done.

gen/passes/GarbageCollect2Stack.cpp
thewilsonator: Done.

gen/optimizer.cpp
JohanEngelen: hope you can find the commit that removed it, so we see whether there was a replacement for it.

gen/modules.cpp
```diff
@@ -544,15 +544,16 @@ void loadInstrProfileData(IRState *irs) {
 
     if (!irs->module.getProfileSummary(
 #if LDC_LLVM_VER >= 1000
-                                       false
+    /*non-context sensitive profile*/  false
```
JohanEngelen: hm, now I'm not sure if this makes it more clear. The `false` indicates **non**context sensitive, but now it reads as if the call is to get the **yes**context sensitive profile, no? hmm.

gen/optimizer.cpp
thewilsonator: https://github.com/llvm-mirror/llvm/commit/6df72b9a11bd6b73192ac716978da8920e80a8c0
TL;DR No, nobody used it it didn't quite do what it was supposed to anyway and nobody noticed.

gen/modules.cpp
thewilsonator: fixed.

gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -443,16 +446,19 @@ static void RemoveCall(CallSite CS, const Analysis &A) {
...
+  Instruction *inst = CS.getInstruction();
+  if (InvokeInst *Invoke = cast<InvokeInst>(inst)) {
     BranchInst::Create(Invoke->getNormalDest(), Invoke);
     Invoke->getUnwindDest()->removePredecessor(CS->getParent());
   }
 
   // Remove the runtime call.
   if (A.CGNode) {
+#if LDC_LLVM_VER >= 1000
+    A.CGNode->removeCallEdgeFor(cast<CallBase>(*inst));
```
thewilsonator: Is this correct for `removeCallEdgeFor(CallBase& cb)` with CallBase a subclass of Instruction? 
My C++fu is not as good as it once was and this was the first thing that compiled without a blizzard of template errors.

Robertorosmaninho: I used a dyn_cast to fix this problem, but this should work as well.

thewilsonator:
@Robertorosmaninho you'll also need to apply [druntime.txt](https://github.com/ldc-developers/ldc/files/3558749/druntime.txt) to druntime to get the runtime to build.

thewilsonator:
Any guesses as to why MSVC think that the call to `make_unique` is ambiguous? are we `using` std, or llvm somewhere in the includes?

kinke:
> Any guesses as to why MSVC think that the call to make_unique is ambiguous? are we using std, or llvm somewhere in the includes?

Nope, no relevant `using namespace std` or `using std::make_unique` here. The problem is apparently [argument-dependent lookup](https://en.cppreference.com/w/cpp/language/adl), i.e., the `std::make_unique` is found automatically because we pass a `std::error_code` argument (*and* now use an unqualified `make_unique` name), so that the `std` namespace is searched too (and MSVC has had `std::make_unique` for ages).
Don't blame MSVC :] - this is clang v8.0.1. ;)

JohanEngelen:
my guess is that we enable c++11 (not 14), which does not have `std::make_unique` and ADL doesn't trigger. Clang being incorrect here doesn't sound likely to me.

kinke:
MS `cl` (from VS 2019) behaves identically, so the `clang-cl` wrapper is correctly compatible here. The MS compiler doesn't support C++11 anymore: `/std:<c++14|c++17|c++latest>`

So this apparently just shows that this PR can currently only be compiled as C++11 and not with any later version (with LLVM < 10).

kinke:
Microsoft is obviously a PITA and chose to use `199711` for predefined `__cplusplus`. So for MSVC, one should check `_MSVC_LANG` instead.

driver/linker-gcc.cpp
JohanEngelen: So as kinke already mentioned, this breaks with c++14 compilation for LLVM < 10. So change the condition to cplusplus version, instead of LLVM version.

driver/codegenerator.cpp
kinke: Here: https://github.com/llvm-mirror/llvm/commit/78ee34a412ec94af550d0881875ce437bab56b9f#diff-b1ec65e53a672d9db72a97451556fe87

Adapted in #3166.

kinke:
This has been incorporated as #3166 and #3187, right?

JohanEngelen:
yeah

Issue #392, opened 2013-06-10
Generate 'inbounds' GEPs where possible


dnadlinger:
We should add the `inbounds` attribute to compiler-generated `getelementptr` instructions where possible.

kinke:
Resolved by #1232 and #3148.

PR #3148, opened 2019-08-31
Add inbounds where possible

baziotis:
Issue: https://github.com/ldc-developers/ldc/issues/392

I don't get the reasoning behind how arrays are handled currently so I didn't change them.

For static arrays and constant index, the generated IR is the expected. i.e. if it's in bounds, then `GEP inbounds` is generated, otherwise compile-time error.
If though say we have a static array and the index is not known (in which case bounds-checking code is generated) e.g.
```
int foo(int[3] a, int i)
{
    return a[i];
}
```
then `inbounds` is _not_ generated for the `bounds.ok` BB. I could "just" change that, but [in the code](https://github.com/baziotis/ldc/blob/73e6f4442744603c55128d8e5ab35667e3d52768/gen/toir.cpp#L1094), it seems as if someone has thought about it (i.e. there is not just a `true` or a `false`).
However, the way I think of it, there are 2 cases:
1) The index is constant, in which case either:
     a) It is out of bounds and it will be caught by the front-end.
     b) The condition is in bounds, in which case `e->indexIsInBounds` is `true`.
2) The index is not constant, in which case if it is in bounds, we should generate `inbounds`.
If it isn't we probably (more on that below) should generate it, but it won't reach that code anway.

Considering the cases, it seems that it should just be `true` there. Same for dynamic arrays.

A note on Clang: [Clang generates](https://godbolt.org/z/IVibHT) `inbouds` even when we're accessing an array of known size _out_ of bounds. [According to the manual](https://llvm.org/docs/LangRef.html#id231) this seems better (it generates a poison value instead of just undefined behavior).

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -65,8 +65,8 @@ static LLValue *DtoSlicePtr(Expression *e) {
   LLValue *val = isStaticArray ? DtoLVal(dval) : makeLValue(loc, dval);
   LLValue *array = DtoRawAlloca(type, 0, ".array");
   LLValue *len = isStaticArray ? DtoArrayLen(dval) : DtoConstSize_t(1);
-  DtoStore(len, DtoGEPi(array, 0, 0));
-  DtoStore(DtoBitCast(val, getVoidPtrType()), DtoGEPi(array, 0, 1));
+  DtoStore(len, DtoGEP(array, 0u, 0));
+  DtoStore(DtoBitCast(val, getVoidPtrType()), DtoGEP(array, 0, 1));
   return array;
 }
 
@@ -160,7 +160,7 @@ static void DtoArrayInit(Loc &loc, LLValue *ptr, LLValue *length,
   LLValue *itr_val = DtoLoad(itr);
   // assign array element value
   DLValue arrayelem(elementValue->type->toBasetype(),
-                    DtoGEP1(ptr, itr_val, true, "arrayinit.arrayelem"));
+                    DtoGEP1(ptr, itr_val, "arrayinit.arrayelem"));
   DtoAssign(loc, &arrayelem, elementValue, TOKblit);
 
   // increment iterator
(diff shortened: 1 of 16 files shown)


kinke:
> it seems as if someone has thought about it

Yes, most GEPs got `inbounds` with https://github.com/ldc-developers/ldc/commit/956061c3c70f2057dc9e4068ecd148636c3a89e5.

> but it won't reach that code anway

The bounds checks can be easily disabled (e.g., only enabled in `@safe` functions with `-release`, plus there's `-boundscheck=off`).

> According to the manual this seems better

I'm wondering when one is supposed NOT to use `inbounds` then.

dnadlinger:
> I'm wondering when one is supposed NOT to use inbounds then.

If things were designed from scratch, I think one would make `inbounds` the default, and have an extra flag for indicating arbitrary address arithmetic that doesn't care about staying within a single object (in the C memory model sense, i.e. allocation) – which is the only reason not to use `inbounds`.

baziotis:
> > it seems as if someone has thought about it
> 
> Yes, most GEPs got `inbounds` with [956061c](https://github.com/ldc-developers/ldc/commit/956061c3c70f2057dc9e4068ecd148636c3a89e5).

I meant that in this case, it did _not_ get `inbounds`.

> I'm wondering when one is supposed NOT to use `inbounds` then.

Same here and I agree with David:

> If things were designed from scratch, I think one would make inbounds the default

kinke:
Could you please add `inbounds` when indexing any array with any index then?

baziotis:
> Could you please add `inbounds` when indexing any array with any index then?

Done. However, from a quick look, it might now not make sense to pass it as argument since in most cases, it is `true`. Maybe we should create a separate function named differently.

kinke:
> However, from a quick look, it might now not make sense to pass it as argument since in most cases, it is true.

Indeed, there's only one remaining call site which might not use `inbounds` (when slicing). I'll refactor this.

baziotis:
> > However, from a quick look, it might now not make sense to pass it as argument since in most cases, it is true.
> 
> Indeed, there's only one remaining call site which might not use `inbounds` (when slicing). I'll refactor this.

Thanks!

Issue #3079, opened 2019-05-29
CMake Error: Some (but not all) targets in this export set were already defined.


Robertorosmaninho:
Hi all!

I just compiled the llvm 8.0.0 and I'm trying to get the ldc, but found this problem and I can't figure out how to solve this. I tried to compile on MacOS 10.12 and MacOS Mojave (same LLVM version).

**I've tried compiling it with the following commands:**

git clone --recursive https://github.com/ldc-developers/ldc.git
cd ldc
mkdir build && cd build
export DMD=/usr/local/Cellar/dmd/2.086.0/bin/dmd
cmake -G Ninja -DLLVM_CONFIG=../../llvm/build/bin/llvm-config ..

Available in: https://wiki.dlang.org/Building_LDC_from_source

**Then I get the following result:**

-- The C compiler identification is AppleClang 10.0.1.10010046
-- The CXX compiler identification is AppleClang 10.0.1.10010046
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found host D compiler /usr/local/Cellar/dmd/2.086.0/bin/dmd, with default flags ''
-- Host D compiler ID: DigitalMars
-- Host D compiler version: DMD64 D Compiler v2.086.0
-- Host D compiler front-end version: 2086
-- Found LLVM: /Users/robertorosmaninho/llvm/build (found suitable version "8.0.0", minimum required is "3.9")
-- Found PkgConfig: /usr/local/bin/pkg-config (found version "0.29.2")
-- Checking for module 'LLVMSPIRVLib'
--   No package 'LLVMSPIRVLib' found
-- Performing Test SUPPORTS_FVISIBILITY_INLINES_HIDDEN_FLAG
-- Performing Test SUPPORTS_FVISIBILITY_INLINES_HIDDEN_FLAG - Success
[...]

PR #3151, opened 2019-09-03
Fix issue #3079 : Some (but not all) targets in this export set were …

thewilsonator:
…already defined


kinke:
So if these targets are part of the LLVM CMake set (and AFAICT, the LLVM CMake files are only included [here](https://github.com/ldc-developers/ldc/blob/a9b9097918ae209d0a35cf5dc2c2b342de51fb83/runtime/jit-rt/DefineBuildJitRT.cmake#L42-L47), i.e., only if building without `LDC_DYNAMIC_COMPILE=False`), the tests would depend on `FileCheck` and `not` being found in PATH.

kinke:
Still a huge improvement over CMake failing.

PR #3152, opened 2019-09-03
Travis lit tests use wrong jit lib (from host compiler)

Hardcode84:
lit tests somehow get linked with host compiler `jit-rt.so` instead of correct one.


Hardcode84:
Deleting `jit-rt.so` from host compiler seems to help. But there can be similar issues with other runtime libs I think. Any ideas how to do this in less hackish way?

ping @kinke

kinke:
Hmm. Are you sure you're always invoking the built compiler? I've just recently found some lit-tests which were using `ldc2 ...` instead of `%ldc ...`, which might invoke the host compiler.
Otherwise, a wrong rpath seems like the most probable reason...

Hardcode84:
jit tests use %ldc

kinke:
Travis uses the [dlang install script](https://dlang.org/install.sh), and that tampers with `LD_LIBRARY_PATH`, so I guess that's the reason.

Hardcode84:
unsetting `LD_LIBRARY_PATH` helped. Thanks

kinke:
> But there can be similar issues with other runtime libs I think.

I guess it should have affected shared druntime and Phobos too, but then I would have expected some failures...

.travis.yml
```diff
@@ -85,6 +85,7 @@ install:
   - python -c "import lit; lit.main();" --version | head -n 1
 
 script:
+  - unset LD_LIBRARY_PATH
```
JohanEngelen: should this be part of the the lit setup python script?

Hardcode84: Do you mean `lit.site.cfg.in`? This hack is too specific for single CI system and I don't want to do this for everyone nor to make complex checks in lit scripts.

ZombineDev:
> Travis uses the dlang install script, and that tampers with LD_LIBRARY_PATH, so I guess that's the reason.

When we built dmd, in most CIs we source the install script, but after dmd is built we call `deactivate` to reset the environment to its previous state, to avoid cases where dmd, dub or libs from the install script's folder take precedence over those that we just built.

Note that due to the way Travis sets up the environment, the `deactivate` may not be defined (IIRC, I recently had such issues with CircleCI). The most robust way may be to call the install script, build LDC and `deactivate` it, all in one bash script invocation, so no environment variable changes remain.

.travis.yml
JohanEngelen: > Do you mean `lit.site.cfg.in`? 

Yes. If we unset it in there, then none of our tests will depend on it which is a good thing. The same issue on the CI system may happen on a user's system too. The lit tests are run in their own environment, so changing it in there does not change it for the user.

Hardcode84: I can add a check which scans `LD_LIBRARY_PATH` for suspicions files and reports them to stdout but changing `LD_LIBRARY_PATH` from lit scripts is still a bad idea IMO.

Hardcode84: Even so I agree on general criticism of `LD_LIBRARY_PATH` let's not do anything funny in our scripts. Randomly ignoring user env variables because we think this is "right thing to do" is bad and can break use cases we are not aware right now. This PR was about simple configuration issue on one test system, let's not try to solve world problems here we were not even aware a week ago.

I added a warning to detect such issues in the future, this should be enough for now IMO.

JohanEngelen: so but if someone uses `install.sh` on his machine to get DMD, then builds LDC, the tests will fail, right?

JohanEngelen: anyway, perhaps it's good that the tests will fail in that case, because programs with ldc-jit won't work well either then :-)

PR #3158, opened 2019-09-08
druntime: Include core.stdcpp.* modules

### LINKED ISSUE #3103: Don't exclude core.stdcpp.* from druntime library
```d
import core.stdcpp.string_view;

void main()
{
    auto s = string_view("hey");
}
```

----

```
onlineapp.d:5: error: undefined reference to '_D4core6stdcpp11string_view3std__T17basic_string_viewTaTSQCdQCbQBxQBn__T11char_traitsTaZQqZQCf6__ctorMFNaNbNcNiNeAxaZSQEmQEkQEgQDw__TQDvTaTQDfZQEf'
```

kinke:
Fixes #3103.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,7 +110,7 @@ include(GetLinuxDistribution)
 #
 
 # Version information
-set(LDC_VERSION "1.21.0") # May be overridden by git hash tag
+set(LDC_VERSION "1.22.0") # May be overridden by git hash tag
 set(DMDFE_MAJOR_VERSION   2)
 set(DMDFE_MINOR_VERSION   0)
 set(DMDFE_PATCH_VERSION   92)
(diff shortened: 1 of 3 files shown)


kinke:
How do you feel about the lib[std]c++ dependency? The upstream PR is https://github.com/dlang/druntime/pull/2788.

kinke:
Got back to this and reworked it to avoid C++ runtime dependencies. A few non-virtual functions are still declared only, but the D vtable should now be populatable and the D implementations shouldn't conflict with the C++ ones (`extern(D)` mangling) anymore.

Note that populating the D vtable with D implementations only works here because the virtual function signatures are trivial and should be ABI-compatible across C++ and D. An `extern(D)` virtual function with 2 params would already suffice to break this, as D callers would call the C++ implementation (if the vptr points to the C++ vtable, i.e., if the instance has been created on the C++ side) with reversed args.

kinke:
Reworked with `@weak extern(C++)` for the 2 problematic `core.stdcpp.{exception,typeinfo}` modules, seems to work fine.

I've also ported druntime's stdcpp integration tests on Windows from DigitalMars make to GNU make, so that they are tested with LDC too. They take about 37 seconds on my box (not parallelized) and cover more modules than Posix hosts (e.g., `core.stdcpp.vector` is currently only supported for MSVC targets) and also test all 4 MSVC runtime flavors (static/dynamic, release/debug).

PR #3160, opened 2019-09-10
druntime: Add {load,store}Unaligned and prefetch to core.simd

kinke:
This tests a revised https://github.com/ldc-developers/druntime/pull/163.


kinke:
Pinging @baziotis.

baziotis:
It took me 2 minutes to find the changes. :P
Indeed, it makes sense to keep for LDC all the unittests since those will be relevant for the DMD "overload". But we don't need to add tests for any other vectors, since those are supposed to be handled by what was already there (the LDC overload).

kinke:
Yes, and the 128-bit vectors should suffice, as it's the same thing over and over.

I had to add an explicit function wrapping the IR implementation, just aliasing wouldn't work. The `ElementType` alias wouldn't work either.

baziotis:
> I had to add an explicit function wrapping the IR implementation, just aliasing wouldn't work. The `ElementType` alias wouldn't work either.

My bad. I'm still struggling to see why.

kinke:
I think it's something along these lines:

* Compiler sees `float4* p; loadUnaligned(p);`.
* It has seen eponymous template `loadUnaligned(V)` with an inner `loadUnaligned` alias and a `loadUnaligned(const V*)` function. In order to expand the alias and check whether it'd be a better fit (e.g., mutable `loadUnaligned(V*)`), the template would have to be instantiated already, as the IR function type depends on `ElementType`, `llvmT` enum etc. - but the compiler is still checking which of the 2 overloads for which `V` template parameter fits best, in order to infer `V` and only then instantiating the template.

baziotis:
> I think it's something along these lines:
> 
> * Compiler sees `float4* p; loadUnaligned(p);`.
> * It has seen eponymous template `loadUnaligned(V)` with an inner `loadUnaligned` alias and a `loadUnaligned(const V*)` function. In order to expand the alias and check whether it'd be a better fit (e.g., mutable `loadUnaligned(float4*)`), the template would have to be instantiated already, as the IR function type depends on `ElementType`, `llvmT` enum etc. - but the compiler is still checking which of the 2 overloads for which `V` template parameter fits best, in order to infer `V` and only then instantiating the template.

Subtle. Thanks!

PR #3164, opened 2019-09-16
Azure CI: Add Android cross-compilation job (incl. prebuilt release package)

kinke:


kinke:
Yay, Microsoft apparently managed to fuck up Azure badly, appending a single quote to variables set in bash scripts (also an issue for LLVM CI). Argh.

kinke:
Ready for comments; will clean it up a bit sometime.

tools/CMakeLists.txt
```diff
@@ -1,61 +1,40 @@
...
 
-# Set tool names
+#############################################################################
+# Build ldc-prune-cache
 set(LDCPRUNECACHE_EXE ldc-prune-cache)
 set(LDCPRUNECACHE_EXE ${LDCPRUNECACHE_EXE} PARENT_SCOPE) # needed for correctly populating lit.site.cfg.in
 set(LDCPRUNECACHE_EXE_NAME ${PROGRAM_PREFIX}${LDCPRUNECACHE_EXE}${PROGRAM_SUFFIX})
 set(LDCPRUNECACHE_EXE_FULL ${PROJECT_BINARY_DIR}/bin/${LDCPRUNECACHE_EXE_NAME}${CMAKE_EXECUTABLE_SUFFIX})
-
-function(build_d_tool output_exe compiler_args linker_args compile_deps link_deps)
```
kinke: The tools now use `build_d_executable` - biggest change: includes LLVM linker flags. I checked the Linux x86_64 `ldc-{profdata,prune-cache}` executables, and their sizes are absolutely identical to master's.

As to why - it's more convenient to use the CMake C++ compiler for linking all D executables (well, if it's gcc or clang), so that one doesn't have to set up both C++ and D compilers for cross-linking.

utils/CMakeLists.txt
```diff
@@ -1,29 +1,32 @@
...
 # GCC-style intrinsics.
-
-# The LLVM_INCLUDE_DIR definition is not always set, e.g. on Windows.
-# strip off anything but the first path
-string(REGEX REPLACE ";.*$" "" LLVM_INC_DIR "${LLVM_INCLUDE_DIRS}")
-find_path(LLVM_INTRINSIC_TD_PATH "Intrinsics.td" PATHS ${LLVM_INC_DIR}/llvm ${LLVM_INC_DIR}/llvm/IR NO_DEFAULT_PATH)
-if (${LLVM_INTRINSIC_TD_PATH} STREQUAL "LLVM_INTRINSIC_TD_PATH-NOTFOUND")
-    message(SEND_ERROR "File Intrinsics.td not found")
+if(CMAKE_CROSSCOMPILING)
+    message(STATUS "NOT building gen_gccbuiltins utility when cross-compiling (you can use host LDC's ldc/gccbuiltins_*.di files)")
```
kinke: Everything is set up for cross-compilation/linking, so this util can most likely not be run on the host.

.azure-pipelines/android.yml
```diff
@@ -0,0 +1,150 @@
...
+    rm android-ndk-r20-linux-x86_64.zip
+    # Set PARALLEL_JOBS env variable for future steps
+    set +x
+    echo "##vso[task.setvariable variable=PARALLEL_JOBS]$(nproc)"
+  displayName: Install prerequisites
+- script: |
+    # DON'T use set -x => issues with task.setvariable (possibly trailing `'`)
+    set -e
+    cd ..
+    # The NDK toolchain file enforces `-g` as base C[XX] flag - remove it.
```
kinke: The difference is enormous - e.g., the `lld` executable with debug infos was ~1GB, now it's something like 35 MB.

.azure-pipelines/android.yml
```diff
@@ -0,0 +1,150 @@
...
+    TOOLCHAIN_FILE=$PWD/android-ndk-r20/build/cmake/android.toolchain.cmake
+    sed -i 's|^  -g$||' $TOOLCHAIN_FILE
+    # Extend EXTRA_CMAKE_FLAGS for future steps
+    echo "##vso[task.setvariable variable=EXTRA_CMAKE_FLAGS]$EXTRA_CMAKE_FLAGS -DCMAKE_TOOLCHAIN_FILE=$TOOLCHAIN_FILE"
```
kinke: I went with this toolchain-file approach, as that's what the Android NDK recommends when building with CMake, keeping things reasonably simple by just having to specify 2 more vars (e.g., `-DANDROID_ABI=armeabi-v7a -DANDROID_NATIVE_API_LEVEL=21`) for setting up the CMake C[++] compilers, linkers etc. I also use it for the runtime libs and don't use Joakim's Android preset.

.azure-pipelines/android.yml
```diff
@@ -0,0 +1,150 @@
...
+    installDir=$PWD/install
+    mkdir build
+    cd build
+    cmake -G Ninja $BUILD_SOURCESDIRECTORY \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DLLVM_ROOT_DIR=$PWD/../llvm \
+      -DD_COMPILER=$DMD \
+      -DCMAKE_INSTALL_PREFIX=$installDir \
+      -DINCLUDE_INSTALL_DIR=$installDir/import \
+      -DD_LINKER_ARGS="-fuse-ld=bfd;-L$PWD/../ldc-build-runtime.tmp/lib;-lphobos2-ldc;-ldruntime-ldc" \
```
kinke: `bfd` is required for the `__bss_end__` symbol, which `gold` apparently doesn't emit (could probably use `_end` instead in druntime, as for the x86 Android archs...).

.azure-pipelines/android.yml
```diff
@@ -0,0 +1,150 @@
...
+      -DLDC_NO_SPIRV=ON \
```
kinke: `LDC_NO_SPIRV` => work around linker error, which is almost certainly due to a wrong position of that lib in the linker cmdline (it's appended after all other LLVM libs; somehow seems to work for the other Posix targets...).

.azure-pipelines/android.yml
```diff
@@ -0,0 +1,150 @@
...
+      $EXTRA_CMAKE_FLAGS
+    ninja -j$PARALLEL_JOBS -v ldc2 ldmd2 ldc-build-runtime ldc-profdata ldc-prune-cache
+  displayName: Cross-compile LDC executables
+- script: |
+    set -ex
+    cd ..
+    mkdir -p install/bin
+    cp build/bin/{ldc2,ldmd2,ldc-build-runtime,ldc-profdata,ldc-prune-cache} install/bin
```
kinke: Need to do all of this manually, as the runtime libs would be compiled with the fresh cross-compiled LDC, and we cannot run that one.

tools/CMakeLists.txt
kinke: Oops, I was wrong, just re-checked: `ldc-prune-cache` size goes up from ~1MB to ~6MB.

kinke: Fixed now, incl. smaller `ldc-build-runtime` tool.

.azure-pipelines/android.yml
kinke: Obsolete; linker cmdline fixed now.

kettes-leulhetsz:
Could you please add an armv8/aarch64 build? I imagine that's what most people would actually want, and Android doesn't typically ship with multilib due to space concerns.

kinke:
That will have to wait a bit, primarily because it requires a prebuilt LLVM and thus a CI extension of our LLVM repo (and a new release). This armv7a package is basically the Andoid prototype package.

adamdruppe:
Would it be a similar process to make a x64 linux -> armv7 android cross compiler release? I'm playing with the android studio stuff and I think most useful to me would probably be linux cross compilers, perhaps even to x86 targets for the emulator.

kinke:
Nope, way simpler. For cross-compilation, any official prebuilt LDC package would do the job, no need to rebuild LDC or LLVM. All you need is
* the prebuilt Linux/Windows/whatever package,
* the precompiled Android libs from the prebuilt armv7a package,
* the Android NDK for cross-linking via clang and the included C libs, and
* a mini extension of etc/ldc2.conf, primarily specifying the path to the Android libs and the path to the NDK's clang (and as a special case for Android armv7a targets, `-linker=bfd` and `-mcpu=cortex-a8` as additional switches).

See https://wiki.dlang.org/Cross-compiling_with_LDC.

For non-armv7a Android targets, there are no precompiled druntime/Phobos libs yet, so you'll to have to compile them yourself for now, that's the only difference.

adamdruppe:
On Thu, Oct 03, 2019 at 09:34:13AM -0700, Martin Kinkelin wrote:
> Nope, way simpler. For cross-compilation, any official prebuilt LDC package would do the job, no need to rebuild LDC or LLVM. All you need is

oh cool, very nice. I'll play with this soon!

tmarplatt:
Hello, I am interested in cross-compiling to Android using ldc.

The documentation I found in the Dlang Wiki and the explanations from this thread have been in some cases contradictory, so here's what I did to make it work. Perhaps you can verify these steps and make the necessary amends to the documentation.

My setup: 
- linux x86_64 host
- installed ldc 1.18 beta2
- NDK version 19b 

1. First I built the ldc runtime libraries. In bash I exported `$CC` to point to my NDK's prebuilt `clang` path, and then ran:
`ldc-build-runtime --ninja --targetPreset=Android-arm --dFlags="-w;-mcpu=cortex-a8"`
I had to add the dFlags parameter to this command, as the [Build D for Android](https://wiki.dlang.org/Build_D_for_Android#linux) article instructs (as opposed to the one @kinke mentioned), otherwise the build process failed with LLVM errors.
2. Copied the built libraries over, and edited etc/ldc2.conf as explained in this thread and the wiki. But ldc did not recognize the target. The problem was in the section name: for some reason **ldc is failing to match** `-mtriple=armv7a--linux-androideabi` with the regex`"armv7a-.*-linux-androideabi$"`. So I renamed the section to `"armv7a-.*-linux-android"` and then it worked.
3. I cross-compiled a simple hello world program.
`ldc2 -mtriple=armv7a--linux-androideabi hello.d`
4. Uploaded the `hello` binary to my Android VM, chmodded the file on Termux and ran it. Easy.

It's worth pointing out that, since version 19, the Android NDK no longer requires `-gcc-toolchain` passed down to the prebuilt `clang`, as `--target` is enough and does the job. Earlier versions will require additional paremeters when cross-compiling, as [detailed in the wiki](https://wiki.dlang.org/Build_D_for_Android#linux).

I also tried running a cross-compiled Sieve of Eratosthenes example [found in the wiki](https://wiki.dlang.org/Build_D_for_Android#Build_a_command-line_executable) but it gets stuck in an endless loop, and after debugging I found out the following line is to blame:
`flags[] = true;`
This bug is possibly unrelated to ldc (e.g. I haven't yet tried to run the binary in my Android phone), but I thought I should mention it. :)

kinke:
The dedicated Android cross-compiling Wiki pages will have to be updated at some point, as the process is fairly straight-forward with recent Android NDKs, so that I'd like to get rid of the 'target preset' and incorporate the proper instructions in the regular cross-compilation Wiki pages (specifically, the page about ldc-build-runtime).

@tmarplatt: Did you have to build the runtime libs yourself, i.e., did the prebuilt ones not work due to your older NDK? Wrt. the triple: LLVM tampers with the triple; the final one can be looked up via `-v` in the 3rd output line (`config`, the triple in parentheses). So `armv7a--linux-androideabi` is transformed to `armv7a-unknown-linux-android`; the Wiki page has been fixed.

adamdruppe:
On Mon, Oct 14, 2019 at 03:22:01AM -0700, Martin Kinkelin wrote:
> The dedicated Android cross-compiling Wiki pages will have to be updated at some point, as the process is fairly straight-forward with recent Android NDKs

I was working on this last week, haven't updated any pages but did
manage to get it working here. It is indeed a lot easier than it
used to be.

I might have a couple small PRs for ldc-build-druntime too

kinke:
I've upgraded the [ldc-build-runtime Wiki page](https://wiki.dlang.org/Building_LDC_runtime_libraries); I'd  be grateful if someone else could be bothered to update/simplify the setup section of the [Android page](https://wiki.dlang.org/Build_D_for_Android).

tmarplatt:
> @tmarplatt: Did you have to build the runtime libs yourself, i.e., did the prebuilt ones not work due to your older NDK?

How can I invoke the prebuilt runtime libs? I'm trying to CC with the ldc 1.18 release and it doesn't work out of the box, not even after editing ldc.conf (without configuring libs path), so I don't know what else to do but build the runtime libraries myself.

tmarplatt:
I just realized my toolchain relies on an older NDK, which is why I suppose I'll continue using my custom built runtime libs.

tmarplatt:
@kinke so I've figured out how to properly cross-compile using the prebuilt runtime libs. I built the sieve example just fine and calling ldc2 only once, which is superb, so congrats.

You should know that running `dub build --arch=armv7a--linux-androideabi` shows an error: `Failed to apply the selected architecture armv7a--linux-androideabi. Got ["arm"].` 

dub still builds 32-bit ARM files with that command. I tested the executable on my phone and it works.

kinke:
Yes I know, the LDC option is `-mtriple`.

tmarplatt:
@kinke when using ldc to link cross-compiled object files into a dynamic library, e.g. when running `ldc2 -mtriple=armv7a--linux-androideabi -shared -of mylib.so file1.o file2.o ...`
Does ldc2 do anything else other than calling `clang` with the appropriate parameters?

kinke:
It might link against the shared versions of druntime and Phobos instead of the static libs, but that doesn't apply to Android (yet); I think that's it.

PR #3166, opened 2019-09-20
Azure & Shippable: Upgrade LDC-LLVM to v9.0.0

kinke:


kinke:
@Hardcode84: jit-rt hits deprecations such as:
> jit_context.cpp:109:7: warning: 'LegacyRTDyldObjectLinkingLayer' is deprecated: ORCv1 layers (layers with the 'Legacy' prefix) are deprecated. Please use ORCv2 (see docs/ORCv2.rst)

Hardcode84:
> jit-rt hits deprecations such as

ok, I'll check it after merge, thanks

kinke:
@Hardcode84: Thanks - there's a [transitioning guide](http://releases.llvm.org/9.0.0/docs/ORCv2.html#transitioning-from-orcv1-to-orcv2) in the release notes.

The PGO counter linkage changes on Windows probably render our [LLVM workaround](https://github.com/ldc-developers/llvm/commit/f21cbc06eee6181b219ff66e806033b0a5858858) obsolete - will check.

kinke:
Okay, only remaining bullet point is how to handle libfuzzer's dropped support for `trace-pc-guard` (see https://reviews.llvm.org/rL352564).

JohanEngelen:
> Okay, only remaining bullet point is how to handle libfuzzer's dropped support for `trace-pc-guard` (see https://reviews.llvm.org/rL352564).

Let's not stress about this. We can simply drop it as well if Clang dropped support.

kinke:
I already dropped it (as libFuzzer would just terminate the program after printing an error). The question was more wrt. how to adapt the tests - duplicate the 4 failing ones without trace-pc-guard stuff for LLVM 9+ or just remove the related checks for all LLVM versions?

JohanEngelen:
For the trace-pc-guard stuff, this is an important Clang commit https://www.mail-archive.com/cfe-commits@lists.llvm.org/msg67181.html
So we can drop testing `-fsanitize-coverage=trace-pc-guard`, and change the code for `-fsanitize=fuzzer` to mean `sancovOpts.Inline8bitCounters = true;` instead of `sancovOpts.TracePCGuard = true;`  (and add `sancovOpts.PCTable =true;` too).

JohanEngelen:
Tested the fuzz changes locally, and fuzzing seems to work as before.

kinke:
Thx Johan!

JohanEngelen:
Ah you already fixed the test :)
I want to add a little bit of extra testing. I'm having trouble pushing to this PR, I thought one can now push to other people's branches if it is a PR?

kinke:
Yes, pushing should work, the only time I wasn't able to was when Stefanos used his default branch as PR branch.

kinke:
`sanitizers/link_fuzzer.d` apparently fails to fail with `-link-no-cpp` with LLVM 9 *and* shared druntime/Phobos (new Travis job)...

JohanEngelen:
> `sanitizers/link_fuzzer.d` apparently fails to fail with `-link-no-cpp` with LLVM 9 _and_ shared druntime/Phobos (new Travis job)...

Hmm, so that simply means that c++ libs are no longer necessary to fuzz?

kinke:
With static druntime/Phobos, the C++ lib is apparently still required, for both Linux and Mac...

JohanEngelen:
I think the test is meant to only test that `-fsanitize=fuzzer -link-no-cpp` means no C++ libs (i.e. that you can override what `-fsanitize=fuzzer` is adding to linker line). We need something like a dry-run `-v` output, without actually invoking the linker.

kinke:
Johan, you're welcome to follow-up with additional tests.

Issue #3168, opened 2019-09-22
How would you do that? Linking statically to Phobos/Druntime in a -shared library


p0nce:
Hello,

So we build shared libraries with `-shared`. 

By an improbably succession of events, we'd rather link with `libphobos2-ldc.a` and `libdruntime-ldc.a` than `libphobos2-ldc-shared.so` and `libdruntime-ldc-shared.so`. 

(Yes this is on Linux and we do know the code won't all be -fPIC)

The reason we do this is that we need the `TypeInfo` and classes stuff. 
The other problem is when using a newer LDC, the end user may not be able to install the right `libphobos-dev` package, it seems the last one for Ubuntu is 2.078.

But:
- the `-shared` flags always append `-shared` to the lib filename which doesn't allow us to override those files
- `-link-defaultlib-shared` exists but not `-link-defaultlib-static`
- `-static` also wants the C runtime to be statically linked, but this isn't what we want, we'd rather just have Phobos and Druntime be statically linked

It turns out we can do it with DMD (with `-defaultlib=libphobos2`) and that works for our use case. **How would you do that with LDC?**

kinke:
`-link-defaultlib-shared=false`.

p0nce:
Thanks!

PR #3169, opened 2019-09-22
Disambiguate that -link-defaultlib-shared can take a parameter

p0nce:
So that nobody asks the #3168 question again.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -521,17 +521,15 @@ cl::opt<bool> dynamicCompileTlsWorkaround(
     cl::Hidden);
 #endif
 
-static cl::extrahelp footer(
-    "\n"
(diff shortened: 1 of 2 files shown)


kinke:
All boolean flags can take a parameter (LLVM default); IIRC, Seb Wilzbach didn't know that either, so maybe we should add a general text somewhere (and just mention for this flag that it defaults to `true` with `-shared` and that this can be explicitly disabled).

p0nce:
The problem being that you won't stumble on the general text so it should be discoverable at the use point no?

dnadlinger:
If it's at the bottom of the `--help` output, hopefully users would stumble over it.

p0nce:
Updated.

kinke:
Slightly revised, incl. getting rid 80 columns limit (no limit for the other help output either).

PR #3187, opened 2019-10-07
Fix compilation with LLVM 10

JohanEngelen:
Replacement for #3144

CMake changes: The matching went wrong with "10.0.0" --> "10.0." (because only 5 chars). Use regexp matching instead. I also got failures with duplicate targets `not` and `FileCheck`. If they are found in LLVM's installation, there is no need to build our own version of them.


JohanEngelen:
With these changes (+druntime change) LDC builds without error on macOS 10.13.6.

driver/cl_options-llvm.cpp
```diff
@@ -90,6 +90,10 @@ TargetOptions InitTargetOptionsFromCodeGenFlags() {
   return ::InitTargetOptionsFromCodeGenFlags();
 }
 
+#if LDC_LLVM_VER >= 1000
+Optional<Reloc::Model> getRelocModelFromCMModel() { return ::getRelocModel(); }
+#endif
+
 #if LDC_LLVM_VER >= 900
```
thewilsonator: `#elif`?

JohanEngelen: No elif, `getRelocModelFromCMModel` is newly needed for llvm 10.

utils/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     target_link_libraries(gen_gccbuiltins dl)
 endif()
 
-if (TARGET FileCheck)
+if ((TARGET FileCheck) OR (EXISTS ${LLVM_ROOT_DIR}/bin/FileCheck))
```
kinke: Do we add the LLVM bin dir to PATH for the lit tests?

JohanEngelen: I thought we did, but apparently we don't (anymore?). I've added it.

JohanEngelen:
Btw, this is to fix compilation. There are 3 lit tests failing on my machine but that's for later work.

JohanEngelen:
Need to take this into account: https://github.com/ldc-developers/ldc/blob/master/tests/sanitizers/lit.local.cfg#L28-L34

JohanEngelen:
Fixed the asan_fiber_main test.

CMakeLists.txt
```diff
@@ -586,6 +586,13 @@ if(LDC_WITH_LLD)
             set(LDC_LINKERFLAG_LIST -llldCOFF -llldCore -llldDriver ${LDC_LINKERFLAG_LIST})
         endif()
     endif()
+    if(NOT (LDC_LLVM_VER LESS 1000))
+        if(MSVC)
+            list(APPEND LDC_LINKERFLAG_LIST LLVMSymbolize.lib)
+        else()
+            set(LDC_LINKERFLAG_LIST -lLLVMSymbolize ${LDC_LINKERFLAG_LIST})
+        endif()
+    endif()
```
kinke: Could this be added as an extra LLVM component for `find_package(LLVM)` and removed in `FindLLVM.cmake` for versions < 10?

kinke: Oh sry, just seen that it's apparently LLD-specific.

JohanEngelen: So should be moved up, before adding the lld libs?

kinke: I guess so (and for MSVC, the order doesn't matter).

JohanEngelen:
Will merge when builders finish successfully.

CMakeLists.txt
JohanEngelen: done

JohanEngelen:
build errors are on timeouts, so this is good to go :)

PR #3189, opened 2019-10-14
specify clang because otherwise it tries /bin/cc which may be gcc on …

adamdruppe:
…some systems and it assumes clang -target argument syntax

### DIFF
--- a/runtime/ldc-build-runtime.d.in
+++ b/runtime/ldc-build-runtime.d.in
@@ -169,6 +169,7 @@ void runCMake() {
         "-DD_FLAGS=" ~ config.dFlags.join(";"),
         "-DRT_CFLAGS=" ~ config.cFlags.join(" "),
         "-DLD_FLAGS=" ~ config.linkerFlags.join(" "),
+	"-DCMAKE_C_COMPILER=clang",
     ];
     if(config.targetPreset.matchFirst("^Android"))
         args ~= ["-DCMAKE_SYSTEM_NAME=Linux", "-DCMAKE_C_COMPILER_WORKS=True"];


kinke:
> it assumes clang -target argument syntax

Where? `ldc-build-runtime` doesn't.

I don't think defaulting to clang over CMake's `cc` default (and probably requiring an explicit `CMAKE_C_COMPILER=cc` in the cmdline to override it) is an improvement. CMake respects the `CC` env variable, and that's encouraged in https://wiki.dlang.org/Building_LDC_runtime_libraries.

If we are talking about Android targets, I'll soon extend that Wiki page with an expanded https://github.com/ldc-developers/ldc/blob/1f9ce7e613eebe8abedfe952ca4494e6b4d20629/.azure-pipelines/posix.yml#L127-L131

adamdruppe:
idk where, it just didn't work without specifying clang. but i don't really care, specifying it is easy enough anyway.

kinke:
For Android/armv7, this seems to work for me on Windows:
```
ldc-build-runtime --ninja "--dFlags=-w;-mtriple=armv7a--linux-androideabi;-mcpu=cortex-a8" "--targetSystem=Android;Linux;UNIX" CMAKE_TOOLCHAIN_FILE=C:\LDC\android-ndk-r20\build\cmake\android.toolchain.cmake ANDROID_ABI=armeabi-v7a ANDROID_NATIVE_API_LEVEL=21
```

adamdruppe:
On Tue, Oct 15, 2019 at 12:36:56PM -0700, Martin Kinkelin wrote:
> For Android/armv7, this seems to work for me on Windows:

Yeah, it is fine if you specify the stuff. I was just trying to get it
to work by default. But eh specifying is prolly better anyway once it
gets better documented.

PR #3191, opened 2019-10-15
CMake: Fix regression for make

### LINKED ISSUE #3188: building from source failed
I downloaded, git submodule init, cmake ., make and it spat this out:

[ 27%] Built target LDCShared
make[2]: *** No rule to make target 'lib/libldmd.a', needed by 'bin/ldmd2'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:126: CMakeFiles/ldmd2.dir/all] Error 2

did i miss something or is the makefile broken?

kinke:
Fixes #3188.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -644,7 +644,7 @@ else()
         "${LDC_EXE_FULL}"
         "${LDC_D_SOURCE_FILES}"
         ""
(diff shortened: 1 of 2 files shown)


kinke:
I've verified it locally; I don't want to add a CI job using `make` due to random concurrency issues (wrt. CMake custom commands as common dependencies) in the past.

PR #3202, opened 2019-10-22
POSIX: Default to cc for linking, not gcc (or clang for FreeBSD 10+)

kinke:


kinke:
Also includes some CLI improvements now.

PR #3204, opened 2019-10-23
Posix x86[_64]: Pass non-POD arguments indirectly by value, not just for extern(C++), and get rid of extra blits for temporaries

kinke:
Streamlining the D ABI with the C++ one and preventing extra blits for rvalue arguments. Non-POD arguments passed by value aren't blitted into the callee params stack any longer, but passed by reference, either directly (rvalue args) or as reference to a bitcopy on the caller stack (lvalue args).

This is also a prerequisite for a potential DIP1014 (opPostMove) implementation, as LLVM (via its `byval` attribute) has been performing the move/blit implicitly as part of the call, with no way for us to invoke the post-move hook afterwards in the callee with the address of the original argument.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -2871,6 +2871,11 @@ bool toInPlaceConstruction(DLValue *lhs, Expression *rhs) {
               Logger::println("success, in-place-constructing temporary");
               auto lhsLVal = DtoLVal(lhs);
               auto rhsLVal = DtoLVal(rhs);
+              if (!llvm::isa<llvm::AllocaInst>(rhsLVal)) {
+                error(rhs->loc, "lvalue of temporary is not an alloca, please "
+                                "file an LDC issue");
+                fatal();
+              }
               rhsLVal->replaceAllUsesWith(lhsLVal);
               return true;
             }
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -505,7 +505,6 @@ public:
 
     // Try to construct the lhs in-place.
     if (lhs->isLVal() && (e->op == TOKconstruct || e->op == TOKblit)) {
-      Logger::println("attempting in-place construction");
       if (toInPlaceConstruction(lhs->isLVal(), e->e2))
         return;
     }
@@ -524,7 +523,6 @@ public:
     if (lhs->isLVal() && (e->op == TOKassign) &&
         ((e->e2->op == TOKstructliteral) &&
          static_cast<StructLiteralExp *>(e->e2)->useStaticInit)) {
-      Logger::println("attempting in-place assignment");
       if (toInPlaceConstruction(lhs->isLVal(), e->e2))
         return;
     }
@@ -2742,9 +2740,34 @@ bool basetypesAreEqualWithoutModifiers(Type *l, Type *r) {
   r = stripModifiers(r->toBasetype(), true);
   return l->equals(r);
 }
+
+Expression *getCommaHead(CommaExp *e) {
+  auto l = e->e1;
+  for (;;) {
+    if (auto ce = l->isCommaExp())
+      l = ce->e1;
+    else
+      break;
+  }
+  return l;
+}
+
+Expression *getCommaTail(CommaExp *e) {
+  auto r = e->e2;
+  for (;;) {
+    if (auto ce = r->isCommaExp())
+      r = ce->e2;
+    else
+      break;
+  }
+  return r;
+}
 }
 
 bool toInPlaceConstruction(DLValue *lhs, Expression *rhs) {
+  Logger::println("attempting in-place construction");
+  LOG_SCOPE;
+
   // Is the rhs the init symbol of a zero-initialized struct?
   // Then aggressively zero-out the lhs, without any type checks, e.g., allowing
   // to initialize a `S[1]` lhs with a `S` rhs.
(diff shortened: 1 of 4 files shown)


kinke:
I've added an experimental in-place-construction extension which gets rid of more blits, for all targets.

tests/codegen/in_place_construct_temporaries.d
```diff
@@ -0,0 +1,83 @@
...
+        if (p != &data[0])
+        {
+            import core.stdc.stdio : printf;
+            printf("Check failed in line %d\n", line);
+            success = false;
+        }
+    }
+}
+
+void func(Container c) { c.check(); } // error
```
kinke: This check requires the ABI change in the 1st commit to prevent blitting the copy-constructed temporary argument in line 61.

gen/toir.cpp
```diff
@@ -2815,17 +2845,41 @@ bool toInPlaceConstruction(DLValue *lhs, Expression *rhs) {
...
+      if (auto de = head->isDeclarationExp()) {
+        if (auto vd = de->declaration->isVarDeclaration()) {
+          if (auto ve = tail->isVarExp()) {
+            if (ve->var == vd) {
+              Logger::println("success, in-place-constructing temporary");
+              auto lhsLVal = DtoLVal(lhs);
+              auto rhsLVal = DtoLVal(rhs);
+              rhsLVal->replaceAllUsesWith(lhsLVal);
+              return true;
+            }
```
SSoulaimane: This works with Posix but Win32 will be more interesting, since you have to construct the arguments directly on their push area.

kinke: Yes, I've looked with horror at our 32-bit x86 ABI implementation yesterday and hoped it was just a bug that non-PODs are passed on the stack. But that's indeed what should happen, and the way to implement that in LLVM IR is going to be a pain (allocating the full params stack on the caller, as an anonymous struct); clang produces [this](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:c%2B%2B,source:'struct+S%0A%7B%0A++++int+i%3B%0A++++S()+%7B%7D%0A++++S(const+S%26)+%7B%7D%0A%7D%3B%0A%0Avoid+takeS(S,+int,+S)+%7B%7D%0A%0A//+Type+your+code+here,+or+load+an+example.%0Avoid+foo()+%7B%0A++++S+s%3B%0A++++takeS(s,+123,+s)%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:33.33333333333333,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:clang900,filters:(b:'1',binary:'1',commentOnly:'1',demangle:'1',directives:'1',execute:'0',intel:'1',libraryCode:'1',trim:'1'),lang:c%2B%2B,libs:!(),options:'-target+i686-pc-windows-msvc',source:1),l:'5',n:'0',o:'x86-64+clang+9.0.0+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:ir,i:(compilerName:'x86-64+clang+9.0.0',editorid:1,j:1,source:'struct+S%0A%7B%0A++++int+i%3B%0A++++S()+%7B%7D%0A++++S(const+S%26)+%7B%7D%0A%7D%3B%0A%0Avoid+takeS(S)+%7B%7D%0A%0A//+Type+your+code+here,+or+load+an+example.%0Avoid+foo()+%7B%0A++++S+s%3B%0A++++takeS(s)%3B%0A%7D'),l:'5',n:'0',o:'x86-64+clang+9.0.0+IR+Viewer+(Editor+%231,+Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

dnadlinger: See also https://llvm.org/docs/InAlloca.html.

kinke: Ugly and unsuited for the current ABI-related LDC code architecture. I definitely don't care enough about Win32 to implement this. Right now, the compiler is free to blit around as much as it wants, so this is just codegen optimization. For extern(D), we could simply violate Walter's ABI and pass non-PODs by ref.

kinke:
[I've verified the 3rd commit in my Linux VM, as CI doesn't run the 32-bit LIT tests on Posix.]

tests/codegen/in_place_construct_temporaries.d
```diff
@@ -0,0 +1,90 @@
...
+
+    Container[1] slit = [v];
+    slit[0].check(); // error
+
+    Container[] dlit = [v];
+    dlit[0].check(); // error
+
+    auto b = B(v);
+    b.m.check(); // error
+
```
SSoulaimane: I couldn't solve the following test case in DMD yet, it seems this PR doesn't solve it for LDC either.
```d
Container[] darr;
darr ~= v;
darr[0].check(); // error
```
This is the machinery work
https://github.com/ldc-developers/ldc/blob/058202a55b4a015b73848b1d59c3319e511df0b1/gen/arrays.cpp#L825-L844
DMD does the same thing. First evaluate into a temporary, expand the array, then assign to `darr[$-1]`.

Any idea on how to solve it?

kinke: In this case, it might be better to let the frontend lower the CatAssignExp to the druntime call for GC expansion and then copy-constructing the last element (copy ctor or blit+postblit). IIRC, there are a few more places where the glue layer needs to insert postblit calls (and where it should now call the copy ctor if available), and ideally we'd get rid of all of them.

PR #3209, opened 2019-10-27
Fix equality/identity comparisons of vectors with length >= 32

### LINKED ISSUE #3208: __vector(byte[32]) wrong code gen for ==
```D
// Tested and fails on x86-64:
import std.stdio;

void main() {
    __vector(byte[32]) a = 0, b = a;
    writeln(a == b); // prints "false"
}
```

kinke:
Fixes #3208.

### DIFF
--- a/gen/binops.cpp
+++ b/gen/binops.cpp
@@ -392,8 +392,7 @@ LLValue *mergeVectorEquals(LLValue *resultsVector, TOK op) {
 
   if (op == TOKequal) {
     // all pairs must be equal for the vectors to be equal
-    LLConstant *allEqual =
-        LLConstantInt::get(integerType, (1 << sizeInBits) - 1);
+    LLConstant *allEqual = LLConstant::getAllOnesValue(integerType);
     return gIR->ir->CreateICmpEQ(v, allEqual);
   } else if (op == TOKnotequal) {
     // any not-equal pair suffices for the vectors to be not-equal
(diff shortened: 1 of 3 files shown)


kinke:
Hmm, I guess LLVM < 8 (or < 7? Travis doesn't test LLVM 7 anymore, maybe we should change the 2nd Mac job from LLVM 8 to 7 again...) fails to cast i1 vectors with length > 64 to an integer with equal size.

gen/binops.cpp
```diff
@@ -392,8 +392,7 @@ LLValue *mergeVectorEquals(LLValue *resultsVector, TOK op) {
 
   if (op == TOKequal) {
     // all pairs must be equal for the vectors to be equal
-    LLConstant *allEqual =
-        LLConstantInt::get(integerType, (1 << sizeInBits) - 1);
```
JohanEngelen: maybe we should ban all shifts by variable amounts to not get hit by these UB shifts!

kinke: The VS IDE showed a warning; I guess we inherit flags disabling these warnings from LLVM.

kinke:
Löl, LLVM 6 now passes with max length of 32, but older LLVM still fails. Might 'just' need enabling of advanced instruction sets, or really fails to cast longer i1 vectors to integers...

JohanEngelen:
hm, strange indeed. Good perseverance ;)

PR #3213, opened 2019-11-01
Shippable: Switch linker from gold to bfd

kinke:


kinke:
This reduces false negatives for dmd-testsuite due to removed linker warnings in checked output; the number of failures is reduced from 45 to 21.

kinke:
Down to 15 failures, which look all valid after a quick glance - most due to unimplemented `core.stdc.stdarg`, some due to lacking C(++) ABI compatibility.

kinke:
Unsure about whether it's safe and desirable to switch to bfd for the Aarch64 release package as well (as it currently is with this PR here). While Shippable is based on Ubuntu 16.04, its gcc is way more recent (8.1 vs 5.3), so I guess binutils too.

PR #3218, opened 2019-11-06
Azure CI Windows: Build LDC itself with LTO

kinke:
We use clang 9 as C++ compiler and the non-LTO'd bootstrap compiler is linked against matching LLVM 9, so everything's set up for LLVM 9 LTO already.

So when building the 2nd, final compiler:
* let clang compile the C++ parts of the ldc2 executable as LTO objects (the LDC parts, i.e., excluding LLVM libs)
* use the bootstrap LDC for archiving the static LDCShared library (as it only contains the C++ LTO objects)
* let the bootstrap compiler produce LTO objects only (for all D executables, not just ldc2)
* let the bootstrap compiler link all D executables with LTO, incl. LTO-able druntime & Phobos (via `lld-link` v9)

### DIFF
--- a/.azure-pipelines/windows.yml
+++ b/.azure-pipelines/windows.yml
@@ -78,7 +78,7 @@ steps:
     clang-cl --version
     mkdir bootstrap-ldc
     cd bootstrap-ldc
-    cmake -G Ninja %BUILD_SOURCESDIRECTORY% -DCMAKE_C_COMPILER:PATH=clang-cl.exe -DCMAKE_CXX_COMPILER:PATH=clang-cl.exe -DCMAKE_BUILD_TYPE=Release -DLLVM_ROOT_DIR=%CD%/../llvm -DD_COMPILER=%CD%/../host-ldc/bin/ldmd2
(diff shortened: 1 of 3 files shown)


kinke:
An extra motivation is not having to worry about performance hits when crossing language boundaries (#3141).

kinke:
I performed a few rough comparisons yesterday; no measurable speed gains at all, and the ldc2, ldc-build-runtime etc. sizes slightly increased by a few hundred KB.

kinke:
Full LTO finally does show a bit of an improvement (~3.3% faster runtime for `ldc2 -c -o- -unittest runtime\phobos\std\typecons.d`). Runtimes (best of 6 runs) and ldc2 executable sizes on my box:
```
No LTO:   4.193 s, 62,192 KB
Thin LTO: 4.161 s, 62,153 KB
Full LTO: 4.055 s, 62,116 KB
```

kinke:
[One annoyance of this is that LTO will have to be temporarily disabled when testing a new LLVM version, as long as there's no corresponding prebuilt clang version.]

PR #3220, opened 2019-11-09
Make llvm_atomic_cmp_xchg signature match LLVM's

kinke:


kinke:
Fixing the 2 CI issues wasn't actually the plan, but nice anyway. ;)

Issue #3216, opened 2019-11-03
CMake swallows -Xlinker option from dmd.conf


mleise:
In my `dmd.conf` I have:
```
[Environment]
DFLAGS=-I/usr/lib/dmd/2.087/import -L--export-dynamic -defaultlib=phobos2 -fPIC
[Environment32]
DFLAGS=%DFLAGS% -L-L/usr/lib/dmd/2.087/lib32 -L-rpath -L/usr/lib/dmd/2.087/lib32
[Environment64]
DFLAGS=%DFLAGS% -L-L/usr/lib/dmd/2.087/lib64 -L-rpath -L/usr/lib/dmd/2.087/lib64
```
That results in the following dmd linker line:
```
cc <name>.o -o <name> -m64 -Xlinker --export-dynamic -Xlinker -rpath -Xlinker /usr/lib/dmd/2.089/lib64 -L/usr/lib/dmd/2.089/lib64 -lphobos2 -lpthread -lm -lrt -ldl
```
As of ldc2 1.18.0 the CMake process outputs the following and swallows the library search path, thinking it is meant to be a library:
```
WARNING: Target "ldmd2" requests linking to directory "/usr/lib/dmd/2.087/lib64".  Targets may link only to libraries.  CMake is dropping the item.
WARNING: Target "ldc2-unittest" requests linking to directory "/usr/lib/dmd/2.087/lib64".  Targets may link only to libraries.  CMake is dropping the item.
WARNING: Target "ldc2" requests linking to directory "/usr/lib/dmd/2.087/lib64".  Targets may link only to libraries.  CMake is dropping the item.
WARNING: Target "ldc-build-runtime" requests linking to directory "/usr/lib/dmd/2.087/lib64".  Targets may link only to libraries.  CMake is dropping the item.
WARNING: Target "ldc-prune-cache" requests linking to directory "/usr/lib/dmd/2.087/lib64".  Targets may link only to libraries.  CMake is dropping the item.
```
As a result the build of these targets fails to locate `-lphobos2`:
```
[...]

PR #3236, opened 2019-11-23
Azure CI: Build LDC itself with full LTO, on Linux and Mac too

kinke:


kinke:
Seems to work after some bash fiddling; I wasn't really expecting this, as I have no idea what LLVM version the Xcode clang is based on. No idea whether full or thin LTO makes a difference as for Windows.

kinke:
Hmm I think there's still a problem, a variant of #3216:
```
D_LINKER_ARGS="-lto_library;/Users/runner/runners/2.160.1/work/1/bootstrap-ldc/lib/libLTO-ldc.dylib;-L/Users/runner/runners/2.160.1/work/1/bootstrap-ldc/lib;-lphobos2-ldc-lto;-ldruntime-ldc-lto;-ldl;-lpthread;-lm;-m64"
```
and then we add those for the C++ linker (clang) as CMake `target_link_libraries` (IIRC, for correct linker flags order) - so the 2nd arg in `-lto_library .../libLTO-ldc.dylib -L.../lib ...` will be swallowed/treated as a dependency to be linked into LDC executable, and clang might just get `-lto_binary -L.../lib ...`.

kinke:
After grepping the clang source for `lto_library`, I've only found this: https://github.com/llvm/llvm-project/blob/0010ea4224c7ff6b5aebeecacf10a7ced7d390b7/clang/lib/Driver/ToolChains/Darwin.cpp#L249-L264. It seems to firstly always specify it in the `ld64` cmdline, and secondly unconditionally use its own LTO plugin version, not the one we'd like it to use (does it silently ignore a user-provided `-lto_library`?!). @JohanEngelen: could you please investigate what's actually happening here?

driver/linker-gcc.cpp
```diff
@@ -169,8 +169,7 @@ std::string getLTOdylibPath() {
 void ArgsBuilder::addDarwinLTOFlags() {
   std::string dylibPath = getLTOdylibPath();
   if (!dylibPath.empty()) {
-    args.push_back("-lto_library");
-    args.push_back(std::move(dylibPath));
+    addLdFlag("-lto_library", dylibPath);
```
kinke: This is the admittedly ugly CMake workaround - invoking clang with `-Wl,-lto_library,/path/to/libLTO.dylib`, so that it can be safely added as single arg via `target_link_libraries` in the 2nd CMake invocation. Seems to work; I assume Apple-clang just forwards `-lto_library <path>` to ld64 too.

kinke:
I quickly checked the D executable file sizes back then; on Linux, they are pretty much the same, while the Mac ones are significantly reduced (~10% for `ldc2` IIRC).

PR #3244, opened 2019-11-30
Upgrade LLVM to v9.0.1 and add Android/AArch64 CI job/package

kinke:


kinke:
This is preparational work for a hopefully pretty timely LLVM 9.0.1. The used LLVM here includes the experimental AVR backend (for the prebuilt x86 packages) as well as a prebuilt Android/AArch64 package.

The 2nd commits add prebuilt Android/i686 libs to the armv7a package, and Android/x86_64 libs to the new aarch64 package, so that all Android platforms are covered with prebuilt libs.

kinke:
Rebased and now based on LLVM 9.0.1-rc3. As the llvm-mirror repos haven't been updated anymore since the end of October, I transplanted our LLVM/LLD/compiler-rt patches incl. CI/package generation to a new fork of upstream's mono-repo, https://github.com/ldc-developers/llvm-project.

kinke:
Are there any preferences wrt. built LLVM src package? Right now, it includes the whole (mono-)repo as-is, so the `.tar.xz` is 70 MB; the previous archive (without clang, libcxx, lldb etc.) was 33 MB. The official LLVM src archives are still separate.

My main problem with the mono-repo source is that the build instructions would have to be touched (and diverge from vanilla LLVM), as the `llvm` subdir is the root dir to be used, and 2 new options are required (`-DLLVM_ENABLE_PROJECTS='compiler-rt;lld' -DCOMPILER_RT_USE_LIBCXX=OFF`). The alternative is to move the `compiler-rt` and `lld` src dirs back into `llvm/{projects,tools}` and archive the `llvm` dir.

PR #3246, opened 2019-12-01
Delete -L--no-warn-search-mismatch on multilib builds

MaskRay:
```
% ldc -linker=lld a.d
ld.lld: error: unknown argument '--no-warn-search-mismatch'
...
```

lld does not implement this option (https://bugs.llvm.org/show_bug.cgi?id=43740).
I think the option not useful because:

1) This usually indicates real library path configuration problems that should not be ignored.
2) If a .so is actually a linker script that includes other libraries,
 but without an OUTPUT_FORMAT directive, GNU ld and gold's behavior of
 skipping incompatible libraries may still not work if the linker script
 contents are different.
 For example, libncurses.so on Debian does not include an OUTPUT_FORMAT directive.

In addition, clang never passes this option.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -234,7 +234,6 @@ endif()
 # DMD and GDC do). OS X doesn't need extra configuration due to the use of
 # fat binaries. Other Posixen might need to be added here.
 if(MULTILIB AND NOT "${TARGET_SYSTEM}" MATCHES "APPLE")
-    set(ADDITIONAL_DEFAULT_LDC_SWITCHES  "${ADDITIONAL_DEFAULT_LDC_SWITCHES}\n        \"-L--no-warn-search-mismatch\",")
     set(MULTILIB_ADDITIONAL_PATH         "\n        \"${CMAKE_BINARY_DIR}/lib${MULTILIB_SUFFIX}\",")
     set(MULTILIB_ADDITIONAL_INSTALL_PATH "\n        \"${CMAKE_INSTALL_PREFIX}/lib${MULTILIB_SUFFIX}\",")
 endif()


kinke:
This is just a workaround for a simpler ldc2.conf, as it allows specifying both 32/64 bit lib dirs via `-L` and make the linker ignore mismatching libs. Specifying the single proper lib dir would require a separate section in the config file, and we'd need to determine the 32-bit triple at CMake config time.

MaskRay:
> This is just a workaround for a simpler ldc2.conf, as it allows specifying both 32/64 bit lib dirs via `-L` and make the linker ignore mismatching libs. Specifying the single proper lib dir would require a separate section in the config file, and we'd need to determine the 32-bit triple at CMake config time.

Can you give me some links how we can properly address the ldc2.conf problem?

kinke:
See https://github.com/ldc-developers/ldc/blob/master/ldc2_install.conf.in - we'd need a separate section for the 32-bit target, something like:
```
"@MULTILIB_SECTION_NAME@":
{
    switches = [
        "-defaultlib=phobos2-ldc,druntime-ldc",@ADDITIONAL_DEFAULT_LDC_SWITCHES@
    ];
    post-switches = [
        "-I@INCLUDE_INSTALL_DIR@",
    ];
    lib-dirs = [
        "@MULTILIB_INSTALL_LIBDIR@",
    ];
    rpath = "@SHARED_LIBS_MULTILIB_INSTALL_RPATH@";
};
```

with appropriate CMake variables replacing the `@VAR@` placeholders in https://github.com/ldc-developers/ldc/blob/ad80f004aeede0b1582bf9831133c000fecfef07/runtime/CMakeLists.txt#L279-L280.

kinke:
Wrt. figuring out the 32-bit triple, that's probably the hardest part. It'd be easy if the config file finalization was delayed until we have built the ldc2 executable (parsing the triple from `ldc2 -m32 -v` output, 3rd line).

kinke:
Superseded by #3276.

PR #3247, opened 2019-12-02
Remove _d_execBss{Beg,End}Addr, used for copy-relocation check

kubo39:
I think we can remove the copy-relocation logic since https://github.com/ldc-developers/druntime/commit/9b97ede74f9b8840c77d7a35aaf573ace012c1d0 was comitted.


kinke:
Thx; I'll merge after the v1.19 final release, just to be on the safe side in case a 3rd party lib depends on these symbols.

PR #3248, opened 2019-12-04
llvm10

thewilsonator:
Fix compilation with head of tree

### DIFF
--- a/driver/toobj.cpp
+++ b/driver/toobj.cpp
@@ -73,7 +73,7 @@ void codegenModule(llvm::TargetMachine &Target, llvm::Module &m,
   if (cb == ComputeBackend::SPIRV) {
 #ifdef LDC_LLVM_SUPPORTED_TARGET_SPIRV
     IF_LOG Logger::println("running createSPIRVWriterPass()");
-#if LDC_LLVM_VER >= 900
+#if LDC_LLVM_VER >= 900 && LDC_LLVM_VER < 1000
     std::ofstream out(filename, std::ofstream::binary);
 #else
     std::error_code errinfo;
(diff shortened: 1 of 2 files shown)


driver/toobj.cpp
```diff
@@ -47,6 +47,14 @@
 #include <cstddef>
 #include <fstream>
 
+#if LDC_LLVM_VER < 1000
+using CGFT_AssemblyFile = llvm::TargetMachine::CGFT_AssemblyFile;
+using CGFT_ObjectFile = llvm::TargetMachine::CGFT_ObjectFile;
```
JohanEngelen: what are these for llvm >= 10 ?

thewilsonator: Nothing? Maybe we have a using hiding something somewhere, but it compiles for me as is.

driver/toobj.cpp
```diff
@@ -57,15 +63,15 @@ namespace {
...
+                   CodeGenFileType fileType) {
   using namespace llvm;
 
   const ComputeBackend::Type cb = getComputeTargetType(&m);
 
   if (cb == ComputeBackend::SPIRV) {
 #ifdef LDC_LLVM_SUPPORTED_TARGET_SPIRV
     IF_LOG Logger::println("running createSPIRVWriterPass()");
-#if LDC_LLVM_VER >= 900
+#if LDC_LLVM_VER == 900
```
kinke: 901 is due in a few weeks.

thewilsonator: Fixed

driver/toobj.cpp
thewilsonator: Derp its the using below.

thewilsonator:
I think this is good to go now.

PR #3257, opened 2019-12-15
Support for LLVM 10 

Robertorosmaninho:
Integer Align parameter on functions replaced by a function called MaybeAlign which can be initialized with an integer.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -1354,10 +1354,6 @@ void DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *llVar,
   DIScope scope = GetSymbolScope(vd);
   llvm::MDNode *Decl = nullptr;
 
-#if LDC_LLVM_VER >= 1000
-  llvm::DIExpression *Expr = nullptr;
-#endif
-  
   if (vd->isDataseg() && vd->toParent()->isAggregateDeclaration()) {
     // static aggregate member
     Decl = StaticDataMemberCache[vd];
@@ -1386,7 +1382,7 @@ void DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *llVar,
 #endif
     
 #if LDC_LLVM_VER >= 1000
-      Expr, // expression
+      nullptr, // expression
 #endif
       Decl // declaration
   );
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -1354,6 +1354,10 @@ void DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *llVar,
   DIScope scope = GetSymbolScope(vd);
   llvm::MDNode *Decl = nullptr;
 
+#if LDC_LLVM_VER >= 1000
+  llvm::DIExpression *Expr = nullptr;
+#endif
+  
   if (vd->isDataseg() && vd->toParent()->isAggregateDeclaration()) {
     // static aggregate member
     Decl = StaticDataMemberCache[vd];
@@ -1379,6 +1383,10 @@ void DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *llVar,
       nullptr, // relative location of field
 #else
       llVar, // value
+#endif
+    
+#if LDC_LLVM_VER >= 1000
+      Expr, // expression
 #endif
       Decl // declaration
   );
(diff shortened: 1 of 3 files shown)


gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -79,7 +79,11 @@ void EmitMemSet(IRBuilder<> &B, Value *Dst, Value *Val, Value *Len,
                 const Analysis &A) {
   Dst = B.CreateBitCast(Dst, PointerType::getUnqual(B.getInt8Ty()));
 
-  auto CS = B.CreateMemSet(Dst, Val, Len, 1 /*Align*/, false /*isVolatile*/);
+#if LDC_LLVM_VER >= 1000
+  llvm::MaybeAlign Align(1);
+#endif
+
+  auto CS = B.CreateMemSet(Dst, Val, Len, Align, false /*isVolatile*/);
```
thewilsonator: `/gen/passes/GarbageCollect2Stack.cpp:86:43: error: 'Align' was not declared in this scope`

You still need to declare the variable for LLVM < 10

Robertorosmaninho: Sorry, I forgot about this detail.

gen/tollvm.cpp
```diff
@@ -327,7 +327,11 @@ void DtoMemSet(LLValue *dst, LLValue *val, LLValue *nbytes, unsigned align) {
 
   dst = DtoBitCast(dst, VoidPtrTy);
 
-  gIR->ir->CreateMemSet(dst, val, nbytes, align, false /*isVolatile*/);
+#if LDC_LLVM_VER >= 1000
+  llvm::MaybeAlign Align(align);
+#endif
+
+  gIR->ir->CreateMemSet(dst, val, nbytes, Align, false /*isVolatile*/);
```
kinke: It's missing here as well - but you can simply use the existing `LLMaybeAlign` macro, as done in https://github.com/ldc-developers/ldc/blob/7e40cf1af5aea8e70313c01440b01d91b0fd2523/gen/tollvm.cpp#L467

Robertorosmaninho: thanks, this should work, right?

gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -79,7 +79,13 @@ void EmitMemSet(IRBuilder<> &B, Value *Dst, Value *Val, Value *Len,
                 const Analysis &A) {
   Dst = B.CreateBitCast(Dst, PointerType::getUnqual(B.getInt8Ty()));
 
-  auto CS = B.CreateMemSet(Dst, Val, Len, 1 /*Align*/, false /*isVolatile*/);
+#if LDC_LLVM_VER >= 1000
+  LLMaybeAlign Align(1);
```
JohanEngelen: you should be able to just use `MaybeAlign` here because of `using namespace llvm` at the top.
(if not, I prefer `llvm::MaybeAlign`).

gen/tollvm.cpp
JohanEngelen: the reason for the `LLMaybeAlign` alias is not to just rename `llvm::MaybeAlign` (that'd be silly), but to make it available regardless of LLVM version. So you can just use it here, without the `#if LDC_LLVM_VER >= 1000` parts.

gen/dibuilder.cpp
```diff
@@ -1354,6 +1354,10 @@ void DIBuilder::EmitGlobalVariable(llvm::GlobalVariable *llVar,
   DIScope scope = GetSymbolScope(vd);
   llvm::MDNode *Decl = nullptr;
 
+#if LDC_LLVM_VER >= 1000
+  llvm::DIExpression *Expr = nullptr;
```
JohanEngelen: `Expr` is not a very descriptive variable name. I would just put `nullptr` as argument to the `createGlobalVariable` call without unnecessarily creating a local variable.

Robertorosmaninho: ok!

gen/passes/GarbageCollect2Stack.cpp
Robertorosmaninho: ok!

PR #3259, opened 2019-12-17
Embed linker directives in ELF and Mach-O object files

### LINKED ISSUE #3245: Support LLD's new dependant-libraries feature via pragma(lib)
LLD 9 introduces support for library dependencies embedded in ELF object files, see https://reviews.llvm.org/rL360984, by listing the library names in a special `.deplibs` section. Not sure if it's really as useful as for Windows, where the embedded linker directives firstly aren't restricted to library names, but are arbitrary linker cmdline flags, and where the linker secondly isn't picky about the libraries order in the cmdline (a later lib depending on a prior one isn't an issue on Windows).

kinke:
Resolves #3245 by adding `pragma(lib, <name>)` library names to `llvm.dependent-libraries` for ELF object files.

For Mach-O, embed appropriate linker options for `pragma(lib)` and support generic `pragma(linkerDirective, <flag>, ...)` as well.

### DIFF
--- a/tests/codegen/linker_directives_linux.d
+++ b/tests/codegen/linker_directives_linux.d
@@ -1,0 +1,11 @@
+// RUN: %ldc -mtriple=x86_64-linux-gnu -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+// REQUIRES: atleast_llvm500, target_X86
+
+// CHECK: !llvm.dependent-libraries = !{!0}
+// CHECK: !0 = !{!"mylib"}
+pragma(lib, "mylib");
+
+// silently ignored because not (yet?) embeddable in ELF object file:
+pragma(linkerDirective, "-myflag");
+pragma(linkerDirective, "-framework", "CoreFoundation");
--- a/dmd/dsymbolsem.d
+++ b/dmd/dsymbolsem.d
@@ -1851,25 +1851,27 @@ version (IN_LLVM)
         const(char)* arg1str = null;
 }
 
-        if (global.params.mscoff)
+        // IN_LLVM: extended pragma(linkerDirective) support - not just for COFF
+        //          object files, and not restricted to a single string arg
+        if (pd.ident == Id.linkerDirective)
         {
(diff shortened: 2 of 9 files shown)


kinke:
I left the direct linker cmdline args as-is for now.

For Linux, they are obviously still required as the special section is only recognized by LLD 9+ for now.

For Mac, we only support `ld64` IIRC, so the cmdline args might be skippable, relying exclusively on the args embedded in the object files, just like for Windows.

JohanEngelen:
Nit: all test files need a `// REQUIRES: target_X86`

tests/codegen/linker_directives_linux.d
```diff
@@ -0,0 +1,10 @@
+// RUN: %ldc -mtriple=x86_64-linux-gnu -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+// REQUIRES: atleast_llvm500
+
+// CHECK: !llvm.dependent-libraries = !{!0}
+// CHECK: !0 = !{!"mylib"}
+pragma(lib, "mylib");
+
+pragma(linkerDirective, "-myflag");
+pragma(linkerDirective, "-framework", "CoreFoundation");
```
JohanEngelen: So these directives are not added to metadata, but on the commandline instead? Let's test that?

kinke:
> Nit: all test files need a // REQUIRES: target_X86

Indeed, I somehow thought generating .ll always works - I'll have to update the Wiki page accordingly too.

tests/codegen/linker_directives_linux.d
kinke: Nope, they are ignored for Linux. From https://dlang.org/spec/pragma.html#linkerDirective:
> The string literal specifies a linker directive to be embedded in the generated object file.

JohanEngelen: Ah, now i understand your original PR comment. So on top of COFF support what DMD does, it's only enabled for macOS. Perhaps add a comment about that in the testcase, or simply remove those lines?

PR #3262, opened 2019-12-21
Upgrade frontend & libs to v2.090.0+

kinke:


kinke:
The CircleCI regression is an annoying re-surfacing of the issue above. The only remaining blocker (Win32 failure) is interesting:
```D
// ldc/profile.d:
void getData(alias F)()
{
    import std.digest.md;
    auto md5hash = md5Of("bla");
}

/// profile_rt_calls.d
import ldc.profile;
void getdataprofile()
{
    getData!getdataprofile();
}
```

When cross-compiling to 32-bit x86 (both Windows and Linux), the `profile_rt_calls` ModuleInfo somehow references the `std.digest.md` and `std.range` ModuleInfos as imported modules. No imported modules for x64 targets with the same compiler binary.

JohanEngelen:
> When cross-compiling to 32-bit x86 (both Windows and Linux), the `profile_rt_calls` ModuleInfo somehow references the `std.digest.md` and `std.range` ModuleInfos as imported modules. No imported modules for x64 targets with the same compiler binary, nor with LDC v1.19.

Interesting indeed. Any idea why importing may be different on x86?

kinke:
It works when importing v1.19/2.089 druntime/Phobos; I guess some library restructurings now reveal an existing compiler bug (I couldn't spot any relevant changes in the diff). The failing PGO test is purely accidental, as the compiled module's ModuleInfo happens to reference the `ldc.profile` ModuleInfo, which is a .di header, so linking fails.

kinke:
> I guess some library restructurings now reveal an existing compiler bug

Confirmed; adding a top-level `import std.math` in `std.format` suffices to make LDC 1.19 reference the 2 ModuleInfos too. It's specific to x86 as `std.math` imports `core.cpuid` for x86 only (to detect SSE availability at runtime...), and `core.cpuid` has a module ctor, so its `Module.needmoduleinfo` gets set, and that flag seems to get set for all importing modules (transitively) too.

kinke:
Interestingly, DMD emits these importedModules references as weak symbols, we don't: https://github.com/dlang/dmd/blob/49dfbe54f4a0f62a0694aec9fb4184739903a0c2/src/dmd/toobj.d#L191-L196

kinke:
Rather surprisingly, `extern_weak` linkage worked out of the box for Win64 - with the MS linker. LLD fails. :(

kinke:
The weak `importedModules` also cause a linking regression for 32-bit Linux x86 and `runnable/testwc[2].d` (`undefined reference to _D3std4file8readImplFNeMAxaMPxakZAv`; the `std.file` ModuleInfo is referenced as `external_weak` in both object files); static Phobos only (shared works). Template instantiations are emitted as `weak_odr` by default; if I understand the LLVM docs correctly, that should be enough to pull in the object file when resolving an `external` declaration.

kinke:
> a linking regression for 32-bit Linux x86

Correction: with gold; bfd works fine in my VM. Also, gold works with `-L/home/martin/build-ldc/lib32` (in gcc cmdline) and only fails with `--no-warn-search-mismatch -L/home/martin/build-ldc/lib -L/home/martin/build-ldc/lib32` => extra motivation for #3246.

PR #3264, opened 2019-12-25
irstate.h: fix -Wdeprecated-copy

MaskRay:
Since C++11, [depr.impldec] says:

> The implicit definition of a copy constructor as defaulted is deprecated
> if the class has a user-declared copy assignment operator or a
> user-declared destructor.

clang 10 will have a warning -Wdeprecated-copy (included by -Wextra)
which flags the code.


thewilsonator:
Thanks!

PR #3265, opened 2019-12-25
Use function attribute "frame-pointer" instead of "no-frame-pointer-elim"/"no-frame-pointer-elim-non-leaf"

MaskRay:
LLVM 8 ([D56351](http://reviews.llvm.org/D56351)) introduced "frame-pointer" which was intended to replace
"no-frame-pointer-elim"/"no-frame-pointer-elim-non-leaf".

-----

In the LLVM monorepo, run `git show origin/release/8.x:llvm/lib/CodeGen/TargetOptionsImpl.cpp` to see the effect of `"frame-pointer"`.

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -480,15 +480,13 @@ void applyTargetMachineAttributes(llvm::Function &func,
 #if LDC_LLVM_VER >= 800
   switch (whichFramePointersToEmit()) {
     case llvm::FramePointer::None:
-      func.addFnAttr("no-frame-pointer-elim", "false");
+      func.addFnAttr("frame-pointer", "none");
       break;
     case llvm::FramePointer::NonLeaf:
-      func.addFnAttr("no-frame-pointer-elim", "false");
-      func.addFnAttr("no-frame-pointer-elim-non-leaf");
+      func.addFnAttr("frame-pointer", "non-leaf");
       break;
     case llvm::FramePointer::All:
-      func.addFnAttr("no-frame-pointer-elim", "true");
-      func.addFnAttr("no-frame-pointer-elim-non-leaf");
+      func.addFnAttr("frame-pointer", "all");
       break;
   }
 #else
@@ -1131,12 +1129,14 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   // disable frame-pointer-elimination for functions with inline asm
   if (fd->hasReturnExp & 8) // has inline asm
   {
+#if LDC_LLVM_VER >= 800
+    func->addFnAttr(
+        llvm::Attribute::get(gIR->context(), "frame-pointer", "all"));
+#else
     func->addAttribute(
         LLAttributeSet::FunctionIndex,
         llvm::Attribute::get(gIR->context(), "no-frame-pointer-elim", "true"));
-    func->addAttribute(
-        LLAttributeSet::FunctionIndex,
-        llvm::Attribute::get(gIR->context(), "no-frame-pointer-elim-non-leaf"));
+#endif
   }
 
   // give the 'this' parameter (an lvalue) storage and debug info
--- a/tests/codegen/attr_targetoptions.d
+++ b/tests/codegen/attr_targetoptions.d
@@ -1,3 +1,4 @@
+// REQUIRES: atleast_llvm800
 // Tests that our TargetMachine options are added as function attributes
 
 // RUN: %ldc -c -output-ll -of=%t.ll %s
@@ -21,7 +22,7 @@ void foo()
 // COMMON-DAG: "no-infs-fp-math"="false"
 // COMMON-DAG: "no-nans-fp-math"="false"
 
-// WITH_FP-DAG: "no-frame-pointer-elim"="true"
-// NO_FP-DAG:   "no-frame-pointer-elim"="false"
+// WITH_FP-DAG: "frame-pointer"="all"
+// NO_FP-DAG:   "frame-pointer"="none"
 
 // ATTR-DAG: "target-features"="{{[^"]*}}+test
--- a/tests/codegen/frame_pointer_x86.d
+++ b/tests/codegen/frame_pointer_x86.d
@@ -1,0 +1,25 @@
+// REQUIRES: target_X86
+
+// RUN: %ldc -c -mtriple=x86_64 -output-s -of=%t.s %s
+// RUN: FileCheck %s --check-prefixes=COMMON,FP < %t.s
+// RUN: %ldc -c -mtriple=x86_64 -output-s -of=%t.s %s -O2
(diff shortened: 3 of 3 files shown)


JohanEngelen:
Looks good to me.
`tests/codegen/attr_targetoptions.d` needs adjustment for this. I'm OK with putting `// REQUIRES: atleast_llvm800` on that test and fixing it for this improved attribute.

gen/functions.cpp
```diff
@@ -1131,12 +1129,13 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
...
   if (fd->hasReturnExp & 8) // has inline asm
   {
-    func->addAttribute(
-        LLAttributeSet::FunctionIndex,
+#if LDC_LLVM_VER >= 800
+    func->addFnAttr(
+        llvm::Attribute::get(gIR->context(), "frame-pointer", "all"));
+#else
+    func->addFnAttr(
         llvm::Attribute::get(gIR->context(), "no-frame-pointer-elim", "true"));
```
kinke: As shown by Travis CI, LLVM 3.9 doesn't feature `addFnAttr()`. You'll probably also need to re-add `no-frame-pointer-elim-non-leaf`.

MaskRay: `"no-frame-pointer-elim-non-leaf"` is ignored when `"no-frame-pointer-elim"="true"` is present.

kinke: I don't mean to be a PITA, but I'd like some reference; it's a breaking change after all, and as we go back to LLVM 3.9, I'm on the conservative side wrt. changes like this. clang 9 also still emits `"no-frame-pointer-elim"="true" "no-frame-pointer-elim-non-leaf"`.

JohanEngelen: Is this a breaking change? I thought it only changes our output to using more recent function attributes?
Code reference: https://github.com/llvm/llvm-project/blob/e5957ac3d7135eca95b4019eaa394703ad6112a6/llvm/lib/CodeGen/TargetOptionsImpl.cpp#L24-L53
(shows that "no-frame-pointer-elim-non-leaf" is ignored when "no-frame-pointer-elim"="true" is present. and that "frame-pointer" is the one we should be using)

Rather than checking LLVM IR output, I think we should test X86 assembly output to make sure we don't silently loose functionality.  (If LLVM would have removed "no-frame-pointer-elim" support, would our testsuite have caught that the framepointer is now removed/kept without us knowing?)

MaskRay: It is not a breaking change for LLVM 3.9.

clang 10 will not have `"no-frame-pointer-elim"="true" "no-frame-pointer-elim-non-leaf"`. I decide to remove "no-frame-pointer-elim" : https://reviews.llvm.org/D71863 .

MaskRay:
My ldc source is at `~/Dev/ldc`. The build directory is `~/Dev/ldc/Debug9/`.

To run `codegen/` tests, I run:

```zsh
cd ~/Dev/ldc/Debug9/tests
python3 ./runlit.py codegen
```

In llvm, I can run `~/llvm/Release/bin/llvm-lit path/to/single-test.ll` to run a single test. Is there a counterpart in ldc?

JohanEngelen:
`./runlit.py codegen` for all tests in `codegen/` indeed.
`./runlit.py codegen/align.d` for the single test.
But I think you already know this, so I don't understand your question.
(afaik we don't have a way to do this by using `ninja`)

gen/functions.cpp
kinke: > Is this a breaking change?

I meant wrt. to the 2nd attribute now not being emitted anymore, basically buying us nothing at all and diverging from what we and clang have been doing for years.

> It is not a breaking change for LLVM 3.9.

Again no code ref. ;) You do seem to be 100% sure about this.

> clang 10 will not have "no-frame-pointer-elim"="true" "no-frame-pointer-elim-non-leaf"

I've seen that clang 10 emits the new attribute and that's fine, my only worry is wrt. LLVM < 8.

MaskRay: > Again no code ref. ;) You do seem to be 100% sure about this.

Yes.

https://github.com/llvm/llvm-project/blob/release/3.9.x/llvm/lib/CodeGen/TargetOptionsImpl.cpp#L25
https://github.com/llvm/llvm-project/blob/release/3.9.x/llvm/lib/CodeGen/TargetFrameLoweringImpl.cpp#L32

MaskRay:
> `./runlit.py codegen` for all tests in `codegen/` indeed.
> `./runlit.py codegen/align.d` for the single test.
> But I think you already know this, so I don't understand your question.

Ah, yes, it works, though `codegen/attr_targetoptions.d` is not in the build directory (shell filename completion does not work).
```
% cd ~/Dev/ldc/Debug9/tests
% python3 ./runlit.py -vv codegen/attr_targetoptions.d
Lit version:  0.9.0
-- Testing: 1 tests, single process --
PASS: LDC :: codegen/attr_targetoptions.d (1 of 1)
Testing Time: 0.32s
  Expected Passes    : 1
```

> (afaik we don't have a way to do this by using `ninja`)

gen/functions.cpp
JohanEngelen: The assembly test should settle this, right?

MaskRay: I believe so.

```sh
cmake -H. -BDebug4 -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=$HOME/Stable/bin/clang -DCMAKE_CXX_COMPILER=$HOME/Stable/bin/clang++ -DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0 -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=lld -DCMAKE_SHARED_LINKER_FLAGS=-fuse-ld=lld -DD_COMPILER=$HOME/dlang/ldc-1.19.0/bin/ldmd2 -DCMAKE_CXX_STANDARD=14 -DLLVM_ROOT_DIR=$HOME/ccls/build/clang+llvm-4.0.1-x86_64-linux-gnu-debian8
```

```sh
cd ~/Dev/ldc/Debug4/tests
python3 ./runlit.py -vv codegen  # Expected Passes    : 114; Unsupported Tests  : 11
```

kinke:
Thx, looks quite good now. I'd suggest extracting the frame pointer tests to a dedicated test file (and then guarding that one with a `REQUIRES: target_X86`) and including the inline-asm special case in that test as well, à la:
```D
void externalFunc();
void inlineAsmLeaf() { asm { nop; } }
void inlineAsmNonLeaf() { asm { nop; } externalFunc(); }
```

MaskRay:
Ping 🤖 🤖 🤖

kinke:
Thx!

PR #3278, opened 2020-01-14
Slightly revise ldc-build-runtime & make --dFlags extend the default D flags

kinke:


kinke:
@jacob-carlborg: If you encounter more ldc-build-runtime annoyances, this is the file to patch. :)

PR #3288, opened 2020-01-19
Add support for iOS, tvOS and watchOS

jacob-carlborg:
I've only tested this on a 64 bit iPhone running iOS 12. 32 bit is obsolete and I don't have access to the other platforms. All unit tests in druntime pass and most in Phobos.

I'm not sure about the correct workflow here but I created PRs for the druntime and Phobos changes to upstream those. I assume those will get pulled in next time LDC updates to the latest frontend version.

There are one or two changes in druntime that should not be upstreamed. How should those be handled?

druntime PR: https://github.com/dlang/druntime/pull/2910.
Phobos PR: https://github.com/dlang/phobos/pull/7355.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -802,6 +802,22 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("AIX");
     VersionCondition::addPredefinedGlobalIdent("Posix");
     break;
+  case llvm::Triple::IOS:
+    VersionCondition::addPredefinedGlobalIdent("iOS");
+    VersionCondition::addPredefinedGlobalIdent("Posix");
+    VersionCondition::addPredefinedGlobalIdent("CppRuntime_Clang");
+    break;
+  case llvm::Triple::TvOS:
+    VersionCondition::addPredefinedGlobalIdent("TVOS");
+    VersionCondition::addPredefinedGlobalIdent("Posix");
+    VersionCondition::addPredefinedGlobalIdent("CppRuntime_Clang");
+    break;
+  case llvm::Triple::WatchOS:
+    VersionCondition::addPredefinedGlobalIdent("WatchOS");
+    VersionCondition::addPredefinedGlobalIdent("Posix");
+    VersionCondition::addPredefinedGlobalIdent("CppRuntime_Clang");
+    break;
+  break;
   default:
     if (triple.getEnvironment() == llvm::Triple::Android) {
       VersionCondition::addPredefinedGlobalIdent("Android");
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -817,7 +817,6 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("Posix");
     VersionCondition::addPredefinedGlobalIdent("CppRuntime_Clang");
     break;
-  break;
   default:
     if (triple.getEnvironment() == llvm::Triple::Android) {
       VersionCondition::addPredefinedGlobalIdent("Android");
--- a/gen/abi-aarch64.cpp
+++ b/gen/abi-aarch64.cpp
@@ -9,6 +9,7 @@
 //
 // The Procedure Call Standard can be found here:
 // http://infocenter.arm.com/help/topic/com.arm.doc.ihi0055b/IHI0055B_aapcs64.pdf
+// https://github.com/ARM-software/software-standards/blob/master/abi/aapcs64/aapcs64.rst
 //
 //===----------------------------------------------------------------------===//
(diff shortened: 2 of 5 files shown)


gen/abi-aarch64.cpp
```diff
@@ -44,7 +44,13 @@ struct AArch64TargetABI : TargetABI {
   IntegerRewrite integerRewrite;
 
   bool isVaList(Type *t) {
-    return t->ty == Tstruct && strcmp(t->toPrettyChars(true),
+    if (global.params.targetTriple->isOSDarwin())
+    {
+      return t->ty == Tpointer &&
+        static_cast<TypePointer*>(t)->next->ty == Tchar;
```
jacob-carlborg: I'm not sure if this is correct.

kinke:
Great, thx for the initiative. I assume you haven't run any C(++) interop tests from dmd-testsuite and thus haven't had to touch the ABI implementations?

Wrt. workflow - yes, druntime and Phobos will be merged before the next LDC release. We can also cherry-pick your stuff for v1.20 final, so that you can complete the 1-2 non-upstream changes earlier, at the cost of a probable merge conflict when advancing to v1.21.

jacob-carlborg:
> Great, thx for the initiative. I assume you haven't run any C(++) interop tests from dmd-testsuite and thus haven't had to touch the ABI implementations?

No, I haven’t run those. I didn’t think of that. How do I run those in the context of LDC?

kinke:
The normal workflow would be building a native compiler for the new platform first, and then simply invoking `ctest -V -R dmd-testsuite` in the build directory on a suitable host. But that's probably overkill at the current stage (need to cross-compile LDC first... - if interested, have a look at the Android specifics in https://github.com/ldc-developers/ldc/blob/master/.azure-pipelines/posix.yml; the native Android compilers are cross-compiled from x64 hosts in there).

So I'd test some cpp* tests manually - cross-compiling the C++ sources via clang, cross-compiling and -linking via LDC, then running the executable on a target device.

kinke:
> There’s no tool to build an executable with the tests for this

Nope, because the executable being tested is the compiler; it doesn't really matter whether it's a cross-compiler or native compiler though. But the point is that DMD's `runnable` tests need to be run natively.

jacob-carlborg:
> But the point is that DMD's runnable tests need to be run natively (and separating compilation/linking from execution probably isn't trivial with DMD's test framework).

Right. But in theory it should be possible to build `runnable` with a cross compiler and run it natively on the device?

kinke:
In theory, yes. :) - Compiling and linking all runnable tests (well, excluding the bash ones) incl. all of their permutations (at one stage, 64 permutations for xtest46.d...), then copying them to a device and running them.

JohanEngelen:
@jacob-carlborg The version identifiers should be reviewed and added to the predefined list:
https://dlang.org/spec/version.html#predefined-versions
PR example: https://github.com/dlang/dmd/pull/6994

jacob-carlborg:
> The version identifiers should be reviewed and added to the predefined list:
https://dlang.org/spec/version.html#predefined-versions
PR example: dlang/dmd#6994

The version identifiers were already added but I added missing tests and documentation:

https://github.com/dlang/dmd/pull/10743
https://github.com/dlang/dlang.org/pull/2740

driver/main.cpp
```diff
@@ -802,6 +802,22 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("AIX");
     VersionCondition::addPredefinedGlobalIdent("Posix");
     break;
+  case llvm::Triple::IOS:
+    VersionCondition::addPredefinedGlobalIdent("iOS");
+    VersionCondition::addPredefinedGlobalIdent("Posix");
+    VersionCondition::addPredefinedGlobalIdent("CppRuntime_Clang");
```
kinke: What about `darwin` then? To me, re-activating that deprecated predefined version is obviously better than having to define `Darwin` manually with the same 8 lines in a number of druntime and Phobos files. Even if Walter wants to restrict these predefines ones to the bare minimum, it's already there, has been deprecated for ages, and is still predefined for macOS with both compilers today.

jacob-carlborg: If you can convince him, sure. I have no interest in arguing with him.

kinke:
I've slightly revised it, incl. an ABI fix (no compiler magic should be required for `va_list` params on Darwin).

jacob-carlborg:
> I've slightly revised it, incl. an ABI fix (no compiler magic should be required for va_list params on Darwin).

Thanks.

jacob-carlborg:
Awesome, thanks.

PR #3294, opened 2020-01-23
ldc-build-runtime directories issue fix

### LINKED ISSUE #3293: ldc-build-runtime: WorkingDirScope isn't works properly?
```
env CC=arm-none-eabi-gcc ~/Dev/ldc/build-ldc/bin/ldc-build-runtime --ninja --dFlags="-w;-de;-dip1000;-march=thumb;-mcpu=cortex-m3" --ldcSrcDir=$PWD/../
```

```
.: Creating build directory: ldc-build-runtime.tmp
.: Invoking: cmake -DLDC_EXE_FULL=/home/denizzz/Dev/ldc/build-ldc/bin/ldc2 -DD_VERSION=2 -DDMDFE_MINOR_VERSION=0 -DDMDFE_PATCH_VERSION=90 '-DD_EXTRA_FLAGS=-w;-de;-dip1000;-march=thumb;-mcpu=cortex-m3' -G Ninja /home/denizzz/Dev/ldc/runtime/../runtime
-- Configuring done
-- Generating done
-- Build files have been written to: /home/denizzz/Dev/ldc/runtime
.: Invoking: ninja all
ninja: error: loading 'build.ninja': No such file or directory
.: Error: command failed with status 1
```

ldc-build-runtime.tmp/ created, but it is empty.
build.ninja created in current directory.

denizzzka:
fix #3293

### DIFF
--- a/runtime/ldc-build-runtime.d.in
+++ b/runtime/ldc-build-runtime.d.in
@@ -114,85 +114,93 @@ void prepareLdcSource() {
 
     // Download & extract LDC source archive if <buildDir>/ldc-src/runtime doesn't exist yet.
 
-    const wd = WorkingDirScope(config.buildDir);
+    workingDirScope(config.buildDir, (){
 
-    auto ldcVersion = "@LDC_VERSION@";
-    void removeVersionSuffix(string beginning) {
-        const suffixIndex = ldcVersion.countUntil(beginning);
-        if (suffixIndex > 0)
-            ldcVersion = ldcVersion[0 .. suffixIndex];
-    }
-    removeVersionSuffix("git-");
-    removeVersionSuffix("-dirty");
-
-    import std.format : format;
-    const localArchiveFile = "ldc-src.zip";
-    if (!localArchiveFile.exists) {
-        const url = "https://github.com/ldc-developers/ldc/releases/download/v%1$s/ldc-%1$s-src.zip".format(ldcVersion);
-        writefln(".: Downloading LDC source archive: %s", url);
-        import std.net.curl : download;
-        download(url, localArchiveFile);
-        if (getSize(localArchiveFile) < 1_048_576) {
-            writefln(".: Error: downloaded file is corrupt; has LDC v%s been released?", ldcVersion);
-            writefln("   You can work around this by manually downloading a src package and moving it to: %s",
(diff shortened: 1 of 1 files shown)


kinke:
I don't see how this would be an improvement. Please try to figure out what the actual issue is.

denizzzka:
This is bug fix, not improvement

Sometimes WorkingDirScope destructor called before exit of scope. Maybe this is because some lowlevel optimizations

kinke:
> Sometimes WorkingDirScope destructor called before exit of scope

Well that's of course not what should happen, so that needs to be debugged. That said, I've used ldc-build-runtime a number of times and never experienced any problem of that sort. It's also used for CI and never caused any trouble.

denizzzka:
For me, first calls of WorkingDirScope ctor works ok, but it fails on build()

Maybe it is need to add something like `scope(exit) destroy wd` after each WorkingDirScope creation? 

Anyway, I think my varian better and clearer because now you clearly see scope where directory will be changed

kinke:
> Maybe it is need to add something like scope(exit) destroy wd after each WorkingDirScope creation?

Nope, the dtor is obviously supposed to run when exiting the scope, that's what RAII is all about. If it doesn't, then that's a huge issue, not just for ldc-build-runtime. IIRC, ldc-build-runtime is built by the D host compiler used to build LDC, so which one are you using?

denizzzka:
> so which one are you using?

LDC from `~ldc`

(It seems to me if the optimizer optimizes call of the destructor early then this does not violate language standard.)

kinke:
Nope, the optimizer cannot reorder the dtor call, it has side effects. And it of course doesn't explain why only you seem to be seeing this issue. Adding some writeln output (+ stdout.flush) in struct ctor + dtor should make clear what happens.

denizzzka:
> Well that's of course not what should happen, so that needs to be debugged.

@kinke you were right! I added debug output into ctor and dtor and found what WorkingDirScope destructor is never called!

affected:
version   1.18.0 (DMD v2.088.1, LLVM 9.0.0)

not affected:
version   1.20.0-git-941542d (DMD v2.090.0, LLVM 7.0.1)

So, this issue is already fixed. I'm sorry to trouble you.

PR #3295, opened 2020-01-24
WebAssembly: Add predefined versions WASI & CRuntime_WASI and emit TLS globals as regular ones

kinke:


kinke:
Pinging @skoppe.

skoppe:
Thank you very much! Might be a small change for you but it would have taken me ages.

Looks good to me.

gen/llvmhelpers.cpp
```diff
@@ -1755,6 +1755,12 @@ llvm::GlobalVariable *declareGlobal(const Loc &loc, llvm::Module &module,
                                     llvm::Type *type,
                                     llvm::StringRef mangledName,
                                     bool isConstant, bool isThreadLocal) {
+  // No TLS support for WebAssembly; spare users from having to add __gshared
+  // everywhere.
```
JohanEngelen: Let's add an IR test for this?

kinke:
> Might be a small change for you but it would have taken me ages.

Exactly my thinking, plus I thought you might slowly return from vacation, so here's a little motivational support. ;)

kinke:
Btw, clang uses 128-bit quadruple precision for `long double` and WASI targets (LDC uses double for `real`): https://cpp.godbolt.org/z/uTXMU8

kinke:
@skoppe: Would you mind taking care of the official predefined version(s) upstream? E.g., whether there should be a `WASI` version when targeting it as 'OS' and/or `CRuntime_WASI` etc. I lack the insight for proper arguing...

skoppe:
Sure. It hasn't been totally clear for me either. To be honest, I have ignored the issue until I really can't anymore, hoping I would have gained some insights by then. Johan brought up some good points on the NG and now you are pushing it so it seems this will need to be crystallised rather sooner than later.

At some point we need to cut out libc, but it seems we are tied to it for the foreseeable future, so a `CRuntime_WASI` version makes complete sense. We can get rid of it when druntime will directly call WASI, but I don't see that happening soon.

We also need the WASI version for cases where we need to do WASI stuff which isn't dependent on the libc we have in-between, but also as a future place to put code that directly calls WASI apis (there are a handful of cases already).

Right now, both `WASI` and `CRuntime_WASI` will be defined when we compile for `wasm32-unknown-wasi`.

kinke:
Thanks! I've added `WASI` here.

skoppe:
Regardless, can this be merged? Or do you first want me to try it out?

kinke:
It's ready and tested, the only thing left to do is to make the versions official. Ideally, that would happen soon, so that this can make it into v1.20 final with a proper release notes entry.

skoppe:
Ahh, and for that they need to be in upstream. I see. Will do tomorrow.

skoppe:
See https://github.com/dlang/dmd/pull/10778 for the versions. I hope that suffices.

PR #3298, opened 2020-01-25
Add Cirrus CI, drop Semaphore

kinke:


kinke:
That speed is just plain awesome - 18 minutes for a full multilib testrun on Linux, ~16 minutes for a macOS job. - It's currently basically a clone of CircleCI with an additional FreeBSD job, which is still allowed to fail (but there's only a very small number of failures).

JohanEngelen:
Nice @kinke

kinke:
Thx, appreciated.

PR #3303, opened 2020-01-27
Phobos: Cherry-pick soft-float improvements

kinke:
dlang/phobos#7362 and dlang/phobos#7366


kinke:
dlang/phobos#7377 is also required, then `std.math` can be compiled successfully, incl. `-unittest`, with `-mtriple=thumb-linux-gnueabi -mcpu=cortex-m3 -float-abi=soft`.

Issue #3299, opened 2020-01-26
Data definition directives inside inline asm are not supported yet.


etcimon:
I'm trying to get Botan to run at its fastest with LDC support, and I think solving this `db` instruction would help with the lack of CPU instructions support for now.

```
PS C:\devpriv\botan> dub test
Generating test runner configuration 'botan-test-full' for 'full' (staticLibrary).
Performing "unittest" build using C:\D2\ldc2\bin\ldc2.exe for x86_64.
botan-math 1.0.3+commit.1.g559d6bf: target for configuration "library" is up to date.
memutils 1.0.1: target for configuration "secure" is up to date.
botan 1.12.11: building configuration "botan-test-full"...
Enhanced memory security is enabled.
Memory debugger enabled
source\botan\entropy\rdrand.d(109,32): Error: Data definition directives inside inline asm are not supported yet.
source\botan\utils\simd\wmmintrin.d-mixin-193(200,39): Error: unknown opcode PCLMULQDQ
source\botan\modes\aead\gcm.d(520,41): Error: template instance botan.utils.simd.wmmintrin._mm_clmulepi64_si128!"0x00" error instantiating
source\botan\utils\simd\wmmintrin.d-mixin-193(200,39): Error: unknown opcode PCLMULQDQ
source\botan\modes\aead\gcm.d(521,41): Error: template instance botan.utils.simd.wmmintrin._mm_clmulepi64_si128!"0x01" error instantiating
source\botan\utils\simd\wmmintrin.d-mixin-193(200,39): Error: unknown opcode PCLMULQDQ
source\botan\modes\aead\gcm.d(522,41): Error: template instance botan.utils.simd.wmmintrin._mm_clmulepi64_si128!"0x10" error instantiating
source\botan\utils\simd\wmmintrin.d-mixin-193(200,39): Error: unknown opcode PCLMULQDQ
source\botan\modes\aead\gcm.d(523,41): Error: template instance botan.utils.simd.wmmintrin._mm_clmulepi64_si128!"0x11" error instantiating
```

kinke:
Looking at that [wmmintrin.d](https://github.com/etcimon/botan/blob/1fed591592c32d476f8ef8f0ff1610a38650006e/source/botan/utils/simd/wmmintrin.d), it looks like the simplest solution is to import the gcc/gdc-compatible builtins from `ldc.gccbuiltins_x86` and enabling the `version(GDC)` path for LDC too.

kinke:
[...]

PR #3304, opened 2020-01-29
Add support for GCC/GDC-style inline asm syntax

kinke:


kinke:
Pinging @ibuclaw - just letting you know your work also benefits LDC and so portability across our compilers (and, well, across C(++) and D too!). Thanks a lot for paving the way for a sane generic asm syntax!

dnadlinger:
The syntax itself isn't very D-like, though, what with the colon-delimited arguments in a `{}` block…

kinke:
Hmm, I don't think that would noticeably improve things (just putting curly braces around all arguments if I understand correctly) while deviating from the gcc syntax, which presumably most interested users are familiar with already.

---

This is most likely going to remain a reasonable-effort solution, just like LLVM's support for gcc constraint codes itself. E.g., clang doesn't even support the `a,b,c,d` names for x86 (I do here because already used in druntime for GDC). I tried to be quite fault-tolerant wrt. constraints, e.g., LLVM-style `*m` is treated just like `m`, outputs don't need to start with `=` and clobbers can start with a `~` prefix.

etcimon:
I'd like to add the bounty here but we don't have any way of adding the issue to DMD. Did you agree to add GCC inline assembly support for DMD as well to create a true compatible compiler state?

kinke:
I don't plan to spend a single second with the DMD backend, as that'd be nothing but lost time IMO.

etcimon:
Ok I understand, I'm finding it hard to keep up with the compiler bugs in DMD as well so I'll align myself with what you're saying and make LDC the main compiler for my projects

kinke:
It's not just the bugs, but mainly its limited scope and usefulness - firstly x86 only, and secondly only really useful for fast unoptimized code generation. `-O -inline` is slower to compile than LDC with equivalent settings while producing obviously slower code. So to me, the DMD backend is clearly a lost cause in the long run.

etcimon:
Wouldn't the goal be to make LDC the default compiler for newcomers and, to phase out DMD? If everyone would contribute to the same compiler it would probably be better to the community at this point

kinke:
Well we need DMD's frontend, and that's still the central part of the dmd repo. Front/backend and glue layer just aren't clearly separated. E.g., I think most dmd constributors don't know that some fix in `e2ir.d` doesn't affect LDC or GDC and so how to prefer pure frontend fixes benefitting all compilers.

etcimon:
> PS: That issue was #3299.

I asked them to remove the additional bounty from that issue because it didn't represent the WIP. I'm only waiting for you to create a tracker/issue to re-assign the bounty properly.

It's also been obvious to me that their support is much better than @dnadlinger claims, maybe because the bounty is a high amount.

kinke:
I think we're talking past each other. ;) - This PR here adds support for GDC-style asm to LDC and is basically finished; the effort was small as expected (I just haven't foreseen having to translate the template string and working around limited LLVM support for GCC constraint codes).

Supporting this syntax in DMD too (so that all 3 D compilers support it) is most likely a whole different beast, but that has only to do with DMD's backend/glue layer and as such is of absolutely no interest to me. If you'd like to see the syntax supported in DMD too, then a DMD bounty might help, but has nothing to do with LDC.

> It's also been obvious to me that their support is much better

It'd better be, considering they take 10%.

dnadlinger:
@kinke:
> Hmm, I don't think that would noticeably improve things […]

This was a misunderstanding – I was merely commenting on the current form of the GDC inline assembly syntax as an aside. I don't think it is tastefully chosen. The DMD-style `asm` syntax uses block as an argument. This makes sense syntactically, as the body is close to D code (no extra quoting, semicolon-delimited, can reference D symbols, …). The GCC C extension uses syntax that resembles a function call/macro, `__asm__( … : … : … : … )`. There is a bit of ad-hoc syntax in the colons to delimit the operands/clobbers, but fair enough. GDC now for reasons unknown uses the curly braces from the DMD syntax, but otherwise keeps GCC's C syntax, with the result loooking quite unnatural to the D-trained eye (Why the quotes inside the curly braces? Why are the clobbers just two values mushed together (`"=a" b, "=b" s, "=c"`)? Etc.).

kinke:
> Maybe positional args could be enough

An AA literal syntax (for out/inputs, and an array literal for clobbers) could be another option: `asm("nop", ["=r": output], ["r": input], ["~rax"]);`. It would look ugly for simple clobber-only cases though: `asm("nop", null, null, ["~a"]);`

ibuclaw:
With all due apologies...  the short version is that as it's been this way since 2004, this is a syntax of a historical note more than anything else.

> GDC now for reasons unknown uses the curly braces from the DMD syntax, but otherwise keeps GCC's C syntax, with the result loooking quite unnatural to the D-trained eye

I suspect  that rather than it being specifically planned that way, the reason is more that it is partially incidental, and partially so as not to interfere too much with parse.c.  The handling was added right in the middle of the parser for DMD-style asm.

https://github.com/D-Programming-GDC/gdc/blob/b25ac31e995c3dc598d7c87c4bc4cf9366c7ac8e/gcc/d/dfrontend/parse.c#L5777-L5786

And the patch that was accepted to dmd uses parse.d as-is, so by the time it reaches asmSemantic (where the deviation from dmd now occurs) it has already been parsed as `asm { tokens; ... }`.

> Why the quotes inside the curly braces? Why are the clobbers just two values mushed together ("=a" b, "=b" s, "=c")?

1: Quotes because it is a instruction template string.
2: I wouldn't object to adding a `check(TOK.leftBracket)`/`check(TOK.rightBracket)` before and after parseAssignExp, first as a deprecation though. https://github.com/dlang/dmd/blob/f7aaa7d1446e4e96aee5c287b2ba9e74437378ec/src/dmd/iasmgcc.d#L82-L84

> This is not really that relevant here if we want to commit to supporting GDC-style syntax for compatibility reasons anyway. It would just be slightly disappointing if this syntax ended up being what the community rallies around…

The parser implementation is in dmd, anyone can modify it as they wish.  So long as the current fields in GccAsmStatement don't change, there's not going to be much room for complaint.

ibuclaw:
> Ouch, AA literal with duplicate keys - ["=r": var1, "=r": var2]...

I don't see that as a problem, as internally AA literals are a pair of arrays (one for keys, another for values), and no attempt is done at compile-time to "merge" duplicates.

ibuclaw:
> An AA literal syntax (for out/inputs, and an array literal for clobbers) could be another option:

Saying that though, a good litmus test would be the following:

```
asm pure nothrow @nogc {
    "cmoveq %1, %2, %[result]"
    : [result] "=r" (result)
    : "r" (test), "r" (newvalue), "[result]" (oldvalue)
    : "memory";
}
```
Any proposed syntax should allows for symbolic names to also fit in is a must as well (AAs don't do this).

PR #3305, opened 2020-01-31
Fix StringRef conversion for LLVM 11 (777180a32b61070a10dd330b4f038bf24e916af1)

MaskRay:
https://github.com/llvm/llvm-project/commit/777180a32b61070a10dd330b4f038bf24e916af1

### DIFF
--- a/driver/configfile.cpp
+++ b/driver/configfile.cpp
@@ -96,7 +96,7 @@ bool ConfigFile::locate(std::string &pathstr) {
   {                                                                            \
     sys::path::append(p, filename);                                            \
     if (sys::fs::exists(p.str())) {                                            \
-      pathstr = p.str();                                                       \
+      pathstr = {p.data(), p.size()};                                          \
       return true;                                                             \
     }                                                                          \
   }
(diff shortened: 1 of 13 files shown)


driver/configfile.cpp
```diff
@@ -96,7 +96,7 @@ bool ConfigFile::locate(std::string &pathstr) {
   {                                                                            \
     sys::path::append(p, filename);                                            \
     if (sys::fs::exists(p.str())) {                                            \
-      pathstr = p.str();                                                       \
+      pathstr = {p.data(), p.size()};                                          \
```
kinke: Has `llvm::SmallString<N>::str()` been removed as well?

MaskRay: You will need `std::string(p.str())`. I use a pair of `data` and `size` to reduce dependency on StringRef utilities that are not available in std::string_view (in case some day StringRef is migrated to std::string_view)

kinke: Ah okay, OTOH I thought it'd return a std::string.

PR #3307, opened 2020-02-04
Upgrade LDC-LLVM to v10.0.0-rc4 and bump Android NDK to r21

kinke:
The 1st LTS version of the NDK IIRC.

Noteworthy change: Neon instructions are enabled by default for 32-bit ARMv7 (in the CMake toolchain file). I think we've already enabled them for generated code (and the D parts of the compiler itself) via `-mcpu=cortex-a8`, so the druntime/Phobos C parts will use them now too, as well as the LDC C++ parts (excl. prebuilt LLVM, whose NDK will be upgraded next time I look at it).


kinke:
This new LLVM 10 build is a couple hours old (RC2+).

kinke:
The Azure pipeline isn't triggered automatically anymore for a couple of days; I've tried different things, nothing worked, and now there's no Webhook anymore (the llvm-project doesn't have one either, but there, the GitHub integration is the 'normal' one via 'checks', presumably via the GitHub 'app'). I can only guess that the GitHub app config needs to be reset/updated, but I don't have access to it. Pinging @dnadlinger and @redstar.

kinke:
The Azure issue seems to be an extremely embarrassing issue on Microsoft's side - https://status.dev.azure.com/_event/179641421.

PR #3313, opened 2020-02-08
MLIR: Add Support to compile flags output MLIR file

Robertorosmaninho:


driver/tomlirfile.cpp
```diff
@@ -0,0 +1,92 @@
+//
+// Created by Roberto Rosmaninho on 09/10/19.
+//
```
thewilsonator: please  follow the format used by the rest of the cpp files

driver/main.cpp
```diff
@@ -503,6 +506,10 @@ void parseCommandLine(Strings &sourceFiles) {
                strcmp(ext, global.s_ext.ptr) == 0) {
       global.params.output_s = OUTPUTFLAGset;
       global.params.output_o = OUTPUTFLAGno;
+    }else if(opts::output_mlir.getNumOccurrences() == 0 &&
```
thewilsonator: `} else if (`, please add spaces

driver/tomlirfile.cpp
```diff
@@ -0,0 +1,92 @@
...
+      fatal();
+    }
+    mlir::OwningModuleRef module = ldc_mlir::mlirGen(mlirContext, m, irs);
+    if(!module){
+      IF_LOG Logger::println("Cannot write MLIR file to '%s'", llpath.c_str());
+      fatal();
+    }
+    module->print(aos);
+  }
+}
```
thewilsonator: please add a newline.

Robertorosmaninho: Solved

driver/main.cpp
Robertorosmaninho: Done

driver/tomlirfile.cpp
Robertorosmaninho: Solved here and in tomlirfile.h

JohanEngelen:
> You will have to put all of your code behind a #ifdef for versions of LDC that are not built with MLIR enabled (i.e. old verions of LLVM and those built without MLIR support).

You can restrict #ifdef to the stuff that actually requires MLIR library calls. The rest of the code (like the commandline parsing) you can keep, and then let LDC output an error message saying that MLIR is not supported in case LDC was built without it (for which you do need to #ifdef stuff ;-) ).

driver/main.cpp
```diff
@@ -85,6 +85,7 @@ void gendocfile(Module *m);
 void generateJson(Modules *modules);
 
 using namespace opts;
+using namespace ldc_mlir;
```
JohanEngelen: what's this needed for?

driver/main.cpp
```diff
@@ -433,11 +434,13 @@ void parseCommandLine(Strings &sourceFiles) {
 
   global.params.output_o =
       (opts::output_o == cl::BOU_UNSET &&
-       !(opts::output_bc || opts::output_ll || opts::output_s))
+       !(opts::output_bc || opts::output_ll || opts::output_s ||
+       opts::output_mlir))
```
JohanEngelen: i'm surprised at the whitespacing. Is this manual or clang-formatted?

driver/main.cpp
```diff
@@ -503,6 +506,10 @@ void parseCommandLine(Strings &sourceFiles) {
                strcmp(ext, global.s_ext.ptr) == 0) {
       global.params.output_s = OUTPUTFLAGset;
       global.params.output_o = OUTPUTFLAGno;
+    } else if (opts::output_mlir.getNumOccurrences() == 0 &&
+             strcmp(ext, global.mlir_ext.ptr) == 0) {
```
JohanEngelen: same whitespace issue here. If clang-format, then OK.

driver/tomlirfile.h
```diff
@@ -0,0 +1,23 @@
+//===-- driver/tomlirfile.h - MLIR file emission ----------------*- C++ -*-===//
+//
+//                         LDC – the LLVM D compiler
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+//
+// Handles emission of "finished" MLIR modules to on-disk object files.
```
JohanEngelen: maybe you can explain here what you mean with "finished"

driver/tomlirfile.cpp
```diff
@@ -0,0 +1,102 @@
...
+        llvm::SmallString<128> buffer(filename);
+        llvm::sys::path::replace_extension(buffer,
+                                           llvm::StringRef(ext.ptr, ext.length));
+        return buffer.str();
+      };
+
+  //Write MLIR
+  if(global.params.output_mlir) {
+    const auto llpath = replaceExtensionWith(global.mlir_ext);
+    Logger::println("Writting MLIR to %s\n", llpath.c_str());
```
JohanEngelen: Writing typo

driver/tomlirfile.cpp
```diff
@@ -0,0 +1,102 @@
...
+    std::error_code errinfo;
+    llvm::raw_fd_ostream aos(llpath, errinfo, llvm::sys::fs::F_None);
+    if(aos.has_error()){
+      error(Loc(), "Cannot write MLIR file '%s':%s", llpath.c_str(),
```
JohanEngelen: nicer to put a space between `:` and `%s`.

driver/tomlirfile.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+#endif
+
+#include "gen/logger.h"
+#include "mlir/IR/Module.h"
+#include "dmd/globals.h"
+#include "gen/MLIR/MLIRGen.h"
+#include "dmd/expression.h"
+
+void writeMLIRModule(Module *m, mlir::MLIRContext &mlirContext,
```
JohanEngelen: I understand this is a copy of toobj.cpp, tailored for MLIR. How much tailoring is really needed here? Why not add it to toobj.cpp and keep the file writing in one location?

driver/tomlirfile.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+#include "dmd/errors.h"
+#include "driver/cl_options.h"
+#include "driver/cache.h"
+#include "driver/targetmachine.h"
+#include "driver/tool.h"
+#include "gen/irstate.h"
+#include "gen/logger.h"
+#include "gen/optimizer.h"
+#include "llvm/IR/AssemblyAnnotationWriter.h"
```
JohanEngelen: This very long #include list seems excessive.

driver/main.cpp
Robertorosmaninho: In this file, nothing, but in the project, this is the namespace that translates D into MLIR Dialect. I'll remove it here! Thanks for the advice.

driver/main.cpp
Robertorosmaninho: I set my environment to wrap into 80 columns but I can change this.

driver/main.cpp
Robertorosmaninho: Same 80 columns.

driver/tomlirfile.h
Robertorosmaninho: by "finished" I mean all D code translated into MLIR

driver/tomlirfile.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "llvm/IR/Verifier.h"
+#include "llvm/Analysis/ModuleSummaryAnalysis.h"
+#if LDC_LLVM_VER >= 400
```
thewilsonator: if MLIR is enabled then we know that `LDC_LLVM_VER >= 1000` so this is always true.

driver/tomlirfile.cpp
```diff
@@ -0,0 +1,102 @@
...
+#endif
+#include "llvm/IR/LegacyPassManager.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/FormattedStream.h"
+#include "llvm/Support/Program.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Target/TargetMachine.h"
+#include "llvm/Analysis/TargetTransformInfo.h"
+#if LDC_LLVM_VER >= 600
```
thewilsonator: ditto.

driver/tomlirfile.cpp
```diff
@@ -0,0 +1,102 @@
...
+#else
+#include "llvm/Target/TargetSubtargetInfo.h"
+#endif
+#include "llvm/Transforms/Utils/Cloning.h"
+#include "llvm/IR/Module.h"
+#include <cstddef>
+#include <fstream>
+#include <mlir/IR/Module.h>
+
+#ifdef LDC_LLVM_SUPPORTED_TARGET_SPIRV
```
thewilsonator: this is unneeded. I will be rewriting the dcompute support to use MLIR when all of this is merged.

driver/tomlirfile.cpp
Robertorosmaninho: Indeed

driver/codegenerator.cpp
```diff
@@ -255,6 +256,7 @@ void CodeGenerator::finishLLModule(Module *m) {
     insertBitcodeFiles(ir_->module, ir_->context(), global.params.bitcodeFiles);
   }
 
+  writeMLIRModule(m, mlirContext_, m->objfile.toChars(), ir_);
```
thewilsonator: `mlirContext_` is only defined for `LDC_MLIR_ENABLED` but used unconditionally here.

Robertorosmaninho: solved

driver/tomlirfile.cpp
Robertorosmaninho: I had a lot of problems with ambiguity even with Module, so it was simpler to create a new file

driver/main.cpp
JohanEngelen: please use clang-format for formatting. In this case, the problem is not wrap at 80 columns, but that `opts::output_mlir` is aligned with `!` instead of with `opts::output_bc`.

driver/main.cpp
JohanEngelen: Not 80 column problem, but wrong alignment of `strcmp`

driver/tomlirfile.cpp
JohanEngelen: It is a big maintainability problem to separate things that don't have to be separated. Ambiguity problems should be solved by simply writing out the full name with namespace. If this was the only reason, please add this code to toobj.cpp (like all the .s and .ir outputs) and we'll see how bad it gets.

driver/main.cpp
Robertorosmaninho: Ok

CMakeLists.txt
```diff
@@ -373,6 +373,7 @@ set(DRV_SRC
     driver/exe_path.cpp
     driver/targetmachine.cpp
     driver/toobj.cpp
+    driver/tomlirfile.cpp
```
thewilsonator: You still got those files in CMake

CMakeLists.txt
```diff
@@ -399,6 +400,7 @@ set(DRV_HDR
     driver/linker.h
     driver/plugins.h
     driver/targetmachine.h
+    driver/tomlirfile.h
```
thewilsonator: ditto.

CMakeLists.txt
Robertorosmaninho: solved

CMakeLists.txt
Robertorosmaninho: solved

driver/toobj.cpp
```diff
@@ -472,3 +473,44 @@ void writeModule(llvm::Module *m, const char *filename) {
     }
   }
 }
+
+#if LDC_MLIR_ENABLED
+void writeMLIRModule(::Module *m, mlir::MLIRContext &mlirContext,
```
JohanEngelen: why is this code not part of `writeModule`? (see e.g. line 415).

Robertorosmaninho: Just to have it in different functions, but I can also put it on ```writeModule```, the problem is that I'll need some extra arguments on this function, so the function should be something like: 
```c++
void writeModule(llvm::Module *m, const char *filename, ::Module *module,
    mlir::MLIRContext &mlirContext, IRState *irs);
```
In addition, ```writeAndFreeLLModule```from CodeGenerator must also be changed to receive a dmodule as an argument to be used on ```write module```

JohanEngelen:
@Robertorosmaninho I looked more into this. I think I was confused and actually it looks like you have to fully rewrite the IR building code. I thought the same code can output IR _and_ MLIR, but now it appears to me that those are completely separate IR build libraries. So you'd have to rewrite all D AST->MLIR generation from scratch. Correct?  In that case, there is no point in trying to merge things into `toobj.cpp`. You have to create a new codegenerator. (this is different from DCompute codegen)

Robertorosmaninho:
@JohanEngelen  I already did that, actually. This PR only introduces the creation of an empty ".mlir" file. My version of CodeGen is available[ here ](https://github.com/Robertorosmaninho/ldc/tree/Introducing-MLIR-translation), I'm preparing this branch to send it as a PR here. Actually, I'm just waiting for this PR to be approved and the revision of @thewilsonator to send it.

JohanEngelen:
OK cool.
What is the expected flow in the future?
I'm thinking: D code --> MLIR --> LLVM IR --> object file. Right? In that case, the code should be set up in such a way that you can feed MLIR module output into an LLVM IR module, and then run through the current exist LLVM IR module output path.

Note that the function `void writeModule(llvm::Module *m, const char *filename);` does not do codegen, and neither should `writeMLIRModule`. That function should therefore not need a D module as argument.

I'd expect the code to look more like this:
```d
llvm::Module irModule;
if (useMLIR) {
   mlir::Module m = codegenMLIRModule(dmodule);
   writeMLIRModule(m); // optionally outputs an mlir file
   irModule = m.generateLLVMIRModule();
} else {
   irModule = codegenIRModule();
}
writeIRModule(irModule);
```

Robertorosmaninho:
> OK cool.
> What is the expected flow in the future?
> I'm thinking: D code --> MLIR --> LLVM IR --> object file. Right? In that case, the code should be set up in such a way that you can feed MLIR module output into an LLVM IR module, and then run through the current exist LLVM IR module output path.

It's something like that. Although, my intention is only to translate D Code --> MLIR and then the user can set up the way that he/she wants to translate it using `mlir-opt`. This point is valid due to a large number of Dialects(IR inside MLIR) options to translate the DDialect and each target has different optimizations and different purposes, translating it directly and obligatorily to the LLVM IR is losing many of the benefits of MLIR framework.  
 
> Note that the function `void writeModule(llvm::Module *m, const char *filename);` does not do codegen, and neither should `writeMLIRModule`. That function should therefore not need a D module as argument.
>
> I'd expect the code to look more like this:
> 
> ```d
> llvm::Module irModule;
> if (useMLIR) {
>    mlir::Module m = codegenMLIRModule(dmodule);
>    writeMLIRModule(m); // optionally outputs an mlir file
>    irModule = m.generateLLVMIRModule();
> } else {
>    irModule = codegenIRModule();
> }
> writeIRModule(irModule);
> ```

Where do you expect this code to be placed?

JohanEngelen:
> Where do you expect this code to be placed?

Maybe in `main.cpp:codegenModules`. That function is a little bit of a mess (hard to see the structure).

JohanEngelen:
I'm sorry if I only added to more confusion, and led you down some dead ends. I'm trying to not have this make the codebase more of a mess ;-)
So perhaps it is indeed best to keep things the way you had it, as separate as possible from the current code path. That way you don't benefit from details (things like "emitLLVMUsedArray" or ABI issues) of which there are a lot; but perhaps they are not needed at this point with MLIR. I think it will be a very long way before we can use the D-->MLIR-->IR-->asm path as a replacement for the non-MLIR path, mainly because of all these particular details.

Robertorosmaninho:
> > Where do you expect this code to be placed?
> 
> Maybe in `main.cpp:codegenModules`. That function is a little bit of a mess (hard to see the structure).

I'll try to do something like this

Robertorosmaninho:
> I'm sorry if I only added to more confusion, and led you down some dead ends. I'm trying to not have this make the codebase more of a mess ;-)
> So perhaps it is indeed best to keep things the way you had it, as separate as possible from the current code path. That way you don't benefit from existing code to deal with tricky details (things like "emitLLVMUsedArray" or ABI issues) of which there are a lot; but perhaps they are not needed at this point with MLIR. I think it will be a very long way before we can use the D-->MLIR-->IR-->asm path as a replacement for the non-MLIR path, mainly because of all these particular details.

Don't worry, I'm here to help the compiler to grown in its own way and learn with it :D 
In addition, I intend to continue expanding the compatibility with D and publish a paper with my improvements and the possible optimizations, this may help the community to adopt MLIR and to help me to improve the compiler in this way.

Robertorosmaninho:
@thewilsonator could you see what i'm doing wrong? All tests are failing to generate runtime stuff that doesn't seem to be related to ldc2 binary.

kinke:
> llvm-ar: runtime/objects/core/atomic.bc.o: No such file or directory.

With CMake option `BUILD_LTO_LIBS=ON`, we build the runtime objects as binary and bitcode at the same time (`atomic.o` and `atomic.bc.o`), using `-output-o -output-bc -flto=thin`, so it looks like your changes broke that.

Robertorosmaninho:
> > llvm-ar: runtime/objects/core/atomic.bc.o: No such file or directory.
> 
> With CMake option `BUILD_LTO_LIBS=ON`, we build the runtime objects as binary and bitcode at the same time (`atomic.o` and `atomic.bc.o`), using `-output-o -output-bc -flto=thin`, so it looks like your changes broke that.

Thanks! I'll solve that.

driver/codegenerator.cpp
```diff
@@ -341,4 +349,73 @@ void CodeGenerator::emit(Module *m) {
...
+  //Write MLIR
+  if(global.params.output_mlir) {
+  const auto llpath = replaceExtensionWith(global.mlir_ext);
+  Logger::println("Writting MLIR to %s\n", llpath.c_str());
+  std::error_code errinfo;
+  llvm::raw_fd_ostream aos(llpath, errinfo, llvm::sys::fs::F_None);
+  if (aos.has_error()) {
+    error(Loc(), "Cannot write MLIR file '%s':%s", llpath.c_str(),
+          errinfo.message().c_str());
+    fatal();
```
thewilsonator, line 413: Indent

.circleci/config.yml
```diff
@@ -148,4 +148,4 @@ workflows:
       - Ubuntu-18.04-multilib-rtSanitizers
       - Ubuntu-19.10-multilib-sharedLibsOnly-gdmd
       - macOS-x64
-      - macOS-x64-sharedLibsOnly
+      - macOS-x64-sharedLibsOnly
```
thewilsonator, line 151: Please restore the newline.

Robertorosmaninho: ok

driver/codegenerator.cpp
Robertorosmaninho: Is it better now?

thewilsonator: Yes.

.circleci/config.yml
Robertorosmaninho: fixed?

thewilsonator: Yes, If I mark the conversation resolved, then the comments no longer apply.

driver/codegenerator.cpp
```diff
@@ -274,7 +283,6 @@ void CodeGenerator::finishLLModule(Module *m) {
   if (moduleCount_ == 1) {
     insertBitcodeFiles(ir_->module, ir_->context(), global.params.bitcodeFiles);
   }
-
```
thewilsonator, line 277: restore this.

Robertorosmaninho: done

.circleci/config.yml
```diff
@@ -148,4 +148,4 @@ workflows:
       - Ubuntu-18.04-multilib-rtSanitizers
       - Ubuntu-19.10-multilib-sharedLibsOnly-gdmd
       - macOS-x64
-      - macOS-x64-sharedLibsOnly
```
kinke, line 151: What's this about?

cmake/Modules/FindMLIR.cmake
```diff
@@ -28,7 +28,16 @@ else()
...
 
     # To be done: add the required MLIR libraries. Hopefully we don't have to manually list all MLIR libs.
-    set(MLIR_LIBRARIES "")
+    
+    if(EXISTS "${MLIR_LIB_DIR}/MLIRIR.lib")
+      message("Found MLIRIR.lib")
+      set(MLIR_LIBRARIES MLIRIR.lib MLIRSupport.lib)
+    elseif(EXISTS "${MLIR_LIB_DIR}/libMLIRIR.a")
+      message("Found libMLIR.lib")
+      set(MLIR_LIBRARIES libMLIRIR.a libMLIRSupport.a)
```
kinke, line 37: Very glad this is such a short list now. :+1:

kinke:
The druntime submodule hash should be restored as well.

.circleci/config.yml
Robertorosmaninho: Nothing important, I'll restore it right now

cmake/Modules/FindMLIR.cmake
Robertorosmaninho: Thanks, although the list needs to grow when new features appear. These 2 libraries only support MLIRContext.

Robertorosmaninho:
> The druntime submodule hash should be restored as well.

@kinke, could you, please, point me the right commit of druntime, phobos and dmd-testsuite that I have to upgrade to get the PR accepted?

kinke:
If you look at the overall diff, e.g., here on GitHub, you'll see that you've only modified the druntime hash at some point, and are now, by the looks of it, trying to set it back to an older commit. I.e., the other 2 submodules are okay.

When you're on your branch and in the repo root dir, you can type `git log runtime/druntime` to see the latest commits modifying the hash, and probably see yours at the top. You can then use `git show <hash from your commit touching druntime>` to see what the hash was.

Robertorosmaninho:
I'm sorry for the mess, I tried to smash some commits and these modifications should work.

cmake/Modules/FindMLIR.cmake
JohanEngelen: You should contact the LLVM/MLIR team about how to be able to do this better in the future. Something similar to `llvm-config` for example.

Robertorosmaninho: Ok! For now, the PR looks good?

driver/main.cpp
```diff
@@ -1088,9 +1094,17 @@ int cppmain() {
 }
 
 void codegenModules(Modules &modules) {
+  bool useMLIR = false;
```
JohanEngelen, line 1097: this variable is not actually used (without mir it is really not used, and with mir it is always true)

driver/toobj.cpp
```diff
@@ -10,6 +10,7 @@
 #include "driver/toobj.h"
 
 #include "dmd/errors.h"
+#include "dmd/module.h"
```
JohanEngelen, line 13: This is not needed?

driver/toobj.h
```diff
@@ -12,6 +12,11 @@
 //===----------------------------------------------------------------------===//
 
 #pragma once
+#if LDC_MLIR_ENABLED
+#include "mlir/IR/MLIRContext.h"
+#include "gen/irstate.h"
+#include "dmd/module.h"
+#endif
```
JohanEngelen, line 19: this is not needed by anything in toobj.h or in toobj.cpp, so best to remove it and add it in the files where it _is_ used.

driver/codegenerator.cpp
```diff
@@ -33,7 +33,7 @@
 
 #if LDC_LLVM_VER < 600
 namespace llvm {
-using ToolOutputFile = tool_output_file;
+  using ToolOutputFile = tool_output_file;
```
JohanEngelen, line 36: the space is counter to our formatting. Can you use clang-format for whitespace?

driver/main.cpp
```diff
@@ -509,6 +511,10 @@ void parseCommandLine(Strings &sourceFiles) {
                strcmp(ext, global.s_ext.ptr) == 0) {
       global.params.output_s = OUTPUTFLAGset;
       global.params.output_o = OUTPUTFLAGno;
+    } else if (opts::output_mlir.getNumOccurrences() == 0 &&
+               strcmp(ext, global.mlir_ext.ptr) == 0) {
+      global.params.output_mlir = OUTPUTFLAGset;
+      global.params.output_o = OUTPUTFLAGno;
```
JohanEngelen, line 517: if MIR is not compiled-in, the user should receive an error with the commandline option, saying that LDC was not compiled with MIR support. Here is a good place to generate that error I think.

JohanEngelen:
See my inline comments, otherwise looks good (I still don't like the addition to codegenerator, but let's live with it for now).
Main thing that is missing is test case(s).

driver/main.cpp
Robertorosmaninho: What would be the best message for this error?

Robertorosmaninho:
> See my inline comments, otherwise looks good (I still don't like the addition to codegenerator, but let's live with it for now).
> Main thing that is missing is test case(s).

Could you point me some test cases that test whether "file.ll" was created?

JohanEngelen:
> Could you point me some test cases that test whether "file.ll" was created?

Create folder `tests/mlir` and add your testcase there. Only enable that folder when MLIR is enabled: See how things are done for another conditional built-in capability like dynamic compile: https://github.com/ldc-developers/ldc/blob/master/tests/lit.site.cfg.in#L61-L63

Add a testcase, similar to anything in `tests/codegen`, like https://github.com/ldc-developers/ldc/blob/master/tests/codegen/wasi.d
More about testcases: https://github.com/ldc-developers/ldc/blob/master/tests/codegen/wasi.d

driver/main.cpp
Robertorosmaninho: I tried to do something like you proposed earlier https://github.com/ldc-developers/ldc/pull/3313#issuecomment-603473567, but I can erase it.

Robertorosmaninho:
> > Could you point me some test cases that test whether "file.ll" was created?
> 
> Create folder `tests/mlir` and add your testcase there. Only enable that folder when MLIR is enabled: See how things are done for another conditional built-in capability like dynamic compile: https://github.com/ldc-developers/ldc/blob/master/tests/lit.site.cfg.in#L61-L63

Thanks!

> Add a testcase, similar to anything in `tests/codegen`, like https://github.com/ldc-developers/ldc/blob/master/tests/codegen/wasi.d
> More about testcases: https://wiki.dlang.org/LDC_Lit-based_testsuite

These tests will be presented with the modifications that actually generate some code, another branch is already being prepared to be sent here as PR (https://github.com/Robertorosmaninho/ldc/tree/Introducing-MLIR-translation/tests/codegen_mlir).

driver/codegenerator.cpp
Robertorosmaninho: done

driver/toobj.h
Robertorosmaninho: done

driver/toobj.cpp
Robertorosmaninho: done

driver/main.cpp
Robertorosmaninho:
```
} else if (opts::output_mlir.getNumOccurrences() == 0 &&
               strcmp(ext, global.mlir_ext.ptr) == 0) {
      #ifndef LDC_MLIR_ENABLED
      error(Loc(), "MLIR only can be printed if LDC is compiled with MLIR support")
      #else
      global.params.output_mlir = OUTPUTFLAGset;
      global.params.output_o = OUTPUTFLAGno;
      #endif
    }
```

What do you think?

JohanEngelen: oh actually this is not the right place. Because when you pass `-output-mlir` this code is not triggered. Just place it outside the if-statement, checking for `global.params.output_mlir == OUTPUTFLAGset`.
For the error message, "printed" is not clear (the compiler writes a file, instead of printing to stdout). How about: "MLIR output requested but this LDC was built without MLIR support". Call `fatal()` after the error so LDC immediately exits there.

JohanEngelen: (remember to use clang-format to get correct whitespacing)

Robertorosmaninho: Ok!

Robertorosmaninho: done

driver/main.cpp
```diff
@@ -1113,7 +1132,12 @@ void codegenModules(Modules &modules) {
       const auto atCompute = hasComputeAttr(m);
       if (atCompute == DComputeCompileFor::hostOnly ||
           atCompute == DComputeCompileFor::hostAndDevice) {
-        cg.emit(m);
+#if LDC_MLIR_ENABLED
+        if(global.params.output_mlir == OUTPUTFLAGset)
```
JohanEngelen, line 1136: whitespace, clang-format! ;)

driver/codegenerator.cpp
```diff
@@ -341,4 +349,73 @@ void CodeGenerator::emit(Module *m) {
...
+  writeMLIRModule(&module, m->objfile.toChars());
+
+  if (m->llvmForceLogging && !loggerWasEnabled) {
+    Logger::disable();
+  }
+}
+
+void CodeGenerator::writeMLIRModule(mlir::OwningModuleRef *module,
+    const char *filename) {
+  const auto outputFlags = {global.params.output_o, global.params.output_bc,
```
JohanEngelen, line 387: Sorry to keep adding things to fix...
This code copy-paste from toobj.cpp will be a maintenance problem. Can you refactor the duplicated code into functions that you can call from both writexxx functions?

Robertorosmaninho: Why not change `writeModule(llvm::Module *m, const char *filename)` to writeModule(llvm::Module *m, mlir::OwningModuleRef *module = nulltpr, const char *filename)? Doing that we can put https://github.com/ldc-developers/ldc/blob/618292b460495122b853f5e2b7193ae7dfdd91c2/driver/codegenerator.cpp#L404-L417 on `wrtieModule` and the lamba expression `replaceExtensionWith` will only be called in this function.

Robertorosmaninho: Then `writeMLIRModule` will not be necessary.

driver/main.cpp
Robertorosmaninho: Sorry, I'll fix it!

driver/codegenerator.cpp
JohanEngelen: Have a look at writeModule. That function already does many different things. Putting writeMLIRModule in that function would make things even worse.
The structure should be D --> MLIR --> LLVM IR --> assembly, and that is not at all reflected in our code.

Robertorosmaninho: Alright, but this sequence is not represented on this code because that’s not the purpose of this PR. In the next one the code generation will be presented as D -> MLIR(D dialect -> Affine Dialect -> LLVM Dialect) -> LLVM IR -> Assembly. Anyway, I’ll make `ReplaceExtensionWith` a static function than can be used for all WriteModules ,

Robertorosmaninho: Done

driver/main.cpp
Robertorosmaninho: Done

driver/toobj.h
```diff
@@ -12,9 +12,13 @@
...
+#include <string>
+#include "dmd/root/dcompat.h"
 
 namespace llvm {
 class Module;
 }
 
 void writeModule(llvm::Module *m, const char *filename);
+std::string replaceExtensionWith(const DArray<const char> &ext,
+                                 const char *filename);
```
JohanEngelen, line 24: newline missing here

.circleci/config.yml
```diff
@@ -149,3 +149,4 @@ workflows:
       - Ubuntu-19.10-multilib-sharedLibsOnly-gdmd
       - macOS-x64
       - macOS-x64-sharedLibsOnly
+
```
JohanEngelen, line 152: needless change here

cmake/Modules/FindMLIR.cmake
```diff
@@ -28,7 +28,12 @@ else()
     set(MLIR_LIB_DIR     ${MLIR_ROOT_DIR}/lib)
 
     # To be done: add the required MLIR libraries. Hopefully we don't have to manually list all MLIR libs.
-    set(MLIR_LIBRARIES "")
+
```
JohanEngelen, line 31: confusing empty line here

driver/codegenerator.cpp
```diff
@@ -219,8 +219,17 @@ void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
 } // anonymous namespace
 
 namespace ldc {
-CodeGenerator::CodeGenerator(llvm::LLVMContext &context, bool singleObj)
-    : context_(context), moduleCount_(0), singleObj_(singleObj), ir_(nullptr) {
+CodeGenerator::CodeGenerator(llvm::LLVMContext &context,
+#if LDC_MLIR_ENABLED
+    mlir::MLIRContext &mlirContext,
+#endif
+bool singleObj)
```
JohanEngelen, line 226: clang-format for whitespacing here

driver/codegenerator.cpp
```diff
@@ -341,4 +349,58 @@ void CodeGenerator::emit(Module *m) {
...
+
+  if (global.errors) {
+    Logger::println("Aborting because of errors");
+    fatal();
+  }
+
+  mlir::OwningModuleRef module;
+  /*module = mlirGen(mlirContext, m, irs);
+  if(!module){
+    IF_LOG Logger::println("Cannot write MLIR file to '%s'", llpath.c_str());
```
JohanEngelen, line 375: the error message here is incorrect

driver/codegenerator.cpp
```diff
@@ -341,4 +349,58 @@ void CodeGenerator::emit(Module *m) {
...
+                                    const char *filename) {
+  // Write MLIR
+  if (global.params.output_mlir) {
+    const auto llpath = replaceExtensionWith(global.mlir_ext, filename);
+    Logger::println("Writting MLIR to %s\n", llpath.c_str());
+    std::error_code errinfo;
+    llvm::raw_fd_ostream aos(llpath, errinfo, llvm::sys::fs::F_None);
+
+    if (aos.has_error()) {
+      error(Loc(), "Cannot write MLIR file '%s':%s", llpath.c_str(),
```
JohanEngelen, line 396: add a space in the error message after the semicolon?

driver/toobj.h
Robertorosmaninho: Between the functions?

driver/codegenerator.cpp
Robertorosmaninho: "Cannot write MLIR module to '%s'" is better?

driver/codegenerator.cpp
Robertorosmaninho: ok

driver/codegenerator.cpp
Robertorosmaninho: ok

cmake/Modules/FindMLIR.cmake
Robertorosmaninho: ok

driver/codegenerator.cpp
JohanEngelen: it's not about writing the file here, it's about generating the MLIR module, right?

Robertorosmaninho: Yes, so I believe that this new message fits well, right?

driver/toobj.h
JohanEngelen: at the end of the file

Robertorosmaninho: done

driver/codegenerator.cpp
JohanEngelen: "Cannot write MLIR module to '%s'" does not help the user. It makes the user think that the _writing_ is the problem (e.g. the file cannot be overwritten because it is protected). "Error generating MLIR" would be a better message (and before that, inside mlirGen I expect there to be more detailed error messages printed).

JohanEngelen:
lgtm

driver/codegenerator.cpp
Robertorosmaninho: Alright! Actually, mlirGen has several error messages printed when it's needed.

cmake/Modules/FindMLIR.cmake
Robertorosmaninho: Hey, @kinke any further comments about this PR?

kinke:
(`ldc2` size increase is < 1% for the Azure packages, so no need to postpone this after v1.22 final.)

Thanks Roberto for hanging on, and the (much needed :)) reviewers. Looking forward to the follow ups.

cmake/Modules/FindMLIR.cmake
Robertorosmaninho: @JohanEngelen the next version of LLVM will support `find_package(MLIR)` https://github.com/llvm/llvm-project/blob/master/mlir/cmake/modules/MLIRConfig.cmake.in
:)

PR #3317, opened 2020-02-10
Merge v2.090.1

kinke:


kinke:
Btw, the `update-dmd-testsuite` script invocation isn't getting any faster. ;) - It's the last script I use from the `ldc-scripts` repo. git now officially recommends a separate Python tool for this task: https://github.com/newren/git-filter-repo

After having it fixed (https://github.com/newren/git-filter-repo/issues/53), it's done in 20 secs, compared to something like 5+ mins for `git filter-branch` (just to add 2-3 commits from DMD `stable`...). Unfortunately, the commit hashes don't match up. Any ideas how to proceed?

dnadlinger:
Regarding dmd-testsuite, I just had a cronjob set up for that way back when keeping the fork up to date, where the time it takes didn't matter.

Now that all of DMD is Boost-licensed, ideally we would just use that repo as a submodule, upstreaming as much of the remaining diff as possible. How realistic that currently is, I'm not sure.

kinke:
> How realistic that currently is, I'm not sure.

I doubt they'll welcome all LDC exceptions (slightly differing output etc.), which are now inside the individual test files...

PR #3320, opened 2020-02-11
Update changelog and README.md

kinke:


kinke:
I'm slowly 'deprecating' ltsmaster and recommend `gdmd` as bootstrap option now, and link to the cross-compilation Wiki page instead of the ldc-build-runtime one (which is linked to on that page). Please approve (or don't, just give feedback ;)).

kinke:
People surely expect the then-latest LDC version to support LLVM 10 when it comes out, so I'll probably look into that before the actual release.

PR #3322, opened 2020-02-12
Fix compilation with LLVM11

thewilsonator:


gen/ctfloat.cpp
```diff
@@ -32,9 +32,15 @@ union CTFloatUnion {
 APFloat parseLiteral(const llvm::fltSemantics &semantics, const char *literal,
                      bool *isOutOfRange = nullptr) {
   APFloat ap(semantics, APFloat::uninitialized);
-  const auto r = ap.convertFromString(literal, APFloat::rmNearestTiesToEven);
```
thewilsonator: For some reason `r.get()` doesn't work on a const object.

JohanEngelen:
Can you also add the usual LLVM upgrade things? FindLLVM.cmake, FileCheck, and profdata. See https://github.com/ldc-developers/ldc/pull/3323/files

kinke:
[`ldc.intrinsics` isn't fit for LLVM 11 either. Btw, I don't mind adding FileCheck and profdata at the end of the LLVM release cycle, as a check whether it's still up-to-date shouldn't be forgotten.]

thewilsonator:
Added llvm config names. If kinke wants to do the intrinsics later, then this is good to go.

gen/irstate.cpp
```diff
@@ -30,6 +30,8 @@ IRScope::IRScope() : builder(gIR->context()) { begin = nullptr; }
 
 IRScope::IRScope(llvm::BasicBlock *b) : begin(b), builder(b) {}
 
+IRScope::IRScope(const IRScope &other) : begin(other.begin), builder(begin) {}
```
thewilsonator: `IRBuilder` had its copy constructor deleted...

gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -231,7 +231,7 @@ class ArrayFI : public TypeInfoFI {
   }
 
   Value *promote(CallSite CS, IRBuilder<> &B, const Analysis &A) override {
-    IRBuilder<> Builder = B;
+    IRBuilder<> &Builder(B);
```
thewilsonator: Hence this change and the one below. I'm pretty sure this is not the right fix but I'm not sure what is.

JohanEngelen: If we don't make a copy, then we should just use `B` instead of adding a reference to it. By not making a copy, my guess is that the insertion point of B is changed after leaving this function, whereas previously the insertion point stayed the same?

kinke:
Superseded by #3546.

PR #3327, opened 2020-02-14
Fixes #3324 - Assertion too strict on vtbl pointer for extern(Windows)

### LINKED ISSUE #3324: Error in assertion for link vtbl pointer in gen/classes.cpp
My (pretty big) project still fails with ldc2 but with this:

```Assertion failed: fdecl->vtblIndex > 0 || (fdecl->vtblIndex == 0 && fdecl->linkage == LINKcpp), file gen\classes.cpp, line 431```

Where name == psl.QueryInterface, and linkage == 4.

Commenting out the assertion makes the build succeed and the application works fine

### PR DESCRIPTION
**Pull request by etcimon**
This loosens a too strict assertion.

### CODE REVIEW

**Comment by kinke** [comment]
As Nick said, a little reduced test would be nice, see https://github.com/ldc-developers/ldc/blob/master/tests/compilable/gh2808.d for an example.

**Comment by kinke** [comment]
While reducing the test case to
```D
extern (System) interface IUnknown { void func(); }

void foo()
{
    IUnknown i;
    i.func();
}
```

I've noticed that it depends on the `IUnknown` identifier and works with others.

kinke:
> it depends on the IUnknown identifier

As designed, as `IUnknown` gets an implicit `extern(C++)` linkage (`extern(Windows)` has no effect on the aggregate itself (normal D class), just on its member fuctions).

PR #3328, opened 2020-02-15
Try to tweak CircleCI

kinke:


kinke:
I hadn't realized that Circle's Mac VMs are twice as powerful as their Linux containers (4 vs. 2 cores, 8 vs. 4 GB of memory), just like Azure Pipelines. We as open-source project cannot use their bigger offerings (`resource_class: large`), as opposed to Cirrus CI.

PR #3333, opened 2020-02-21
Upgrade frontend & libs to v2.091.0

kinke:


kinke:
Something bad must have happened wrt. `static if` and semantic analysis order or so:
```
core/stdc/math.d(4573): Error: undefined identifier `c_long`
```
when compiling `core.stdc.math` separately for non-Windows targets, which defines the `c_long` alias itself. It works when replacing the `static if` (https://github.com/dlang/druntime/blob/f80a07e1a0140f01878a4b13e2eea61faf6a3063/src/core/stdc/config.d#L118) by `version (D_LP64)`.

kinke:
Found a workaround for the `c_long` issue (localized import). The issue isn't new, e.g., LDC v1.20 also fails to compile `core.stdc.config` for Posix when importing `core.math` in `core.internal.convert`. The new global import was also a problem for neither-Windows-nor-Posix targets (wasm, Android...) because `core.stdc.config.c_long` isn't defined for those.

kinke:
The only remaining problem is the ltsmaster compilability regression due to modernized contract syntax.

kinke:
We're now at a stage where ltsmaster apparently miscompiles the compiler, so that new ownership-borrow tests segfault if compiled with ltsmaster (Travis job): `compilable/ob1.d`, `fail_compilation/fob{1,2}.d`.

PR #3339, opened 2020-02-25
Extend inline asm diagnostics by D source location

kinke:


JohanEngelen:
nice

driver/codegenerator.cpp
```diff
@@ -196,8 +193,31 @@ void emitLLVMUsedArray(IRState &irs) {
   llvmUsed->setSection("llvm.metadata");
 }
 
+void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
+                                unsigned locCookie) {
+  if (d.getKind() == llvm::SourceMgr::DK_Error)
+    ++global.errors;
```
kinke: Without extra handler, LLVM aborts itself, after printing something like 'LLVM error while parsing inline asm'. I didn't want to dig deeper into this, but I don't miss that (non-colored) line; we apparently have to manually take care of aborting by increasing the errors count.

gen/naked.cpp
```diff
@@ -492,5 +492,8 @@ llvm::CallInst *DtoInlineAsmExpr(
   bool sideeffect = true;
   llvm::InlineAsm *ia = llvm::InlineAsm::get(FT, code, constraints, sideeffect);
 
-  return gIR->ir->CreateCall(ia, irArgs, "");
+  llvm::CallInst *call = gIR->ir->CreateCall(ia, irArgs, "");
+  gIR->addInlineAsmSrcLoc(loc, call);
```
kinke: We could actually attach an array of cookies, mapping the inline asm lines to D source lines, but that seemed overkill for now (`__asm("a\n\bn\c")` => 3 inline asm lines mapping to 1 D line...), so only the Loc of the containing`asm` statement / `__asm` expression is used.

gen/irstate.h
```diff
@@ -113,6 +113,7 @@ struct IRState {
 private:
   std::vector<std::pair<llvm::GlobalVariable *, llvm::Constant *>>
       globalsToReplace;
+  std::vector<const Loc *> inlineAsmLocs;
```
kinke: These `Loc`s should probably be kept alive for safety (`-lowmem`), by storing the array in GC-tracked memory. `Array` has a one-element small-array-optimization...

JohanEngelen: I don't fully understand your comment, but sizeof(Loc) is only twice as large as a pointer. So why not store `Loc`s (instead of `Loc*`) inside the inlineAsmLocs vector (Loc does contain a pointer to the filename, but the filename string is probably kept alive longer than the Loc?)? Would that help?

kinke: That's also an option, but has a similar issue - the GC should know about these `Loc` instances in order not to collect the filename. It would with `Array`, but we'd need to make sure its elements are on the tracked heap, not on the C++ heap (where `IRState` lives, and so `IRState::inlineAsmLocs` and due to small-array-optimization, its element(s) directly if length <= 1).

JohanEngelen: `Array::reserve` a larger amount?

kinke: That's an option, but I went for simply allocating the Array itself on the D heap (to be precise, only if running with `-lowmem`), as I've recently done for other problematic `Strings` as well, see the last commit.

kinke: Argh, that's still no enough - the GC needs to know about the `IRState::inlineAsmLocs` pointer to that Array, in order to keep that alive...

gen/irstate.cpp
```diff
@@ -40,9 +40,10 @@ IRScope &IRScope::operator=(const IRScope &rhs) {
 IRState::IRState(const char *name, llvm::LLVMContext &context)
     : module(name, context), objc(module), DBuilder(this) {
   ir.state = this;
+  mem.addRange(&inlineAsmLocs, sizeof(inlineAsmLocs));
 }
 
-IRState::~IRState() {}
+IRState::~IRState() { mem.removeRange(&inlineAsmLocs); }
```
JohanEngelen: (  I see a small helper class on the horizon if we have to add more C++ variables to the GC range.
Something like `GCTracked< T >` with a ctor calling `addRange` and dtor calling `removeRange`. ;-P  )

kinke: Yeah, definitely. It's the 2nd such usage for now (the other is `IrFuncTyArg::type`).

PR #3341, opened 2020-02-25
Windows: Do not leak MSVC-environment-setup into `-run` child processes

kinke:


kinke:
Btw, the reason for setting the `ldc2.exe` environment, as opposed to just the linker child process, is `-link-internally`.

JohanEngelen:
Would be good to add a testcase for this if possible.

kinke:
Testing this via CI would be cumbersome, as we obviously don't auto-detect the environment for each linking operation when running the testsuite, and it's only auto-detection (requires an `LDC_VSDIR` env variable *and* a cleared `VSINSTALLDIR`) that leaked.

Issue #3346, opened 2020-03-03
Shared static this ordering problem for module importing another module with ctor


JohanEngelen:
The following code crashes inside druntime:

```d
// import core.thread.osthread;  // uncomment to fix
import core.thread;

void foo() {}

shared static this()
{
    auto f = new Fiber(&foo); // depends on shared static this in core.thread.osthread.
}

void main() {}
```

Backtrace (fixed up with irrelevant info:
```
#0  0x00000000006414a5 in core.thread.fiber.Fiber.allocStack(ulong, ulong) ()
#1  0x0000000000641573 in _D4core6thread5fiber5Fiber6__ctorMFNbDFZvmmZCQBrQBpQBlQBi ()
#2  0x00000000005ae46b in example_module_ctor
```

The crash happens on this line because PAGESIZE has not been set by core.thread.osthread's shared static this yet (still zero): https://github.com/ldc-developers/druntime/blob/e96b6f2743274648ace75126488954c671ec4140/src/core/thread/fiber.d#L1195

The spec says that the order of execution is well defined: https://dlang.org/spec/module.html#order_of_static_ctor
That is: the example imports core.thread, which imports core.thread.osthread, therefore core.thread.osthread's module ctor must be run _before_ the example's module ctor.

kinke:
According to godbolt, it's a 1.20 regression.

kinke:
The respective ordering code in `rt.minfo` relies on `importedModules`, so `core.thread` not being pulled in newly yields a `null` weak ref for its `importedModules` slot in the sample module's ModuleInfo, so the transitive dependencies are lost. Not sure how this works for DMD, but then we do need strong refs and pull in the otherwise unused object. This also affects the no-moduleinfo pragma and problematic .di dependencies.

kinke:
AFAICT, DMD emits strong refs, so I've been fooled by the comment in DMD src (and the flag apparently being ignored)...

mario-tripaneer:
Does this change also affect the import/static ctor call behavior described in #3343 ?

kinke:
Yep, should fix that too.

kinke:
Fixed in #3347 and v1.20.1 & master.

PR #3347, opened 2020-03-04
Revert to strong importedModules ModuleInfo refs

kinke:
Resolves issue #3346.

### DIFF
--- a/dmd/dsymbolsem.d
+++ b/dmd/dsymbolsem.d
@@ -1609,20 +1609,16 @@ version (IN_LLVM)
             // if inside a template instantiation, the instantianting
             // module gets the import.
             // https://issues.dlang.org/show_bug.cgi?id=17181
+            Module importer = sc._module;
             if (sc.minst && sc.tinst)
             {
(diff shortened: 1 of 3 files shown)


dmd/dsymbolsem.d
```diff
@@ -1674,8 +1670,8 @@ version (IN_LLVM)
 
             if (imp.mod.needmoduleinfo)
             {
-                //printf("module4 %s because of %s\n", sc.module.toChars(), mod.toChars());
-                sc._module.needmoduleinfo = 1;
+                //printf("module4 %s because of %s\n", importer.toChars(), imp.mod.toChars());
+                importer.needmoduleinfo = 1;
```
kinke: This gets rid of the previously referenced `ldc.profile` ModuleInfo for `tests/PGO/profile_rt_calls` and 32-bit x86 targets, the original issue I've tried to work around with the weak refs. That issue isn't an issue anymore, since the `std.math` import in `std.format` was reverted, so I've tested this manually. Will be upstreamed if I find the energy to find a nice small testcase.

PR #3348, opened 2020-03-05
WebAssembly: Switch from legacy linked-list ModuleInfo registry to __minfo section

kinke:
As LLD [**edit**: 9+] supports the magic `__{start,stop}___minfo` linker symbols for wasm targets.

### DIFF
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -101,10 +101,10 @@ enum class RegistryStyle {
   /// global.
   legacyLinkedList,
 
-  /// Module references are emitted into the .minfo section.
-  sectionMSVC,
+  /// Module references are emitted into the .minfo/__minfo section.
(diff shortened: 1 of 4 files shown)


skoppe:
It took me a while to understand it, but by emitting all moduleinfo pointers into their own (sub)section (which ends up in the data section) you dont need a linked list anymore, and you just have an array of pointers.

Pretty awesome, I didn't know you could do that.

I reckon I have to modify my druntime code for this as well.

kinke:
> I reckon I have to modify my druntime code for this as well.

Yes, `rt/sections_ldc.d` won't do anymore; you could add a separate `sections_wasm.d` similar to https://github.com/ldc-developers/druntime/blob/ldc/src/rt/sections_android.d (but even simpler because no TLS).

PR #3365, opened 2020-03-14
Support PS4

thewilsonator:
cc @TurkeyMan Try this.


thewilsonator:
Theres some LTO stuff that I haven't changed yet, because I assume that you will be doing the linking manually.

TurkeyMan:
Wow, that was fast!
Yeah, we'll link externally.

I don't understand how this works, is that a method from llvm?
How does this psych stop the error I was seeing?
And what are the proper args to build a PS4 binary?

thewilsonator:
> Yeah, we'll link externally.

Good.

> is that a method from llvm?

Yes. https://llvm.org/doxygen/Triple_8h_source.html#l00655

> How does this patch stop the error I was seeing?

Ah, I forgot that one. Done, see line 846 of the diff.

thewilsonator:
> And what are the proper args to build a PS4 binary?

Dunno.  You obviously need to supply `-mtriple=x86_64-scei-ps4`. Hopefully that should be all that is needed.

TurkeyMan:
That should target the system, but do you reckon that triple will drive defaults for the CPU and potential other build flags, or do I need to collect the proper set and supply them all manually?

thewilsonator:
I would suspect the LLVM would do that for you, but that is only a suspicion. Either way it will be the same as for what clang does. If you specify it for that I would presume that you also need it for LDC.

TurkeyMan:
Well the official compiler is built configured as default. You don't specify anything to that compiler. We need to match all the flags they enable by default I guess.

thewilsonator:
can you post the result of `clang -v` (and `clang -###` if it i different)?

kinke:
* There's a reserved `PS4` version.
* The target CPU should probably default to `btver2`.
* Linking should work with LDC too, like any other target. So specifying the SDK clang as `-gcc` option should be enough; @TurkeyMan: please test and e.g. see if we have to modify the hardcoded C libs.
* Manu: I assume you can run and test console processes on the device/simulator. In that case, building the druntime/Phobos test runners via [ldc-build-runtime](https://wiki.dlang.org/Building_LDC_runtime_libraries) and running them would be very helpful.

thewilsonator:
> There's a reserved PS4 version.

OK, we should probably still set `version(FreeBSD)` though, since it is a FreeBSD.

> The target CPU should probably default to btver2

How do I do that?

kinke:
> OK, we should probably still set version(FreeBSD) though, since it is a FreeBSD.

Of course, otherwise adapting druntime/Phobos would be very much fun.

> How do I do that?

https://github.com/ldc-developers/ldc/blob/bdb613d0ad81c40a13ece387709b91d8d6deb1e2/driver/targetmachine.cpp#L120-L126

kinke:
There are more FreeBSD cases, e.g., https://github.com/ldc-developers/ldc/blob/bdb613d0ad81c40a13ece387709b91d8d6deb1e2/gen/target.cpp#L69-L72

JohanEngelen:
Needs a testcase with `x86_64-scei-ps4` !
What does "scei" stand for?

JohanEngelen:
> * There's a reserved `PS4` version.

Apparently the decision landed on `PlayStation4` instead of PS4.
https://dlang.org/spec/version.html#predefined-versions

TurkeyMan:
Do you suppose the intention is to define `PlayStation4` AND `PlayStation`??

TurkeyMan:
That way `PlayStation5` will also have the common `PlayStation` version for PSx general code that doesn't care which one?

TurkeyMan:
"scei" = Sony Computer Entertainment International

driver/main.cpp
```diff
@@ -790,6 +790,9 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("Posix");
     VersionCondition::addPredefinedGlobalIdent("CppRuntime_Clang");
     break;
+  case llvm::Triple::PS4:
+    VersionCondition::addPredefinedGlobalIdent("PS4");
```
TurkeyMan: It seems the spec lists 'PlayStation4', and there's also the un-numbered version key which might also be intended?

thewilsonator:
Of course, but this is still very much WIP. I'll be waiting to get the green light from Manu and Pursche and then polish this up.

driver/main.cpp
thewilsonator: Changed the version to PlayStation4. I guess we can add the unnumbered version when we support more than one version of PS.

runtime/PresetRuntimeConfiguration.cmake
```diff
@@ -113,7 +113,7 @@ if(NOT LDC_TARGET_PRESET STREQUAL "")
         endif()
     elseif(LDC_TARGET_PRESET MATCHES "PlayStation4")
         set(LLVM_TARGET_TRIPLE "x86_64-scei-ps4")
-        list(APPEND D_FLAGS "-mtriple=${LLVM_TARGET_TRIPLE}")
+        append("-mtriple=x86_64-scei-ps4" D_FLAGS)
```
kinke: Hardly worth a dedicated preset (I was planning on getting rid of) and not sufficient, because the C flags aren't set up. - Building the libs shouldn't be much harder than `CC=preconfigured-SDK-clang ldc-build-runtime --ninja --dFlags="-mtriple=x86_64-scei-ps4" --targetSystem="FreeBSD;UNIX"`, i.e., the same as for most other targets, and explained in the Wiki.

thewilsonator: Is there a way to set the default target of LDC other than building LLVM with LLVM_TARGET_TRIPLE set to a particular value?

kinke: Yes, creating a little bash file. ;)

kinke:
> That way PlayStation5 will also have the common PlayStation version for PSx general code that doesn't care which one?

IMO, `PlayStation` as a generalization for all PS consoles is useless; AFAIK, the used hardware (no idea about the software) has changed dramatically over the years. And we'll yet have to see whether there's a justification for any PS reserved version at all, if it's really nothing but a FreeBSD on specific x64 CPUs.

I find it pretty much incredible that there's `PlayStation[4]`, but `darwin` is still deprecated (but would be very useful for recent iOS-derived platforms support, as there are hardly any relevant divergences across Darwin-based flavors for a specific bitness).

Pursche:
> > That way PlayStation5 will also have the common PlayStation version for PSx general code that doesn't care which one?
> 
> IMO, `PlayStation` as a generalization for all PS consoles is useless; AFAIK, the used hardware (no idea about the software) has changed dramatically over the years. And we'll yet have to see whether there's a justification for any PS reserved version at all, if it's really nothing but a FreeBSD on specific x64 CPUs.
> 
> I find it pretty much incredible that there's `PlayStation[4]`, but `darwin` is still deprecated (but would be very useful for recent iOS-derived platforms support, as there are hardly any relevant divergences across Darwin-based flavors for a specific bitness).

I can tell you that it definitely isn't "just" FreeBSD on specific x64 CPUs, I expect to have to make significant changes to the DRuntime when it comes to threading, networking and possibly memory.

JohanEngelen:
> Of course, but this is still very much WIP. I'll be waiting to get the green light from Manu and Pursche and then polish this up.

I was trying to remind you that you should build up your testcase along with the code you are writing. ;-) Forces you to think a little harder about all the changes needed, and also points review to cases to look into.

driver/main.cpp
```diff
@@ -1017,8 +1020,9 @@ int cppmain() {
   }
 
   auto relocModel = getRelocModel();
-  if (global.params.dll && !relocModel.hasValue()) {
-    relocModel = llvm::Reloc::PIC_;
+  if (!relocModel.hasValue()) {
+    if (global.params.dll || llvm::Triple(mTargetTriple).isPS4())
+      relocModel = llvm::Reloc::PIC_;
```
JohanEngelen: Better to change to PIC inside this function: https://github.com/ldc-developers/ldc/blob/0a007a55f87d91f54d10366621d46c379b268cf6/driver/targetmachine.cpp#L413

thewilsonator: Fixed.

driver/targetmachine.cpp
```diff
@@ -426,6 +426,9 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
       // these OSes.
       // On Android, PIC is default as well.
       relocModel = llvm::Reloc::PIC_;
+    } else if (triple.isPS4()) {
+      // PS4 also requires PIC, exact reason unknown.
```
TurkeyMan: because all ps4 software is a dll.

gen/llvmhelpers.cpp
```diff
@@ -322,6 +322,9 @@ void DtoCAssert(Module *M, Loc &loc, LLValue *msg) {
     args.push_back(file);
     args.push_back(line);
     args.push_back(msg);
+  } else if (triple.isPS4()) {
+    args.push_back(file);
+    args.push_back(msg);
```
TurkeyMan: This arg should be the name of the calling function.

gen/llvmhelpers.cpp
```diff
@@ -322,6 +322,9 @@ void DtoCAssert(Module *M, Loc &loc, LLValue *msg) {
     args.push_back(file);
     args.push_back(line);
     args.push_back(msg);
+  } else if (triple.isPS4()) {
+    args.push_back(file);
```
TurkeyMan: This arg is supposed to be: `format("%s(%d) : Assertion Failed ( %s )", __FILE__, __LINE__, ExpressionAsString)`

thewilsonator: I know. This was done to avoid errors in compilation more than it was to get a nice error message. I assume that people will be using the D asserts once they are working. If a nice C assert message is a desired feature then I can invest some time to do it, but its not particularly high on my list.

Its probably easier to redirect assert to a custom function that does the formatting and then call `_Assert` from that.

kinke: > Its probably easier to redirect assert to a custom function that does the formatting and then call _Assert from that.

Absolutely, it's utter nonsense that each compiler backends needs to take care of the betterC asserts itself. As long as we don't have a minimal/betterC (d)runtime, object.d could be augmented by such a (force-inlined) function for `version (BetterC)`, taking care of the different platform signatures,

JohanEngelen:
@thewilsonator It's a pity this PR has not been merged yet. Can you address the comments and patch it up?

Pursche:
I have been made aware that there has been internal discussions between Sony and a developer about running Rust on PS4/5 by having their LLVM based compiler output LLVM IR, and then finishing compilation with the provided Sony specific toolchain.

The outcome of the discussion was Sony telling the developer that combining a custom LLVM front end via LLVM IR is not allowed because it is likely to break with the customized ABI. There is also a specific technical requirement (that you need to pass to get certified before release) which prohibits using anything but the provided toolchain.

The suggested "workaround" seems to be to cross-compile to C/C++ first, and then compile using the full toolchain, but I don't know how viable that sounds.

PR #3312, opened 2020-02-07
MLIR: Updating CMakeLists and adding FindMLIR.cmake

Robertorosmaninho:
This updates will allow LDC to build the code that will allow the support for MLIR

PR #3368, opened 2020-03-16
Updating CMake and FindMLIR to work on Windows and CI distribution

Robertorosmaninho:


kinke:
I'm still waiting for an answer to https://github.com/ldc-developers/ldc/pull/3312#issuecomment-596230544. v1.21.0-beta1 will be based on LLVM 10 rc4, so there'd still be time to modify MLIR in our LLVM fork.

JohanEngelen:
This is not going in the right direction. Fixed it here: https://github.com/ldc-developers/ldc/pull/3371

PR #3371, opened 2020-03-17
CMake: Fix finding MLIR.

JohanEngelen:
Fixes https://github.com/ldc-developers/ldc/pull/3312


JohanEngelen:
@Robertorosmaninho Should we only add MLIR support from a certain LLVM version upwards? Since which LLVM version is MLIR included?

kinke:
v10.

PR #3373, opened 2020-03-18
Cherry-pick Phobos iOS support & extend Mac package by prebuilt iOS/arm64 libraries

kinke:


kinke:
@jacob-carlborg: Looking bad, needs druntime stuff. How have you tested this then? Merging both druntime and Phobos upstream masters into the ldc branches here?

jacob-carlborg:
> Looking bad, needs druntime stuff

Yes, needs this: https://github.com/dlang/druntime/pull/2952.

> How have you tested this then? Merging both druntime and Phobos upstream masters into the ldc branches here?

No. I just explicitly applied the changes I made (in upstream) on top of LDC master:

https://github.com/jacob-carlborg/phobos/tree/ios-ldc-upstream
https://github.com/jacob-carlborg/druntime/tree/ios-ldc-upstream

jacob-carlborg:
Perhaps it's easier to wait. Then more things will hopefully be ready. Like a CI pipeline, perhaps better Objective-C integration and a blog post.

kinke:
v1.21 final will likely take another month (v2.091.1...), so that might be enough time to set up CI for package generation (just the libs).

.azure-pipelines/posix.yml
```diff
@@ -254,6 +254,20 @@ steps:
     cp build-libs-$arch/lib/*.a installed/lib-$arch
   displayName: 'Android: Cross-compile x86 libraries and copy to install dir'
   condition: and(succeeded(), eq(variables['CI_OS'], 'android'))
+- script: |
+    set -ex
+    cd ..
+    build/bin/ldc-build-runtime --ninja -j $PARALLEL_JOBS \
+      --buildDir=build-libs-arm64 \
+      --cFlags='-target;arm64-apple-ios' \
```
jacob-carlborg: Not sure what happens if you don't include an iOS version.

kinke: Yeah, that's what I want to figure out. I can't imagine we really need different libs for the different iOS versions.

jacob-carlborg: > Yeah, that's what I want to figure out.

I just checked. It will set the iOS version to 7.

> I can't imagine we really need different libs for the different iOS versions.

No, we should specify the minimum version we want to support. It's the same idea with `MACOSX_DEPLOYMENT_TARGET` on macOS. In Apple's SDK, symbols are annotated with in which version of the OS they were added. `arm64-apple-ios12.0` and `MACOSX_DEPLOYMENT_TARGET` specifies the minimum deployment version. That is, the oldest version of the OS the application should be able to run on. If you're trying to use a symbol that was added in a later version than the specified deployment version, the compiler will force you to check (a runtime check) that the symbol exists.

When it comes to picking the oldest version we want to support I think it makes sense to either pick the oldest version the CI system supports or we can pick the latest version of the OS of the oldest supported device (which would be the first device with 64bit support). When it comes to the latter, that would be iPhone 5S and iOS 12. Current version of iOS is 13. iOS 14 is expected to be announced during WWDC in June.

jacob-carlborg:
When it comes to packaging there's the question of what's best. To package only binary code, only LLVM bitcode or both (not sure if it's possible to do that with universal libraries). New Xcode projects defaults to have bitcode enabled. For now, for simplicity, I have disabled bitcode in the Xcode projects I've been using.

kinke:
Wrt. bitcode, `BUILD_LTO_LIBS=ON` would build bitcode libs (`libdruntime-ldc-lto.a` etc.). They include the ThinLTO summaries, but that probably doesn't matter. I assume a user package's bitcode is supposed to be pre-linked at the IR level, which I don't think is currently supported by LDC.

.azure-pipelines/posix.yml
kinke: Thx for checking; I'll adapt it later.

.azure-pipelines/posix.yml
```diff
@@ -254,6 +254,46 @@ steps:
...
+    "arm64-apple-ios":
+    {
+        switches = [
+            "-defaultlib=phobos2-ldc,druntime-ldc",
+            "-link-defaultlib-shared=false",
+            "-fvisibility=hidden",
+            "-Xcc=-target",
+            "-Xcc=arm64-apple-ios12.0",
+            "-Xcc=--isysroot",
+            "-Xcc=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk",
```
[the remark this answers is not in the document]
kinke: I know that it might not be correct, and I let you Apple guys decide whether it's worth it. It makes LDC easier to use for iOS cross-compiling than clang itself (the smoke test just turned green). But I'm okay with the user having to add the corresponding `-Xcc` switches himself, or adapting the ldc2.conf file for now.

kinke:
For anyone interested in playing with this (on a Mac): https://github.com/ldc-developers/ldc/releases/download/CI/ldc2-7aaf2905-osx-x86_64.tar.xz

.azure-pipelines/posix.yml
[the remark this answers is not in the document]
jacob-carlborg: This is a good start and will be correct (at least until Apple changes the structure) if one is running a recent enough version of Xcode.

jacob-carlborg:
> For anyone interested in playing with this (on a Mac): https://github.com/ldc-developers/ldc/releases/download/CI/ldc2-7aaf2905-osx-x86_64.tar.xz

I'll give this a try.

kinke:
dub will most likely need to be slightly extended too, at least for the .dylib extension: https://github.com/dlang/dub/blob/68c6fc2dfb4aa62b41ea7e289520e0495fe905e5/source/dub/compilers/ldc.d#L196-L197 (and that needs an extension of the compiler [probe file](https://github.com/dlang/dub/blob/master/source/dub/platform.d)). [It's missing `version (darwin)` all over again.]

kinke:
=> https://github.com/dlang/dub/pull/1903

jacob-carlborg:
> dub will most likely need to be slightly extended too, at least for the .dylib extension: https://github.com/dlang/dub/blob/68c6fc2dfb4aa62b41ea7e289520e0495fe905e5/source/dub/compilers/ldc.d#L196-L197 (and that needs an extension of the compiler probe file). [It's missing version (darwin) all over again.]

Yeah. That's also why I don't want to rush anything. There's a lot polishing that can be done. If we announce this as it is now, I think people will get disappointed. One needs to integrate with Xcode, which is a bit complicate, unless you know what you're doing. The Objective-C integration is very basic in LDC, which will mean it's going to be a PITA to actual write a meaningful app.

Or do you prefer to release early and often and that's why you're pushing this?

kinke:
I'm not trying to rush, but believe in incremental improvements. If we wanted super-polished Android support before announcing anything, people still wouldn't have a clue about Android support and might dismiss D as an option because of it etc. And from the compiler perspective, iOS support seems pretty complete with this by now, except for ABI issues and the apparently planned, but orthogonal Objective-C improvements. This should be enough for enthusiasts to play with it, share their experiences with it and build their stuff on it, without having to build LDC themselves or wait for 2-3 months.

jacob-carlborg:
Fair enough.

kinke:
I'd be glad to coordinate the release log entry. I'd propose something like:
* **iOS** (incl. watchOS and tvOS) support has landed in druntime and Phobos (thanks Jacob!). All unittests are green on iOS/arm64.
The prebuilt macOS package includes prebuilt druntime & Phobos libraries for iOS/arm64, for first `-mtriple=arm64-apple-ios12.0` cross-compilation experiments.

PR #3382, opened 2020-03-22
Give user possibility to avoid passing -fuse-ld to cc with `--linker=`.

JohanEngelen:
This is needed for example when you want to use `lld` but `-fuse-ld=lld` is not supported by the C compiler yet. Manually symlinking `ld` to `ld.lld` and passing `-linker=` to LDC will work with this patch.

### DIFF
--- a/driver/tool.cpp
+++ b/driver/tool.cpp
@@ -27,12 +27,12 @@
 //////////////////////////////////////////////////////////////////////////////
 
 namespace opts {
-llvm::cl::opt<std::string>
-    linker("linker", llvm::cl::ZeroOrMore, llvm::cl::desc("Linker to use"),
-           llvm::cl::value_desc(
-               "lld-link|lld|gold|bfd|...|<nothing>. When explicitly set to "
-               "nothing, prevents LDC from passing -fuse-ld to cc."),
-           llvm::cl::cat(opts::linkingCategory));
+llvm::cl::opt<std::string> linker(
+    "linker", llvm::cl::ZeroOrMore,
+    llvm::cl::value_desc("lld-link|lld|gold|bfd|..."),
+    llvm::cl::desc("Set the linker to use. When explicitly set to '' "
+                   "(nothing), prevents LDC from passing `-fuse-ld` to `cc`."),
+    llvm::cl::cat(opts::linkingCategory));
 }
 
 static llvm::cl::opt<std::string>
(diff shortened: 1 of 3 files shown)


driver/tool.cpp
```diff
@@ -29,7 +29,9 @@
 namespace opts {
 llvm::cl::opt<std::string>
     linker("linker", llvm::cl::ZeroOrMore, llvm::cl::desc("Linker to use"),
-           llvm::cl::value_desc("lld-link|lld|gold|bfd|..."),
+           llvm::cl::value_desc(
+               "lld-link|lld|gold|bfd|...|<nothing>. When explicitly set to "
+               "nothing, prevents LDC from passing -fuse-ld to cc."),
```
kinke: I don't think that'll look great in the help output (move extra sentence to `cl::desc`?).

tests/linking/linker_empty.d
```diff
@@ -0,0 +1,9 @@
+// Check that the user can override LDC passing `-fuse-ld` to `cc`.
+
+// REQUIRES: target_X86
+
+// RUN: %ldc --gcc=echo --mtriple=x86_64-linux --linker= %s | FileCheck %s
```
kinke: Not bad, the super-duper gcc. ;) - `echo` is an intrinsic on Windows, but git ships with a `bash.exe` tool, so it probably works.

driver/tool.cpp
JohanEngelen: what do you think of the text itself? I struggled a bit with this.

kinke: I think it's okay; mentioning `cc` should make it halfway clear that this applies to Posix targets only. - We should have probably put the Linux default `-linker` in the config file/CMake instead of hardcoding it, that would have prevented the need for this...

JohanEngelen: yes indeed. Can still do both.

JohanEngelen: > I don't think that'll look great in the help output (move extra sentence to `cl::desc`?).

Oof indeed looks terrible, sorry for not checking.

JohanEngelen:
good to go imo

MaskRay:
Does --linker accept an absolute path? clang supports `clang -fuse-ld=/path/to/ld` but GCC doesn't (https://gcc.gnu.org/pipermail/gcc-patches/2020-February/540060.html)

kinke:
Sure, we don't artificially restrict that arg, it's passed verbatim as `-fuse-ld` value (for Posix targets; for Windows and wasm, it's name or path to the actual linker).

Issue #3102, opened 2019-07-03
Bad column data in debuginfo


TurkeyMan:
Bad column data causes VS debugger to show a big ugly yellow block at the start of the line where the execution cursor is stopped.
This makes debugging LDC binaries really suck.

I suspect it's a problem with column data. @rainers understands the issue.

Repro:
- Launch an LDC binary in VS
- Put a breakpoint
- Look for the big yellow block.

PR #3388, opened 2020-03-26
Don't emit any column infos for CodeView by default, like clang

kinke:
Fixing #3102; see https://reviews.llvm.org/D23720.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -39,7 +39,13 @@
 
 ////////////////////////////////////////////////////////////////////////////////
 
+namespace cl = llvm::cl;
 using LLMetadata = llvm::Metadata;
+
+static cl::opt<cl::boolOrDefault> emitColumnInfo(
+    "gcolumn-info", cl::ZeroOrMore, cl::Hidden,
+    cl::desc("Include column numbers in line debug infos. Defaults to "
+             "true for non-MSVC targets."));
 
 namespace ldc {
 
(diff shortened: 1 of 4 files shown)


gen/dibuilder.cpp
```diff
@@ -97,6 +97,12 @@ bool DIBuilder::mustEmitLocationsDebugInfo() {
   return (global.params.symdebug > 0) || global.params.outputSourceLocations;
 }
 
+unsigned DIBuilder::getColumn(const Loc &loc) {
+  // like clang, don't emit any column info for CodeView
+  // (https://reviews.llvm.org/D23720)
+  return (loc.linnum && !isTargetMSVC) ? loc.charnum : 0;
```
PetarKirov: Perhaps make this configurable, like clang's `-gcolumn-info`?

kinke: I can't think of any use for it, so I decided against a new option.

PetarKirov: I don't have a use case, so it's not important to me. I'm just wary of unconditionally removing information, without giving the user a choice.
-> Feel free to keep it like this, and add it later if someone else actually needs it.

kinke: The clang PR mentions that it's still useful for sanitizers, so the option is available now.

tests/debuginfo/msvc_no_colinfo.d
```diff
@@ -0,0 +1,15 @@
+// REQUIRES: Windows
```
kinke: Only restricted to Windows hosts to test the default setting.

gen/dibuilder.cpp
PetarKirov: > The clang PR mentions that it's still useful for sanitizers, so the option is available now.

Yes, that's why I opened the question regarding the option, sorry if it was unclear.

PR #3389, opened 2020-03-27
Do not fallback to host for critical section size of unknown targets

kinke:
Instead, error out whenever requested by an expression-less `synchronized` statement, including source LoC to track it down.

This is safer, especially since the previous initial warning may likely be suppressed, and makes this host-agnostic.

Also suppress previous warnings about unknown empty and `none` OS, treating these like `unknown`.

This is a follow-up on https://forum.dlang.org/post/agupeguvxftrucpjvpho@forum.dlang.org.

### DIFF
--- a/dmd/target.d
+++ b/dmd/target.d
@@ -105,12 +105,12 @@ version (IN_LLVM)
     uint alignsize(Type type);
     uint fieldalign(Type type);
 
-    uint critsecsize()
+    uint critsecsize(const ref Loc loc)
     {
         if (c.criticalSectionSize == 0)
         {
             import dmd.errors;
-            error(Loc.initial, "Unknown critical section size");
+            error(loc, "unknown critical section size for the selected target");
             fatal();
         }
         return c.criticalSectionSize;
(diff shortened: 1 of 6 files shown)


dmd/target.d
```diff
@@ -105,12 +105,12 @@ version (IN_LLVM)
...
     uint fieldalign(Type type);
 
-    uint critsecsize()
+    uint critsecsize(const ref Loc loc)
     {
         if (c.criticalSectionSize == 0)
         {
             import dmd.errors;
-            error(Loc.initial, "Unknown critical section size");
+            error(loc, "unknown critical section size for the selected target");
```
thewilsonator: Wy not submit this upstream?

kinke: The upstream version is a one-liner, and DMD doesn't cross-compile. It might be useful for GDC, but AFAIK, Iain has his own Target class too. Plus I can't make the changes and add the corresponding tests at once, like here.

JohanEngelen: It would be nice to only upstream the `uint critsecsize(const ref Loc loc)` API change, I think. The specific target implementation is different indeed.

kinke: Maybe next time I revise our frontend diff (not too soon, last time was a few weeks back).

PR #3392, opened 2020-03-27
Ios testing3

jacob-carlborg:
Just testing to see if making a new PR makes any difference with Bitrise.


kinke:
I've pushed this (rebased) as `ios-testing3` branch on our repo, that seems to work. I've seen that the certificates are exposed to PR builds, so no idea what's missing. Maybe they still have a reference to your original repo and only expose it to PRs to your fork, not the new upstream repo here.

jacob-carlborg:
> Maybe they still have a reference to your original repo and only expose it to PRs to your fork, not the new upstream repo here.

I don't see why that would be. I created a new project and exported the certificates again. I suspecting there's a bug in Bitrise with regard to this setting, that exposes the certificates to PRs.

I temporarily changed the URL for the repository to point to my fork and pushed directly to a branch in my fork. That worked.

Could please try and push this branch directly to LDC (I added dummy Xcode project to reduce the build times). This will hopefully determine if there's a bug in the Bitrise settings.

kinke:
> Could please try and push this branch directly to LDC

I already did yesterday.

jacob-carlborg:
> I already did yesterday (https://github.com/ldc-developers/ldc/tree/ios-testing3).

Ah, I see. Then it looks to me like there's a bug in Bitrise.

kinke:
Still no good news?

jacob-carlborg:
> Still no good news?

No, unfortunately not. I've reported the issue to the Bitrise support. I got the reply that they have escalated the issue to the technical support engineers for further investigation.

jacob-carlborg:
> No, unfortunately not. I've reported the issue to the Bitrise support. I got the reply that they have escalated the issue to the technical support engineers for further investigation.

I've finally got an answer for the Bitrise support. Apparently exposing certificates to pull requests are not supported for public projects.

PR #3393, opened 2020-03-29
AArch64: ABI fixups & add va_arg implementations

kinke:
[Just a testing aid for now.]


kinke:
Shippable CI: down from 16 dmd-testsuite-debug failures to 3, one of which being a harmless gdb failure.

kinke:
Now down to 2 dmd-testsuite-debug failures; the gdb one (`runnable/b18504.d`) and `runnable/argufilem.d`, which seems to be the only test in the whole testsuite using the non-templated, TypeInfo-based `va_arg` implementation, which still isn't implemented (and I'm not going to in this PR). So all C and C++ ABI tests seem to work now.

Wrt. the fundamental `isHFVA` check, working with the IR types seemed a no-brainer at first, as that gets rid of uninteresting differences (const/shared, signed/unsigned, `char` vs. `byte`, `ifloat` vs. `float`, `double` vs 64-bit `real` etc. etc.). As the annoying `va_arg` stuff needs it too, I've added a corresponding trait, and currently re-map the HFVA type (IR static array) to a D type for that, which isn't really nice.

JohanEngelen:
pretty nice. Do you intend to keep the special trait?

kinke:
I don't see how we could get the templated AAPCS64 (non-Apple) `va_arg` to work otherwise, unfortunately. GDC will most likely need it too, so it should probably be upstreamed (and thus work with the D types).

kinke:
Well I suppose compile-time reflection could be used to determine HFVA-ness without a trait; the downside is duplicated logic in compiler and druntime, but as the complexity isn't really high, it might be worth exploring.

JohanEngelen:
That may go down a path of a more general ABI trait ....

kinke:
There is something like this, the special `is(T U == __argTypes)` expression (currently only used for x86_64, exactly for `va_arg`), which returns a type tuple. We could implement that for AArch64 too and return an appropriate static array for HFVAs, which can then easily be checked for HFVA-ness in druntime. The downside is that we need to / should handle other types as well (e.g., small POD structs passed in GP registers and rewritten as int or static array of `long`s).

dmd/argtypes_aarch64.d
```diff
@@ -0,0 +1,178 @@
...
+{
+    if (t == Type.terror)
+        return new TypeTuple(t);
+
+    const size = cast(size_t) t.size();
+    if (size == 0)
+        return null;
+
+    Type tb = t.toBasetype();
+    if (tb.ty == Tstruct || tb.ty == Tsarray || tb.ty == Tdelegate || tb.iscomplex())
```
thewilsonator:
```d
if (tb.ty != Tstruct && tb.ty != Tsarray && tb.ty != Tdelegate && !tb.iscomplex())
    return new TypeTuple(t);
...
```

kinke:
Superseded by a bunch of upstream PRs (those above + dlang/druntime#3039, dlang/druntime#3040, dlang/druntime#3086), #3416 and finally #3421.

PR #3401, opened 2020-04-13
Try to use memcmp for (in)equality of non-mutable static arrays and mutable slices

### LINKED ISSUE #3400: Optimization doesn't work for const array
```d
bool f(int[] b, ref const int[4] c)
{
    return b[0..4]==c;
}
```
If rhs value is const or immutable, the compiler produces comparison through _adEq2 and typeinfos.

kinke:
Fixes #3400. The frontend pre-casts the static array operand to the slice type for cases like `bool[4] == bool[]`, but that's not possible in case the static array is non-mutable but the slice type is mutable.

So check the element types for equivalence, not the array types, as static arrays and slices are NOT equivalent.

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -1011,7 +1011,7 @@ LLValue *DtoArrayEqCmp_impl(Loc &loc, const char *func, DValue *l,
 bool validCompareWithMemcmpType(Type *t) {
   switch (t->ty) {
   case Tsarray: {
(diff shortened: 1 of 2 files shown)


anon17:
It doesn't cast `shared` away?

kinke:
@anon17: Is there an actual issue or is that just speculation based on the code here? `Type.equivalent()` should handle `shared` diffs.

Issue #3405, opened 2020-04-15
gcc builtin for @llvm.x86.avx512.mask.pshuf.b.512 and pshuf friends are missing


9il:
gcc builtin for `@llvm.x86.avx512.mask.pshuf.b.512` and `pshuf` friends are missing

kinke:
Thx for the report.

PR #3406, opened 2020-04-15
 gccbuiltins: Lift 256-bit vector limit, adding 174 builtins for x86 with LLVM 10

kinke:
Among which:
```D
pragma(LDC_intrinsic, "llvm.x86.avx512.pshuf.b.512")
    byte64 __builtin_ia32_pshufb512(byte64, byte64) pure @safe;
```
and thus fixing #3405.

### DIFF
--- a/utils/gen_gccbuiltins.cpp
+++ b/utils/gen_gccbuiltins.cpp
@@ -52,18 +52,6 @@ string dtype(Record* rec, bool readOnlyMem)
         type = type.substr(i);
     }
 
-    if(vec.size() > 0 && type.size() > 0)
-    {
-        int typeSize, vecElements;
-        if(
-            sscanf(vec.c_str(), "%d", &vecElements) == 1 &&
-            sscanf(type.c_str() + 1, "%d", &typeSize) == 1 &&
-            typeSize * vecElements > 256)
-        {
-            return "";
-        }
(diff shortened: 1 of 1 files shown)


utils/gen_gccbuiltins.cpp
```diff
@@ -52,18 +52,6 @@ string dtype(Record* rec, bool readOnlyMem)
...
-        int typeSize, vecElements;
-        if(
-            sscanf(vec.c_str(), "%d", &vecElements) == 1 &&
-            sscanf(type.c_str() + 1, "%d", &typeSize) == 1 &&
-            typeSize * vecElements > 256)
-        {
-            return "";
-        }
-    }
-
```
kinke, line 66: Instead of ignoring builtins with vector types > 256 bits, this now allows all vector types. As soon as there are builtins for vectors > 512 bits, the generated `gccbuiltins_*.di` header is unusable (missing `core.simd` aliases), which should be much easier to detect and fix. - There's one lit-test importing `ldc.gccbuiltins_x86`, so that one's tested.

kinke:
The remaining missing 114 x86 builtins (a total of 989 for LLVM 10) are all missing because we don't have a mapping for the `x86_mmx` IR type.

PR #3408, opened 2020-04-19
Trivial: Add a .gitignore

Geod24:
Using this project as a (submodule) dependency is quite painful otherwise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,28 @@
+# CMake output
+CMakeFiles/
+*.cmake
+CMakeCache.txt
+Makefile
+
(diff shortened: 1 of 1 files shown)


kinke:
That's only the case if the build dir is inside the src tree, which is a bad idea, unless for a quick'n dirty build-install-and-throw-src-away cycle.

Geod24:
Ah, my familiarity with CMake shows. Indeed adjusting things to work in a subdirectory is much saner. I really wish `cmake` was doing this by default though, but it's already properly documented on the wiki, so closing.

PR #3411, opened 2020-04-25
gccbuiltins: Add AMD GCN and Nvidia PTX

kinke:
This is sort of a reboot of #2556. @thewilsonator: The generated files are .di headers, do those really need any special dcompute handling?

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -337,8 +337,8 @@ if(TARGET gen_gccbuiltins)
     )
   endfunction()
 
-  set(target_arch "AArch64;ARM;Mips;PowerPC;SystemZ;X86")
-  set(target_name "aarch64;arm;mips;ppc;s390;x86")
+  set(target_arch "AArch64;AMDGPU;ARM;Mips;NVPTX;PowerPC;SystemZ;X86")
+  set(target_name "aarch64;amdgcn;arm;mips;nvvm;ppc;s390;x86")
 
   foreach(target ${LLVM_TARGETS_TO_BUILD})
     list(FIND target_arch ${target} idx)


thewilsonator:
Shouldn't. LLVM handles the intrinsics.

kinke:
Yeah that's what I thought too. - I've only checked the generated NVPTX file, which contains ~550 builtins, and this works:
```D
// ldc2 -c bla.d -betterC -mtriple=nvptx64 -mcpu=sm_50
import ldc.gccbuiltins_nvvm;
double foo(double a, double b) { return __nvvm_add_rm_d(a, b); }
```

My LLVM doesn't include the AMDGPU backend.

runtime/CMakeLists.txt
```diff
@@ -337,8 +337,8 @@ if(TARGET gen_gccbuiltins)
     )
   endfunction()
 
-  set(target_arch "AArch64;ARM;Mips;PowerPC;SystemZ;X86")
-  set(target_name "aarch64;arm;mips;ppc;s390;x86")
+  set(target_arch "AArch64;AMDGPU;ARM;Mips;NVPTX;PowerPC;SystemZ;X86")
+  set(target_name "aarch64;amdgcn;arm;mips;nvvm;ppc;s390;x86")
```
thewilsonator, line 341: Why is this done as a whitelist anyway? Are there any targets we want to exclude builtins for?

kinke: * IIRC, there are targets like BPF and Hexagon (no idea what those are) with a handful of GCC builtins.
* AFAICT, you need to know the target prefix (`nvvm`) for the tblgen invocation.
* We additionally filter this list and only keep those for which the LLVM backend is available. That restriction can probably be lifted for a uniform `ldc.gccbuiltins_*` imports set, independent from the LLVM build config.

thewilsonator: > AFAICT, you need to know the target prefix (nvvm) for the tblgen invocation.

Thats annoying.

Issue #3270, opened 2020-01-09
LDC does not compile with SystemZ host


Geod24:
While trying to compile LDC v1.19.0 (+ patches from stable) for s390x-musl, I hit [this assert](https://github.com/ldc-developers/ldc/blob/8da4e6f9cf2f1c695902f142ea22eac9d8998845/gen/ctfloat.cpp#L61).

The package definition can be found here: https://gitlab.alpinelinux.org/alpine/aports/merge_requests/2873

kinke:
That's probably just the first of many hurdles (and trivial to fix once we know how to detect such hosts via C++ predefines and which `long double` format the architecture uses). Does gdc already properly support SystemZ? I'd have guessed that druntime support etc. is still severely lacking.

Geod24:
>  Does gdc already properly support SystemZ? I'd have guessed that druntime support etc. is still severely lacking.

Surprisingly not. I had to fiddle with a couple of things, but the druntime support is already there (see e.g. https://github.com/dlang/druntime/pull/2357). Dub compiles and passes `dub test` for example: https://gitlab.alpinelinux.org/alpine/aports/merge_requests/2484

Geod24:
Actually I'm seeing failures on a bunch of platforms.

ARMv7:
```
LLVM ERROR: Cannot select: 0x41b3f28: f64 = bitcast 0x41b4120
  0x41b4120: i64 = build_pair 0x41b3af0, 0x41b4168
    0x41b3af0: i32,ch = CopyFromReg 0x4154bcc, Register:i32 %15
      0x41b4708: i32 = Register %15
    0x41b4168: i32,ch = CopyFromReg 0x41b3af0:1, Register:i32 %16
      0x41b4828: i32 = Register %16
In function: _memset80
make[2]: *** [runtime/CMakeFiles/druntime-ldc.dir/build.make:335: runtime/objects/core/atomic.o] Error 1
make[2]: Leaving directory '/builds/Geod24/aports/testing/ldc/src/ldc-1.19.0-src/stage1'
make[1]: *** [CMakeFiles/Makefile2:856: runtime/CMakeFiles/druntime-ldc.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
LLVM ERROR: Cannot select: 0x3ed0f78: f64 = bitcast 0x3ed1170
  0x3ed1170: i64 = build_pair 0x3ed0b40, 0x3ed11b8
[...]

PR #3414, opened 2020-04-28
Improve detection of host real_t semantics

kinke:
Supporting more hosts and fixing early stoppers such as issue #3270.

### DIFF
--- a/gen/ctfloat.cpp
+++ b/gen/ctfloat.cpp
@@ -51,21 +51,25 @@ void CTFloat::initialize() {
   if (apSemantics)
     return;
 
-  static_assert(sizeof(real_t) >= 8, "real_t < 64 bits?");
-
-  if (sizeof(real_t) == 8) {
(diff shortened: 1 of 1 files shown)


gen/ctfloat.cpp
```diff
@@ -51,21 +51,25 @@ void CTFloat::initialize() {
   if (apSemantics)
     return;
 
-  static_assert(sizeof(real_t) >= 8, "real_t < 64 bits?");
-
-  if (sizeof(real_t) == 8) {
+#ifdef _MSC_VER
+  // MSVC hosts use dmd.root.longdouble (80-bit x87)
```
JohanEngelen, line 55: would this not be handled correctly with `real_t` as for the other platforms below?

kinke: Nope, that type is a struct, not D `real` / C `long double`.

Issue #3402, opened 2020-04-14
weird linking issues on windows


9il:
Hi folks,

What we are doing wrong?
Is there a way to make LDC automatically link with required win-runtime libs?

----

So, linking the `libcmt.lib` or `libcmtd.lib` forces you to link two dependent libs as well.
Here are all the linked libs in `dub.sdl`:

```sdl
lflags `C:\Program Files (x86)\Windows Kits\10\Lib\10.0.18362.0\um\x64\BufferOverflowU.lib` platform="windows-x86_64"
lflags `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\lib\x64\libcmt.lib` platform="windows-x86_64"
lflags `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\lib\x64\libvcruntime.lib` platform="windows-x86_64"
lflags `C:\Program Files (x86)\Windows Kits\10\Lib\10.0.18362.0\ucrt\x64\libucrt.lib` platform="windows-x86_64"
```

Now, when I try to compile the project I get:

```
C:\Devel\mir_benchmarks>dub build --compiler=ldc2 --build=release
Performing "release" build using ldc2 for x86_64.
mir-core 1.1.2: building configuration "library"...
mir-algorithm 3.7.28: building configuration "default"...
mir-blas 1.1.13: building configuration "mkl-tbb-thread-dll"...
mir-random 2.2.14: building configuration "default"...
mir_benchmarks ~master: building configuration "application"...
lld-link: error: undefined symbol: open
>>> referenced by phobos2-ldc.lib(gzlib.c.obj):(gz_open)

lld-link: error: undefined symbol: fileno
>>> referenced by druntime-ldc.lib(dmain2.obj):(_d_print_throwable)
>>> referenced by druntime-ldc.lib(cover.obj):(_D2rt5cover25_sharedStaticDtor_L188_C1FZv)
>>> referenced by druntime-ldc.lib(cover.obj):(_D2rt5cover25_sharedStaticDtor_L188_C1FZv)

lld-link: error: undefined symbol: execv
>>> referenced by phobos2-ldc.lib(process.obj):(_D3std7process6execv_FxAyaxAQfZi)

lld-link: error: undefined symbol: execve
>>> referenced by phobos2-ldc.lib(process.obj):(_D3std7process7execve_FxAyaxAQfxQeZi)

[...]

PR #3415, opened 2020-04-29
Adopt DMD's MSVC toolchain detection

kinke:
This reduces the overhead for auto-detecting and setting up an MSVC toolchain from a very rough 1 second to about 8 milliseconds on my box. Enabling the auto-detection by default (and so preferring MSVC over the 'internal' toolchain if there's a Visual C++ installation) is now possible, fixing issues like #3402.

The MSVC setup now consists of the bare minimum - prepending 3 directories to the `LIB` env var and 1-2 directories to `PATH`.


kinke:
@rainers: Looks usable, thx for this. I still didn't expect almost 10 ms for this stuff; the COM stuff alone takes 5-6 ms on my box (and I have 2 installations to iterate over).

I changed it to use the most recent installation found via COM, not the first one. I also don't want to reintroduce any narrow WinAPI calls and will switch to the wide API (e.g., when getting the installation dir from the registry); is that an option for DMD too? Otherwise the diff might get pretty ugly.

JohanEngelen:
> This reduces the overhead for auto-detecting and setting up an MSVC toolchain from a very rough 1 second to about 8 milliseconds on my box. Enabling the auto-detection by default (and so preferring MSVC over the 'internal' toolchain if there's a Visual C++ installation) is now possible, fixing issues like #3402.

Very nice.

kinke:
For Azure (with unknown number of VS installations):
> MSVC setup took 137592 microseconds

2nd time a few seconds later:
> MSVC setup took 10358 microseconds

rainers:
> I still didn't expect almost 10 ms for this stuff; the COM stuff alone takes 5-6 ms on my box (and I have 2 installations to iterate over).

I guess that's mostly loading DLLs as part of creating the COM object (I just stepped over it in the debugger, and it showed 6 ms, too - for 10 DLLs on a rather fast SSD).

> I changed it to use the latest version found via COM, not the first one (and check for an existing `VC` subdir).

Should be fine. That probably also results in more predictable behvior.

> I also don't want to reintroduce any narrow WinAPI calls and will switch to the wide API (e.g., when getting the installation dir from the registry); is that an option for DMD too?

Should be no problem as far as I can see.

dmd/vsoptions.d
```diff
@@ -635,18 +662,54 @@ const(char)* detectVSInstallDirViaCOM()
...
+        if (instance.GetInstallationVersion(&thisVersionString) != S_OK)
+            continue;
+        scope(exit) SysFreeString(thisVersionString);
+
+        BSTR thisInstallDir;
+        if (instance.GetInstallationPath(&thisInstallDir) != S_OK)
             continue;
+        scope(exit) SysFreeString(thisInstallDir);
+
+        if (versionString && wcscmp(thisVersionString, versionString) <= 0)
```
rainers, line 682: Is string comparison of the version good enough? Probably ok, as 3 digit versions are still some time in the future.

dmd/vsoptions.d
```diff
@@ -635,18 +662,54 @@ const(char)* detectVSInstallDirViaCOM()
...
+            continue; // not a newer version, skip
+
+        const installDirLength = wcslen(thisInstallDir);
```
rainers, line 685: There is also `SysStringLen` (https://docs.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-sysstringlen), which is slightly more correct.

dmd/vsoptions.d
```diff
@@ -635,18 +662,54 @@ const(char)* detectVSInstallDirViaCOM()
...
+        const vcInstallDirLength = installDirLength + 4;
+        auto vcInstallDir = (cast(wchar*) mem.xmalloc_noscan(vcInstallDirLength * wchar.sizeof))[0 .. vcInstallDirLength];
```
rainers, line 687: The result is currently limited to 260 narrow characters below anyway, so I don't think it's necessary to cope with longer strings here, i.e. a stack allocated array should be fine. Or adapt the conversion below, too.

dmd/vsoptions.d
```diff
@@ -635,18 +662,54 @@ const(char)* detectVSInstallDirViaCOM()
...
-        int len = WideCharToMultiByte(CP_UTF8, 0, bstrInstallDir, -1, path.ptr, 260, null, null);
-        SysFreeString(bstrInstallDir);
+        if (versionString)
+        {
+            SysFreeString(versionString);
+            SysFreeString(installDir);
+        }
+        versionString = thisVersionString;
+        installDir = thisInstallDir;
+        thisVersionString = null;
```
rainers, line 701: resetting these local variables that go out of scope seems gratuitous.

packaging/README.txt
```diff
@@ -7,34 +7,21 @@ The compiler configuration file is etc\ldc2.conf and can be easily customized
...
-(vcruntime140.dll, ucrtbase.dll etc.).
+If you have an installed Visual C++ toolchain (Visual Studio/Build Tools 2015 or
+newer), LDC defaults to using linker and libraries of the latest Visual C++
+installation it can find.
+You can set the LDC_VSDIR environment variable to select a specific version,
+e.g., 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community'.
+The MSVC toolchain setup is skipped if LDC is run inside a 'VS Native/Cross
+Tools Command Prompt' (more precisely, if the VSINSTALLDIR environment variable
+is set). You can set the LDC_VSDIR_FORCE environment variable (to some
+non-empty value) to enforce setting up the MSVC toolchain.
```
rainers, line 18: I haven't checked the C++ code whether it is actually doing this, but I think it should be the default behavior to adapt to the architecture even if the VS/VC installation path is given by the environment variable. It is pretty annoying if you have to call `vcvarsall.bat` or similar to adapt the LIB environment variable although you have the option -m32/64 to specify just that to LDC.

packaging/README.txt
```diff
@@ -7,34 +7,21 @@ The compiler configuration file is etc\ldc2.conf and can be easily customized
...
-WinSDK & Visual C++ runtime (import) libraries based on MinGW-w64. In order to
-run the generated binaries, a Visual C++ 2015 runtime installation is required
-(vcruntime140.dll, ucrtbase.dll etc.).
+If you have an installed Visual C++ toolchain (Visual Studio/Build Tools 2015 or
+newer), LDC defaults to using linker and libraries of the latest Visual C++
+installation it can find.
+You can set the LDC_VSDIR environment variable to select a specific version,
+e.g., 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community'.
+The MSVC toolchain setup is skipped if LDC is run inside a 'VS Native/Cross
+Tools Command Prompt' (more precisely, if the VSINSTALLDIR environment variable
```
rainers, line 16: Do the Build Tools set VSINSTALLDIR aswell? Shouldn't this check VCINSTALLDIR, too?

dmd/vsoptions.d
kinke: I've verified that `1.x` < `10.x`.

dmd/vsoptions.d
kinke: This is to prevent them from getting free'd, as they escape the loop iteration via `versionString` and `installDir`.

packaging/README.txt
kinke: Nobody except you has ever complained about that ;), and I still have no idea why that's needed for the DMD build setup.

packaging/README.txt
kinke: It's been like this for years, I've tested it with Build Tools 2015 IIRC years ago, no complaints.

dmd/vsoptions.d
rainers: I meant 100.x is not less than 99.x.

dmd/vsoptions.d
rainers: BSTR doesn't have a destructor.

dmd/vsoptions.d
kinke: Argh. ;)

dmd/vsoptions.d
kinke: `scope(exit) SysFreeString(this...);`

packaging/README.txt
kinke: I think I have a solution now, checking the `VSCMD_ARG_TGT_ARCH` env var (VS 2017+) for a mismatch and *not* skipping the MSVC setup in that case. The environment is still left untouched (thus avoiding potential issues) if the targets match up.

rainers: Sounds good to me, thanks.

PR #3420, opened 2020-05-02
AVR: Add predefined version AVR and emit TLS globals as regular ones

ErnyTech:
The AVR target on LLVM and AVR-GCC does not have support for TLS, so
it is necessary to emit global variables as NotThreadLocal.

Based on PR #3295

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -651,6 +651,10 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("AArch64");
     registerPredefinedFloatABI("ARM_SoftFloat", "ARM_HardFloat", "ARM_SoftFP");
     break;
+  case llvm::Triple::avr:
+    VersionCondition::addPredefinedGlobalIdent("AVR");
+    VersionCondition::addPredefinedGlobalIdent("D_SoftFloat");
+    break;
   case llvm::Triple::mips:
   case llvm::Triple::mipsel:
     VersionCondition::addPredefinedGlobalIdent("MIPS");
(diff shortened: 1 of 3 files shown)


thewilsonator:
Can you also please submit a DMD PR to reserve the `AVR` identifier.

JohanEngelen:
most important one is to add it to the spec. See: https://github.com/dlang/dlang.org/pull/2753

ErnyTech:
@thewilsonator @JohanEngelen Done

https://github.com/dlang/dmd/pull/11094
https://github.com/dlang/dlang.org/pull/2783

ErnyTech:
All the PRs on the dlang side have been merged

driver/main.cpp
```diff
@@ -651,6 +651,9 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("AArch64");
     registerPredefinedFloatABI("ARM_SoftFloat", "ARM_HardFloat", "ARM_SoftFP");
     break;
+  case llvm::Triple::avr:
+    VersionCondition::addPredefinedGlobalIdent("AVR");
+    break;
```
JohanEngelen, line 656: do you need to set anything regarding float abi here?

ErnyTech: The AVR architecture does not have floating-point units, therefore only the soft float is supported depending on the implementation of the compiler or external libraries.

Should I register AVR_SoftFloat?

JohanEngelen: If there is only one option anyway, then no distinction is necessary. Let's not add AVR_SoftFloat.

JohanEngelen:
Can you add a testcase similar to https://github.com/ldc-developers/ldc/blob/master/tests/codegen/wasi.d ?

Also the llvmhelpers.cpp line change looks long, did you clang-format?

LGTM otherwise.

driver/main.cpp
kinke: But `D_SoftFloat` should be added then.

ErnyTech:
I don't know if I can add the test because the AVR target is currently still experimental, so it is only available if enabled in the llvm build flags

kinke:
You can when using `target_AVR` instead of `target_WebAssembly`.

ErnyTech:
I added the test and D_SoftFloat

PR #3424, opened 2020-05-07
Emulate @weak functions on Windows and don't emit COMDATs for ELF anymore

kinke:
The test adaptations are mostly a revert of 3893840f.

`@weak` global variables are still left out but could be emulated exactly the same way.

### DIFF
--- a/gen/tollvm.cpp
+++ b/gen/tollvm.cpp
@@ -223,25 +223,25 @@ LinkageWithCOMDAT DtoLinkage(Dsymbol *sym) {
     linkage = LLGlobalValue::WeakAnyLinkage;
   }
 
-  return {linkage, supportsCOMDAT()};
+  return {linkage, needsCOMDAT()};
 }
 
-bool supportsCOMDAT() {
-  const auto &triple = *global.params.targetTriple;
-  return !(triple.isOSBinFormatMachO() ||
-#if LDC_LLVM_VER >= 500
-           triple.isOSBinFormatWasm()
-#else
-           triple.getArch() == llvm::Triple::wasm32 ||
-           triple.getArch() == llvm::Triple::wasm64
-#endif
-  );
+bool needsCOMDAT() {
+  /* For MSVC targets (and probably MinGW too), linkonce[_odr] and weak[_odr]
+   * linkages don't work and need to be emulated via COMDATs to prevent multiple
+   * definition errors when linking.
+   * Simply emit all functions in COMDATs, not just templates, for aggressive
+   * linker stripping (/OPT:REF and /OPT:ICF with MS linker/LLD), analogous to
+   * using /Gy with the MS compiler.
+   * https://docs.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=vs-2019
+   */
+  return global.params.targetTriple->isOSBinFormatCOFF();
 }
 
 void setLinkage(LinkageWithCOMDAT lwc, llvm::GlobalObject *obj) {
   obj->setLinkage(lwc.first);
-  if (lwc.second)
-    obj->setComdat(gIR->module.getOrInsertComdat(obj->getName()));
+  obj->setComdat(lwc.second ? gIR->module.getOrInsertComdat(obj->getName())
+                            : nullptr);
 }
 
 void setLinkageAndVisibility(Dsymbol *sym, llvm::GlobalObject *obj) {
--- a/tests/PGO/irbased_indirect_calls.d
+++ b/tests/PGO/irbased_indirect_calls.d
@@ -2,17 +2,20 @@
 
 // REQUIRES: PGO_RT
 
+// with LLVM 4.0, @optStrategy apparently doesn't suffice to prevent eliding the hot() call
+// XFAIL: llvm400
+
 // RUN: %ldc -O3 -fprofile-generate=%t.profraw -run %s  \
 // RUN:   &&  %profdata merge %t.profraw -o %t.profdata \
 // RUN:   &&  %ldc -O3 -c -output-ll -of=%t.use.ll -fprofile-use=%t.profdata %s \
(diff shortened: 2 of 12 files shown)


kinke:
I've looked into this because it might be a solution for #3158, at least for LDC - only emitting weak `extern(C++)` `std::exception::what()` etc. in druntime, so that they are ABI-compatible with the C++ ones and don't conflict with a potentially linked in C++ runtime.

tests/codegen/attr_weak.d
```diff
@@ -1,20 +1,23 @@
...
-// though, so the call below is not inlined.
+// should take precedence over and not conflict with weak attr_weak_input.return_two
+extern(C) int return_two() {
+    return 123;
+}
+
+// should be overridden by strong attr_weak_input.return_seven
 extern(C) @weak int return_seven() {
-  return 1;
+  return 456;
```
kinke, line 17: This fails on Mac, so this still requires a matching object files order (later ones not overriding earlier weak symbols). Is that the discussion about Comdats and their selection kind again?

kinke:
Is there any point in emitting Comdats for ELF? It prevents `@weak` from working properly (linker apparently prefers the version in the 1st object file, independent of whether it's weak or not), and clang doesn't emit them for Linux...

tests/PGO/irbased_indirect_calls.d
```diff
@@ -27,7 +27,7 @@ extern (C)
 
     void function() foo;
 
-    @weak // disable reasoning about this function
+    pragma(inline, false)
```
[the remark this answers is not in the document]
kinke: I just wanted to avoid additional complexity for string matching; for `hot`, disabling inlining wasn't enough, as the call was still elided. What worked was `pragma(inline, false); llvm_sideeffect();` in the `hot` body, but `optStrategy` seemed nicer and had the same effect.

For `select_func` here, I thought disabling inlining would be enough (for string matching in the main body wrt. the `foo` function ptr); re-adding `@weak` wouldn't affect any strings though, and might even fix the regression with LLVM 4.0.

kinke:
@rainers: Any suggestions wrt. how to 'best' adapt the mangled name (for backtrace/debugger/...)?

rainers:
> @rainers: Any suggestions wrt. how to 'best' adapt the mangled name (for backtrace/debugger/...)?

If you want it to be demangled with some modifications, prepending a package `weak` by changing the `_D` prefix to `_D4weak` should work.

Your use case in https://github.com/ldc-developers/ldc/pull/3158 probably needs C++ symbols, though. VC mangling doesn't encode identifier length, so patching the first identifier might work, but it can be back referenced so it appears multiple times in the demangled name.

kinke:
Cleaned up & put a little more effort into the renamed mangle.

clang emits COMDATs for templates and inline functions only (with `linkonce_odr`), not for regular functions, for both MSVC and Linux targets. I have no idea whether ELF COMDATs provide any benefit and so whether I should try harder and only get rid of them for regular functions, to make `@weak` work properly.

JohanEngelen:
You could ask on the forums about the COMDATs on Linux, but I doubt you'll get useful feedback.
If Clang does not emit it, then I think it is fairly safe to omit them.

kinke:
Yeah I wanted to know if you guys know more about this, otherwise I'd propose to include this in the upcoming beta and see how it goes.

kinke:
Interestingly, the produced static libraries are now noticeably smaller for the Azure Linux x64 job (compared against latest master run):
* libdruntime-ldc.a: 2,898 vs. 3,210 KB (-9.7%)
* libphobos2-ldc.a: 9,466 vs. 10,393 KB (-8.9%)

The shared libs are minimally larger:
* libdruntime-ldc-shared.so: 1,625 vs. 1,619 KB
* libphobos2-ldc-shared.so: 6,202 vs. 6,173 KB (+0.5%)

gen/tollvm.cpp
```diff
@@ -223,25 +223,25 @@ LinkageWithCOMDAT DtoLinkage(Dsymbol *sym) {
...
+   */
+  return global.params.targetTriple->isOSBinFormatCOFF();
 }
 
 void setLinkage(LinkageWithCOMDAT lwc, llvm::GlobalObject *obj) {
   obj->setLinkage(lwc.first);
-  if (lwc.second)
-    obj->setComdat(gIR->module.getOrInsertComdat(obj->getName()));
+  obj->setComdat(lwc.second ? gIR->module.getOrInsertComdat(obj->getName())
+                            : nullptr);
```
kinke, line 244: FWIW, I've also tried setting the COMDAT's selection kind to 'exact match' (instead of 'any'), but that only led back to the multiple definition errors.

JohanEngelen:
I just found out that the ELF COMDAT change is causing trouble at Weka. It prevents merging of symbols, leaving excess data around in a section where Weka's tooling wants there to be no garbage (or well, it just exceeds the amount of data that the system can deal with (`ushort`...)). Re-enabling COMDAT fixes it.
I guess we have to start doing what @kinke noted about clang (I think that would fix Weka's problem):
> clang emits COMDATs for templates and inline functions only (with linkonce_odr), not for regular functions,

I'll work on a small testcase, so you can see what the problem is.

JohanEngelen:
Opened an issue for it with testcase: https://github.com/ldc-developers/ldc/issues/3589

JohanEngelen:
LLVM 11 release notes:
"Windows target:  Produce COFF weak external symbols for IR level weak symbols without a comdat (e.g. for __attribute__((weak)) in C)"
https://releases.llvm.org/11.0.0/docs/ReleaseNotes.html#changes-to-the-windows-target

@kinke Does that remove the need for the "emulation" of this PR ?

kinke:
I've read that, found it a nice conincidence, but haven't tested whether it works similar to this manual emulation here. Anyway, it's not restricted to LLVM11+, and we always emit comdats on Windows.

PR #3427, opened 2020-05-08
[WIP] [AVR] Implement setGlobalArrayAddressSpace for global ctors/dtors

ErnyTech:
For Harvard architectures, LLVM uses address space to rappresent the "program memory" and the "data space".
Specifically for AVR, the program memory is represented by address space number 1 while the data space is represented by the address space number 0.

To add the Global Ctors/Dtors we use appendToGlobalCtors and appendToGlobalDtors, which depends on appendToGlobalArray, and do not support address spaces other than 0 (See https://github.com/llvm/llvm-project/blob/b38d77f185c53cb4bf108f0c29df1d6d7cfe13b4/llvm/lib/Transforms/Utils/ModuleUtils.cpp#L34) therefore it would not be possible to use it on AVR because the functions are on program memory.

Currently clang sets the ctors/dtors without using appendToGlobalCtors and appendToGlobalDtors and correctly sets the address space (See https://github.com/llvm/llvm-project/blob/b38d77f185c53cb4bf108f0c29df1d6d7cfe13b4/clang/lib/CodeGen/CodeGenModule.cpp#L1229).

With this patch I tried to implement the same behavior of clang, going to overwrite the array already defined by appendToGlobalArray, now LDC_global_crt_ctor and LDC_global_crt_dtor generates a correct ir code for the target AVR


gen/llvmhelpers.cpp
```diff
@@ -1453,13 +1454,55 @@ void printLabelName(std::ostream &target, const char *func_mangle,
...
+        "llvm.global_ctors", gIR->module,
+        gIR->module.getDataLayout().getProgramAddressSpace());
+#endif
   } else {
     llvm::appendToGlobalDtors(gIR->module, func, priority);
+#if LDC_LLVM_VER >= 800
+    setGlobalArrayAddressSpace(
+        "llvm.global_dtors", gIR->module,
+        gIR->module.getDataLayout().getProgramAddressSpace());
+#endif
```
kinke, line 1505: Instead of doing this repeatedly for each ctor/dtor, you can defer this to https://github.com/ldc-developers/ldc/blob/1b52bf06cd13ecad1cff1b6300bd0baf793e9fa4/driver/codegenerator.cpp#L281 where we do other finalization stuff for each IR module.

gen/llvmhelpers.cpp
```diff
@@ -1453,13 +1454,55 @@ void printLabelName(std::ostream &target, const char *func_mangle,
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// Workaround for LLVM shit
```
JohanEngelen, line 1457: language...

ErnyTech:
Ok moved to codegenerator

kinke:
I've just tried this with an LLVM with enabled assertions, which hits an assertion below `llvm::appendToGlobalCtors()`.

ErnyTech:
I currently believe that the failure of the assert is caused by the fact that appendToGlobalArray generates a wrong GlobalVariable (`void` () *` instead of `void () addrspace (1) *`)

I don't think we can do much, the only one would be to reimplement appendToGlobalArray but it's ugly

kinke:
Yeah that's IMO clear, and the proper location to fix would be LLVM anyway.

ErnyTech:
@kinke Do you think it's better to patch our llvm fork?

JohanEngelen:
@ErnyTech Would be much better to submit a patch to LLVM! You can discuss it on their mailing list if you want some guidance first. It's a friendly community and I'm sure you can get your patch landed there.

PR #3438, opened 2020-05-20
GCC-style asm: Add support for indirect input operands

kinke:
I.e., use the expression's lvalue as operand for `"m"` memory inputs, like GDC. Previously, this was restricted to output operands (`"=m"`); input operands were still using the expression's rvalue (and so required an explicit address-of operator for variables).


kinke:
[Amended an lvalue check and according error instead of ICE'ing for rvalues.]

PR #3440, opened 2020-05-20
druntime: Revise upstream diff a bit & speed-up core.math.ldexp for x87 real

kinke:
Testing https://github.com/ldc-developers/druntime/pull/185.

### DIFF
--- a/dmd/builtin.d
+++ b/dmd/builtin.d
@@ -666,7 +666,7 @@ public extern (C++) void builtin_init()
 {
 version (IN_LLVM)
 {
-    builtins._init(204);
+    builtins._init(205);
 }
 else
 {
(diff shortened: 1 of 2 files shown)


dmd/builtin.d
```diff
@@ -736,16 +736,13 @@ else
...
+    // LDC's core.math.ldexp might be an alias of C ldexp[l]
     add_builtin("ldexpl", &eval_ldexp);
-    add_builtin("ldexp", &eval_ldexp); // required for Windows/MSVC targets
-}
-else
-{
-    // @safe @nogc pure nothrow T function(T, int)
-    add_builtin("_D4core4math5ldexpFNaNbNiNfeiZe", &eval_ldexp);
+    add_builtin("ldexp", &eval_ldexp);
 }
```
thewilsonator, line 746: This function has been completely reworked in upstream https://github.com/dlang/dmd/pull/11069

kinke: Looks like more unthankful work for LDC 1.23, oh well.

thewilsonator: Well thankfully the new code is much nicer and more flexible. I'm a bit sick atm (oh the joys of nieces), but I'm happy to help.

kinke: Help is always welcome. - If you find the time and energy, you could prepare the ~90 extra builtins (mostly LLVM intrinsics) for LDC already. Preferably without adding new `Identifier` globals and without relying on them being declared in `ldc.intrinsics` (might be declared manually by other libs as well, and we don't want to break CTFE-ability for them).

PR #3443, opened 2020-05-24
Add --fno-plt option to avoid PLT external calls

kubo39:
This patch gives new option which avoids the PLT and lazy binding while
making external calls.
Implementation inspired by `-fno-plt` support to Clang.

Clang's patch: https://reviews.llvm.org/D39079

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -460,6 +460,10 @@ cl::opt<bool> disableLinkerStripDead(
     cl::desc("Do not try to remove unused symbols during linking"),
     cl::cat(linkingCategory));
 
+cl::opt<bool> noPLT(
+    "fno-plt", cl::ZeroOrMore,
+    cl::desc("Do not use the PLT to make function calls"));
+
 // Math options
 bool fFastMath; // Storage for the dynamically created ffast-math option.
 llvm::FastMathFlags defaultFMF;
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -634,9 +634,15 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
   // add func to IRFunc
   irFunc->setLLVMFunc(func);
 
-  // First apply the TargetMachine attributes, such that they can be overridden
-  // by UDAs.
+  // First apply the TargetMachine attributes and NonLazyBind attribute,
+  // such that they can be overridden by UDAs.
   applyTargetMachineAttributes(*func, *gTargetMachine);
+  if (!fdecl->fbody && opts::noPLT) {
+      // Add `NonLazyBind` attribute to function declarations,
+      // the codegen options allow skipping PLT.
+      func->addFnAttr(LLAttribute::NonLazyBind);
+  }
+
   applyFuncDeclUDAs(fdecl, irFunc);
 
   // parameter attributes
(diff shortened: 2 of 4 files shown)


tests/codegen/noplt.d
```diff
@@ -0,0 +1,8 @@
+// RUN: %ldc --output-ll --fno-plt -of=%t.ll %s && FileCheck %s < %t.ll
+
+// CHECK: Function Attrs: nonlazybind
```
JohanEngelen, line 3: can you improve the checking to make sure that the attribute is applied to `foo` only?

gen/functions.cpp
```diff
@@ -647,6 +647,12 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
     }
   }
 
+  // Add `NonLazyBind` attribute to function declarations,
```
JohanEngelen, line 650: nit: move this up to line 640, before applying UDAs, such that in future there is no change needed when a UDA is added to override this.

gen/functions.cpp
```diff
@@ -647,6 +647,12 @@ void DtoDeclareFunction(FuncDeclaration *fdecl) {
...
+  // the codegen options allow skipping PLT.
+  if (!fdecl->fbody && global.params.disablePLT) {
```
JohanEngelen, line 652: You can just use `opts::disablePLT` here, such that there is no need to add `disablePLT` to `global.params`.

tests/codegen/noplt.d
JohanEngelen: (actually, that is what I now see in the clang PR aswell, so you can see how it is done there)

driver/cl_options.cpp
```diff
@@ -460,6 +460,10 @@ cl::opt<bool> disableLinkerStripDead(
     cl::desc("Do not try to remove unused symbols during linking"),
     cl::cat(linkingCategory));
 
+cl::opt<bool> disablePLT(
+    "fno-plt", cl::ZeroOrMore,
+    cl::desc("disabling PLT"));
```
JohanEngelen, line 465: The description is too brief (and slightly wrong?), better to copy Clang's description.

tests/codegen/noplt.d
```diff
@@ -0,0 +1,8 @@
+// RUN: %ldc --output-ll --fno-plt -of=%t.ll %s && FileCheck %s -check-prefix=CHECK-NOPLT < %t.ll
+
+// CHECK-NOPLT: Function Attrs: nonlazybind
```
JohanEngelen, line 3: you don't need a specific check-prefix here.
Probably a check-next: define{{.*}}foo is good enough.

JohanEngelen:
apart from the test fix, lgtm

tests/codegen/noplt.d
kubo39: Ah, I misunderstood.

tests/codegen/noplt.d
```diff
@@ -0,0 +1,9 @@
...
+// CHECK-NOPLT-NEXT: declare i32 @_D5noplt3fooFZi
```
JohanEngelen, line 4: This fails on Windows because of different mangling.
Have a look at some of the other codegen tests to see how to match on a function name on all platforms

JohanEngelen:
Thanks for the work.

PR #3444, opened 2020-05-25
Add option for RELRO

kubo39:
This patch gives support for fullRELRO/partialRELRO/noRELRO option.
Some linux distributions apply fullRELRO(`-Wl,-z,relro -Wl,-z,now`) by
default, but this option would be useful for others.

### DIFF
--- a/tests/linking/relro_level.d
+++ b/tests/linking/relro_level.d
@@ -1,0 +1,6 @@
+// REQUIRES: Linux
+// RUN: %ldc --gcc=echo --relro-level=full %s | grep -e \\-Wl,-z,relro -e \\-Wl,-z,now
+// RUN: %ldc --gcc=echo --relro-level=partial %s | grep \\-Wl,-z,relro
+// RUN: %ldc --gcc=echo --relro-level=off %s | grep \\-Wl,-z,norelro
+
+void main() {}
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -41,6 +41,28 @@ static llvm::cl::opt<bool> linkNoCpp(
     "link-no-cpp", llvm::cl::ZeroOrMore, llvm::cl::Hidden,
     llvm::cl::desc("Disable automatic linking with the C++ standard library."));
 
+// RELRO options
+enum RELROKind {
+  RELRO_None,
+  RELRO_Full,
+  RELRO_Partial,
+  RELRO_Off,
+};
+
+static llvm::cl::opt<RELROKind> relroLevel(
+    "relro-level", llvm::cl::ZeroOrMore,
+    llvm::cl::desc("Set the read-only relocation hardeing level for Posix target"),
+    llvm::cl::init(RELRO_None),
+    clEnumValues(
+        clEnumValN(RELRO_Full, "full", "full RELRO (equivalent to '-L-z -Lrelro -L-z -Lnow')"),
+        clEnumValN(RELRO_Partial, "partial", "partial RELRO (equivalent to '-L-z -Lrelro')"),
+        clEnumValN(RELRO_Off, "off", "no RELRO (equivalent to '-L-z -Lnorelro')")));
+
+static inline bool isUsingRELRO() { return relroLevel != RELRO_None; }
+static inline bool enableFullRELRO() { return relroLevel == RELRO_Full; }
+static inline bool enablePartialRELRO() { return relroLevel == RELRO_Partial; }
+static inline bool disableRELRO() { return relroLevel == RELRO_Off; }
+
 //////////////////////////////////////////////////////////////////////////////
 
 namespace {
@@ -508,6 +530,17 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
 
   args.push_back("-o");
   args.push_back(std::string(outputPath));
+
+  if (isUsingRELRO()) {
+      if (enableFullRELRO()) {
+          addLdFlag("-z", "relro");
+          addLdFlag("-z", "now");
+      } else if (enablePartialRELRO()) {
+          addLdFlag("-z", "relro");
+      } else if (disableRELRO()) {
+          addLdFlag("-z", "norelro");
+      }
+  }
 
   addSanitizers(*global.params.targetTriple);
 


driver/cl_options.cpp
```diff
@@ -518,6 +518,14 @@ cl::opt<LTOKind> ltoMode(
...
                    "Parallel importing and codegen (faster than 'full')")));
 
+cl::opt<RELROKind> relroLevel(
+    "relro-level", cl::ZeroOrMore, cl::desc("Set the RELRO level"),
+    cl::init(RELRO_None),
+    clEnumValues(
+        clEnumValN(RELRO_Full, "full", "full RELRO (equivalent to '-Wl,-z,relro -Wl,-z,now')"),
+        clEnumValN(RELRO_Partial, "partial", "partial RELRO (equivalent to '-Wl,-z,relro')"),
+        clEnumValN(RELRO_Off, "off", "no RELRO (equivalent to '-Wl,-z,norelro')")));
+
```
kinke, line 528: Should be moved to `linker-gcc.cpp`, adding it to the linker CLI category as well. The description could use a little more context (something like 'Set the read-only relocation hardening level for Posix targets'). Wrt. 'equivalent to', the LDC syntax would probably be more intuitive ('-L-z -Lrelro -L-z -Lnow').

tests/linking/relro_level.d
```diff
@@ -0,0 +1,7 @@
+// REQUIRES: Linux
+// RUN: %ldc --relro-level=full %s -of=%t && (readelf -l %t | grep -q GNU_RELRO)
+// RUN: %ldc --relro-level=full %s -of=%t && (readelf -d %t | grep -q BIND_NOW)
+// RUN: %ldc --relro-level=partial %s -of=%t && (readelf -l %t | grep -q GNU_RELRO)
+// RUN: %ldc --relro-level=off %s -of=%t && (! readelf -l %t | grep -q GNU_RELRO)
```
kinke, line 5: Simpler to test by checking the cc cmdline output via `-v`, like other similar tests. Or using `-gcc=echo` to output it directly, as done in https://github.com/ldc-developers/ldc/blob/b8b7615696e9932ab452e855798f547ebf921e3d/tests/linking/linker_empty.d#L5.

JohanEngelen:
@kubo39 Just curious, do you need this for anything in particular?

MaskRay:
`--enable-relro` configured binutils enables `-z relro` by default.
LLD enables `-z relro` by default.

Why isn't `-z relro` `-z now` enough and you want a dedicated driver option?

kubo39:
> @kubo39 Just curious, do you need this for anything in particular?

The reason I want to add this option is enable `--fno-plt` (which prevents us from [retpoline](https://reviews.llvm.org/D41723) I guess) by default if `--relro-level=full` is passed.

> Why isn't -z relro -z now enough and you want a dedicated driver option?

I though the compiler needs to know `--rerlo-level=full` is passed to enable `--fno-plt` by default, but it may be enough to check the linker flags..?

kubo39:
Anyway, `--relro-level=full` is more obvious than `-z relro -z now` I think.

tests/linking/relro_level.d
```diff
@@ -0,0 +1,6 @@
+// REQUIRES: Linux
```
kubo39, line 1: Ahh, is it better to use `!Windows` here?

kinke: As usual, I'm having a hard time finding Apple docs, but according to https://github.com/EionRobb/purple-discord/issues/294, Mac's ld64 linker doesn't support it.

kubo39: Thanks for letting me know, I'll keep this up!

MaskRay:
> > @kubo39 Just curious, do you need this for anything in particular?
> 
> The reason I want to add this option is enable `--fno-plt` (which prevents us from [retpoline](https://reviews.llvm.org/D41723) I guess) by default if `--relro-level=full` is passed.
> 
> > Why isn't -z relro -z now enough and you want a dedicated driver option?
> 
> I though the compiler needs to know `--rerlo-level=full` is passed to enable `--fno-plt` by default, but it may be enough to check the linker flags..?

(OK, a main contributor of LLD here, I know -z retpolineplt and have implemented the missing piece of -fno-plt (`__tls_get_addr`) in clang and other GOT/PLT related dso_local stuff in LLVM)

I am still not sure the driver option will be very useful and why `-fno-plt` needs to be enabled by default.

Note, -fno-plt is nice on x86-64 when there are certain amount of shared objects. It does avoid PLT entries but it changes 5-byte call/jump to 6-byte call/jump (R_X86_64_GOTPCREL or R_X86_64_GOTPCRELX (since cl/282068431)): binary size regression.

```
# -fno-plt, 6 bytes
callq   *676915(%rip)  # object file uses R_X86_64_GOTPCREL. Unneeded GOT indirection
addr32 callq   31155 <fwrite> # object file uses R_X86_64_GOTPCRELX

# without -fno-plt, 5 bytes
callq 31155 <fwrite>  # object file uses R_X86_64_PLT32
```

Without R_X86_64_GOTPCRELX, there is even a regression if the function turns out to be defined in the executable (in all of GNU ld, gold, and lld)

-fno-plt is only effective on x86-32/x86-64 (GCC and clang). The option essentially inlines the call stubs at the call sites. On a RISC architecture, the great code size increase can make such a trick impractical.

kubo39:
@MaskRay First of all, hanks a lot for your reviews!

> Note, -fno-plt is nice on x86-64 when there are certain amount of shared objects. It does avoid PLT entries but it changes 5-byte call/jump to 6-byte call/jump (R_X86_64_GOTPCREL or R_X86_64_GOTPCRELX): binary size regression.
> 
> ```
> # -fno-plt, 6 bytes
> callq   *676915(%rip)  # object file uses R_X86_64_GOTPCREL. Unneeded GOT indirection
> addr32 callq   31155 <fwrite> # object file uses R_X86_64_GOTPCRELX
> 
> # without -fno-plt, 5 bytes
> callq 31155 <fwrite>  # object file uses R_X86_64_PLT32
> ```
> 
> Without R_X86_64_GOTPCRELX, there is even a regression if the function turns out to be defined in the executable (in all of GNU ld, gold, and lld)
> 

Hmm, I thought fno-plt is nice x86 with pic too by reading [this article](https://ewontfix.com/18/) (though [gcc improved x86 with pic since 5.0](https://software.intel.com/content/www/us/en/develop/blogs/new-optimizations-for-x86-in-upcoming-gcc-50-32bit-pic-mode.html)) .
Anyway, I got that this has a downside for binary size.

> -fno-plt is only effective on x86-32/x86-64 (GCC and clang). The option essentially inlines the call stubs at the call sites. On a RISC architecture, the great code size increase can make such a trick impractical.

I didn't know that.

Okay, I changed my mind. So closing this.

PR #3445, opened 2020-05-26
Remove bitrig support

kubo39:
Bitrig is now defunct OpenBSD fork. It released 1.0 in 2014 and hasn't had another release since.
I'm not sure if LDC officially supports it, just delete an occurence of bitrig from the source.


kubo39:
CI failure appears to be irrelevant.

kinke:
No trace of this in the current clang logic either: https://github.com/llvm/llvm-project/blob/d11155d273af00f75c2b40a5ca3007463f9808c1/clang/lib/Driver/ToolChains/Arch/X86.cpp#L23-L106

PR #3446, opened 2020-05-26
druntime: Adopt fast ldexp implementation from libmir

kinke:
On Win64 and its double-precision real, it seems to be roughly 14x faster than Microsoft's C `ldexp` on my box (i5-3550).

On Ubuntu 20.04, the speed-up is something like 9x, again for double precision. Double and single precision versions aren't enabled yet, but will be with LDC v1.23 (already available in upstream druntime master).

### DIFF
--- a/dmd/builtin.d
+++ b/dmd/builtin.d
@@ -666,7 +666,7 @@ public extern (C++) void builtin_init()
 {
 version (IN_LLVM)
 {
-    builtins._init(205);
+    builtins._init(203);
 }
 else
 {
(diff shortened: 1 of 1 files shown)


kinke:
On Ubuntu, using the x87 asm version for double precision is about 50% slower than `ldexpImpl`. On Windows, the performance was horrific.

kinke:
@9il: This is a long over-due follow-up to https://forum.dlang.org/post/gdqqwoqphvlfdlpqercm@forum.dlang.org. Are you okay with the adoption?

9il:
LGTM. BTW, how ldexpImpl becomes ldexp, I can’t find where it called.

kinke:
With LDC v1.23, the commented-out double and float versions of core.math.ldexp will be enabled. In the meantime, only the real version for non-x87 targets is enabled. It's right below ldexpImp.

PR #3449, opened 2020-05-28
Sync .editorconfig with dmd

kubo39:


.editorconfig
```diff
@@ -1,9 +1,25 @@
+# EditorConfig file: http://editorconfig.org/
+
 root = true
 
-[*.{c,h,d,di,dd}]
-end_of_line = lf
+# Default settings for all file types:
+# + save files in UTF-8 encoding
+# + use Unix line-endings and a final newline
```
kinke, line 7: Seems like the Unix line-endings are gone.

.editorconfig
```diff
@@ -1,9 +1,25 @@
...
+# + remove whitespace at the end of lines
+[*]
+charset = utf-8
 insert_final_newline = true
+trim_trailing_whitespace = true
+
+# File type specific settings:
+[*.{asm,S,c,cpp,m,h,d,di,dd,ddoc,dt,json,sdl,R,bat,bash,fish,sh,zsh}]
```
kinke, line 15: The .cpp files should use `indent_size = 2` (one of the clang-format settings).

.editorconfig
kubo39: DMD removes `end_of_line = lf` in [this PR](https://github.com/dlang/dmd/pull/8656).

kinke: Yeah that's probably fine, I meant the comment.

kubo39: Ahh, I got!

kinke:
Requesting a review from Johan as he added this file. [Seems like no editor I'm using considers this file.]

JohanEngelen:
Hm, I've never used this file either for an editor. I see in the logs that I added it for `dfmt`. I stopped using dfmt too because the output quality is not sufficient to be used in practice.
So I don't really have an opinion on this file, nor a way to test it.

Issue #3108, opened 2019-07-09
FreeBSD: switch to using external (libexecinfo) backtrace function like GDC does


myfreeweb:
The custom backtrace impl at https://github.com/ldc-developers/druntime/blob/ldc/src/core/sys/freebsd/execinfo.d is not working properly on my amd64 machine currently (only shows `_d_createTrace`), even though it might have worked on aarch64 when I added the `llvmasm` thing for it..

GDC is actually using libexecinfo instead of this mess (`version (GNU) version = BacktraceExternal;`). I think LDC should do the same. It actually works :) and it's portable to all architectures.

The plan is:

- druntime: `version (LDC) version = BacktraceExternal;`
- druntime: `version (LDC) { numframes = cast(int)backtrace( callstack.ptr, MAXFRAMES ); }` (execinfo's backtrace returns a `size_t`, add an int cast) (hmm this is only in ldc's fork!)
- LDC/runtime/cmake: add execinfo to `C_SYSTEM_LIBS` for shared druntime
- LDC/driver/linker-gcc: add execinfo to `addDefaultPlatformLibs` for static druntime (is this the right place???)

Do you agree?

---

https://github.com/myfreeweb/ldc/commit/573cb1456b1074bb4ab604a838658f30beb6a797
https://github.com/myfreeweb/druntime/commit/4cc671669d45f14cc8fd270857d5cbeae6b38be8

kinke:
I think this should be discussed upstream, the only LDC-specific aspect here is the question about how to implement the proposed additional C lib dependency - I'd propose a `pragma(lib, "execinfo");` for FreeBSD there instead of tampering with CMake and new hardcoded lib. The ~150 D lines don't look that bad though IMO; trying to figure out & fix the issue(s) might be worthwhile.

myfreeweb:
yeah, `pragma` is a good idea!

> due to non-x86 archs

Indeed — the worst part of those D lines, other than that they're reinventing part of the system's public API, is that it touches inline assembly.

> should be discussed upstream

Well, upstream would probably like to know that LDC has agreed to work with the `version (LDC) version = BacktraceExternal` change :)

kinke:
> yeah, pragma is a good idea!

It might not work for non-Windows though; I think the module with the pragma needs to be imported somewhere in the cmdline performing the linking. For MSVC, we can embed linker cmdline args in the object file, so there the pragma should be sufficient.

[...]

PR #3453, opened 2020-05-30
CirrusCI: Fix the last few FreeBSD failures

kinke:
As the `std.format` unittest failure (missing `-` sign when printing a negative `double` as hex string) smells like a CTFE issue and might be related to a buggy host druntime's `core.stdc.math.log2l`.

The previous host compiler was FreeBSD's LDC v1.14, possibly including unknown patches.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -178,6 +178,18 @@ task:
     cd $CIRRUS_WORKING_DIR/..
     sysctl -n hw.ncpu
     pkg install -y git cmake ninja gmake llvm ldc bash
-    # Use FreeBSD's ldc as host D compiler
-    ln -s /usr/local host-ldc
+  clone_submodules_early_script: |
+    cd $CIRRUS_WORKING_DIR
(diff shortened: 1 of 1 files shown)


kinke:
* The CTFE issue for `std.format` was fixed with https://github.com/dlang/druntime/pull/3119.
* The dmd-testsuite-release failures were all backtrace-related and fixed by switching to FreeBSD's libexecinfo (#3108). Embedding the library dependency via `pragma(lib, "execinfo")` in druntime works nicely, working for both shared and static druntime without having to touch CMake/ldc2.conf/hardcoded libraries, *as long as* using lld 9+. Unfortunately, FreeBSD 12.1 (current stable) comes with clang 8 and lld 8 by default.
* When switching to clang 9 (and apparently implicitly to lld 9), the previous single lit-test failure wrt. invalid relocation for TLS vars and dynamic-compile vanished.
* The `std.path` issue is being fixed in https://github.com/dlang/phobos/pull/7510.

PR #3457, opened 2020-06-04
[RISCV] getTargetCPU should returns generic-rvXX, not generic

kubo39:


JohanEngelen:
What versions of LLVM are people using with risc? (perhaps all the ifdeffing is not really needed?)

kubo39:
> What versions of LLVM are people using with risc? (perhaps all the ifdeffing is not really needed?)

Hmm.. not sure, I just copy-and-paste from driver/main.cpp to suppress CI failure.
These lines were added in [this commit](https://github.com/ldc-developers/ldc/commit/0d9ab693bf854a0cfe4f02e594bf013d4e31c9c4).

kubo39:
RISC-V seems to have been ** **officially** ** added since LLVM 9.0.
https://releases.llvm.org/9.0.0/docs/ReleaseNotes.html#changes-to-the-riscv-target

driver/main.cpp
```diff
@@ -684,12 +684,8 @@ void registerPredefinedTargetVersions() {
   case llvm::Triple::msp430:
     VersionCondition::addPredefinedGlobalIdent("MSP430");
     break;
-#if defined RISCV_LLVM_DEV || LDC_LLVM_VER >= 400
-#if defined RISCV_LLVM_DEV
-  case llvm::Triple::riscv:
-#else
+#if LDC_LLVM_VER >= 900
```
kubo39, line 687: I'm guessing that no one is using RISC-V with LDC in before LLVM9...

kinke: While that might be true, it has just been promoted to a non-experimental target with LLVM 9. The previous `RISCV_LLVM_DEV` was apparently just something for LLVM < 4, i.e., by now 3.9 only. So getting rid of that CMake option and predefine seems fine, but I'd keep on enabling RISC-V support for LLVM 4+.

PR #3459, opened 2020-06-06
[RISCV] GCC RISC-V toolchain does not recognize -m32/m64 switches

kubo39:


kinke:
Ah yeah, another common annoyance for exotic targets. - I've just grepped the gcc 10.1 [cmdline docs](https://gcc.gnu.org/onlinedocs/gcc-10.1.0/gcc-html.tar.gz) for `-m32` and `-m64`. It looks like we should use a whitelist here instead of the current blacklist - gcc apparently supports these only for x86, PowerPC, SPARC and Nvidia PTX. [Some TILE targets, whatever that is, would feature it too, and SystemZ apparently has `-m64` and `-m31` (really, no typo! :]).

kubo39:
Oh, I didn't know that! :+1: for use a whitelist.
Should I close this for now?

kinke:
Superseded by #3460.

PR #3460, opened 2020-06-06
Switch to a white list of archs wrt. -m32/-m64 in cc cmdlines

kinke:
According to the gcc 10.1 cmdline docs.

### DIFF
--- a/driver/tool.cpp
+++ b/driver/tool.cpp
@@ -89,17 +89,20 @@ void appendTargetArgsForGcc(std::vector<std::string> &args) {
 
   const auto &triple = *global.params.targetTriple;
   const auto arch64 = triple.get64BitArchVariant().getArch();
-  const auto arch32 = triple.get32BitArchVariant().getArch();
 
-  // Only specify -m32/-m64 for architectures where the two variants actually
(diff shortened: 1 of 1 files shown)


kubo39:
Maybe it would be better to use `-mx32` for x86_64-unknown-linux-gnux32, not sure LDC supports this architecture.
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

kubo39:
This works fine for me.

```console
$ ./build-ldc/bin/ldc2 --mtriple=riscv64-unknown-none-elf -mcpu=generic-rv64 \
  -mabi=lp64 -mattr=+m,+a,+c --gcc=./gcc/bin/riscv64-unknown-elf-gcc \
  -Xcc=-march=rv64imac -Xcc=-mabi=lp64 -g -betterC hello.d
$ spike pk hello
bbl loaded
Hello, World!
```

Full description: https://github.com/kubo39/ldc-riscv-hello

kinke:
Nice. Is the `-mabi` option for LDC required? It's the first time I see it being used.

kubo39:
Ahh, it's not needed. Only `-Xcc=-abi=lp64` is required.

Issue #3461, opened 2020-06-07
TLS-related issue in Termux LDC 1.21.0-1 release


PetarKirov:
Hi @kinke I have been using LDC 1.20 release in Termux for while without any issues. Recently I updated to LDC 1.21, and when I tried it I noticed that it doesn't work anymore.

Steps to reproduce:
1. Install Termux on an Android mobile device (in my case Android 10, Aarch64)
2. Run `pkg install ldc`
3. Run any command involving ldc: `ldc2 --help`
4. Observe the command failing with the following output:
```
Aborting from rt/sections_android.d(226) expected .tdata and .tbss sections to be contiguousfish: Job 2, 'ldc2 --help' terminated by signal SIGABRT (Abort)
```

(The `contiguousfish` artifact in the output is because of a missing newline and because I'm using Fish shell. But the issue is same if I use bash.)

kinke:
What ld linker are you using, and can you try with the others? - See https://github.com/ldc-developers/ldc/issues/3350#issuecomment-597904606.

kinke:
Btw thx for letting me know it actually works; I've never tested the Termux stuff. With v1.22, Termux won't need any special patches anymore; someone else (who's actually using it) taking over Termux maintenance would be great. ;)

PetarKirov:
> What ld linker are you using, and can you try with the others? 

The problem is not that my code when compiled on Termux crashes, the problem is that LDC itself crashes, so I have no control over the linker - LDC has already been compiled and linked by Termux ;)

> Btw thx for letting me know it actually works; I've never tested the Termux stuff.

It has been at least a year since I have built successfully LDC from source on Termux. Before I did it from the `lts-master` C++ branch, and more recently (a couple of months back) by using LDC as provided by the Termux package manager.

kinke:
Oh, I see now. Can you list the sections of the `ldc2` executable, specifically, `.tdata` and `.tbss`, and post their memory address ranges?

PetarKirov:
Here's what I get:
```
llvm-readelf --sections /data/data/com.termux/files/usr/bin/ldc2

There are 33 section headers, starting at offset 0x2ccdd90:

Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .interp           PROGBITS        0000000000000238 000238 000015 00   A  0   0  1
[...]

PR #3462, opened 2020-06-07
[Android] druntime: Handle TLS init issues

kinke:
Wrt. #3461 and the encountered issue with the rdmd executable.

### DIFF
--- a/.azure-pipelines/posix.yml
+++ b/.azure-pipelines/posix.yml
@@ -165,7 +165,7 @@ steps:
       -DD_COMPILER=$PWD/../bootstrap-ldc/bin/ldmd2 \
       -DCMAKE_INSTALL_PREFIX=$installDir \
       -DINCLUDE_INSTALL_DIR=$installDir/import \
-      -DD_LINKER_ARGS="-L$PWD/../ldc-build-runtime.tmp/lib;-lphobos2-ldc;-ldruntime-ldc" \
+      -DD_LINKER_ARGS="-fuse-ld=bfd;-L$PWD/../ldc-build-runtime.tmp/lib;-lphobos2-ldc;-ldruntime-ldc" \
       "${extraFlags[@]}"
     ninja -j$PARALLEL_JOBS -v ldc2 ldmd2 ldc-build-runtime ldc-profdata ldc-prune-cache
   displayName: 'Android: Cross-compile LDC executables'
--- a/.azure-pipelines/posix.yml
+++ b/.azure-pipelines/posix.yml
@@ -114,12 +114,12 @@ steps:
     mkdir llvm-$ARCH
     tar -xf llvm-$ARCH.tar.xz --strip 1 -C llvm-$ARCH
     # Download & extract Android NDK
-    curl -OL https://dl.google.com/android/repository/android-ndk-r21-linux-x86_64.zip
-    unzip android-ndk-r21-linux-x86_64.zip > /dev/null
-    rm android-ndk-r21-linux-x86_64.zip
+    curl -OL https://dl.google.com/android/repository/android-ndk-$ANDROID_NDK_VERSION-linux-x86_64.zip
+    unzip android-ndk-$ANDROID_NDK_VERSION-linux-x86_64.zip > /dev/null
+    rm android-ndk-$ANDROID_NDK_VERSION-linux-x86_64.zip
     # The NDK toolchain file enforces `-g` as base C[XX] flag - remove it to
     # *significantly* reduce executable sizes
-    TOOLCHAIN_FILE=$PWD/android-ndk-r21/build/cmake/android.toolchain.cmake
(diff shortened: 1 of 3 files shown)


MaskRay:
You only need to overalign the first section in PT_TLS (typically .tdata, but .tss is possible as well).

https://reviews.llvm.org/D62055

kinke:
The code currently makes no assumptions about which section comes first in the final binary.

kinke:
Hmm, doesn't seem to have any effect. The section sizes are identical, so it looks like the dummy symbols are stripped, ignoring the `llvm.used` directives.

MaskRay:
Then the comment is wrong.

   // Enforce some minimum alignment for both .tdata and .tbss sections,
    // to have the first section appropriately padded and hence the second
    // section (expected to be adjacent) ending up suitably aligned in each
    // thread's TLS block.

kinke:
The comment is incomplete anyway (but writing down all details seemed not worth it), and should do if you don't relate `first` and `second` to the order in the opening sentence, but the sections order in the final binary, as traversed by the code.

JohanEngelen:
> Hmm, doesn't seem to have any effect. The section sizes are identical, so it looks like the dummy symbols are stripped, ignoring the `llvm.used` directives.

I recently found out that llvm.used / assumeUsed is only a directive to the "compiler" but not to the "linker". So the linker may still discard the symbol...

kinke:
It works fine for our ModuleInfos...

kinke:
Now apparently working as expected - for AArch64, but not for ARMv7-A (sections not adjacent; alignment would be fine). I've checked the 32-bit ARM ld scripts included in the used Android NDK r21, and all of them specify .tdata right before .tbss, but that doesn't correspond to the final layout.

kinke:
Okay, I've checked all 3 linkers for both AArch64 and ARMv7-A - the only one placing both sections adjacent to each other is bfd, for both archs. So it looks like the NDK uses different default linkers, depending on arch...

JohanEngelen:
> It (assumeUsed) works fine for our ModuleInfo refs...
> 
> **Edit**: And https://github.com/ldc-developers/druntime/blob/f5e154743d844c5ec1ea3b086abb6daca6e1713b/src/rt/sections_elf_shared.d#L971 also appears to assume this just works (not even using @assumeUsed there).

I discovered that things are kept when explicitly accessing the magic start/stop linker symbols. Without that, `lld` discarded many symbols at Weka.

kinke:
@PetarKirov: Please give the aarch64 artifact at https://dev.azure.com/ldc-developers/ldc/_build/results?buildId=1780&view=results a quick check. The executables should all work.

There's one remaining issue though, an apparent disagreement of the NDK and Termux wrt. `-fuse-ld=bfd` - for the NDK, it's required (as default `ld` is gold for all archs but AArch64, where it's bfd); according to your tests, Termux doesn't support it on AArch64, and so requires an explicitly empty `-linker=` (or `-linker=ld`) with this PR. I can add that as default option for future Termux packages, but the official package wouldn't work out of the box in Termux anymore (at least the AArch64 one).

kinke:
> I discovered that things are kept when explicitly accessing the magic start/stop linker symbols. Without that, lld discarded many symbols at Weka.

Yeah, I remember the issues we've had with ModuleInfos refs potentially being stripped by older bfd versions, in combination with `--gc-sections`. And as it happens, the bundled bfd for the latest Android NDK (a few days old) is v2.27, which is from 2016. - It's a pain in the ass trying to outsmart the optimizer.

kinke:
The complications with 2 separate sections being aggregated to a single contiguous memory block could presumably also be handled in our LLVM fork, e.g., by emitting zero-initialized TLS stuff into `.tdata` too. I think that would only increase the executable file size; with a single section, all linkers would work, and users will never be forced to overalign `.tdata` manually.

kinke:
@PetarKirov: Will you be able to quickly check this in the next couple of days? It'd be nice for the final 1.22 release, but I won't delay it by much more.

PetarKirov:
Hi @kinke, sorry for the delay, but as I mentioned previously I'm not able to keep up with GitHub notifications, so it's best to contact me either via the dlang slack or via email.

I've checked both the ldc 1.22.0 package from Termux and the [ldc2-1.22.0-android-aarch64](https://github.com/ldc-developers/ldc/releases/download/v1.22.0/ldc2-1.22.0-android-aarch64.tar.xz) archive from GitHub. The Termux package works out of the box and the official one works after I add `-linker=ld`  to the switches section of ldc2.conf, as otherwise it fails with:
```
clang-9: error: invalid linker name in argument '-fuse-ld=bfd'
Error: /data/data/com.termux/files/usr/bin/cc failed with status: 1
```

This is because BFD is named just `ld`, whereas the other linkers are `ld.gold` and `ld.lld` respectively.

kinke:
Yeah I've known about missing `ld.bfd` for Termux, at least on AArch64. I consider that a Termux issue, as the NDK obviously has the symlink.

PR #3464, opened 2020-06-11
CirrusCI: Add prebuilt package for FreeBSD

kinke:


kinke:
In case someone cares to check (@myfreeweb maybe?): https://github.com/ldc-developers/ldc/releases/download/CI/ldc2-f376f6ad-freebsd-x86_64.tar.xz

It's built on FreeBSD 12.1 x64, using the latest `llvm` package from ports (currently, LLVM 9): https://www.freshports.org/devel/llvm
LDC is linked dynamically against that LLVM, that's why the package is < 18 MB. It's also no multilib package, i.e., only contains the x64 libs (but includes LTO libs, compiler-rt libs etc.). In order to have `libexecinfo` pulled in automatically, it should be coupled with matching clang v9 (and so lld v9) from that ports llvm package (e.g., via `CC` env var or a `-gcc` default switch in `etc/ldc2.conf`).

```
> llvm-readelf --needed-libs ldc2
NeededLibraries [
  libLLVM-9.so
  libc++.so.1
  libc.so.7
  libcxxrt.so.1
  libexecinfo.so.1
  libgcc_s.so.1
  libm.so.5
  libthr.so.3
]
> llvm-readelf --needed-libs dub
NeededLibraries [
  libc.so.7
  libgcc_s.so.1
  libm.so.5
  libthr.so.3
]
```

I don't have a VM, so I haven't run anything myself. And I have absolutely no idea whether this package would also work for FreeBSD 11/13...

kinke:
Note that the Android NDK isn't available for FreeBSD, so using FreeBSD's LLVM instead of our fork shouldn't be a problem.

myfreeweb:
`llvm` is a metapackage that just depends on the "default" llvm, `llvm90` right now. But there's [10](https://www.freshports.org/devel/llvm10) and [devel](https://www.freshports.org/devel/llvm-devel) available as well.

Works on 13, unsurprisingly (system libs are backwards compatible and I don't see how LLVM would have different ABI between OS versions).

I would expect the compiler to work on 11, but dub wouldn't work, it is built with the 64-bit inode ABI:

```
% readelf -s bin/ldc2 | rg readdir
  1485: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND readdir@FBSD_1.0 (2)
 33451: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND readdir
% readelf -s bin/rdmd | rg readdir
    42: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND readdir@FBSD_1.0 (2)
  3414: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND readdir
% readelf -s bin/dub | rg readdir
    53: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND readdir@FBSD_1.5 (4)
 21445: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND readdir
```

kinke:
Thanks a lot for checking! :+1:

Wrt. dub: ~~no idea why that differs from `rdmd` - both are compiled with the final compiler + druntime/Phobos.~~ Oh, looks like I forgot to export the `DMD` env var, so it's using the FreeBSD LDC...

Is there anything to be improved? E.g., did you have to mess with `CC` or `ldc2.conf` for proper linking? Should we specify `-gcc=clang-<LLVM major>` in `ldc2.conf` for a smooth experience, or set `-defaultlib=phobos2-ldc,druntime-ldc,execinfo` to work with all lld versions?

myfreeweb:
Linking just works (it's using the `cc` command).

kinke:
On the FreeBSD 12.1 CI machine, `cc` is the non-ports default clang v8.

myfreeweb:
Yes, and that shouldn't be a problem for linking

kinke:
It should be a problem (and was for CI), because of the `libexecinfo` dependency embedded into static druntime's `.deplibs` section, which is only supported since lld v9.

myfreeweb:
oh, okay.

> Should we specify -gcc=clang-<LLVM major> in ldc2.conf for a smooth experience, or set -defaultlib=phobos2-ldc,druntime-ldc,execinfo to work with all lld versions?

idk, if either works.. I guess linking using clang9 would make LTO work so it's better?

kinke:
Okay, new package available at: https://github.com/ldc-developers/ldc/releases/download/CI/ldc2-4c0fe365-freebsd-x86_64.tar.xz

PR #3469, opened 2020-06-14
Travis CI: Add arm64 job

kinke:


kinke:
Hmm, seems like we have a successor for Shippable. The latter has been skipping builds for some days, apparently because of no nodes in the shared AArch64 node pool. Googling a bit suggested that Shippable might be slowly dying (after having been acquired...).

I dismissed the (experimental) Travis support for AArch64 so far because the containers are officially spec'd as having 2 ARMv8-A cores and 4 GB of memory (vs. the 32-96? cores for Shippable, and 32-128 GB of memory). But Travis' AArch64 integration is supposedly using the same Packet service, so I've checked:
```
$ nproc
32
$ free
              total        used        free      shared  buff/cache   available

Mem:        3906248      371460     2047912       62172     1486876     3534788

Swap:      31256164      367476    30888688
```

Using 16 jobs for compiling LDC (incl. unittests) works and takes just above 17 mins, suggesting that it's indeed usable.

Travis' D integration via generic install script also working on AArch64 out of the box was a pleasent surprise as well.

kinke:
Shippable is working again after letting them know, so this isn't required (yet).

kinke:
Reactivated, as Shippable's queue timings have gone through the roof lately (~2 hours...).

PR #3474, opened 2020-06-17
CI: Link libstdc++ statically for libldc-jit.so of prebuilt Linux packages

### LINKED ISSUE #3473: -enable-dynamic-compile on d.godbolt.org
Below is compiled on ldc 1.21.0 with the flag -enable-dynamic-compile. However, I am having an error when compiling it with [d.godbolt.org](http://d.godbolt.org) ([here](https://d.godbolt.org/z/YqeYSC)). The output is
```
ASM generation compiler returned: 0

Execution build compiler returned: 0

Program returned: 1

./output.s: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /opt/compiler-explorer/ldc1.21.0/ldc2-1.21.0-linux-x86_64/lib/libldc-jit.so.91)
```
Here is the code (reduced from [simple.d](https://github.com/ldc-developers/ldc/blob/00c3e89933b58354117daf49b6c8f48479bd3ed9/tests/dynamiccompile/simple.d)):
```
import ldc.attributes;

@dynamicCompile int foo()
{
  return 5;
}

void main() { }
```

kinke:
Should fix #3473.


kinke:
Seems to work; .so size increase is ~2.7% for the Azure package [with enabled LLVM assertions].

Issue #3374, opened 2020-03-19
Add an option to control if and which system libraries are added by default to the linker command line when using LDC2 for compile and link.


dbankov-vmware:
It would be good to be possible to control which and if system libraries are linked by default when using ldc2 for compiling and linking a binary. The latter is desirable because the code in an application may not need some of these or may need none at all.

The particular use case that I have in mind is when building UEFI application/service/driver where UEFI binaries are built with 'x86_64-unknown-windows-coff' target triple but don't need any of the default windows system libraries. More important it is incorrect to link with these.

kinke:
Yes, I think that's useful for baremetal usages. Something like our existing `-defaultlib` (for the default D libs), which can be used to clear (`-defaultlib=`) or override them (`-defaultlib=lib1,lib2`). [Adding the libs manually via `-L` would probably insert them in the wrong position (before the default D libs).]

Open for suggestions wrt. how to name the new switch, and what it should include *exactly* (I'd propose all non-default-D implicit libraries, except for LLVM compiler-rt libraries).

dbankov-vmware:
Yes, that's exactly what I had in mind. I was thinking about (`-platformlib=`) based on the comment at: https://github.com/ldc-developers/ldc/blob/b56bef247431a4cad4b9b7cb8b0d86364787634a/driver/linker-msvc.cpp#L204-L220
And I think as a start it should include the libs that are listed at the above location which are not controlled with a switch at the moment.

BTW I was planning to send a pull request but decided to file an issue first to discuss the topic and then something urgent appeared so I had to postpone the pull request work. In any-case its great to discuss the details here because whoever makes the change will get less comments during the review ;).

Thanks a lot for taking a look at this!

dbankov-vmware:
I've prepared a PR for this issue but since this is my first PR for LDC I am not sure if what I did is correct at all.
Please let me know if there is a better way to address this.

PR #3475, opened 2020-06-19
Fix for #3374

dbankov-vmware:
Introduce -platformlib option to control libraries added automatically for Windows targets.

### DIFF
--- a/driver/linker-msvc.cpp
+++ b/driver/linker-msvc.cpp
@@ -206,23 +206,33 @@ int linkObjToBinaryMSVC(llvm::StringRef outputPath,
     addSwitch(ls);
   }
 
-  // default platform libs
-  // TODO check which libaries are necessary
-  args.push_back("kernel32.lib");
-  args.push_back("user32.lib");
-  args.push_back("gdi32.lib");
-  args.push_back("winspool.lib");
-  args.push_back("shell32.lib"); // required for dmain2.d
-  args.push_back("ole32.lib");
-  args.push_back("oleaut32.lib");
-  args.push_back("uuid.lib");
-  args.push_back("comdlg32.lib");
-  args.push_back("advapi32.lib");
+  llvm::Optional<std::vector<std::string>> platformLibNames =
+    getExplicitPlatformLibs();
 
-  // these get pulled in by druntime (rt/msvc.c); include explicitly for
-  // -betterC convenience (issue #3035)
-  args.push_back("oldnames.lib");
-  args.push_back("legacy_stdio_definitions.lib");
+  if (platformLibNames.hasValue()) {
+    args.insert(
+      args.end(),
+      std::make_move_iterator(platformLibNames.getValue().begin()),
+      std::make_move_iterator(platformLibNames.getValue().end()));
+  } else {
+    // default platform libs
+    // TODO check which libaries are necessary
+    args.push_back("kernel32.lib");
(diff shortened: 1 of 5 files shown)


driver/linker-msvc.cpp
```diff
@@ -85,6 +94,26 @@ void addSanitizerLibs(std::vector<std::string> &args) {
 
 //////////////////////////////////////////////////////////////////////////////
 
+static std::vector<std::string> getPlatformLibNames() {
```
kinke, line 97: Thx for following up. - Please move this to `linker.cpp` (perhaps as `getExplicitPlatformLibs()` returning an `llvm::Optional<std::vector<string>>`) as we'll support this option for Posix targets too (see [here](https://github.com/ldc-developers/ldc/blob/9eb9f5a568ec21f84449d039132e25ce2abfc238/driver/linker-gcc.cpp#L630-L681)). The tokenizing can then be extracted as little helper, used for both `-defaultlib` and `-platformlib`.

dbankov-vmware: Finally got some time to do this. I'm sorry for the delay.
Please take a look and let me know if I should change something.

tests/linking/platformlib.d
```diff
@@ -0,0 +1,21 @@
+/* Make sure -platformlib overrides the default platform libraries list.
+ * We only care about the platform libs in the linker command line;
+ * make sure linking fails in all cases (no main()) as linking would
+ * fail without the platform libraries anyway. Finally this option is
+ * relevant only for windows targets so make sure we target Windows with
+ * -mtriple=x86_64-unknown-windows-coff.
+ */
+
+// RUN: not %ldc -v -mtriple=x86_64-unknown-windows-coff -platformlib= %s | FileCheck %s
```
JohanEngelen, line 9: can you also add a testcase where you pass some strings through the new option?

driver/linker-msvc.cpp
kinke: No worries. - I've pushed a commit, slightly revising a few things, extending it to Posix targets too as well as extending the tests as requested by Johan.

dbankov-vmware: Thank you very much Kinke!
I was going to take care of this over the weekend as I've been quite busy at work lately...

kinke:
[I've apparently hit some GitHub bug while merging this - this shouldn't have been a squash-merge, plus I've customized the merge commit msg - then got an error and pressed a retry button, which squash-merged it without further confirmation...]

dbankov-vmware:
I'm not particularly proud of my part of the change so IMO it is better that things happened this way :).

kinke:
Well, if it wasn't for you and your initiative, we wouldn't have this. :) - This wasn't a squash merge, but a normal merge with the actual merge commit (title + msg) looking like a squash-merge-commit...

denizzzka:
What exactly is the difference between `--platformlib` and `--defaultlib`?

kinke:
IIRC, mainly that `-defaultlib` doesn't apply to `-betterC`, and that the default lib names are affected by `-link-defaultlib-{shared,debug}` (extra suffix). And for a simpler overriding of the default libs, e.g., `-defaultlib=druntime-ldc` when wanting to exclude Phobos, or `-defaultlib=phobos2-ldc-lto,druntime-ldc-to` to use the LTO variants.

PR #3478, opened 2020-06-26
Fix naked DMD-style asm emission for non-Mac x86 Darwin targets & add prebuilt druntime/Phobos for iOS/x86_64 to macOS package

kinke:
This fixes the issue reported in https://forum.dlang.org/thread/hgkramhteqhfxirigxpq@forum.dlang.org.

### DIFF
--- a/.azure-pipelines/posix.yml
+++ b/.azure-pipelines/posix.yml
@@ -267,27 +267,40 @@ steps:
   displayName: 'Android: Cross-compile x86 libraries and copy to install dir'
   condition: and(succeeded(), eq(variables['CI_OS'], 'android'))
 
-# Mac: add iOS/arm64 libraries
+# Mac: add iOS libraries (arm64 and x86_64)
 - script: |
     set -ex
     cd ..
     export PATH="$PWD/ninja:$PATH"
-    triple="arm64-apple-ios$IOS_DEPLOYMENT_TARGET"
-    # TODO: shared libs too, and look into `-fvisibility=hidden` requirement
+    triple_arm64="arm64-apple-ios$IOS_DEPLOYMENT_TARGET"
+    triple_x64="x86_64-apple-ios$IOS_DEPLOYMENT_TARGET"
+    # TODO: build shared libs too, and look into `-fvisibility=hidden` requirement
     build/bin/ldc-build-runtime --ninja -j $PARALLEL_JOBS \
       --buildDir=build-libs-arm64 \
-      --cFlags="-target;$triple" \
-      --dFlags="-mtriple=$triple;-fvisibility=hidden" \
+      --cFlags="-target;$triple_arm64" \
+      --dFlags="-mtriple=$triple_arm64;-fvisibility=hidden" \
       --ldcSrcDir=$BUILD_SOURCESDIRECTORY \
       CMAKE_SYSTEM_NAME=iOS \
       CMAKE_OSX_ARCHITECTURES=arm64 \
       CMAKE_OSX_DEPLOYMENT_TARGET=$IOS_DEPLOYMENT_TARGET \
       BUILD_SHARED_LIBS=OFF \
       BUILD_LTO_LIBS=ON
-    mkdir installed/lib-ios-arm64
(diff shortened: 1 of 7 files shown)


.azure-pipelines/posix.yml
```diff
@@ -267,35 +267,48 @@ steps:
...
-# Mac: add iOS/arm64 libraries
+# Mac: add iOS libraries (arm64 and x86_64)
 - script: |
     set -ex
     cd ..
     export PATH="$PWD/ninja:$PATH"
-    triple="arm64-apple-ios$IOS_DEPLOYMENT_TARGET"
-    # TODO: shared libs too, and look into `-fvisibility=hidden` requirement
+    triple_arm64="arm64-apple-ios$IOS_DEPLOYMENT_TARGET"
+    triple_x64="x86_64-apple-ios$IOS_DEPLOYMENT_TARGET"
```
jacob-carlborg, line 276: I’m not sure if it matters but Xcode does not use this triple for the simulator. I’m not currently at the computer but, IIRC, it uses `iphonesimulator` instead of `iOS`.

.azure-pipelines/posix.yml
```diff
@@ -267,35 +267,48 @@ steps:
...
       BUILD_SHARED_LIBS=OFF \
       BUILD_LTO_LIBS=ON
-    mkdir installed/lib-ios-arm64
+    build/bin/ldc-build-runtime --ninja -j $PARALLEL_JOBS \
+      --buildDir=build-libs-x86_64 \
+      --cFlags="-target;$triple_x64" \
+      --dFlags="-mtriple=$triple_x64;-fvisibility=hidden" \
+      --ldcSrcDir=$BUILD_SOURCESDIRECTORY \
+      CMAKE_SYSTEM_NAME=iOS \
+      CMAKE_OSX_SYSROOT=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk \
```
jacob-carlborg, line 294: Hard coding the path to the SDK should be avoided. The following command can be used retrieve the path: `xcrun —show-sdk-path —sdk iphonesimulator`.

.azure-pipelines/posix.yml
```diff
@@ -304,10 +317,28 @@ steps:
...
+    {
+        switches = [
+            \"-defaultlib=phobos2-ldc,druntime-ldc\",
+            \"-link-defaultlib-shared=false\",
+            \"-fvisibility=hidden\",
+            \"-Xcc=-target\",
+            \"-Xcc=$triple_x64\",
+            \"-Xcc=-miphoneos-version-min=$IOS_DEPLOYMENT_TARGET\",
+            \"-Xcc=-isysroot\",
+            \"-Xcc=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk\",
```
jacob-carlborg, line 332: Is it possible to run the above mentioned command to avoid hard coding the path?

kinke: It's hardcoded just like the one for the arm64 triple, and noone has complained about that so far.

jacob-carlborg: I know. I'm just saying it would be better if it wasn't hard coded. Not for arm64 either. If someone only installs the command line tools and not Xcode, the path to the SDKs will be different. Also, Apple may decide to change the path. Originally when I started developing on Mac, the SDK was located somewhere in `/Developer`.

kinke: If we wanted to do this properly for the few users who might run into troubles (and aren't able to fix it themselves by adapting ldc2.conf), we would probably have to do something similar to the MSVC toolchain setup, invoking that `xcrun` cmdline and parsing its result to obtain the proper path. Of course none of this would be necessary (none of these `-Xcc` flags) if Apple just used a little preconfigured wrapper à la `x86_64-apple-i{os,phonesimulator}12.0-clang` like the Android NDK or most cross-gcc toolchains.

PR #3480, opened 2020-06-28
Stop generating _d_eh_personality with -betterC

### LINKED ISSUE #3479: linker error with scope(exit) with betterC
so i dont really know much about compilers or linkers but i managed to create a reproducible code snippet which breaks only when using ldc.

https://run.dlang.io/is/NH9MGM
```d
extern(C) int main() {
    int a = 5;
    scope(exit) a = 6;

    create(0, 1, "2");

    return 0;
}

void create(uint a, uint b, string c) {
}
```

the error is:
```
onlineapp-2f059b3.o:onlineapp.d:DW.ref._d_eh_personality: error: undefined reference to '_d_eh_personality'
```

if you uncomment either the `scope(exit)` line or the `create` line everything compilers and works ok.

i dont know how to go even further with finding the cause of the issue so just gonna drop this here. this might be something obvious or an actual compiler/linker bug.

kubo39:
Closes #3479

### DIFF
--- a/gen/funcgenstate.cpp
+++ b/gen/funcgenstate.cpp
@@ -9,6 +9,7 @@
 
 #include "gen/funcgenstate.h"
 
+#include "dmd/globals.h"
 #include "dmd/identifier.h"
 #include "gen/llvm.h"
 #include "gen/llvmhelpers.h"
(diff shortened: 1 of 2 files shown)


kubo39:
Hmm, it seems ldc should not generate landing pads when `--checkaction=halt` was given.

kinke:
Thx for tackling this. - The issue should be the same as reported in the forum a short while ago, and which I've redirected to https://issues.dlang.org/show_bug.cgi?id=20897, because I think the frontend should take care of not emitting any TryFinallyStatements in betterC mode (everything that can be handled there is to be preferred, so that we don't have these ugly special cases in each compiler backend). It generates this for the testcase:
```D
int a = 5;
try
{
    create(0u, 1u, "2");
    return 0;
}
finally
    a = 6;
```

kinke:
Ouch, I've completely overlooked that `return 0`...

kubo39:
Thanks for having a look, I agree with that so closing this.

PR #3481, opened 2020-06-28
Adds output of targetOptions.ExceptionModel to log

denizzzka:


JohanEngelen:
now that you are working on this, can you also add the exceptionmodel to the cache hash calculation, just to be safe? Thanks!
See https://github.com/ldc-developers/ldc/blob/master/driver/cache.cpp#L298-L309

denizzzka:
@JohanEngelen I made this draft just to skip building LDC binary by myself

PR #3482, opened 2020-06-28
-betterC: Don't use unsupported EH for handling clean-ups

kinke:
A reboot of #3480, fixing #3479 and related issues.


kinke:
Pinging @kubo39 - looks like you were absolutely right, sorry for the noise. The extended test hopefully also covers the other guy's issue with structs with dtors.

PR #3485, opened 2020-07-01
Set Module Metadata "RtLibUseGOT" when fno-plt is used

kubo39:
Follow-up https://reviews.llvm.org/D42217

### DIFF
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -138,6 +138,10 @@ void emitLLVMUsedArray(IRState &irs) {
   llvmUsed->setSection("llvm.metadata");
 }
 
+void emitRtLibUseGOT(llvm::Module &M) {
+  M.addModuleFlag(llvm::Module::Max, "RtLibUseGOT", 1);
+}
+
 bool inlineAsmDiagnostic(IRState *irs, const llvm::SMDiagnostic &d,
                          unsigned locCookie) {
   if (!locCookie) {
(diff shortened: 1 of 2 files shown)


gen/functions.cpp
```diff
@@ -641,10 +641,11 @@ void DtoDeclareFunction(FuncDeclaration *fdecl, const bool willDefine) {
...
   applyTargetMachineAttributes(*func, *gTargetMachine);
-  if (!fdecl->fbody && opts::noPLT) {
-      // Add `NonLazyBind` attribute to function declarations,
-      // the codegen options allow skipping PLT.
-      func->addFnAttr(LLAttribute::NonLazyBind);
+
+  // Add `NonLazyBind` attribute to function declarations,
+  // the codegen options allow skipping PLT.
+  if (opts::noPLT && !fdecl->ir->isDefined()) {
+    func->addFnAttr(LLAttribute::NonLazyBind);
```
kinke, line 648: At this point, `fdecl->ir->isDefined()` should always be false, regardless of whether the function is defined later on. So if you want to clear it for defined functions, you should do so in `DtoDefineFunction()`.

kubo39: > At this point, fdecl->ir->isDefined() should always be false, regardless of whether the function is defined later on. 

Ahh.. indeed, thanks for having a look.
And now I guess it is safe for adding `NoLazyBind` for any functions, but I'm not sure...

kinke: So you've changed this again, this time to body-less functions only - is this sufficient?

JohanEngelen:
@kubo30 why did you close this PR?

kubo39:
@JohanEngelen sorry, it was an accident.

JohanEngelen:
LGTM
Can we pull this? (no longer draft status?)

PR #3486, opened 2020-07-01
Refactoring: Specialize IrAggr into IrStruct and IrClass

kinke:
And move ClassInfo generation from `gen/classes.cpp` to `ir/irclass.cpp`.


kinke:
This is preparational work for moving `TypeInfo_Struct` generation to a new `irstruct.cpp`. I plan on revamping the TypeInfo emission for structs, moving to only-emit-in-owning-module, just like ClassInfos and just like the struct init symbols (if non-zero). This hopefully enables to get rid of the xopCmp/xopEquals etc. special cases, incl. extra semantic and emission into non-owning modules.

kinke:
Does anyone know what that `GENERATE_OFFTI` CMake option is and whether someone's using it?

JohanEngelen:
(I prefer if you could keep this PR a "NFC", only doing refactoring. Then it's easier to see the functional changes in the other PR)

No clue what the OFFTI option is...

JohanEngelen:
So basically we have not been emitting complete typeinfo since... forever?

kinke:
See latest commit. I've read your latest comment just now, after pushing the removal of GENERATE_OFFTI. I don't think it really qualifies as functional change though, because I don't think anyone has been using this. Anyway, if there are concerns, I can transplant it to another PR. - The real functional changes are fully intended to land in the other PR, exactly for the visibility reason you mentioned.

kinke:
As with any refactoring of this size, slight functional changes creep in. E.g., ClassInfos for interfaces should now feature `m_RTInfo = 0` (=> no pointers) when it was previously `1` (but their `m_flags` still don't have `ClassFlags.noPointers` set) - analogous to DMD.

JohanEngelen:
"the normal `TypeInfo_Class.interfaces` array contains the offsets nowadays"
Is that then mapped to `object.TypeInfo.offTi` ? https://dlang.org/library/object/type_info.off_ti.html   Or is this always `null`, also for DMD, and perhaps it should be deprecated in dlang?

kinke:
Always null (but the `ClassFlags.hasOffTi` is *always* set, incl. interfaces...). Should IMO clearly be deprecated upstream.

PR #3489, opened 2020-07-02
Set PIC level metadata when relocation model is PIC

kubo39:


kinke:
Please give a rationale - is this only required for MIPS, and only because LLVM doesn't have the right default setting?

tests/codegen/linker_directives_linux.d
```diff
@@ -2,8 +2,8 @@
 
 // REQUIRES: atleast_llvm500, target_X86
 
-// CHECK: !llvm.dependent-libraries = !{!0}
-// CHECK: !0 = !{!"mylib"}
+// CHECK: !llvm.dependent-libraries = !{!{{[0-9]+}}}
```
kinke, line 5: Regexes can be named (`[[LIB:[0-9]+]]`) and later referenced in other places (`[[LIB]]`), in order not to weaken the tests. See https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-string-substitution-blocks.

kubo39:
> Please give a rationale - is this only required for MIPS, and only because LLVM doesn't have the right default setting?

Sorry for that, I re-read recent LLVM source code and noticed that this only affects for PowerPC 32-bit architecture to generate PLT references. Closing this, sorry for noise.

PR #3491, opened 2020-07-03
Emit struct TypeInfos in referencing compilation units only

kinke:
**First commit only**:

Analogous to ClassInfos, incl. normal linkage (external for non-templates, weak_odr for templates).

This enables to get rid of frontend logic wrt. whether to add `TypeInfoStructDeclarations` to a module's members tree - previously,
it was defined as linkonce_odr in the owning module and each referencing module (unless speculative) - and related extra semantic and codegen for the special member functions.
I've gone a bit further and moved the entire TypeInfo emission for LDC to the codegen layer; no `TypeInfoDeclarations` are added to the module members anymore. Whenever we need a TypeInfo symbol during codegen, it is declared or defined, and we don't need to rely on brittle frontend logic with speculative-ness complications.

This might slightly increase compilation speed due to less emitted TypeInfos and functions (possibly less work for the linker too).

It might require LTO to avoid performance regressions though and delegates the job of stripping unused struct TypeInfos to the linker, as the TypeInfo is guaranteed to end up in the owning object file due to no linkonce_odr.

Re-emitting the struct TypeInfos (and optionally the special member functions too) into each referencing CU could be handled in our codegen layer, which should be much simpler and more robust than the upstream scheme.

### DIFF
--- a/dmd/aggregate.h
+++ b/dmd/aggregate.h
@@ -167,10 +167,12 @@ public:
     bool hasIdentityEquals;     // true if has identity opEquals
     bool hasNoFields;           // has no fields
     bool hasCopyCtor;           // copy constructor
+#if !IN_LLVM
     // Even if struct is defined as non-root symbol, some built-in operations
     // (e.g. TypeidExp, NewExp, ArrayLiteralExp, etc) request its TypeInfo.
     // For those, today TypeInfo_Struct is generated in COMDAT.
     bool requestTypeInfo;
+#endif
 
     FuncDeclarations postblits; // Array of postblit functions
     FuncDeclaration *postblit;  // aggregate postblit
--- a/dmd/dstruct.d
+++ b/dmd/dstruct.d
@@ -104,7 +104,10 @@ extern (C++) void semanticTypeInfo(Scope* sc, Type t)
             Scope scx;
             scx._module = sd.getModule();
             getTypeInfoType(sd.loc, t, &scx);
+version (IN_LLVM) {} else
+{
             sd.requestTypeInfo = true;
+}
         }
         else if (!sc.minst)
         {
@@ -114,7 +117,10 @@ extern (C++) void semanticTypeInfo(Scope* sc, Type t)
         else
         {
             getTypeInfoType(sd.loc, t, sc);
+version (IN_LLVM) {} else
+{
             sd.requestTypeInfo = true;
+}
 
             // https://issues.dlang.org/show_bug.cgi?id=15149
             // if the typeid operand type comes from a
@@ -122,6 +128,13 @@ extern (C++) void semanticTypeInfo(Scope* sc, Type t)
             // unSpeculative(sc, sd);
         }
 
+version (IN_LLVM)
+{
+        // LDC defines a struct's TypeInfo (only) once in its owning module,
+        // including the special members, as part of StructDeclaration codegen.
+}
+else
+{
         /* Step 2: If the TypeInfo generation requires sd.semantic3, run it later.
          * This should be done even if typeid(T) exists in speculative scope.
          * Because it may appear later in non-speculative scope.
@@ -150,6 +163,7 @@ extern (C++) void semanticTypeInfo(Scope* sc, Type t)
                 Module.addDeferredSemantic3(sd);
             }
         }
+} // !IN_LLVM
     }
 
     void visitTuple(TypeTuple t)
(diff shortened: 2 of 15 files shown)


kinke:
Thoughts and/or compile/runtime numbers would be appreciated.

kinke:
@JohanEngelen: This also eliminates some `needsCodegen()` calls (e.g., in disabled `isSpeculativeType()`).

JohanEngelen:
@kinke  It will be a while before I can look at this and test it , sorry

>It might require LTO to avoid performance regressions 

What specific type of perf regression do you expect?

kinke:
I'm in no hurry wrt. this, so take your time. :)

> What specific type of perf regression do you expect?

I was thinking about cases where the optimizer might have been able to inline xtoHash/xopEquals/xopCmp/xtoString/xdtor[ti]/xpostblit indirections. But after looking at `TypeInfo_Struct`, these (public) members are probably hardly used directly, and instead used by the virtual TypeInfo_Struct methods equals/compare/getHash/destroy/postblit, which are in druntime's `object.o` object file only, so that's probably not an issue at all.

kinke:
Okay to merge? The runtime performance concerns shouldn't apply as mentioned above; size-wise, I've noticed some slight increases for built executables, but nothing dramatic IIRC. - It's a prerequisite for a clean reboot of #3422.

kinke:
This does have quite an impact - some binary size numbers (in bytes) for Win64, based on the Azure packages:

| | master | this |
|---|---|---|
| bin dir (self-hosted) | 130,823,680 | 130,816,512 (-0%) |
| static druntime debug | 9,873,312 | 9,859,848 (-0.1%) |
| static druntime | 4,498,708 | 5,617,442 (+24.9%) |
| static Phobos debug | 18,536,984 | 18,477,206 (-0.3%) |
| static Phobos | 8,916,078 | 9,155,182 (+2.7%) |

So e.g. apparently lots of directly-unused struct TypeInfos in druntime which aren't stripped anymore with enabled optimizations (but e.g. also aren't emitted in user-code and Phobos CUs anymore if the structs aren't inside templates).

kinke:
And the debug lib sizes probably show that at least for druntime and Phobos, there are hardly any extra emissions in non-owning CUs.

Maybe we should instead strive for emitting them *only* in referencing CUs as linkonce_odr, not in the owning one at all if not referenced. At least the TypeInfos themselves, not sure about the special member functions.

JohanEngelen:
> Okay to merge? The runtime performance concerns shouldn't apply as mentioned above; size-wise, I've noticed some slight increases for built executables, but nothing dramatic IIRC. - It's a prerequisite for a clean reboot of #3422.

I'll try this on weka's code this week.

kinke:
Wrt. the 2nd approach in the 2nd commit, I don't have a good idea wrt. the missing dtor linker failure in `runnable\link15017.d`, and why it works for Linux but doesn't work Windows and Mac. But I haven't looked at it in detail yet. Anyway, the numbers are quite interesting (percentages relative to master again):

| | this (2nd commit) |
|---|---|
| static druntime debug | 8,746,420 (-11.4%) |
| static druntime | 4,514,284 (+0.3%) |
| static Phobos debug | 18,241,660 (-1.6%) |
| static Phobos | 8,948,460 (+0.4%) |

kinke:
> I don't have a good idea wrt. the missing dtor linker failure in runnable\link15017.d

Looked at it (on Windows), and for some bizarre reason, the extra `semanticTypeInfoMembers()` call leads to the (already sema3'd!) dtor losing the `pure @safe` attributes...

JohanEngelen:
No linking errors at Weka, good news. Attribute deduction problems (`pure @safe`) are very common at Weka after dlang frontend changes, so it's good news that this didn't introduce any for that codebase.

Some compilation size data from Weka:
```
 Before:        After:
   2338208     2223504
   1196240     1082032
  45730696    44155640
    178768      155232
     59984       39816
   3738096     3660928
   2100416     2024160
    916296      884192
   1172320     1151992
    557808      496680
    448144      386880
  10752160    10269976
1005558456  1004152488   # yes the unoptimized binary is 1 GB.
    314144       72432
```

That's a net size reduction across the board. Measuring compilation time will be very flawed, but at least there was no obvious slow down / speed up.

kinke:
Thx for the numbers! [And I'm surprised there were no linker errors...]

Upstream, Walter and Andrei are working on getting rid of these compiler-built TypeInfos (struct TypeInfos at least) and replacing them by straight-forward templates in druntime. So I'm not sure how long that takes and whether getting to the bottom of this weird attributes-inference problem in the meantime is worth it.

kinke:
Found the reason for the failure; it was (deprecated) `delete` specific (a general front-end issue apparently), thus not reall surprising there weren't any linker issues for Weka.

kinke:
To elaborate a bit: when a function hasn't had its attributes inferred yet, but is checked for being pure/safe/..., it is set as impure/non-safe, and a later sema doesn't infer them anymore...

JohanEngelen:
> To elaborate a bit: when a function hasn't had its attributes inferred yet, but is checked for being pure/safe/..., it is set as impure/non-safe (wtf?), and a later sema doesn't infer them anymore...

We generally do have linking errors related to attribute inference. Weka's codebase has a bunch of explicit attributes on templates to fix those.

kinke:
That's why I elaborated a bit. The proposed upstream fix for this particular case is https://github.com/dlang/dmd/pull/11854, maybe it sparks a deeper discussion about why things are being done this way.

Issue #3490, opened 2020-07-03
Excessive Allocation for StringExp with Cache


looked-at-me:
There's an issue where `e->toChars()` linked below, allocates new memory and the string can be rather large sometimes. It is building a "header" pretty version of the string, so for a large string it allocates quite a bit extra memory (x3) each time. I tried it with peekString() and was able to reduce compile time of one my projects by about 1 second (from a 19 second total). I `import("...")`  quite a few files but the sizes aren't that big but still quite significant considering what this code is doing.

Don't think peekString() would be portable though, all my strings are just regular strings. Any better ideas?

Some of the other optimizations below wouldn't be necessary if something else was used instead of the actual string. But otherwise they are quite slow for larger strings.

https://github.com/ldc-developers/ldc/blob/v1.22.0/gen/toir.cpp#L394
```D
    llvm::StringMap<llvm::GlobalVariable *> *stringLiteralCache =
        stringLiteralCacheForType(cty);
    LLConstant *_init = buildStringLiteralConstant(e, true);           // *** These can go into the if statement below don't think they have side effects
                                                                       // *** only used to construct gvar first time
    const auto at = _init->getType();                                  // *** ditto

    llvm::StringRef key(e->toChars());                                 // *** allocates memory excessively
    llvm::GlobalVariable *gvar =
        (stringLiteralCache->find(key) == stringLiteralCache->end())   // *** can eliminate the operator[] here, avoid a second scan.
            ? nullptr
            : (*stringLiteralCache)[key];
    if (gvar == nullptr) {
      llvm::GlobalValue::LinkageTypes _linkage =
          llvm::GlobalValue::PrivateLinkage;
      IF_LOG {
        Logger::cout() << "type: " << *at << '\n';
        Logger::cout() << "init: " << *_init << '\n';
      }
      gvar = new llvm::GlobalVariable(gIR->module, at, true, _linkage, _init,
                                      ".str");
      gvar->setUnnamedAddr(llvm::GlobalValue::UnnamedAddr::Global);
      (*stringLiteralCache)[key] = gvar;
    }
```

kinke:
Seems to makes perfect sense, would you mind opening a PR with your suggested changes?

looked-at-me:
[...]

PR #3492, opened 2020-07-04
Refactor toir visit(StringExp) to avoid un-necessary computations.

looked-at-me:
This doesn't include peekData() as described in https://github.com/ldc-developers/ldc/issues/3490 not sure about the process about updating the interface files, if it's fine I'll just add it into this PR.

### DIFF
--- a/dmd/expression.h
+++ b/dmd/expression.h
@@ -390,6 +390,11 @@ public:
         assert(sz == 1);
         return {len, static_cast<const char *>(string)};
     }
+
+    DArray<const unsigned char> peekData() const
+    {
+        return {len * sz, static_cast<const unsigned char *>(string)};
+    }
 #endif
     size_t numberOfCodeUnits(int tynto = 0) const;
     void writeTo(void* dest, bool zero, int tyto = 0) const;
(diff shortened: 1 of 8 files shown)


kinke:
> not sure about the process about updating the interface files

You could just add an inline wrapper returning a `DString`, analogous to https://github.com/ldc-developers/ldc/blob/1494343d6e2687234339fa6c6a242dd5a7003eff/dmd/expression.h#L386-L393

kinke:
Thx for figuring out the test failures. - Seems like the code duplication is pointless, so this should be further refactored. It doesn't have to be you, and it doesn't have to be in this PR, but it's ugly as hell once one knows about it.

kinke:
Thx! I've pushed 2 more commits and am finally satisfied. ;)

looked-at-me:
LGTM, the failure is due to a timeout. Not sure if you want to run it again?

kinke:
I've learned to accept unfortunate sporadic failures and CI hickups. [Plus I can't retrigger or cancel Travis jobs as I used to.]

In case you're wondering about the formatting - we use clang-format for a consistent C++ style.

PR #3493, opened 2020-07-04
Raise min LLVM version to 6.0

kinke:


kinke:
I wasn't really sure whether to raise to 5 or 6, but went with 6 in the end, mostly because most non-ancient distro versions ship with v6 or newer, see https://pkgs.org/download/llvm. E.g., Ubuntu jumped directly from v3.8 (16.04) to v6.0 (18.04), and Debian from v3.8 (stretch) to v7.0 (buster), so most users shouldn't be affected.

dnadlinger:
LLVM 6 was released in March 2018, so should be fine in that respect.

The only reason to hold off on this I can see would be if someone has a weird custom backend that is pre-6 only.

JohanEngelen:
The amount of code lines this removes is incredible.

kinke:
Heh yeah, but most of it is just our version-specific profdata/FileCheck copies.

kinke:
I've given the guys a week to respond to the forum thread, I think that's enough.

Geod24:
This broke Buildkite: https://buildkite.com/dlang/dmd/builds/14258#2579093e-dd2a-478a-8ae0-ef6f30275231

PR #3494, opened 2020-07-05
CMake: Set up Visual D when using the Visual Studio generator

kinke:
For a smooth out-of-the-box experience after something like:
```
cmake C:\LDC\ldc
      -G "Visual Studio 16 2019"
      -A x64
      -DLLVM_ROOT_DIR=C:\LDC\LLVM-x64
      -DD_COMPILER=C:\LDC\ldc2-1.22.0-windows-multilib\bin\ldmd2
start ldc.sln
```


kinke:
FYI @rainers. Apparently it's not a perfect solution (https://cmake.org/cmake/help/latest/prop_tgt/VS_USER_PROPS.html), but it seems to suffice in my case (edit & debug in VS, build in the shell with ninja) - tested with Visual D v1.0. There are still no Visual D properties for the project, I guess that would require adding some dummy `<DCompile />` item.

kinke:
Btw Rainer, the back button on my mouse doesn't navigate backward after jumping with F12. The navigation buttons in the toolbar work as expected, but the extra mouse buttons don't (but do for C++ files).

rainers:
> There are still no Visual D properties for the project, I guess that would require adding some dummy `<DCompile />` item (CMake adds all .d files as `<None />` items).

Indeed, the options only appear if there is at least one file included for rule `<DCompile>`. Maybe you can generate this instead of the custom rules with a combination of `VS_TOOL_OVERRIDE` and `VS_SOURCE_SETTINGS_<tool>`.

rainers:
> Btw Rainer, the back button on my mouse doesn't navigate backward after jumping with F12

IIRC that functionality is not covered by any command that can be reimplemented, but has to be done by explicit mouse handling hooking into the text editor. Something similar has been added recently to Visual D, so I might be able to revisit that.

kinke:
Thx for the `VS_TOOL_OVERRIDE` hint, that works indeed to make them all `<DCompile>`. I got rid of the custom command (hey, now there's an actual `ldc2` project in the VS solution...). Building in VS this way, i.e., letting Visual D compile the .d files as oart of the `LDCShared` lib (usually, just the C++ parts), seems to almost work, the linker is just complaining about conflicting MSCRT versions. Plus there's something like 17k warnings (LLVM headers), and it takes ages due to serially building the .cpp modules out of the box - so since I prefer ninja anyway, this is good enough for me.

rainers:
> letting Visual D compile the .d files as part of the `LDCShared` lib (usually, just the C++ parts), seems to almost work

Nice.

>, the linker is just complaining about conflicting MSCRT versions.

Adding this to the DCompile section in your prop file should work:
```
      <CRuntimeLibrary>None</CRuntimeLibrary>
```



> and it takes ages due to serially building the .cpp modules out of the box

I think cmake doesn't enable parallel compilation by default, adding this to your prop file might help:
```
    <ClCompile>
      <MultiProcessorCompilation>true</MultiProcessorCompilation>
    </ClCompile>
```

kinke:
Thx, `MultiProcessorCompilation` works fine. Building the `ldc2` project, i.e., ldc2.exe, now (rather surprisingly) works out of the box with both `Release` and `RelWithDebInfo` configs - using our prebuilt LLVM (which was built with clang, which probably accounts for the ~17k C++ warnings).

kinke:
Btw, I can't test DMD via the IDE, the custom .props file seems to expose a small issue:

image

When I try to select DMD explicitly and apply, the compiler is still LDC.

rainers:
> When I try to select DMD explicitly and apply, the compiler is still LDC.

I guess that happens because you switched the default compiler in the props file. The position of the import might also be a factor,

kinke:
I've just transplanted that property from .vcxproj to .props, just like the other settings, which can all be overridden from the GUI. It's not even overridable in the .vcxproj. - Anway, no big deal at all.

kinke:
> > Btw Rainer, the back button on my mouse doesn't navigate backward after jumping with F12
> 
> IIRC that functionality is not covered by any command that can be reimplemented, but has to be done by explicit mouse handling hooking into the text editor. Something similar has been added recently to Visual D, so I might be able to revisit that.

@rainers: Can you point me to some commit/PR? It's really a minor feature I can hardly live without... :]

rainers:
I was thinking about one of these mouse handlers:
https://github.com/dlang/visuald/blob/92ec7a24278a2e32583137f250f2ddae4e67c41d/vdextensions/gotodef.cs#L221
https://github.com/dlang/visuald/blob/92ec7a24278a2e32583137f250f2ddae4e67c41d/vdextensions/quickinfo.cs#L65
but maybe they are actually not so useful to implement the back button.

Maybe this extension is helpful: https://marketplace.visualstudio.com/items?itemName=SamHarwell.MouseNavigation

kinke:
Thx; I've had a go, but failed to build the vdextensions project and gave up. I'd be more than happy to test a build - here's the patch:
```diff
--- a/vdextensions/gotodef.cs
+++ b/vdextensions/gotodef.cs
@@ -182,6 +182,7 @@ namespace vdext15
             var buffer = view.TextBuffer;

             IOleCommandTarget shellCommandDispatcher = GetShellCommandDispatcher(view);
+            IVsUIShell vsUIShell = GetVsUIShell();

             if (shellCommandDispatcher == null)
                 return null;
@@ -196,6 +197,7 @@ namespace vdext15

                        return new GoToDefMouseHandler(view,
                                            shellCommandDispatcher,
+                                           vsUIShell,
                                            _aggregatorFactory.GetClassifier(buffer),
                                            _navigatorService.GetTextStructureNavigator(buffer),
                                            CtrlKeyState.GetStateForView(view));
@@ -211,6 +213,14 @@ namespace vdext15
             return _globalServiceProvider.GetService(typeof(SUIHostCommandDispatcher)) as IOleCommandTarget;
         }

+        /// <summary>
+        /// Get the SVsUIShell from the global service provider.
+        /// </summary>
+        private IVsUIShell GetVsUIShell()
+        {
+            return _globalServiceProvider.GetService(typeof(SVsUIShell)) as IVsUIShell;
+        }
+
         #endregion
     }

@@ -225,12 +235,14 @@ namespace vdext15
         private IClassifier _aggregator;
         private ITextStructureNavigator _navigator;
         private IOleCommandTarget _commandTarget;
+        private IVsUIShell _vsUIShell;

-        public GoToDefMouseHandler(IWpfTextView view, IOleCommandTarget commandTarget,
+        public GoToDefMouseHandler(IWpfTextView view, IOleCommandTarget commandTarget, IVsUIShell vsUIShell,
             IClassifier aggregator, ITextStructureNavigator navigator, CtrlKeyState state)
         {
             _view = view;
             _commandTarget = commandTarget;
+            _vsUIShell = vsUIShell;
             _state = state;
             _aggregator = aggregator;
             _navigator = navigator;
@@ -315,6 +327,27 @@ namespace vdext15
             _mouseDownAnchorPoint = null;
         }

+        // Support backward/forward navigation via extended mouse buttons.
+        // Derived from https://github.com/tunnelvisionlabs/MouseNavigation/blob/master/Tvl.VisualStudio.MouseNavigation/MouseNavigationProcessor.cs.
+        public override void PostprocessMouseUp(MouseButtonEventArgs e)
+        {
+            if (_vsUIShell == null)
+                return;
+
+            uint cmdId;
+            if (e.ChangedButton == MouseButton.XButton1)
+                cmdId = (uint)VSConstants.VSStd97CmdID.ShellNavBackward;
+            else if (e.ChangedButton == MouseButton.XButton2)
+                cmdId = (uint)VSConstants.VSStd97CmdID.ShellNavForward;
+            else
+                return;
+
+            object obj = null;
+            _vsUIShell.PostExecCommand(VSConstants.GUID_VSStandardCommandSet97, cmdId,
+                (uint)OLECMDEXECOPT.OLECMDEXECOPT_DONTPROMPTUSER, ref obj);
+
+            e.Handled = true;
+        }

         #endregion
```

rainers:
Not exactly the same build as a release, but you can try the installer from https://ci.appveyor.com/project/rainers/visuald/builds/34611854/artifacts

kinke:
Thx! I've fortunately not yet uninstalled VS 2017 (installer wouldn't detect VS 2019), and the back/forward buttons work like a charm.

rainers:
> installer wouldn't detect VS 2019

That's the Appveyor build only, as it runs in a VS2017 environment and cannot build some components for VS2019.

PR #3495, opened 2020-07-06
Set Module Metadata "RtLibUseGOT" when noplt UDA is used

kubo39:
blocked by https://github.com/ldc-developers/druntime/pull/189

### DIFF
--- a/gen/uda.cpp
+++ b/gen/uda.cpp
@@ -215,6 +215,13 @@ void applyAttrLLVMAttr(StructLiteralExp *sle, llvm::AttrBuilder &attrs) {
     const auto kind = llvm::getAttrKindFromName(key);
     if (kind != llvm::Attribute::None) {
       attrs.addAttribute(kind);
+#if LDC_LLVM_VER >= 500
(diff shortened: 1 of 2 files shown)


kinke:
Is that really desirable/needed? According to https://reviews.llvm.org/D42216, this metadata seems like some kind of workaround for intrinsics. They don't specify what they mean by `RTLib`; I guess it's the LLVM compiler-rt libs, but it's unclear. If that's the case, this doesn't seem suitable for each module containing some `@noplt` for some functions.

kubo39:
Thank you for having a look.

>According to https://reviews.llvm.org/D42216, this metadata seems like some kind of workaround for intrinsics. They don't specify what they mean by RTLib; I guess it's the LLVM compiler-rt libs, but it's unclear. If that's the case, this doesn't seem suitable for each module containing some @noplt for some functions.

Yes, that's true. According to https://reviews.llvm.org/D42217
> This allows processing of calls to intrinsics easily.

It's just for intrinsics, so this patch is not correct. Closing now.

PR #3498, opened 2020-07-07
Don't use excess precision when parsing single- and double-precision literals

kinke:


kinke:
This fixes https://issues.dlang.org/show_bug.cgi?id=20951 for us. I'll upstream it too.

a8nJ:
https://github.com/dlang/dmd/pull/11387
https://github.com/dlang/phobos/pull/7591

https://github.com/ldc-developers/ldc/pull/3498

Issue #3501, opened 2020-07-10
Assertion triggered while compiling vibe-d with master


Geod24:
```console
$ dub test --compiler=$HOME/dlang/ldc-0d09a171/bin/ldc2
Generating test runner configuration 'vibe-d-test-vibe-core' for 'vibe-core' (library).
Running pre-generate commands for vibe-d:tls...
Performing "unittest" build using /Users/geod24/dlang/ldc-0d09a171/bin/ldc2 for x86_64.
taggedalgebraic 0.11.8: target for configuration "library" is up to date.
eventcore 0.8.48: target for configuration "kqueue" is up to date.
stdx-allocator 2.77.5: target for configuration "library" is up to date.
vibe-core 1.8.1: building configuration "kqueue"...
../../../.dub/packages/eventcore-0.8.48/eventcore/source/eventcore/drivers/posix/watchers.d(405,44): Deprecation: precision argument th.msg.length for format specification "%.*s" must be int, not ulong
Assertion failed: (global->isDeclaration() && "Global variable already defined"), function defineGlobal, file /Users/runner/work/1/s/gen/llvmhelpers.cpp, line 1806.
0  ldc2                     0x000000010fb3ca85 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  ldc2                     0x000000010fb3b836 llvm::sys::RunSignalHandlers() + 198
2  ldc2                     0x000000010fb3d09c SignalHandler(int) + 268
3  libsystem_platform.dylib 0x00007fff700275fd _sigtramp + 29
4  libsystem_platform.dylib 0x00000000000078c0 _sigtramp + 18446603338636985056
5  libsystem_c.dylib        0x00007fff6fefd808 abort + 120
6  libsystem_c.dylib        0x00007fff6fefcac6 err + 0
7  ldc2                     0x000000010fe473bc (anonymous namespace)::DeclareOrDefineVisitor::visit(TypeInfoStructDeclaration*) + 4636
8  ldc2                     0x000000010fdad654 CodegenVisitor::visit(TypeInfoDeclaration*) + 116
9  ldc2                     0x000000010fdde8bf codegenModule(IRState*, Module*) + 4255
10 ldc2                     0x000000010fe60831 ldc::CodeGenerator::emit(Module*) + 881
11 ldc2                     0x000000010fe7a0b6 codegenModules(Array<Module*>&) + 822
[...]

PR #3502, opened 2020-07-10
Fix issue #3501: Don't re-define struct TypeInfos

kinke:
Surfacing now as #3486 added a corresponding assertion.

I've tested this with an older vibe-d v0.8.6 lying around on my box.
The assertion was hit for the TypeInfo of a nested struct:

https://github.com/vibe-d/vibe.d/blob/3c7ae13989003b7ebc59de23c5b6a0fb0466f119/http/vibe/http/router.d#L570-L574

I wasn't able to quickly reduce it to a stand-alone testcase and am reluctant to put any more effort into this.

### DIFF
--- a/gen/typinf.cpp
+++ b/gen/typinf.cpp
@@ -507,8 +507,12 @@ class DeclareOrDefineVisitor : public Visitor {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


kinke:
I've just verified that the issue would be fixed by #3491 as well.

PR #3506, opened 2020-07-14
[Trivial] refactor doLTO in writeModule

kubo39:


driver/toobj.cpp
```diff
@@ -368,7 +364,7 @@ void writeModule(llvm::Module *m, const char *filename) {
     auto M = m;
 #endif
 
-    if (opts::isUsingThinLTO()) {
+    if (doLTO) {
```
kinke, line 367: This is the problem; the summary is only to be emitted for ThinLTO.

kubo39: ahh, mistake..

PR #3507, opened 2020-07-14
Revert "Raise min LLVM version to 6.0"

Geod24:
Reverts ldc-developers/ldc#3493

Because Buildkite needs to be updated: https://buildkite.com/dlang/dmd/builds/14258#2579093e-dd2a-478a-8ae0-ef6f30275231
You probably want to do something akin to: https://github.com/dlang/ci/pull/317


kinke:
> You probably want to do

Not really. There's more than enough on my plate already, and I don't want to dive into the Buildkite details. Plus I'm not sure DMD's CI should depend on LDC master; AFAIK, most other projects use a tagged version (and that's been blocking some of my PRs).

PetarKirov:
It's a good practice to always test the bleeding edge, though such builds should be allowed to fail.

kinke:
Stuff to figure out for the upstream CI. AFAICT, testing LDC, i.e., building LDC with the fresh DMD, is mostly useful to catch C++ interop regressions; for that, using LDC master isn't really needed.

Anyway, I'm definitely not accepting the revert as a band-aid for the DMD CI regression.

Geod24:
Then could someone raise a PR to either use a tag or drop LDC? I'd do it if I was not on a phone.

kinke:
Proposed band-aid: https://github.com/dlang/ci/pull/420

dnadlinger:
See https://github.com/dlang/ci/pull/421 – bumping the LLVM version shouldn't be hard, but I don't have the setup to easily test the changes myself right now.

PR #3511, opened 2020-07-21
Fix partial discarding of function hash info for PGO

JohanEngelen:
See Clang PR: https://reviews.llvm.org/D79961

Note that we do not include versioning code for our PGO files: PGO files from older LDC versions are not guarantee to work with newer versions of LDC. The user will see warnings about different function hashes, so he is aware the profile is not applied for those functions. Hash clashes may occur (old version hash corresponding to different function newer version hash), which is unfortunate but I think we do not need to go to the length of solving that as it only concerns optimization, not miscompilation. (it is similar that we don't guarantee object files from older compiler to work with newer compiler due to potential ABI changes which may also be silent)

Thanks @jondegenhardt

### DIFF
--- a/gen/pgo_ASTbased.cpp
+++ b/gen/pgo_ASTbased.cpp
@@ -135,8 +135,11 @@ public:
       return Working;
 
     // Check for remaining work in Working.
-    if (Working)
-      MD5.update(Working);
+    if (Working) {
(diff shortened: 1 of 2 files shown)


JohanEngelen:
Now with testcase fixup commit

Issue #3375, opened 2020-03-19
PGO build failures involving std.path.buildPath


jondegenhardt:
I've been having PGO build failures in an app that creates a fair number of files in a loop. I've narrowed it down to having multiple calls to `std.path.buildPath`. The build failure occurs when building the final app, and is dependent on the number of calls to `buildPath()` during profile data collection.

Failures occur on both macOS and Ubuntu. For the minimal case testing I've been using LDC 1.20.1. In the full app I am building I tried with a number of compiler versions. In that app the problem started occurring with LDC 1.12.0. Earlier LDC versions work fine.

There are two files attached that demonstrate the problem (remove the `.txt` extensions to get the real files):
*  `build_path.d` - The minimal app. It takes two command line parameters, a directory name and the number of times to call `buildPath`.
* `build.sh` - This builds the `build_path` app. It builds an instrumented build and calls it one or more times to create profile data, then builds the final version of the app. It's during this build of the final version of the app that the crash occurs. It doesn't occur with all `buildPath` call counts, so `build.sh` takes arguments that allow experimenting with the number of `build_path` (app) invocations and the number of `std.path.buildPath` function calls made. Modify the `ldc_home` variable near the top of the script to use it.

The main piece of code in `build_path.d` is:
```
    foreach (i; 0 .. numFiles)
    {
        auto filename = buildPath(dir, format("file_%d.txt", i));
        writeln(filename);
    }
```
where `dir` and `numFiles` are taken from command line arguments.

The problem disappears if the `auto filename = ...` line is replace with:
```
    auto filename = format("%s/file_%d.txt", dir, i);
```

The failures are not always the same, and there are some oddities to the pattern (described below). However, the most common failures on macOS and Ubuntu are as follows:

**macOS**
```
/wrk/dlang/ldc2-1.20.1-osx-x86_64/bin/ldc2 -release -O3 -boundscheck=off -singleobj -flto=thin -fprofile-instr-use=app.profdata -defaultlib=phobos2-ldc-lto,druntime-ldc-lto -odobj -ofbin/build_path build_path.d
0  ldc2                     0x000000010a65cc75 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  ldc2                     0x000000010a65bc66 llvm::sys::RunSignalHandlers() + 198
2  ldc2                     0x000000010a65d278 SignalHandler(int) + 264
3  libsystem_platform.dylib 0x000000010cc12b5d _sigtramp + 29
[...]

PR #3512, opened 2020-07-21
Fix PGO instrumentation crash for final switch(enum) statements together with -release

JohanEngelen:
See GH issue #3375 (it fixes the instrumentation part of that bug on my machine, but apparently there are other bugs troubling PGO still).

This change includes checking of whether there is a default statement when using `case <variable>` (LLVM IR switch instruction cannot be used), although that is currently not allowed in the D language. Therefore there is no testcase that covers that.


JohanEngelen:
thanks for the review

PR #3513, opened 2020-07-22
CI: Bump LDC-LLVM to v10.0.1

kinke:


.azure-pipelines/posix.yml
```diff
@@ -18,16 +18,15 @@ steps:
       tar -xf ninja-mac.zip -C ninja
     else
       export DEBIAN_FRONTEND=noninteractive
+      wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+      CLANG_MAJOR=${CLANG_VERSION%%\.*}
+      sudo add-apt-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-$CLANG_MAJOR main"
```
kinke, line 23: I've switched to their official 'nightlies' apt repo because they've somehow moved to Ubuntu 18.04 for their x86_64 package with LLVM 10 (and that doesn't work on our 16.04, I've already tried it). They still use Ubuntu 16.04 for their PPC64le builds, so I guess this inconsistency stems from different people uploading their builds...

kinke:
Apple users: is `-llibxml2.tbd` supposed to ever work? I've heard about .tbd files for the first time a few weeks ago by @jacob-carlborg.

JohanEngelen:
I am still on macOS 10.13.6 (High Sierra), and llvm-config (trunk) --system-libs spits out `-lz -lcurses -lm -lxml2`

jacob-carlborg:
> Apple users: is -llibxml2.tbd supposed to ever work (apparently newly spit out as LLVM linker flag)

I can't tell for sure (not currently at the computer) but I'm pretty sure a regular `-l` flag should be used, i.e. drop the prefix and file extension. The switch to tbd files are supposed to be transparent.

kinke:
Okay thx, that's what I've done, but wasn't sure whether it should be something like `-Wl,libxml2.tbd`.

jacob-carlborg:
> Apple users: is -llibxml2.tbd supposed to ever work

I've double checked now, here's the result:

* `-llibxml2.tbd` does not work
* `-lxml2` does work
* `-Wl,libxml2.tbd` does not work
* The `-Wl` flag + full path to the `.tbd` file works

`-lxml2` is the correct flag.

`llvm-config --system-libs`, using LLVM 10.0.1-rc1, outputs this for me:  `-lz -lcurses -lm -lxml2`.

.azure-pipelines/posix.yml
jacob-carlborg: I've run into that issue several times in the past. I've asked about the issue and the answer was that builds are manually created and uploaded. They also said that https://releases.llvm.org is more or less only the official place to get binaries for Windows. Other platforms are expected to get them from the package manager.

kinke:
Thx again. FWIW, the Azure Mac image (for our LLVM CI) was bumped from 10.14 to 10.15, so that's probably what somehow caused this (I did some grepping in LLVM src but didn't find anything of relevance, so it looks like CMake's `find_package` gets it wrong for unknown reason).

.azure-pipelines/posix.yml
kinke: Yeah that's really not very professional from their side, leading to stuff like https://github.com/ldc-developers/ldc/blob/10cf5f0a46195d4623745ef34326d11b4d717d18/.travis.yml#L53-L57.

jacob-carlborg:
> FWIW, the Azure Mac image (for our LLVM CI) was bumped from 10.14 to 10.15. 

I'm using 10.15. Perhaps CMake was updated in the process and is broken.

PR #3524, opened 2020-07-25
Add missing `UnrolledLoopStatement` to PGO region count calculator. 

JohanEngelen:
Also adds extra precautions to catch this problem earlier in the future.

Fixes issue #3375

### DIFF
--- a/gen/pgo_ASTbased.cpp
+++ b/gen/pgo_ASTbased.cpp
@@ -102,6 +102,7 @@ public:
     ConditionalExpr,
     AndAndExpr,
     OrOrExpr,
+    UnrolledLoopIterationScope,
 
     // Keep this last.  It's for the static assert that follows.
     LastHashType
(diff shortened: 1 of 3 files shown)


tests/PGO/gh3375.d
```diff
@@ -0,0 +1,19 @@
+// REQUIRES: PGO_RT
```
jondegenhardt, line 1: Will this test get run with an LDC compiled with asserts enabled?

gen/pgo_ASTbased.cpp
```diff
@@ -176,11 +176,40 @@ struct MapRegionCounters : public StoppableVisitor {
...
-  void visit(Statement *stmt) override {}
-  void visit(Expression *exp) override {}
-  void visit(Declaration *decl) override {}
-  void visit(Initializer *init) override {}
-  void visit(Dsymbol *dsym) override {}
+  void visit(Statement *stmt) override {
+    // If this assert fails, a new statement type was added to the frontend, and
+    // thus we need to decide on how to handle PGO calculations for that, both
+    // in MapRegionCounters and in ComputeRegionCounts.
+    assert(0 && "All statement types should be explicitly handled to avoid "
```
jondegenhardt, line 183: Does it make sense to do something in addition or other than `assert` so that detection is explicit in release mode builds that end users will use?

JohanEngelen:
Although it fixes the crash, it is not a correct fix and will result in wrong execution count numbers. Working on better fix.

tests/PGO/gh3375.d
JohanEngelen: @kinke do we have a CI system that runs with asserts enabled? I thought that the builds that output our latest-ci binaries do have asserts enabled, right?

gen/pgo_ASTbased.cpp
JohanEngelen: I'm not sure. The error may not be user-actionable, and things may also just work without change needed. Would you suggest a warning saying that "Statement on line x is not supported by PGO" ?

JohanEngelen: (note that this is the same for `CaseRangeStatement`)

jondegenhardt: > Would you suggest a warning saying that "Statement on line x is not supported by PGO" ?
Something like that. Perhaps requesting that an issue be filed for the purpose of improving LDC. This might make more sense if similar things are already being done elsewhere in the code base.

This rationale is that it might be difficult to catch these cases in the context of LDC development, and instead they'd be more likely to occur when running against user code. Do you think this is the case? If not, then such a message is not needed.

The other part is that PGO computation should proceed reasonably even if the assert condition arises. Ie. Not crash. I didn't try to review the code from this angle.

jondegenhardt:
I tested the PR on all the different reduced cases of bug #3375  I put together, the tool in `tsv-utils` where the problem first arose, and the full set of `tsv-utils` tools. Everything works fine.

gen/pgo_ASTbased.cpp
JohanEngelen: Mind if I disregard this? We have many asserts in the codebase where we don't really report it to the user. (I think it is not very valuable for the user, to be honest)  I do agree that the assert is not very likely to catch things during development, but it might when people use the beta build or latest-ci to have asserts enabled.

jondegenhardt: > Mind if I disregard this?

I don't mind. It was a suggestion for your consideration, nothing else. And it makes complete sense to be consistent with similar cases in the LDC code base.

tests/PGO/gh3375.d
kinke: Yeah, Azure (excl. Android) and Shippable feature enabled assertions.

JohanEngelen:
I guess the Azure failure is a random error? Only `debuginfo/scopes_cdb.d` Lit test fails, but this PGO code change has no impact on that test at all.

PR #3525, opened 2020-07-26
Fix ThreadSanitizer support by removing detaching of main thread upon program termination.

JohanEngelen:
let's see what CI thinks of this... :)


kinke:
Thx!

PR #3546, opened 2020-08-23
Add support for LLVM 11

kinke:
Quite a PITA this time.


driver/cl_options-llvm.cpp
```diff
@@ -57,18 +83,39 @@ cl::boolOrDefault disableFPElim() {
...
+#endif
   if (MCPU == "help")
     return true;
   return std::any_of(MAttrs.begin(), MAttrs.end(),
                      [](const std::string &a) { return a == "help"; });
 }
 
 TargetOptions InitTargetOptionsFromCodeGenFlags() {
+#if LDC_LLVM_VER >= 1100
+  return codegen::InitTargetOptionsFromCodeGenFlags();
```
JohanEngelen, line 98: you already did all the work, but maybe a `using codegen` would've worked well too.

JohanEngelen:
oof indeed. Nice work

kinke:
Should be finally almost ready, just a couple of test regressions wrt. optimized code (`std.bitmanip` for Mac & Linux x64, 2 other failures for Win32) + failing `instrument/xray_link.d` for Mac. I guess the necessary adaptations would have been even more painful when still supporting LLVM 3.9...

JohanEngelen:
I'm working on the macOS Xray failure right now

kinke:
I've had a look at the `std.bitmanip` issue. It looks pretty insane:
```D
import std.bitmanip : bitsSet;
import std.algorithm.comparison : equal;

void main()
{
    auto a = bitsSet(3);
    assert(a.length == 2);
    int[2] expected = [0, 1];
    assert(a.equal(expected[]));
}
```

```
$ ldc2 -O -run bla.d
core.exception.AssertError@/home/martin/ldc/runtime/phobos/std/range/primitives.d(2432): Attempting to fetch the front of an empty array of int
```

It works fine when making the `expected` array `static immutable`.

The optimized .ll is incredible - no traces of the `bitsSet` left:
```ll
assertPassed:
  ; 1) allocate and store `expected` array
  %expected = alloca [2 x i32], align 4           ; [#uses = 3, size/byte = 8]
  %.fca.0.gep = getelementptr inbounds [2 x i32], [2 x i32]* %expected, i64 0, i64 0 ; [#uses = 2, type = i32*]
  store i32 0, i32* %.fca.0.gep, align 4
  %.fca.1.gep = getelementptr inbounds [2 x i32], [2 x i32]* %expected, i64 0, i64 1 ; [#uses = 1, type = i32*]
  store i32 1, i32* %.fca.1.gep, align 4
  ; 2) load the 1st elem and assert it's 0 - it's just been stored!
  %0 = load i32, i32* %.fca.0.gep, align 4        ; [#uses = 1]
  %1 = icmp eq i32 %0, 0                          ; [#uses = 1]
  br i1 %1, label %assertPassed.i.i, label %assertFailed2

assertPassed.i.i:                                 ; preds = %assertPassed
  ; 3) load the 2nd elem and assert it's 1, as if it hadn't just been stored...
  %lowerbound.i.i = getelementptr inbounds [2 x i32], [2 x i32]* %expected, i64 0, i64 1 ; [#uses = 1, type = i32*]
  %2 = load i32, i32* %lowerbound.i.i, align 4    ; [#uses = 1]
  %3 = icmp eq i32 %2, 1                          ; [#uses = 1]
  br i1 %3, label %assertPassed.i.i.1, label %assertFailed2

assertPassed1:                                    ; preds = %assertPassed.i.i.1
  ret i32 0

assertFailed2:                                    ; preds = %assertPassed.i.i, %assertPassed
  tail call void @_d_assert({ i64, i8* } { i64 12, i8* getelementptr inbounds ([13 x i8], [13 x i8]* @.str, i32 0, i32 0) }, i32 9) #2
  unreachable

assertPassed.i.i.1:                               ; preds = %assertPassed.i.i
  ; 4) FUN! *undefined* branch, apparently jumping to forbody.i.2 which throws the assertion error...
  br i1 undef, label %assertPassed1, label %forbody.i.2

forbody.i.2:                                      ; preds = %assertPassed.i.i.1
  tail call void @_d_assert_msg({ i64, i8* } { i64 54, i8* getelementptr inbounds ([55 x i8], [55 x i8]* @.str.4, i32 0, i32 0) }, { i64, i8* } { i64 54, i8* getelementptr inbounds ([55 x i8], [55 x i8]* @.str.5, i32 0, i32 0) }, i32 2432) #2
  unreachable
}
```

kinke:
> It works fine when making the `expected` array `static immutable`.

Yeah well, just by accident:
```ll
forbody.preheader.i:
  br i1 undef, label %assertPassed1, label %forbody.i.2

assertPassed1:                                    ; preds = %forbody.preheader.i
  ret i32 0

forbody.i.2:                                      ; preds = %forbody.preheader.i
  tail call void @_d_assert_msg({ i64, i8* } { i64 54, i8* getelementptr inbounds ([55 x i8], [55 x i8]* @.str.4, i32 0, i32 0) }, { i64, i8* } { i64 54, i8* getelementptr inbounds ([55 x i8], [55 x i8]* @.str.5, i32 0, i32 0) }, i32 2432) #4
  unreachable
}
```

kinke:
With `-mtriple=x86_64-pc-windows-msvc`, the `static immutable` thing is optimized to the expected `return 0`, while the stack variant contains the same poor optimization checking the previous stores, but no undefined branches. So IMO, this has got to be related to EH.

kinke:
The optimized 32-bit Linux x86 .ll is ugly as hell (incl. the `static immutable` version), but no undefined branches either. EH-wise, I don't think there should be any relevant divergences...

kinke:
Filed as https://bugs.llvm.org/show_bug.cgi?id=47322.

kinke:
They are working on it; it's actually an older bug (at least also happening with LLVM 10) which just happened to surface with LLVM 11 for our testcase. The other regression for Win32 is apparently just some custom NaN not being propagated and can probably be neglected, so yeah, almost ready.

kinke:
I'm trying to move the AArch64 package generation from Shippable to Travis, as Shippable is constantly timing out again. I need someone to enable the llvm-project repo on Travis though (I've tried setting up a webhook alone, but that doesn't suffice). Maintaining CI would definitely be a lot easier if I could manage the GitHub apps for the ldc-developers organization, but I've already requested those rights multiple times in the last year without a single reply - trying again @dnadlinger @redstar.

kinke:
Thx to whoever did what they did, it's working now. - Looks like it's too slow unfortunately, plus the SPIR-V translator doesn't compile with gcc 5.4 (and MLIR neither, but that was already the case with IIRC gcc 8 for Shippable, apparently some LLVM 11 regression).

kinke:
There's just no way Travis can build LLVM in under 50 mins (I've tried 32-48-64 threads, ld.bfd and ld.gold, excluding the LLVM tools...). According to https://blog.travis-ci.com/2020-09-11-arm-on-aws, migrating to travis-ci.**com** would allow to experiment with a supposedly much faster AWS Graviton2 CPU. Looks like the domain transition wouldn't come with any real drawbacks, see https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/.

kinke:
That CPU is apparently totally different from the Shippable one - no high-latency-high-throughput monster like Shippable with huge memory requirements to feed the many cores, but showing a similar performance to x64 server cores, and apparently being restricted for our purposes to something like 4 gigs of memory. With ~8 threads (retrying after OOM kills), the 50 minutes seem just enough, see https://travis-ci.com/github/kinke/llvm-project/builds/186751206.

kinke:
> Looks like the domain transition wouldn't come with any real drawbacks

Well, encrypting stuff with the (annoyingly, Ruby-based) `travis` CLI tool needs an extra step, authenticating with a GitHub token, see https://docs.travis-ci.com/user/encryption-keys#usage.

---

There's an annoying AArch64 regression with LLVM 11 wrt. the `llvm_umul...` intrinsic (apparently wrongly reporting overflows), only detected yesterday with the first successful LLVM 11 build.

I'm newly including the `AMDGPU` backend for the x86 builds; we've included the nvidia one for a while, and excluding AMD seems not really fair, plus we have gccbuiltins support for AMD GCN now too...

kinke:
AArch64 regression filed as https://bugs.llvm.org/show_bug.cgi?id=47679.

kinke:
Finally - merging this with non-final LLVM 11 for the prebuilt packages, as the final will most likely not be ready for the beta.

jacob-carlborg:
> There's just no way Travis can build LLVM in under 50 mins

@kinke FYI, GitHub actions are significantly faster than Travis, in my experience.

jacob-carlborg:
BTW, all tests pass for me on iOS using LLVM 11.

kinke:
> FYI, GitHub actions are significantly faster than Travis, in my experience.

But don't support arm64 AFAIK.

jacob-carlborg:
> But don't support arm64 AFAIK

Oh, I didn't know Travis support non-x86 architectures. That explains it.

jacob-carlborg:
> But don't support arm64 AFAIK

There's this action: https://github.com/marketplace/actions/run-on-architecture, but that's using emulation.

kinke:
That's going to be waaay too slow. - Travis.**com** would be fast enough for LLVM too as stated above (travis.**org** already is fast enough for LDC itself, taking less than 30 mins).

---

Thx for checking LLVM 11; I'll use it for Bitrise too once the final is out.

kinke:
Yay, Bitrise is indeed green again for master builds (where the unittests are actually run, as opposed to PRs where they are only compiled).

jacob-carlborg:
Awesome, let's keep it that way 😀.

Issue #3548, opened 2020-08-30
Duplicate symbol link error with `__initZ` of `pragma(inline,true)` function inner struct


JohanEngelen:
The link error happens only with separate compilation.

Link error (errors at least on macOS and on Linux):
```
duplicate symbol __D1a1S11innerStructMFZ11InnerStruct6__initZ in:
    exe.o
    a.o
```

Testcase:
```d
// File a.d
module a;

struct S {
    auto innerStruct() {
        pragma(inline, true); // <-- this triggers the error
        static struct InnerStruct {
            int nonZero = 1;
        }
        return InnerStruct();
    }
}

/+ //Not needed but simply making sure we generate code for the struct.
auto foo() {
    S s;
    return s.innerStruct();
}
+/
```

```d
// File b.d
module b;

import a;

void main() {
    S s;
    s.innerStruct();
}
```

Reproduce with:
```
ldc2 a.d -c -of=a.o
ldc2 b.d a.o -of=exe
```

JohanEngelen:
(working on a fix btw. At least we know inlining is happening now ;-))

JohanEngelen:
Does not only affect structs, but all aggregates: structs, classes, and interfaces.

JohanEngelen:
Workaround is templetizing the to-be-inlined function:
```diff
- auto innerStruct() {
+ auto innerStruct()() {
```

JohanEngelen:
See #3549 for some implementation discussion.

Implementation is not trivial, and I think forcing the users to template the function or inner aggregate is good enough.

PR #3549, opened 2020-08-30
Don't define __initZ and friends, but instead only declare, for nonroot structs, classes, and interfaces (happens when force-inlining a function and that function contains internal aggregates)

JohanEngelen:
Fixes issue #3548

I'm not super happy with this fix, but it works. If you agree, I will add more commentary to explain the if() statements. 

Note that this does not prevent inlining/optimization of the init values themselves: when only compiling module b (the one that imports the struct), then functions are still optimized to
```llvm
define %gh3548a.S.innerInnerStruct.InnerStruct.InnerInnerStruct @_D1b3fooFNaNbNiNfZS7gh3548a1S16innerInnerStructMFZ11InnerStruct16InnerInnerStruct() local_unnamed_addr #0 {
  ret %gh3548a.S.innerInnerStruct.InnerStruct.InnerInnerStruct { i32 1 }
}
```
so no referencing to some opaque `__initZ` variable.

I did not check for vtables. It will probably hurt devirtualization, because the vtbl is unknown. But I think that should be part of a different PR, probably we then always want to emit vtbls as linkonce_odr.

### DIFF
--- a/gen/declarations.cpp
+++ b/gen/declarations.cpp
@@ -102,10 +102,12 @@ public:
       }
 
       // Emit TypeInfo.
-      IrClass *ir = getIrAggr(decl);
-      if (!ir->suppressTypeInfo() && !isSpeculativeType(decl->type)) {
-        llvm::GlobalVariable *interfaceZ = ir->getClassInfoSymbol();
-        defineGlobal(interfaceZ, ir->getClassInfoInit(), decl);
+      if (!decl->inNonRoot()) {
+        IrClass *ir = getIrAggr(decl);
+        if (!ir->suppressTypeInfo() && !isSpeculativeType(decl->type)) {
+          llvm::GlobalVariable *interfaceZ = ir->getClassInfoSymbol();
+          defineGlobal(interfaceZ, ir->getClassInfoInit(), decl);
+        }
       }
     }
   }
@@ -144,7 +146,10 @@ public:
       IrStruct *ir = getIrAggr(decl);
 
       // Define the __initZ symbol.
-      if (!decl->zeroInit) {
+      // Don't define it if it is all-zeros, or if the struct is from another
+      // module (codegen reached from a force inlined function from another
+      // module).
+      if (!decl->zeroInit && !decl->inNonRoot()) {
         auto &initZ = ir->getInitSymbol();
         auto initGlobal = llvm::cast<LLGlobalVariable>(initZ);
         initZ = irs->setGlobalVarInitializer(initGlobal, ir->getDefaultInit());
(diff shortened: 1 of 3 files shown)


kinke:
A probably pretty simple fix would be using `weak_odr` for aggregates inside templates. Declaring in non-root modules is probably not enough when compiling multiple object files at once.

JohanEngelen:
> Declaring if in non-root module is probably not enough when compiling multiple object files at once.

What case do you mean exactly?

Emitting always as linkonce_odr also works (it's what I tried to first to make sure it would fix the bug in the original code), but I am not sure if that may merge symbols with the same name that shouldn't be merged. Can we have symbols that have the same name but must stay private to compilation modules and should not be merged?

kinke:
> What case do you mean exactly?

E.g., try compiling your 2 test files at once via `-c` and then try to link the 2 objects. The struct is then definitely in a root module and should be defined in both CUs.

`linkonce_odr` shouldn't work in case the init symbol is unused in the CU where it is defined, not making it to the assembling stage, but required in another CU where it's just declared.

JohanEngelen:
> E.g., try compiling your 2 test files at once via `-c` and then try to link the 2 objects. The struct is then definitely in a root module and should be defined in both CUs.

Yep indeed. This fails:
```
// RUN: %ldc -I%S/inputs %S/inputs/gh3548a.d %s -c -od=%T
// RUN: %ldc %T/gh3548.o %T/gh3548a.o -of=gh3548
```

kinke:
Have you checked whether this generally works with DMD? [Well, I wouldn't be too surprised if it unknowingly did... :)]

Just not sure whether supporting such constructs is worth it - the function body itself is only `available_externally`, but nested aggregates are fully emitted, and e.g. nested classes probably even make it to the list of classes in the ModuleInfo. I assume this comes from Weka; would dummy-templatizing these functions be a feasible workaround?

JohanEngelen:
I think we should support it; it doesn't seem to be too much effort. Would `weak_odr` have any downside over what we currently already do? 

Indeed it's at Weka. In one case I can dummy-template. But in the other case, I cannot (or there would be breakage every now and then), because a general fix is needed (it's caused by auto-generated function enty/exit code).

JohanEngelen:
We could apply weak_odr only to structs that are inside pragma(inline) functions, but I first thought that'd be too much checking to do, and instead just always emit them as weak_odr.

JohanEngelen:
> Have you checked whether this generally works with DMD? [Well, I wouldn't be too surprised if it unknowingly did... :)]

DMD works. (but I think dmd just emits everything as weak_odr)

kinke:
If we really want to support this, it would probably mean that we have to treat these nested aggregates exactly as if they were inside a template - emitting everything as weak_odr, not just init symbol, TypeInfo, vtables, but also its member functions and further nested aggregates. Doing so for *all* non-templated aggregates sounds way too dangerous to me, just to account for this exotic special case. Selectively for those inside force-inlined function parents is certainly doable, but I'm still not convinced this is worth it over just requiring dummy-templatization...

kinke:
Another option would be *not* codegen'ing them at all, in case the parent function is emitted in a non-owning module (available_externally). But that probably undermines the purpose of force-inlining then...

JohanEngelen:
I'm closing this, but let's keep the issue open.

PR #3568, opened 2020-09-27
Refactoring: Introduce getIrType()

kinke:
See commit messages for details.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -761,9 +761,8 @@ DISubroutineType DIBuilder::CreateFunctionType(Type *type) {
 
   // The calling convention has to be recorded to distinguish
   // extern(D) functions from extern(C++) ones.
-  DtoType(t);
-  assert(t->ctype);
-  unsigned CC = t->ctype->getIrFuncTy().reverseParams ? DW_CC_D_dmd : 0;
+  unsigned CC =
+      getIrType(t, true)->getIrFuncTy().reverseParams ? DW_CC_D_dmd : 0;
 
   return DBuilder.createSubroutineType(paramsArray, DIFlags::FlagZero, CC);
 }
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -1806,9 +1806,9 @@ LLValue *DtoIndexAggregate(LLValue *src, AggregateDeclaration *ad,
   // Look up field to index and any offset to apply.
   unsigned fieldIndex;
   unsigned byteOffset;
-  assert(ad->type->ctype->isAggr());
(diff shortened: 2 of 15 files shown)


kinke:
There's probably a real frontend bug somewhere. With the 3rd commit and the printf uncommented, there's a single aggregate across druntime/Phobos (incl. unittests) showing this 'multiple Types' syndrome:
```
Aggregate with multiple Types detected: std.demangle.__unittest_L50_C9.main.__lambda1!string.__lambda1.ChunkByImpl!(__lambda2, string).ChunkByImpl.Impl
```

kinke:
@rainers: Don't you find that prettyName a bit odd? I'd expect something like
```
std.algorithm.iteration.ChunkByImpl!(std.demangle.__unittest_L50_C9.main.__lambda1!string.__lambda1.__lambda2!(immutable(char)).__lambda2, string).ChunkByImpl.Impl
```

rainers:
> @rainers: Don't you find that prettyName a bit odd? I'd expect something like

I think you are right. The struct ChunkByImpl should not get the instantiating scope in the fully qualified name. Is this the result of ambiguous mangling?

kinke:
> Is this the result of ambiguous mangling?

In case you mean this 'multiple Types'-thing: just a very wild guess. ;) I was more wondering in general and found it confusing, and if the user sees something like this in error msges etc., (s)he'll probably be confused too.

rainers:
IIUC what happens is that for some TypeStruct ts, ts.sym.type != type. I don't think this should happen. I suspect the types have not been merged for some reason. Have you compared the deco of the types?

Regarding the prettyChars: yeah, probably some issue in dmd.hdrgen.

kinke:
The decos minimally diverge:
```
// ts->sym->type->deco: S3std9algorithm9iteration__T11ChunkByImplSQBp8demangle17__unittest_L50_C9FZ4mainMFZ__T9__lambda1TAyaZQpMFQiZ9__lambda2TQwZQDq4Impl
// ts->type->deco:
S3std9algorithm9iteration__T11ChunkByImplSQBp8demangle17__unittest_L50_C9FZ4mainMFZ__T9__lambda1TAyaZQpFQhZ9__lambda2TQvZQDp4Impl
```

rainers:
> The decos minimally diverge:

Both demangle to `std.algorithm.iteration.ChunkByImpl!(std.demangle.__unittest_L50_C9().main().__lambda1!(immutable(char)[]).__lambda1(immutable(char)[]).__lambda2, immutable(char)[]).ChunkByImpl.Impl`, the only difference is that there is an additional `M` in there which means that `__lambda1` is a delegate in one type, but a regular function in the other. I suspect that this changes late during semantic analysis.

PR #3522, opened 2020-07-24
Copy LLVM's ThreadSanitizer (TSan) library to LDC's libraries like we do for ASan, and set the correct linker flags

JohanEngelen:
TSan is not supported on Windows.

See https://github.com/ldc-developers/ldc/issues/3519

_Execution is still failing for me on macOS, let's see how our CI systems fare. With LLVM 8 I get:_
```
❯ lldb /Users/johan/ldc/johan/build80/tests/sanitizers/Output/fsanitize_thread.d.tmp
(lldb) target create "/Users/johan/ldc/johan/build80/tests/sanitizers/Output/fsanitize_thread.d.tmp"
Current executable set to '/Users/johan/ldc/johan/build80/tests/sanitizers/Output/fsanitize_thread.d.tmp' (x86_64).
(lldb) r
Process 10307 launched: '/Users/johan/ldc/johan/build80/tests/sanitizers/Output/fsanitize_thread.d.tmp' (x86_64)
FATAL: ThreadSanitizer CHECK failed: /Users/johan/llvm/llvm/compiler-rt/lib/tsan/rtl/tsan_rtl_thread.cc:311 "((tid)) < ((kMaxTid))" (0xffffffffffffffff, 0x1fc0)
    #0 __tsan::TsanCheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) <null> (libldc_rt.tsan.dylib:x86_64h+0x62ac9)

Process 10307 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff72962b66 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff72962b66 <+10>: jae    0x7fff72962b70            ; <+20>
    0x7fff72962b68 <+12>: movq   %rax, %rdi
    0x7fff72962b6b <+15>: jmp    0x7fff72959ae5            ; cerror_nocancel
    0x7fff72962b70 <+20>: retq   
Target 0: (fsanitize_thread.d.tmp) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff72962b66 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff72b2d080 libsystem_pthread.dylib`pthread_kill + 333
[...]

PR #3572, opened 2020-10-01
Fix ThreadSanitizer support by removing detaching of main thread upon program termination

kinke:
A reboot of #3525 after #3522.


kinke:
I guess this doesn't work with *shared* druntime...

JohanEngelen:
lgtm

JohanEngelen (approved):

PR #3575, opened 2020-10-02
CMake: Tweak build order for D executables

kinke:
It's been bothering me for a while that building `ldc2.exe` makes poor use of parallelization, because the C++ parts are compiled first, and compiling the D frontend to a single object file, the step which takes *way* longer than any C++ file, gets scheduled very late.

I've just stumbled upon this in the CMake [docs](https://cmake.org/cmake/help/latest/command/add_custom_command.html) for the custom-command dependencies:

> If any dependency is an OUTPUT of another custom command in the same directory (CMakeLists.txt file), CMake automatically brings the other custom command into the target in which this command is built.

And indeed, using an extra dummy custom-target inbetween makes ninja schedule the D object compilation very early in the process. On my box with `ninja -j4 ldc2`, the wall-time is reduced from ~196 to ~133 secs!


JohanEngelen:
Nice, that's a good find.
Perhaps add a comment about that in the cmake file so that the "unnecessary" in-between target is not removed.

PR #3578, opened 2020-10-04
Implement ABI stuff for -preview=in

kinke:
I.e., the ABI-specific decision for which parameter types to prefer passing a ref over a value.

The x86_64 and AArch64 ABIs are pretty clear - use a ref if the POD type cannot be passed in registers. For all others, use a simple default heuristic based on the type size - if larger than 2 machine words, use a ref. Note that this includes x87 `real` for the 32-bit x86 ABI.

### DIFF
--- a/dmd/target.d
+++ b/dmd/target.d
@@ -661,6 +661,8 @@ version (IN_LLVM)
     TypeTuple toArgTypes(Type t);
     bool isReturnOnStack(TypeFunction tf, bool needsThis);
     // unused: ulong parameterSize(const ref Loc loc, Type t);
+    bool preferPassByRef(Type t);
+    Expression getTargetInfo(const(char)* name, const ref Loc loc);
 }
 else // !IN_LLVM
 {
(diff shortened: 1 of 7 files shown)


JohanEngelen:
I am very worried about the semantic meaning of `-preview=in` being unclear. If this PR only implements the ABI part, and does not change the semantic meaning of code, it's all good with me; but I'm worried the frontend implementation is incorrect and does not provide the correct guarantees when passing by ref (i.e. creating a temporary local copy if parameters alias).

My worry is that with this PR, we combine the decision that we _must_ pass by reference to not violate D language semantics, with the decision that we _may_ pass by ref as an optimization (without changing language semantics).

kinke:
Yeah, this just implements the ABI-specific part for LDC (well, at least the 3 major ABIs, but the 32-bit x86 ABI is IMO okay too this way). The upstream ABI part for DMD needs work - I've promised to implement it, so the frontend changes in here will probably be upstreamed more or less as-is.

kinke:
> i.e. creating a temporary local copy if parameters alias

This is hard to prove (could be alias via global state as well), and explicit `in` + `-preview=in` shifts the burden to the user.

thewilsonator:
> The upstream ABI part for DMD needs work - I've promised to implement it, so the frontend changes in here will probably be upstreamed more or less as-is.

Ping me when you make that PR.

PR #3580, opened 2020-10-09
Add IR alignment attribute for ref/out and pointer parameters

### LINKED ISSUE #3562: Assume alignment when loading from pointer
As far as I know, unaligned memory access is UB in D. We don't yet make use of this in all cases.
```d
align(64)
struct int64 {
    int i;
    alias i this;
}

int64* g();

int foo() {
    auto t = g();
    return *t;
}

int f(int64* i) { return *i;}
```
creates the IR:
```llvm
define i32 @_D7example3fooFZi() local_unnamed_addr #1 {
  %1 = tail call %example.int64* @_D7example1gFZPSQo5int64() #0
  %2 = getelementptr inbounds %example.int64, %example.int64* %1, i64 0, i32 0
  %3 = load i32, i32* %2, align 4
  ret i32 %3
}

define i32 @_D7example1fFPSQn5int64Zi(%example.int64* nocapture readonly %i_arg) local_unnamed_addr #2 {
  %1 = getelementptr inbounds %example.int64, %example.int64* %i_arg, i64 0, i32 0
  %2 = load i32, i32* %1, align 4
  ret i32 %2
}
```

Note that `align 4` instead of `align 64`. (The `align 4` is actually deduced by LLVM (loading an `i32`), and not emitted by LDC.)

I think we can be aggressive here and emit `align 64`.
Clang does that too.

kinke:
Resolves #3562.


thewilsonator (approved):

JohanEngelen:
> I thought about it, but class alignments can be violated by the GC, so that's definitely dangerous.

Ah I remember something about that... It's pretty bad actually. Do we have these violations when newing a struct?

(hey! how do you get a reply inline indented like that?)

kinke:
> It's pretty bad actually. Do we have these violations when newing a struct?

Yes - the GC doesn't guarantee any alignment > 16 IIRC.

> how do you get a reply inline indented like that?

Heh, I think GitHub is just playing around - your comment doesn't show up in the diff anymore, and the inline diff lines are missing here in the 'Conversation' tab as well...

JohanEngelen:
Considering the "fix sentinel pointer commit. So a side effect of this PR is that invalid pointer values are also assumed to be aligned... hmm. I'm not sure whether that is correct. If a pointer parameter is known to be `dereferencable`, in that case we can assume alignment, otherwise not.
Maybe better to add alignment when loading from a pointer, and not add alignment info on function parameters (unless they are `dereferencable`) ?

kinke:
> So a side effect of this PR is that invalid pointer values are also assumed to be aligned

IIRC, the pointers are actually valid, at least the `object.rtinfo...` ones - they are `void*`. They become invalid later when casting to `size_t*` or `TypeInfo` etc. - I don't think these potentially rather common magic values like 1 and -1 need a special case (but definitely a changelog entry); they can still be checked without altering the values when typed as `void*` or `size_t`.

JohanEngelen:
> **Edit2**: Invalid constant values could be detected and rejected by the compiler, e.g., `cast(size_t*) 1`.

Yeah I think this is quite a common use case. Because pointer value 1 is more obviously a wrong value.
I'll see if I can try this on Weka's codebase + runtime tests, see if it breaks something.

JohanEngelen:
> Maybe better to add alignment when loading from a pointer, and not add alignment info on function parameters (unless they are `dereferencable`) ?

I still think that theoretically it is the better solution to only assume alignment when the pointer value is used to read from memory (so add align(x) on `load` only, not on func params). People may also do extreme data packing in pointers (using the last bit of a pointer to store some boolean for example), that may break with putting alignment assumptions on func params (or any pointer value read from memory).

kinke:
Restricting it to loads would surely be the safe variant, but the alignment would be lost when forwarding the pointer, e.g., for
`void clear(int4* p) { memset(p, 0, 16); }`...

JohanEngelen:
> Restricting it to loads (**edit**: and stores) would surely be the safe variant, but the alignment would be lost when forwarding the pointer, e.g., for `void clear(int4* p) { memset(p, 0, 16); }`...

Yep, too bad. We have to be correct before we can optimize for performance.
Look at this code:
```d
bool foo(int* i) {
    return cast(size_t)i & 1;
}
```
In LLVM IR:
```llvm
define zeroext i1 @aligned(i32* %i_arg) {
  %1 = ptrtoint i32* %i_arg to i64
  %2 = and i64 %1, 1              
  %3 = icmp ne i64 %2, 0
  ret i1 %3
}
```
When alignment is added to the parameter, the IR is optimized to `ret i1 false` (as it should, ofcourse).
https://llvm.godbolt.org/z/zKcz4r
That means that it is impossible to write an `bool is_aligned(int*)` function or do a boolean inside pointer packing. Very unexpected, I think, and non-compliant to D language spec.

dnadlinger:
> That means that it is impossible to write an bool is_aligned(int*) function or do a boolean inside pointer packing.

You can still write `bool is_aligned(T)(void*)`, and a wrapper struct that packs the boolean inside a `void*` together with a `T*` pointer.

> Very unexpected, I think, and non-compliant to D language spec.

I guess that's quite subjective. If the language disallows creating unaligned pointers, I'd expect it to behave like that. IIRC, the D spec isn't very explicit about this (but imho should), so erring on the side of caution is probably the way to go.

JohanEngelen:
> > That means that it is impossible to write an bool is_aligned(int*) function or do a boolean inside pointer packing.
> 
> You can still write `bool is_aligned(T)(void*)`, and a wrapper struct that packs the boolean inside a `void*` together with a `T*` pointer.

Note that if one would receive a typed pointer as function parameter, it would already be tagged with alignment assumption, so there is no point trying to write a templated is_aligned function: the alignment would be forced and by inlining the is_aligned function the optimizer will have a go at it :)

> > Very unexpected, I think, and non-compliant to D language spec.
> 
> I guess that's quite subjective. If the language disallows creating unaligned pointers, I'd expect it to behave like that. IIRC, the D spec isn't very explicit about this (but imho should), so erring on the side of caution is probably the way to go.

It's easy to create an unaligned pointer by taking the address of packed align(1) struct members. (and maybe using that pointer is UB then...?)
Hmm, so even the safe way of only adding alignment assumptions on load/stores, it should be disabled for load/stores to align(1) struct members. (if the alignment of the specific variable is applied to IR, instead of the alignment of typeof(variable), things will work fine I think).

kinke:
> Hmm, so even the safe way of only adding alignment assumptions on load/stores, it should be disabled for load/stores to align(1) struct members.

On the contrary, those invalidly-aligned members would need extra alignment annotations if we really wanted to make them work. `DtoAlignment()` doesn't use the field alignments IIRC.

A memcpy is probably the usual way to handle alignment violations, such as in https://github.com/dlang/druntime/blob/afd61bceb25e6c76b4183cb763e0d75b44a169a1/src/rt/backtrace/dwarf.d#L483.

dnadlinger:
> Note that if one would receive a typed pointer as function parameter, it would already be tagged with alignment assumption, so there is no point trying to write a templated is_aligned function: the alignment would be forced and by inlining the is_aligned function the optimizer will have a go at it :)

You'll note that my example had `void*` as the argument type.

JohanEngelen:
> > Note that if one would receive a typed pointer as function parameter, it would already be tagged with alignment assumption, so there is no point trying to write a templated is_aligned function: the alignment would be forced and by inlining the is_aligned function the optimizer will have a go at it :)
> 
> You'll note that my example had `void*` as the argument type.

Yes, the `is_aligned` function. I meant the case where the caller of `is_aligned` will not have `void*` as parameter.

kinke:
> I meant the case where the caller of `is_aligned` will not have `void*` as parameter.

Yeah, that's the problem when combined with inlining. I've tried an extra intermediate function for `ldc.simd.storeUnaligned()` taking a `void*`, but the `align 1` is still overridden, and the `impl` function would need an (obviously unacceptable) `pragma(inline, false)` to prevent that:
```D
template storeUnaligned(V)
if(is(typeof(llvmVecType!V)))
{
    pragma(inline, true):

    private void impl(V value, void* p)
    {
        enum llvmV = llvmVecType!V;
        enum ir = `
            %p = bitcast i8* %1 to `~llvmV~`*
            store `~llvmV~` %0, `~llvmV~`* %p, align 1`;
        __ir_pure!(ir, void)(value, p);
    }

    void storeUnaligned(V value, BaseType!V* p)
    {
        impl(value, p);
    }

    /// Deprecated: This is the DMD interface, use it only for DMD compatibility. Otherwise, please use the LDC interface.
    V storeUnaligned(V* p, V value) // <--- invalid p
    {
        impl(value, p);
        return value;
    }
}
```

=>

```LL
; Function Attrs: alwaysinline uwtable
define weak_odr <2 x i64> @_D3ldc4simd__T14storeUnalignedTNhG2lZQxFNaNbNiNePQsQuZQx(<2 x i64> %value_arg, <2 x i64>* align 16 %p_arg) local_unnamed_addr #4 {
  store <2 x i64> %value_arg, <2 x i64>* %p_arg, align 16
  ret <2 x i64> %value_arg
}

; Function Attrs: alwaysinline uwtable
define weak_odr void @_D3ldc4simd__T14storeUnalignedTNhG2lZ4implFNaNbNiNeQuPvZv(i8* align 1 %p_arg, <2 x i64> %value_arg) local_unnamed_addr #4 {
  %p.i = bitcast i8* %p_arg to <2 x i64>*         ; [#uses = 1]
  store <2 x i64> %value_arg, <2 x i64>* %p.i, align 1
  ret void
}
```

PR #3583, opened 2020-10-09
Azure CI: Include macOS arm64 libraries in prebuilt Mac package

kinke:
For simple cross-compilation via `-mtriple=arm64-apple-macos` as long as a suited Xcode version >= 12.0 (?) is installed.

### DIFF
--- a/.azure-pipelines/posix.yml
+++ b/.azure-pipelines/posix.yml
@@ -12,6 +12,7 @@ steps:
     set -ex
     cd ..
     if [ "$CI_OS" = "osx" ]; then
+      sudo xcode-select --switch /Applications/Xcode_12.2.app/Contents/Developer
       # Download & extract Ninja
       curl -L -o ninja-mac.zip https://github.com/ninja-build/ninja/releases/download/v1.10.0/ninja-mac.zip
       mkdir ninja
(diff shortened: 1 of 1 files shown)


kinke:
@p0nce: This should be usable now; you can download an artifact from https://dev.azure.com/ldc-developers/ldc/_build/results?buildId=2232&view=artifacts&type=publishedArtifacts.

There are gazillions of linker warnings like
> ld: warning: direct access in function 'ltmp2' from file 'objects-debug-shared/core/demangle.o' to global weak symbol '__D4core8demangle__T8DemangleTSQBcQBa7NoHooksZQBa15parseSymbolNameMFNaNlNfZv' from file 'objects-debug-shared/core/demangle.o' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.

They happened for iOS too, but have been suppressed via `-fvisibility=hidden`. For macOS arm64 here, the object files are used for the shared libs too (not just static libs as for iOS), so that's not an option, and `-fvisibility-inlines-hidden` neither due to D's template emission. I think these warnings can be safely ignored and just mean that there might be multiple identical symbol definitions in a process (defined in different .dylibs / executable); actual weak symbols (`@weak`) are hopefully not accessed directly and can still be overridden.

Anyway, this shouldn't be related in any way to LDC itself, as the IR should be equivalent in this regard for macOS on x64 and arm64. After a quick search, I couldn't find a way to suppress the linker warning.

.azure-pipelines/posix.yml
```diff
@@ -267,48 +268,74 @@ steps:
...
+    rm installed/lib-{arm64,ios-arm64,ios-x86_64}/*-lto-debug.a
     sections="
+    \"arm64-apple-macos\":
+    {
+        switches = [
+            \"-defaultlib=phobos2-ldc,druntime-ldc\",
+            \"-Xcc=-target\",
+            \"-Xcc=$triple_arm64\",
+            \"-Xcc=-isysroot\",
+            \"-Xcc=/Applications/Xcode_12.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk\",
```
kinke, line 323: It would be good to know whether this `-isysroot ...` is really required for linking when using an Xcode with SDK >= 11.0, or whether it's better to leave it out and just assume the user uses an appropriate default Xcode version.

p0nce: Hello, I'll get back to you this evening (GMT+1) about testing this.

p0nce: I also get the gazillions of such warning, and I had assumed `-fvisibility=hidden`  would be possible. Well I trust your judgement on this. :)

p0nce:
- I'm able to use the compiler for arm64 without problems except the linker warnings.
- the `-isysroot` parameters doesn't seem to be needed. My Xcode here is named `Xcode-beta.app` and possibly has a slightly different SDK.
- I was able to build an arm64 shared library (that doesn't enable the D runtime) and call it like a C lib. Haven't tried enabling the runtime from a shared library.

kinke:
> the -isysroot parameters doesn't seem to be needed

Tested on native arm64, right? Well, I'll just try it here too with Azure on native x64, we'll see...

p0nce:
> Tested on native arm64, right? 
Yes.

kinke:
Specifying the sysroot is needed for Azure at least; not sure if that's because it's native x64 or because it has something like 10 installed Xcode versions.

PR #3585, opened 2020-10-11
Refactoring: Prepare for proper -linkonce-templates

kinke:
Some stuff extracted and cleaned up from #3422.

### DIFF
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -56,7 +56,7 @@ void DtoResolveClass(ClassDeclaration *cd) {
   DtoType(cd->type);
 
   // create IrAggr
-  IrClass *irAggr = getIrAggr(cd, true);
+  getIrAggr(cd, true);
 
   // make sure all fields really get their ir field
   for (auto vd : cd->fields) {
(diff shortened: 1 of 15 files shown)


gen/classes.cpp
```diff
@@ -56,7 +56,7 @@ void DtoResolveClass(ClassDeclaration *cd) {
   DtoType(cd->type);
 
   // create IrAggr
-  IrClass *irAggr = getIrAggr(cd, true);
+  getIrAggr(cd, true);
```
JohanEngelen, line 59: maybe add something to the name of all these `get*` functions, to make it easier to read here that something is happening inside (define)? It's pretty easy to confuse a `get*` function as just getting something that isn't stored anywhere and thus can be removed.

kinke: The `getIr*(*, bool create = false)` and new `get<Symbol>(bool define = false)` functions are similar, but the former make sure the `Ir*` instances are either already created or to be created lazily with `true`, and the latter lazily declare the IR symbol and optionally define it too with `true`. They all have in common that there's usually a bunch of normal gets and one location where it is created/defined.
I agree that it's not very nice to read, but no better name comes to my mind immediately, and splitting the creation/definition off into a separate call for the single call site seems a bit excessive IMO.

PR #3598, opened 2020-10-25
Mention the VisualD Windows installer in the README.

p0nce:
My undecisive suggestion would be to just link to the VisualD installer for those users who would rather have a branded experience. I don't really have any opinion on this.

README.md would look like:
image


JohanEngelen:
Nice, thanks.

Issue #2538, opened 2018-01-26
Improve coverage for templates (2.078.1)


kinke:
There's no coverage at all for templates if they aren't codegen'd in their declaring module, see https://github.com/ldc-developers/ldc/pull/2486#issuecomment-360307728. dmd-testsuite's `runnable/test13742.sh` has been disabled for now.

PR #3600, opened 2020-10-25
Extend -linkonce-templates by matching template emission scheme

kinke:
[In essence, a revised version of the remaining stuff from #3422.]

I.e., *define* templated symbols in each referencing compilation unit (=> object file) when using discardable `linkonce_odr` linkage, analogous to C++.

This makes each compilation unit self-sufficient wrt. templated symbols, which also means increased opportunity for inlining and less need for LTO. There should be no more undefined symbol issues caused by buggy template culling.

The biggest advantage is that the optimizer can discard unused `linkonce_odr` symbols early instead of optimizing and forwarding to the assembler. So this is especially useful with `-O` to decrease compilation times and can at least in some scenarios greatly outweigh the (potentially very much) higher number of symbols defined by the glue layer.

Libraries compiled with `-linkonce-templates` can generally not be linked against dependent code compiled without `-linkonce-templates`; the other way around works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -167,9 +167,6 @@ script:
   - |
     if [ "$TRAVIS_CPU_ARCH" = "arm64" ]; then
       DMD_TESTSUITE_MAKE_ARGS="-j${PARALLELISM:-3} GDB_FLAGS=OFF" ctest -V -R "dmd-testsuite"
-    elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
-      # dmd-testsuite-debug only to reduce time-outs
-      DMD_TESTSUITE_MAKE_ARGS=-j${PARALLELISM:-3} ctest -V -R "build-run-dmd-testsuite|dmd-testsuite-debug"
     else
       DMD_TESTSUITE_MAKE_ARGS=-j${PARALLELISM:-3} ctest -V -R "dmd-testsuite"
     fi
--- a/dmd/dsymbolsem.d
+++ b/dmd/dsymbolsem.d
@@ -6071,8 +6071,13 @@ void templateInstanceSemantic(TemplateInstance tempinst, Scope* sc, Expressions*
             tempinst.inst.gagged = tempinst.gagged;
         }
 
-        tempinst.tnext = tempinst.inst.tnext;
-        tempinst.inst.tnext = tempinst;
+        // LDC: the tnext linked list is only used by TemplateInstance.needsCodegen(),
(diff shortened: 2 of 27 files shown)


JohanEngelen:
@kinke  Let me know when this is ready to be tested on Weka's codebase.

kinke:
It should be ready, first tests results looking good (benign failures after a quick glance). [It currently defaults to `-linkonce-templates` for testing.]

kinke:
Sorry, there was still a problem for unreferenced symbols like module ctors, fixed now. Only 2 failing tests here; lit-test `codegen/array_equals_memcmp_dyn.d` fails because of discarded (after inlining) function being checked in asm output, i.e., clearly benign, and druntime integration test `profile`, which is probably related to #2538.

kinke:
I've enabled `-linkonce-templates` for all druntime and Phobos unittests, + dmd-testsuite-release. The unittests primarily because of the significant speed-up, and dmd-testsuite for some extra coverage. Note that `-unittest` has a deep impact on the normal template emission (each unique template instantiation is assigned to a root module, even if exclusively instantiated in non-root modules), so I don't think we really lose test coverage for the normal template emission this way.

Some rough timings (only 1 run) and static unittest lib sizes compared to master, using LLVM 11 *without* assertions on Win64 on my personal quad-core:

| | Master | This (rebased) |
| --- | --- | --- |
| dub | 1m 4s | 50s (-22%) |
| phobos2-test-runner | 6m 20s, 191 MB | 4m 3s (-36%), 69 MB (-64%) |
| phobos2-test-runner-debug | 2m 9s, 368 MB | 2m (-7%), 339 MB (-8%) |
| druntime-test-runner | 22s, 9.5 MB | 19s (-14%), 6.2 MB (-35%)  |
| druntime-test-runner-debug | 14s, 21 MB | 14s (±0%), 19 MB (-10%)  |

This is in line with some early tests in #3422. The fact that each unittest lib is smaller than before, including the debug ones without any optimization, shows that the normal template emission for `-unittest` is extremely conservative and yields more defined symbols than with new `-linkonce-templates`. Most likely because this scheme here (define-on-1st-declaration for templated symbols and each compilation unit) works on a symbol basis - i.e., if we are only referencing a method `MyStruct!(...).foo()`, only that one is defined, not the whole `MyStruct!(...)` struct instantiation.

ibuclaw:
Nice, maybe ping @atilaneves and maybe align this up with what he's doing for dmd?

PR #3601, opened 2020-10-27
Add CI-automated Windows installer based on Inno Setup

kinke:
With the option to add the `bin` dir to the current user's `PATH`, choosing x64 and/or x86 libraries, and optionally opening `README.txt` after installation.

An uninstaller is generated and accessible from 'add/remove programs'.

The `major.minor` version part (or first 8 commit hash chars for CI builds) constitute the 'app ID' - v1.24.1 upgrades v1.24.0, but
v1.25.0-beta1 is a separate 1.25 family installed side-by-side.

Also reflect installed LDC versions in (the 32-bit view of) the registry - e.g., key `HKLM\SOFTWARE\LDC\1.24` with values `InstallationFolder` and `Version` (`1.24.0`).

image
image
image
image

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -42,7 +42,9 @@ jobs:
   dependsOn:
     - Windows
   steps:
-    - checkout: none
+    - checkout: self
+      submodules: false
+      fetchDepth: 50
     - task: DownloadPipelineArtifact@2
       inputs:
         artifactName: windows-x64
@@ -60,7 +62,8 @@ jobs:
           set ARTIFACT_ID=%BUILD_SOURCEVERSION:~0,8%
         )
         set ARTIFACT_NAME=ldc2-%ARTIFACT_ID%-windows-multilib
-        :: Persist the ARTIFACT_NAME env variable for future steps
+        :: Persist the ARTIFACT_{ID,NAME} env variables for future steps
+        echo ##vso[task.setvariable variable=ARTIFACT_ID]%ARTIFACT_ID%
         echo ##vso[task.setvariable variable=ARTIFACT_NAME]%ARTIFACT_NAME%
         7z x artifacts/ldc2-*-x64.7z > nul
         7z x artifacts/ldc2-*-x86.7z > nul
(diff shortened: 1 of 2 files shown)


kinke:
Should also be nice to integrate in other projects via `/VERYSILENT [/CURRENTUSER] [/DIR="C:\bla"]` - no need for a 7z unpacker.

@rainers: What do you think?

rainers:
> Should also be nice to integrate in other projects via `/VERYSILENT [/CURRENTUSER] [/DIR="C:\bla"]` - no need for a 7z unpacker.
> 
> @rainers: What do you think?

Does this automatically work with Inno-Setup? Sounds nice, but I'd probably stay with the 7z files for Visual D for now. A 7za.dll is also used to unpack dmd.

BTW: Your install script looks a lot cleaner than the mess that NSIS needs, but maybe the current task is also simpler.

> The `major.minor` version part (or first 8 commit hash chars for CI builds) constitute the 'app ID' - v1.24.1 upgrades v1.24.0, but
> v1.25.0-beta1 is a separate 1.25 family installed side-by-side.

It might get confusing when setting the PATH variable multiple times.

Maybe it would be worth avoiding to have LDC installed both by the installer and Visual D. If you want to be compatible with how Visual D updates the compilers: it unpacks the multilib 7z-files into a user specified root folder, which results in "c:\D\ldc2-1.24.0-windows-multilib", for example. That doesn't look as clean as "LDC 1.24", though.

kinke:
> Does this automatically work with Inno-Setup? Sounds nice, but I'd probably stay with the 7z files for Visual D for now. A 7za.dll is also used to unpack dmd.

Sure, if you need a 7za.dll anyway, then keep on using the .7z, it's something like 15% (~8 MB) smaller.

> BTW: Your install script looks a lot cleaner than the mess that NSIS needs, but maybe the current task is also simpler.

Years ago, I went with Inno for another project, and remembered it as way simpler than NSIS and WiX at the time, at least for simple setups like this. With the DelphiScript support it's also quite powerful for more complex tasks.

> It might get confusing when setting the PATH variable multiple times.

Each new install prepends it and thus takes precedence (unless it's already in user PATH). And removes it when uninstalling. So you can give v1.25.0-beta1 a try and use it as default compiler; if the code needs 2.095 adaptations you don't want to get into, you can uninstall it and use the previous version as default compiler again.

> Maybe it would be worth avoiding to have LDC installed both by the installer and Visual D.

Yes, that's why I've added the registry stuff here, so that you can auto-detect installed versions.

PR #3604, opened 2020-11-01
Fix LTO with -link-internally for MSVC targets

### LINKED ISSUE #2657: Integrated LLD (-link-internally) not usable for LTO
See https://github.com/ldc-developers/ldc/pull/2203#issuecomment-339167131 for the problem description (LLVM globals apparently not properly shared between LDC and LLD).

kinke:
Fixes #2657 by deferring the `llvm_shutdown()` call.

This allows to remove bundled `lld-link.exe` for the prebuilt Windows packages.

### DIFF
--- a/.azure-pipelines/windows.yml
+++ b/.azure-pipelines/windows.yml
@@ -68,9 +68,8 @@ steps:
 - script: |
     echo on
     cd ..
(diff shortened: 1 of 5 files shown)


JohanEngelen:
I recently saw that we do llvm_shutdown quite early, so this fixes some other potential issues too. 👍

PR #3608, opened 2020-11-07
Predefine version `FreeStanding` when targeting bare-metal

### LINKED ISSUE #3607: `FreeStanding` version support
Bare-metal targets needs `FreeStanding` version to be pre-defined.

As I see currently it is not so:
```
$ ldc2 -mtriple=thumbv7m-none-none -mcpu=cortex-m3 -float-abi=soft -v -o- -betterC -nogc - </dev/null | grep predefs
predefs   LDC all D_Version2 assert D_BetterC ARM Thumb ARM_Thumb ARM_SoftFloat D_SoftFloat LittleEndian LDC_LLVM_1100
```

```
$ ldc2 --version
LDC - the LLVM D compiler (1.24.0):
  based on DMD v2.094.1 and LLVM 11.0.0
  built with LDC - the LLVM D compiler (1.23.0)
  Default target: x86_64-unknown-linux-gnu
  Host CPU: sandybridge
  http://dlang.org - http://wiki.dlang.org/LDC

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_32 - AArch64 (little endian ILP32)
    aarch64_be - AArch64 (big endian)
    amdgcn     - AMD GCN GPUs
    arm        - ARM
    arm64      - ARM64 (little endian)
    arm64_32   - ARM64 (little endian ILP32)
    armeb      - ARM (big endian)
    avr        - Atmel AVR Microcontroller
    bpf        - BPF (host endian)
    bpfeb      - BPF (big endian)
    bpfel      - BPF (little endian)
    hexagon    - Hexagon
    lanai      - Lanai
    mips       - MIPS (32-bit big endian)
    mips64     - MIPS (64-bit big endian)
    mips64el   - MIPS (64-bit little endian)
    mipsel     - MIPS (32-bit little endian)
    msp430     - MSP430 [experimental]
    nvptx      - NVIDIA PTX 32-bit
    nvptx64    - NVIDIA PTX 64-bit
    ppc32      - PowerPC 32
    ppc64      - PowerPC 64
    ppc64le    - PowerPC 64 LE
    r600       - AMD GPUs HD2XXX-HD6XXX
    riscv32    - 32-bit RISC-V
    riscv64    - 64-bit RISC-V
    sparc      - Sparc
    sparcel    - Sparc LE
    sparcv9    - Sparc V9
    systemz    - SystemZ
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
    wasm32     - WebAssembly 32-bit
    wasm64     - WebAssembly 64-bit
    x86        - 32-bit X86: Pentium-Pro and above
    x86-64     - 64-bit X86: EM64T and AMD64
    xcore      - XCore
```

kinke:
Resolves #3607.

This has been added 5+ years ago with dlang/dlang.org@4c9a20a75f3a577edc89abacc81902f1cb11c32c.

### DIFF
--- a/tests/codegen/unknown_critical_section_size.d
+++ b/tests/codegen/unknown_critical_section_size.d
@@ -2,13 +2,16 @@
 // RUN:     %ldc -mtriple=x86_64 -c -w %s
 // RUN: not %ldc -mtriple=x86_64 -c -w -d-version=WithSynchronized %s 2>&1 | FileCheck %s
 
+// no target OS
+version (FreeStanding) {} else static assert(0);
+
 void foo() {}
 
 version (WithSynchronized)
 void bar()
 {
     __gshared int global;
-    // CHECK: unknown_critical_section_size.d(12): Error: unknown critical section size for the selected target
+    // CHECK: unknown_critical_section_size.d(15): Error: unknown critical section size for the selected target
     synchronized
     {
         global += 10;
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -830,7 +830,9 @@ void registerPredefinedTargetVersions() {
       VersionCondition::addPredefinedGlobalIdent("Android");
     } else {
       llvm::StringRef osName = triple.getOSName();
-      if (!osName.empty() && osName != "unknown" && osName != "none") {
+      if (osName.empty() || osName == "unknown" || osName == "none") {
+        VersionCondition::addPredefinedGlobalIdent("FreeStanding");
+      } else {
         warning(Loc(), "unknown target OS: %s", osName.str().c_str());
       }
     }


thewilsonator (approved):

tests/codegen/unknown_critical_section_size.d
```diff
@@ -2,13 +2,16 @@
 // RUN:     %ldc -mtriple=x86_64 -c -w %s
 // RUN: not %ldc -mtriple=x86_64 -c -w -d-version=WithSynchronized %s 2>&1 | FileCheck %s
```
JohanEngelen, line 3: hm, I always thought that specifying a triple without the 2nd/3rd parts would select the default for the host OS, instead of selecting "no OS"....

tests/codegen/unknown_critical_section_size.d
```diff
@@ -2,13 +2,16 @@
...
+// no target OS
+version (FreeStanding) {} else static assert(0);
+
 void foo() {}
 
 version (WithSynchronized)
 void bar()
 {
     __gshared int global;
-    // CHECK: unknown_critical_section_size.d(12): Error: unknown critical section size for the selected target
```
JohanEngelen, line 11: fyi, FileCheck has `@LINE` for such cases:  `...size.d([[@LINE+1]]): Error...`
http://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-pseudo-numeric-variables

tests/codegen/unknown_critical_section_size.d
JohanEngelen: See Clang's documentation https://clang.llvm.org/docs/CrossCompilation.html#target-triple
Clang also uses the _default_ when a triple part is empty, instead of assuming `none`.
`-ffreestanding` is what we should implement, in my opinion.
See also: https://chromium.googlesource.com/external/llvm.org/clang/+/google/stable/test/CodeGen/PR3589-freestanding-libcalls.c

kinke: > Clang also uses the _default_ when a triple part is empty, instead of assuming `none`.

Doesn't seem to be the case: https://cpp.godbolt.org/z/ThW5jW. I think `-march=x86-64` is to be used when only wanting to target another arch, with the same host OS and environment.

JohanEngelen: Ah OK. Indeed.
I presume `FreeStanding` can be used to disable all parts of druntime that depend on a host OS. So it'd be good to be able to still set the arch without changing OS.
Is LDCs behavior changed now? (e.g. that before `-mtriple=x86_64` would still compile-in all Linux/Windows/... parts.)
 Don't think so right?

kinke: Nothing has changed (`FreeStanding` isn't used anywhere in druntime/Phobos yet); `Windows`, `linux`, `OSX` etc. haven't been predefined before either. I think it just might make disabling parts you've mentioned easier.

PR #3609, opened 2020-11-07
Win64 ABI: Pass/return delegates like slices - in (up to) 2 GP registers

kinke:
I expect this to be slightly more performant than the previous behavior, where a delegate was treated like a corresponding struct, passed via hidden pointer and returned via sret.

The primary motivation is a smooth preparation for PR #3578 - in order to allow people to experiment with `-preview=in` without recompiling druntime and Phobos, `in` slices and delegates must not be passed by-ref with `-preview=in` (see dlang/dmd#11828). This would have required a special case for delegates on Win64, which is IMO better handled this way.

### DIFF
--- a/gen/abi-win64.cpp
+++ b/gen/abi-win64.cpp
@@ -71,7 +71,11 @@ private:
 
     // Remaining aggregates which can NOT be rewritten as integers (size > 8
     // bytes or not a power of 2) are passed by ref to hidden copy.
-    return isAggregate(t) && !canRewriteAsInt(t);
+    // LDC-specific exceptions: slices and delegates are left alone (as non-
+    // rewritten IR structs) and passed/returned as 2 separate args => passed in
+    // up to 2 GP registers and returned in RAX & RDX.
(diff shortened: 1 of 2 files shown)


gen/abi-win64.cpp
```diff
@@ -71,7 +71,11 @@ struct Win64TargetABI : TargetABI {
 
     // Remaining aggregates which can NOT be rewritten as integers (size > 8
     // bytes or not a power of 2) are passed by ref to hidden copy.
-    return isAggregate(t) && !canRewriteAsInt(t);
+    // LDC-specific exceptions: slices and delegates are left alone (as non-
+    // rewritten IR structs) and passed/returned as 2 separate args => passed in
+    // up to 2 GP registers and returned in RAX & RDX.
+    return isAggregate(t) && !canRewriteAsInt(t) && t->ty != Tarray &&
```
kinke, line 77: [`isAggregate(t)` already excludes `Tarray`, so no change for slices. - That's a general limitation, as we cannot ABI-rewrite slices because we don't ABI-rewrite the druntime hooks.]

PR #3613, opened 2020-11-11
druntime: Define rt.aaA.AA as naked pointer, no struct wrapper

kinke:
To fix ABI incompatibilities between the previous struct and `void*` used by the hardcoded compiler prototypes, e.g., for WebAssembly, which uses the default `UnknownTargetABI`, which uses LLVM `byval` for all structs and static arrays.

### DIFF
--- a/gen/runtime.cpp
+++ b/gen/runtime.cpp
@@ -488,7 +488,7 @@ static void buildRuntimeModule() {
   Type *wstringTy = Type::twchar->arrayOf();
   Type *dstringTy = Type::tdchar->arrayOf();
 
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Can we come up with a testcase that would catch this?

kinke:
~Any usage of AAs for such targets should very quickly show that something's off. A lit-test could theoretically import `rt.aaA` and check that `AA` is a pointer, but that's IMO not much of a test.

JohanEngelen:
ah yes, I was confused about the WASM case, but that errored because of a user-written runtime.

PR #3620, opened 2020-11-26
Upgrade frontend & libs to v2.095.0+

kinke:


gen/functions.cpp
```diff
@@ -136,8 +136,6 @@ llvm::FunctionType *DtoFunctionType(Type *type, IrFuncTy &irFty, Type *thistype,
   if (fd && fd->linkage == LINK::objc && thistype) {
     if (fd->objc.selector) {
       hasObjCSelector = true;
-    } else if (fd->parent->isClassDeclaration()) {
-      fd->error("Objective-C `@selector` is missing");
```
kinke, line 140: @jacob-carlborg: Is this removed check okay? Question triggered by https://github.com/dlang/dmd/blob/860d0b2d2f24f3d0ef5cc12e4f43f12aa0e406c8/test/dshell/extra-files/cpp_header_gen/library.d#L116.

jacob-carlborg: Well, `@selector` is always required. But DMD doesn't check that currently.

kinke: Hmm... there's at least one other occurrence: https://github.com/dlang/dmd/blob/860d0b2d2f24f3d0ef5cc12e4f43f12aa0e406c8/test/compilable/objc_class.d#L22

jacob-carlborg: That's a bug and should be fixed. It only breaks if the methods is actually called.

kinke: Okay. I guess the `cpp_header_gen` test only verifies the method isn't put into the vtable, and so guarding it with `version (D_ObjectiveC)` should be fine?

jacob-carlborg: I don't really know what the test does. `extern(Objective-C)` should only be used when `version (D_ObjectiveC)` is active. Perhaps `extern(Objective-C)` is only checked at the class level and not at function level.

jacob-carlborg: https://github.com/dlang/dmd/pull/12004

jacob-carlborg:
Looks like this broke the Bitrise build:

```
1.018s PASS debug64 std.socket
****** FAIL debug64 std.stdio
Invalid UTF sequence: e4x - Invalid UTF-8 sequence (at index 1)
0.084s PASS debug64 std.string
```

https://app.bitrise.io/build/1e25821ef4cba327#?tab=log

kinke:
Yeah I've seen it; nothing I can do about it. I was also wondering about the missing stacktrace.

PR #3621, opened 2020-11-27
CI: Pin bundled dlang-tools version, don't use master anymore

kinke:
A recent bigger dustmite upgrade broke our manual build steps (new module). Pin the git tag in a text file, just like for dub.


PetarKirov:
@kinke do you have a bug report for the Dutmite problem?

kinke:
It's not a dustmite problem, only a 'problem' for our *manual* build steps - we don't use the dlang-tools Makefiles.

PR #3624, opened 2020-11-29
Add --ftime-trace functionality.

JohanEngelen:
This uses LLVM's TimeProfiler functionality, such that LLVM's work is traced in the same profile (optimization and machine code gen).
Functionality is meant to be identical to Clang and LLD's --ftime-trace.

### DIFF
--- a/driver/timetrace_sema.d
+++ b/driver/timetrace_sema.d
@@ -190,35 +190,39 @@ extern(C++) final class SemanticTimeTraceVisitor(SemaVisitor) : Visitor
     override void visit(InterfaceDeclaration idec) { semavisitor.visit(idec); }
 }
 
-// Note: this is not very careful check, but is hopefully good enough to trigger upon relevant changes to the DMD frontend.
+// Note: this is not a very careful check, but is hopefully good enough to trigger upon relevant changes to the DMD frontend.
 // If the parent of First and Second already contains an override
 // function and Second overrides it, this function always returns true even if First does not override it.
 private bool checkFirstOverridesAllSecondOverides(First, Second)() {
-    foreach (der2;  __traits(derivedMembers, Second))
+    // Due to access rights limits of __traits(derivedMembers,...) we require a newer dlang version to do the check
+    static if (__VERSION__ >= 2086)
     {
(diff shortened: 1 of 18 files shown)


kinke:
Oh wow, that's probably a valuable addition...

JohanEngelen:
@kinke  What do you think of the way I added it? I tried to touch DMD frontend code as minimal as possible.

Yeah, it's quite insightful to see the traces. I'm using it to try and speed up compile times at Weka (e.g. now you can see clearly that importing a particular module costs 5 seconds).

If you want to try it, I recommend using Tracy for viewing the profile. Get the profile from LDC, convert it using Tracy's `import-chrome` tool, and view it in Tracy's profile GUI. It's a _much much much_ faster UI than chrome::tracing.

driver/timetrace_sema.d
```diff
@@ -0,0 +1,230 @@
...
+        enum pretext = "Sema3: ";
+    else
+        static assert (false, "did not recognize SemaVisitor ==" ~ SemaVisitor);
+
+    this(SemaVisitor semavisitor)
+    {
+        this.semavisitor = semavisitor;
+    }
+
+    alias visit = Visitor.visit;
```
kinke, line 98: Perhaps just use `alias visit = SemaVisitor.visit`? Or do you want to be notified whenever a new override is added to `SemaVisitor` and decide whether to just forward in the 90+% of cases or to add a new trace scope?

JohanEngelen: That can't work because I have to forward the visit call to `semavisitor.visit` instead of `this.SemaVisitor.visit`. The semantic analysis visitors are `final` which is probably good for performance, but requires this nuisance of manually overriding all overrides (and hence the static if check to fail when something gets added....)

kinke: Ouch yeah I remember some PR which attempted to make them non-final and I opposed against it. ;)

JohanEngelen:
Added a testcase before merging.

JohanEngelen:
Merging. The Azure failures are due to LLVM package not being available for Ubuntu 16 (but they are for 20).

PR #3632, opened 2020-12-13
Fix missing location information for error reporting of TypeInfo in b…

DaveP1776:
…etterC

This fixes an issue where any instantiation of TypeInfo in the final
output would lead to a cryptic error with no file or line information.
This change brings ldc in line with dmd's reporting of the same error,
which at least gives file and line information to discover the problem.


kinke:
Thanks! I took the liberty of slightly revising (incl. an unrelated long-overdue mutable-to-const conversion) and adding a test.

thewilsonator (approved):

PR #3639, opened 2021-01-08
CI: Bump LDC-LLVM to v11.0.1

kinke:


Geod24:
Just wondering why you don't upgrade to 18.04 or 20.04 since 16.04 is EOL in 3-4 months ?

kinke:
I'll just wait these few more months before bumping Ubuntu to 18.04 - I've been using oldest non-EOL Ubuntu for years now; last time, there have been 'complaints', so we're actually using build tweaks so that the prebuilt libs can still be linked with Ubuntu 14's binutils...

PR #3663, opened 2021-02-11
Add support for LLVM 12

kinke:


kinke:
Tested with current vanilla `release/12.x` branch (rc1+) on Linux x64; all tests pass except for `druntime-test-profile` which I haven't looked into. There's a severe issue wrt. emitting bitcode files though - e.g., `ldc2 -output-bc .../druntime/src/core/cpuid.d` hits an assertion deep inside LLVM. Other druntime modules worked, so I suspect once again an issue related to inline asm.

kinke:
Merging eagerly as anticipation for distro packagers requiring LLVM 12 support for v1.25 once the LLVM final is out.

PR #3664, opened 2021-02-12
Ignore -enable-cross-module-inlining if inlining is generally disabled

kinke:
People use `dflags "-enable-cross-module-inlining" platform="ldc"` in their `dub.sdl` files; default and debug builds without `-O` thus incur a superfluous compile-time cost (extra `available_externally` emissions) for no benefit.

### DIFF
--- a/tests/codegen/inlining_disablecross.d
+++ b/tests/codegen/inlining_disablecross.d
@@ -1,21 +1,18 @@
-// Test disabling/enabling of cross-module inlining
+// Test disabling of cross-module inlining
 
-// RUN: %ldc %s -I%S -c -output-ll  -enable-cross-module-inlining -O0 -of=%t.ENA.ll && FileCheck %s --check-prefix ENABLED  < %t.ENA.ll
-// RUN: %ldc %s -I%S -c -output-ll -disable-cross-module-inlining -O3 -of=%t.DIS.ll && FileCheck %s --check-prefix DISABLED < %t.DIS.ll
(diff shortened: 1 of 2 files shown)


tests/codegen/inlining_disablecross.d
```diff
@@ -1,21 +1,18 @@
...
 extern (C): // simplify mangling for easier matching
 
-// DISABLED-LABEL: define{{.*}} @call_easily_inlinable(
-// ENABLED-LABEL: define{{.*}} @call_easily_inlinable(
+// CHECK-LABEL: define{{.*}} @call_easily_inlinable(
 int call_easily_inlinable(int i)
 {
-    // DISABLED: call {{.*}} @easily_inlinable(
+    // CHECK: call {{.*}} @easily_inlinable(
     return easily_inlinable(i);
```
kinke, line 14: The call was previously there for `-enable-cross-module-inlining -O0` too.

JohanEngelen:
Makes total sense. The flag is meant to tweak inlining behavior, and when inlining generally is not enabled, it should not be enabled by this flag for some special cases.

Does this impact `pragma(inline,true)` functions cross-module?

kinke:
> Does this impact pragma(inline,true) functions cross-module?

Nope, these functions circumvent this check: https://github.com/ldc-developers/ldc/blob/aac18988ffb63a12acf933ed46d8a0bc2470a21c/gen/function-inlining.cpp#L104-L110

JohanEngelen (approved):

PR #3673, opened 2021-02-20
Posix: Make LDC_INSTALL_LTOPLUGIN & LDC_INSTALL_LLVM_RUNTIME_LIBS default to ON if LLVM is linked statically

kinke:
And to `OFF` if LLVM is linked dynamically. As we depend on the LLVM .so anyway in that case, it should be safe to depend on its LTO plugin and the associated (clang) compiler-rt libraries as well. The assumption is that most users choosing a shared LLVM prefer dependencies over renamed duplicates, and that users of a static LLVM strive for self-sufficiency or at least not conflicting with potential other LLVM installations.

Use the default values for CI, which means that linker plugin and compiler-rt libs are still included in our prebuilt packages (except for FreeBSD) and that the code locating non-bundled libraries is tested for CI jobs using a shared distro LLVM, like Cirrus and Circle CI on Ubuntu and FreeBSD. That code had to be extended to include the LLVM lib dir when using a shared LLVM (e.g., the install dir on FreeBSD is `/usr/local/llvm90`). And the constraints for according lit tests have been considerably lowered too, simply assuming the compiler-rt libraries are available somewhere (and that the linker chosen by LDC supports the LTO plugin).


CMakeLists.txt
```diff
@@ -804,12 +812,6 @@ if (LDC_INSTALL_LLVM_RUNTIME_LIBS)
         copy_compilerrt_lib("windows/clang_rt.builtins-${compilerrt_arch_suffix}.lib"   "ldc_rt.builtins.lib" FALSE)
         copy_compilerrt_lib("windows/clang_rt.profile-${compilerrt_arch_suffix}.lib"    "ldc_rt.profile.lib"  FALSE)
         copy_compilerrt_lib("windows/clang_rt.fuzzer-${compilerrt_arch_suffix}.lib"     "ldc_rt.fuzzer.lib"   FALSE)
-        copy_compilerrt_lib("windows/clang_rt.xray-${compilerrt_arch_suffix}.lib"       "ldc_rt.xray.lib"     FALSE)
```
JohanEngelen, line 807: is XRay still not available on Windows?

kinke: Nope, sadly not, at least not up to & including v11; I haven't checked 12 yet.

PR #3674, opened 2021-02-21
Prepare for v1.25.0 final: Bump dlang-tools & dub and add changelog

kinke:
As DMD v2.095.1 has been tagged. - Assumes #3673 makes it into the final.


kalev:
Thanks for the release, @kinke! I went ahead and updated the ldc package in Fedora 34 to 1.25.0.

kinke:
Thanks!

PR #3677, opened 2021-02-23
druntime: Fix DWARF backtrace segfault regression in case executable cannot be opened

kinke:
This sadly crept in with the backtrace cherry-pick for v1.25 and requires a hotfix. See dlang/druntime#3382.


JohanEngelen:
Can we add a testcase that would have caught this and related bugs?

kinke:
An undefined 'we' surely can ;) - should be upstream though.

kinke:
[I'll probably wait for a few more days to see if there are other reports and then release v1.25.1.]

PR #3680, opened 2021-03-09
CI: Use `-k0` for ninja cmdlines potentially running into out-of-memory errors

kinke:
This can significantly improve throughput, because ninja keeps on going after a child process dies and tries to build as many targets as it can in the current pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ matrix:
     - os: linux
       arch: arm64
       d: ldc-beta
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -55,14 +55,12 @@ script:
...
     targets='all ldc2-unittest all-test-runners'
-    currentNumJobs=$numJobs
-    for i in {1..5}; do
-      ninja -j$currentNumJobs $targets && break || true
-      currentNumJobs=$(( ($numJobs + $i - 1) / $i ))
-    done
+    ninja -j8 -k0 $targets || \
+      ninja -j8 -k0 $targets || \
+      ninja -j2 -k0 $targets || \
+      ninja -j1 $targets
```
kinke, line 63: Not happy with this setup at all, but the situation is very special for the Travis AArch64 box with apparently just slightly more than 2 GB of memory...

PR #3684, opened 2021-03-13
Fix pip bootstrap URL

thewilsonator:
This error was hit on https://github.com/ldc-developers/ldc/pull/3683

### DIFF
--- a/.azure-pipelines/1-posix-setup.yml
+++ b/.azure-pipelines/1-posix-setup.yml
@@ -40,7 +40,7 @@ steps:
       tar -xf clang.tar.xz --strip 1 -C clang
     fi
     # Install lit
-    curl -OL https://bootstrap.pypa.io/2.7/get-pip.py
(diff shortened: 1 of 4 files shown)


thewilsonator:
Azure linux/ android gives
```
g++-multilib : Depends: gcc-multilib (>= 4:5.3.1-1ubuntu1) but it is not going to be installed
                Depends: g++-5-multilib (>= 5.3.1-3~) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```
noticed on `dlang/dmd` too, unrelated to this.

Geod24:
https://github.com/ldc-developers/ldc/pull/3680

kinke:
Thx; I wanted to keep on experimenting with #3680 for Travis AArch64 but lost all interest in the meantime. Merged now, so at least this annoyance is resolved; for Azure Linux, where MS manages to f*ck this up regularly, we'll probably have to wait for a fix on their end.

thewilsonator:
Thanks!

PR #3694, opened 2021-04-04
Fix overzealous toBasetype'ing for debuginfo types

### LINKED ISSUE #3688: Enums are not showing up correctly as members of a struct when debugging.
```
enum B : byte {x,y}

struct Y {B A;}
int main()
{
Y y = { A:B.x };
return 0;
}

```
Set breakpoint at ```return 0;``` and you will see A as A = 0

kinke:
Resolves #3688.

### DIFF
--- a/tests/debuginfo/args_cdb.d
+++ b/tests/debuginfo/args_cdb.d
@@ -87,7 +87,8 @@ int byValue(ubyte ub, ushort us, uint ui, ulong ul,
 // CHECK: int delegate()
 // CHECK-NEXT: context :
 // CHECK-NEXT: funcptr :
-// CHECK-SAME: args_cdb.main.__lambda
+// CHECK-G-SAME: args_cdb.main.__lambda
+// CHECK-GC-SAME: args_cdb::main::__lambda
 
 // CDB: ?? slice
 // CHECK-G:  struct int[]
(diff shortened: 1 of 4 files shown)


tests/debuginfo/args_cdb.d
```diff
@@ -134,7 +135,7 @@ int byValue(ubyte ub, ushort us, uint ui, ulong ul,
  */
 size_t byValueShort(Large large)
 {
-// CDB: bp1 /1 `args_cdb.d:137`
+// CDB: bp1 /1 `args_cdb.d:138`
```
JohanEngelen, line 138: fyi, FileCheck also gives you `[[@LINE]]` or `[[@LINE+<offset>]]` to check for linenumber output. https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-pseudo-numeric-variables

kinke: I know, but this isn't run through FileCheck.

JohanEngelen: ah yes of course, sorry for silly remark!

kinke: Don't be, thanks for taking a look!

PR #3696, opened 2021-04-10
Cirrus CI: Work around strange FreeBSD std.experimental.allocator unittest regression

kinke:


MoonlightSentinel:
What fails on FreeBSD?

kinke:
```
****** FAIL release64 std.experimental.allocator
core.exception.AssertError@std/experimental/allocator/package.d(3738): Assertion failure
```

Twice in a row for #3678 (I thought it was just spurious when seeing it the first time). Only with enabled optimizations *and* shared druntime/Phobos, and by the looks of it, not related to any recent changes. But it was green earlier today with master, and green before today's additions to #3678, which included these frontend changes.

MoonlightSentinel:
Note that the failing test was disabled for a long time due to a spurious `__EOF__`, see dlang/phobos#7778.
Did `master` already include this PR?.

kinke:
Wow, wtf is/was that `__EOF__`?! Dude...

Still fails without this last-minute frontend diff, so something else added today in #3678 must have brought this issue to light. The `__EOF__` was already removed in #3678 before today (Phobos wasn't touched today at all).

PR #3701, opened 2021-04-18
CI: Bump LDC-LLVM to v12.0.0

kinke:
Including bumping the Ubuntu base image from 16.04 to 18.04.


kinke:
Remaining failures:
- [ ] Windows: lit-test `PGO/indirect_calls.d`
- [ ] 32-bit x86: `std.conv` unittests with enabled optimizations, lines 5876, 5885, …?

tests/PGO/indirect_calls.d
```diff
@@ -2,6 +2,9 @@
 
 // REQUIRES: PGO_RT
 
+// PROFUSE IR matching has regressed on Windows since LLVM 12 (due to our manual @weak emulation?)
```
JohanEngelen, line 5: Does it work if you remove `@weak` and add inline asm noop or something else that prevents the optimizer from knowing what the function does?

kinke: Oh yeah, `asm {}` makes it work. :+1:

PR #3710, opened 2021-04-30
Use new OpenFlags enumeration

kubo39:
[This](https://reviews.llvm.org/D101506) drops old deprecated OpenFlags support.
OF_{Append,None,Text} has been available since LLVM 9.

- https://reviews.llvm.org/rG1f67a3cba9b09636c56e2109d8a35ae96dc15782


kinke:
Thx!

PR #3711, opened 2021-05-02
Implement --cppstdlib

omerfirmak:
This flag allows user to select from GCC/Clang C++ runtimes
on Linux similar to `--stdlib` on `clang`.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -784,7 +784,10 @@ void registerPredefinedTargetVersions() {
       VersionCondition::addPredefinedGlobalIdent("CRuntime_UClibc");
     } else {
       VersionCondition::addPredefinedGlobalIdent("CRuntime_Glibc");
-      VersionCondition::addPredefinedGlobalIdent("CppRuntime_Gcc");
+      if (global.params.cppstdlib.compare("libc++") == 0)
+        VersionCondition::addPredefinedGlobalIdent("CppRuntime_Clang");
+      else
+        VersionCondition::addPredefinedGlobalIdent("CppRuntime_Gcc");
     }
     break;
   case llvm::Triple::Haiku:
--- a/dmd/globals.d
+++ b/dmd/globals.d
@@ -351,6 +351,7 @@ version (IN_LLVM)
     // target stuff
     const(void)* targetTriple; // const llvm::Triple*
     bool isUClibcEnvironment;
+    const(char)[] cppstdlib; // Cpp runtime library
 
     // Codegen cl options
     bool disableRedZone;
--- a/dmd/globals.h
+++ b/dmd/globals.h
@@ -313,6 +313,7 @@ struct Param
 
     const llvm::Triple *targetTriple;
     bool isUClibcEnvironment; // not directly supported by LLVM
+    DString cppstdlib; // Cpp runtime library
 
     // Codegen cl options
     bool disableRedZone;
--- a/dmd/mars.d
+++ b/dmd/mars.d
@@ -83,6 +83,7 @@ version (IN_LLVM)
     const(char)* getPathToProducedBinary();
     void deleteExeFile();
     int runProgram();
+    const (char)* getExplicitCppRuntimeName();
 }
 else
 {
@@ -431,6 +432,8 @@ else
 
 version (IN_LLVM)
 {
+    import dmd.root.string : toDString;
+    global.params.cppstdlib = toDString(getExplicitCppRuntimeName());
     registerPredefinedVersions();
 }
 else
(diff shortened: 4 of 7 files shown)


thewilsonator:
cc @Geod42

This will probably will be implemented upstream too

Geod24:
I think that should be done through `mtriple` tho, if possible at all ?

omerfirmak:
AFAICS, llvm::triple does not have support for this.
Even `clang` has a separate flag for it.

Geod24:
Then I guess `-cppstdlib` it is. What do you think @kinke ?

thewilsonator:
I'd prefer that this be done upstream and then adapted for LDC as a) DMD will need this too, and b) target.d and the command line processing for target specific things has changed a lot recently and this will only add to that.

driver/main.cpp
```diff
@@ -784,7 +784,10 @@ void registerPredefinedTargetVersions() {
       VersionCondition::addPredefinedGlobalIdent("CRuntime_UClibc");
     } else {
       VersionCondition::addPredefinedGlobalIdent("CRuntime_Glibc");
-      VersionCondition::addPredefinedGlobalIdent("CppRuntime_Gcc");
+      if (std::string(global.params.cppstdlib.ptr) == "libc++")
```
JohanEngelen, line 787: Because `cppstdlib` is a D string, it is not necessarily null terminated. This should be a `stricmp` call, using `cppstdlib.length`.

JohanEngelen: Probably best to add a helper function next to where Dstring helper struct is defined.

omerfirmak: Thanks, done.

omerfirmak:
> I'd prefer that this be done upstream and then adapted for LDC as a) DMD will need this too, and b) target.d and the command line processing for target specific things has changed a lot recently and this will only add to that.

Is there a consensus over this?

kinke:
Sorry for the late reply, but this PR popped up while I was on vacation. - I'm not a big fan of a new switch, nor of extending the 'triple' in some LDC-specific, non-LLVM-compatible way. Mainly because it's only useful for the few `core.stdcpp` modules and the according `CppRuntime_*` versions they require. If it was about linking only, that's already covered via `-Xcc=--stdlib=...` etc. The problem with the druntime modules is that they are precompiled using a specific `CppRuntime_*`, so a user cannot simply take a prebuilt LDC package and change the C++ runtime on the fly and expect everything to work; druntime has to be recompiled.

Extracting the few C++ stdlib bindings into a separate dub package, with e.g. dub configurations enabling to switch between libstdc++ and libc++ would IMO make much more sense.

Wrt. triples in general, even a C++ runtime addition wouldn't define a target platform sufficiently. We have to guess the used C compiler (gcc/clang) in some ABI-specific parts, where clang skips empty-struct parameters while gcc doesn't etc.

JohanEngelen:
> Sorry for the late reply, but this PR popped up while I was on vacation. - I'm not a big fan of a new switch, nor of extending the 'triple' in some LDC-specific, non-LLVM-compatible way. Mainly because it's only useful for the few `core.stdcpp` modules and the according `CppRuntime_*` versions they require. If it was about linking only, that's already covered via `-Xcc=--stdlib=...` etc. The problem with the druntime modules is that they are precompiled using a specific `CppRuntime_*`, so a user cannot simply take a prebuilt LDC package and change the C++ runtime on the fly and expect everything to work; druntime has to be recompiled.
> 
> Extracting the few C++ stdlib bindings into a separate dub package, with e.g. dub configurations enabling to switch between libstdc++ and libc++ would IMO make much more sense.

I agree about not adding this to the triple.
Splitting the stdlib bindings into a separate package makes sense, I think so too, but still requires a compiler change to not set a `CppRuntime_*` version at all and let the user explicitly `-d-version=..` it (e.g. inside `ldc2.conf` file).
Adding a commandline flag or removing the hardcoded `CppRuntime_*` version (and put it in ldc2.conf by default) are the only short term solutions I see.

Geod24:
I fully agree with you guys, but Andrei didn't: https://github.com/dlang/druntime/pull/2286#issuecomment-416704437
I guess I will have to email him.

Geod24:
@kinke : I just found out that no `CppRuntime` is defined on Musl. Is that on purpose ?

kinke:
Just no idea about the dominant C++ library for musl, uclibc, some BSDs… - I preferred defining none over a wild guess, in the hopes that people using such platforms would notice and open according PRs when trying to use `core.stdcpp`, instead of seeing crashes if the guess was bad.

Geod24:
Will do that soon (currently testing https://github.com/bosagora/docker-agora-builder/commit/e7fab4679eaff5c8149665a1385fd0ab4881d76f )

PR #3704, opened 2021-04-24
CMake: Support BUILD_SHARED_LIBS for Windows too

kinke:
By compiling druntime/Phobos with `-fvisibility=public` for the new DLLs ~~(=> #3703 included as prerequisite for now)~~.

Compiling and linking druntime works fine:
```
> ls -lh lib/druntime-ldc*
-rwxr-xr-x 1 Martin 197121 1.9M Apr 26 21:35 lib/druntime-ldc-debug-shared.dll
-rw-r--r-- 1 Martin 197121 3.0M Apr 26 18:51 lib/druntime-ldc-debug-shared.lib
-rw-r--r-- 1 Martin 197121 6.0M Apr 26 21:35 lib/druntime-ldc-debug-shared.pdb
-rw-r--r-- 1 Martin 197121 8.6M Apr 26 21:35 lib/druntime-ldc-debug.lib
-rw-r--r-- 1 Martin 197121 4.7M Apr 26 21:35 lib/druntime-ldc-lto.lib
-rwxr-xr-x 1 Martin 197121 1.7M Apr 26 22:10 lib/druntime-ldc-shared.dll
-rw-r--r-- 1 Martin 197121 2.9M Apr 26 21:59 lib/druntime-ldc-shared.lib
-rw-r--r-- 1 Martin 197121 2.5M Apr 26 22:10 lib/druntime-ldc-shared.pdb
-rw-r--r-- 1 Martin 197121 4.4M Apr 26 21:35 lib/druntime-ldc.lib
```

The druntime DLL exports more than 10k symbols. ~~The MS C runtime is linked statically; that should probably be changed to the MS DLLs too.~~ **Edit**: done.

Linking a non-Phobos hello-world app against the druntime DLL works; it unsurprisingly crashes at runtime though.

The Phobos DLL can be compiled too, but linking fails (builtin TypeInfo init symbols and vtables etc.).

Issue #2783, opened 2018-07-18
LLVM Runtime Preemption specifier


JohanEngelen:
Have a look at this: https://llvm.org/docs/LangRef.html#runtime-preemption-specifiers
Looks like it is new in LLVM. Clang applies `dso_local` to normal functions.
https://godbolt.org/g/RKDtBU

PR #3713, opened 2021-05-03
Mark a function or global variable as dso_local if possible

kubo39:
dso_local allows the compiler to assume a function or global variable will resolve to a symbol within the same linkage unit.

reference: https://llvm.org/docs/LangRef.html#runtime-preemption-specifiers

### DIFF
--- a/tests/codegen/dso_local.d
+++ b/tests/codegen/dso_local.d
@@ -1,0 +1,46 @@
+// RUN: %ldc -mtriple=x86_64-linux-gnu --relocation-model=static --output-ll -of=%t.ll %s && FileCheck --check-prefix=ELF_STATIC_RELOC %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-linux-gnu --relocation-model=pic --output-ll -of=%t.ll %s && FileCheck --check-prefix=ELF_PIC_DEFAULT %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-linux-gnu -fvisibility=hidden --relocation-model=pic --output-ll -of=%t.ll %s && FileCheck --check-prefix=ELF_PIC_HIDDEN %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-apple-darwin --relocation-model=static --output-ll -of=%t.ll %s && FileCheck --check-prefix=MACHO_STATIC_RELOC %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-apple-darwin --relocation-model=pic --output-ll -of=%t.ll %s && FileCheck --check-prefix=MACHO_PIC %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-windows-coff -output-ll -of=%t.ll %s && FileCheck --check-prefix=COFF %s < %t.ll
+
+import ldc.attributes : weak;
+
+// ELF_STATIC_RELOC: define{{( dso_local)?}} i32 @{{.*}}3foo
+// ELF_PIC_DEFAULT: define i32 @{{.*}}3foo
+// ELF_PIC_HIDDEN: define{{( dso_local)?}} hidden i32 @{{.*}}3foo
+// MACHO_STATIC_RELOC: define dso_local i32 @{{.*}}3foo
+// MACHO_PIC: define dso_local i32 @{{.*}}3foo
+// COFF: define dso_local {{.*}} i32 @{{.*}}3foo
+private int foo() { return 42; }
+
+
+// ELF_STATIC_RELOC: define{{( dso_local)?}} i32 @{{.*}}3bar
+// ELF_PIC_DEFAULT: define i32 @{{.*}}3bar
+// ELF_PIC_HIDDEN: define hidden i32 @{{.*}}3bar
+// MACHO_STATIC_RELOC: define dso_local i32 @{{.*}}3bar
+// MACHO_PIC: define dso_local i32 @{{.*}}3bar
+// COFF: define dso_local {{.*}} i32 @{{.*}}3bar
+public int bar() { return foo(); }
+
+
+// ELF_STATIC_RELOC: define weak{{( dso_local)?}} i32 @{{.*}}3baz
+// ELF_PIC_DEFAULT: define weak i32 @{{.*}}3baz
+// ELF_PIC_HIDDEN: define weak hidden i32 @{{.*}}3baz
+// MACHO_STATIC_RELOC: define weak i32 @{{.*}}3baz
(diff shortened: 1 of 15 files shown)


JohanEngelen:
See https://github.com/ldc-developers/ldc/issues/2783

JohanEngelen:
Is `dso_local` correct in all cases? 
- weak functions
- with all relocation models
- DLL export
- ...?

kubo39:
@JohanEngelen 

> Is dso_local correct in all cases?

It depends on binary format(i.e,` on COFF, extern_weak or dllimport GlobalValues must be dso_preemptable), These rules have been implemented in LLVM(shouldAssumeDSOLocal function) and I use it.

My concern is that LLVM 12 [drops  implied dso_local for definitions in ELF static relocation model/PIE](https://github.com/llvm/llvm-project/commit/2047c10c22b071cccc57a7e2779d6603512e9113) to avoid copy relocation.
Clang/Rust implemented the original shouldAssumeDSOLocal lfunctions which implies dso_local in such case, I'm not sure it'd better in out case.

JohanEngelen:
Please add testcases for all known interesting cases. Then automatically we will notice when LLVM's behavior is changing.

tests/codegen/dso_local_extern_weak.d
```diff
@@ -0,0 +1,10 @@
+// RUN: %ldc -mtriple=x86_64-windows-coff -c -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+// CHECK: declare extern_weak void @weakreffunction
+pragma(LDC_extern_weak) extern(C) void weakreffunction();
+
```
JohanEngelen, line 5: this should also have a test for `@weak`, see https://wiki.dlang.org/LDC-specific_language_changes#.40.28ldc.attributes.weak.29

tests/codegen/dso_local_elf_with_static_reloc.d
```diff
@@ -0,0 +1,5 @@
+// REQUIRES: atleast_llvm1200
```
JohanEngelen, line 1: what makes this test LLVM>=12?   Is the behavior still correct for <12 ?

JohanEngelen:
For the testing, I think it is more clear if you put all possibilities in one file, and have multiple `//RUN` lines with different triples and flags, and distinguish between the different cases using a prefix for FileCheck. That way, it is easier to see that the testing is complete and what the behavior differences are (and should be) for different triples.

tests/codegen/dso_local.d
```diff
@@ -0,0 +1,46 @@
+// RUN: %ldc -mtriple=x86_64-linux-gnu --relocation-model=static --output-ll -of=%t.ll %s && FileCheck --check-prefix=ELF_STATIC_RELOC %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-linux-gnu --relocation-model=pic --output-ll -of=%t.ll %s && FileCheck --check-prefix=ELF_PIC_DEFAULT %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-linux-gnu -fvisibility=hidden --relocation-model=pic --output-ll -of=%t.ll %s && FileCheck --check-prefix=ELF_PIC_HIDDEN %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-apple-darwin --relocation-model=static --output-ll -of=%t.ll %s && FileCheck --check-prefix=MACHO_STATIC_RELOC %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-apple-darwin --relocation-model=pic --output-ll -of=%t.ll %s && FileCheck --check-prefix=MACHO_PIC %s < %t.ll
+// RUN: %ldc -mtriple=x86_64-windows-coff -output-ll -of=%t.ll %s && FileCheck --check-prefix=COFF %s < %t.ll
+
+import ldc.attributes : weak;
+
+// ELF_STATIC_RELOC: define{{( dso_local)?}} i32 @{{.*}}3foo
```
JohanEngelen, line 10: Don't we want dso_local here and in the other cases? (not optional)

tests/codegen/dso_local.d
```diff
@@ -0,0 +1,46 @@
...
+// ELF_STATIC_RELOC: define weak{{( dso_local)?}} i32 @{{.*}}3baz
+// ELF_PIC_DEFAULT: define weak i32 @{{.*}}3baz
+// ELF_PIC_HIDDEN: define weak hidden i32 @{{.*}}3baz
+// MACHO_STATIC_RELOC: define weak i32 @{{.*}}3baz
+// MACHO_PIC: define weak i32 @{{.*}}3baz
+// COFF: define x86_vectorcallcc i32 @{{.*}}3baz
+@weak int baz() { return 42; }
+
+
+version(Windows)
```
JohanEngelen, line 37: why the version(Windows) ? This should be tested for all triples. The extern_weak attribute is afaik only useful for ELF object files.

tests/codegen/dso_local.d
```diff
@@ -0,0 +1,46 @@
...
+
+import ldc.attributes : weak;
+
+// ELF_STATIC_RELOC: define{{( dso_local)?}} i32 @{{.*}}3foo
+// ELF_PIC_DEFAULT: define i32 @{{.*}}3foo
+// ELF_PIC_HIDDEN: define{{( dso_local)?}} hidden i32 @{{.*}}3foo
+// MACHO_STATIC_RELOC: define dso_local i32 @{{.*}}3foo
+// MACHO_PIC: define dso_local i32 @{{.*}}3foo
+// COFF: define dso_local {{.*}} i32 @{{.*}}3foo
+private int foo() { return 42; }
```
JohanEngelen, line 16: besides functions, can you also add global variables? (`@weak` is also applicable to global variables btw)

kinke:
I doubt this has any real effect, because the spec talks about

> Direct access will be generated even if the definition is not within this compilation unit.

but the attribute is currently set for *definitions* only. So I think the interesting case is the opposite, *declarations* in other IR modules.

It most likely also won't affect Windows at all, because the loader doesn't support such runtime preemption AFAIK.

Whether these direct accesses are safe is another question; TypeInfos and instantiated symbols might be defined in multiple shared objects (exe/.so), so identity checks relying on the loader uniquing such symbols will probably break. AFAIK, that's why MSVC++ EH is based on type *strings*, and why comparing a `TypeInfo` will probably have to be based on string comparisons as well - e.g., I think this accounts for the last remaining failure in #3704.

kubo39:
Ah,, true. I'm wrong so closing now., sorry for noise.

PR #3720, opened 2021-05-13
Fix compilation on MacOS with read-only root filesystems

thewilsonator:
Supply a `WORKING_DIRECTORY` option so that it choose something other than `/`


cmake/Modules/ExtractDMDSystemLinker.cmake
```diff
@@ -33,6 +33,7 @@ execute_process(
     RESULT_VARIABLE result_code
     OUTPUT_VARIABLE stdout
     OUTPUT_STRIP_TRAILING_WHITESPACE
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
```
thewilsonator, line 36: I'm not quite sure what the best choice of directory is here but this is better than nothing.

JohanEngelen (approved):

PR #3721, opened 2021-05-13
Fix compilation with recent LLVM

thewilsonator:
There are still a bunch of deprecations with `CreateLoad` that need to be addressed at some point.

This is based off of #3720 because I can't build without that.


driver/codegenerator.cpp
```diff
@@ -263,9 +263,9 @@ void CodeGenerator::writeAndFreeLLModule(const char *filename) {
   Version.append(global.ldc_version.ptr, global.ldc_version.length);
   llvm::Metadata *IdentNode[] = {llvm::MDString::get(ir_->context(), Version)};
   IdentMetadata->addOperand(llvm::MDNode::get(ir_->context(), IdentNode));
-
+#if LDC_LLVM_VER < 1300
   context_.setInlineAsmDiagnosticHandler(inlineAsmDiagnosticHandler, ir_);
-
+#endif
```
thewilsonator, line 268: This seems to have been removed. I'm not sure what, if anything has replaced it.

kubo39: looks like unified in llvm::DiagnoseInfoSrcMgr.
https://reviews.llvm.org/rG5de2d189e6ad466a1f0616195e8c524a4eb3cbc0

gen/coverage.cpp
```diff
@@ -38,6 +38,9 @@ void emitCoverageLinecountInc(const Loc &loc) {
 
   // Do an atomic increment, so this works when multiple threads are executed.
   gIR->ir->CreateAtomicRMW(llvm::AtomicRMWInst::Add, ptr, DtoConstUint(1),
+#if LDC_LLVM_VER >= 1300
+                           llvm::MaybeAlign(),
```
JohanEngelen, line 42: this can be explicit 4-byte (uint) alignment

thewilsonator:
superseded by #3822

PR #3724, opened 2021-05-16
Add -cov-increment option to specify the type of increment instruction

JohanEngelen:
This gives the user the ability to get a more performant coverage count execution, at the cost of thread-safety or just boolean true/false on a line (no more counting).

I restricted the testing to Linux, due to the test commandline crazyness (the coverage output filename is very annoying and cannot be set by the user).


JohanEngelen:
Rebased and pushed to retrigger CI.

JohanEngelen:
finally CI is green :)

thewilsonator (approved):

PR #3727, opened 2021-05-16
Fix Circle CI: use python 3

JohanEngelen:


JohanEngelen:
Note: PEP394 recommends to use `python3` explicitly, so let's move to that instead of keeping using `python` https://legacy.python.org/dev/peps/pep-0394/  . New PR incoming...

PR #3728, opened 2021-05-17
Switch to using Python 3

JohanEngelen:
PEP394 recommends to use `python3` explicitly, so let's move to that instead of keeping using `python` https://legacy.python.org/dev/peps/pep-0394/  .

### DIFF
--- a/.azure-pipelines/1-posix-setup.yml
+++ b/.azure-pipelines/1-posix-setup.yml
@@ -33,17 +33,15 @@ steps:
       sudo apt-get -yq install \
         git-core cmake ninja-build g++-multilib \
         libcurl3 libcurl3:i386 \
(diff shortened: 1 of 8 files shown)


JohanEngelen:
We need Python >= 3.6, see https://llvm.org/docs/GettingStarted.html#software

JohanEngelen:
Azure info on installing python: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/use-python-version?view=azure-devops

JohanEngelen:
Ubuntu 16.04 will be dropped soon, moving to 18 on Azure.  https://github.com/actions/virtual-environments/issues/3287

JohanEngelen:
The remaining errors are gdb related

JohanEngelen:
Merged eagerly to improve our CI testing.

PR #3732, opened 2021-05-20
Fix some test regressions with recent GDB versions

kinke:


kinke:
@ibuclaw: Do you happen to know something about GDB 10 changes wrt. globals? Like added C++ module support or so? - Still seeing errors like `No symbol "globVal" in current context.` and `'classtypes_gdb.gvar' has unknown type; cast it to its declared type`...

ibuclaw:
@kinke haven't been keeping tabs on it for a while now, so not surprised if it is slowly deteriorating.

PR #3739, opened 2021-05-25
CI: Ship reggae with prebuilt packages

kinke:
Incompatibilities with LDC and Windows in general have recently been fixed. @atilaneves is fine with reggae being included.


.azure-pipelines/6-package.yml
```diff
@@ -43,6 +44,20 @@ steps:
     $DMD -w -de -dip1000 DustMite/dustmite.d DustMite/splitter.d DustMite/polyhash.d -of=bin/dustmite
     cp bin/{rdmd,ddemangle,dustmite} ../installed/bin
   displayName: Build & copy dlang tools
+- bash: |
+    set -ex
+    cd ..
+    git clone --recursive https://github.com/atilaneves/reggae.git
+    cd reggae
+    git checkout "$(cat $BUILD_SOURCESDIRECTORY/packaging/reggae_version)"
+    DFLAGS="-O $DFLAGS" ../host-ldc/bin/dub build -v --build-mode=allAtOnce --combined \
```
kinke, line 53: `--build-mode=allAtOnce` is a workaround for the dub issue fixed in https://github.com/dlang/dub/pull/2135 - compiling & linking at once avoids the 'D-linker-flag' filtering.

`--combined` for a similar effect as LTO, compiling reggae and its dub library dependency to a single object file.

PR #3753, opened 2021-06-08
Build druntime and Phobos with --enable-cross-module-inlining.

JohanEngelen:
This should give maximum performance benefit (it would have prevented the SHA performance degradation fixed by ldc-developers/druntime#199), at the cost of longer compile time (which should not be an issue as the standard library is only built by a fraction of the users).
As an extra plus, this automatically tests the flag on a larger codebase, in preparation of always enabling the flag at -O3.

@kinke what do you think?


kinke:
I think this is high risk. At work, we've recently had troubles because of it; IIRC, the problem with `= __FILE__` default template params.

Issue #2782, opened 2018-07-18
Non matching mangled name for C functions using structs


jacob-carlborg:
Example:

```d
module main;

struct Foo
{
    int a;
}

extern (C) void bar(Foo);
```

```d
module foo;

struct Foo
{
    int a;
}

extern (C) void bar(Foo);
```

```
$ ldc2 main.d foo.d
main.d(8): Error: Function type does not match previously declared function with the same mangled name: bar
```

C functions doesn't have a mangling. I suspect the mangled name that would be used for a D function for the `Foo` struct is used.

kinke:
This is a duplicate of long-standing https://github.com/ldc-developers/ldc/issues/1020.

A quick look at the IR shows the problem - 2 equivalent `Foo` structs in different D modules:
```LLVM
declare void @bar(%main.Foo) #0
declare void @bar(%foo.Foo) #0
```

This is only a problem with (mostly default/enforced) `-singleobj` though (i.e., `-c` should work), where code from all D modules is emitted into a single IR module (object file), and we can only have one IR function for a particular (mangled) name.

We make sure the IR types match if there are multiple declarations (and only emit the first definition in that case). If we didn't, we'd have to adapt the call sites, e.g., repainting IR `foo.Foo` call args as `main.Foo` ones.

jacob-carlborg:
Wouldn't the proper way to do this be to not look at the actual name of the struct but instead its members? Basically building up a mangled/IR name combining the mangled names of the members, recursively.

kinke:
That's definitely something I haven't thought of before, but only imagining the length of some of these names (need to take alignments into account as well, unions...) makes my head ache. I look at IR a lot, and having type names >1k isn't bearable IMO.

jacob-carlborg:
Yeah, the might be a problem. But as I mentioned in the other issue https://github.com/ldc-developers/ldc/issues/1020 it's easy to detect if a class/struct is `extern(D)` or some other linkage.

kinke:
> it's easy to detect if a class/struct is extern(D) or some other linkage.

That has just been added with 2.081 AFAIK. But as your example shows, people don't expect having to declare a struct `extern(C[++])` just because it's used in some C[++] function.

We could keep a map of struct identifier (excl. module prefix) to IR types, per IR module. When trying to emit a new IR type, we could check if its IR layout is equivalent to an existing type with the same identifier, and use that existing one instead.

JohanEngelen:
[...]

PR #3757, opened 2021-06-10
CMake: Compile shared Phobos libs to a single object file

kinke:
(Only) when compiling them all-at-once anyway; primarily for implicit cross-module inlining across the same library.

This won't work as long as there are multiple conflicting declarations across a library's modules - issue #2782.


kinke:
Argh, it's not just conflicting declarations in the libraries, but also mismatches regarding the (simplified) compiler forward declarations...

kinke:
Phobos actually already works, what a nice surprise. :)

Issue #3737, opened 2021-05-24
Add extra allocator lib when linking LDC


JohanEngelen:
At weka we've found that using mimalloc (https://microsoft.github.io/mimalloc/index.html) gives a huge LDC performance boost. We can use LD_PRELOAD, but I'd rather build this into the LDC executable. We can do this by linking an extra object file (`mimalloc.o`) when linking LDC.
It's not clear to me how best to add this as an option to our CMake script. A generalized option to add object files to the link line would be sufficient).
Something like this in `BuildDExecutable.cmake`:
```diff
    # Link to an executable.
    if(LDC_LINK_MANUALLY)
-        add_executable(${target_name} ${object_files})"
+        add_executable(${target_name} "${user_supplied_object_files};${object_files})"
```

Shall I add a general cmake option for this?
`cmake -DLINK_WITH_EXTRA_OBJECT_FILES="/usr/local/lib/mimalloc-1.7/mimalloc.o"`
Or a (mi)malloc-specific option? (other allocators use the same trick)

kinke:
Might be feasible by setting `D_LINKER_ARGS` instead of detecting them; those flags are the last ones being appended AFAICT, so if the order is still okay, it should work.

JohanEngelen:
Adding it to `D_LINKER_ARGS` fails building `ldmd2` because first `libldmd` is created (with the object file added) and then again the object file is added when linking `ldmd2`.  But it does work to build `ldc2` that way (with correct override of malloc).

JohanEngelen:
Perhaps add a `D_EXTRA_LINKER_ARGS` option that is only used for the final links?

JohanEngelen:
I'm at a loss. For some reason, `D_LINKER_ARGS` is _partly_ copied twice on the linker line for `ldmd2` and not for `ldc2`. So LDC links correctly and LDMD doesn't. I've tried a bunch of hacks to our CMake script but can't get it to behave different.

```
[125/283] Linking CXX executable bin/ldmd2
[...]

PR #3758, opened 2021-06-11
CMake: Add `ALTERNATIVE_MALLOC_O` option for user to link LDC with a malloc-overriding allocator like mimalloc.o

JohanEngelen:
Fixes https://github.com/ldc-developers/ldc/issues/3737

Usage example:
```
cmake -DALTERNATIVE_MALLOC_O=/usr/local/lib/mimalloc-1.7/mimalloc.o
```

Especially codegen performance is greatly improved with mimalloc (https://microsoft.github.io/mimalloc/index.html), around ~30% on a large codebase.


kinke:
Building it is apparently a breeze, so I'll look into integrating mimalloc for our Azure packages - at least the non-cross-compile builds (as e.g. support for Apple M1 is apparently still experimental in the last version).

PR #3759, opened 2021-06-12
Azure CI: Use mimalloc for Linux and macOS packages

kinke:


kinke:
For Windows, the trade-off is that we'd have to link `ldc2.exe` against the *shared* MSVC++ runtime, meaning that using it would newly require an installed MSVC++ runtime (not sure if the one from VS 2015 would suffice (which might come with Win10), or whether it'd have to be the 2019 one). I'm a bit sceptical.

kinke:
@JohanEngelen: Have you tried it on macOS as well? Does it need some extra CMake option there?

JohanEngelen:
I've only tried it on Linux. I know that Orip at Weka also had similar problems on macOS (`Aborting from rt/sections_elf_shared.d(195) Could not determine TLS range`).  The documentation, https://microsoft.github.io/mimalloc/overrides.html, is not so clear about whether it should work on macOS or not  (`static.c` does look like it supports macOS though).

jacob-carlborg:
>  I know that Orip at Weka also had similar problems on macOS (Aborting from rt/sections_elf_shared.d(195) Could not determine TLS range)

I'm not sure if it's related or not, but implementation to get the TLS range on Darwin is using `malloc_size`. If `malloc_size` is not overridden (according to https://microsoft.github.io/mimalloc/overrides.html, it looks like it's not), it might cause problems.

kinke:
[It's apparently working fine now with the `MI_OSX_ZONE` CMake option, but I have no idea what it does.]

JohanEngelen:
Using MI_OSX_ZONE seems correct: https://github.com/microsoft/mimalloc/commit/a83bca72b3f98b55c564713176c40e44698e2c43

`malloc_size` is overridden: https://github.com/microsoft/mimalloc/blob/752594e76423526e108413731518a26e3322b9ca/src/alloc-override.c#L175

kinke:
@p0nce: It'd be awesome if you could download the macOS arm64 artifact and give it a quick whirl (any little module should do), to check whether they really fixed the M1 issues in the latest version.

p0nce:
@kinke can do before 23h CEST , I'm not at home right now

kinke:
Perfect, thx, no hurry.

p0nce:
- both macOS artifacts (x86_64 and arm64) pass intel-intrinsics test
- both macOS artifacts (x86_64 and arm64) can build a working hello world
- for some reason I have trouble building a shared library with the arm64 native LDC, but the x86_64 works for both target
So all in all I'd say no problem with these artifacts.

kinke:
Thx a lot for testing Guillaume.

> for some reason I have trouble building a shared library with the arm64 native LDC

Care to elaborate a bit? I can't imagine this having anything to do with the allocator here (and would expect the same issue with v1.27.0-beta1, maybe related to LLVM 12), but I'm still interested.

p0nce:
Yes, probably unrelated.
My build log finish with: https://pastebin.com/raw/xk73aLdy
There are mentions of x86_64 even though everything was built with `-mtriple=arm64-apple-macos`
No such problem with the cross-compiler artifact.

kinke:
Okay that looks like your Apple clang for linking somehow defaults to x64 here, probably requiring an explicit `-Xcc=-target -Xcc=arm64-apple-macos` in the LDC cmdline.

Issue #3690, opened 2021-03-28
v1.25+: -linkonce-templates miscompiles wrt. nested lambdas


kinke:
I haven't been able to nicely reduce this yet, but host compilers v1.25+ miscompile the v2.096 frontend with `-linkonce-templates`. The `&fad` address in the lambda in https://github.com/dlang/dmd/blob/9191fdc12266cd3191d9d51c44e5a9d98b583745/src/dmd/root/file.d#L292 is null when compiling with `-linkonce-templates` (non-optimized too), so it looks like the context pointer is lost somewhere in the chain of lambdas, probably related to semantic order.

The symptom only shows up on Windows because it's some Windows-specific code construct newly added in v2.096.0. E.g., trying to create any .di file via `-Hd=.` makes the compiler segfault.

kinke:
I think I found the cause in the IR:
```diff
-define weak_odr x86_vectorcallcc i32 @"\01_D3dmd4root8filename__T16extendedPathThenSQBoQBn4file4File6updateFNbPxaxAvZ9__lambda6ZQClMFNbAxaZi"(i8* nonnull %.nest_arg, { i64, i8* } %path_arg) #1 comdat {
+define linkonce_odr x86_vectorcallcc i32 @"\01_D3dmd4root8filename__T16extendedPathThenSQBoQBn4file4File6updateFNbPxaxAvZ9__lambda6ZQClMFNbAxaZi"(i8* nonnull %.nest_arg, { i64, i8* } %path_arg) #1 comdat {
   %path = alloca { i64, i8* }, align 8            ; [#uses = 3, size/byte = 16]
   store { i64, i8* } %path_arg, { i64, i8* }* %path, align 8
   %1 = getelementptr inbounds { i64, i8* }, { i64, i8* }* %path, i32 0, i32 0 ; [#uses = 1, type = i64*]
   %.len = load i64, i64* %1, align 8              ; [#uses = 1]
   %2 = icmp ne i64 %.len, 0                       ; [#uses = 1]
   %3 = icmp eq i1 %2, false                       ; [#uses = 1]
   br i1 %3, label %if, label %endif

 if:                                               ; preds = %0
   %4 = call x86_vectorcallcc i32 @"\01_D3dmd4root4file4File6updateFNbPxaxAvZ__T9__lambda6TAuZQoMFNbNiQlZi"(i8* nonnull %.nest_arg, { i64, i16* } zeroinitializer) #1 ; [#uses = 1]
[...]

PR #3763, opened 2021-06-14
Windows: Add -dllimport CLI option for more selective implicit dllimports

kinke:
* Newly require `-link-defaultlib-shared` for implicit dllimport.
  E.g., this enables to compile druntime DLL with `-fvisibility=public` for pure exports and no (local) imports (such as builtin TypeInfos).
* `-link-defaultlib-shared` alone now only implicitly imports symbols from druntime/Phobos.
  This simplifies building complex DLLs linked against a bunch of static libs (dub only supports static lib dependencies!); the static
  libs don't need to be compiled with `-fvisibility=public` anymore (if the DLL itself isn't either), `-link-defaultlib-shared` is sufficient.
  This is mainly useful for existing DLLs with explicit exports, to make them link against *shared* druntime/Phobos and so end up with a single druntime/Phobos for the whole process.

### DIFF
--- a/dmd/globals.d
+++ b/dmd/globals.d
@@ -109,6 +109,14 @@ enum FeatureState : byte
     enabled = 1    /// Specified as `-preview=`
 }
 
+version (IN_LLVM)
+enum DLLImport : byte
+{
+    none,
+    defaultLibsOnly, // only symbols from druntime/Phobos
+    all
+}
+
 // Put command line switches in here
 extern (C++) struct Param
 {
@@ -304,8 +312,8 @@ version (IN_LLVM)
     bool linkonceTemplates; // -linkonce-templates
 
     // Windows-specific:
-    bool dllexport; // dllexport ~all defined symbols?
-    bool dllimport; // dllimport data symbols not defined in any root module?
+    bool dllexport;      // dllexport ~all defined symbols?
+    DLLImport dllimport; // dllimport data symbols not defined in any root module?
 } // IN_LLVM
 }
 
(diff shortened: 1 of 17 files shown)


kinke:
With this proposal, the option of only dllimporting (~all external D globals) without own dllexports (previous `-link-defaultlib-shared`) is lost. This might be a bit of a hassle (I guess mainly for executables linked against non-defaultlib DLLs).

Until now, I've tried to handle these implicit dllimports without extra cmdline option, to keep things reasonably simple. Maybe an explicit option is better; having to use `-fvisibility=public -link-defaultlib-shared` for betterC executables (to dllimport external globals) isn't really intuitive. ;)

kinke:
I've tested this for an executable with ~150 dub dependencies (incl. big ones like vibe-d; all static libs), wanting to link it against shared druntime/Phobos - by enforcing `-link-defaultlib-shared` as implicit compile flag for all packages via `reggae --dflags=-link-defaultlib-shared`.

This led to 21 undefined symbols; some TLS variables (`std.getopt` - can be worked around in Phobos) and, annoyingly, culled templates (pre-instantiated in Phobos). `-linkonce-templates` to the rescue, and down to 8 undefined symbols.

kinke:
After some fixes, down to 1 undefined symbol - there are 3rd-party packages in `std`, e.g., `std.io` (https://github.com/MartinNowak/io/)...

kinke:
No more linker errors, but hitting https://github.com/ldc-developers/ldc/issues/3690 apparently.

PR #3766, opened 2021-06-18
Fix DtoCreateNestedContextType() for -linkonce-templates

### LINKED ISSUE #3690: v1.25+: -linkonce-templates miscompiles wrt. nested lambdas

kinke:
Fixes #3690.

### DIFF
--- a/.azure-pipelines/6-package.yml
+++ b/.azure-pipelines/6-package.yml
@@ -4,8 +4,9 @@
 # Required env vars:
 # - ARCH
 # - CI_OS
-# - EXTRA_CMAKE_FLAGS_CROSS (optional)
 # - CROSS_TRIPLE (optional)
+# - DFLAGS (optional)
+# - EXTRA_CMAKE_FLAGS_CROSS (optional)
 
 steps:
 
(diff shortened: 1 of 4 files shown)


kinke:
I can confirm that this fixes the issue for the executable mentioned in https://github.com/ldc-developers/ldc/pull/3763#issuecomment-861527116. So both `-link-defaultlib-shared` and `-linkonce-templates` seem to work fine for a complex project on Windows now ('seem' because I haven't extensively tested it yet).

thewilsonator (approved):

kinke:
I haven't fully analyzed what happened before this change (generating the `-vv` output took something like half an hour on a fast box (as the issue only showed up when compiling the front-end all-at-once), and grepping an absolute horror due to https://issues.dlang.org/show_bug.cgi?id=21387), but seeing that `irFunc.nestedContextCreated` is set right before recursively walking up the parents hierarchy seemed dangerous wrt. reentrancy for the original function.

Crafting an according testcase would probably take *significant* effort.

PR #3768, opened 2021-06-20
README: Add latest release badge (incl. prereleases) and update notes about prebuilt packages

kinke:
Incl. emphasizing that the package managers are community-maintained, I'm tired of reading the same questions again and again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,28 +33,28 @@ Installation
 
 ### From a pre-built package
 
-Portable stand-alone binary builds for common platforms (incl. Linux,
-macOS and Windows) are available at the
-[GitHub release page](https://github.com/ldc-developers/ldc/releases).
+Portable stand-alone binary builds (and a Windows installer) for common
+platforms (incl. Linux, macOS, Windows, FreeBSD and Android) are available
+at the [GitHub release page](https://github.com/ldc-developers/ldc/releases).
+For Windows, the [Visual D installer](https://rainers.github.io/visuald/visuald/StartPage.html)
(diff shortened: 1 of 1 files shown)


Geod24:
Do you also want to include a link to `setup-dlang` for CIs ?

kinke:
Hmm - it'd be a single CI service only, but my main point would be that the target audience is different - this readme is primarily for first-time users, not guys already familiar with D and setting up CI for their D repos.

JohanEngelen:
> Incl. emphasizing that the package managers are community-maintained, I'm tired of reading the same questions again and again.

to avoid variable interpretation of "community-maintained", perhaps "are not maintained by the LDC project but by the larger D language community" would work even better?

kinke:
I think it's a sufficiently self-describing and established term, some Google results:
* https://github.com/ziglibs/repository: `This is one community-maintained repository of zig packages.`
* https://github.com/openwrt/packages: `This is the OpenWrt "packages"-feed containing community-maintained build scripts...`
* https://community.chocolatey.org/packages: `There are 8513 Community Maintained Packages`

kinke:
@Geod24: I've noticed Alpine is missing - is `apk add ldc` sufficient?

Geod24:
Yes. `edge` is usually updated pretty quickly, stable releases get whatever's current when the release is made.

kinke:
Thx; wrt. dub, yeah that seems to be pretty standard for distro packages (and makes sense).

Issue #3773, opened 2021-06-27
Hacky init symbol declaration can cause undefined symbols with -linkonce-templates


kinke:
https://github.com/dlang/druntime/pull/3467 (landed in v2.097) accesses struct init symbols via `pragma(mangle)` - while nice to use it directly, it's an issue for templated structs with `-linkonce-templates`, because these init symbols are defined as `linkonce_odr` in that case (in the owning CU and each other CU referencing it *regularly*).

There'd be a simple workaround for this by emitting them as `weak_odr` with `-linkonce-templates` too, but I think a proper solution à la `__traits(initSymbol, <aggregate type>)` (returning null for zero-initialized structs) would be more appropriate, allowing to access class init symbols directly too without `TypeInfo_Class.initializer()` indirection (and e.g. so also available for `-betterC`).

kinke:
Resolved by #3774 and https://github.com/ldc-developers/druntime/pull/201.

PR #3774, opened 2021-06-27
Add __traits(initSymbol, <aggregate type>)

kinke:
A clean way of directly accessing struct and class init symbols. It yields a `const(void[])` slice; the length reflecting the struct/class instance size, and the pointer either pointing to the init symbol or being null for zero-initialized structs.

Paves the way for resolving #3773 in druntime as well as accessing class init symbols without `TypeInfo_Class.initializer()` indirection, and so with `-betterC` as well.

### DIFF
--- a/dmd/declaration.d
+++ b/dmd/declaration.d
@@ -1711,6 +1711,19 @@ version (IN_LLVM)
  */
 extern (C++) final class SymbolDeclaration : Declaration
 {
+version (IN_LLVM)
+{
+    AggregateDeclaration dsym;
+
+    extern (D) this(const ref Loc loc, AggregateDeclaration dsym)
+    {
+        super(loc, dsym.ident);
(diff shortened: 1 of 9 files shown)


thewilsonator:
Please do submit this upstream

kinke:
I have 0 interest to tackle the DMD glue stuff but it's all prepared for simple upstreaming (that's why it's not in `ldctraits.d`).

thewilsonator:
Noted, added upstream PR to reduce frontend diff

JohanEngelen:
Nice, much better than the pragma hack indeed.

kinke:
Reverted from `immutable` to `const`, as class init symbols aren't really immutable (monitor field, plus druntime emplacing some `Error`s in their init symbols to prevent allocations...).

PR #3775, opened 2021-06-27
druntime: Prefer __traits(initSymbol) over TypeInfo.initializer()

kinke:
Testing #3774 with an adapted druntime (https://github.com/ldc-developers/druntime/pull/201).


kinke:
Confirmed to fix #3773, making `-linkonce-templates -O` work for a huge project.

PR #3776, opened 2021-07-02
Travis CI: fix cache directory name

kubo39:


.travis.yml
```diff
@@ -13,7 +13,7 @@ matrix:
 
 cache:
   directories:
-    - llvm-11.0.1
+    - llvm-12.0.0
```
kinke, line 16: Ah thanks, forgot this, but it's currently dead here anway. Please try `llvm-$LLVM_VERSION` to prevent having to update this each time.

PR #3779, opened 2021-07-05
Fix #3778 - in-place construct <temporary>.ctor(args)

### LINKED ISSUE #3778: RVO not done when return statement expression is a constructor call
A variant of https://godbolt.org/z/e9xEhb6xa from GDC bugzilla report https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101273
```
struct S
{
    int x;
    S* impl; 
    this(int x)
    {
        this.x = x;
        this.impl = &this;
    }
    ~this() { }
}

S makeS()
{
    return S(2);
}

S nrvo()
{
    S ret = makeS();
    return ret;
}

S rvo()
{
    return makeS();
}

void main()
{
    auto nrvo = nrvo();
    assert(&nrvo is nrvo.impl);

    auto rvo = rvo();
    assert(&rvo is rvo.impl);
}
```

What happens is that - because of the presence of a constructor - a temporary is created by the DMD front-end in the form of:
```
S makeS()
{
    return ((S __slS58 = S(0);) , __slS58).this(2);
}
```

The fix was to detect the declaration of a new temporary _inside_ a ReturnStatement and swap the temporary for the hidden return slot/pointer.

kinke:


gen/toir.cpp
```diff
@@ -2792,74 +2817,56 @@ bool toInPlaceConstruction(DLValue *lhs, Expression *rhs) {
...
       Logger::println("success, in-place-constructing array literal");
       initializeArrayLiteral(gIR, al, DtoLVal(lhs));
       return true;
     }
   }
-
-  // vector literals too
-  if (auto ve = rhs->isVectorExp()) {
+  // and vector literals
+  else if (auto ve = rhs->isVectorExp()) {
```
JohanEngelen, line 2860: Because there is not a guaranteed return statement in the previous else-if (line 2852), this is slight change in behavior. Was that intentional?

PR #3782, opened 2021-07-07
CMake: Export zlib symbols from Phobos DLL on Windows

kinke:


kinke:
This is needed because some of these C symbols are directly exposed (declared) in `etc.c.zlib`.

PR #3783, opened 2021-07-07
Make sure special struct TypeInfo members are semantically analyzed before emitting the TypeInfo

kinke:
As there are apparently cases where the struct hasn't had semantic3 yet, at least with `-linkonce-templates`. This prevents erroneous `xopCmp` etc. (failing to compile) from being referenced in the TypeInfo (=> linker errors) / being analyzed for `-linkonce-templates` codegen. `semanticTypeInfoMembers()` replaces them by `object._xopCmp()` etc. if compilation fails.

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -602,7 +602,6 @@ void DtoDeclareFunction(FuncDeclaration *fdecl, const bool willDefine) {
     Logger::println("Function is inside a linkonce_odr template, will be "
                     "defined after declaration.");
     if (fdecl->semanticRun < PASSsemantic3done) {
-      // this can e.g. happen for special __xtoHash member functions
       Logger::println("Function hasn't had sema3 run yet, running it now.");
       const bool semaSuccess = fdecl->functionSemantic3();
       (void)semaSuccess;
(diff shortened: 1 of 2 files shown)


gen/functions.cpp
```diff
@@ -601,14 +601,7 @@ void DtoDeclareFunction(FuncDeclaration *fdecl, const bool willDefine) {
...
                     "defined after declaration.");
-    if (fdecl->semanticRun < PASSsemantic3done) {
-      // this can e.g. happen for special __xtoHash member functions
-      Logger::println("Function hasn't had sema3 run yet, running it now.");
-      const bool semaSuccess = fdecl->functionSemantic3();
-      (void)semaSuccess;
-      assert(semaSuccess);
-      Module::runDeferredSemantic3();
-    }
+    assert(fdecl->semanticRun == PASSsemantic3done);
```
kinke, line 604: No idea if that's too optimistic - works fine for the project exposing the original issue though.

kinke: It definitely was, CI immediately errored out...

kinke:
Sorry, can't be bothered to spend hours on setting up a dustmite run for a nicely reduced testcase...

PR #3797, opened 2021-07-27
Improve ftime-trace implementation.

JohanEngelen:
- Rewrite ftime-trace to our own implementatation instead of using LLVM's time trace code. The disadvantage is that this removes LLVM's work from the trace (optimization), but has the large benefit of being able to tailor the tracing output to our needs.
- Add memory tracing to ftime-trace (not possible with LLVM's implementation)
- Do not output the sum for each "category"/named string. This causes the LLVM output to be _very_ long, because we put more information in each time segment name. Tooling that processes the time trace output can do this summing itself (i.e. Tracy), and makes the time trace much more pleasant to view in trace viewers.
- Add source file location info to time trace (Tracy understands it) (not possible with LLVM's implementation)


kinke:
Sounds nice! Excluding the LLVM stats seems fine to me, after all it's mostly about introspecting time spent in frontend and glue layer.

Wrt. memory stats, AFAICT you're tracking frontend memory consumption only. Not sure if glue layer and LLVM consumption is really relevant, but at least with mimalloc, we could probably get the overall heap stats via [mi_stats_print_out()](https://microsoft.github.io/mimalloc/group__extended.html#ga537f13b299ddf801e49a5a94fde02c79) or so.

JohanEngelen:
> Sounds nice! Excluding the LLVM stats seems fine to me, after all it's mostly about introspecting time spent in frontend and glue layer.
> 
> Wrt. memory stats, AFAICT you're tracking frontend memory consumption only. Not sure if glue layer and LLVM consumption is really relevant, but at least with mimalloc, we could probably get the overall heap stats via [mi_stats_print_out()](https://microsoft.github.io/mimalloc/group__extended.html#ga537f13b299ddf801e49a5a94fde02c79) or so.

Memory stats definitely needs more work, I'll leave that for the future. I'm actually not sure if I'm interpreting the GC stats correctly. It's like you say, this does not track the C++ memory usage. But I think the frontend usage is a lot more (CTFE...) than that.
We can also think of other useful stats (like number of template instantiations?). As for timing regions, CTFE is not tracked well I think.

JohanEngelen:
Apparently `gdmd` does not define `D_InlineAsm_X86_64` ? Very strange. But OK, I've implemented a fallback option, with only time in seconds as resolution. Based on https://github.com/D-Programming-GDC/gdc/blob/e5a495a8490b4eb3acb78448f79920e2ee8a9ca9/libphobos/libdruntime/rt/util/random.d#L17-L63

kinke:
[GDC doesn't support DMD-style inline asm (which that predefined version stands for).]

JohanEngelen:
And now I need a compile-time (!) `wait(1 second)` :D

dmd/root/rmem.d
```diff
@@ -157,6 +157,7 @@ enum CHUNK_SIZE = (256 * 4096 - 64);
 
 __gshared size_t heapleft = 0;
 __gshared void* heapp;
+version(IN_LLVM) __gshared size_t heaptotal = 0; // Total amount of memory allocated using malloc
```
kinke, line 160: Please use the `version (IN_LLVM)` convention (grepping).

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
+void timeTraceProfilerBegin(const(char)* name_ptr, const(char)* detail_ptr, Loc loc)
+{
+    import dmd.root.rmem;
+    import core.stdc.string : strdup;
+
+    assert(timeTraceProfiler);
+
+    // `loc` contains a pointer to a string, so we need to duplicate that string too.
+    if (loc.filename)
+        loc.filename = strdup(loc.filename);
```
kinke, line 122: [I understand this is the fail-safe variant; IIRC, `loc.filename`s are never cleaned up, and name/detail strings might persist too (e.g., `toChars()` is cached IIRC), so there might be potential for less copying.]

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
 
-static if (LLVM_VERSION_MAJOR >= 10)
+    timeTraceProfiler.beginScope(mem.xstrdup(name_ptr).toDString(),
+                                 mem.xstrdup(detail_ptr).toDString(), loc);
```
kinke, line 125: `rmem.xarraydup(name_ptr.toDString())` uses a single strlen.

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
+            QueryPerformanceFrequency(&freq);
+            time_scale = freq / 1_000_000;
+        }
+
+        timer_t time;
+        QueryPerformanceCounter(&time);
+        this.beginningOfTime = time / time_scale;
+    }
+
+    void beginScope(const(char)[] name, const(char)[] details, Loc loc)
```
kinke, line 192: `const ref Loc`?

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
+        static if (is(typeof(&QueryPerformanceFrequency)))
+        {
+            timer_t freq;
+            QueryPerformanceFrequency(&freq);
+            time_scale = freq / 1_000_000;
+        }
+
+        timer_t time;
+        QueryPerformanceCounter(&time);
+        this.beginningOfTime = time / time_scale;
```
kinke, line 189: This `timer_t bla; QueryPerformanceCounter(&bla); return bla / time_scale` seems suited for extraction as separate function, as it's used at least 3 times.

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
         {
-            extern __gshared TimeTraceProfiler* TimeTraceProfilerInstance;
-            auto getTimeTraceProfilerInstance() { return TimeTraceProfilerInstance; }
+            static if (__VERSION__ >= 2085)
+            {
+                import core.memory : GC;
+                auto stats = GC.stats();
+                auto profileStats = GC.profileStats();
+
+                counters.allocatedMemory = stats.allocatedInCurrentThread;
```
kinke, line 236: I use: https://github.com/ldc-developers/ldc/blob/30bf997718e06a1d1d1bd174abbfbba8ae47acfd/dmd/mars.d#L740

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
-        this(string name, lazy string detail) {
-            if (timeTraceProfilerEnabled())
-                timeTraceProfilerBegin(name.length, name.ptr, detail.length, detail.ptr);
+    }
+
+    void writeDurationEvents(OutBuffer* buf)
+    {
+        // {"ph":"X","name": "Sema1: somename","ts":111,"dur":222,"loc":"filename.d:123","args": {"detail": "something", "loc":"filename.d:123"},"pid":0,"tid":0}
+
+        void writeLocation(Loc loc)
```
kinke, line 322: const ref?

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
+        buf.write(`"},"cat":"","name":"thread_name",`);
+        buf.write(pidtid_string);
+        buf.write("},\n");
+    }
+
+    void writeCounterEvents(OutBuffer* buf)
+    {
+        // {"ph":"C","name":"ctr","ts":111,"args": {"Allocated_Memory_bytes":  0, "hello":  0}},
+
+        foreach (event; counterEvents)
```
kinke, line 302: const ref

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
-        this(string name, lazy string detail) { }
+alias long timer_t;
 
-        ~this() { }
+version (Windows)
+{
+    extern (Windows)
+    {
+        export int QueryPerformanceCounter(timer_t *);
+        export int QueryPerformanceFrequency(timer_t *);
```
kinke, line 463: Hmm, I wouldn't have expected this to work at all (enforcing `dllimport`, even though we link against the static MSVCRT). It's in `rt.trace`, oh well.

driver/timetrace.d
```diff
@@ -8,81 +8,521 @@
...
+                ret                     ;
+            }
+        }
+        else
+        {
+            // Fallback to libc timestamp
+            import core.stdc.time : time;
+            auto t = time(null); // unit is seconds
+            *ctr = cast(timer_t) t * 1_000_000_000; // we expect nanoseconds
+        }
```
kinke, line 526: It might be worth considering switching to C++ `std::chrono::steady_clock`, seems way simpler and surely has a much nicer fallback.

kinke: Or `core.time.MonoTime`, I guess it's been there for long enough (and you're already depending on `std.conv` etc.).

driver/timetrace.d
JohanEngelen: Let's optimize later if needed.

driver/timetrace.d
JohanEngelen: Loc is small (size of two pointers), so I chose to copy it, because it's probably faster.

driver/timetrace.d
JohanEngelen: Loc is small (size of two pointers), so I chose to copy it, because it's probably faster.

driver/timetrace.d
JohanEngelen: thnx!

driver/timetrace.d
JohanEngelen: OK, let's use that here too then :)

JohanEngelen:
Good to merge?

thewilsonator (approved):
LGTM apart from the `const ref` on the `Loc` which is only bad because for the frontend source uses `const ref Loc` everywhere and not using it is surprising.

driver/timetrace.d
kinke: [It works because these functions are OS functions, not CRT ones, and the WinAPI consists of DLLs only.]

driver/timetrace.d
kinke: Fine by me.

driver/timetrace.d
kinke: It doesn't really matter, hence the `?`; it was primarily for consistency with the frontend, which *mostly* uses `const ref Loc` but isn't totally consistent either.

driver/timetrace.d
kinke: I actually meant using that in general, not just as fallback. If `core.time.MonoTime` is indeed significantly slower for x86[_64] and AArch64 than the LLVM intrinsic / the asm, we should probably optimize that in druntime directly.

On Windows, `core.time.MonoTime` uses the same WinAPI functions, thus it shouldn't matter there. For Darwin, it's using `mach_absolute_time()`, for other Posixen, `clock_gettime()`.

kinke: Even more convinced about this after looking at https://stackoverflow.com/questions/6498972/faster-equivalent-of-gettimeofday/13096917#13096917 (`core.time.MonoTime` defaults to `CLOCK_MONOTONIC` AFAICT) - almost as fast as RDTSC, but avoiding its pitfalls. [readcyclecounter](https://llvm.org/docs/LangRef.html#llvm-readcyclecounter-intrinsic) doesn't seem to guarantee any frequency, so the assumed 1ns might likely be wrong, plus they warn about rather quick overflows (9 seconds on Alpha).

JohanEngelen: Just tested this on a 27sec sema-only compilation, 15MB timetrace. Did not see a difference in time it takes. So I gather that the perf is OK :)  Any perf improvements should come from improving druntime instead, fully agree.

JohanEngelen:
Conversion to core.time.MonoTime complete, ready now.

kinke (approved):
LGTM, thanks, also nice to have less integer divisions now. I guess the test will remain a bit brittle (OS context switches etc.), but we'll see.

kinke:
@JohanEngelen: I'm about to merge v2.097.2 and release LDC v1.27.1, mainly to re-add an AArch64 Linux package. Including this might be nice, so please merge if you agree.

JohanEngelen:
Merged. Risk of introducing badness to 1.27 is very low.

PR #3799, opened 2021-07-30
Merge upstream stable and bump bundled dub & dlang tools

kinke:


kinke:
As preparation for v1.27 final; I've run out of patience waiting for DMD v2.097.2.

Issue #3806, opened 2021-08-07
core.atomics.pause is not implemented


KytoDragon:
Right now, the pause function in core.atomics is not implemented in LDC:

```
void pause() pure nothrow @nogc @trusted
{
    // LDC: TODO?
}

```

I don't know if there is a LLVM-intrinsic for that, but it can be done via LLVM-ASM:

```
void pause() pure nothrow @nogc @trusted
{
    __asm("pause","");
}
```

PR #3807, opened 2021-08-07
Fix issue #3806 - Implement core.atomic.pause() for some archs

kinke:


JohanEngelen:
From some internet searching, `yield` seems to be the ARM instruction corresponding to `pause` on x86. `pause` also works for MIPS. See code here: https://chromium.googlesource.com/chromium/src/third_party/WebKit/Source/wtf/+/823d62cdecdbd5f161634177e130e5ac01eb7b48/SpinLock.cpp
```C
#if CPU(X86_64) || CPU(X86)
#define YIELD_PROCESSOR __asm__ __volatile__("pause")
#elif CPU(ARM) || CPU(ARM64)
#define YIELD_PROCESSOR __asm__ __volatile__("yield")
#elif CPU(MIPS)
// The MIPS32 docs state that the PAUSE instruction is a no-op on older
// architectures (first added in MIPS32r2). To avoid assembler errors when
// targeting pre-r2, we must encode the instruction manually.
#define YIELD_PROCESSOR __asm__ __volatile__(".word 0x00000140")
#elif CPU(MIPS64) && __mips_isa_rev >= 2
// Don't bother doing using .word here since r2 is the lowest supported mips64
// that Chromium supports.
#define YIELD_PROCESSOR __asm__ __volatile__("pause")
#endif
```

kinke:
Thx for looking it up. I came up with
```D
    void pause() pure nothrow @nogc @trusted
    {
        version (X86)
            enum inst = "pause";
        else version (X86_64)
            enum inst = "pause";
        else version (ARM)
            enum inst = "yield"; // apparently requires v6k+ (e.g., -mtriple=armv6k-linux-gnueabihf)
        else version (AArch64)
            enum inst = "yield";
        else version (MIPS32)
            enum inst = "pause"; // apparently requires ISA r2+ (e.g., -mcpu=mips32r2)
        else version (MIPS64)
            enum inst = "pause"; // apparently requires ISA r2+ (e.g., -mcpu=mips64r2)
        else
            enum inst = null; // TODO?

        static if (inst !is null)
            asm pure nothrow @nogc @trusted { (inst); }
    }
```
but see the according comments. Neither that internal function nor the `core.atomic.pause()` wrapper are templates, so this might easily lead to druntime compilation trouble for such archs, for probably extremely little benefit. The AArch64 addition seems trouble-free and trivial to add though.

kinke:
Nevermind, `__traits(targetHasFeature)` works nicely, tested manually.

JohanEngelen (approved):

Issue #3802, opened 2021-08-05
ICE comparing strings at compile time for mips64el-linux


jacob-carlborg:
The following code:

```d
enum a = "foo" == "bar";
```

Results in an ICE when targeting mips64el-linux:

```
 $ ldc2 --mtriple mips64el-linux main.d -c -betterC                                                                                                                                                                    10:02:10
Both operands to a binary operator are not of the same type!
  %26 = mul i64 %25, i32 1
Invalid bitcast
  %27 = bitcast i64 %26 to i32
Invalid bitcast
  %35 = bitcast i64 %34 to i32
in function _D4core8internal5array8equality__T8__equalsTaTaZQoFNaNbNiNeMxAaMxQeZb
LLVM ERROR: Broken function found, compilation aborted!
Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
0  ldc2                     0x0000000106ecfbb0 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 56
1  ldc2                     0x0000000106eceae0 llvm::sys::RunSignalHandlers() + 72
2  ldc2                     0x0000000106ed0234 SignalHandler(int) + 304
3  libsystem_platform.dylib 0x000000019440ac44 _sigtramp + 56
4  libsystem_pthread.dylib  0x00000001943bf43c pthread_kill + 292
5  libsystem_c.dylib        0x0000000194307460 abort + 104
6  ldc2                     0x0000000106e31518 llvm::report_fatal_error(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool) + 0
7  ldc2                     0x0000000106e31370 llvm::report_fatal_error(llvm::Twine const&, bool) + 0
8  ldc2                     0x0000000106cd5b1c void llvm::VerifierSupport::WriteTs<llvm::Instruction*, llvm::MDNode const*>(llvm::Instruction* const&, llvm::MDNode const* const&) + 0
9  ldc2                     0x0000000106c5af0c llvm::FPPassManager::runOnFunction(llvm::Function&) + 656
10 ldc2                     0x0000000106c5aa24 llvm::legacy::FunctionPassManagerImpl::run(llvm::Function&) + 76
11 ldc2                     0x0000000106c6074c llvm::legacy::FunctionPassManager::run(llvm::Function&) + 296
[...]

PR #3808, opened 2021-08-07
Fix issue #3802 - Fix size_t size for 32bit ABI on 64bit architectures.

JohanEngelen:
I don't know how to trigger the `_d_array_slice_copy` simplification by SimplifyDRuntimeCalls, so can't test that part of the PR.


kinke:
This triggers the LLVM pass (call optimized away with `-O`, *not* optimized with `-O -disable-d-passes`):
```D
void foo(int[] a, int[] b)
{
    a[] = b[];
}

auto bar()
{
    int[2] a = [1, 2];
    int[2] b = [3, 4];
    foo(a, b);
    return a;
}
```

kinke:
I took the liberty of extending the test, fixing the target C `long` size as suggested and predefining `D_X32`, to make this land in v1.27.1.

Issue #3795, opened 2021-07-21
Bootstrapping LDC 1.26 with GDC 9.3 results in LDC2 binary being linked dynamically with druntime and phobos.


dbankov-vmware:
While attempting to bootstrap LDC 1.26 with GDC 9.3 located at a custom location LDC build failed complaining that it couldn't load libgphobos.so.76. Investigating a bit revealed that the link parameters extracted at https://github.com/ldc-developers/ldc/blob/6de8bfde5aa5bc5b528e0771b576281063fcb473/cmake/Modules/ExtractDMDSystemLinker.cmake#L72 from GDMD link line and stored in D_LINKER_ARGS are later passed to the a C++ compiler at https://github.com/ldc-developers/ldc/blob/6de8bfde5aa5bc5b528e0771b576281063fcb473/cmake/Modules/BuildDExecutable.cmake#L77 not being prefixed with "-Wl,". This works fine for "-l.\*" and "-L.\*" arguments but not for "-B.\*" ones. The result is that a "-Bstatic" argument before "-lgphobos" and "-lgdruntime" arguments and a "-Bdynamic" argument after these are simply dropped resulting in both Phobos and DRuntime being linked dynamically with the LDC2 binary. Later during LDC build when LDC2 is used to build some D code it can't load Phobos and DRuntime from GDC 9.3 as it is located at custom location which is not indicated in any way in the environment used with LDC2 and there is no "rpath" in LDC2 pointing to these.

I was able to resolve this problem by prefixing each item from "D_LINKER_ARGS" with "-Wl," which unfortunately revealed another problem.

When Phobos and DRuntime are linked statically to LDC2, the linker complains that there are several duplicate symbols namely "_d_allocmemory", "_d_newclass", "_d_newitemT" and "_d_newitemiT" which appeared to be defined in https://github.com/ldc-developers/ldc/blob/master/dmd/root/rmem.d.

I was able to resolve the second problem by putting the mentioned functions within "version(LDC)" blocks (similar to other functions in this file) but am not sure if this is the correct solution.

I'll post a pull request with the solutions I came up with shortly.

kinke:
1. You can use `-DLDC_LINK_MANUALLY=OFF` in the CMake command-line to let the D host compiler perform the linking of the ldc executables, instead of going through the C++ compiler.
[...]

PR #3733, opened 2021-05-20
CI: Remove Shippable & move AArch64 package generation to Travis

kinke:
As the service is sadly no more.

PR #3810, opened 2021-08-11
Fix for 3795

dbankov-vmware:
Prepend each argument in D_LINKER_ARGS with "-Wl," when using GDMD.
The above is needed because the arguments reported by GDMD are what is
passed to COLLECT2 while D_LINKER_ARGS are later passed to CXX.
The problem with that is that without "-Wl," prefix the -Bstatic and
-Bdynamic arguments before and after -lgphobos and -libgdruntime are
dropped. And that is a problem because without these the produced LDC is
linked dynamically to libgphobos and libgdruntime so these have to be
available whereever LDC is used.

Once libgphobos and libgdruntime were linked statically several symbol
conflicts were revealed which are fixed by moving the offending symbols
within "version (LDC)" blocks.


dmd/root/rmem.d
```diff
@@ -220,9 +220,12 @@ static if (OVERRIDE_MEMALLOC)
     // That scheme is faster and comes with less memory overhead than using a
     // disabled GC alone.
 
-    extern (C) void* _d_allocmemory(size_t m_size) nothrow
+    version (LDC)
```
JohanEngelen, line 223: This is too restrictive if it only meant to fix building with GDC. (e.g. this now also changes the binary when compiling with DMD)
If GDC does not support this code, then lines 204-209 should be changed.

kinke: Yes, as I already said in https://github.com/ldc-developers/ldc/issues/3795#issuecomment-884513903. But that's upstream code, i.e., DMD would be affected by the same problem. Overriding the symbols seems to work fine with shared GDC druntime, only the static one seems problematic (no `@weak` for GDC @ibuclaw?). So I'm clearly against patching LDC here in this respect.

dbankov-vmware: Thanks a lot for the reviews!

If I understand correctly newer versions of GDC (I'm using GDC from GCC 9.3) should not have the duplicated symbols problem. So is there a way to figure out if GDC would work (e.g. check its version or something else) and only disable OVERRIDE_MEMALLOC for versions that are known not to work?

JohanEngelen: `__VERSION__` may be of help here. https://dlang.org/spec/lex.html#specialtokens

JohanEngelen: as @kinke says, please file the patch upstream (with dmd)  https://github.com/dlang/dmd/blob/master/src/dmd/root/rmem.d

dbankov-vmware: I'm sorry I hadn't realised that rmem.d is from DMD code base.
Thanks for pointing this as now I better understand @kinke's comment as well.
Unfortunately this means that I'd have to wait until the patch in DMD is consumed in LDC and only after that the change in ExtractDMDSystemLinker.cmake could go in.
Let me look a bit deeper into this to try to find other options if any.

dbankov-vmware: It seems both GDC versions that I have access to (9.3 and 11.2) fail with the duplicate symbols error so it isn't worth trying to disable OVERRIDE_MEMALLOC based on the GDC version (not to mention that __VERSION__ is replaced with 2076 by both GDC versions).

I was able to workaround the problem by passing "-Wl,--allow-multiple-definition" in LDC build script when using GDMD. Does this sound as a reasonable workaround?

If not another option seems to be to mark the corresponding symbols as "weak" in the GDC DRuntime implementation (but I haven't tried this yet). 

Finally there is the option to make it possible to control the OVERRIDE_MEMALLOC value through a command line option and have this change as a PR for DMD (I haven't tried this either). Once this is consumed in LDC we could modify LDC build script to pass this option to GDMD when LDC is bootstrapped with GDMD.

Please let me know if there is a better option to resolve this issue or which of the above is the preferred one.

kinke: > another option seems to be to mark the corresponding symbols as "weak" in the GDC DRuntime implementation

Yes, that would be the real solution. It's important to make the compiler really avoid the GC for performance and memory effectiveness [except when run with `-lowmem`], so straight-forwardly setting `OVERRIDE_MEMALLOC` to false would imply worse performance. `-Wl,--allow-multiple-definition` might be a sufficiently satisfying workaround with a probably similar effect to linking against the shared libs (=> GC functions not overridden in druntime/Phobos themselves I guess).

Another thing for gdmd would be the support of `-Xcc` for dmd/ldmd2 compatibility and thus being able to further simplify matters via `LDC_LINK_MANUALLY=OFF`.

If you just want to get things working now without waiting for all requirements to land, I recommend again https://github.com/ldc-developers/ldc/issues/3795#issuecomment-884519462. It's straight-forward, just invoke CMake twice; once with `D_COMPILER=gdmd`, then build, then another time in a separate dir with `D_COMPILER=path/to/just-built/ldmd2`. All official packages are built this way; e.g., see https://github.com/ldc-developers/ldc/pull/3733/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R57-R95.

dbankov-vmware: Thanks a lot for your feedback @kinke!

It seems to me that even if I fix the `LDC_LINK_MANUALLY=OFF` path by adding `-Xcc` support in GDMD I would face the duplicate symbols problem again so I'll leave this as is for now (I do plan to fix it but not right now).

Just to give some context. I'm integrating D compilers in our internal build infrastructure which produces components where we tend to expose the result of a single configure/build/install procedure per component and then we use what is exposed during the build of other components potentially on another machine and from a different location. This is why I hit the problem with LDC being dynamically linked with `libgphobos.so` and `libgdruntime.so` initially because GDC libraries are not part of the first stage LDC component deliverables thus when I tried to use this LDC to build the second stage LDC the first stage LDC couldn't even start on the machine where the second stage LDC was about to be built.

I may be able to tweak our build infrastructure to execute two configure/build/install steps for a single component but prefer to avoid this if I can. This is why I preferred to fix the static linkage problem and I focused on that because it seemed to be the original intention.

I'll check if marking the corresponding symbols as "weak" in GDC implementation of DRuntime solves the problem locally and if so will try to upstream the change in GDC.

Unfortunately even if GDC is fixed I'm not sure what would be a safe way to introduce the fix in LDC build scripts because they would have to behave differently based on the version of GDC used. Anyway I'll think about this when we reach that point.

Thanks a lot for yours and @JohanEngelen's support and looking forward achieving the next step for this fix.

kinke: Okay so the main problem is the inflexibility of the existing build infrastructure. If you really can't simply hook into it to prepend a manual CMake invocation + build for the 1st compiler (no installation necessary), there might be 2 other options:
* Install GDC for the 2nd stage too, just to have its shared libs available when running the host compiler from the 1st stage.
* Or hook into the post-installation of the 1st stage, adding the shared GDC libs and setting the rpath of the `ldc2` and `ldmd2` executables (e.g., via patchelf).

> It seems to me that even if I fix the LDC_LINK_MANUALLY=OFF path by adding -Xcc support in GDMD I would face the duplicate symbols problem again

Yes, but the main advantage is that you wouldn't have to touch LDC CMake stuff at all - you could e.g. simply specify host compiler flags via `-DD_COMPILER_FLAGS="<use static libs> -L--allow-multiple-definition"`.

dbankov-vmware: I already have a solution on our side. First it was what I posted initially then I updated it to include only the "-Wl," prefix and to pass `--allow-multiple-definition` and now that I marked the conflicting symbols as weak in gcc/src/libphobos/libdruntime/rt/lifetime.d (which worked as expected) I only need the "-Wl," change in LDC build scripts and to consume GDC with weak symbols.

I posted this PR because we prefer not to have patches on top of the upstream code which we use in our build infrastructure.

On the two other suggestions - before touching the build scripts I tried something similar to a combination of the two points. I consumed GCC for the second stage LDC build but in this case it is located in a different location (compared to when LDC stage one was built) so had to patch LDC/LDMD before starting with patchelf to be able to locate libgphobos.so and libgruntime.so. This worked but seemed too require too much hacking (not to mention that had to consume the patchelf component itself) and didn't seem robust enough. That's when I decided to try to figure out why I end up with this dynamic linkage in the first place. And that was because I knew that GDC from homebrew (linuxbrew) gcc 11 formulae produced binaries linked statically with Phobos and DRuntime by default.

On the "-Xcc" topic I tried adding -Xcc support in GDMD and hit another issue. GDMD calls GDC for the link step and while it is part of GCC it seems to not honour the -static-libgcc and -static-libstdc++ so LDC build fails with missing symbols from stdc++. I also tried calling gcc/g++ in GDMD but this time it didn't link Phobos and DRuntime (as these are not explicitly listed on the command line) so currently don't see a good way to implement "-Xcc" in a way to work exactly like "-Xcc" in LDMD or DMD.

I may play a bit more with "-Xcc" as indeed having both changes (weak symbols and -Xcc support that works with LDC build as is) seems to be a possible solution but may not have a lot more time for this task.

In the mean time I'll update this PR to include the "-Wl," change and to automatically pass "--allow-multiple-definition" when using GDMD. I don't expect this to be merged in LDC but prefer to have a working patch to be available in case someone faces the same issue.

kinke: > I consumed GCC for the second stage LDC build but in this case it is located in a different location (compared to when LDC stage one was built) so had to patch LDC/LDMD before start (with patchelf) so these to be able to locate libgphobos.so and libgruntime.so. This worked but seemed to require too much hacking (not to mention that had to consume the patchelf component itself) and didn't seem robust enough.

Setting env var `LD_LIBRARY_PATH` would be another option, not requiring extra tools.

dbankov-vmware: >Setting env var LD_LIBRARY_PATH would be another option, not requiring extra tools.
That's a good point but at that time I had another reason to go with patchelf and that was trying to use newer GLIBC on an otherwise very old build host (gcc-4.4.3 and glibc-2.5). That's when I tried linking everything statically (even GLIBC) which didn't work well for GLIBC so eventually moved to a newer build host (gcc-4.8.5, glibc-2.17) so GLIBC was no longer an issue but the desire to have everything else linked statically persisted :).

dbankov-vmware: Now that I thought of it it should probably be the best to implement -Xcc in GDC itself but I'm not sure I'm for that task.

ibuclaw: No idea what -Xcc is supposed to do, there is no C compiler involved in gdc.

dbankov-vmware: @ibuclaw what I did was to look at how "-Xcc" is implemented in LDMD2 and it appeared it just forwards "-Xcc" args directly to LDC so this seemed a possible approach for GDC as well.

Looking again at GDC I now think I was wrong stating the above because (as you mentioned) GDC doesn't use C/CPP related parts from GCC and would not be able to include correct C/CPP related link flags (which is what seems to be expected when using (L)DMD -Xcc).

Looking again at GDMD however it seems remotely possible to implement "-Xcc" in GDMD because it actually starts a separate process for linking and it could be made to call G++ instead of GDC for linking. BTW I tried that earlier (making "-Xcc" work like "-q," and GDMD calling G++ for linking) but it appeared that I would also need the flags that GDC would include on the link line with the passed options and this seemed too complex for me to implement.

In any case this seems harder to implement compared to the patch in this thread so I'll probably stick with what I have provided here for our build infrastructure.

JohanEngelen: LDC calls the C _compiler_ to do linking. You cannot use the `-L` flag for passing a flag to the C compiler, because when you pass, say, `-L -foo`, what will be passed to the C compiler (for linking) is `-Xlinker -foo` (such that indeed the flag is used for the _linker_ that is called by the C compiler).
My guess: because GDC _is also_ the C compiler (and thus calls linker directly), I think if you pass `-Xcc -foo` to LDC, passing just `-foo` to GDC should be equivalent.

dbankov-vmware: I was just able to bootstrap LDC with -DLDC_LINK_MANUALLY=OFF in our build infrastructure with this (https://github.com/dbankov-vmware/gdmd/commit/3ae6155dd9ea2f3df58881deaec59e646580d18c) patch on top of GDMD.

As I mentioned earlier it seems LDC build scripts expect that when -DLDC_LINK_MANUALLY=OFF the D compiler will link stdc++ automatically. Because of that expectation I've implemented -Xcc support in GDMD to automatically link stdc++ and also to honour `-static-libstdc++` option so static linking to work as well (which is what is needed in my case).

@ibuclaw Please let me know if this patch looks acceptable for GDMD so to post a PR if necessary.

dbankov-vmware: Forgot to mention that with GDMD from https://github.com/dbankov-vmware/gdmd/commit/3ae6155dd9ea2f3df58881deaec59e646580d18c and `-DLDC_LINK_MANUALLY=OFF` I again hit the symbol conflicts when I use unmodified version of libgphobos (where these are not marked as weak).

cmake/Modules/ExtractDMDSystemLinker.cmake
```diff
@@ -65,13 +65,63 @@ list(GET linker_line 0 D_LINKER_COMMAND)
...
-    # Filter linker arguments for those we know can be safely reused
+    # Filter linker arguments for those we know can be safely reused.
+    # Prepend arguments starting with "-B" in D_LINKER_ARGS with "-Wl," when
+    # using GDMD. The latter is needed because the arguments reported by GDMD
+    # are what is passed to COLLECT2 while D_LINKER_ARGS are later passed to
+    # CXX. The problem with that is that without "-Wl," prefix the -Bstatic
+    # and -Bdynamic arguments before and after -lgphobos and -lgdruntime are
+    # dropped. And that is a problem because without these the produced LDC is
+    # linked dynamically to libgphobos and libgdruntime so these have to be
+    # available whereever LDC is used.
```
ibuclaw, line 76: So what you are describing is an something specific to Debian and all its derivatives (Ubuntu, etc...)

https://salsa.debian.org/toolchain-team/gcc/-/blob/gcc-9-debian/debian/patches/gdc-dynamic-link-phobos.diff

There exists the driver option `-static-libphobos` which is probably more desirable, as there's no need to mess with `-Bdynamic` and `-Bstatic`, just `-static-libphobos -Wl,--allow-multiple-definition`

dmd/root/rmem.d
kinke: > As I mentioned earlier it seems LDC build scripts expect that when -DLDC_LINK_MANUALLY=OFF the D compiler will link stdc++ automatically.

Now that's really something that's LDC-specific and should be fixed. This (LDC_LINK_MANUALLY=OFF on Posix) currently only works with ldmd2 host compilers, by instructing it to invoke the C++ compiler for linking, not the C compiler: https://github.com/ldc-developers/ldc/blob/723904183fcd79bcdf47c43fd5cfb1cdf3a8f507/cmake/Modules/BuildDExecutable.cmake#L96-L99

cmake/Modules/ExtractDMDSystemLinker.cmake
dbankov-vmware: @ibuclaw I'm building GDC from the source in GCC 9.3 and if used directly my GDC version links both Phobos and DRuntime statically by default.

The problem occurs only when building LDC because when `-DLDC_LINK_MANUALLY=ON` (which is the default) instead of calling GDMD for linking LDC build script calls CXX for linking. Now to make sure that the correct libraries (phobos and druntime) are linked by CXX, LDC build script extracts the link line from a call to GDMD with `-v` and then copies all arguments which start with `-L`, `-l` and `-B` in a list which is later passed to CXX. The problem here is that for GDMD the link line reported by `-v` is what is passed to COLLECT2 (which are practically parameters passed to LD)  and then when LDC build script passed these directly to CXX only parameters starting with -L and -l are honoured while these starting with `-B` are simply dropped. The result is that although the link line reported by GDMD contains -Bstatic before -llibgphobos and -lgdruntime and -Bdynamic after these when CXX is linking LDC it drops the parameters starting -B and LDC ends up being linked dynamically with Phobos and DRuntime. What was needed was for parameters starting with `-B` form the link line reported by GDMD to be prepended with `-Wl,`.

dmd/root/rmem.d
dbankov-vmware: Yes I saw that and also for `LDC_LINK_MANUALLY=ON` it is the same:
https://github.com/ldc-developers/ldc/blob/723904183fcd79bcdf47c43fd5cfb1cdf3a8f507/cmake/Modules/BuildDExecutable.cmake#L73-L76
After fiddling so much with this I think there is a legitimate problem when linking C/C++ and D code which is that neither of the compilers knows how to link both languages correctly with their runtime/standard libraries. I haven't thought through this yet but it seems to me that what LDMD/LDC does is a good approach as the other options that I could think of (what I did in GDMD and generating the link line from scratch) seem worse to me. What I'm saying is that probably this doesn't need fixing in LDC build scripts but that maybe all D compilers should allow specifying the linking agent and should be able to pass through options from the customer to it but also include their own based on parameters that they recognise.

dbankov-vmware: I implemented what I mentioned above for GDMD in https://github.com/D-Programming-GDC/gdmd/pull/13 and with this version of GDMD am able to achieve what I need without touching LDC build scripts. Unfortunately I once again need a version of GCC with GDC which has _d_allocmemory, _d_newclass, _d_newitemiT and _d_newitemT symbols marked as weak so I'll post a PR with this update shortly.

Trying to figure out why bootstrapping LDC with both DMD and LDMD2 works without failures even though both link their versions of DRuntime and Phobos statically with LDC revealed that the duplicated symbols in both libdruntime.a (as well as in libdruntime-ldc.a) and libphobos.a (as well as in libphobos-ldc.a) are already marked as weak. Now for the LDC versions of the libraries this is achieved with the `@weak` attribute but for DMD ones it seems that all symbols are marked as `weak` by default and it was interesting to me (if somebody knows) wether this is the desired behaviour or has happened by accident.

PR #3816, opened 2021-08-16
-dllimport=defaultLibsOnly: Avoid -linkonce-templates requirement

kinke:
Via a sorts-of '-linkonce-templates light', only defining all *data* symbols instantiated from druntime/Phobos templates in each referencing CU (as `weak_odr`).


kinke:
Tested on Symmetry's code base and apparently working fine, without a measurable compile-time impact (but fixing hundreds of undefined symbol linker errors).

PR #3821, opened 2021-09-06
Upgrade frontend & libs to early v2.098.0

kinke:


kinke:
Looking good, almost done. For x86, apparently just some importC stuff to be fixed (not sure I can be bothered at this early stage). Some remaining druntime regressions for Android, some usual compatibility trouble with ltsmaster, and druntime integration tests hanging on AArch64 (probably new tests using the concurrent GC).

JohanEngelen:
@kinke I just noticed that this did not update all files in `dmd/root`. For example `dmd/root/man.d` is still an old version, not 2.098.

kinke:
I obviously can't rule out that I missed some parts, but v2.098 root/man.d hasn't been changed for a year: https://github.com/dlang/dmd/blob/v2.098.0/src/dmd/root/man.d

JohanEngelen:
> I obviously can't rule out that I missed some parts, but v2.098 root/man.d hasn't been changed for a year: https://github.com/dlang/dmd/blob/v2.098.0/src/dmd/root/man.d

Shouldn't git merge always update all the files, no way to miss it? Our `dmd/root/man.d` is old, you can see the delta on line 29 where the old `body` keyword is still used.

kinke:
It's not old, it just features `body` ugliness for LTS support.

PR #3822, opened 2021-09-07
Fix compilation with LLVM 14

thewilsonator:


driver/codegenerator.cpp
```diff
@@ -264,7 +264,9 @@ void CodeGenerator::writeAndFreeLLModule(const char *filename) {
   llvm::Metadata *IdentNode[] = {llvm::MDString::get(ir_->context(), Version)};
   IdentMetadata->addOperand(llvm::MDNode::get(ir_->context(), IdentNode));
 
+#if LDC_LLVM_VER < 1400
   context_.setInlineAsmDiagnosticHandler(inlineAsmDiagnosticHandler, ir_);
+#endif
```
thewilsonator, line 269: I feel like this has come up before, this is obviously not ideal

driver/cache.cpp
```diff
@@ -322,7 +322,7 @@ void outputIR2ObjRelevantCmdlineArgs(llvm::raw_ostream &hash_os) {
 #if LDC_LLVM_VER >= 800
   const auto framePointerUsage = opts::framePointerUsage();
   if (framePointerUsage.hasValue())
-    hash_os << framePointerUsage.getValue();
+    hash_os << (int)framePointerUsage.getValue();
```
thewilsonator, line 325: This fixed compilation, without this it errored with can't convert (return type) to `llvm::NoneType` or something like it. I'm open to suggestions to improve this.

gen/coverage.cpp
```diff
@@ -43,6 +43,9 @@ void emitCoverageLinecountInc(const Loc &loc) {
   case opts::CoverageIncrement::atomic:
     // Do an atomic increment, so this works when multiple threads are executed.
     gIR->ir->CreateAtomicRMW(llvm::AtomicRMWInst::Add, ptr, DtoConstUint(1),
+#if LDC_LLVM_VER >= 1400
+                             LLAlign(4),
```
thewilsonator, line 47: align(4) is what `uint` gets

gen/tocall.cpp
```diff
@@ -508,8 +508,12 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
       fatal();
     }
 
-    auto ret = p->ir->CreateAtomicCmpXchg(ptr, cmp, val, successOrdering,
-                                           failureOrdering);
+    auto ret = p->ir->CreateAtomicCmpXchg(ptr, cmp, val,
+#if LDC_LLVM_VER >= 1400
+                                          LLMaybeAlign(),
```
thewilsonator, line 513: unsure if we have a priori knowledge of what the alignment should be here

gen/tocall.cpp
```diff
@@ -554,6 +558,9 @@ bool DtoLowerMagicIntrinsic(IRState *p, FuncDeclaration *fndecl, CallExp *e,
     LLValue *val = DtoRVal(exp2);
     LLValue *ret =
         p->ir->CreateAtomicRMW(llvm::AtomicRMWInst::BinOp(op), ptr, val,
+#if LDC_LLVM_VER >= 1400
+                               LLMaybeAlign(),
```
thewilsonator, line 562: ditto

driver/codegenerator.cpp
thewilsonator: Ahh, https://github.com/ldc-developers/ldc/pull/3721#discussion_r631575006

thewilsonator:
This should be ready for review.

driver/codegenerator.cpp
thewilsonator: solved in bbd0091

driver/codegenerator.cpp
```diff
@@ -173,6 +174,40 @@ void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
...
+
+    // keep on using llvm::SMDiagnostic::print() for nice, colorful output
+    llvm::SMDiagnostic d2(*d.getSourceMgr(), d.getLoc(), filename, d.getLineNo(),
+                          d.getColumnNo(), d.getKind(), d.getMessage(),
+                          d.getLineContents(), d.getRanges(), d.getFixIts());
+    d2.print(nullptr, llvm::errs());
+    return true;
+  }
+};
+#endif
```
kinke, line 210: The 2 functions share most code - it might make sense to extract an impl function (returning a bool etc.) and call that in the new `InlineAsmDiagnosticHandler` as well as the old callback function.

driver/codegenerator.cpp
```diff
@@ -173,6 +174,40 @@ void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
...
+        return false;
+
+    const auto &DISM = llvm::cast<llvm::DiagnosticInfoSrcMgr>(DI);
+    if (DI.getKind() == llvm::SourceMgr::DK_Error)
+      ++global.errors;
+    unsigned locCookie = DISM.getLocCookie();
+    auto & d = DISM.getSMDiag();
+    if (!locCookie) {
+      d.print(nullptr, llvm::errs());
+      return false;
```
kinke, line 194: Returning `false` might imply that LLVM prints it again.

driver/codegenerator.cpp
thewilsonator: Will do.

driver/codegenerator.cpp
```diff
@@ -173,6 +174,40 @@ void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
...
+struct InlineAsmDiagnosticHandler : public llvm::DiagnosticHandler {
+  IRState *irs;
+  InlineAsmDiagnosticHandler(IRState *irs) : irs(irs) {}
+
+    // return false to defer to LLVMContext::diagnose()
+  bool handleDiagnostics(const llvm::DiagnosticInfo &DI) override {
+    if (DI.getKind() != llvm::DK_SrcMgr)
+        return false;
+
+    const auto &DISM = llvm::cast<llvm::DiagnosticInfoSrcMgr>(DI);
```
kinke, line 187: Does that return a ref or does the type param need to be a ref for that?

driver/codegenerator.cpp
thewilsonator: Will double check, but the documentation wasn't very thorough.

driver/cache.cpp
kinke: So `{has,get}Value()` aren't enough anymore to skip the not-specified case?!

driver/codegenerator.cpp
thewilsonator: No idea, copied this directly from one the similar examples in the diff that obsoleted the `setInlineAsmDiagnosticHandler`.

driver/linker-msvc.cpp
```diff
@@ -36,15 +36,19 @@ void addMscrtLibs(bool useInternalToolchain, std::vector<std::string> &args) {
   // We need the vcruntime lib for druntime's exception handling (ldc.eh_msvc).
   // Pick one of the 4 variants matching the selected main UCRT lib.
 
+#if LDC_LLVM_VER < 1400
+#define contains_insensitive contains_lower
+#define endswith_insensitive endswith_lower
+#endif
```
kinke, line 42: I'd define/`using` it the other way around - IIRC, `contains_lower` required the needle to be in lowercase and only converted the haystack to lowercase - but I might be wrong.

thewilsonator: OK

gen/tocall.cpp
kinke: Something like `LLMaybeAlign(gDataLayout->getABITypeAlignment(val->getType()))`.

gen/tocall.cpp
```diff
@@ -1028,9 +1035,15 @@ DValue *DtoCallFunction(const Loc &loc, Type *resulttype, DValue *fnval,
     call->setCallingConv(callconv);
   }
   // merge in function attributes set in callOrInvoke
+#if LDC_LLVM_VER >= 1400
+attrlist = attrlist.addFnAttributes(
+        gIR->context(),
+        llvm::AttrBuilder(call->getAttributes(), LLAttributeList::FunctionIndex));
```
kinke, line 1041: clang-format?

driver/codegenerator.cpp
thewilsonator: Full comment on `llvm:: DiagnosticHandler:: handleDiagnostics` is 
```
/// Override handleDiagnostics to provide custom implementation.
/// Return true if it handles diagnostics reporting properly otherwise
/// return false to make LLVMContext::diagnose() to print the message
/// with a prefix based on the severity.
```
Oh I see what you mean now, will fix that.

driver/codegenerator.cpp
thewilsonator: Type param is a ref, but `auto` is always a value type, never a pointer or a reference, for those we need to use `auto*` and ` auto&`.

driver/linker-msvc.cpp
thewilsonator: well the needles are all lower case here anyway and there are 6 occurrences in total between the top of them. 
Will swap to `using`, I keep forgetting C++ has that.

driver/cache.cpp
thewilsonator: exact error is 
```
ldc/driver/cache.cpp:325:13: error: invalid operands to binary expression ('llvm::raw_ostream' and 'const llvm::FramePointerKind')
    hash_os << framePointerUsage.getValue();
```
followed by a large number of 
```
no known conversion from 'const llvm::FramePointerKind' to 'foo'
```
Given we only need this for a hash, I think the int cast should be fine.

gen/tocall.cpp
thewilsonator: done

gen/tocall.cpp
thewilsonator: done

driver/linker-msvc.cpp
kinke: > well the needles are all lower case here anyway

Yes, but `contains_insensitive` suggests it doesn't have to, contrary to the old name.

driver/cache.cpp
kinke: Okay, but a `static_cast<int>` then please, I don't think we have C-style casts anymore.

thewilsonator: Done

driver/linker-msvc.cpp
thewilsonator: Done

kinke:
Thx!

thewilsonator:
Thank you for the excellent reviews!

PR #3834, opened 2021-09-19
[dcompute] fix some ABI issues

thewilsonator:
I'll add tests for these in the next PR that will add support for OpenCL images.

### DIFF
--- a/gen/abi-spirv.cpp
+++ b/gen/abi-spirv.cpp
@@ -34,9 +34,19 @@ struct SPIRVTargetABI : TargetABI {
       if (!arg->byref)
         rewriteArgument(fty, *arg);
     }
+    if (!skipReturnValueRewrite(fty))
+      rewriteArgument(fty, *fty.ret);
   }
   bool returnInArg(TypeFunction *tf, bool) override {
-    return !tf->isref() && DtoIsInMemoryOnly(tf->next);
+    if (tf->isref())
+      return false;
+    Type *retty = tf->next->toBasetype();
+    if (retty->ty == TY::Tsarray)
+      return true;
+    else if (auto st = retty->isTypeStruct())
+      return !toDcomputePointer(st->sym);
+    else
+      return false;
   }
   void rewriteArgument(IrFuncTy &fty, IrFuncTyArg &arg) override {
     Type *ty = arg.type->toBasetype();
--- a/gen/dcompute/abi-rewrites.h
+++ b/gen/dcompute/abi-rewrites.h
@@ -15,7 +15,19 @@
 
 #include "gen/abi-generic.h"
 
-struct DComputePointerRewrite : BaseBitcastABIRewrite {
+struct DComputePointerRewrite : ABIRewrite {
+  LLValue *put(DValue *v, bool isLValueExp, bool) override {
+    LLValue *address = DtoLVal(v);
+    address = DtoGEP(address, 0u, 0u);
+    return DtoLoad(address, ".DComputePointerRewrite_arg");
+  }
+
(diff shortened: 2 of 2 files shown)


gen/dcompute/abi-rewrites.h
```diff
@@ -15,7 +15,21 @@
...
-struct DComputePointerRewrite : BaseBitcastABIRewrite {
+struct DComputePointerRewrite : ABIRewrite {
+  LLValue *put(DValue *v, bool isLValueExp, bool) override {
+    LLValue *address = DtoLVal(v);
+    address = DtoGEP(address, 0u, 0u);
+    return DtoLoad(address, ".DComputePointerRewrite_arg");
+  }
+
+  LLValue *getLVal(Type *dty, LLValue *v) override {
+    const unsigned alignment = BaseBitcastABIRewrite::getMaxAlignment(v->getType(), dty);
```
kinke, line 26: Are you worried the alignment of the addrspace pointer can actually be higher than the struct wrapper's? In that case, it would probably make sense to fix up the struct alignment. If the alignments match, a simple `DtoAlloca()` would suffice.

gen/dcompute/abi-rewrites.h
```diff
@@ -15,7 +15,21 @@
...
+    LLType *asMemType = DtoMemType(dty);
+    LLValue *mem = DtoRawAlloca(asMemType, alignment, ".DComputePointerRewrite_param_storage");
+    DtoStoreZextI8(v, DtoGEP(mem, 0u, 0u));
```
kinke, line 29: `DtoStore()` should suffice, a pointer can never be an `i1`.

gen/abi-spirv.cpp
```diff
@@ -34,9 +34,18 @@ struct SPIRVTargetABI : TargetABI {
       if (!arg->byref)
         rewriteArgument(fty, *arg);
     }
+    rewriteArgument(fty, *fty.ret);
```
kinke, line 37: You'd rewrite a Dcomputer pointer struct returned by ref too this way. I'd suggest using a
```
if (!skipReturnValueRewrite(fty))
  rewriteArgument(fty, *fty.ret);
```
as most other TargetABIs.

gen/abi-spirv.cpp
```diff
@@ -34,9 +34,18 @@ struct SPIRVTargetABI : TargetABI {
...
   }
   bool returnInArg(TypeFunction *tf, bool) override {
-    return !tf->isref() && DtoIsInMemoryOnly(tf->next);
+    if (tf->isref())
+      return false;
+    Type *retty = tf->next->toBasetype();
+    if (retty->ty == TY::Tsarray)
+      return true;
+    else if (retty->ty == TY::Tstruct)
+      return !toDcomputePointer(retty->isTypeStruct()->sym);
```
kinke, line 46: [a bit more pleasing to the eye IMO: `else if (auto st = retty->isTypeStruct()) return !toDcomputePointer(st->sym);`]

gen/dcompute/abi-rewrites.h
thewilsonator: No that was a relic of pasting in the body of `BaseBitcastABIRewrite:: getLVal` to simplify and figure out how to generate GEPs instead of bitcasts. fixed.

gen/abi-spirv.cpp
thewilsonator: done

gen/abi-spirv.cpp
thewilsonator: done

thewilsonator:
Review addressed.

PR #3835, opened 2021-09-20
[dcompute] add support for OpenCL image I/O

thewilsonator:


tests/codegen/dcompute_cl_images.d
```diff
@@ -0,0 +1,46 @@
...
+import ldc.dcompute;
+import opencl;
+
+// CHECK: %opencl.image1d_ro_t = type opaque
+// CHECK: %opencl.image1d_wo_t = type opaque
+// CHECK: %"ldc.dcompute.Pointer!(AddrSpace.Global, image1d_ro_t).Pointer" = type { %opencl.image1d_ro_t addrspace(1)* }
+// CHECK: %"ldc.dcompute.Pointer!(AddrSpace.Global, image1d_wo_t).Pointer" = type { %opencl.image1d_wo_t addrspace(1)* }
+// CHECK: %"ldc.dcompute.Pointer!(AddrSpace.Shared, sampler_t).Pointer" = type { %opencl.sampler_t addrspace(2)* }
+// CHECK: %opencl.sampler_t = type opaque
+// CHECK: %opencl.image1d_rw_t = type opaque
```
thewilsonator, line 13: It seems the order of these declarations is sensitive. does `lit` have an inverse of `CHECK-NEXT`? i.e. check the input is present but not required to be ordered w.r.t other `CHECKS`?

JohanEngelen: Have a look at `CHECK-DAG`

tests/codegen/dcompute_cl_images.d
```diff
@@ -0,0 +1,46 @@
+// REQUIRES: target_SPIRV
+// RUN: %ldc -c -mdcompute-targets=ocl-220 -m64 -I%S/inputs -mdcompute-file-prefix=images -output-ll -output-o %s && FileCheck %s < images_ocl220_64.ll
```
JohanEngelen, line 2: to avoid conflict with other testcases, shouldn't the file prefix be something with `%t` ?

dmd/id.d
```diff
@@ -553,6 +553,7 @@ immutable Msgtable[] msgtable =
     { "dcPointer", "Pointer" },
     { "dcReflect", "__dcompute_reflect" },
     { "RTInfoImpl" },
+    { "opencl" },
```
JohanEngelen, line 556: do you use this anywhere?

gen/dcompute/targetOCL.cpp
```diff
@@ -174,17 +172,33 @@ class TargetOCL : public DComputeTarget {
     if (v->type->ty == TY::Tstruct &&
         (ptr = toDcomputePointer(static_cast<TypeStruct *>(v->type)->sym))) {
       addrspace = ptr->addrspace;
-      tyName = basicTypeToString(ptr->type) + "*";
-      baseTyName = tyName;
+      tyName = basicTypeToString(ptr->type);
+
+      auto sd = ptr->type->isTypeStruct();
+      auto mod = sd ? sd->sym->getModule() : nullptr;
+      if (mod && mod->md && mod->md->id == Id::opencl) {
```
JohanEngelen, line 179: So this is where you hardcode that the module name must be `opencl`. Should this not be `ldc.opencl` ? That way, we can ship LDC with that hardcoded module.
Or am I misunderstanding things?

gen/dcompute/targetOCL.cpp
```diff
@@ -174,17 +172,33 @@ class TargetOCL : public DComputeTarget {
...
+        // TODO: Pipes
+        auto name = std::string(sd->toChars());
+        name = name.substr(name.length()-4, 2);
```
JohanEngelen, line 182: another big hardcoded assumption!

gen/dcompute/targetOCL.cpp
thewilsonator: > Should this not be ldc.opencl?

Alas it must be, the SPIR-V backend depends on exclusively the type e.g. `opencl.image1d_ro_t` and I don't know how to force the name of a type.

gen/dcompute/targetOCL.cpp
thewilsonator: all the types in that module are `imageXd_YY_y` (X=(1,2, 3) (YY = (ro,wo,rw) or an `array`, `msaa` and/or `depth` variant thereof, or `sampler_t` (I've not looked at pipes but they are "access qualified" (i.e. `read_only`, `write_only` or `read_write`) too so I presume they are the same). What this code does is extracts YY from the name from which we get the access qualifier.

tests/codegen/dcompute_cl_images.d
thewilsonator: Will do, its been a while since I've used `lit`.

gen/dcompute/targetOCL.cpp
JohanEngelen: I think it would work if you put `module opencl;` at the top of the file, while still having it in `ldc/opencl.d`.

thewilsonator: true, `druntime/src/ldc` would have to be on the import path in addition to `druntime/src`.

I'll do that in a separate PR, because only a few of the types are needed to test the code generation and I've not finished the file yet, I still need to figure out which combinations of `{1d,2d,3d}, {ro,wo,rw}, {array, msaa, depth, buffer,(none)}` are valid and the msaa ones are only valid with OpenGL sharing or something. 

(Also I still don't know how submodule update PRs work.)

JohanEngelen: > true, `druntime/src/ldc` would have to be on the import path in addition to `druntime/src`.

That would be a big change (because then suddenly `import attributes;` would 'work').
I would expect the type names to be generated from the name specified by `module <name>`, rather than the how it is imported using `import <name>`.
Zooming out a little, I think requiring that the type name in D is exactly how LLVM wants it internally is very brittle. For example, if LLVM expects a different type name, then all user code would have to change aswell (perhaps solvable with  a bunch of stable `alias` types in LDC's `opencl.d`?). I'd say first priority is getting things to work, but soon after there needs to be some thought behind the future/longer term stability of this functionality working.

> I'll do that in a separate PR

ok

thewilsonator: >That would be a big change (because then suddenly import attributes; would 'work')

Indeed. We could have it along side `object.d`, I currently have it under [`dcompute/source`](https://github.com/thewilsonator/dcompute/tree/cl-image2/source)

> requiring that the type name in D is exactly how LLVM wants it internally is very brittle

It may be brittle, but it is exceedingly unlikely to change (SPIR (n.b. not SPIR-V) is a standard after all) unless we change backends (which I would love to do, but there is not official SPIR-V that is part of the LLVM project) in which case they would probably all be `i8 addrspace(1)*` to intrinsic functions which would be completely handled by the dcompute library.

>perhaps solvable with a bunch of stable alias types in LDC's opencl.d?

Of course, there's no point calling it an `GlobalPointer!image1d_ro_t` when you can alias it to `ReadOnlyImage!1`.
The only mildly annoying concern is the conditionally declared types that depend on if OpenGL image sharing is enabled.

tests/codegen/dcompute_cl_images.d
```diff
@@ -0,0 +1,45 @@
+// REQUIRES: target_SPIRV
```
JohanEngelen, line 1: do we have any CI system that runs these tests?

thewilsonator:
```
ci/circleci: macOS-x64
macOS 10.15 x64
Azure Pipeline (Windows x86)
Azure Pipeline (Windows x64)
```

(at least) all have target_SPIRV enabled. I think its part of our LLVM build process

thewilsonator:
Any final comments? It would be good to get this in before `v1.28` is tagged.

gen/dcompute/targetOCL.cpp
kinke: > Should this not be ldc.opencl?

Yeah, that would be much nicer indeed IMO, keeping extra LDC stuff in its package.

> I don't know how to force the name of a type.

The name is currently initialized and then untouched AFAICT in https://github.com/ldc-developers/ldc/blob/f9ba17248eb53021c91130b0cb7c6c00a3f8074c/ir/irtypeaggr.cpp#L206. You *could* override it in https://github.com/ldc-developers/ldc/blob/f9ba17248eb53021c91130b0cb7c6c00a3f8074c/ir/irtypestruct.cpp#L49-L55 (before returning for opaque structs, as all of these magic structs are opaque at the moment).

> Also I still don't know how submodule update PRs work.

You push a branch onto the LDC repo and then reference it here in an LDC PR. After merging the PR here, the druntime branch needs to be merged into druntime's `ldc` branch (and must be a fast-forward merge, i.e., the druntime branch must branch off the current `ldc` head).

gen/dcompute/targetOCL.cpp
kinke: A little comment in the code (`parse access qualifiers from ldc.opencl.*_XX_t types like ldc.opencl.image1d_ro_t`) would be nice - there are hardly any comments here explaining what this stuff is all about.

tests/codegen/dcompute_cl_images.d
kinke: Yep, all CIs using our LLVM fork, most notably all Azure jobs.

gen/dcompute/targetOCL.cpp
```diff
@@ -38,6 +39,7 @@
 
 namespace {
 class TargetOCL : public DComputeTarget {
+    bool usedImage;
```
kinke, line 42: Seems like you read from it but don't set it.

thewilsonator: fixed in 1f24be6 stackoverflow tells me that `s.rfind("needle", 0) == 0` is how you're supposed to do `startsWith` on a string prior to C++20.

gen/dcompute/targetOCL.cpp
thewilsonator: comment added in fac6fec

gen/dcompute/targetOCL.cpp
thewilsonator: > You could override it before returning for opaque structs

Thanks, will do some investigation tomorrow about this.

> as all of these magic structs are opaque at the moment

correct, and I don't see that changing any time soon.

> You push a branch onto the LDC repo and then reference it here in an LDC PR. After merging the PR here, the druntime branch needs to be merged into druntime's ldc branch (and must be a fast-forward merge, i.e., the druntime branch must branch off the current ldc head).

THANKS!!! will bookmark this.

thewilsonator: done

thewilsonator:
Hmm, I seem to have a merge conflict that I can't figure out how to fix.

thewilsonator:
Hmm, the druntime commit seems to be 
```diff
-e1ba20fec08568c9f7d81ceedbac84d78cfdcd34
+e3738e482e9854cf7eb05c8f79abe5405fd1a8a9
```
instead of 
```diff
-06fdf5464471d9877bce2aeadc80d527b7e8db1d
+e3738e482e9854cf7eb05c8f79abe5405fd1a8a9
```
hence the merge conflict. I've no idea why its doing that though or how to fix it.

thewilsonator:
Yay, it worked!

kinke:
That's probably caused by master pointing to a newer druntime hash (I've merged upstream stable recently) than when this branch was branched off - rebase this branch in that case (and yes, it's going to complain about a druntime conflict, just commit your change, your druntime branch is fine).

thewilsonator:
Yay! I've still got some fixes to do on the druntime bit.

kinke:
You could try making `ldc.opencl` a `.di` - it has opaque structs and aliases only, shouldn't need to be compiled (no init symbols etc., neither for `GlobalPointer!…` instantiations as they are most likely zero-initialized).

thewilsonator:
That fixes it locally.

thewilsonator:
Ubuntu 16.04 x64 ltsmaster fails with: 
> Unable to checkout 'f2a6ae9802aceef9e958ac0d62b0128b85c1aeae' in submodule path 'runtime/druntime'

is that a problem with ltsmaster CI?

kinke:
I've just retried it (thought you might have pushed the druntime commit a bit too late after pushing here), but it failed identically. Might have to do with an old git of that vanilla Ubuntu 16 image and some strange `--depth 50` shallow-cloning behavior. The job is new, so no history of this problem so far.

kinke:
You could try removing the `--depth` bit in https://github.com/ldc-developers/ldc/blob/f9ba17248eb53021c91130b0cb7c6c00a3f8074c/.cirrus.yml#L186

thewilsonator:
Hmm, didn't work.

kinke:
You've patched the wrong line.

thewilsonator:
Thanks, that seems to have fixed it.

thewilsonator:
I think this is good to go.

ir/irtypestruct.cpp
```diff
@@ -56,6 +56,11 @@ IrTypeStruct *IrTypeStruct::get(StructDeclaration *sd) {
 
   // if it's a forward declaration, all bets are off, stick with the opaque
   if (sd->sizeok != Sizeok::done) {
+    // but rewrite the name for special OpenCL types
+    if (isFromLDC_OpenCL(sd)) {
+      const int prefixlen = 4; // == sizeof("ldc.")
+      t->type = LLStructType::create(gIR->context(), sd->toPrettyChars()+prefixlen);
```
kinke, line 62: `LLStructType` should have a `{get,set}Name()`, where you could simply strip off the first 4 chars.

gen/dcompute/targetOCL.cpp
kinke: You could use `isFromLDC_OpenCL(sd)` for the if-condition now, that would simplify things.

kinke:
[As for druntime, I suppose you could get rid of `@compute(CompileFor.deviceOnly)` as it's a `.di` anyway.]

ir/irtypestruct.cpp
thewilsonator: Fixed. Sorry I meant to do this earlier but I forgot to push.

thewilsonator:
Good to go?

PR #3842, opened 2021-10-03
Add support for LLVM 13 & bump LDC-LLVM to v13.0.1

kinke:


kinke:
[Preliminary support extracted to #3846.]

calvin2021y:
Maybe add  static libxml2 and libz to the build process,  I think with this we can reuse the glibc binary from musl. 

The current glibc binary not work with alpine glibc-compat because:

```sh
 ./ldc2 -v
./ldc2: Relink `/usr/lib/liblzma.so.5' with `/usr/glibc-compat/lib/libc.so.6' for IFUNC symbol `memmove'
Segmentation fault (core dumped)
```


build with libxml2 with options and link with it should fix the problem.
```sh
 ./configure --enable-static --enable-shared=no --without-lzma --without-icu --without-python --without-docbook  --without-iconv
```

kinke:
The remaining failures on Linux are fixed by switching from ld.gold to lld v13, on my box at least. So this seems to be the inverse of https://github.com/ldc-developers/ldc/issues/3861 - LLVM 13 causing ModuleInfo-related troubles with non-lld linkers. I hope this will be fixed by #3850 too.

kinke:
Seems like it, I've got a single `std.experimental.allocator.building_blocks.allocator_list-shared_32` failure with ld.gold when merging #3850.

kinke:
I could reproduce strange dmd-testsuite failures with gold on Ubuntu20. Merged current LLVM `release/13.x` head (13.0.1-rc2+), and 32/64-bit dmd-testsuite is green, so looks like this apparently serious LLVM 13 issue was fixed in the meantime.

kinke:
Hmm, the vanilla 13.0.0 GH Actions job now seems to pass consistently with its gold linker. The CI image version hasn't changed, the runner has - no idea.

kinke:
The GH Actions LLVM 13 job with old ld.gold is driving me somewhat nuts. Yesterday, it went from green to red inside an hour without any real change. I've then switched from vanilla v13.0.0 to just-released LDC-LLVM v13.0.1, and the first attempt led to some (different) failures, incl. something hanging for a *long* time. Retried today, green. So these failures seem totally indeterministic.

I'll go ahead and merge this now that v13.0.1 is available; if the job failure is frequent, I'll do something about it.

Issue #3786, opened 2021-07-10
LLD and `-flto=thin` calls `ldc.register_dso` multiple times instead of once at startup/shutdown


JohanEngelen:
When using LLD with `-flto=thin`, we are running into the problem that each D module's call to `ldc.register_dso` in `llvm.global_ctors` is _not_ folded into one call. We are relying on LLVM functionality where if a D module's `ldc.register_dso` is "discarded" because of another module's `ldc.register_dso` overriding it, then the entry in `llvm.global_ctors` is also discarded. This is apparently happening with `ld.gold`, but not with `ld.lld`.

It's related to this work: https://github.com/ldc-developers/ldc/pull/2870

Strangely, when I manually put the ctor/dtor in `.init_array`/`.fini_array` sections, the symbols are not merged either (not with bfd, gold, or lld). That is, replacing https://github.com/ldc-developers/ldc/blob/2ec2fd7814fdb46c617a7ac5bbe76dee0a610aec/gen/modules.cpp#L367-L368 with
```cpp
    const auto dsoCtor = defineDSOGlobal("ldc.dso_ctor", registerDSO);
    const auto dsoDtor = defineDSOGlobal("ldc.dso_dtor", registerDSO);
    gIR->usedArray.push_back(dsoCtor);
    gIR->usedArray.push_back(dsoDtor);
    dsoCtor->setSection(".init_array");
    dsoDtor->setSection(".fini_array");
```

JohanEngelen:
(testcase = https://github.com/ldc-developers/ldc/blob/master/tests/linking/thinlto_modulecdtors.d)

kinke:
Sounds more like an LLD issue to me. FWIW, I think the way to tackle and further simplify this would be to adapt the Windows scheme with a special prebuilt module: https://github.com/ldc-developers/druntime/blob/ldc/src/rt/dso_windows.d

GDC does a similar thing apparently. All code would be in druntime, and people wouldn't wonder about the extra emissions when inspecting produced asm either. All we'd lose is the ability to link arbitrary D object files to binaries when bypassing LDC for linking, in which case the extra module would need to be linked in manually.

JohanEngelen:
Reproducer for LLVM (will submit to LLVM and ask about it):

```llvm
; File: main.ll
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

[...]

PR #3848, opened 2021-10-13
Linux: Don't default to ld.gold linker & fix TLS range with static x86/x64 druntime

kinke:
I hope the issues with old bfd aren't really a thing anymore.

The main reason is that the user might have set LLD as default `/usr/bin/ld` symlink, and LDC not using it by default isn't intuitive. And LLD can be significantly faster than gold.

### DIFF
--- a/.azure-pipelines/1-posix-setup.yml
+++ b/.azure-pipelines/1-posix-setup.yml
@@ -65,7 +65,9 @@ steps:
     tar -xf llvm.tar.xz --strip 1 -C llvm
     # Non-Mac: make lld the default linker (possibly with enabled assertions unfortunately)
     if [ "$CI_OS" != "osx" ]; then
-      sudo ln -sf "$PWD/llvm/bin/ld.lld" /usr/bin/ld
+      #sudo ln -sf "$PWD/llvm/bin/ld.lld" /usr/bin/ld
+      # FIXME: https://github.com/ldc-developers/ldc/issues/3786
+      sudo ln -sf /usr/bin/ld.gold /usr/bin/ld
     fi
     # Set PARALLEL_JOBS env variable and persist it for future steps
     if [ "$CI_OS" = "osx" ]; then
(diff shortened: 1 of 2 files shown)


kinke:
For reference, here are some linking times on Ubuntu 20.04 for a pretty large executable (~1.1 GB with debuginfos) using the distro linkers:
```
bfd v2.34:
	Elapsed (wall clock) time (h:mm:ss or m:ss): 1:26.59
	Maximum resident set size (kbytes): 16211232

gold v1.16 (from binutils v2.34):
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:18.23
	Maximum resident set size (kbytes): 10524112

lld v10.0.0:
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:05.48
	Maximum resident set size (kbytes): 6338564
```

kinke:
Oh, lld is more picky than gold, and rightfully so:
```
ld: error: duplicate symbol: std::bad_exception::~bad_exception()
>>> defined in /usr/lib/gcc/x86_64-linux-gnu/10/libstdc++.a(eh_exception.o)
>>> defined in /home/vsts/work/1/bootstrap-ldc/lib/libdruntime-ldc-lto.a(exception.bc.o)

ld: error: duplicate symbol: std::bad_cast::~bad_cast()
>>> defined in /home/vsts/work/1/bootstrap-ldc/lib/libdruntime-ldc-lto.a(typeinfo.bc.o)
>>> defined in /usr/lib/gcc/x86_64-linux-gnu/10/libstdc++.a(bad_cast.o)

ld: error: duplicate symbol: std::bad_typeid::~bad_typeid()
>>> defined in /home/vsts/work/1/bootstrap-ldc/lib/libdruntime-ldc-lto.a(typeinfo.bc.o)
>>> defined in /usr/lib/gcc/x86_64-linux-gnu/10/libstdc++.a(bad_typeid.o)
```

These are auto-generated dtors which don't inherit the `@weak` UDA of the dtor in the base class...

kinke:
The frontend workaround is now proposed as https://github.com/dlang/dmd/pull/13167. It's really just a workaround for simple cases like this, and just addressing the IMO oddity that derived C++ classes get a superfluous dtor, even if they don't introduce any new fields requiring destruction or user dtors. The `@weak` UDA limitation only affecting the user dtor is still in place.

Not using gold has led to #3849. And the LTO lit-tests with lld run into https://github.com/ldc-developers/ldc/issues/3786 apparently.

PR #3854, opened 2021-10-20
Support 'raw mangles' via leading "\1" in pragma(mangle) strings

kinke:
Such mangled names aren't affected by a target-specific default (C) prefix, such as `_` on Mac.


Geod24:
What prevents this from being upstream ?

kinke:
A DMD backend/glue layer implementation and an agreed prefix/notion, as the "\1" is AFAIK just an LLVM convention.

PR #3860, opened 2021-10-29
Draft: Remove -unittest special cases wrt. template emission

kinke:
Tests https://github.com/dlang/dmd/pull/13224.

### DIFF
--- a/dmd/dtemplate.d
+++ b/dmd/dtemplate.d
@@ -6230,39 +6230,43 @@ version (IN_LLVM)
         assert(!global.params.linkonceTemplates);
 }
 
+        // minst is finalized after the 1st invocation.
+        // tnext and tinst are only needed for the 1st invocation and
+        // cleared for further invocations.
+        TemplateInstance tnext = this.tnext;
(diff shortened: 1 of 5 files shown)


kinke:
Some numbers for runs *without* default `-linkonce-templates`, comparing the total size of all *separately compiled* (`COMPILE_ALL_D_FILES_AT_ONCE=OFF`) druntime/Phobos object files with master:
* with `-unittest`: about -8.7% (almost identical percentage across release and debug versions) with the new template emission alone, -11% with additional ignoring of non-root unittests. New `-allinst` adds some 0.1% to the new size and thus seems to hardly make a difference.
* without `-unittest`: +0.5% with the new template emission, and new `-allinst` again hardly changes that number.

`version (unittest)` blocks in templates can still be a source for problems, and ignoring unittests in instantiations of templates not declared in a root module was required to fix undefined symbols for the upstream `vibe-d/vibe.d+base` Buildkite test.

kinke:
This has landed in D 2.099 and LDC master.

PR #3862, opened 2021-11-03
Place __minfo global variables in llvm.used

### LINKED ISSUE #3861: ELF: `__minfo` global variables may be garbage collected by ld.lld>=13.0.0 if built with llvm-project<13.0.0
I maintain lld/ELF nowadays. Sorry for your inconvenience but for the benefit of better garbage collection for metadata sections (PGO/sanitizer and more future instrumentation), I changed a garbage collection rule in LLD 13.0.0 (https://releases.llvm.org/13.0.0/tools/lld/docs/ReleaseNotes.html):

A `__start___minfo` reference from a live input section no longer keeps a `__minfo` section live.

(The topic is complex. You may read https://maskray.me/blog/2021-01-31-metadata-sections-comdat-and-shf-link-order#garbage-collection-on-metadata-sections for more information.)

FreeBSD is trying to upgrade its `/usr/bin/ld` to LLD 13.0.0. If ldc is built with llvm-project<13.0.0 when host ld.lld >= 13.0.0, there can be breakage:

* ld.lld>=13.0.0 defaults to `-z start-stop-gc` and discards unreferenced `__minfo` sections.
* A `__minfo` section will be retained if it has the `SHF_GNU_RETAIN` flag.
* Before llvm-project 13.0.0, `llvm.used` does not set `SHF_GNU_RETAIN` flag.

Some fixes:

* Detect host linker flavor and version. If `LDC_LLVM_VER < 1300 && ld.lld >= 13.0.0`, pass `-z nostart-stop-gc`. This can be a CMake-time check.
* FreeBSD uses llvm-project 13.0.0 to build ldc. Note: for LTO we really like the compiler and the linker to use the same LLVM version, though larger version linker usually works. IIUC FreeBSD's usage is not friendly to LTO.
* Somehow places `__minfo` in a section group with a `SHT_INIT_ARRAY` section. This is a hack which should retire ASAP.
* Disable `--gc-sections`
* FreeBSD packager switches to a bundled ld.lld with a version matching llvm-project. This is the recommended way using LTO.

---

ld64 on Mach-O behaves similar to `ld.lld -z nostart-stop-gc` in the presence of `-dead_strip`.

MaskRay:
This removes an abuse of ELF linker behaviors while keeping
Mach-O/COFF/WebAssembly linker behaviors unchanged.

LLD 13.0.0 (default) and GNU ld 2.37 support `-z start-stop-gc` which enables
the linker to linker to garbage collect unreferenced C identifier name sections
in the presence of `__start_/__stop_`. The `__minfo` section is otherwise
unreferenced and may be garbage collected.

Add the global variables to `llvm.used` can prevent garbage collection.

Fixes #3861

---

I don't know how to add test and I may not have enough time following up. If this needs larger changes, please just take over.

### DIFF
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -325,6 +325,8 @@ void emitModuleRefToSection(RegistryStyle style, std::string moduleMangle,
   auto thismref = defineDSOGlobal(thismrefIRMangle,
                                   DtoBitCast(thisModuleInfo, moduleInfoPtrTy));
   thismref->setSection(moduleInfoRefsSectionName);
+  // Ensure __minfo will not be discarded by ld -z start-stop-gc.
+  appendToUsed(gIR->module, {thismref});
   gIR->usedArray.push_back(thismref);
 
   if (!isFirst || style == RegistryStyle::sectionSimple) {


MaskRay:
Seems that `llvm.used` is already used. Closing...

PR #3866, opened 2021-11-08
DI: delegate context pointer should be .ptr instead of .context

ljmf00:
According to the ABI specification, when describing the delegates memory
layout, the context pointer is named .ptr .

https://dlang.org/spec/abi.html#delegates


ljmf00:
I'm not sure if this is being tested. Going to convert to draft until workflows approval.

kinke:
It is tested, see test failure (https://github.com/ldc-developers/ldc/blob/master/tests/debuginfo/args_cdb.d#L88).

thewilsonator (approved):

PR #3871, opened 2021-11-13
Azure & Cirrus CI: Bump macOS image to 11.x (Big Sur)

kinke:


jacob-carlborg:
> LDC cannot be linked due to missing libm, libz, libxml2… I'm wondering whether that has to do with

I had a quick look. For this build https://github.com/ldc-developers/ldc/runs/4200111986 it looks like it doesn't link with zlib. I don't see `-lz` on the command line.

As for the Cirrus CI problem https://github.com/ldc-developers/ldc/pull/3871/checks?check_run_id=4200111962. I don't know, libxml2 is available in the SDK.

For the Azure Pipeline, I cannot access the logs, the build as been removed: https://dev.azure.com/ldc-developers/ldc/_build/results?buildId=3189&view=logs&jobId=1bd59a18-c57a-55f0-9beb-7ebf5e9c6ff2.

I made a few inline comments as well.

.cirrus.yml
```diff
@@ -213,9 +213,9 @@ task:
     PARALLELISM: 12
     matrix:
       - TASK_NAME_SUFFIX: x64
-        EXTRA_CMAKE_FLAGS: "-DBUILD_LTO_LIBS=ON"
+        EXTRA_CMAKE_FLAGS: "-DD_COMPILER_FLAGS=-platformlib=pthread -DBUILD_LTO_LIBS=ON"
```
jacob-carlborg, line 216: You don't need to explicitly link with libpthread. libpthread is bundled in libSystem. libSystem is an umbrella/wrapper library that depends on may libraries. libpthread is symlink to libSystem so it is possible to link with libpthread if you really want to.

cmake/Modules/FindLLVM.cmake
```diff
@@ -138,8 +138,13 @@ else()
     llvm_set(LDFLAGS ldflags)
     llvm_set(SYSTEM_LIBS system-libs)
     string(REPLACE "\n" " " LLVM_LDFLAGS "${LLVM_LDFLAGS} ${LLVM_SYSTEM_LIBS}")
-    if(APPLE) # unclear why/how this happens
+    if(APPLE)
+        # unclear why/how this happens
         string(REPLACE "-llibxml2.tbd" "-lxml2" LLVM_LDFLAGS ${LLVM_LDFLAGS})
+        # remove libm (apparently not available for more recent Xcode versions)
+        string(REPLACE " -lm" "" LLVM_LDFLAGS ${LLVM_LDFLAGS})
```
jacob-carlborg, line 145: Same with libm as with libpthread.

cmake/Modules/FindLLVM.cmake
```diff
@@ -138,8 +138,13 @@ else()
...
+        # and libz too
+        string(REPLACE " -lz" "" LLVM_LDFLAGS ${LLVM_LDFLAGS})
```
jacob-carlborg, line 147: libz is its own separate library and needs to be explicitly linked.

jacob-carlborg:
I can add that Big Sur doesn't ship with the system provided dynamic libraries in the filesystem. They're instead present in a cache. This can cause scripts and build tools that try to identify if a library is present to fail. The errors I've seen doesn't seem to be caused by this, but just a heads up. The correct way is to use `dlopen`.

kinke:
Thx for trying to help; the linker was simply complaining about not finding a libm (`ld: library not found for -lxml2`), so I read about it being an alias to libSystem or something along these lines and thought 'okay, maybe Apple removed that alias, let's remove it'. Then it complained about not finding libz, and I went 'okay, let's try without that too'. Then it complained about libxml2, and I gave up. The 2 latter libs are needed for LLVM, and `dlopen` definitely no option.

jacob-carlborg:
You definitely need to link with libz and libxml2. Unfortunately I can't say anymore without doing a more thorough investigation on what's going on.

PR #3876, opened 2021-11-19
Add Void Linux to package manager list

Jlobblet:
Maintained at https://github.com/void-linux/void-packages/blob/master/srcpkgs/ldc/template


JohanEngelen:
Thanks

PR #3882, opened 2021-12-07
wasm: Add new default lld switches to protect from stackoverflow overwriting global memory

ryuukk:
I was stuck on trying to fix a bug on my code, whenever i was manipulating around large stucts, it'd change global data

I first thought of a memory corruption somewhere, but it turns out it was a stackoverflow issue!

Increasing the default stack (from 64kb) to 1mb fixed the issue... temporarily! as i was playing with more data, i hit the issue again.. so i looked around and found this PR on the zig repo: https://github.com/ziglang/zig/issues/4496

Tada! now instead of overwritting my global memory, it throws a runtime error! that's better! i can now debug the issue properly!

Let's save people days of headache and make these options the default!

Zig/Rust also bumped the default stack size to 1mb (emscripten has a stack size of 5mb)

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -250,14 +250,15 @@ elseif(${BUILD_SHARED_LIBS} STREQUAL "OFF")
     set(ADDITIONAL_DEFAULT_LDC_SWITCHES "${ADDITIONAL_DEFAULT_LDC_SWITCHES}\n        \"-link-defaultlib-shared=false\",")
 endif()
 
+
+# Default wasm stack is only 64kb, this is rather small, let's bump it to 1mb
+set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-L-z\", \"-Lstack-size=1048576\",")
+
+# Protect from stack overflow overwriting global memory
+set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-L--stack-first\",")
+
 if(LDC_WITH_LLD)
     set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-link-internally\",")
-
-    # Default wasm stack is only 64kb, this is rather small, let's bump it to 1mb
-    set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-L-z\", \"-Lstack-size=1048576\",")
-
-    # Protect from stack overflow overwriting global memory
-    set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-L--stack-first\",")
 endif()
 # LLD 8+ requires (new) `--export-dynamic` for WebAssembly (https://github.com/ldc-developers/ldc/issues/3023).
 if(NOT (LDC_LLVM_VER LESS 800))
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -254,10 +254,10 @@ if(LDC_WITH_LLD)
     set(WASM_DEFAULT_LDC_SWITCHES "${WASM_DEFAULT_LDC_SWITCHES}\n        \"-link-internally\",")
 
     # Default wasm stack is only 64kb, this is rather small, let's bump it to 1mb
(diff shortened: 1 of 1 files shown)


PetarKirov:
It's failing with:


```
FAIL: LDC :: baremetal/wasm2.d (17 of 308)
******************** TEST 'LDC :: baremetal/wasm2.d' FAILED ********************
Script:
--
: 'RUN: at line 6';   /var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/bin/ldc2 -mtriple=wasm32-unknown-unknown-wasm -betterC -w /private/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/cirrus-ci-build/tests/baremetal/wasm2.d -of=/private/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/tests/baremetal/Output/wasm2.d.tmp.wasm
: 'RUN: at line 7';   /var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/bin/ldc2 -mtriple=wasm32-unknown-unknown-wasm -betterC -w -fvisibility=hidden /private/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/cirrus-ci-build/tests/baremetal/wasm2.d -of=/private/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/tests/baremetal/Output/wasm2.d.tmp_hidden.wasm
: 'RUN: at line 10';   grep myExportedFoo /private/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/tests/baremetal/Output/wasm2.d.tmp.wasm
: 'RUN: at line 11';   grep myExportedFoo /private/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/tests/baremetal/Output/wasm2.d.tmp_hidden.wasm
--
Exit Code: 1
Command Output (stdout):
--
$ ":" "RUN: at line 6"
$ "/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/bin/ldc2" "-mtriple=wasm32-unknown-unknown-wasm" "-betterC" "-w" "/private/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/cirrus-ci-build/tests/baremetal/wasm2.d" "-of=/private/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/tests/baremetal/Output/wasm2.d.tmp.wasm"
# command stderr:
ldc2: Unknown command line argument '-z'.  Try: '/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/bin/ldc2 --help'
ldc2: Did you mean '-D'?
ldc2: Unknown command line argument '--stack-first'.  Try: '/var/folders/wh/z5_y2cv53sg24tzvtw_f_y1m0000gn/T/build/bin/ldc2 --help'
ldc2: Did you mean '--help-list'?
error: command failed with exit status: 1
```

So most like you have to use a different syntax to set these options for WASM.

Also, I think it would be cleaner if you set the options in the config file here:

https://github.com/ldc-developers/ldc/blob/5a28329e2f027ff2e3ea77a299f245e5baa4247c/ldc2.conf.in#L37-L40

Geod24:
`dub -v` will give you the command line DUB is using. From a quick look, those are lflags, so you might want to use `-L`.

JohanEngelen:
> Also, I think it would be cleaner if you set the options in the config file here:
> 
> https://github.com/ldc-developers/ldc/blob/5a28329e2f027ff2e3ea77a299f245e5baa4247c/ldc2.conf.in#L37-L40

Indeed. The current change in the CMakeFile will only help if LDC is built with LLD built-in. The case where an external LLD is used for linking, would still need manual adding of these flags.  If you add the flags to `ldc/ldc2.conf.in` directly, then that is solved.

kinke:
> or you add the flags unconditionally to the variable in CMakeFile. Perhaps better

Yep, exactly.

ryuukk:
it needs approval to trigger builds

kinke:
Yeah that's annoying, and IIRC, requires a new approval after each push. - Wrt. making LDC specify these flags without integrated LLD too, please just move the new lines right before the `if(LDC_WITH_LLD)` block (without empty lines to make clear this contiguous CMake part is all about setting up `WASM_DEFAULT_LDC_SWITCHES`).

ryuukk:
done!

kinke:
> without empty lines

Sorry about being so pedantic, but the CMake scripts are messy enough already. ;)

ryuukk:
Sorry about that! done!

kinke:
Thanks!

PR #3886, opened 2021-12-19
Merge upstream stable

kinke:


JohanEngelen:
Hi @kinke , do you intend to release `.0` and `.1` versions of LDC just like DMD in the future?  
I'm personally against releasing based on `.0` DMD source, because there are already enough bugs in the `.1` versions. (but it's easy for me to start ignoring `.0` releases, as long as I know ;))

kinke:
I'm not sure why the DMD release cycle has changed lately and what the new plan is; it's been > 2 months since v2.098.0. I didn't expect a 2.098.1 to be released anymore and was expecting a 2.099 this month. - I just don't want to postpone an LDC release for months waiting for the first point release of DMD.

That said, all (final) LDC releases are tested to work with Symmetry's code base on Linux and Windows, so they're lately definitely better tested than they used to. As a side note, a DMD regression inbetween v2.098.0-rc.1 and v2.098.0 final currently prevents us from using DMD v2.098.0 (using the rc1 instead). There's lately more action on upstream's stable branches AFAICT, and so the danger of regressions on that branch has increased.

PR #3887, opened 2021-12-20
Fix ABI build for FreeBSD on PowerPC64

lenoil98:
This fixes an issue when building on PowerPC64 BE on FreeBSD.  FreeBSD version 12 uses elfv1 but version 13 uses elfv2.  Changes allow LDC to build against the appropriate ABI for PowerPC.

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -75,10 +75,12 @@ static const char *getABI(const llvm::Triple &triple) {
         return "eabi";
       break;
     case llvm::Triple::ppc64:
-    case llvm::Triple::ppc64le:
       if (ABIName.startswith("elfv1"))
         return "elfv1";
       if (ABIName.startswith("elfv2"))
+        return "elfv2";
+      break;
+    case llvm::Triple::ppc64le:
         return "elfv2";
       break;
     default:
@@ -93,7 +95,10 @@ static const char *getABI(const llvm::Triple &triple) {
   case llvm::Triple::mips64el:
     return "n32";
   case llvm::Triple::ppc64:
-    return "elfv1";
+    if (ABIName.startswith("elfv1"))
+      return "elfv1";
+    if (ABIName.startswith("elfv2"))
+      return "elfv2";
   case llvm::Triple::ppc64le:
     return "elfv2";
   default:


driver/targetmachine.cpp
```diff
@@ -75,12 +75,14 @@ static const char *getABI(const llvm::Triple &triple) {
...
     case llvm::Triple::ppc64:
-    case llvm::Triple::ppc64le:
       if (ABIName.startswith("elfv1"))
         return "elfv1";
       if (ABIName.startswith("elfv2"))
         return "elfv2";
       break;
+    case llvm::Triple::ppc64le:
+        return "elfv2";
+      break;
```
kinke, line 85: This doesn't seem like an improvement - silently enforcing v2 for little-endian. Do you know whether ppc64le is *guaranteed* to be v2 for all OSs?

driver/targetmachine.cpp
```diff
@@ -93,7 +95,10 @@ static const char *getABI(const llvm::Triple &triple) {
   case llvm::Triple::mips64el:
     return "n32";
   case llvm::Triple::ppc64:
-    return "elfv1";
+    if (ABIName.startswith("elfv1"))
+      return "elfv1";
+    if (ABIName.startswith("elfv2"))
+      return "elfv2";
```
kinke, line 101: These 2 conditions will never yield true with the current code; the only real and breaking change is the new fall-through defaulting to v2 for all ppc64 targets. Is that really safe/desirable, or should that be FreeBSD-specific, potentially even restricted to a major version 13 or whatever?

JohanEngelen:
Note there is also something about ELF in target.d : https://github.com/ldc-developers/ldc/blob/ea19de1d1a22f58c924e5d3d79b7d6d1c8a02de3/dmd/target.d#L369-L370

target.d also contains code to find the FreeBSD version. So you will already have to specify the FreeBSD version in the triple to get correct compilation.
Related issue: https://issues.dlang.org/show_bug.cgi?id=22615

kinke:
> So you will already have to specify the FreeBSD version in the triple to get correct compilation.

LLVM should figure that out by itself by default; according to the Cirrus FreeBSD job, it uses `Default target: x86_64-portbld-freebsd12.2`.

> target.d also contains code to find the FreeBSD version.

We don't use that brittle/buggy logic and have https://github.com/ldc-developers/ldc/blob/ea19de1d1a22f58c924e5d3d79b7d6d1c8a02de3/driver/main.cpp#L814-L816 instead, so nothing to be done for LDC's compiler part to fix that unrelated issue.

JohanEngelen:
So the triple indeed already contains the OS version number. So that should be used for setting elfv1 or v2.

driver/targetmachine.cpp
lenoil98: In general little endian is only supported on power8 and newer processors, which is supported by OpenPower 64-bit ELFv2 ABI.  As far as I know,  only Linux and FreeBSD support little endian (ppc64le) and both are elfv2.  OpenBSD PowerPC64 release uses v2 but is big endian only.

driver/targetmachine.cpp
lenoil98: Okay I will rewrite and submit a revised version.

PR #3891, opened 2022-01-06
gen/target: refactor function to discover real type

ljmf00:
This patch refactors the real type discover function to use triple methods
instead of reinventing them.

### DIFF
--- a/gen/target.cpp
+++ b/gen/target.cpp
@@ -35,20 +35,17 @@ llvm::Type *getRealType(const llvm::Triple &triple) {
   auto &ctx = getGlobalContext();
 
   const auto a = triple.getArch();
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

ljmf00:
This seems supported only on recent versions of LLVM. Is LDC required to compile to LLVM 9 or older?

kinke:
Yes, we're currently supporting LLVM 6-12, as tested by CI.

ljmf00:
> Yes, we're currently supporting LLVM 6-12, as tested by CI.

Well, so I should stale this or close it?

kinke:
I guess close it, seems hardly worth it to special case this via `#if LDC_LLVM_VER`.

PR #3892, opened 2022-01-07
gen/target: make real 128bit on RISC-V targets

ljmf00:
RISC-V ILP32/LP64 ABI defines `long double` as 128bit.

Reference: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#cc-type-details


kinke:
I took the liberty of fixing this for LLVM < 9, refactoring the function while at it. The switch can be replaced by a terser `if (triple.isX86()) … else if …` chain once support for LLVM < 9 is dropped.

ljmf00:
The ultimate goal would be to use clang::TargetInfo, which gives you information about all target-specific types, sizes, etc... I didn't search in which version this is available, but I think the downside is the dependency to libclang. But if you do it statically, maybe it is not a big of an hassle.

PR #3893, opened 2022-01-07
Update ldmd2 to support -vasm

maxhaton:


maxhaton:
@kinke

thewilsonator (approved):

thewilsonator:
does DMD output the asm in addition to the binary? If so you might need additional `--output-*` arguments.

kinke:
According to Walter's PR description, it prints the asm (per function, so not the full object file) and doesn't output a dedicated file like `-output-s`. I'm not convinced LDC needs a similar functionality, given what it already supports.

maxhaton:
I suppose, but it doesn't hurt at all to add this to ldmd2. If there's an
easy way of mimicking the slightly bizarre way Walter chose to do it (he
didn't seem to be aware that GCC and LLVM have both supported this for
years) then I guess we could do that but I don't think there is. This is
better than nothing though + saves me some typing - angels die when you use
AT&T syntax.

maxhaton:
> According to Walter's PR description, it prints the asm (per function, so not the full object file) and doesn't output a dedicated file like `-output-s`. I'm not convinced LDC needs a similar functionality, given what it already supports.

I've made it mimic dmd a bit more. This doesn't add any functionality, just mimics dmd as best as is currently possible. I assume one could inject a hook somewhere to print the instructions properly, but who cares since it's just a dmd compatibility layer.

PR #3895, opened 2022-01-10
Add rdtscp to the dmd-style inline assembler.

maxhaton:
This meant adding a new pattern for the 0 args, EDX:EAX,ECX clobber set.


thewilsonator (approved):

maxhaton:
@thewilsonator merge?

PR #3899, opened 2022-01-15
-linkonce-templates: Cull 'discardable' template instances

kinke:
Such instances are guaranteed not to need codegen (e.g., CTFE only).

Extracted from #3860.


kinke:
C++ header addition upstreamed in https://github.com/dlang/dmd/pull/13581.

PR #3900, opened 2022-01-16
Output error code and message string upon cache file errors.

JohanEngelen:
This makes the error message look like this, for example:
```
Error: Could not create name of temporary file in the cache (errno 13: Permission denied)
                                                            \---------------------------/ this is the new part
```

### DIFF
--- a/driver/cache.cpp
+++ b/driver/cache.cpp
@@ -54,13 +54,20 @@
 
 #if LDC_POSIX
 #include <unistd.h>
+#include <errno.h>
 // Returns true upon error.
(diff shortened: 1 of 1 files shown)


driver/cache.cpp
```diff
@@ -54,13 +54,20 @@
 
 #if LDC_POSIX
 #include <unistd.h>
+#include <errno.h>
 // Returns true upon error.
```
kinke, line 58: Can kill this and the other 2 identical comments.

PR #3912, opened 2022-02-06
Simplify selection of IR fields for unions

kinke:
As the frontend has been more strict for some time.


kinke:
Oh no, the failing LLD 13 assertion on Win32 is back, with lld v13.0.1 now. Well, first time I see it again after a while, so another sporadic issue. Argh.

PR #3923, opened 2022-02-26
Fix #3916 - undefined symbols with `-dllimport=all` on Windows

### LINKED ISSUE #3916: [Windows] Unresolved symbol in shared static this.
```D
// Source/Game/Core/Math/package.d
module Game.Core.Math;
```

```D
// Source/Game/Core/Random.d

module Game.Core.Random;

import std.random : Xorshift, unpredictableSeed;

import Game.Core.Math;


private __gshared Xorshift engine;

shared static this()
{
    engine = Xorshift(unpredictableSeed);
}
```

Output from the build:
```
ldmd2 -m64 -g -shared -I../Build/Imports/ -JText/ -JData/ -mscrtlib=msvcrt -of../Build/Bin/Windows/runtime_out.dll Source\Game\Core\Random.d Source\Game\Core\Math\package.d -L/LIBPATH:../Build/Lib/Windows/x64 -L/PDB:../Build/Obj/runtime-20220214-113857.pdb OpenAL32.lib vulkan-1.lib SDL2.lib imgui.lib libvorbis.lib libvorbisfile.lib

runtime_out.obj : error LNK2019: unresolved external symbol _D3std6random__T14XorshiftEngineTkVki128Vii11ViN8ViN19ZQBn6__initZ referenced in function _D4Game4Core6Random24_sharedStaticCtor_L11_C1FZv
..\Build\Bin\Windows\runtime_out.dll : fatal error LNK1120: 1 unresolved externals
```

Turning it to a function makes it compile and resolve successfully, also needs to include the math package as the bare minimum.

kinke:
Instantiated data symbols were previously never dllimported with `-dllimport=all`. So if the parent template instance wasn't codegen'd into the binary directly, it remained undefined.

For `-dllimport=defaultLibsOnly`, the 'solution' to this problem was to define-on-declare data symbols instantiated from druntime/Phobos templates, making sure each binary defines all such symbols it references.

In both cases, switch to an approach where we dllimport all instantiated data symbols (or druntime/Phobos symbols only), and
dllexport them whenever defining them (so that other object files or binaries can import them). This may lead to more 'importing
locally defined symbol' linker warnings, but may also lead to less duplicates and possibly 'proper' sharing of instantiated globals
across the whole process.

This is superfluous and skipped with `-linkonce-templates`, as that mode defines all referenced instantiated symbols in each binary anyway, and so has already been a workaround.

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -1607,7 +1607,7 @@ DValue *DtoSymbolAddress(const Loc &loc, Type *type, Declaration *decl) {
   if (SymbolDeclaration *sdecl = decl->isSymbolDeclaration()) {
     // this is the static initialiser (init symbol) for aggregates
     AggregateDeclaration *ad = sdecl->dsym;
-    IF_LOG Logger::print("Sym: ad=%s\n", ad->toChars());
+    IF_LOG Logger::print("init symbol of %s\n", ad->toChars());
     DtoResolveDsymbol(ad);
     auto sd = ad->isStructDeclaration();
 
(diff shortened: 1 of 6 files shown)


kinke:
At work, we're using `-dllimport=defaultLibsOnly`, for a huge project consisting of ~140 dub packages/static libs. There are lots of 'importing locally defined symbol' linker warnings now, but I think they can be suppressed via `/warning:<number>`, need to test. And annoyingly, **one** symbol remains undefined - a `std.variant.VariantN!16` init symbol referenced by its TypeInfo.

PR #3924, opened 2022-02-27
Make -linkonce-templates less aggressive by default & add -linkonce-templates-aggressive

kinke:
Make `-linkonce-templates` *not* tamper with the general template emission algorithm anymore (so on top of default non-allinst or `-allinst` modes), and keep those tweaks as experimental `-linkonce-templates-aggressive`.

Compiling the druntime/Phobos unittests is only marginally slowed down compared to the more aggressive variant (~1.5% for debug, ~2.5% for release). It does show some rough 10% increase in required memory, but that's in line with non-linkonce-templates.

The more aggressive variant has the advantage of skipping `needsCodegen()` and potentially codegen'ing less symbols. The problem is that if an instantiated symbol isn't explicitly referenced, for instance a CRT ctor, it might not be codegen'd at all.

### DIFF
--- a/dmd/dsymbolsem.d
+++ b/dmd/dsymbolsem.d
@@ -5890,8 +5890,8 @@ void templateInstanceSemantic(TemplateInstance tempinst, Scope* sc, Expressions*
         }
 
         // LDC: the tnext linked list is only used by TemplateInstance.needsCodegen(),
-        //      which is skipped with -linkonce-templates
-        if (!(IN_LLVM && global.params.linkonceTemplates))
+        //      which is skipped with -linkonce-templates-aggressive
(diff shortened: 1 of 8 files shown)


kinke:
Runtimes and peak working sets (best-of-3) of `/usr/bin/time -v ninja -j8 druntime-ldc-unittest[-debug] phobos2-ldc-unittest[-debug]` on Linux x64, with an LLVM v13.0.1 without assertions:

| | debug | release |
| --- | --- | --- |
| normal | 0:30.54, 2.29 GB | 1:42.78, 2.12 GB |
| `-linkonce-templates` | 0:30.28, 2.29 GB | 1:17.86, 1.95 GB |
| `-linkonce-templates-aggressive` | 0:29.91, 2.19 GB | 1:17.34, 1.85 GB |

PR #3930, opened 2022-03-02
Fix @ldc.attributes.noSanitize compilation with `-w -de`.

JohanEngelen:


thewilsonator (approved):

kinke:
[CI is unfortunately pretty flaky at the moment - GitHub downloads (`curl --retry-all` is pretty recent IIRC and should make `--retry 3` work as expected), and the very annoying spurious lld 13 assertion on Win32, causing either a `debuginfo/msvc_dwarf.d` failure or crashes when linking LDC or its unittests itself.]

PR #3936, opened 2022-03-07
Prepare for LLVM15: Headers, attributes

thewilsonator:
`llvm::AttrBuilder` was changed to require an LLVMContext in its constructor, and can no longer be copied.

declare `DEBUG_TYPE` after includes for the GC2Stack pass as `llvm/Support/GenericDomTreeConstruction.h `(included from `llvm/IR/Dominators.h` `#define`s and `#undef`s it.

`llvm::DIBuilder::createExpression` now takes an `ArrayRef<uint64_t>` instead of `int64_t`. EmitLocalVariable has similarly changed

There is still to go the changing of load instructions to specify a type when loading, but that is significantly larger and more complicated. Will do that in a follow up PR.

### DIFF
--- a/ir/irfuncty.h
+++ b/ir/irfuncty.h
@@ -15,11 +15,6 @@
 #pragma once
 
 #include "gen/attributes.h"
-
-// "gen/llvmhelpers.h" declares `getGlobalContext`, but includes this header
-// so we can't #include it to rely on its definitions
-llvm::LLVMContext& getGlobalContext();
-
 #include "llvm/ADT/SmallVector.h"
 #include <vector>
 
--- a/ir/irfuncty.h
+++ b/ir/irfuncty.h
@@ -81,13 +81,8 @@ struct IrFuncTyArg {
    *  @param byref Initial value for the 'byref' field. If true the initial
    *               LLVM Type will be of DtoType(type->pointerTo()), instead
    *               of just DtoType(type) */
-  IrFuncTyArg(Type *t, bool byref,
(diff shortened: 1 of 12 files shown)


thewilsonator:
Ping @kinke

ir/irfuncty.h
```diff
@@ -76,7 +81,13 @@ struct IrFuncTyArg {
...
    *               LLVM Type will be of DtoType(type->pointerTo()), instead
    *               of just DtoType(type) */
-  IrFuncTyArg(Type *t, bool byref, llvm::AttrBuilder attrs = {});
+  IrFuncTyArg(Type *t, bool byref,
+#if LDC_LLVM_VER < 1500
+              llvm::AttrBuilder attrs = llvm::AttrBuilder {}
+#else
+              llvm::AttrBuilder attrs = llvm::AttrBuilder(getGlobalContext())
+#endif
+              );
```
kinke, line 90: Can't say that the API change looks like an improvement, sigh. - Maybe add a 2nd ctor taking 2 params only, to keep the ugliness in the .cpp only?

thewilsonator: Fixed

kinke: Thx. You should be able to remove the extra `getGlobalContext()` declaration in the .h then.

kinke (approved):

kinke:
Cheers!

thewilsonator:
Thanks for the reviews!

PR #3938, opened 2022-03-09
Fix -main regression when used for linking only, without source module

kinke:
This is a pretty wild one. `-main` handling was changed in v2.099 - the module is now optionally appended *after* full semantic for the source modules, so that it can be skipped if a `main` was already encountered.

Unlike regular source modules, it didn't get an entry in `global.params.objfiles` anymore this way though. So in non-singleobj
mode, its object file was ignored for archiving/linking. In singleobj mode, the problem was the case with no regular source
modules at all, the main module being the only module to be compiled (in singleobj mode) - where we assume `global.params.objfile[0]` to be the path to that object file.

Some meson build at Symmetry uses `-main` for linking separately compiled object files. The compiler then overwrote the first given object file with the compiled main module, and the linker later luckily complained about an undefined symbol, defined in the original first object file...

### DIFF
--- a/dmd/mars.d
+++ b/dmd/mars.d
@@ -689,7 +689,22 @@ version (IN_LLVM)
     }
 
     if (params.addMain && !global.hasMainFunction)
+    {
+version (IN_LLVM)
+{
+        auto mainModule = moduleWithEmptyMain();
+        modules.push(mainModule);
+
+        if (!params.oneobj)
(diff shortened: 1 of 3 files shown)


thewilsonator:
Can we push this change upstream?

kinke:
Seems like it's necessary, yes, as DMD v2.099.0 fails to link with `dmd -main foo.o` too (undefined `main`).

PR #3942, opened 2022-03-12
druntime: Simplify & somewhat accelerate getTLSRange() on Darwin

kinke:


kinke:
If this works, the shared libs support for macOS should be upstreamable without requiring DMD compiler changes. FYI @jacob-carlborg.

kinke:
A PITA without being able to test locally. Maybe the TLS block incl. `TLVDescriptor` are lazily allocated, so that it was with the previous `getTLSAnchor()` call but now isn't anymore.

jacob-carlborg:
> If this works, the shared libs support for macOS should be upstreamable without requiring DMD compiler changes

Even though this change won't require DMD compiler changes, this change will not make dynamic libraries with DMD, right?

jacob-carlborg:
> A PITA without being able to test locally.

I recommend setting up a VM with macOS. Here's some instructions: https://gist.github.com/MartinNowak/8156507.

kinke:
> this change will not make dynamic libraries with DMD, right?

Not alone; the according `rt.sections_elf_shared` modifications would need to be upstreamed of course, but that should be almost it. Perhaps still requires a tiny DMD adaptation to emit the `CompilerDSOData` struct and `d_dso_register` calls as on Linux, the only difference being a different mangle for the bracketing symbols, but at least no extra `getTLSAnchor()` function + TLS global in each object file anymore.

> I recommend setting up a VM with macOS.

If I really wanted to I'd have one, but I don't and definitely don't plan to.

PR #3946, opened 2022-03-16
Added support for newlib

### LINKED ISSUE #3945: LDC can't set some reserved version identifiers (CRuntime_*)
During the process of some experimental porting, I've noticed that there is a disconnect between how DMD handles triple parsing and how LDC does it. In particular, DMD allows appending a final '-cruntime' component to the triple, allowing specifications such as `arm-none-newlib` to appropriately set the CRuntime_Newlib version identifier. LDC appears to fall back on LLVM's triple parsing and cannot set arbitrary CRuntime version identifiers based directly on the triple.

This actually leads to the particularly frustrating situation where version identifiers such as CRuntime_Newlib cannot be set by LDC's triple parsing, but also can't be set by specific -d-version command line flags because it is still a reserved identifier; in fact the only way to actually set it is to hack it in to driver/main.cpp, such as the following diff:

```
diff --git a/driver/main.cpp b/driver/main.cpp
index 09e15e96..e764dcd0 100644
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -871,6 +871,10 @@ void registerPredefinedTargetVersions() {
       llvm::StringRef osName = triple.getOSName();
       if (osName.empty() || osName == "unknown" || osName == "none") {
         VersionCondition::addPredefinedGlobalIdent("FreeStanding");
+        /* Assume newlib; LLVM triples can't support *-none-newlib like
+         * DMD's triple parsing
+         */
+        VersionCondition::addPredefinedGlobalIdent("CRuntime_Newlib");
       } else {
         warning(Loc(), "unknown target OS: %s", osName.str().c_str());
       }
diff --git a/runtime/druntime b/runtime/druntime
--- a/runtime/druntime
+++ b/runtime/druntime
@@ -1 +1 @@
-Subproject commit 533e0178fc917bf56e8e791ab73731264bd4a2c3
+Subproject commit 533e0178fc917bf56e8e791ab73731264bd4a2c3-dirty
```

This is obviously not an ideal solution. There should be some way to get command-line support for CRuntime_Newlib (and various other reserved version identifiers) into LDC. The core issue is that LDC has inherited the full list of "reserved" version identifiers from DMD, yet lacks a of actually setting certain reserved identifiers implicitly based on platform specification due to limitations of LLVM's triple parsing. One approach would be relaxing LDC's enforcement of the restricted version identifiers so they can at least be finagled in by -d-version. A better fix would be to expand LDC's triple parsing to be capable of fully supporting DMD-style `[-cruntime]` suffixes - in fact, targeting parity with DMD for the triple:version identifiers mapping would probably be a reasonable goal. If this seems like a good approach, I can start working on a general solution to the problem and come up with an (actually useful) PR.

drahosj:
Fixes #3945 by handling newlib like uclibc.

Implemented similar to fixupUClibc. newlib will be removed from the
environment part of the triple, if present. The rest of the environment
is preserved (such as eabi). For example, the triple
arm-none-eabi-newlib will be converted to arm-none-eabi and 
CRuntime_Newlib will be set.

### DIFF
--- a/dmd/globals.d
+++ b/dmd/globals.d
@@ -316,6 +316,7 @@ version (IN_LLVM)
     // target stuff
     const(void)* targetTriple; // const llvm::Triple*
     bool isUClibcEnvironment;
+    bool isNewlibEnvironment;
 
     // Codegen cl options
     bool disableRedZone;
--- a/dmd/globals.h
+++ b/dmd/globals.h
@@ -284,6 +284,7 @@ struct Param
 
     const llvm::Triple *targetTriple;
     bool isUClibcEnvironment; // not directly supported by LLVM
+    bool isNewlibEnvironment; // not directly supported by LLVM
 
     // Codegen cl options
     bool disableRedZone;
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -595,6 +595,24 @@ void fixupUClibcEnv() {
   global.params.isUClibcEnvironment = true;
 }
 
+// Check for -newlib and strip out (supports -newlib and -eabi-newlib)
+void fixupNewlibEnv() {
+  llvm::Triple triple(mTargetTriple);
+  std::string envName(triple.getEnvironmentName());
+  size_t pos = envName.find("newlib");
+
+  if (pos == std::string::npos)
+    return;
+
+  envName.replace(pos, 6, "");
+  if (envName[envName.length() - 1] == '-') {
+      envName.replace(envName.length() - 1, 1, "");
+  }
+  triple.setEnvironmentName(envName);
+  mTargetTriple = triple.normalize();
+  global.params.isNewlibEnvironment = true;
+}
+
 /// Register the float ABI.
 /// Also defines D_HardFloat or D_SoftFloat depending if FPU should be used
 void registerPredefinedFloatABI(const char *soft, const char *hard,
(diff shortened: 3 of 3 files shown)


kinke:
On 32-bit ARM, triples looked like `armv6-linux-gnueabihf`, so I'd have expected `arm-none-newlibeabi`. Are there any precedents for `arm-none-eabi-newlib` you could link? Btw, you can use `ldc2 -v -mtriple=…` to check the normalized triple (a quadruple actually) - the last value in the `config` line (`arm-none-unknown-eabi-newlib` - `arm` arch, `none` vendor, `unknown` OS, `eabi-newline` environment).

JohanEngelen:
To me it seems best to upstream to LLVM (and only have a temporary fix in LDC if needed). This prevents the D triple name diverging from the rest of the world. The changes needed are very simple: see e.g. https://gist.github.com/crabtw/4481604

kinke:
I took the liberty of slightly refactoring the two env-patching functions to a single consistent one. Thanks for noticing the hassle wrt. having to specify a full quadruple (`arm-none-newlibeabi` not working, requiring sth. like `arm-none--newlibeabi`); string-matching the first `-{uclibc,newlib}` should be good enough indeed.

I've just checked the LLVM 14 source, apparently still no official uclibc support.

drahosj:
The refactor is definitely an improvement and keeps things relatively clean for what is ultimately a hack. While there may be some hassle with the triple parsing, having _any_ command-line/build-environment way of getting the CRuntime_Newlib version identifier set without having to use a patched version of LDC is a massive convenience for any runtime development or porting effort.

kinke:
> having any command-line/build-environment way of getting the CRuntime_Newlib version identifier set without having to use a patched version of LDC is a massive convenience for any runtime development or porting effort.

This is the price to pay for predefined version identifiers; the according logic (and FWIW, tests) preventing the user from setting one is in the shared DMD frontend, so nothing LDC specific. Initial ports with an arbitrary custom version are surely simpler to perform, and adding the predefined version later. FWIW, I've been annoyed by those checks a few times myself. ;)

PR #3951, opened 2022-03-30
Add support for LLVM 14

kinke:


JohanEngelen:
I don't know much about macos versions to be honest.
10.12 Sierra was released 6 years ago. https://en.wikipedia.org/wiki/MacOS_version_history#Version_10.12:_%22Sierra%22
I'd say that is acceptable for LDC. Including BOLT would be great.

kinke:
Thx, 10.12 it'll be then.

kinke:
Ouch, the LLVM 14 optimizer seems too aggressive. Derived from dmd-testsuite's `runnable/nested.d`:
```D
import core.stdc.stdio;

void foo14846(Dg)(scope Dg code)
{
    code();
}

void test14846()
{
    int x;

    struct S
    {
        this(int n) { x = n; }
        ~this() { printf("dtor: %p, %p\n", &this, &x); x = 99; }
    }

    foo14846({ S* p = new S(1); });
}

void main()
{
    test14846();
}
```

This is what optimized `main()` looks like:
```LL
define i32 @_Dmain({ i64, { i64, i8* }* } %unnamed) #5 {
  %.gc_struct.i.i = tail call noalias i8* @_d_newitemT(%object.TypeInfo* bitcast (%object.TypeInfo_Struct* @_D28TypeInfo_S3mod9test14846FZ1S6__initZ to %object.TypeInfo*)) #1 ; [#uses = 0]
  ret i32 0
}
```

So it does the GC allocation alone - without allocating the nested context, initializing the hidden `p.vthis` field accordingly, modifying `x` in the `S.this()` constructor etc. Most likely because it has no idea that the instance is destructed indirectly via the GC at program termination...

JohanEngelen:
Perhaps it is because of the `noalias` attribute that we put on the return value of `new`. Perhaps that means that noone else will meaningfully access the values put in that memory location, thus it is OK to remove stores to it, because the pointer does not escape our function --> data will never be used.
It's strange that it removes the `_d_allocmemory` completely, because that call could have side effects...

kinke:
Right on, Johan - removing the `noalias` attribute from `_d_newitemT` seems to be enough to 'fix' the testcase.

From https://llvm.org/docs/LangRef.html#parameter-attributes:
> On function return values, the noalias attribute indicates that the function acts like a system memory allocation function, returning a pointer to allocated storage disjoint from the storage for any other object accessible to the caller.

I think that should still apply here? I mean even malloc will need some book-keeping for the allocated chunks...

JohanEngelen:
> From https://llvm.org/docs/LangRef.html#parameter-attributes:
> 
> > On function return values, the noalias attribute indicates that the function acts like a system memory allocation function, returning a pointer to allocated storage disjoint from the storage for any other object accessible to the caller.
> 
> I think that should still apply here? I mean even malloc will need some book-keeping for the allocated chunks...

I think the reasoning is that `malloc` needs to do bookkeeping, but it does not need to access the contents of memory allocated for the user. So whatever the users stores in it, it is not used by `malloc`. Our GC indirectly _does_ access it in destructors.

kinke:
Yeah that makes sense. :+1: (plus the GC might access it during scanning)

PR #3955, opened 2022-04-08
CI: Bump some image versions and dependencies

kinke:


kinke:
Argh, looks like failing LLD assertions on Windows now also surface on Win64 with the newer Azure Windows image (and VS 2022). It's a different assertion than the spurious Win32 one (introduced with LLVM 13) though, and might even trigger consistently.

PR #3956, opened 2022-04-08
Drop compilability with ltsmaster

kinke:


kinke:
@ibuclaw: Do you know the actual bootstrap requirements for the frontend? The currently available GDC versions are still based on D v2.076 (and work for bootstrapping), so is it 2.076, or is there an exception for GDC? If the latter, what's the min GDC version?

ibuclaw:
> @ibuclaw: Do you know the actual bootstrap requirements for the frontend (I thought it was 2.079)? The currently available GDC versions are still based on D v2.076 (and work for bootstrapping), so is it 2.076, or is there an exception for GDC? If the latter, what's the min GDC version?

GDC 9.4 (i.e the latest minor version) is the recommended version for bootstrapping Linux. It's based on 2.076 (without static foreach though, that got added in GDC 10). For other platforms - in particular the non-x86 architectures for them - it can get a bit more hairy due to belated support being added for them.  The last version that uses the C++ front-end (GDC 11.x) has the best chance of working on whatever you choose to bootstrap from.

Last time I checked building dmd from the dmd-cxx branch (the upstream for GDC up to 11.x) was able to bootstrap master without problems on Linux.

kinke:
Thx Iain, I've added a note about GDC v9.4+. FWIW, the LDC CI here tests GDC v10 (Ubuntu 20) and v11 (Ubuntu rolling) at the moment, x86_64 only.

Issue #3864, opened 2021-11-04
ldc2 with -g flag fails to link on Apple M1 Pro (aarch64) with unaligned pointer


schveiguy:
I removed everything I could think of:

```console
% cat testit.d
extern(C) void main() {}
% ldc2 -betterC testit.d
% ./testit              
% ldc2 -betterC -g testit.d
ld: warning: pointer not aligned at address 0x10000402D (anon + 45 from testit.o)
ld: unaligned pointer(s) for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Error: /usr/bin/cc failed with status: 1
```

thewilsonator:
Some sleuthing reveals that `-platform_version macos 12.0.0 12.0` is the argument passed to the linker cause this to fail. On Big Sur (OS version 11)  `-platform_version macos 11.0.0 12.0` is passed instead. 

This latter command works, even with the same object file.

jacob-carlborg:
As a workaround, if I set the `MACOSX_DEPLOYMENT_TARGET` environment variable to `11` it will link successfully.

kinke:
I guess ld64 pedantically checks DWARF alignment specs and complains about an invalid one; I guess alignment of pointer itself vs. pointee alignment or something along these lines.

kinke:
Thx for testing, Steven. 2nd guess: https://github.com/ldc-developers/ldc/blob/5a28329e2f027ff2e3ea77a299f245e5baa4247c/driver/targetmachine.cpp#L40-L50

(testable via `-preserve-dwarf-line-section=false`)

schveiguy:
> 2nd guess

That seems to have worked. However, I now do not have file/line numbers as I did when targeting MacOS 11.

```console
steves@MacBook-Pro objdraw % export MACOSX_DEPLOYMENT_TARGET=11                                                                      
steves@MacBook-Pro objdraw % cat testthrow.d                   
void foo()
{
    throw new Exception("hi");
}

void main()
{
    foo();
}
steves@MacBook-Pro objdraw % ldc2 -g testthrow.d 
steves@MacBook-Pro objdraw % ./testthrow 
object.Exception@testthrow.d(3): hi
----------------
testthrow.d:3 void testthrow.foo() [0x1021d422b]
testthrow.d:8 _Dmain [0x1021d4237]
steves@MacBook-Pro objdraw % ldc2 -g -preserve-dwarf-line-section=false testthrow.d
[...]

PR #3958, opened 2022-04-09
Add universal prebuilt Mac package (x86_64 + arm64)

kinke:


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 #### Big news
 - LLVM for prebuilt packages bumped to v14.0.0. (#3952)
+- New prebuilt *universal* macOS package, runnable on both x86_64 and arm64, and enabling x86_64/arm64 macOS/iOS cross-compilation targets out of the box (`-mtriple={x86_64,arm64}-apple-{macos,ios12.0}`). The x86_64 package doesn't bundle any arm64 libs anymore; the arm64 package newly bundles iOS libs (arm64). (#3958)
```
kinke, line 5: This seems like less trouble than anticipated; according to CI, it's fine on x86_64 hosts. Some quick smoke tests on actual Apple Silicon (for all 4 targets) would be appreciated; if you have time, maybe also checking the arm64 package and that it can target iOS too. Pinging @p0nce, @dnadlinger, @schveiguy.

CI artifact(s) downloadable from https://dev.azure.com/ldc-developers/ldc/_build/results?buildId=3437&view=artifacts&pathAsName=false&type=publishedArtifacts.

dnadlinger: Seems to work fine for building a Phobos Hello World on 12.2.1/arm64, for both arm64 and x86_64 targets (with the 1.29.0-git-788b8ef binaries from your comment).

p0nce: All tests on macOS Monterey 12.0.1, Xcode 13.0

I'm testing intel-intrinsics in unitest-release mode, to avoid the -g bug in Monterey.

(Reference run with LDC 1.28 x86_64
`dub test -a x86_64-apple-macos -b unittest-release` => OK
`dub test -a arm64-apple-macos -b unittest-release` => OK)



## LDC is Universal Binary

`dub test -a x86_64-apple-macos -b unittest-release --compiler /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-universal/bin/ldc2` => OK

`dub test -a arm64-apple-macos -b unittest-release --compiler /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-universal/bin/ldc2` => OK

`dub test -a x86_64-apple-ios -b release --compiler /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-universal/bin/ldc2 ` => OK, successfully link and runs, `crash the with dyld[1918]: DYLD_ROOT_PATH not set for simulator program` which sounds alright.

`dub test -a arm64-apple-ios -b release --compiler /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-universal/bin/ldc2 ` => OK, successfully link and crash the unittests with error code 9, probably my fault

So yeah I guess it can become the one and only LDC on macOS :)
Also tests the now obsolete packages:


## LDC arm64 package

`dub test -a x86_64-apple-macos -b unittest-release --compiler /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-arm64/bin/ldc2` => OK, link failure as expected, as arm64 LDC doesn't have x86_64 target.

`dub test -a arm64-apple-macos -b unittest-release --compiler /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-arm64/bin/ldc2`
=> link failure, not sure why, not a concern if universal supersedes that
```
ld: warning: ignoring file /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-arm64/lib/libphobos2-ldc.a, building for macOS-x86_64 but attempting to link with file built for unknown-arm64
ld: warning: ignoring file /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-arm64/lib/libdruntime-ldc.a, building for macOS-x86_64 but attempting to link with file built for unknown-arm64
ld: warning: ignoring file .dub/build/intel-intrinsics-test-library-unittest-release-posix.osx.darwin-aarch64.arm_hardfloat-ldc_v1.29.0-git-788b8ef-6D0F504A78D2DED12452D86E33078DC9/intel-intrinsics-test-library.o, building for macOS-x86_64 but attempting to link with file built for unknown-arm64
  "_main", referenced from:
     implicit entry/start for main executable
```


## LDC is x86_64

dub test -a x86_64-apple-macos -b unittest-release --compiler /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-x86_84/bin/ldc2 => OK

dub test -a arm64-apple-macos -b unittest-release --compiler /Users/guill/Desktop/compilers/test/ldc2-788b8ef0-osx-x86_84/bin/ldc2 => OK, link failure, but expected since x86_64 package can't target arm64 anymore.

schveiguy: Well, this isn't working for me on OSX 12.3.1. Specifically, the -g flag doesn't work still (get linker errors).

Note also, with a previous LDC, if I set the environment variable: `MACOSX_DEPLOYMENT_TARGET=11` it worked. In this version, that workaround *doesn't work*. So I would not be able to use this binary.

The specific error is:

```
steves@MacBook-Pro testd % ~/Downloads/osx-universal/ldc2-788b8ef0-osx-universal/bin/ldc2 -g testbadlink.d
ld: warning: pointer not aligned at address 0x100008032 (anon + 50 from testbadlink.o)
ld: unaligned pointer(s) for architecture arm64
```

See #3864 for more details.

kinke: Thx for testing. - This doesn't affect/fix #3864 at all; if there are changes compared to before, that's either because of LLVM 14 or some further Xcode changes or whatever.

schveiguy: @jacob-carlborg any insights here? You were the one I think who figured out the workaround last time.

kinke:
@dnadlinger: Can you confirm Guillaume's trouble with the arm64 package? The `cc` invocation apparently defaulting to x86_64, not arm64 as expected. `ldc2.conf` hasn't changed for the macOS-arm64 target, so I'd have expected according bug reports for the existing arm64 package if linking wouldn't work.

Please also test the `arm64-apple-ios12.0` target with the arm64 package. And also the default targets for both universal and arm64 packages, without any `-mtriple` (should obviously default to macOS-arm64).

CHANGELOG.md
kinke: Off-topic, but there is another change for this CI artifact here compared to the latest releases - the macOS image for building LDC, druntime and Phobos was bumped from 10.15 to 11.x, and so the Xcode SDK version with it.

`MACOSX_DEPLOYMENT_TARGET` during the build was raised from 10.9 to 10.12 for the x86_64 package; still at 11.0 for the arm64 package.

One workaround is adding `-preserve-dwarf-line-section=false` to the cmdline / `ldc2.conf` file, implying no file/line infos in druntime exception backtraces. A debugger would presumably still offer those infos.

p0nce: 10.12 as target is more than OK, because once notarized a lot of software doesn't launch on 10.11 anyway except if you use a special flag.

jacob-carlborg: > any insights here? You were the one I think who figured out the workaround last time.

Unfortunately no. I think the best way forward, assuming the `-preserve-dwarf-line-section=false` flag works, is to read the debug info from the object files instead of the executable, that's what the debugger is doing. Debug info can also be compiled to its own .dSYM file for distribution. The best would be if all three methods were supported by druntime. @Geod24 is working on this, IIRC.

p0nce:
I didn't use the dub from the archive, so maybe related. Anyway, I don't think it's important given the new package and its reach.

schveiguy:
Oh yeah! I have used in the past an X86_64 dub while specifying the ldc2 compiler, and it doesn't work properly. And thinking about it, I can't imagine why that is true, unless rosetta does something weird when executing the linker from an x86 application?

dnadlinger:
@kinke:

> Can you confirm Guillaume's trouble with the arm64 package? 

Sorry, I'm a bit out of the loop here – what exactly is the issue? It appears to do what it should on a standard macOS 12.2 arm64 install (MacBook Pro M1 Max).

```
~ ❱ ~/Downloads/ldc2-788b8ef0-osx-universal/bin/ldc2 hello.d
~ ❱ file hello
hello: Mach-O 64-bit executable arm64
~ ❱ ./hello
~ ❱ rm hello
~ ❱ ~/Downloads/ldc2-788b8ef0-osx-arm64/bin/ldc2 hello.d
~ ❱ file hello
hello: Mach-O 64-bit executable arm64
~ ❱ ./hello
Hello world!
```

schveiguy:
> what exactly is the issue?

Almost certainly, the issue was with using an x86_64 dub executable. For what reason, I'm not sure, but I had the exact same experience a few months ago. Switching to the ldc2-provided dub package works (maybe @p0nce can confirm it works for him?)

p0nce:
If it work for everybody but for me, it is certainly DUB. I'm not home and I tested already so... there is probably no issue.

schveiguy:
> macOS will choose to run the x86_64 version of a child process binary if the (any?) parent is x86_64

Oooh! that explains it. Dub was running `ldc2` for linking, and when I copied the exact command, it worked (because now, I'm running it from the shell). I don't think dub expects to pass arch to the linker.

Is there a doc explaining this somewhere?

PR #3960, opened 2022-04-15
Raise min LLVM version to 9.0

kinke:
LDC v1.23.0 supports LLVM 6-10; we've been adding support for 4 new versions in the meantime, so drop some (3) again. [FreeBSD 12.2 uses LLVM 9.]


thewilsonator:
FYI this broke DMD's build kite job for LDC, complains it can't find LLVM of an appropriate version.

kinke:
=> https://github.com/dlang/ci/pull/455

PR #3970, opened 2022-04-26
Upgrade frontend & libs to v2.100.0+

kinke:


kinke:
The only remaining problem is a `sanitizers/asan_fakestack_GC.d` lit-test regression. `core.thread` wasn't touched; failing https://github.com/ldc-developers/ldc/blob/490bdcb0dd1c23f9552f116a3e9b68198fd36355/tests/sanitizers/asan_fakestack_GC.d#L44 might show a GC regression (the GC has been worked on).

kinke:
[That same test just failed in another PR: https://cirrus-ci.com/task/6100113155162112.]

JohanEngelen:
> The only remaining problem is a `sanitizers/asan_fakestack_GC.d` lit-test regression. `core.thread` wasn't touched; failing
> 
> https://github.com/ldc-developers/ldc/blob/490bdcb0dd1c23f9552f116a3e9b68198fd36355/tests/sanitizers/asan_fakestack_GC.d#L44
> 
> might show a GC regression (the GC has been worked on).
> **Edit**: Unlikely, as the test works fine without `SupportSanitizers`.

The `test_nulling_triggers_collection` function can be removed. https://github.com/ldc-developers/ldc/blob/490bdcb0dd1c23f9552f116a3e9b68198fd36355/tests/sanitizers/asan_fakestack_GC.d#L29
It is testing GC functionality (which is not _required_ to collect unused memory), and not fakestack functionality (done by `test_non_null_does_not_trigger_collection`).

kinke:
> The `test_nulling_triggers_collection` function can be removed. [...] It is testing GC functionality (which is not required to collect unused memory), and not fakestack functionality (done by test_non_null_does_not_trigger_collection).

Hmm, I still find it somewhat worrying that it now apparently never collects the unreferenced garbage anymore; no idea how the fakestack works, but is there a chance the nullified pointer is still non-nullified in some GC-scanned memory block? And would that be considered legal?

PR #3975, opened 2022-05-01
GH Actions / Cirrus CI: Use lld from host clang (without assertions) as default linker

kinke:
Instead of the lld from the prebuilt LDC-LLVM package, with enabled assertions for untagged CI builds. Besides slowing down linking, we also suffer from sporadic failing LLD assertions (since v13) for MSVC targets.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -359,8 +359,6 @@ task:
     mkdir -p llvm
     tar -xf llvm.tar.xz --strip 1 -C llvm
     rm llvm.tar.xz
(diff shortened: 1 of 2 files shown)


kinke:
Argh, this doesn't help with the LLD assertions on Windows for the lit-tests.

kinke:
I haven't seen those LLD assertions on Windows in a while. - Updated this PR nonetheless to not waste time for lld assertions, especially on the slow GH Actions CI runners.

Issue #3961, opened 2022-04-15
Port passes to LLVM's new pass manager


MaskRay:
The new pass manager has been on by default since LLVM 13.0.0. It is scheduled to be removed https://reviews.llvm.org/D109080

I plan to remove instrumentation passes from 15.0.0, which include at least PGO and sanitizers which are currently used by ldc.

thewilsonator:
Thanks for the heads up, we've just updated to LLVM14 for the latest release (~1 week ago).

MaskRay:
I have created https://reviews.llvm.org/D123834 which will remove `createPGOIndirectCallPromotionLegacyPass`.

I will likely create patches to remove sanitizers' legacy PM passes like `createAddressSanitizerFunctionPass`, `createMemorySanitizerLegacyPassPass`.

thewilsonator:
https://github.com/ldc-developers/ldc/pull/4105

PR #3976, opened 2022-05-06
Move to use new sanitizer passes

thewilsonator:
see #3961

I chose to enable them from LLVM13+ partly because that's when the old ones were deprecated and partly to ensure that the CI tests them. That can be easily changed if desired.

Still to go is replacement for `createPGOIndirectCallPromotionLegacyPass`.

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -38,6 +38,13 @@
 #if LDC_LLVM_VER >= 1000
 #include "llvm/Transforms/Instrumentation/SanitizerCoverage.h"
 #endif
+#define LDC_LLVM_VER_NEW_PASSES 1300
+#if LDC_LLVM_VER >= LDC_LLVM_VER_NEW_PASSES
+#include "llvm/Passes/OptimizationLevel.h"
+#include "llvm/Passes/PassBuilder.h"
+#include "llvm/Transforms/Scalar/EarlyCSE.h"
+#include "llvm/Transforms/Instrumentation/PGOInstrumentation.h"
+#endif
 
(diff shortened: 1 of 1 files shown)


kinke:
The sanitizers are probably just a minor aspect of switching to the 'new' pass manager (https://llvm.org/docs/NewPassManager.html). We also need to port *our* passes (https://forum.dlang.org/thread/mqcmbyvhdeoitcipyzkj@forum.dlang.org) before we can use it.

gen/optimizer.cpp
```diff
@@ -221,6 +229,73 @@ static void addSanitizerCoveragePass(const PassManagerBuilder &Builder,
...
+static void addSanitizers(PassBuilder &PB) {
+
+  auto addXSanPasses = [&](SanitizerBits mask, void (*passAdder)(const PassManagerBuilder &Builder,
+                                                           PassManagerBase &PM)) {
+    if (opts::isSanitizerEnabled(mask) {
+      builder.addExtension(PassManagerBuilder::EP_OptimizerLast, passAdder);
+      builder.addExtension(PassManagerBuilder::EP_EnabledOnOptLevel0, passAdder);
+    }
+  }
+  addXSanPasses(opts::AddressSanitizer,  addAddressSanitizerPasses);
```
JohanEngelen, line 243: did you check whether this puts the passes in the same order as with the old passmanager?

JohanEngelen:
I think it is better to make new/old passmanager selectable through a runtime option (commandflag) in the transition period. Easier to investigate bugs that way.

thewilsonator:
superseded by https://github.com/ldc-developers/ldc/pull/4105

PR #3980, opened 2022-05-13
GH Actions Linux: Build LDC with PGO'd frontend

kinke:


kinke:
Cool, seems to work. - Let's check more targets. :]

kinke:
No idea how much sense it makes for cross-compiled macOS-arm64 - based on the x86_64 PGO profile, resulting in slightly more than 100 `function control flow change detected (hash mismatch)` warnings. The number seems pretty low to me, so I guess it does make some sense at least, and shouldn't do any harm (well, CI times obviously suffer a bit, but it's not too bad IMO).

kinke:
Merging for beta1; I'll check the impact on compile speed later.

kinke:
Merging for beta1; I'll check the impact on compile speed later.

JohanEngelen:
> No idea how much sense it makes for cross-compiled macOS-arm64 - based on the x86_64 PGO profile, resulting in slightly more than 100 `function control flow change detected (hash mismatch)` warnings. 

Hm, I wonder what the mismatch is... Do we do a static-if on anything arm/x86 ?

JohanEngelen:
(nevertheless, this is benign of course)

kinke:
Perhaps some ABI diffs for args/return values, and/or `real_t` differences (double-precision vs. x87).

kinke:
Wrt. macOS arm64 hash mismatches - one contributor for non-tagged CI pipelines is that the x86_64 PGO build uses a native LLVM with enabled assertions, but the arm64 build doesn't (only no-assertions available). This controls the `-release` flag for the frontend and so assertions etc.

For the tagged v1.30.0-beta1 pipeline, the arm64 build still had 79 hash mismatches though.

PR #3982, opened 2022-05-15
GitHub Actions: Extend PGO to C++ parts of LDC

kinke:
[on top of #3980]


kinke:
* Doesn't work on macOS yet, because it's not using a host clang with the same LLVM version.
* Loads of hash mismatches on both Windows and Linux. The main difference between instrumented and PGO'd build is that the former does *not* use LTO at the moment. This wasn't an issue for the frontend, but that's compiled to a single object file.

kinke:
> Loads of hash mismatches on both Windows and Linux.

Using LTO for the instrumented binary too hasn't gotten rid of those, although the relevant CMake flags should be identical for both instrumented and PGO'd builds now. Perhaps due to vanilla clang 14.0.0 vs. LDC-LLVM v14.0.3?!

JohanEngelen:
> Ah, so `-fprofile-instr-*` != `-fprofile-*` for clang either

Sorry for not being able to help you with the issues. But yes, I did not reinvent anything new, the LDC commandline options are meant to do exactly the thing that clang does.

kinke:
Okay, that's sorted now too; those hash mismatches originated from the D parts of the ldc2-unittest executable, which don't use the PGO profiles anymore now.

kinke:
For some Symmetry code and a Linux debug build, I'm seeing some ~1% reduction in compile times over frontend-PGO only.

PR #3991, opened 2022-05-30
Revise DValue repainting

### LINKED ISSUE #3990: LDC crashing on compiling code in 32-bit mode
Add `-m32` or `--arch=x86`
```
struct Vector
{
    float x, y, z;
}

struct QAngle
{
    float x, y, z;

    QAngle opOpAssign(string op)(const(QAngle))
    {
        return this;
    }
}

void OnUserCmdPre()
{
    Vector ss;
    QAngle dd;
    dd -= cast(QAngle)ss;
}

void main() {}

```

Error message
```
#0 0x0000000005303243 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/ndf/dlang/ldc-1.30.0-beta1/bin/ldc2+0x5303243)
/home/ndf/dlang/ldc-1.30.0-beta1/bin/ldc2 failed with exit code -11.
```
Tested with LDC 1.30.0 beta1 and with LDC-nightly and they both crash.

kinke:
Fixes #3990.


thewilsonator (approved):

thewilsonator:
Ping, this good to go?

kinke:
I'm a bit unsure about `DFuncValue`; that looks ill-designed from the start.

kinke:
To be more precise: a bit unsure about whether non-delegate `DFuncValue`s can ever end up in here (and what to do if they do). Anyway, that unexpected case would be reported as a 'proper' fatal error now.

PR #3993, opened 2022-06-06
Add test for NVPTX intrinsics

thewilsonator:
cc @JohanEngelen 

https://github.com/ldc-developers/ldc/pull/2556/#issuecomment-1147265717

Whats the best way to do this for all targets that LDC supports?

### DIFF
--- a/tests/compilable/intrinsics.d
+++ b/tests/compilable/intrinsics.d
@@ -1,0 +1,5 @@
+// REQUIRES: target_NVPTX
+// RUN: %ldc -c %s
+
+import ldc.gccbuiltins_nvvm;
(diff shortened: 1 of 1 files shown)


JohanEngelen:
This test should include actual usage of an intrinsic (adding `dcompute_` as prefix to the filename would be nice too).
The idea is not to test the existence of the intrinsc file, but its correctness (for example, the test should catch what Kai forgot to add orginally: "The generated file will have to be marked @compute(CompileFor.deviceOnly), with the corresponding import of ldc.dcompute.").

Please make it part of #2556 (or add those changes to this PR).

PR #4007, opened 2022-06-28
RISCV64: Initial support

Ast-x64:
This adds support for riscv64gc with the default LP64D ABI.
This work with https://github.com/ldc-developers/druntime/pull/204 and https://github.com/ldc-developers/phobos/pull/71.
Enable attribute `+d` by default to support riscv64gc D standard extension.
ABI is implemented according to RISCV Calling Convention.

Currently it is not able to pass all tests.

Most failed tests are related to NaN payload issue. But this maybe a issue related to the platform. e.g., after `atan` calculation, NaN payload will be cleared. This behavior is the same with GDC. C codes compiled with GCC and Clang act the same.

The `zlib-debug` and `zlib-debug-shared` test failure is probably caused by a wrong signed/unsigned compiler optimization. This needs further investigation.

The `druntime-test-gc-*` test also fails on x86-64 according to my test. Failing assertion is the same one.

Many GDB tests and profdata tests fail. This needs further investigation.
Sanitizers seem to be unusable.
LTO should not be enabled because of soft-float and double-float merging issue. This error may be triggered with Clang too. This seems to be a common issue for LLVM-based compilers.

In `dmd-testsuite*`, `runnable/argufilem.d` fails with a comment inside explaining this may be a known issue. `dshell/dll_cxx.d` fails because `c++` on riscv64 do not accept argument `-m64`. `fail_compilation/test15703.d` seem to output some extra unexpected explanations for compilation error.

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -421,6 +421,7 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
   }
 
   if (triple.getArch() == llvm::Triple::riscv64 && !hasFeature("d")) {
+    // Support Double-Precision Floating-Point by default.
     features.push_back("+d");
   }
 
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -102,6 +102,8 @@ static const char *getABI(const llvm::Triple &triple) {
     return "elfv1";
   case llvm::Triple::ppc64le:
     return "elfv2";
+  case llvm::Triple::riscv64:
+    return "lp64d";
   default:
     return "";
   }
@@ -416,6 +418,10 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
   // if the user did not make an explicit choice.
   if (cpu == "x86-64" && !hasFeature("cx16")) {
     features.push_back("+cx16");
+  }
+
+  if (triple.getArch() == llvm::Triple::riscv64 && !hasFeature("d")) {
+    features.push_back("+d");
   }
 
   // Handle cases where LLVM picks wrong default relocModel
--- a/driver/tool.cpp
+++ b/driver/tool.cpp
@@ -120,6 +120,11 @@ void appendTargetArgsForGcc(std::vector<std::string> &args) {
     }
     return;
 
+  case Triple::riscv64:
+      args.push_back("-march=rv64gc");
+      args.push_back("-mabi=lp64d");
+    return;
+
   default:
     break;
   }
--- a/gen/abi-riscv64.cpp
+++ b/gen/abi-riscv64.cpp
@@ -23,21 +23,21 @@ struct Integer2Rewrite : BaseBitcastABIRewrite {
   }
 };
 
-struct flattenedFields {
-  struct flattenedField {
+struct FlattenedFields {
+  struct FlattenedField {
     Type *ty = nullptr;
     unsigned offset = 0;
   };
-  flattenedField fields[2];
+  FlattenedField fields[2];
   int length = 0; // use -1 to represent "no need to rewrite" condition
 };
 
-flattenedFields visitStructFields(Type *ty, unsigned baseOffset) {
+FlattenedFields visitStructFields(Type *ty, unsigned baseOffset) {
   // recursively visit a POD struct to flatten it
   // FIXME: may cause low performance
   // dmd may cache argtypes in some other architectures as a TypeTuple, but we
   // need to additionally store field offsets to realign later
-  flattenedFields result;
+  FlattenedFields result;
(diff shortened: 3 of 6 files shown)


driver/tool.cpp
```diff
@@ -120,6 +120,11 @@ void appendTargetArgsForGcc(std::vector<std::string> &args) {
     }
     return;
 
+  case Triple::riscv64:
+      args.push_back("-march=rv64gc");
+      args.push_back("-mabi=lp64d");
+    return;
+
```
kinke, line 127: Hmm, is that required even for a native gcc/clang?

kinke:
Cool, thanks for working on this!

driver/tool.cpp
Ast-x64: Thanks for your review. I tested it on a physical riscv board, and indeed this is the default switch (i.e., not required) for both gcc and clang.
In the standard specification, `LP64D` is *strongly recommended* to use on `RV64G`, but I haven't found standards showing that this is a forced behavior. So I guess the best practice is to specify these args explicitly.

driver/targetmachine.cpp
```diff
@@ -418,6 +420,10 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
     features.push_back("+cx16");
   }
 
+  if (triple.getArch() == llvm::Triple::riscv64 && !hasFeature("d")) {
+    features.push_back("+d");
+  }
```
JohanEngelen, line 425: can you add a comment why this is needed? Is this related to `-mabi=lp64d` ?

Ast-x64: In order to support riscv64gc (i.e., riscv64imacfdZicsrZifencei), we should enable D standard extension by default. This is also required to compile some asm codes in phobos.
A comment is added in https://github.com/ldc-developers/ldc/pull/4007/commits/299285e53dc2c29aa61dcf8d9428985a560df4bb.

kinke:
Wrt. ABI - is there any PDF or so that you could link?

And how did you approach this? Have you checked the C[++] ABI interop tests in dmd-testsuite and tried to fix those? And/or looked at generated clang IR and tried to replicate that?

gen/abi-riscv64.cpp
```diff
@@ -0,0 +1,234 @@
...
+    if (!isPOD(arg.type)) {
+      // non-PODs should be passed in memory
+      indirectByvalRewrite.applyTo(arg);
+      return;
+    }
+
+    if (isAggregate(ty) && ty->size() && ty->size() <= 16) {
+      if (ty->size() > 8 && DtoAlignment(ty) < 16) {
+        // pass the aggregate as {int64, int64} to avoid wrong alignment
+        integer2Rewrite.applyToIfNotObsolete(arg);
```
kinke, line 225: I guess this is not about alignment, but making it pass POD aggregates > 64 bits and <= 128 bits in 2 integer registers? Incl. e.g. `int[4]`, which would otherwise probably be passed like 4 scalar ints.

Ast-x64: Yes it's true. Maybe I did not explain it clearly. Those with 2*XLEN-bit alignment should be treated as an `int128`, while others should be treated as two `int64`. This matters when passing variadic arguments.

Ast-x64:
> Wrt. ABI - is there any PDF or so that you could link?
> 
> And how did you approach this? Have you checked the C[++] ABI interop tests in dmd-testsuite and tried to fix those? And/or looked at generated clang IR and tried to replicate that?

Thanks for your reply. I mainly implement this according to the [RISC-V Calling Conventions](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc), and I did fix some errors to pass all tests related to abi in `dmd-testsuite*`, by comparing verbose outputs with LLVM IRs generated from corresponding C codes using Clang.

JohanEngelen:
Is there anything blocking this?

kinke:
I haven't found time for a proper review yet. And a link to the spec as a code comment would be appreciated.

kinke:
Thanks @Ast-x64, much appreciated.

driver/targetmachine.cpp
zyedidia: Sorry to revive this conversation, but the most recent release of LDC broke a build of mine due to this change. It's not a big deal because I can just add `-mattr=-d` to fix it, but I'm not sure why this change needed to happen in the first place. Why should the D extension be enabled by default when the M, A, C, F, etc... extensions are not? Can't the D extension be manually enabled for compiling Phobos with `-mattr=+d`, just like all the other extensions?

PR #4016, opened 2022-08-06
Use `Record::getValueAsOptionalString()` when available

thewilsonator:
I get a link error with `getValue` with LLVM truck, not sure exactly when it was changed. `getValueAsOptionalString` was added in
https://github.com/llvm/llvm-project/commit/76419525fba62c93d5c337acdb0b80d6e42b00c9 tagged `llvmorg-12.0.0-rc1`


thewilsonator:
Hmm, maybe this was because I was using the wrong LLVM here,

PR #4017, opened 2022-08-07
Remove usage of `getPointerElementType` and `getPointeeType`

thewilsonator:
These no longer exist in LLVM15 as part of the opaque pointer migration.

### DIFF
--- a/gen/dvalue.cpp
+++ b/gen/dvalue.cpp
@@ -125,6 +125,10 @@ DLValue::DLValue(Type *t, LLValue *v) : DValue(t, v) {
          stripAddrSpaces(v->getType()) == DtoPtrToType(t));
 }
 
+llvm::Type *DLValue::memoryType() {
+    return DtoMemType(type);
+}
+
 DRValue *DLValue::getRVal() {
   if (DtoIsInMemoryOnly(type)) {
     llvm_unreachable("getRVal() for memory-only type");
--- a/gen/typinf.cpp
+++ b/gen/typinf.cpp
@@ -435,7 +435,7 @@ void buildTypeInfo(TypeInfoDeclaration *decl) {
     assert(isBuiltin && "existing global expected to be the init symbol of a "
                         "built-in TypeInfo");
   } else {
-    LLType *type = DtoType(decl->type)->getPointerElementType();
+      LLType *type = DtoType(decl->type)->getPointerElementType();
     // We need to keep the symbol mutable as the type is not declared as
     // immutable on the D side, and e.g. synchronized() can be used on the
     // implicit monitor.
--- a/ir/irclass.cpp
+++ b/ir/irclass.cpp
@@ -110,8 +110,7 @@ LLGlobalVariable *IrClass::getClassInfoSymbol(bool define) {
       emitTypeInfoMetadata(typeInfo, aggrdecl->type);
 
       // Gather information
-      LLType *type = DtoType(aggrdecl->type);
-      LLType *bodyType = type->getPointerElementType();
+      LLType *bodyType = llStructType;
       bool hasDestructor = (aggrdecl->dtor != nullptr);
       // Construct the fields
       llvm::Metadata *mdVals[CD_NumFields];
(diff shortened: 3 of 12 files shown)


thewilsonator:
There are still a few usages of `getPointeeType` in the load functions, but I'll leave those for another PR as their usage is quite pervasive.

gen/typinf.cpp
```diff
@@ -435,7 +435,7 @@ void buildTypeInfo(TypeInfoDeclaration *decl) {
     assert(isBuiltin && "existing global expected to be the init symbol of a "
                         "built-in TypeInfo");
   } else {
-    LLType *type = DtoType(decl->type)->getPointerElementType();
```
thewilsonator, line 438: So this doesn't work. Any Ideas?

ir/irclass.cpp
```diff
@@ -512,7 +511,7 @@ LLConstant *IrClass::getInterfaceVtblInit(BaseClass *b,
 
       llvm::GlobalVariable *interfaceInfosZ = getInterfaceArraySymbol();
       llvm::Constant *c = llvm::ConstantExpr::getGetElementPtr(
-          getPointeeType(interfaceInfosZ), interfaceInfosZ, idxs, true);
+            DtoType(getInterfacesArrayType()->nextOf()), interfaceInfosZ, idxs, true);
```
kinke, line 514: For globals, there's `getValueType()`.

gen/typinf.cpp
kinke: It's a class type, where `DtoType()` returns a IR pointer, but there's no `nextOf()` for the frontend type.

gen/dvalue.cpp
```diff
@@ -125,6 +125,10 @@ DLValue::DLValue(Type *t, LLValue *v) : DValue(t, v) {
          stripAddrSpaces(v->getType()) == DtoPtrToType(t));
 }
 
+llvm::Type *DLValue::memoryType() {
+    return DtoType(type);
```
kinke, line 129: `DtoMemType()` returns the proper type (i1 => i8).

gen/typinf.cpp
thewilsonator: Is `decl->type` always a class type in this branch? The function says "Builds all non-struct/class TypeInfos", but I guess that's referring to `decl->tinfo`.

gen/tocall.cpp
```diff
@@ -93,14 +93,13 @@ LLValue *DtoCallableValue(DValue *fn) {
...
+  Type *type = val->type;
+  if (type->ty == TY::Tfunction) {
+    return static_cast<LLFunctionType *>(DtoType(type));
   }
-  if (LLPointerType *pty = isaPointer(type)) {
-    if (LLFunctionType *fty = isaFunction(pty->getPointerElementType())) {
-      return fty;
-    }
+  else if (type->ty == TY::Tdelegate) {
+    return static_cast<LLFunctionType *>(DtoType(type->nextOf()));
```
thewilsonator, line 102: this is wrong, will look into it tomorrow. Any ideas appreciated.

gen/typinf.cpp
kinke: It's a `TypeInfo` subtype, so yeah, always a class.

thewilsonator: Ah, that is very obvious in hindsight. Thanks.

gen/tocall.cpp
kinke: I definitely wouldn't cast it statically, but use a LLVM dynamic cast to be notified appropriately. - Might have to care of function *pointers* here (and do a `toBasetype()`).

thewilsonator:
Hmm, I'm going to leave that `DtoCallableType` to a separate PR, to reduce surface area for unrelated failures and make some progress.

gen/typinf.cpp
```diff
@@ -435,7 +435,7 @@ void buildTypeInfo(TypeInfoDeclaration *decl) {
...
+    LLType *type = getIrType(decl->type)->isClass()->getMemoryLLType();
```
thewilsonator, line 438: So this still doesn't work

thewilsonator: Im going to defer this as well. I want to get some progress.

PR #4019, opened 2022-08-13
[rm-getPointerElementType] Remove usage of getPointerElementType in g…

thewilsonator:
…etClassInfoSymbol

see #4017

### DIFF
--- a/ir/iraggr.h
+++ b/ir/iraggr.h
@@ -61,6 +61,9 @@ public:
   llvm::Constant *getInitSymbol(bool define = false);
   /// Builds the __initZ initializer constant lazily.
   llvm::Constant *getDefaultInit();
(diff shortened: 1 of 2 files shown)


ir/iraggr.h
```diff
@@ -61,6 +61,9 @@ class IrAggr {
   llvm::Constant *getInitSymbol(bool define = false);
   /// Builds the __initZ initializer constant lazily.
   llvm::Constant *getDefaultInit();
+  /// Return the LLVM type of this Aggregate (w/o the reference for classes)
+  llvm::StructType *getLLStructType();
```
thewilsonator, line 65: This also needs to be used in `gen/dibuilder.cpp`, hence why this needs to be  public

PR #4021, opened 2022-08-13
Remove use of `getPointerElementType` in `gen/arrays.cpp`

thewilsonator:


gen/arrays.cpp
```diff
@@ -853,6 +833,27 @@ DSliceValue *DtoCatAssignArray(const Loc &loc, DValue *arr, Expression *exp) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+static LLValue *DtoSlicePtr(DValue *dval) {
```
thewilsonator, line 836: I moved this closer to its only user

thewilsonator (commented):
Yay finally! This is ready for review.

gen/arrays.cpp
JohanEngelen: it's now really hard to see what you changed, vs simply moved.

thewilsonator: Hmm, I must have left that in when trying to figure out the return type.
I can revert that back if desired, but the changes are:

* `s/type/i8arrty` and moved its definition to the point of use, past the early return, and 
* split off the `toElem(e)` into a separate function.

PR #4033, opened 2022-08-21
Remove `getPointeeType` from `tocall.cpp`

thewilsonator:


gen/tocall.cpp
```diff
@@ -30,6 +30,7 @@
 #include "ir/irfunction.h"
 #include "ir/irtype.h"
 #include "llvm/IR/LLVMContext.h"
+#include "llvm/IR/DerivedTypes.h"
```
thewilsonator, line 33: Need this for `LLType::IntegerType::get` I'm pretty sure we have a way to do this already somewhere...

PR #4035, opened 2022-08-21
[NFC] Use early return in jit compile `generateBind:genBind`

thewilsonator:


thewilsonator:
Trying to understand the web of `BindOverride` that should have is signature changed to accomodate the type to replace the `getPointerElementType`

PR #4038, opened 2022-08-21
Provide pointee type with gcc-asm to `createInlineAsmCall`

thewilsonator:


gen/irstate.cpp
```diff
@@ -276,7 +277,7 @@ IRState::createInlineAsmCall(const Loc &loc, llvm::InlineAsm *ia,
     if (constraintInfo.isIndirect) {
       call->addParamAttr(i, llvm::Attribute::get(context(),
                                                  llvm::Attribute::ElementType,
-                                                 getPointeeType(args[i])));
+                                                 types[i]));
```
thewilsonator, line 280:
```
%5 = call x86_fp80 asm sideeffect "fldt $1\0A\09fldt $2\0A\09fpatan ", "={st},*m,*m,~{st(1)},~{st(2)},~{st(3)},~{st(4)},~{st(5)},~{st(6)},~{st(7)},~{st}"(x86_fp80* elementtype(x86_fp80*) %3, x86_fp80* elementtype(x86_fp80*) %4), !srcloc !1
361
Attribute 'elementtype' type does not match parameter!
362
  %5 = call x86_fp80 asm sideeffect "fldt $1\0A\09fldt $2\0A\09fpatan ", "={st},*m,*m,~{st(1)},~{st(2)},~{st(3)},~{st(4)},~{st(5)},~{st(6)},~{st(7)},~{st}"(x86_fp80* elementtype(x86_fp80*) %3, x86_fp80* elementtype(x86_fp80*) %4), !srcloc !1
[363](https://github.com/ldc-developers/ldc/runs/7936964545?check_suite_focus=true#step:4:367)
in function _D3std4math12trigonometry8atan2AsmFNaNbNiNeeeZe
```

thewilsonator: Hmm, so I guess the types need to be one level of indirection less.

kinke: Yep, but those are only available & required for indirect operands. `GccAsmStatement_toIR()` has those infos for GCC-style inline asm; I think we only really support indirect operands there.

kinke: Erm nope, `atan2Asm()` has DMD-style inline asm.

thewilsonator: Indeed, I wasn't quite sure how that was done. It looks to be part of the constraint string that LLVM then parses. Are they readily available from somewhere else?

kinke: I'm afraid I don't think so, and the D types are probably only available a few levels further up...

PR #4039, opened 2022-08-21
Remove one use of `getPointeeType` in` DtoNestedVariable`

thewilsonator:


gen/nested.cpp
```diff
@@ -173,7 +173,7 @@ DValue *DtoNestedVariable(const Loc &loc, Type *astype, VarDeclaration *vd,
     // Handled appropriately by makeVarDValue() and EmitLocalVariable(), pass
     // storage of pointer (reference lvalue).
   } else if (byref || captureByRef(vd)) {
-    val = DtoAlignedLoad(getPointeeType(val), val);
+    val = DtoAlignedLoad(DtoType(vd->type)->getPointerTo(), val);
```
kinke, line 176: No getPointerTo.

thewilsonator: Hmm, I think I tried that before, I got an invalid bitcast which seems to be the cause of failure now. I'have another look tomorrow.

kinke: Oh right sorry, we load the actual pointer from the context field here.

PR #4041, opened 2022-08-21
Rework `DtoCallable` to remove `getPointerElementType`

thewilsonator:
I'm not at all sure about this. Any inputs welcome. (also idk if this is the best code style either)

### DIFF
--- a/gen/tocall.cpp
+++ b/gen/tocall.cpp
@@ -24,6 +24,7 @@
 #include "gen/llvmhelpers.h"
 #include "gen/logger.h"
 #include "gen/nested.h"
+#include "gen/mangling.h"
 #include "gen/pragma.h"
 #include "gen/tollvm.h"
 #include "gen/runtime.h"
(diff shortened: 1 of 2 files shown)


gen/tocall.cpp
```diff
@@ -874,9 +863,9 @@ DValue *DtoCallFunction(const Loc &loc, Type *resulttype, DValue *fnval,
   }
 
   // get callee llvm value
-  LLValue *callable = DtoCallableValue(fnval);
-  LLFunctionType *const callableTy =
-      DtoExtractFunctionType(callable->getType());
+  LLValue *callable;
+  LLFunctionType *callableTy;
```
kinke, line 867: Hmm, `callableTy = DtoType(tf)` might suffice here.

gen/tocall.cpp
```diff
@@ -72,41 +72,30 @@ TypeFunction *DtoTypeFunction(DValue *fnval) {
...
+    v = DtoRVal(fn);
+    f = DtoType(fn->type);
   }
   if (type->ty == TY::Tdelegate) {
     if (fn->isLVal()) {
       LLValue *dg = DtoLVal(fn);
       LLValue *funcptr = DtoGEP(dg, 0, 1);
-      return DtoLoad(funcptr, ".funcptr");
+      v = DtoLoad(funcptr, ".funcptr");
+      f = DtoType(type->nextOf());
```
kinke, line 86: missing early return

thewilsonator: I'll try this, see if it works, then try `callableTy = DtoType(tf)`, if that works use it if not revert back.

gen/tocall.cpp
thewilsonator: Actually on second thoughts, for a delegate that would return the delegate type, not the underlying type of the function pointer of the delegate.

kinke: ? It's already a `TypeFunction*`, so delegates are already handled by https://github.com/ldc-developers/ldc/blob/f142ac2e830b5a2865d8f5be8c34e5e2cd866069/gen/tocall.cpp#L46-L71.

thewilsonator: OK trying this

thewilsonator: `callableTy = isaFunction(DtoType(tf));` does not appear to work.

thewilsonator: trying `callableTy = irFty.funcType;`

thewilsonator:
OK this mostly works, except for `>>> TARGET FAILED: runnable/test22592.c`

thewilsonator:
So the problem seems to be that 
```c
int testfn(void);
...
int testfn() { ... }
``` 
is treated as 
```llvm
call i32 (...) @testfn()
```
instead of 
```llvm
call i32 @testfn()
```
(I'm guessing)

kinke:
The mismatch is in the frontend IIRC. This changed recently due K&R style functions support for importC; IIRC (fuzzy), a non-variadic fwd declaration can later get 'overwritten' by a variadic definition. See https://github.com/ldc-developers/ldc/blob/bd3d562d2a2eb41c2b8880473e3fb58484a471b6/gen/functions.cpp#L662-L668.

thewilsonator:
Thanks, I'll see what I can make of it tomorrow.

gen/tocall.cpp
```diff
@@ -877,10 +863,14 @@ DValue *DtoCallFunction(const Loc &loc, Type *resulttype, DValue *fnval,
   }
 
   // get callee llvm value
-  LLValue *callable = DtoCallableValue(fnval);
-  LLFunctionType *const callableTy =
-      DtoExtractFunctionType(callable->getType());
-  assert(callableTy);
+  LLValue *callable = DtoCallableValue(irFty.funcType, fnval);
+  LLFunctionType *callableTy = irFty.funcType;
+  if (dfnval && dfnval->func->isCsymbol()) {
```
thewilsonator, line 868: this could made probably quite a bit faster with a function signature check, will get around to that soon.

Issue #4047, opened 2022-08-27
static Windows build


89z:
Using this program:

~~~d
import stdio = std.stdio;

void main() {
   stdio.writeln("hello world");
}
~~~

and this command:

~~~
ldc2 app.d
~~~

I do not get a static build. Namely, the file `vcruntime140.dll` is required. So I tried this:

~~~
ldc2 -static app.d
~~~

same result. It seems LDC includes the static library:

~~~
    Directory: D:\ldc2\lib\mingw

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        2022-07-20  1:10 PM          32542 vcruntime140.lib
~~~

so it would appear it should be possible to do what I am trying to do.

kinke:
You'll need an MSVC++ installation for that; then linking the MSVCRT statically is the default behavior (except for DLLs). See https://github.com/ldc-developers/ldc/blob/master/packaging/README.txt#L10-L23.

89z:
I dont understand. If LDC has its own linker, and also includes the static library needed, why cant LDC just link the static library?

kinke:
It's a custom tiny CRT skeleton, based on the MSVC *DLLs*. If you really want to know more details, see https://github.com/ldc-developers/mingw-w64-libs (where the `vcruntime140.lib` is from).

89z:
OK fair enough, but it seems you could at least make the process simpler, for people who are in this position. For example, I dont see any links for people who want to download the Redistributable. Including that in the Readme or similar would be helpful to Windows users:

https://docs.microsoft.com/cpp/windows/latest-supported-vc-redist

kinke:
PRs are welcome. - Not sure what position you are talking about - I thought you wanted to link statically, that requires an VS/Build tools installation as mentioned in the readme. The redistributable only allows running the built stuff linked without VS.

89z:
https://github.com/ldc-developers/ldc/pull/4049

89z:
> PRs are welcome.

done

> Not sure what position you are talking about - I thought you wanted to link statically

I do

> that requires an VS/Build tools installation as mentioned in the readme.

right. Hopefully you understand, that Visual Studio is a giant bloated dependency. So given that, I would rather just deal with the DLL requirement.

> and AFAIK, comes with every non-ancient Windows

[...]

PR #4049, opened 2022-08-27
README.md: add Microsoft Visual C++ Redistributable

89z:


89z:
https://github.com/ldc-developers/ldc/issues/4047

kinke:
Sorry, I don't think that's the right place for it; `packaging/README.txt` would seem way more suitable to me. It's not required for the prebuilt binaries themselves (all linked statically), just for running the binaries in case there's no VS installation.

89z:
This is why I don't do PRs. You already know exactly what you want, but you want someone else to do it. Thanks for the comment, but I am done here.

PR #4050, opened 2022-08-28
Remove use of untyped `DtoLoad` in `dvalue.cpp`

thewilsonator:
- [x] TODO: fix dcompute addrspace pointers with this change

### DIFF
--- a/gen/dvalue.cpp
+++ b/gen/dvalue.cpp
@@ -118,7 +118,24 @@ bool DFuncValue::definedInFuncEntryBB() {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-DLValue::DLValue(Type *t, LLValue *v) : DValue(t, v) {
+DLValue::DLValue(Type *t, LLValue *v) :
+  DLValue(t, t->toBasetype()->ty == TY::Ttuple ?
+                nullptr :
+                DtoMemType(t),
+          v)
+{}
+
+DLValue::DLValue(Type *t, LLType * llt, LLValue *v) : DValue(t, v) {
+  lltype = llt;
+#if 1600 > LDC_LLVM_VER && LDC_LLVM_VER >= 1400
+  if (isSpecialRef())
+    assert(lltype == nullptr || isaPointer(val->getType())->isOpaque() || 
+           lltype == val->getType()->getPointerElementType()
+                                   ->getPointerElementType());
+  else
+    assert(lltype == nullptr ||
+           isaPointer(val->getType())->isOpaqueOrPointeeTypeMatches(lltype));
+#endif
   // v may be an addrspace qualified pointer so strip it before doing a pointer
   // equality check.
   assert(t->toBasetype()->ty == TY::Ttuple ||
@@ -130,8 +147,12 @@ DRValue *DLValue::getRVal() {
     llvm_unreachable("getRVal() for memory-only type");
     return nullptr;
   }
+  if (!lltype) {
+    llvm_unreachable("getRVal() for a tuple or noreturn");
+    return nullptr;
+  }
 
-  LLValue *rval = DtoLoad(val->getType()->getPointerElementType(), val);
+  LLValue *rval = DtoLoad(lltype, val);
 
   const auto ty = type->toBasetype()->ty;
   if (ty == TY::Tbool) {
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -950,6 +950,13 @@ public:
 
   //////////////////////////////////////////////////////////////////////////////
 
+  static llvm::PointerType * getWithSamePointeeType(llvm::PointerType *p, unsigned as) {
+#if LDC_LLVM_VER >= 1300
+    return llvm::PointerType::getWithSamePointeeType(p, as);
+#else
+    return p->getPointerElementType()->getPointerTo(as);
+#endif
+  }
   void visit(DotVarExp *e) override {
     IF_LOG Logger::print("DotVarExp::toElem: %s @ %s\n", e->toChars(),
                          e->type->toChars());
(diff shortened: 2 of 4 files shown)


gen/logger.cpp
```diff
@@ -59,7 +59,7 @@ void indent() {
 }
 
 void undent() {
-  if (_Logger_enabled) {
+  if (_Logger_enabled && !indent_str.empty()) {
```
thewilsonator, line 62: (this is only so that I can enable the logger wherever I please.)

gen/dvalue.cpp
```diff
@@ -131,7 +131,7 @@ DRValue *DLValue::getRVal() {
     return nullptr;
   }
 
-  LLValue *rval = DtoLoad(val);
+  LLValue *rval = DtoLoad(DtoMemType(type), val);
```
thewilsonator, line 134: `DtoMemType(type)` should always be `getPointeeType(val)` here. This is true, except when e.g.  
`DtoMemType(type) == float*` and `getPointeeType(val) == float addrspace(n)*`

but because the (D) type is `float*` and not e.g. `GlobalPointer!float` I've got no immediate way to fix this.

gen/toir.cpp
```diff
@@ -892,11 +889,22 @@ class ToElemVisitor : public Visitor {
...
+      _Logger_enabled = true;
+      IF_LOG Logger::println("PtrExp::toElem: %s @ %s", e->toChars(),
+                             e->type->toChars());
+      LOG_SCOPE;
 
     // get the rvalue and return it as an lvalue
-    LLValue *V = DtoRVal(e->e1);
-
-    result = new DLValue(e->type, DtoBitCast(V, DtoPtrToType(e->type)));
+    DValue *dv = toElem(e->e1);
```
thewilsonator, line 898: `dv` always seems to be an lvalue here, I'm not sure why we cast to a real and then make an lval out of it again, but trying to skip that seems to result in the unreachable at the top of `DLValue::getRVal` being triggered.

thewilsonator (commented):
I've hit a brick wall here

gen/dvalue.cpp
JohanEngelen: Two comments:
- shouldn't `DtoMemType` be called by the DLValue constructor and stored inside DLValue, and then you can simply use the stored value here? (eg. we store D and LLVM type in DLValue)
- alongside storing the type in DLValue, you can store the address space, just add it as optional constructor parameter?

thewilsonator:
So this works for dcompute but not for struct containing anon unions.

thewilsonator:
This work when I disable the commented out asserts. They fail for class/subclass pointer mismatches. I'd like some more feedback before I go ahead with this as is.

thewilsonator:
Hmm, this passes locally.

thewilsonator:
Splitting this up.

thewilsonator:
```
struct
Indexing aggregate field core.internal.lifetime.emplaceRef!(S[2][2][2], S[2][2][2], S).emplaceRef.S.payload:
* Value:   %4 = getelementptr inbounds %"core.internal.lifetime.emplaceRef!(S[2][2][2], S[2][2][2], S).emplaceRef.S", %"core.internal.lifetime.emplaceRef!(S[2][2][2], S[2][2][2], S).emplaceRef.S"* %0, i32 0, i32 0
	post
[2 x [2 x [2 x %core.lifetime.__unittest_L1086_C9.S]]]
arrllty[2 x [2 x [2 x %core.lifetime.__unittest_L1086_C9.S]]]
Building type: S[8]
tmp = [8 x %core.lifetime.__unittest_L1086_C9.S]*
e->type = S[8]
e->e1->type = S
DtoType(e->type) = [8 x %core.lifetime.__unittest_L1086_C9.S]
Assertion failed: (lltype == nullptr || isaPointer(val->getType())->isOpaqueOrPointeeTypeMatches(lltype))
```

gen/toir.cpp
```diff
@@ -938,8 +938,19 @@ class ToElemVisitor : public Visitor {
...
       // Logger::cout() << "mem: " << *DtoLVal(arrptr) << '\n';
-      result = new DLValue(e->type, DtoBitCast(DtoLVal(arrptr),
-                                               DtoPtrToType(e->type)));
+      LLType *arrllty = nullptr;
+      if (arrptr->lltype == nullptr) {
+        // noreturn
+      }
+      else if (llvm::PointerType *p = isaPointer(arrptr->lltype)) {
+        unsigned as = isaPointer(arrptr->lltype)->getAddressSpace();
+        arrllty = llvm::PointerType::getWithSamePointeeType(
```
thewilsonator, line 947: This is only available in LLVM13+

kinke:
Please don't merge this yourself; I'm not really a fan of adding a 2nd LL type to `DLValue`, I hope we can find a better solution.

JohanEngelen:
> Please don't merge this yourself; I'm not really a fan of adding a 2nd LL type to `DLValue`, I hope we can find a better solution.

Where is the second LLVM type? Or do you mean adding a llvm::Type at all? (a second type, next to the D type?)
The issue with opaque pointers is the the llvm::Value no longer carries any type information. So we need to store it somewhere else, hence the llvm::Type next to it. To me that makes sense. It effectively is the same situation we have pre LLVM15, where llvm::Value carries address and type. Or do I misunderstand?

thewilsonator:
> So we need to store it somewhere else, hence the llvm::Type next to it.

For non-dcompute we can derive `lltype` from `type` as `DtoMemType(type)` but for dcompute where address spaces may be involved, I don't see any other way of retaining the correct point type, because as Johan notes, the only type info `val` gives is "this is a pointer (maybe with an address space) to something"

Without this, it breaks when `typeof(val) == float addrspace(n)**` and using DtoMemType gives us `load float*` instead of `load float addrspace(n)*`

kinke:
> Where is the second LLVM type? Or do you mean adding a llvm::Type at all? (a second type, next to the D type?)

Oh right, forgot for a moment that it's the D type we have.

> but for dcompute where address spaces may be involved, I don't see any other way of retaining the correct point type

Have you tried detecting and handling the special dcompute pointer structs in `DtoMemType()`, lowering them to the address-space-qualified pointers directly?

thewilsonator:
> Have you tried detecting and handling the special dcompute pointer structs in DtoMemType(), lowering them to the address-space-qualified pointers directly?

the problem is not e,g. `GlobalPointer!float` which becomes `{ float addrspace(n)* }` its when you try to do `*gf.ptr = whatever` and the `DotVarExp`'s type is `float*` which `DtoMemType` has no way of knowing, and neither does the `llvm::Value` because all it knows is that it is a pointer.

This wouldn't be so bad if it could be isolated to just `DotVarExp` and `DtoIndexAggregate` but the result of `toElem` on the `DotVarExp` is a D(L)Value that is then used elsewhere , so that's not possible.

kinke:
Okay that makes sense. What about a special `DLValue` subclass for those `DotVarExp` of special dcompute structs? That's how I handled the bitfields.

thewilsonator:
Might be doable, I'll have to check.

PR #4063, opened 2022-08-28
Use typed GEP in `binops.cpp`

thewilsonator:


thewilsonator:
This is out by one level of indirection

PR #4071, opened 2022-08-28
Use typed GEP in `llvmhelpers.cpp`

thewilsonator:


thewilsonator:
This is mostly correct but breaks dcompute.

PR #4077, opened 2022-08-29
Use typed load/gep in `indexVThis`

thewilsonator:


gen/nested.cpp
```diff
@@ -46,7 +46,10 @@ LLValue *loadThisPtr(AggregateDeclaration *ad, IrFunction &irfunc) {
 }
 
 LLValue *indexVThis(AggregateDeclaration *ad,  LLValue* val) {
-    return DtoLoad(DtoGEP(val, 0, getVthisIdx(ad), ".vthis"));
+  llvm::StructType *st = getIrAggr(ad, true)->getLLStructType();
```
thewilsonator, line 49: this fails the already created assert  without the  `, true`

PR #4078, opened 2022-08-29
Use typed loads & geps in `nested.cpp`

thewilsonator:


gen/nested.cpp
```diff
@@ -171,9 +172,10 @@ DValue *DtoNestedVariable(const Loc &loc, Type *astype, VarDeclaration *vd,
   } else {
     // Load frame pointer and index that...
     IF_LOG Logger::println("Lower depth");
-    offsetToNthField(vardepth);
+    LLType *t = offsetToNthField(vardepth);
     IF_LOG Logger::cout() << "Frame index: " << *val << '\n';
     dereference((std::string(".frame.") + vdparent->toChars()).c_str());
+    currFrame = newframe;
```
thewilsonator, line 178: where do I get the type of the new frame from? the `t` above is a pointer type which doesn't help me.

PR #4105, opened 2022-09-03
Attempted to put the basic framework for new pass manger in place.

jamesragray:
This is related to https://github.com/ldc-developers/ldc/issues/3961

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -19,7 +19,6 @@
 #include "driver/targetmachine.h"
 #include "llvm/ADT/Triple.h"
 #include "llvm/Analysis/InlineCost.h"
-#include "llvm/Analysis/LoopPass.h"
 #include "llvm/Analysis/TargetLibraryInfo.h"
 #include "llvm/Analysis/TargetTransformInfo.h"
 #include "llvm/IR/DataLayout.h"
@@ -38,9 +37,11 @@
 #include "llvm/Transforms/Instrumentation/ThreadSanitizer.h"
 #include "llvm/Transforms/Instrumentation/AddressSanitizer.h"
 #include "llvm/Transforms/Instrumentation/AddressSanitizerOptions.h"
+#if LDC_LLVM_VER >= 1400
 #include "llvm/Transforms/Scalar/EarlyCSE.h"
 #include "llvm/Transforms/Scalar/LICM.h"
 #include "llvm/Transforms/Scalar/Reassociate.h"
+#endif
 #if LDC_LLVM_VER >= 1000
 #include "llvm/Transforms/Instrumentation/SanitizerCoverage.h"
 #endif
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -429,6 +429,21 @@ bool legacy_ldc_optimize_module(llvm::Module *M) {
 
 #if LDC_LLVM_VER >= 1400
 
+static OptimizationLevel getOptimizationLevel(){
+  switch(optimizeLevel) {
+    case 0: return OptimizationLevel::O0;
+    case 1: return OptimizationLevel::O1;
+    case 2: return OptimizationLevel::O2;
+    case 3:
+    case 4:
+    case 5: return OptimizationLevel::O3;
+    case -1: return OptimizationLevel::Os;
+    case -2: return OptimizationLevel::Oz;
+  }
+  //This should never be reached
+  return OptimizationLevel::O0;
+}
+
 static void addAddressSanitizerPasses(ModulePassManager &mpm,
                                       OptimizationLevel level ) {
   AddressSanitizerOptions aso;
@@ -595,29 +610,7 @@ void runOptimizationPasses(llvm::Module *M) {
 //  builder.populateFunctionPassManager(fpm);
 //  builder.populateModulePassManager(mpm);
 
-  OptimizationLevel level;
-  switch(optimizeLevel) {
-    case 0:
-      level = OptimizationLevel::O0;
-      break;
-    case 1:
-      level = OptimizationLevel::O1;
-      break;
-    case 2:
-      level = OptimizationLevel::O2;
-      break;
-    case 3:
-    case 4:
-    case 5:
-      level = OptimizationLevel::O3;
-      break;
-    case -1:
-      level = OptimizationLevel::Os;
-      break;
-    case -2:
-      level = OptimizationLevel::Oz;
-      break;
-  }  
+  OptimizationLevel level = getOptimizationLevel();
 
   if (optLevelVal == 0) {
     mpm = pb.buildO0DefaultPipeline(level, opts::isUsingLTO() || opts::isUsingThinLTO());
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -120,8 +120,8 @@ static cl::opt<int> fSanitizeMemoryTrackOrigins(
     cl::desc(
         "Enable origins tracking in MemorySanitizer (0=disabled, default)"));
 
-static cl::opt<signed char> passmanager("passmanger",
-    cl::desc("Setting the passmanger (new,legacy):"), cl::ZeroOrMore, cl::init(1),
+static cl::opt<signed char> passmanager("passmanager",
+    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(1),
     cl::values(
         clEnumValN(0, "legacy", "Use the legacy passmanager (available for LLVM14 and below) "),
         clEnumValN(1, "new", "Use the new passmanager (available for LLVM14 and above)")));
@@ -476,27 +476,26 @@ static void addMemorySanitizerPass(FunctionPassManager &fpm,
   if (level != OptimizationLevel::O0) {
     fpm.addPass(EarlyCSEPass());
     fpm.addPass(ReassociatePass());
-//FIXME: Can't figure this out yet
-//    fpm.addPass(LICMPass());
+    fpm.addPass(createFunctionToLoopPassAdaptor(LICMPass()));
     fpm.addPass(GVNPass());
-//FIXME: Ditto
-//    fpm.addPass(InstructionCombiningPass());
-//    fpm.addPass(DeadStoreEliminationPass());
+    //FIXME: Not sure what to do with these?
+    //fpm.addPass(InstructionCombiningPass());
+    //fpm.addPass(DeadStoreEliminationPass());
   }
 }
-static void addThreadSanitizerPass(ModulePassManager &mpm, FunctionPassManager &fpm,
+static void addThreadSanitizerPass(ModulePassManager &mpm,
                                       OptimizationLevel level ) {
   mpm.addPass(ModuleThreadSanitizerPass());
-  fpm.addPass(ThreadSanitizerPass());
+  mpm.addPass(createModuleToFunctionPassAdaptor(ThreadSanitizerPass()));
 }
 
-static void addSanitizerCoveragePass(ModulePassManager &mpm, FunctionPassManager &fpm,
+static void addSanitizerCoveragePass(ModulePassManager &mpm,
                                       OptimizationLevel level ) {
   mpm.addPass(ModuleSanitizerCoveragePass(
       opts::getSanitizerCoverageOptions()));
 }
 // Adds PGO instrumentation generation and use passes.
-static void addPGOPasses(ModulePassManager &mpm, FunctionPassManager &fpm,
+static void addPGOPasses(ModulePassManager &mpm,
                                       OptimizationLevel level ) {
   if (opts::isInstrumentingForASTBasedPGO()) {
     InstrProfOptions options;
@@ -513,7 +512,7 @@ static void addPGOPasses(ModulePassManager &mpm, FunctionPassManager &fpm,
   }
 }
 
-static void addStripExternalsPass(ModulePassManager &mpm, FunctionPassManager &fpm,
+static void addStripExternalsPass(ModulePassManager &mpm,
                                       OptimizationLevel level ) {
 
   if (level == OptimizationLevel::O1 || level == OptimizationLevel::O2 ||
@@ -532,8 +531,8 @@ static void addStripExternalsPass(ModulePassManager &mpm, FunctionPassManager &f
   }
 }
 
-static void addSimplifyDRuntimeCallsPass(ModulePassManager &mpm, FunctionPassManager &fpm,
-                                      LoopPassManager &lpm, OptimizationLevel level ) {
+static void addSimplifyDRuntimeCallsPass(ModulePassManager &mpm,
+                                      OptimizationLevel level ) {
   if (level == OptimizationLevel::O2  || level == OptimizationLevel::O3) {
 //FIXME: Update and uncomment this once gen/passes/SimplifyDRuntimeCalls.cpp is updated to
 // work with the new pass manager.
@@ -544,8 +543,8 @@ static void addSimplifyDRuntimeCallsPass(ModulePassManager &mpm, FunctionPassMan
   }
 }
 
-static void addGarbageCollect2StackPass(ModulePassManager &mpm, FunctionPassManager &fpm,
-                                      LoopPassManager &lpm, OptimizationLevel level ) {
+static void addGarbageCollect2StackPass(ModulePassManager &mpm,
+                                         OptimizationLevel level ) {
   if (level == OptimizationLevel::O2  || level == OptimizationLevel::O3) {
 //FIXME: Update and uncomment this once gen/passes/GarbageCollect2Stack.cpp is updated to
 // work with the new pass manager.
@@ -640,14 +639,8 @@ void runOptimizationPasses(llvm::Module *M) {
 
   ModulePassManager mpm;
 
-  // Also set up a manager for the per-function passes.
-  FunctionPassManager fpm;
-
-  // Also set up manger for the per-loop passes.
-  LoopPassManager lpm;
-
   if (!noVerify) {
-    fpm.addPass(VerifierPass());
+    mpm.addPass(createModuleToFunctionPassAdaptor(VerifierPass()));
   }
 
   if (opts::isSanitizerEnabled(opts::AddressSanitizer)) {
@@ -660,21 +653,23 @@ void runOptimizationPasses(llvm::Module *M) {
   if (opts::isSanitizerEnabled(opts::MemorySanitizer)) {
     pb.registerOptimizerLastEPCallback([&](ModulePassManager &mpm,
                                         OptimizationLevel level) {
+      FunctionPassManager fpm;
       addMemorySanitizerPass(fpm,level);
+      mpm.addPass(createModuleToFunctionPassAdaptor(std::move(fpm)));
     });
   }
 
   if (opts::isSanitizerEnabled(opts::ThreadSanitizer)) {
     pb.registerOptimizerLastEPCallback([&](ModulePassManager &mpm,
                                         OptimizationLevel level) {
-      addThreadSanitizerPass(mpm, fpm,level);
+      addThreadSanitizerPass(mpm, level);
     });
   }
 
   if (opts::isSanitizerEnabled(opts::CoverageSanitizer)) {
     pb.registerOptimizerLastEPCallback([&](ModulePassManager &mpm,
                                         OptimizationLevel level) {
-      addSanitizerCoveragePass(mpm, fpm,level);
+      addSanitizerCoveragePass(mpm,level);
     });
   }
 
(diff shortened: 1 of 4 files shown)


gen/optimizer.cpp
```diff
@@ -148,44 +159,54 @@ llvm::CodeGenOpt::Level codeGenOptLevel() {
...
 
   if (verifyEach) {
     pm.add(createVerifierPass());
   }
 }
+#endif
 
-static void addStripExternalsPass(const PassManagerBuilder &builder,
+#if LDC_LLVM_VER < 1500
+static void legacyAddStripExternalsPass(const PassManagerBuilder &builder,
```
thewilsonator, line 173: Given the preprocessor directives are going to be confusing as it is, I'd recommend having one block for _all_ the legacy passes, i.e. get rid of the above `#endif` and `#if LDC_LLVM_VER < 1500` for all except the top most `#if LDC_LLVM_VER < 1500` and bottom most `#endif`

Similarly when you get around to it, one preprocessed version block for the new pass function

jamesragray: Okay I will do that.

jamesragray: I have done this now.

gen/optimizer.cpp
```diff
@@ -110,6 +110,17 @@ static cl::opt<int> fSanitizeMemoryTrackOrigins(
...
+    "use-new-pass-manager", cl::ZeroOrMore,
+     cl::desc("Use the new pass manger instead of the old"));
+
+static cl::opt<signed char> passmanager(
+    cl::desc("Setting the passmanger (new,legacy):"), cl::ZeroOrMore, cl::init(0),
+    cl::values(
+        clEnumValN(0, "legacy", "Use the legacy passmanager (available with LLVM14 and below) "),
+        clEnumValN(1, "new", "Use the new passmanager (available LLVM14 and above)")),
+    cl::init(0));
+
```
JohanEngelen, line 123: Why did you add two commandline switches for choosing pass manager?
You forgot to give `passmanager` a commandline switch name `--... ?`, and duplicated `cl::init(0)`.

JohanEngelen:
Thanks for working on this!

gen/optimizer.cpp
jamesragray: Thank you for the feedback.. I removed the first one (it was what I thought to use first and forgot to remove).  I removed the duplicated cl::init(0) and added commandline switch name.

gen/optimizer.cpp
```diff
@@ -389,7 +395,233 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+  if (verifyEach) {
+    mpm.addPass(VerifierPass());
+  }
+}
+*/
+static void addAddressSanitizerPasses(ModulePassManager &mpm,
+                                      OptimizationLevel level ) {
+  AddressSanitizerOptions aso;
+  //FIXME: Where do we get these options from
+  aso.CompileKernel = false;
```
thewilsonator, line 439: https://github.com/ldc-developers/ldc/blob/master/gen/optimizer.cpp#L190-L192
https://github.com/ldc-developers/ldc/blob/master/gen/optimizer.cpp#L108-L111

gen/optimizer.cpp
```diff
@@ -389,7 +395,233 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+
+  // Report that we run some passes.
+  return true;
+}
+#endif
+
+#if LDC_LLVM_VER >= 1400
+//FIXME: Something wrong with this:
+/*template <typename PassT>
+static inline void addPass<PassT>(ModulePassManager &mpm, PassT &pass) {
```
thewilsonator, line 427:
```diff
template <typename PassT>
-static inline void addPass<PassT>(ModulePassManager &mpm, PassT &pass)
+static inline void addPass(ModulePassManager &mpm, PassT &pass)
```

gen/optimizer.cpp
```diff
@@ -389,7 +395,233 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+  //FIXME: Where can we find these
+  Optional<llvm::CodeModel::Model> cm = None;
+  std::string featuresStr;
+  Optional<llvm::Reloc::Model> rm = None;
+  llvm::TargetOptions options;
+  //FIXME: Not sure that this works in all cases
+  CodeGenOpt::Level optLev = (CodeGenOpt::Level) optLevelVal;
+                                            
+
+  tm = target->createTargetMachine(triple, "", featuresStr, options, rm, cm, optLev);
```
thewilsonator, line 478: You should use https://github.com/ldc-developers/ldc/blob/c5e4501eebd2755502994521ea39bcce49ba73de/gen/irstate.h#L43

gen/optimizer.cpp
```diff
@@ -21,6 +22,7 @@
 #include "llvm/Analysis/InlineCost.h"
 #include "llvm/Analysis/TargetLibraryInfo.h"
 #include "llvm/Analysis/TargetTransformInfo.h"
+#include "llvm/MC/TargetRegistry.h"
```
thewilsonator, line 25: I think this is no longer needed if you use `gTargetMachine`, but if it is, you will need to guard it with `#if LDC_LLVM_VER >= 1400`

gen/optimizer.cpp
```diff
@@ -13,6 +13,7 @@
 #include "gen/cl_helpers.h"
 #include "gen/logger.h"
 #include "gen/passes/Passes.h"
+#include "llvm/Passes/PassBuilder.h"
```
thewilsonator, line 16: please keep the include list sorted, put this after `#include "llvm/LinkAllPasses.h"`

gen/optimizer.cpp
```diff
@@ -414,6 +663,24 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+// or new version using the new pass managr
+// Returns true if any optimization passes were invoked.
+bool ldc_optimize_module(llvm::Module *M) {
+#if LDC_LLVM_VER < 1400
+  return legacy_ldc_optimize_module(M);
+#elif LDC_LLVM_VER < 1500
+  return passmanager==0 ? legacy_ldc_optimize_module(M)
+                        : new_ldc_optimize_module(M);
+#else
+  return ldc_optimize_module(M);
```
thewilsonator, line 679: This should be `return new_ldc_optimize_module(M);`

gen/optimizer.cpp
jamesragray: Thanks. Still not sure if and how fSanitizeMemoryTrackOrigins fits in here.

thewilsonator: Hmm, I guess that's one of `aso.UseAfterScope` or `aso.UseAfterReturn`
For now I'd go with if `fSanitizeMemoryTrackOrigins` is set, then set 
```
aso.UseAfterScope = true;
aso.UseAfterReturn = AsanDetectStackUseAfterReturnMode.Always;
```

We can come back to this if that is undesirable.

gen/optimizer.cpp
jamesragray: Thanks haven't written a template in c++ for a few years. Still can't get that to work though.

gen/optimizer.cpp
jamesragray: This is a bad error. Sorry about that.

gen/optimizer.cpp
thewilsonator: > mpm.addPass(pass);

should be `addPass(mpm, pass);` no UFCS ;(

gen/optimizer.cpp
thewilsonator: no problems, caught by `-Winfinte-recursion`, not by me scratching my head.

gen/optimizer.cpp
```diff
@@ -413,6 +419,252 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+  }
+}
+
+static void addAddressSanitizerPasses(ModulePassManager &mpm,
+                                      OptimizationLevel level ) {
+  AddressSanitizerOptions aso;
+  aso.CompileKernel = false;
+  aso.Recover = false;
+  aso.UseAfterScope = true;
+  aso.UseAfterReturn = AsanDetectStackUseAfterReturnMode::Always;
```
JohanEngelen, line 441: `UseAfterScope` and `UseAfterReturn` should both be off. It's a bit surprising to see these options here, because you usually set them by setting an environment variable. Perhaps that has changed? Or maybe this is just additional such that you can turn them on without env variable. Regardless, LDC (druntime) does not have full support for these options yet.

JohanEngelen: **Edit2**:  `llvm::AsanDetectStackUseAfterReturnMode::Runtime` is probably the one that we want.

jamesragray: Thanks. I changed it as suggested.

gen/optimizer.cpp
```diff
@@ -389,7 +400,272 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+    case 5:
+      level = OptimizationLevel::O3;
+      break;
+    case -1:
+      level = OptimizationLevel::Os;
+      break;
+    case -2:
+      level = OptimizationLevel::Oz;
+      break;
+  }
```
thewilsonator, line 619: You should split this off into a separate function

gen/optimizer.cpp
```diff
@@ -389,7 +400,265 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+//Run optimization passes using the new pass manager
+void runOptimizationPasses(llvm::Module *M) {
+  // Create a ModulePassManager to hold and optimize the collection of
+  // per-module passes we are about to build.
+  
+  unsigned optLevelVal = optLevel();
+  unsigned sizeLevelVal = sizeLevel();
+  
+
+  PipelineTuningOptions pto;
```
thewilsonator, line 509: since `PipelineTuningOptions` is plain old data and its construction is essentially `pure`, you might want to break the out into a separate function. You can probably then do `PassBuilder pb(gTargetMachine, getPipleineTuningOptions(arguments as appropriate));`

gen/optimizer.cpp
```diff
@@ -110,6 +117,12 @@ static cl::opt<int> fSanitizeMemoryTrackOrigins(
     cl::desc(
         "Enable origins tracking in MemorySanitizer (0=disabled, default)"));
 
+static cl::opt<signed char> passmanager("passmanger",
+    cl::desc("Setting the passmanger (new,legacy):"), cl::ZeroOrMore, cl::init(0),
```
thewilsonator, line 121: when you think this is close to usable, change this `cl::init(0)` to `cl::init(1)`, to get all the appropriate CI (i.e. with LLVM version >= 14) to test it, but then set it back prior to merging.

gen/optimizer.cpp
```diff
@@ -389,7 +400,265 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+    case 0: return OptimizationLevel::O0;
+    case 1: return OptimizationLevel::O1;
+    case 2: return OptimizationLevel::O2;
+    case 3:
+    case 4:
+    case 5: return OptimizationLevel::O3;
+    case -1: return OptimizationLevel::Os;
+    case -2: return OptimizationLevel::Oz;
+  }
+  //This should never be reached
```
thewilsonator, line 442: there is `llvm_unreachable("reason");` for that.

gen/optimizer.cpp
jamesragray: Put it to 1 for testing now.

gen/optimizer.cpp
```diff
@@ -389,7 +402,269 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+                                      OptimizationLevel level ) {
+  mpm.addPass(ModuleSanitizerCoveragePass(
+      opts::getSanitizerCoverageOptions()));
+}
+static PipelineTuningOptions getPipelineTuningOptions(unsigned optLevelVal, unsigned sizeLevelVal) {
+  PipelineTuningOptions pto;
+
+  pto.LoopUnrolling = optLevelVal > 0;
+
+  pto.LoopUnrolling = !((disableLoopUnrolling.getNumOccurrences() > 0)
```
thewilsonator, line 501: Silly question: why is this unconditionally assigned to back to back?

jamesragray: No reason I just tried to keep it the same logic as the original version:
  builder.DisableUnrollLoops = optLevel == 0;

  builder.DisableUnrollLoops = (disableLoopUnrolling.getNumOccurrences() > 0)
                                   ? disableLoopUnrolling
                                   : optLevel == 0;

gen/optimizer.cpp
```diff
@@ -413,6 +427,391 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+                     debugInfoForProfiling, pseudoProbeForProfiling);
+  } else if (opts::isUsingIRBasedPGOProfile()) {
+    return PGOOptions(global.params.datafileInstrProf, "", "",
+                     PGOOptions::PGOAction::IRUse,
+                     PGOOptions::CSPGOAction::NoCSAction,
+                     debugInfoForProfiling, pseudoProbeForProfiling);
+  }
+  return PGOOptions("", "", "",
+                    PGOOptions::PGOAction::NoAction,
+                    PGOOptions::CSPGOAction::NoCSAction,
```
thewilsonator, line 576: Looks like LLVM asserts that these values are not passes to the constructor.
You should make this function return a `llvm::Optional<getPGOOptions>` and return `Optional::None` for this case where PGO is not enabled.

jamesragray: Thanks should have realized why clang does that.

gen/optimizer.cpp
```diff
@@ -413,6 +428,381 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+
+  // MemorySanitizer inserts complex instrumentation that mostly follows
+  // the logic of the original code, but operates on "shadow" values.
+  // It can benefit from re-running some general purpose optimization passes.
+  if (level != OptimizationLevel::O0) {
+    fpm.addPass(EarlyCSEPass());
+    fpm.addPass(ReassociatePass());
+    fpm.addPass(createFunctionToLoopPassAdaptor(LICMPass()));
+    fpm.addPass(GVNPass());
+    //FIXME: Not sure what to do with these?
```
JohanEngelen, line 481: We should look at what Clang does here (that's probably where I copied it from originally).

gen/optimizer.cpp
```diff
@@ -413,6 +428,381 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+    }
+  }
+
+  pb.registerOptimizerLastEPCallback([&](ModulePassManager &mpm,
+                                      OptimizationLevel level) {
+    addStripExternalsPass(mpm, level);
+  });
+  OptimizationLevel level = getOptimizationLevel();
+
+  addPGOPasses(mpm, level);
```
JohanEngelen, line 698: you are going to assign (overwrite) to `mpm` later, so this line has to move after that.

JohanEngelen: check where Clang adds the PGO passes. We want to add them before optimization passes (not all the way at the end).

jamesragray: Yes, you are of course correct.

gen/optimizer.cpp
jamesragray: I think the relevant part is here: https://github.com/llvm/llvm-project/blob/8d2a447bf92700e576e037c5f6495f74153cf5e9/clang/lib/CodeGen/BackendUtil.cpp#L648

They seem to have dropped some of the passes. The old passes are here: https://github.com/llvm/llvm-project/blob/ce5379f0f0675592fd10a522009fd5b1561ca72b/clang/lib/CodeGen/BackendUtil.cpp#L333

gen/optimizer.cpp
jamesragray: I checked and I think clang does it at the start. I will do the same.

jamesragray:
Not sure what to do about the tests in: /runtime/phobos/std/math/hardware.d
When I disable all custom passes and only leave only those parses coming from here the test still fails for -O1 -O2 -O3: https://github.com/ldc-developers/ldc/blob/388f322dc9a9b1fa049dbac990b7f780421f897c/gen/optimizer.cpp#L709

JohanEngelen:
> Not sure what to do about the tests in: /runtime/phobos/std/math/hardware.d When I disable all custom passes and only leave only those parses coming from here the test still fails for -O1 -O2 -O3:

You can place the unittest in a standalone file, and compile with both pass managers and look at -output-s output (assembly output). You can also try and check with --print-after (print llvm output after every optimizer step) for more in-depth checking of where the two (new/legacy) pass pipelines diverge.

I see that the unittest is marked with `@optStrategy("none")`. Perhaps good to check and verify that indeed the behavior is the same for both pass pipelines.

jamesragray:
I copy here what I hope is a relevant part of the output of compiling : /runtime/phobos/std/math/hardware.d with -O3
For context they both produce:
```
; [#uses = 1]
; Function Attrs: noinline optnone uwtable
define void @_D3std4math8hardware18__unittest_L337_C7FNfZv() local_unnamed_addr #2 {
  %1 = alloca %std.math.hardware.IeeeFlags, align 4 ; [#uses = 2, size/byte = 4]
  %2 = alloca %std.math.hardware.IeeeFlags, align 4 ; [#uses = 2, size/byte = 4]
  %3 = alloca %std.math.hardware.IeeeFlags, align 4 ; [#uses = 2, size/byte = 4]
```
Followed for the legacy passmanger by 
```
  %f = alloca %std.math.hardware.IeeeFlags, align 4 ; [#uses = 2, size/byte = 4]
  %4 = alloca %std.math.hardware.IeeeFlags, align 4 ; [#uses = 2, size/byte = 4]
  %5 = alloca %std.math.hardware.IeeeFlags, align 4 ; [#uses = 2, size/byte = 4]
  store x86_fp80 0xK4000E000000000000000, x86_fp80* %a, align 16
  call void @_D3std4math8hardware14resetIeeeFlagsFNbNiNeZv() #9
  %6 = call %std.math.hardware.IeeeFlags @_D3std4math8hardware9ieeeFlagsFNaNbNdNiNeZSQBpQBoQBm9IeeeFlags() #9 ; [#uses = 1]
  store %std.math.hardware.IeeeFlags %6, %std.math.hardware.IeeeFlags* %1, align 4
  %7 = call zeroext i1 @_D3std4math8hardware9IeeeFlags9divByZeroMxFNbNdNiNfZb(%std.math.hardware.IeeeFlags* nonnull %1) #5 ; [#uses = 1]
  %8 = icmp eq i1 %7, false                       ; [#uses = 1]
  br i1 %8, label %assertPassed, label %assertFailed
  ```
while the new pass manager produces:
```
  tail call void @_D3std4math8hardware14resetIeeeFlagsFNbNiNeZv() #5
  %4 = tail call %std.math.hardware.IeeeFlags @_D3std4math8hardware9ieeeFlagsFNaNbNdNiNeZSQBpQBoQBm9IeeeFlags() #5 ; [#uses = 1]
  %.fca.0.extract17 = extractvalue %std.math.hardware.IeeeFlags %4, 0 ; [#uses = 1]
  %.fca.0.gep18 = getelementptr inbounds %std.math.hardware.IeeeFlags, %std.math.hardware.IeeeFlags* %1, i64 0, i32 0 ; [#uses = 1, type = i32*]
  store i32 %.fca.0.extract17, i32* %.fca.0.gep18, align 4
  %5 = call zeroext i1 @_D3std4math8hardware9IeeeFlags9divByZeroMxFNbNdNiNfZb(%std.math.hardware.IeeeFlags* nonnull %1) #0 ; [#uses = 1]
  br i1 %5, label %assertFailed, label %assertPassed1
````
I don't know llvm ir well but yo me it looks like the new pass manager code is optimizing even with noopt?

**Comment by thewilsonator** [comment]
The presence of `tail call` and `extractvalue` suggests to me that the new pass manager is indeed performing at leat _some_ optimisations.

**Comment by jamesragray** [comment]
I cut down the size of hardware.d but made sure that assertion failure was still there and tested with --print-after-all. What I found is that firstly the legacy passmanager seems to have several extra passes, but after those passes the first difference appears in:
*** IR Dump After Pre-ISel Intrinsic Lowering (pre-isel-intrinsic-lowering) ***

File: gen/optimizer.cpp [hunk 18]
```diff
@@ -413,6 +428,390 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+  fpm.addPass(MemorySanitizerPass(
+      MemorySanitizerOptions{trackOrigins, recover, kernel}));
+
+  // MemorySanitizer inserts complex instrumentation that mostly follows
+  // the logic of the original code, but operates on "shadow" values.
+  // It can benefit from re-running some general purpose optimization passes.
+  if (level != OptimizationLevel::O0) {
+    fpm.addPass(EarlyCSEPass());
+    fpm.addPass(ReassociatePass());
+    fpm.addPass(createFunctionToLoopPassAdaptor(LICMPass()));
```
thewilsonator, line 479: This line doesn't compile for me with LLVM15, it complains it needs `LICMOpts` or 2 unsigned and a bool`.

I've put `LICMPass(128,128,false)` arbitrarily as a placeholder in my local copy of this branch.
```
unsigned LicmMssaOptCap,
unsigned LicmMssaNoAccForPromotionCap,
bool LicmAllowSpeculation
```

gen/optimizer.cpp
```diff
@@ -110,6 +120,12 @@ static cl::opt<int> fSanitizeMemoryTrackOrigins(
     cl::desc(
         "Enable origins tracking in MemorySanitizer (0=disabled, default)"));
 
+static cl::opt<signed char> passmanager("passmanager",
+    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(1),
```
thewilsonator, line 124: given there is only one failing test not accounted for (`std.math.hardware`) i think its fine to merge this once you change the `cl::init(1)` back to `cl::init(0)`. 
That would enable you to 
* work on porting the custom passes independently 
* add and LLVM15 CI
* fix the `optnone` issue causing the `std.math.hardware`

independently.

Thanks for all the hard work you've put into this!

thewilsonator:
Superseded by https://github.com/ldc-developers/ldc/pull/4175

PR #4107, opened 2022-09-04
Remove `getPointerElementType` from DLLImportRelocation.cpp

thewilsonator:


gen/passes/DLLImportRelocation.cpp
```diff
@@ -188,18 +188,21 @@ struct Impl {
...
       if (i <= 0xFFFFFFFFu) {
         address = b.CreateConstInBoundsGEP2_32(t, address, 0,
                                                static_cast<unsigned>(i));
       } else {
         address = b.CreateConstInBoundsGEP2_64(t, address, 0, i);
       }
+      if (StructType *st = dyn_cast<StructType>(t))
+        t = st->getElementType(i);
+      else if (PointerType *pt = dyn_cast<PointerType>(t))
+        assert(0 && "Have to give this some more thought");
```
thewilsonator, line 202: Hmm, I think I also need to account for `ArrayType` here too.

thewilsonator:
Is this pass actually tested by the CI or am I just getting lucky here?

kinke:
It is minimally tested by a few lit-tests at least. It should be absolutely safe to assume only IR structs and arrays of constant initializers are being GEP'd.

thewilsonator:
Thanks. OK, I'll clean up the assert message and then merge this

PR #4112, opened 2022-09-04
Make `DtoIndexAggregate` return a `DLValue`

thewilsonator:


kinke:
Oh no, this won't work with bit fields.

thewilsonator:
Ah. because it returns an Lvalue? I had just assumed it would always be an Lvalue, but I don't think I require it to be.

thewilsonator:
I've had a look at #4015, this doesn't look to be _completely_ add odds with the implementation there. 
It should be possible to have the return type of DtoIndexAggregate return a `DValue` which can be either a `DLValue` or a bitfield.

If you could pull out some of the refactoring bits of that PR e.g. `void visit(DotVarExp *e)` that would be good, I'll leave #4050 open for now and we can hopefully get both working.

kinke:
Yeah, I was merely 'complaining' about the increasing merge conflicts in the bitfields PR.

PR #4121, opened 2022-09-07
Provide indirect types to `createInlineAsmCall`

thewilsonator:


gen/asmstmt.cpp
```diff
@@ -676,6 +692,14 @@ void CompoundAsmStatement_toIR(CompoundAsmStatement *stmt, IRState *p) {
   std::vector<LLValue *> args;
   args.insert(args.end(), out.args.begin(), out.args.end());
   args.insert(args.end(), in.args.begin(), in.args.end());
+  std::vector<LLType *> indirectTypes;
+  indirectTypes.insert(indirectTypes.end(),
+                       out.indirectTypes.begin(),
+                       out.indirectTypes.end());
+  indirectTypes.insert(indirectTypes.end(),
+                       in.indirectTypes.begin(),
+                       in.indirectTypes.end());
```
thewilsonator, line 701: this segfaults somehow, will look into tomorrow.

thewilsonator: maybe its because its empty...

thewilsonator: No.

thewilsonator: Trying a reserve to avoid the realloc call?

thewilsonator: Segfault is only on linux (glibc). Windows and macOS don't crash.

PR #4127, opened 2022-09-09
Remove redundant check from `createInlineAsmCall`

thewilsonator:


gen/irstate.cpp
```diff
@@ -277,7 +277,7 @@ IRState::createInlineAsmCall(const Loc &loc, llvm::InlineAsm *ia,
     
   for (const auto &constraintInfo : ia->ParseConstraints()) {
     if (constraintInfo.isIndirect) {
-      llvm::Type *indirectType = indirectLen != 0 && indirectTypes[indirectIdx] ?
```
JohanEngelen, line 280: it's redundant because `indirectTypes[indirectIdx]` is never null?

thewilsonator: correct, it was null when the constraint was not an indirect constraint, but that approach didn't work so well.

PR #4128, opened 2022-09-09
Pass indirect types in `DtoInlineAsmExpr`

thewilsonator:


gen/naked.cpp
```diff
@@ -427,19 +427,29 @@ DValue *DtoInlineAsmExpr(const Loc &loc, FuncDeclaration *fd,
...
 
+  auto constraintInfo = llvm::InlineAsm::ParseConstraints(constraints);
   // build runtime arguments
   const size_t n = arguments->length - 2;
   LLSmallVector<LLValue *, 8> operands;
+  LLSmallVector<LLType *, 8> indirectTypes;
   operands.reserve(n);
+  if (constraintInfo.size() < n)
+      se->error("number of constraints (%d) must be greater than or equal to the  number of additional arguments (%d)",
+                constraintInfo.size(), n);
```
thewilsonator, line 438: I've no idea if this is at all reasonable, but I've put it here as a sanity check for the time being.

thewilsonator: I wonder, should this be `constraintInfo.size() ==  n + 1`? i.e. one additional constraint for the return type (assuming its not void?)?

PR #4131, opened 2022-09-10
README.txt: Clarify there's no way around MSVC++ toolchain for static linking

kinke:
Wrt. #4047.

### DIFF
--- a/packaging/README.txt
+++ b/packaging/README.txt
@@ -16,11 +16,14 @@ MSVC toolchain detection and setup is skipped if LDC is run inside a
 'VS Native/Cross Tools Command Prompt' (more precisely, if the VSINSTALLDIR
 environment variable is set).
 
-If you don't have a Visual C++ installation, LDC falls back to LLD (the LLVM
-linker) and the bundled WinSDK & Visual C++ runtime (import) libraries based on
-MinGW-w64. In that case, the generated executables and DLLs depend on an
-installed (redistributable) Visual C++ 2015+ runtime (vcruntime140.dll,
-ucrtbase.dll etc.).
+If you don't have a Visual C++ installation, LDC falls back to the integrated
+LLD (the LLVM linker) and the bundled WinSDK & Visual C++ runtime (import)
+libraries based on MinGW-w64. In that case, the generated executables depend on
+an installed (redistributable) Visual C++ 2015+ runtime (vcruntime140.dll,
+ucrtbase.dll etc.). Linking the Visual C++ runtime statically requires the
+static libraries from a Visual C++ toolchain, which cannot be bundled with
+3rd-party compilers such as LDC due to their restrictive license; see
+https://github.com/ldc-developers/mingw-w64-libs for details.
 
 For further information, including on how to report bugs, please refer to the
 LDC wiki: http://wiki.dlang.org/LDC.


thewilsonator (approved):

89z:
> Linking the Visual C++ runtime statically requires the static libraries from a Visual C++ toolchain, which cannot be bundled with 3rd-party compilers such as LDC due to their restrictive license; see https://github.com/ldc-developers/mingw-w64-libs for details.

Couldn't you just create an installer, something similar to this:

<https://github.com/c3lang/c3c/blob/master/msvc_build_libraries.py>

89z:
I think this need more justification than is currently being provided. Several other compiled languages, including Go, Nim, Rust, Zig and others, don't put this burden on developers. What is it about D, that requires developers to install Visual Studio just to get a static build?

kinke:
Don't expect an answer from me, it's all there. And noone has complained about it in years - except for you. Complain to Microsoft instead.

89z:
> Don't expect an answer from me, it's all there.

Its not. I read the updated notes twice, and I don't see the explanation as to why D has to have this requirement, and other languages don't. For example, I download Go [1], which is 156 MB. Then I can download **nothing else**. Then I can run `go build`, and I get a fully static build. Why cant D do that?

> And noone has complained about it in years - except for you.

Hopefully you can realize this is a poor argument. Behind every person who comments, might exist 10 or 100 that do not. Some people might just look at my comments, agree with them, and decide to never use D because of those issues. If you don't care about that, thats fine, but I don't think its healthy to try to paint me as some kind of outlier. If I was, then the four other languages I named would have the same requirements as D. They don't.

> Complain to Microsoft instead.

I will defer to my previous comment. If other languages don't have this requirement, what is it about D makes installing Visual Studio a requirement?


1. https://go.dev/dl

kinke:
If you know *how* the other languages circumvent the MS restrictions in a legal way, without requiring to run some sort of installer, then absolutely feel free to adopt it and open an MR. D is primarily installed by simply unpacking fully standalone, portable archives. These archives and the included libs are also very convenient for cross-building on Posix systems; the MinGW-based libs just work after extracting the Windows package (https://wiki.dlang.org/Cross-compiling_with_LDC).

Running some sort of Visual-Studio-Build-Tools-install-wrapper might be an option for our Windows installer (downloaded 30x less often than the .7z packages for LDC v1.30). I personally don't know any native Windows devs (really working on Windows, not just quickly wanting to compile some Posix-y source on Windows via MinGW or so in a VM) that don't have a VS or at least Build Tools installation anyway, or see that as a huge hassle (done in a few mins).

Years back, I created a little .7z archive containing all MS libs, after merging them from different dirs and converting them all to lowercase. I think the compressed size was < 50 MB, and it worked just fine on Posix systems too. It was pretty trivial but unfortunately it's not allowed to include them, thanks to Microsoft's license. So Rainer Schütze and me put quite some effort into the MinGW-based libs to please the casual Windows user without VS installation and make stuff link out of the box. Never heard complaints again, except for... ;)

Anyway, I'm not going to argue any longer or waste any more time about this topic. Deleting posts on GitHub etc. doesn't really help in bonding.

89z:
> If you know _how_ the other languages circumvent the MS restrictions in a legal way, without requiring to run some sort of installer, then absolutely feel free to adopt it and open an MR.

I think they do it, by only relying on libraries included in a standard Windows installation. For example on my system (with no redistributable installed), I still have these:

~~~
msvcirt.dll
msvcp110_win.dll
msvcp120_clr0400.dll
msvcp140_clr0400.dll
msvcp60.dll
msvcp_win.dll
msvcr100_clr0400.dll
msvcr120_clr0400.dll
msvcrt.dll
~~~

So if you can write software that only uses these, then you remove the need for a user to install Visual Studio, or even a redistributable.

> I personally don't know any native Windows devs (really working on Windows, not just quickly wanting to compile some Posix-y source on Windows via MinGW or so in a VM) that don't have a VS or at least Build Tools installation anyway,

By your own admission, this is anecdotal. Also, again, why should I need to install Visual Studio? **I am a native Windows dev**, for over a decade, both personal and professional. At least four other major languages don't need Visual Studio, so why would I just have it on my computer?

> or see that as a huge hassle (done in a few mins).

It is a huge hassle, and it doesn't take a few minutes. Unless you are a "native Windows dev", I don't think you are in a position to be downplaying the heft of Visual Studio. But if you want to go down that road, I am happy to provide data to shut down that argument.

> So Rainer Schütze and me put quite some effort into the MinGW-based libs to please the casual Windows user without VS installation and make stuff link out of the box.

Thats nice, but it seems you cant (or wont) explain why Visual Studio is still needed? If you say "oh well the standard Windows installation doesn't work for X reason", that is totally fine. But until that justification is provided, its just "oh well you need Visual Studio, because we said so".

> Never heard complaints again, except for... ;)

https://wikipedia.org/wiki/Ad_hominem

> Anyway, I'm not going to argue any longer or waste any more time about this topic.

You seem like a Linux user, who does not have interest in Windows. If thats the case, thats fine. I actually think Linux is better than Windows in many ways. However I see the current state of D for Windows, and I think it could be better. I think its better to remove requirements, unless they are absolutely needed. If they are needed, then its helpful to provide to justification for the users, so they can understand the decision that was made. But it seems like you haven't discovered for yourself why Visual Studio is actually needed. So when someone confronts you with that fact, you get defensive. I would ask you to resist this instinct, and realize that I like D, and just want to see the user experience improve.

kinke:
As said, not going to argue any longer. Peace and out.

PR #4135, opened 2022-09-11
Fix issue 4130

### LINKED ISSUE #4130: My code crashes the LDC compiler, while compiles just fine for DMD
The code in question: https://github.com/ZILtoid1991/newxml

Under LDC2, it crashes with an access violation/segfault. So far, I have tried Windows/x86-64 and Linux/AArch64, all the same issue. With this code, I used to have an issue of segfaulting DMD too, but that was a constructor call issue that I had to solve with the `DOMImplementation.Element.createMap` function, since calling the ctor of `DOMImplementation.Element.Map` from `DOMImplementation.Document` caused issues.

thewilsonator:
This fixes the issue, I'm not sure if the correct thing to do is to make DMD not emit bogus ASTs to begin with.

fixes #4130

### DIFF
--- a/gen/nested.cpp
+++ b/gen/nested.cpp
@@ -307,7 +307,23 @@ LLValue *DtoNestedContext(const Loc &loc, Dsymbol *sym) {
     IF_LOG Logger::println("Current function is %s", ctxfd->toChars());
     if (fromParent) {
       ctxfd = getParentFunc(ctxfd);
-      assert(ctxfd && "Context from outer function, but no outer function?");
+
+      /*
+       The following can happen with
+       class Outer {
+           class Inner {
+               auto opSlice() {
+                   struct Range {
+                       void foo() {}
+                   }
+                   return Range();
+               }
+           }
+       }
+       see https://github.com/ldc-developers/ldc/issues/4130
+       */
+      if (!ctxfd)
+        ctxfd = irFunc.decl;
     }
     IF_LOG Logger::println("Context is from %s", ctxfd->toChars());
 
--- a/tests/codegen/gh4130.d
+++ b/tests/codegen/gh4130.d
@@ -1,0 +1,21 @@
+// RUN: %ldc -run %s
+
+class Outer {
+    Inner inner() {return new Inner; }
+    class Inner {
+        auto bar() {
+            struct Range {
+                void foo() {}
+            }
+            return Range();
+        }
+    }
+}
+
+void main()
+{
+    auto o = new Outer;
+    auto i = o.inner();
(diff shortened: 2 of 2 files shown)


tests/codegen/gh4130.d
```diff
@@ -0,0 +1,21 @@
+// RUN: %ldc -c %s
+
+class A
+{
+    class B { B b; }
+    class C
+    {
+        B b;
```
JohanEngelen, line 8: I think it is nicer to not name every variable `b`. It makes it harder to figure out what is going on.

gen/nested.cpp
```diff
@@ -307,7 +307,8 @@ LLValue *DtoNestedContext(const Loc &loc, Dsymbol *sym) {
     IF_LOG Logger::println("Current function is %s", ctxfd->toChars());
     if (fromParent) {
       ctxfd = getParentFunc(ctxfd);
-      assert(ctxfd && "Context from outer function, but no outer function?");
+      if (!ctxfd)
+        ctxfd = irFunc.decl;
```
JohanEngelen, line 311: add log output about this strange case?

JohanEngelen: does this actually produce correct code now?

thewilsonator: It doesn't crash, which is what it did before. As for the correctness of the code,I'm not really sure, but I don't know what its supposed to do other than iterate through the linked list of classes. @ZILtoid1991 could you provide a reduced test case of this code in action?

ZILtoid1991:
```d
class DOMImplementation {
class Attr  {  Attr  _nextAttr; }
class Map  {
    Attr firstAttr;
    auto opSlice()
    {
         struct Range
         {
             Attr currentAttr;

             auto front() { return currentAttr; }
             void popFront() { currentAttr = currentAttr._nextAttr; }
             bool empty() { return currentAttr is null; }
        }
       return Range(firstAttr);
   }
}
}
```

kinke: Thx, somewhat further reduced:
```D
class Outer {
    class Inner {
        auto opSlice() {
            struct Range {
                void foo() {}
            }
            return Range();
        }
    }
}
```

**2** outer aggregates are apparently required, and both need to be classes (works with any one of them being a struct).

JohanEngelen: > It doesn't crash, which is what it did before. As for the correctness of the code,I'm not really sure, but I don't know what its supposed to do other than iterate through the linked list of classes. 

If unsure about the correct behavior, then better to exit gracefully instead of silently continuing.

PR #4136, opened 2022-09-11
Fix issue 4134

### LINKED ISSUE #4134: DustMite shipped with `ldc-latest-ci` corrupts files
Not sure what exactly is the problem but I think LDC might be miscompiling DustMite.

Reproducer:
```
$ ~/dlang/install.sh install ldc-latest-ci
...

$ cat test.d
extern(C) int printf(const scope char*, ...);
 
void main()
{
    printf("This text should be shown!\n");
}

$ ~/dlang/ldc-69a54c94/bin/dustmite --no-redirect test.d 'cat test.d ; exit 1'
Loading test.d
None => ;

{
;
}
No
Hint: read https://github.com/CyberShadow/DustMite/wiki#initial-test-fails
object.Exception@DustMite/dustmite.d(367): Initial test fails: Test script "cat test.d ; exit 1" exited with exit code 1 (failure)
----------------
??:? [0x55a4b1a01406]
??:? [0x55a4b1a02a06]
??:? [0x55a4b1a01cdf]
??:? [0x55a4b18fe7bd]
??:? [0x55a4b1a011bb]
??:? [0x55a4b1a010b5]
??:? [0x55a4b1a00f0d]
??:? [0x55a4b1927591]
??:? [0x7f12e793f2cf]
??:? __libc_start_main [0x7f12e793f389]
??:? [0x55a4b18ea369]
```

So, DustMite reads in:

```d
extern(C) int printf(const scope char*, ...);
 
void main()
{
    printf("This text should be shown!\n");
}
```

but when it should write out the same file with no changes for the initial test, it instead writes:

```d
;

{
;
}
```

Seems to be a recent regression.
Downstream issue: https://github.com/dlang/dub/pull/2446#issuecomment-1242714549

thewilsonator:
partial revert of https://github.com/ldc-developers/ldc/commit/37aca9c9405226d10bb0aeb082ffeb508da78b9e fixes #4134, tested locally (e: not with an LLVM with assertions enabled though apparently)

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -919,10 +919,12 @@ DSliceValue *DtoCatArrays(const Loc &loc, Type *arrayType, Expression *exp1,
 
     auto loadArray = [fn](Expression* e, int paramTypeIdx) {
       DValue * dval = toElem(e);
-      LLValue *val = DtoLoad(DtoSlicePtrType(dval), DtoSlicePtr(e));
+      LLValue *val = DtoLoad(DtoSlicePtrType(dval), DtoSlicePtr(dval));
       return DtoAggrPaint(val, fn->getFunctionType()->getParamType(paramTypeIdx));
     };
+    // byte[] x
     args.push_back(loadArray(exp1,1));
+    // byte[] y
     args.push_back(loadArray(exp2,2));
   }
 
--- a/tests/codegen/gh4134.d
+++ b/tests/codegen/gh4134.d
@@ -1,0 +1,14 @@
+// https://github.com/ldc-developers/ldc/issues/4134
+// RUN: %ldc -run %s
+
+int i;
+
+string getString() {
+    ++i;
+    return "Abc";
+}
+
+void main() {
+    const r = getString() ~ getString();
+    assert(i == 2);
+}


JohanEngelen:
Can you add a testcase for the bug that was introduced, to prevent future regression?

gen/arrays.cpp
```diff
@@ -919,13 +919,16 @@ DSliceValue *DtoCatArrays(const Loc &loc, Type *arrayType, Expression *exp1,
     // TypeInfo ti
     args.push_back(DtoTypeInfoOf(loc, arrayType));
 
-    auto loadArray = [fn](Expression* e, int paramTypeIdx) {
-      DValue * dval = toElem(e);
-      LLValue *val = DtoLoad(DtoSlicePtrType(dval), DtoSlicePtr(e));
```
kinke, line 924: Ah, so most likely a double `toElem(e)`, as you didn't pass `dval` to `DSlicePtr()`, but `e` again.

thewilsonator: Yep that's the issue

thewilsonator:
> Can you add a testcase for the bug that was introduced, to prevent future regression?

I think that would require dustmiting dutmite using ldc master/this pr, but I'm not very experienced with using dustmite let alone trying to dustmite dustmite. 

@CyberShadow could you help here? I presume the issue is with `splitter.d`, but that's just a guess.

CyberShadow:
How can I get a Linux LDC build without this bug? (Preferably which doesn't involve compiling LLVM)

thewilsonator:
You can get a precompiled LLVM from https://github.com/ldc-developers/llvm-project/tags but I think you should be able to use the latest (non-CI) LDC release [here](https://github.com/ldc-developers/ldc/releases/tag/v1.30.0), it won't have this bug. 

I _think_ that all you need is an LDC with the bug (ldc-latest-ci) and one without in order to construct the appropriate shell script to feed to dustmite.

kinke:
Way simpler to create an artificial test based on the bug/fix itself:
```D
int i;

string getString() {
    ++i;
    return "Abc";
}

void main() {
    const r = getString() ~ getString();
    assert(i == 2);
}
```

But not too much use IMO (only covers this one specific case); the compiler evaluating an expression multiple times isn't really something one would expect and test.

JohanEngelen:
> But not too much use IMO (only covers this one specific case); the compiler evaluating an expression multiple times isn't really something one would expect and test.

I disagree. D is very much undertested, hence all the bugs that are introduced regularly. Please add it.

thewilsonator:
Test added.

tests/codegen/gh4134.d
```diff
@@ -0,0 +1,14 @@
+// https://github.com/ldc-developers/ldc/issues/4134
+// RUN: ldc -run %s
```
Geod24, line 2:
```suggestion
// RUN: ldc2 -run %s 
```

thewilsonator: should be `%ldc`

PR #4141, opened 2022-09-12
Upgrade frontend & libs to pre-DMD-monorepo state

kinke:


thewilsonator:
Apologies for the merge conflict, I've got a couple of changes I need to make to druntime, namely making the intrinsic names for pointers not include the type for LLVM 15 (well technically opaque pointers, which in theory we could enable for LLVM14 or disable for 15) e.g. 
```diff
-pragma(LDC_intrinsic,  "llvm.memcpy.p0i8.p0i8.i#")
+pragma(LDC_intrinsic, LLVM_version >= 1500 ? "llvm.memcpy.p0.p0.i#" : "llvm.memcpy.p0i8.p0i8.i#")
    void llvm_memcpy(T)(void* dst, const(void)* src, T len, bool volatile_ = false)
        if (__traits(isIntegral, T));
```
ditto for `memmove`, `memset` `prefetch` and `frameaddress`.

kinke:
The undefined `_d_arrayctor` symbol for the druntime unittest runners is caused by https://issues.dlang.org/show_bug.cgi?id=23337 and needs a frontend fix...

tests/semantic/dcompute.d
```diff
@@ -31,7 +31,7 @@ void func()
     int[] quux;
     //CHECK: dcompute.d([[@LINE+1]]): Error: can only call functions from other `@compute` modules in `@compute` code
     quux.length = 1;
-    //CHECK: dcompute.d([[@LINE+1]]): Error: cannot use operator `~=` in `@compute` code
+    //CHECK: dcompute.d([[@LINE+1]]): Error: can only call functions from other `@compute` modules in `@compute` code
```
JohanEngelen, line 34: this looks like a regression to me. @thewilsonator can you help here to keep the better old error message?

kinke: These are caused by new frontend lowerings (the templated 'hooks').

JohanEngelen: Perhaps we can hardcode the recognition of names of the hooks, and give a bit nicer feedback to the user. It will not be easy to understand that `~=` leads to a "call to a function from another non-`@compute` module".

JohanEngelen: I'm fine with making an open issue for it, shouldn't block the monorepo work.

kinke: Okay: #4272

PR #4144, opened 2022-09-13
Remove use of `ptrty->getContainedType(0)` from `DtoSetArray`

thewilsonator:


gen/arrays.cpp
```diff
@@ -323,13 +323,13 @@ void DtoArrayAssign(const Loc &loc, DValue *lhs, DValue *rhs, EXP op,
 
 ////////////////////////////////////////////////////////////////////////////////
 
-static void DtoSetArray(DValue *array, LLValue *dim, LLValue *ptr) {
+static void DtoSetArray(DValue *array, DValue *rhs) {
   IF_LOG Logger::println("SetArray");
   LLValue *arr = DtoLVal(array);
-  LLType *s = isaStruct(arr->getType()->getContainedType(0));
+  LLType *s = DtoType(array->type->nextOf());
```
kinke, line 329: This is supposed to be the slice LL struct, not the slice element type.

PR #4155, opened 2022-09-13
Remove `getContainedType` in `toir.cpp`

thewilsonator:


gen/toir.cpp
```diff
@@ -843,7 +843,7 @@ class ToElemVisitor : public Visitor {
     if (e->offset == 0) {
       offsetValue = baseValue;
     } else {
-      LLType *elemType = baseValue->getType()->getContainedType(0);
+      LLType *elemType = DtoType(base->type->nextOf());
```
thewilsonator, line 846: this is wrong, will come back to this

PR #4159, opened 2022-09-13
Remove `getContainedType` in `tollvm.cpp`

thewilsonator:


gen/tollvm.cpp
```diff
@@ -543,7 +543,6 @@ void DtoVolatileStore(LLValue *src, LLValue *dst) {
 void DtoStoreZextI8(LLValue *src, LLValue *dst) {
   if (src->getType()->isIntegerTy(1)) {
     llvm::Type *i8 = llvm::Type::getInt8Ty(gIR->context());
-    assert(dst->getType()->getContainedType(0) == i8);
```
thewilsonator, line 546: this assert is not expressible with opaque pointers

PR #4165, opened 2022-09-14
Remove `getContainedType` from `DtoArrayAssign`

thewilsonator:


gen/arrays.cpp
```diff
@@ -293,7 +293,7 @@ void DtoArrayAssign(const Loc &loc, DValue *lhs, DValue *rhs, EXP op,
     if (!needsDestruction && !needsPostblit) {
       // fast version
       const size_t lhsElementSize =
-          getTypeAllocSize(realLhsPtr->getType()->getContainedType(0));
```
thewilsonator, line 296: Wait, why is the _left_ hand side, element size determined by the _right_ hand size???

PR #4168, opened 2022-09-14
`InlineAsm::verify` returns an `Error` in LLVM15

thewilsonator:


gen/naked.cpp
```diff
@@ -503,17 +503,20 @@ llvm::CallInst *DtoInlineAsmExpr(const Loc &loc, llvm::StringRef code,
...
+      error(loc, "inline asm constraints are invalid");
+      fatal();
+  }
 #else
-      verify
-#endif
-      (FT, constraints)) {
+  if (auto err = llvm::InlineAsm::verify(FT, constraints)) {
     error(loc, "inline asm constraints are invalid");
+    llvm::errs() << err;
```
thewilsonator, line 515: not quite sure how to convert this into a supplemental error, but that's a pretty low priority given this is followed immediately by `fatal();`

PR #4169, opened 2022-09-14
Uwtable

jamesragray:
Added a parameter to uwtable attribute required by LLVM15.

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -1216,6 +1216,7 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   // function attributes
   if (gABI->needsUnwindTables()) {
     func->addFnAttr(LLAttribute::UWTable);
+    func->setUWTableKind(llvm::UWTableKind::Default);
   }
   if (opts::isAnySanitizerEnabled() &&
       !opts::functionIsInSanitizerBlacklist(fd)) {
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -295,6 +295,7 @@ void addCoverageAnalysis(Module *m) {
     // Set function attributes. See functions.cpp:DtoDefineFunction()
     if (global.params.targetTriple->getArch() == llvm::Triple::x86_64) {
       ctor->addFnAttr(LLAttribute::UWTable);
+      ctor->setUWTableKind(llvm::UWTableKind::Default);
     }
 
     llvm::BasicBlock *bb = llvm::BasicBlock::Create(gIR->context(), "", ctor);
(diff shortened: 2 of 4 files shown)


gen/functions.cpp
```diff
@@ -1216,6 +1216,7 @@ void DtoDefineFunction(FuncDeclaration *fd, bool linkageAvailableExternally) {
   // function attributes
   if (gABI->needsUnwindTables()) {
     func->addFnAttr(LLAttribute::UWTable);
+    func->setUWTableKind(llvm::UWTableKind::Default);
```
thewilsonator, line 1219: this will have only been added in recent LLVM. You will need to guard it with `#if LDC_LLVM_VER >= 1400` (1400 is just a guess)

JohanEngelen:
I worry that `default` means generating `async` tables, which apparently have a large overhead: https://groups.google.com/g/llvm-dev/c/KH4QAKPBQ5A
Better to explicitly set to `sync` because we do not use `async` ?

LLVM Langref:
> uwtable[(sync|async)]
This attribute indicates that the ABI being targeted requires that an unwind table entry be produced for this function even if we can show that no exceptions passes by it. This is normally the case for the ELF x86-64 abi, but it can be disabled for some compilation units. The optional parameter describes what kind of unwind tables to generate: sync for normal unwind tables, async for asynchronous (instruction precise) unwind tables. Without the parameter, the attribute uwtable is equivalent to uwtable(async).

jamesragray:
Okay will change the default. I will put a command line option for now as well.

JohanEngelen:
> Okay will change the default. I will put a command line option for now as well.

I've read the history of this a bit more. `default` is what we currently do, so let's keep it, sorry for the confusion. But it would indeed be good to add a commandline flag for it. (please duplicate the clang commandline option naming)
`sync` would also work, but is a change of LDC behavior, and might hurt advanced users with sampling profilers, debuggers. It'd be good to revisit this in the future, but for now I propose we stick to our current behavior.

jamesragray:
If I am not mistaken (I only found the information by looking at https://reviews.llvm.org/D114543) clang uses:
-uwtable with value 0, 1, or 2, which I guess correspond to what 
I found in (https://llvm.org/doxygen/CodeGen_8h_source.html)
```
 enum class UWTableKind {
     None = 0,  ///< No unwind table requested
     Sync = 1,  ///< "Synchronous" unwind tables
     Async = 2, ///< "Asynchronous" unwind tables (instr precise)
     Default = 2,
   };
```
which also seems to suggest that the Default is Async?

JohanEngelen:
> which also seems to suggest that the Default is Async?

yes.  `-funwind-tables    -fno-asynchronous-unwind-tables` is needed for clang to output `unwind(sync)` attribute. https://cpp.godbolt.org/z/hnofz7ddY

So we should use `default` and give user the option to select sync.

thewilsonator:
We'll figure out what command line options to control in a follow up PR. 
In the mean time, this fixes compilation with LLVM15 on `x86_64`.

PR #4175, opened 2022-09-15
New passmanager clean

jamesragray:
This is related to https://github.com/ldc-developers/ldc/issues/3961
Starting to produce optimization pipeline using the new pass manager for LLVM15.

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -121,7 +121,7 @@ static cl::opt<int> fSanitizeMemoryTrackOrigins(
         "Enable origins tracking in MemorySanitizer (0=disabled, default)"));
 
 static cl::opt<signed char> passmanager("passmanager",
-    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(1),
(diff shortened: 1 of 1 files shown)


gen/optimizer.cpp
```diff
@@ -413,6 +430,391 @@ bool ldc_optimize_module(llvm::Module *M) {
...
+      mpm.addPass(PGOIndirectCallPromotion());
+    }
+  }
+}
+
+static void addStripExternalsPass(ModulePassManager &mpm,
+                                      OptimizationLevel level ) {
+
+  if (level == OptimizationLevel::O1 || level == OptimizationLevel::O2 ||
+      level == OptimizationLevel::O3) {
```
thewilsonator, line 522: N.B. this misses `Os` and `Oz`, probably best to have this as `level != OptimizationLevel::O0`, but we can change that when the passes are added properly

thewilsonator (approved):
Nice! Will merge this and then we can work on the fixemes.

JohanEngelen:
FYI, this merge means that the new passmanager is now available on Godbolt's compiler explorer. If you want to check what differences there are in passes run, you can use compiler explorer's Opt pipeline viewer: https://d.godbolt.org/z/eP5f3be51

PR #4176, opened 2022-09-16
Fix compilation with LLVM16

thewilsonator:


driver/toobj.cpp
```diff
@@ -35,8 +35,10 @@
 #include "llvm/Transforms/Utils/Cloning.h"
 #include "llvm/IR/Module.h"
 #ifdef LDC_LLVM_SUPPORTED_TARGET_SPIRV
+#if LDC_LLVM_VER < 1600
 #include "LLVMSPIRVLib/LLVMSPIRVLib.h"
 #endif
+#endif
```
thewilsonator, line 41: I plan to switch to the LLVM backend for LLVM16

PR #4179, opened 2022-09-17
remove `warnInvalidPrintfCall`, it is handled by dmd

thewilsonator:
```d
pragma(printf)
extern(C) int printf(const char*,...);

void main()
{
    size_t i;
    printf("%d"); // Deprecation: more format specifiers than 0 arguments
    printf("%d", i); // Deprecation: argument `i` for format specification `"%d"` must be `int`, not `ulong`
}
```


thewilsonator:
There is still 
```c++
llvm::cl::opt<bool> checkPrintf(
    "check-printf-calls", llvm::cl::ZeroOrMore,
    llvm::cl::desc("Validate printf call format strings against arguments"));
```
at the top of `toir.cpp`, how do we deprecate command line args? or should we just remove it?

JohanEngelen:
By the looks of it, this was never functional beyond a `wchar` check...?

thewilsonator:
Indeed, I don't imagine anyone would have enabled that switch.

kinke:
> Indeed, I don't imagine anyone would have enabled that switch.

Yeah well: https://github.com/PhilippeSigaud/Pegged/blob/master/dub.json#L21

> how do we deprecate command line args

Making them hidden/really hidden, and issuing a deprecation if used, like https://github.com/ldc-developers/ldc/blob/23fa11d320c04542ad937362ae68d274cb73efd6/driver/linker.cpp#L169-L172

PR #4181, opened 2022-09-17
Add support for dcompute/directX

thewilsonator:
Still very WIP, the backend is also in development. Just putting this up here.


thewilsonator:
cc @TurkeyMan

tests/codegen/dcompute_directx.d
```diff
@@ -0,0 +1,6 @@
+// REQUIRES: target_DirectX
+// RUN: %ldc -c -mdcompute-targets=dx-630 -m64 -mdcompute-file-prefix=directx -output-ll %s
```
JohanEngelen, line 2: perhaps add a basic sanity check for the IR output?

thewilsonator: Whence why this is still WIP, I'm trying to figure out what that is supposed to be (and have more than a no-op kernel as an example.

PR #4182, opened 2022-09-17
Fix for optnone not being respected with new passmanager

jamesragray:
Found an explanation of what was happening with optnone attribute being ignored here: https://www.duskborn.com/posts/llvm-new-pass-manager/#the-curious-case-of-the-optimized-optnone-functions
I made the change suggested there.

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -122,7 +122,7 @@ static cl::opt<int> fSanitizeMemoryTrackOrigins(
         "Enable origins tracking in MemorySanitizer (0=disabled, default)"));
 
 static cl::opt<signed char> passmanager("passmanager",
-    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(1),
+    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(0),
     cl::values(
         clEnumValN(0, "legacy", "Use the legacy passmanager (available for LLVM14 and below) "),
         clEnumValN(1, "new", "Use the new passmanager (available for LLVM14 and above)")));
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -641,22 +641,25 @@ void runOptimizationPasses(llvm::Module *M) {
 //  } else {
 //    builder.Inliner = createAlwaysInlinerLegacyPass();
 //  }
-
-   llvm::PassInstrumentationCallbacks passInstrumentationCallbacks;
-
-   llvm::OptNoneInstrumentation optNoneInstrumentation(false);
-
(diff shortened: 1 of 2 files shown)


thewilsonator:
Looks good, some tests will need to be updated to account for the changes in whitespace, specifically the regex 
`// CHECK-DAG: attributes #[[OPTNONE]] = {{.*}} noinline {{.*}} optnone` in `codegen/attr_optstrat.d:64` should become 
`// CHECK-DAG: attributes #[[OPTNONE]] = {{.*}} noinline {{.*}}optnone`. Ditto for the other failing test I suspect.

gen/optimizer.cpp
```diff
@@ -641,8 +642,16 @@ void runOptimizationPasses(llvm::Module *M) {
...
+
+   llvm::OptNoneInstrumentation optNoneInstrumentation(false);
+
+   optNoneInstrumentation.registerCallbacks(passInstrumentationCallbacks);
+
+
+
   PassBuilder pb(gTargetMachine, getPipelineTuningOptions(optLevelVal, sizeLevelVal),
-                 getPGOOptions());
+                 getPGOOptions(), &passInstrumentationCallbacks);
```
JohanEngelen, line 654: can you verify that this does not override some default instrumentationcallback?

jamesragray: Okay I will check that.

jamesragray: According to https://llvm.org/doxygen/classllvm_1_1PassBuilder.html#aef3b2db337187e91efc9923f2f688abd the default is a null pointer.

jamesragray: Here is what clang does:
```
 FunctionAnalysisManager FAM;
  CGSCCAnalysisManager CGAM;
  ModuleAnalysisManager MAM;

  bool DebugPassStructure = CodeGenOpts.DebugPass == "Structure";
  PassInstrumentationCallbacks PIC;
  PrintPassOptions PrintPassOpts;
  PrintPassOpts.Indent = DebugPassStructure;
  PrintPassOpts.SkipAnalyses = DebugPassStructure;
  StandardInstrumentations SI(CodeGenOpts.DebugPassManager ||
                                  DebugPassStructure,
                              /*VerifyEach*/ false, PrintPassOpts);
  SI.registerCallbacks(PIC, &FAM);
  PassBuilder PB(TM.get(), PTO, PGOOpt, &PIC);
```

jamesragray: Updating to follow clang.

thewilsonator:
Looks like this only fails gc2stack and custom plugin pass, both of which are expected. 
Set the default pass manager back to legacy and I think this should be good to go.

thewilsonator (approved):

PR #4187, opened 2022-09-18
Use LLVM's SPIRV backend for LLVM 16

thewilsonator:
The test has this at the top of it.
```
// REQUIRES: target_SPIRV
// RUN: %ldc -c -mdcompute-targets=ocl-220 -m64 -mdcompute-file-prefix=addrspace -output-ll -output-o %s && FileCheck %s --check-prefix=LL < addrspace_ocl220_64.ll \
// RUN: && %llvm-spirv -to-text addrspace_ocl220_64.spv && FileCheck %s --check-prefix=SPT < addrspace_ocl220_64.spt
```

judging by e.g. https://github.com/llvm/llvm-project/blob/main/llvm/test/CodeGen/SPIRV/ExecutionMode.ll it seems we should be able to use `llc` to then get text that we can verify. How do I `//RUN:` a different set of commands based on the LLVM version that LDC was built against?

### DIFF
--- a/driver/toobj.cpp
+++ b/driver/toobj.cpp
@@ -72,12 +72,12 @@ void codegenModule(llvm::TargetMachine &Target, llvm::Module &m,
     std::ofstream out(filename, std::ofstream::binary);
     llvm::createSPIRVWriterPass(out)->runOnModule(m);
     IF_LOG Logger::println("Success.");
+    return;
 #endif
 #else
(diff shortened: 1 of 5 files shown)


thewilsonator:
bummer I made this PR from `master`.

JohanEngelen:
>  How do I `//RUN:` a different set of commands based on the LLVM version that LDC was built against?

You can't. You have to make two different files to do that. (and then REQUIRE llvm_atleast ...)

PR #4188, opened 2022-09-18
Gc2stack

jamesragray:
Work in progress new pass manager GarbageCollect2StackPass.

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -122,7 +122,7 @@ static cl::opt<int> fSanitizeMemoryTrackOrigins(
         "Enable origins tracking in MemorySanitizer (0=disabled, default)"));
 
 static cl::opt<signed char> passmanager("passmanager",
-    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(1),
+    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(0),
     cl::values(
         clEnumValN(0, "legacy", "Use the legacy passmanager (available for LLVM14 and below) "),
         clEnumValN(1, "new", "Use the new passmanager (available for LLVM14 and above)")));
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -11,6 +11,7 @@
 
 #include "dmd/errors.h"
 #include "gen/logger.h"
+#include "gen/passes/GarbageCollect2Stack.h"
 #include "gen/passes/Passes.h"
 #include "driver/cl_options.h"
 #include "driver/cl_options_instrumentation.h"
@@ -121,7 +122,7 @@ static cl::opt<int> fSanitizeMemoryTrackOrigins(
         "Enable origins tracking in MemorySanitizer (0=disabled, default)"));
 
 static cl::opt<signed char> passmanager("passmanager",
-    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(0),
+    cl::desc("Setting the passmanager (new,legacy):"), cl::ZeroOrMore, cl::init(1),
     cl::values(
         clEnumValN(0, "legacy", "Use the legacy passmanager (available for LLVM14 and below) "),
         clEnumValN(1, "new", "Use the new passmanager (available for LLVM14 and above)")));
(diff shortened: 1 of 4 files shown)


gen/optimizer.cpp
```diff
@@ -567,10 +568,11 @@ static void addGarbageCollect2StackPass(ModulePassManager &mpm,
   if (level == OptimizationLevel::O2  || level == OptimizationLevel::O3) {
 //FIXME: Update and uncomment this once gen/passes/GarbageCollect2Stack.cpp is updated to
 // work with the new pass manager.
-//    mpm.addPass(GarbageCollect2Stack());
-//    if (verifyEach) {
-//      mpm.addPass(VerifierPass());
-//    }
+    FunctionPassManager fpm;
+    fpm.addPass(GarbageCollect2StackPass());
```
JohanEngelen, line 572: looks like you forgot to add this functionpassmanager to the modulepassmanager.

jamesragray: Thanks

gen/passes/GarbageCollect2Stack.h
```diff
@@ -0,0 +1,156 @@
+#pragma once
+#include "gen/llvm.h"
+#include "gen/passes/Passes.h"
+#include "llvm/Analysis/CallGraph.h"
+#include "llvm/IR/DataLayout.h"
+#include "llvm/IR/Dominators.h"
+
+namespace {
+struct Analysis {
```
thewilsonator, line 9: You could forward declare this here and leave the implementation in the `.cpp` file. the only use of `Analysis` is by `const &`, I think you would have to remove the `
 namespace { ... }` though. (and probably call it something less generic than analysis, e.g. `GC2StackAnalysis`)

thewilsonator:
The remaining failure is 
```
#4  D main (args=...) at src/rt_trap_exceptions_drt.d:9
        myLocal = <error reading variable>
grep "in D main (args=...) at .*rt_trap_exceptions_drt.d:9" > /dev/null < /home/runner/work/ldc/build/runtime/druntime-test-exceptions-release/rt_trap_exceptions_drt_gdb.output
```

I'm not sure why the address is not part of the output here, maybe because it is at the start of the function?
In `runtime/druntime/test/exceptions/rt_trap_exceptions_drt.d`. The test checks for `myLocal`, but doesn't check for not ` <error reading variable>`. I'm not sure if that is an omission on the part of the test.
```
cat $(ROOT)/rt_trap_exceptions_drt_gdb.output
grep "in D main (args=...) at .*rt_trap_exceptions_drt.d:9" > /dev/null < $(ROOT)/rt_trap_exceptions_drt_gdb.output
grep 'myLocal' > /dev/null < $(ROOT)/rt_trap_exceptions_drt_gdb.output
! grep "No stack." > /dev/null < $(ROOT)/rt_trap_exceptions_drt_gdb.output
```

I note that the invalid memory operation test does 
```
ifeq (ldmd,$(findstring ldmd,$(DMD)))
# LDC: Make sure allocation intended to provoke exception is not elided.
$(ROOT)/invalid_memory_operation: DFLAGS+=-disable-gc2stack
endif
```

thewilsonator (approved):

tests/plugins/addFuncEntryCall/testPlugin.d
```diff
@@ -1,7 +1,7 @@
 // REQUIRES: Plugins
 
 // RUN: %gnu_make -f %S/Makefile
-// RUN: %ldc -c -output-ll -plugin=./addFuncEntryCallPass.so -of=%t.ll %s
+// RUN: %ldc -passmanager=legacy -c -output-ll -plugin=./addFuncEntryCallPass.so -of=%t.ll %s
```
JohanEngelen, line 4: how is this related to GC2Stack ?

gen/optimizer.cpp
```diff
@@ -717,9 +714,13 @@ void runOptimizationPasses(llvm::Module *M) {
...
       });
     }
-
     if (!disableGCToStack) {
-      pb.registerLoopOptimizerEndEPCallback([&](LoopPassManager &lpm,
+//FIX ME: This should be checked
+      fam.registerPass([&] { return DominatorTreeAnalysis(); });
+      mam.registerPass([&] { return  CallGraphAnalysis(); });
+//FIX ME: Is this registerOptimizerLastEPCallback correct here
+//(had registerLoopOptimizerEndEPCallback) but that seems wrong
```
JohanEngelen, line 722: please create an open issue to revisit pass ordering after ldc is fully functional with the new passmanager is done (and creation, preferably not create a new function passmanager for this pass as that probably means the ordering/parallelization is suboptimal).

jamesragray: https://github.com/ldc-developers/ldc/issues/4222#issue-1383901032

PR #4210, opened 2022-09-23
Opaque Pointers: `codegen/gh3692.d`

thewilsonator:


thewilsonator:
Hmm, might need to split this into a pre LLVM15 and post LLVM, there is much less bit casting required

JohanEngelen:
Fixed as part of https://github.com/ldc-developers/ldc/pull/4252

PR #4225, opened 2022-09-25
Pass slices as not a struct

thewilsonator:
Exploring the idea of splitting `T[]` arguments into `i(32|64), (T*|ptr)` on the call level to enable adding `@restrict` to `T[]` arguments.

Currently (one among many) thing that fails is doesn't produce the correct `alloca`, it produces `%a = alloca i64, align 8` and then only stores the length in `a_arg_length`. I can't seem to figure out where the code that produces the bad alloca comes from.

Still todo, runtime functions (`_d_.*`), test suite update, ...
Very much WIP.


JohanEngelen:
I don't know if it is worth changing ABI for this. Another approach would be to add preamble code that reads from the argument with `norestrict` on the read operation?  (into a dummy local variable that can then be used for the rest of the function codegen).

JohanEngelen:
see also: https://llvm.org/devmtg/2017-02-04/Restrict-Qualified-Pointers-in-LLVM.pdf

JohanEngelen:
Did anyone ever ask about this issue on LLVM's maillist?

thewilsonator:
we can limit the ABI changes to just functions that take `@restrict`, but that would break function pointers...

Issue #2511, opened 2018-01-15
Add target-independent controlflow protection `-fcf-protection`


JohanEngelen:
See:
- https://reviews.llvm.org/D40478
- https://reviews.llvm.org/D41880

There is more CFI stuff in Clang that may be interesting for someone to implement for LDC.

PR #4221, opened 2022-09-23
Add LLVM 15 (Vanilla) CI Job

thewilsonator:
This is expected to fail for the time being, but we're almost there!

fixes #4042

PR #4239, opened 2022-10-24
Add support for control-flow protection

kubo39:
This implementation is fully inspired by `-fcf-protection` in Clang, with options `none|branch|return|full`.

- related: #2511
- LLVM side: https://reviews.llvm.org/D40478

### DIFF
--- a/driver/codegenerator.cpp
+++ b/driver/codegenerator.cpp
@@ -237,14 +237,20 @@ void CodeGenerator::prepareLLModule(Module *m) {
 
   // Currently, controll-flow architecture protection is x86 only.
   auto arch = global.params.targetTriple->getArch();
-  if (arch == llvm::Triple::x86 || arch == llvm::Triple::x86_64) {
-    if (opts::cfProtection == opts::CFProtection::branch || opts::cfProtection == opts::CFProtection::full) {
+  if (opts::cfProtection == opts::CFProtection::branch || opts::cfProtection == opts::CFProtection::full) {
+    if (arch == llvm::Triple::x86 || arch == llvm::Triple::x86_64) {
       // Indicate that we want to instrument branch control flow protection.
       ir_->module.addModuleFlag(llvm::Module::Override, "cf-protection-branch", 1);
+    } else {
+      warning(Loc(), "Ignoring -fcf-protection options: branch control flow protection is not supported this architecture");
     }
-    if (opts::cfProtection == opts::CFProtection::return_ || opts::cfProtection == opts::CFProtection::full) {
+  }
+  if (opts::cfProtection == opts::CFProtection::return_ || opts::cfProtection == opts::CFProtection::full) {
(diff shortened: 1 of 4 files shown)


driver/codegenerator.cpp
```diff
@@ -235,6 +235,19 @@ void CodeGenerator::prepareLLModule(Module *m) {
   ir_->module.setTargetTriple(global.params.targetTriple->str());
   ir_->module.setDataLayout(*gDataLayout);
 
+  // Currently, controll-flow architecture protection is x86 only.
```
JohanEngelen, line 238: I think it's good to emit a warning in case the user supplied these cmdline flags for non-supported platforms (telling the user that the cmdline args are ignored)

tests/codegen/cfprotection.d
```diff
@@ -0,0 +1,30 @@
...
+// UNDEFINED-NOT: !"cf-protection-branch"
+// NONE-NOT: !"cf-protection-branch"
+// BRANCH: !"cf-protection-branch", i32 1
+// RETURN-NOT: !"cf-protection-branch"
+// FULL: !"cf-protection-branch", i32 1
+
+// UNDEFINED-NOT: !"cf-protection-return"
+// NONE-NOT: !"cf-protection-return"
+// BRANCH-NOT: !"cf-protection-return"
+// RETURN: !"cf-protection-return", i32 1
```
JohanEngelen, line 29: the ordering of -NOT and normal checks is always annoying... You have to put a -NOT before and after something that does need to be there:
```
CHECK-NOT: something
CHECK: needs to be there
CHECK-NOT: something
```

JohanEngelen:
Cool that you implemented this. Did you try it? I'm wondering how it works / what it does. Do you need runtime support? (i.e. link with an LLVM lib?)

JohanEngelen:
Reference to current Clang implementation: https://github.com/llvm/llvm-project/blob/5f5e0199c1a650c41e1ea6e0c18d3c0b29f45023/clang/lib/CodeGen/CodeGenModule.cpp#L753-L764

Can you look into the difference between `llvm::Module::Min` and `Override` ?

thewilsonator:
If this is only enabled for LLVM 15, then do note that we do not currently have any CI set up to test this. see #4221

JohanEngelen:
> > Did you try it? I'm wondering how it works / what it does. Do you need runtime support? (i.e. link with an LLVM lib?)
> 
> Wow, good catch! I tested under my old Linux/WSL with Intel SDE and this did not work as expected. (evil.d is copied and modified from https://smallkirby.hatenablog.com/entry/2020/09/10/230629)

Can you add this as a runnable test case?

(we also have such very basic integration tests for e.g. address sanitizer)

Note that there is still value for this feature even if it does not (yet) work fully by itself. If runtime support is needed, that can be added separately from the LLVM IR modifications that you do in this PR.

kubo39:
duplicated with https://github.com/ldc-developers/ldc/pull/4437

PR #4249, opened 2022-11-09
Add druntime .di source files to build dependencies.

JohanEngelen:
This rebuilds druntime when a .di file is changed (e.g. `ldc.intrinsics` which is used in some places in druntime source).


Imperatorn:
Merge?

PR #4254, opened 2022-11-11
Fix loading pass plugins while using the new passmanager. (LLVM >= 14)

JohanEngelen:


thewilsonator (approved):
Will probably need to `#if` out the legacy implementation for LLVM15+ (16+?) where the legacy option os no longer available.

PR #4278, opened 2022-11-25
Travis & Cirrus CI: Do not upload artifacts to GitHub release for tags without `v` prefix

kinke:
GitHub Actions already skips these other tags: https://github.com/ldc-developers/ldc/blob/5c3478c129492acf2201222d5398090a07072f66/.github/workflows/main.yml#L268


kinke:
[This is needed for the still missing `pre-monorepo` tag.]

PR #4279, opened 2022-11-25
Upgrade frontend & libs to v2.101.2

kinke:


kinke:
There's a weird regression for `runnable/mixin2.d` when optimized - a `scope` class allocation is totally optimized away, while its destructor (invoked indirectly via `_d_callfinalizer()`) performs a required global state modification. The unoptimized IR is hardly different from master's though - just a tiny dtor mangle diff (all class dtors are now implicitly `scope`), and few differently ordered load/GEP instructions (which shouldn't have any effect though).

The `Main` GH Actions workflow shows that the compiler quickly segfaults when compiling itself. I've reproduced it locally; it's apparently caused by `NewExp.newtype` somehow having a `.ty` of `Tident` (6), not `Tclass` (7) - unexpected when computing LDC-specific `VarDeclaration.onstackWithDtor`.

kinke:
Finally almost ready, after revising our elide-scope-class-finalization optimization and disabling a few betterC/importC tests.

Just one remaining issue on 32-bit Windows, where a 64-bit GEP offset changed from `i64 -10` to `i64 4294967286`.

kinke:
That offset 'regression' was introduced by https://github.com/dlang/dmd/pull/14299. It doesn't depend on the compiler bitness (happens with 64-bit compiler too), but on the target bitness (so e.g. happens for 32-bit Linux targets too).

kinke:
These symbol-offsets are optimized/const-folded in the frontend, clamping to 32-bit for 32-bit targets [e.g., when negating `10` to `-10`]. So we should probably either revert to 32-bit GEPs for targets with less than 64 bits, or restore the `SymOffExp` special case for subtractions (using full 64-bit precision, independent from target) in the frontend again. Not sure what's better.

PR #4283, opened 2022-11-28
Fix miscompilation when adding a 64-bit constant to a symbol

### LINKED ISSUE #4264: Incorrect compilation when adding a 64-bit constant to a link-time address
This code adds a large constant to the address of main, but the addition doesn't actually seem to be performed.

```d
import core.stdc.stdio;
enum offset = 0xFFFF_FFFF_0000_0000UL;
extern (C) void main() {
    // auto offset = offset;
    printf("0x%llx\n", cast(ulong)&main);
    printf("0x%llx\n", cast(ulong)&main + offset);
    assert((cast(ulong)&main) != (cast(ulong)&main + offset));
}
```

`&main + offset` should not be equal to `main`, but it is with LDC 1.30.0 and DMD 2.098.0 (causing an assertion failure). The test passes with GDC 10.3.0 though. The problem only seems to occur when `offset` is a constant greater than 32 bits.

If I uncomment the line `auto offset = offset`, then it passes on all compilers.

Thanks for your help.

zyedidia:
The `DtoGEP` functions assume the offset is a 32-bit value, but when used with a `SymOffExp`, the offset may be 64-bit. In that case, the GEP should use an i64 for the offset.

Fixes #4264.

Let me know if I should make any changes. Thanks!

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -847,7 +847,7 @@ public:
         uint64_t elemSize = gDataLayout->getTypeAllocSize(elemType);
         if (e->offset % elemSize == 0) {
           // We can turn this into a "nice" GEP.
-          offsetValue = DtoGEP1(DtoType(base->type), baseValue, e->offset / elemSize);
+          offsetValue = DtoGEP1u64(DtoType(base->type), baseValue, e->offset / elemSize);
         }
       }
 
@@ -855,7 +855,7 @@ public:
         // Offset isn't a multiple of base type size, just cast to i8* and
         // apply the byte offset.
         offsetValue =
-            DtoGEP1(LLType::getInt8Ty(gIR->context()),
+            DtoGEP1u64(LLType::getInt8Ty(gIR->context()),
                     DtoBitCast(baseValue, getVoidPtrType()), e->offset);
       }
     }
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -847,7 +847,12 @@ public:
         uint64_t elemSize = gDataLayout->getTypeAllocSize(elemType);
         if (e->offset % elemSize == 0) {
           // We can turn this into a "nice" GEP.
-          offsetValue = DtoGEP1i64(DtoType(base->type), baseValue, e->offset / elemSize);
+          if (elemType->isStructTy()) {
+            // LLVM requires that struct offsets are 32-bit.
+            offsetValue = DtoGEP1(DtoType(base->type), baseValue, e->offset / elemSize);
+          } else {
+            offsetValue = DtoGEP1i64(DtoType(base->type), baseValue, e->offset / elemSize);
+          }
         }
       }
 
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -847,7 +847,7 @@ public:
         uint64_t elemSize = gDataLayout->getTypeAllocSize(elemType);
         if (e->offset % elemSize == 0) {
           // We can turn this into a "nice" GEP.
-          offsetValue = DtoGEP1u64(DtoType(base->type), baseValue, e->offset / elemSize);
+          offsetValue = DtoGEP1i64(DtoType(base->type), baseValue, e->offset / elemSize);
         }
       }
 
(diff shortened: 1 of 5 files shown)


gen/tollvm.cpp
```diff
@@ -385,6 +385,15 @@ LLConstant *DtoGEP(LLType *pointeeTy, LLConstant *ptr, unsigned i0,
                                               /* InBounds = */ true);
 }
 
+LLValue *DtoGEP1i64(LLType *pointeeTy, LLValue *ptr, int64_t i0, const char *name,
+                    llvm::BasicBlock *bb) {
+  // Use a 32-bit offset if i0 can fit, otherwise use a 64-bit offset.
+  if (i0 <= INT32_MAX && i0 >= INT32_MIN) {
```
JohanEngelen, line 391: is this needed?  (why not always use 64bit?)

JohanEngelen:
Thanks!
I'm wondering if we cannot just remove all 32bit versions of DtoGEP?

gen/tollvm.cpp
zyedidia: I wanted to avoid modifying the behavior when it's not necessary. I tried always using 64-bit, but then I had to change a test to expect `i64`, and then the test failed on Windows x86, possibly because some other part of the compiler is shrinking it down for that target? You can see the test log here: https://github.com/ldc-developers/ldc/actions/runs/3561370076/jobs/5982226332. I will look into it tonight.

zyedidia:
We could try doing that with a bit more effort. I initially tried to just make the inputs to `DtoGEP` `uint64_t` instead of `unsigned`, but I think I had errors regarding implicit conversion ambiguity (many inputs are `unsigned`, and maybe they could be implicitly converted to `uint64_t` or to `llvm::Value`, so the compiler doesn't know which to choose?). With some additional manual casts we could probably make all `DtoGEP` calls use 64-bit. I'll look into it tonight.

JohanEngelen:
Note that the LLVM Langref says that GEPs into structs can only have 32bit index.

tests/codegen/gh4264.d
```diff
@@ -0,0 +1,6 @@
+// RUN: %ldc -run %s
```
JohanEngelen, line 1: In addition to running this test, I think it's good to check the generated LLVM IR.

zyedidia:
I've updated the code so that it always uses a 64-bit offset in getelementptr except when the base is a struct type, and the test also checks the codegen. Not sure if we also want to replace all/most existing DtoGEPs with 64-bit versions since that would be a larger change.

zyedidia:
Hmm actually checking the generated IR is difficult because on 32-bit systems it doesn't even generate a getelementptr because the frontend does something different. Let me know if you have a solution in mind for that.

zyedidia:
I'm not sure if the macOS x64 test is failing because of this change, because the failure seems unrelated?

JohanEngelen:
> Hmm actually checking the generated IR is difficult because on 32-bit systems it doesn't even generate a getelementptr because the frontend does something different. Let me know if you have a solution in mind for that.

You can explicitly specify a 64bit target triple. Perhaps it's not worth it in a test though.

JohanEngelen:
> I'm not sure if the macOS x64 test is failing because of this change, because the failure seems unrelated?

looks unrelated to me too

gen/toir.cpp
```diff
@@ -847,15 +847,20 @@ class ToElemVisitor : public Visitor {
         uint64_t elemSize = gDataLayout->getTypeAllocSize(elemType);
         if (e->offset % elemSize == 0) {
           // We can turn this into a "nice" GEP.
-          offsetValue = DtoGEP1(DtoType(base->type), baseValue, e->offset / elemSize);
+          if (elemType->isStructTy()) {
+            // LLVM requires that struct offsets are 32-bit.
+            offsetValue = DtoGEP1(DtoType(base->type), baseValue, e->offset / elemSize);
```
JohanEngelen, line 852: https://llvm.org/docs/LangRef.html#getelementptr-instruction

JohanEngelen (approved):

zyedidia:
Sounds good, let me know if there are more changes you'd like me to make.

JohanEngelen:
let me think for a little. I wonder if there can be a negative performance impact on architectures where 64bit indexing is potentially costing more instructions than, say, 16bit indexing.

zyedidia:
I would imagine that when getelementptr is lowered to machine code, LLVM might select more efficient instructions if the offset can fit into 32 bits or 16 bits, etc.

JohanEngelen:
thanks again

PR #4297, opened 2023-01-11
Add -fno-delete-null-pointer-checks

JohanEngelen:
The commandline option name stems from GCC and Clang.


thewilsonator (approved):
Changelog?

PR #4299, opened 2023-01-15
Add ldc.attributes.callingConvention("...")

JohanEngelen:


gen/uda.cpp
```diff
@@ -366,6 +387,105 @@ void applyAttrAssumeUsed(IRState &irs, StructLiteralExp *sle,
   irs.usedArray.push_back(symbol);
 }
 
+/// Tries to recognize the calling convention,
+/// TODO: Add support "cc <n>"
+/// Returns true if succesful, with the calling convention in 'callconv'.
+/// Returns false if unsuccesful.
+bool parseCallingConvention(llvm::StringRef name,
```
thewilsonator, line 394: can we not defer to LLVM for this?

JohanEngelen: nope.

ljmf00: Would be cool to add this to LLVM itself, I'm sure clang and other compilers would see this useful there.

thewilsonator: Well how is calling convention parsed for LLVM IR? are the routines for that not public?

ljmf00: It uses a LLVM IR token based parser. See https://github.com/llvm/llvm-project/blob/main/llvm/lib/AsmParser/LLParser.cpp#L2046 .

kinke:
Just curious - what convention(s) are you guys interested in?

JohanEngelen:
> Just curious - what convention(s) are you guys interested in?

I don't know to be honest :)  maybe `vectorcall`.

gen/uda.cpp
JohanEngelen: this is just like many other string->enum things, most of it is not public. Like the supported asm instructions, or CPU target capabilities ("sse2" etc.)

ljmf00:
> Just curious - what convention(s) are you guys interested in?

`vectorcall` can be beneficial on cases where a very hot call is used and the compiler decides not to inline it due to frequent usage of it. Therefore the compiler generates prologue/epilogue based on the calling convention. `vectorcall` follows what `stdcall` and `fastcall` does but it uses `XMM` and `YMM` registers for 128-bit and 256-bit data types instead of incrementing the stack.

For very specific tasks, it can save a few registers on e.g. memory to memory move operations and therefore save a few cycles per call on those cases.

ljmf00 (approved):

kinke:
On which platform? On x86_64, I'd be surprised if it made a difference - the SysV ABI already uses registers for vectors, and for Windows, we already use `vectorcall` for `extern(D)`.

ljmf00:
> On which platform? On x86_64, I'd be surprised if it made a difference - the SysV ABI already uses registers for vectors, and for Windows, we already use `vectorcall` for `extern(D)`.

I thought that only vectorcall prioritizes XMM/YMM registers on the first parameters and also would use XMM registers for types with 128bit/256bit size, even though not being vector types. Isn't at least the later supposed to happen? I'm less confident that it will improve performance on my usecase now :/

JohanEngelen:
Indeed, judging by this table, `vectorcall` would be a pessimization over the standard convention on x86_64: https://en.wikipedia.org/wiki/X86_calling_conventions#List_of_x86_calling_conventions
(less registers and ZMM not used for vectorcall).

JohanEngelen:
(I figured it might be useful for embedded too, btw)

kinke:
The resulting ABI is a combination of our rewrites (e.g., https://github.com/ldc-developers/ldc/blob/master/gen/abi/x86-64.cpp, which is nowadays based on frontend code in https://github.com/dlang/dmd/blob/c0053682f50feabe79a2629b412e81fef776c72d/compiler/src/dmd/argtypes_sysv_x64.d#L21-L101 - it hasn't been extended to 512-bit vectors yet, but supports 256-bit ones IIRC) and the LLVM behavior (depending on the IR calling convention etc.). For LLVM, I would expect `vectorcall` to be equivalent to default `C` for *non-Windows* x86_64 targets.

> I thought that only vectorcall prioritizes XMM/YMM registers on the first parameters and also would use XMM registers for types with 128bit/256bit size, even though not being vector types. Isn't at least the later supposed to happen?

*Some* aggregates are passed in vector registers too. E.g., wrappers like `struct { double4 x; }`.

> (I figured it might be useful for embedded too, btw)

It certainly can come in handy in some situations. E.g., for `extern(C[++])` on Windows, we expect the default MSVC++ calling convention. This UDA would allow to interop with C[++] code using `__vectorcall` (either explicitly, or using a compiler flag IIRC).

PR #4301, opened 2023-01-21
Cirrus CI: Migrate Mac jobs to M1

kinke:


JohanEngelen:
nice that you are working on this.
I'm a bit lost with all the test systems we have (and the reasons for them), is it an idea to have a table in the readme with short explanation?

kinke:
This needs some more ironing out:
* The x64 emulation jobs fail some tests which involve Apple-clang (LLVM-plugin lit-tests, some druntime integration tests, more?), as it defaults to the arm64 target. And CRT destructors apparently aren't invoked anymore (a macOS 12 change? or some Rosetta problem?).
* The native arm64 jobs with vanilla LLVM 15 fail because there's no `FileCheck` for the lit-tests, then there are 3 failing dmd-testsuite tests, and the druntime integration tests hang.

Johan, you have an M1 box, right? ;)

4 Mac jobs are probably overkill for Cirrus (due to their limits - apparently only one Apple job can run at a time), so we could just test native arm64 with Cirrus, and keep on testing Mac x64 with Circle and GitHub or so.

Btw, Cirrus now also supports Linux arm64 containers, so we probably don't really *need* Travis anymore.

---

A table surely wouldn't hurt, but it might become outdated pretty quickly. Let me quickly summarize the status quo here:
* GitHub Actions:
  * `Main`: the big fat jobs producing most packages, incl. 3-stage compilation for most platforms (bootstrap compiler, PGO-instrumented compiler, then final compiler). Generating a package includes some integration tests, building dlang tools, dub, reggae and .tar.xz-ing the package etc. For non-tags, the jobs also use an LLVM with enabled assertions, further slowing them down (but enabling LLVM+LDC assertions).
  * `Vanilla`: rather quick jobs with a single stage, mainly testing different vanilla LLVM versions and D host compilers
* Travis: a 2-stage job (bootstrap + final compiler) producing the Linux AArch64 package. Also using an LLVM with enabled assertions for non-tags.
* Cirrus CI:
  * FreeBSD: a 2-stage job producing the FreeBSD package. Linked dynamically against the FreeBSD LLVM package.
  * Others: basically duplicates of the CircleCI jobs - quick 1-stage jobs (against the distro LLVM version on Linux), with some permutations of host compilers and multilib/shared-libs-only variants etc.
* CircleCI: see above; basically what the Travis jobs were years ago. I don't care much about these, just keeping them around for redundancy in case Cirrus goes down or imposes stricter limits for open-source projects etc.

kinke:
The performance of these M1 runners is outstanding. If the last step (defaultlib tests) wasn't hanging (most of the time - in some runs, it passed), we could expect the native arm64 CI jobs to take around 8 minutes. And that with 4 CPU cores - the Cirrus Linux x64 containers need 8 cores for similar timings.

JohanEngelen:
> The performance of these M1 runners is outstanding. If the last step (defaultlib tests) wasn't hanging (most of the time - in some runs, it passed), we could expect the native arm64 CI jobs to take around 8 minutes. And that with 4 CPU cores - the Cirrus Linux x64 containers need 8 cores for similar timings.

On my dev machine (M1, macOS Ventura) the sanitizer tests seem to take much longer than before (~8sec), and some just hang (so bad that the whole system becomes unresponsive and I have to hard shutdown of the system :S). Haven't found the cause yet...

JohanEngelen:
> On my dev machine (M1, macOS Ventura) the sanitizer tests seem to take much longer than before (~8sec), and some just hang (so bad that the whole system becomes unresponsive and I have to hard shutdown of the system :S). Haven't found the cause yet...

The underlying cause I mean. The hanging tests are TSan and LSan tests. Perhaps we should mark them as `// UNSUPPORTED: Darwin`

kinke:
For CI here, the situation has improved since switching to LDC-LLVM v14.0.3 (I forgot we already cross-build an arm64 package). Failures:
* lit-tests: `sanitizers/lsan_memleak.d`
* dmd-testsuite:
  * `runnable/test17868.d` (=> `crt_destructor` not invoked)
  * `runnable/test17868b.d` (=> `crt_destructor` not invoked)
  * `runnable/ldc_extern_weak.d`
  * `runnable_cxx/test22898.d
  * `fail_compilation/test15703.d`
  * `runnable/test34.d` (only failing with enabled optimizations)
* stdlib tests: `druntime-test-exceptions-release` (potentially only failing with shared druntime; worked with static druntime for the last CI job at least), then hanging (potentially also just with shared libs - didn't hang for the last `macOS 12 arm64` CI job)

The jobs are allowed to fail for now in this PR, so that these remaining issues can be ironed out later.

Issue #4293, opened 2022-12-31
Errors in inline assembly don't stop compilation


zyedidia:
Here is a program with some invalid inline assembly:

```d
void main() {
    asm {
        "a";
    }
}
```

I compile with

```
$ ldc2 asm.d
asm.d(3):1:2: error: invalid instruction mnemonic 'a'
        a
        ^
```

It correctly shows an error, but also produces an executable (it seems like it just removed the inline asm statement and compiled everything else), and the return code from compiling is 0 (success). This could go unnoticed if checking for success is automated based on the shell command's return code. I think the problem is that this error occurs during codegen, and there is no code to handle a codegen error in `driver/toobj.cpp` (that I could find).

PR #4302, opened 2023-01-22
driver: add support for a more generic diagnostic handler

ljmf00:

---

I can't add you as a reviewer @JohanEngelen since I don't have permissions on LDC org. This is something I need cherry-picked in weka to diagnose traces with possibly bigger stack usage than permitted.

The default seem to warn on stdout rather than stderr and also doesn't respect `-w` and `-wi` command line options and therefore it seem to not working on our build system diagnostics.

This also further improves the existing logic on diagnostic handler as it was wrongly checking the `DiagnosticInfoSrcMgr` severity and not incrementing the error counter.

I'm going to add tests for this tomorrow.


ljmf00:
Perhaps, we can get rid of default llvm diagnostic printer and use DMD error/warning functions.

thewilsonator:
Do you have a test case for this?

tests/codegen/attr_warn_stack_size.d
```diff
@@ -0,0 +1,14 @@
+module attr_warn_stack_size;
+// Test diagnostics for warnings on stack size
+
+// RUN: not %ldc -w -c %s 2>&1 | FileCheck %s
+
+import ldc.attributes;
+
+// CHECK: {{.*}}Warning: stack frame size (8) exceeds limit (1) in '_D20attr_warn_stack_size6foobarFiZf'
```
ljmf00, line 8: I'm defering to LLVM demangler. I'm working on upstreaming the demangler, but for now it like this, should I abstract this from the test too?

ljmf00:
> Do you have a test case for this?

From my PR description :)
> I'm going to add tests for this tomorrow.

driver/codegenerator.cpp
```diff
@@ -166,19 +167,68 @@ void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
...
+        inlineAsmDiagnostic(irs, DISM.getSMDiag(), DISM.getLocCookie());
+        return true;
+      }
+      case llvm::DK_StackSize:
+      {
+        const auto &DISS = llvm::cast<llvm::DiagnosticInfoStackSize>(DI);
+#if LDC_LLVM_VER > 1400
+        llvm::StringRef fname;
+        unsigned line, column;
+        DISS.getLocation(fname, line, column);
```
JohanEngelen, line 203: is getting the location generalizable for all cases? (i.e. call it before `switch` using `DI`)

driver/codegenerator.cpp
```diff
@@ -166,19 +167,68 @@ void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
...
+        const auto loc = Loc(fname.str().c_str(), line, column);
+#else
+        const auto loc = Loc(nullptr, 0, 0);
+#endif
+        switch (DISS.getSeverity())
```
JohanEngelen, line 208: this also seems generalizable (i.e. not individual switches for severity for each kind of diag info type?)

driver/codegenerator.cpp
```diff
@@ -166,19 +167,68 @@ void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
...
+            return false;
+        }
+
+        inlineAsmDiagnostic(irs, DISM.getSMDiag(), DISM.getLocCookie());
+        return true;
+      }
+      case llvm::DK_StackSize:
+      {
+        const auto &DISS = llvm::cast<llvm::DiagnosticInfoStackSize>(DI);
+#if LDC_LLVM_VER > 1400
```
JohanEngelen, line 200: this is likely wrong, because this disables the code for 14.0.0 but enables it for 14.0.1.  Should be `>=`.

driver/codegenerator.cpp
```diff
@@ -166,19 +167,68 @@ void inlineAsmDiagnosticHandler(const llvm::SMDiagnostic &d, void *context,
...
+#else
+        const auto loc = Loc(nullptr, 0, 0);
+#endif
+        switch (DISS.getSeverity())
+        {
+          case llvm::DS_Error:
+            error(loc, "stack frame size (%ld) exceeds limit (%ld) in '%s'",
+              DISS.getStackSize(),
+              DISS.getStackLimit(),
+              llvm::demangle(DISS.getFunction().getName().str()).c_str()
```
JohanEngelen, line 214: do we use the LLVM demangler anywhere else in the codebase? Not better to use D's own demangler?

JohanEngelen: (because then it is always up-to-date)

tests/codegen/attr_warn_stack_size.d
JohanEngelen: you can also use `{{.*}}foobar` to not depend on yes/no demangling. If demangling works, I'd make it part of the test, if it doesn't then remove demangling yes/no assumption from the test.

JohanEngelen:
Nice. I wonder if there are other LLVM attributes that can generate errors/warnings, such as `@target`. If so, please make a TODO issue :)

tests/codegen/attr_warn_stack_size.d
```diff
@@ -0,0 +1,14 @@
+module attr_warn_stack_size;
+// Test diagnostics for warnings on stack size
+
+// RUN: %ldc -wi -c %s 2>&1 | FileCheck %s
+
+import ldc.attributes;
+
+// CHECK: {{.*}}Warning: stack frame size (8) exceeds limit (1) in '{{.*}}foobar'
```
JohanEngelen, line 8: I think this will need {{.*}} after foobar too, because of the trailing `'`, (or just remove the `'` ;-)

ljmf00:
> Edit: even if we defer for other diagnostic kinds to the default LLVM's diagnostic function, we can still ++ global.errors or warnings, such that -w is respected.

Yeah, makes sense to me, but keep in mind that apparently, warning severities does write to stdout.

driver/codegenerator.cpp
ljmf00: > (because then it is always up-to-date, when building LDC with itself)

I don't think so. Makes sense to use the official one due to that, but the tradeoff is that is very slow. `core.demangle` use exceptions for control flow, which is an horrible implementation. I have a project to update the official demangler to a decent and fast parser, but I endedup getting out of time for these things :/

driver/codegenerator.cpp
ljmf00: I think I can change this to use the IR state loc, when its not available through the diagnostic info object.

ljmf00:
I can't really understand what I'm doing wrong in the test. Maybe its trivial but I'm not getting there.

kinke:
Btw, wrt. `++global.errors` etc. - according to https://github.com/ldc-developers/ldc/issues/4293, there's a missing check for such LLVM errors.

driver/codegenerator.cpp
JohanEngelen: The compiler is erroring, don't worry about performance, it's fine if it takes an extra second even. ;)

JohanEngelen:
> I can't really understand what I'm doing wrong in the test. Maybe its trivial but I'm not getting there.

To run lit-based tests locally, go into your build folder, go to `tests` folder, then run `./runlit.py -v .` or `./runlit.py -v codegen/attr_warn_stack_size.d`.

tests/codegen/attr_warn_stack_size.d
```diff
@@ -0,0 +1,15 @@
+// Test diagnostics for warnings on stack size
+
+// REQUIRES: atleast_llvm1400
+
+// RUN: %ldc -wi -c %s 2>&1 | FileCheck %s
```
JohanEngelen, line 5: This should be a `not` if you want it to error out, I think you had that correct in the first commit. But you need a fix for kinke's comment: https://github.com/ldc-developers/ldc/issues/4293

ljmf00:
> Btw, wrt. `++global.errors` etc. - according to #4293, there's a missing check for such LLVM errors.

Yes, it fixes it! I'll also add a test case for it.

PR #4304, opened 2023-01-22
Replace `sprintf` with safer `snprintf`

JohanEngelen:
fixes compilation warning on newer clang versions.

See https://issues.dlang.org/show_bug.cgi?id=23648 for frontend.

### DIFF
--- a/dmd/root/ctfloat.d
+++ b/dmd/root/ctfloat.d
@@ -208,6 +208,8 @@ extern (C++) struct CTFloat
     static real_t parse(const(char)* literal, bool* isOutOfRange = null);
     @system
     static int sprint(char* str, char fmt, real_t x);
+    @system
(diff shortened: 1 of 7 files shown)


thewilsonator (approved):

kinke:
Isn't there some C++ template which takes a ref to a char array, inferring the length automatically? I don't use C++ that much anymore...

JohanEngelen:
> Isn't there some C++ template which takes a ref to a char array, inferring the length automatically? I don't use C++ that much anymore...

yes: https://stackoverflow.com/questions/2384107/how-does-this-template-magic-determine-array-parameter-size

JohanEngelen:
> Isn't there some C++ template which takes a ref to a char array, inferring the length automatically? I don't use C++ that much anymore...

but this won't work for calls from D code. All use of CTFloat::snprint is through CTFloat::sprint, called from D code.

PR #4257, opened 2022-11-12
Add CI job for macOS with LLVM 15

JohanEngelen:

Issue #4261, opened 2022-11-13
Pointer comparisons no longer work with opaque pointers, code review needed


JohanEngelen:
With opaque pointers, we can no longer check if types are the same by comparing pointer types.
It lead to a miscompilation bug here: https://github.com/ldc-developers/ldc/pull/4257#issuecomment-1312846914 , fixed by https://github.com/ldc-developers/ldc/pull/4257/commits/e24567bf8bb0e6583310573fd7271af21ed17364

I think we need to review all uses of `->getType() ==` (and more of this type of code)
Potential bugs with opaque pointers (the list is longer):

```d
gen/irstate.cpp:
  137                                   LLConstant *initializer,
  138                                   Dsymbol *symbolForLinkageAndVisibility) {
  139:   if (initializer->getType() == globalVar->getValueType()) {
  140      defineGlobal(globalVar, initializer, symbolForLinkageAndVisibility);
```
```d
gen/llvmhelpers.cpp:
  462        }
  463  #else
  464:       assert(r->getType() == lit);
  465  #endif
  466      }
  ...
  514  
  515    LLValue *rval = DtoRVal(val);
  516:   if (rval->getType() == tolltype) {
  517      return new DImValue(_to, rval);
  518    }
  ...
  615      if (fromsz == tosz) {
  616        rval = DtoRVal(val);
  617:       assert(rval->getType() == tolltype);
  618      } else if (fromsz < tosz) {
  619        rval = new llvm::FPExtInst(DtoRVal(val), tolltype, "", gIR->scopebb());
  ...
  798      LLType *tll = DtoType(tb);
  799  
  800:     if (rval->getType() == tll) {
  801        return new DImValue(to, rval);
  802      }
```

Assert bugs:
```d
gen/dvalue.cpp:
   77  
   78  DConstValue::DConstValue(Type *t, LLConstant *con) : DRValue(t, con) {
   79:   assert(con->getType() == DtoType(t));
   80  }
   ...
  158  
  159  DSpecialRefValue::DSpecialRefValue(Type *t, LLValue *v) : DLValue(v, t) {
  160:   assert(v->getType() == DtoPtrToType(t)->getPointerTo());
  161  }
```

```d
gen/functions.cpp:
  896          // The argument is an appropriate lvalue passed by reference.
  897          // Use the passed pointer as parameter storage.
  898:         assert(irparam->value->getType() == DtoPtrToType(paramType));
```

JohanEngelen:
[...]

PR #4307, opened 2023-01-23
CI: Add vanilla LLVM 15 job on Linux

kinke:
Just testing a rebased #4257 + an LLVM 15 job on Linux, as v15.0.6 has an x86_64 Ubuntu package again...


kinke:
The defaultlib tests with LLVM 15 are hanging on Linux too...

JohanEngelen:
> The defaultlib tests with LLVM 15 are hanging on Linux too...

:(   I think it's https://github.com/ldc-developers/ldc/issues/4261 ?

JohanEngelen:
I forgot, do we default to `--opaque-pointers=false` with LLVM15? (higher chance of working)

thewilsonator:
Unless we changed it to true, no.

JohanEngelen:
> I forgot, do we default to `--opaque-pointers=false` with LLVM15? (higher chance of working)

Just tested, we default to `--opaque-pointers=true`.
We can force opaque pointers like this hack (needs better location but this works):
```diff
diff --git a/driver/main.cpp b/driver/main.cpp
index b36971195b..72890c14c5 100644
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -998,6 +998,7 @@ void registerPredefinedVersions() {
   }

 #if LDC_LLVM_VER >= 1400
+  getGlobalContext().setOpaquePointers(false);
   // A version identifier for whether opaque pointers are enabled or not. (needed e.g. for intrinsic mangling)
   if (!getGlobalContext().supportsTypedPointers()) {
     VersionCondition::addPredefinedGlobalIdent("LDC_LLVM_OpaquePointers");
```
It breaks some tests because we wrote the IR matching for opaque ptr work (e.g. missing bitcasts), but on my machine I also run into some LLVM assertions (possibly bugs in our code with pointer comparisons or smth).

LLVM15 fully supports typed pointers, so it's OK to force typed pointers if it makes LDC work. LLVM16 only supports typed pointers on "best-effort" basis, untested.

kinke:
Right, hitting stuff like this now after rebase + hack:
```
Stored value type does not match pointer operand type!
  store %nest.__unittest_L291_C9* %8, i8** %18, align 8, !dbg !516
```

kinke:
Okay, I seem to have found & fixed the culprit, see last commit. - With typed pointers now, the LLVM 15 job on Mac is green, while there's a single `druntime-test-exceptions-release` failure on Linux.

kinke:
Oh and I override LLVM's `-opaque-pointers` option now - in order to default to `false` for now, independent from LLVM version. When we have proper support for opaque pointers, we can get rid of it again and use LLVM's default setting.

thewilsonator (approved):

PR #4010, opened 2022-07-11
Bump LDC-LLVM to v14.0.6

kinke:

PR #4311, opened 2023-01-29
CI: Use LDC-LLVM v15.0.7 for prebuilt packages

kinke:


kinke:
We seem to be hitting a new assertion with `-cov` in general, which should account for most failures (`Attribute::isEnumAttrKind(Kind) && "Not an enum attribute"`).

And lit-test `codegen/dcompute_cl_images.d` still hits the same assertion as in https://github.com/ldc-developers/ldc/pull/4010#issuecomment-1191820165. FYI @thewilsonator

kinke:
The only remaining regression/blocker is `codegen/dcompute_cl_images.d`.

kinke:
And looks like the macOS-arm64 `crt_destructor` issues for dmd-testsuite (Cirrus CI) are gone with LLVM 15.

kinke:
Oh and 1 more TODO: checking for new CLI options inherited from LLVM 15, and hiding those.

kinke:
@thewilsonator: I went with marking `tests/codegen/dcompute_cl_images.d` as XFAIL with LLVM 15+ for now, as suggested in https://github.com/ldc-developers/ldc/pull/4010#issuecomment-1192142131.

kinke:
`tests/sanitizers/fuzz_asan.d` seems flaky again, sigh...

PR #4320, opened 2023-02-10
Be less conservative wrt. linker dead code elimination

kinke:


kinke:
Ouch, Apple's `ld64` apparently strips `rt_options` globals from executables linked against *shared* druntime. This might be a blocker for Apple targets.

kinke:
Very unsure about the last [`[tentative fix]`](https://github.com/ldc-developers/ldc/pull/4320/commits/c61516a501f24ed986a42ea70d039ae77c961e77) commit - has this really just worked accidentally so far (and only started to show up on macOS with shared druntime now with dead_strip)? It's about https://github.com/ldc-developers/ldc/blob/95ac5cadb89e2fde1d9501f3aac2dc8a077c5278/runtime/druntime/test/profile/src/profile.d#L17-L19

kinke:
The size of the bundled D executables (except for ldc2 and ldc-profdata, which are unchanged) from the macOS-x64 CI package is cut in half - from 59 MB to 30.4 MB; e.g., `ldmd2` from 4.4 to 1.2 MB.

JohanEngelen:
> Very unsure about the last [`[tentative fix]`](https://github.com/ldc-developers/ldc/pull/4320/commits/c61516a501f24ed986a42ea70d039ae77c961e77) commit - has this really just worked accidentally so far (and only started to show up on macOS with shared druntime now with dead_strip)? It's about
> 
> https://github.com/ldc-developers/ldc/blob/95ac5cadb89e2fde1d9501f3aac2dc8a077c5278/runtime/druntime/test/profile/src/profile.d#L17-L19

Sorry I don't fully understand. Why was the fix needed?

kinke:
The produced `.def` file didn't include the two string-compare functions (`dstrcmp` and `__cmp`) expected in https://github.com/ldc-developers/ldc/blob/95ac5cadb89e2fde1d9501f3aac2dc8a077c5278/runtime/druntime/test/profile/mytrace.def.exp - on macOS only, and only when linking *shared* druntime (tested by 2 CI jobs, both failed), and only with `dead_strip`. I don't know anything about how this profiling code works; and how the 2 functions were apparently in there before the `dead_strip`, even though the string comparisons are AFAICT exclusively triggered behind assertions, which are ignored with `-release` when running that test in release mode.

JohanEngelen:
> The produced `.def` file didn't include the two string-compare functions (`dstrcmp` and `__cmp`) expected in https://github.com/ldc-developers/ldc/blob/95ac5cadb89e2fde1d9501f3aac2dc8a077c5278/runtime/druntime/test/profile/mytrace.def.exp - on macOS only, and only when linking _shared_ druntime (tested by 2 CI jobs, both failed), and only with `dead_strip`. I don't know anything about how this profiling code works; and how the 2 functions were apparently in there before the `dead_strip`, even though the string comparisons are AFAICT exclusively triggered behind assertions, which are ignored with `-release` when running that test in release mode.

yeah ok... I'm OK with your fix :)

PR #4322, opened 2023-02-12
Allow RISC-V ABI to be configured by mabi

### LINKED ISSUE #4321: No options to control the RISC-V ABI
The latest release of LDC changed the default ABI of RISCV64 to `lp64d` (default used to be `lp64`). Unfortunately this value is not currently configurable via a command-line flag (with GCC one would use `-mabi=lp64`). I would like to continue using `lp64` (because I am not using float/double support), but can't do so with the latest release of LDC. It would be great to have a new option (`-mabi`) to let the user configure the ABI.

Alternatively, the RISCV ABI could be set to `lp64` by default, and automatically change to `lp64d` if the `+d` attribute is passed in.

zyedidia:
This PR allows the RISC-V ABI to be configured by the user. RISC-V has the following ABIs: `lp64`, `lp64f`, `lp64d` (64-bit only), and `ilp32`, `ilp32f`, `ilp32d` (32-bit only). See [here](https://wiki.gentoo.org/wiki/RISC-V_ABIs) for a slightly more detailed description, and [here](https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html) for the GCC documentation on this.

The latest LDC release changed the default 64-bit ABI from `lp64` to `lp64d`. I think the default should continue to be `lp64` because it is the most basic, and additional features/extensions should be opt-in (this PR changes back to `lp64` default). The latest release also enabled the D extension by default. This PR undoes that and leaves the D extension as opt-in (with `-mattr=+d`), just like every other extension.

The ABI is now configurable using `-mabi=...` instead of hard-coding to `lp64d`. The `mabi` option is also no longer hidden. The default ABI on riscv64 is now `lp64` and on riscv32 is `ilp32`. Also, the backend linker now gets passed the appropriate `mabi` flags and `march` flags. Instead of hard-coding to `mabi=lp64d` and `march=rv64gc` (only for 64-bit though), now the value passed via `mabi` is used (or the default ABI), and `march` is properly constructed from the enabled features (passed via `mattr`). For example, to compile for `rv64gc` (shorthand for `rv64imacfdc_zicsr_zifencei`) with `lp64d` one would now pass:

```
-mtriple=riscv64-unknown-elf -mattr=+m,+a,+c,+f,+d,+c -mabi=lp64d
```

The backend linker/gcc driver would be passed `-march=rv64gc -mabi=lp64d`. Here are some other examples:

```
-mtriple=riscv32-unknown-elf -mattr=+m,+a,+c -mabi=ilp32
gcc/linker flags: -march=rv32imac_zicsr_zifencei -mabi=ilp32
```

```
-mtriple=riscv64-unknown-elf -mattr=+m,+a,+c,+f -mabi=lp64f
gcc/linker flags: -march=rv64imafc_zicsr_zifencei -mabi=lp64f
```

```
-mtriple=riscv64-unknown-elf
gcc/linker flags: -march=rv64i_zicsr_zifencei -mabi=lp64
```

Since Phobos probably needs to be built with `rv64gc` and `lp64d` the flags used to invoke the build should be

```
-mtriple=riscv64-unknown-elf -mattr=+m,+a,+c,+f,+d,+c -mabi=lp64d
```

I'm not sure if code in other repos needs to be changed as a result of this.

Fixes #4321.


driver/tool.cpp
```diff
@@ -120,13 +121,46 @@ void appendTargetArgsForGcc(std::vector<std::string> &args) {
...
+      std::string march;
+      if (triple.isArch64Bit())
+        march = "rv64";
+      else
+        march = "rv32";
+      bool m = features.find("+m") != std::string::npos;
+      bool a = features.find("+a") != std::string::npos;
+      bool f = features.find("+f") != std::string::npos;
+      bool d = features.find("+d") != std::string::npos;
+      bool c = features.find("+c") != std::string::npos;
```
kinke, line 140: Thanks for the quick reaction! Looks pretty good in general. I'm just a bit worried here wrt. feature parsing without splitting the string at commas; e.g., a potential `+myfeature` being mistaken as `+m`.

JohanEngelen: Indeed only "+m" should do the right thing, and not e.g. "+mm".
Also, what should `"+m,+d,-m"` do?

zyedidia: Good points, thanks! I’ll look into it tonight.

zyedidia: I have improved the parsing of the feature string. Now it will split based on commas, and it will take the last value for the feature (this follows how LLVM internally parses `-mattr` with multiple conflicting occurrences of the same feature). For example, `+m,+d,-m` becomes `rv64id_zicsr_zifencei` (`m` is disabled).

driver/tool.cpp
```diff
@@ -123,21 +123,34 @@ void appendTargetArgsForGcc(std::vector<std::string> &args) {
...
       std::string mabi = getABI(triple);
       args.push_back("-mabi=" + mabi);
 
+      extern llvm::TargetMachine* gTargetMachine;
+      auto featuresStr = gTargetMachine->getTargetFeatureString();
+      llvm::SmallVector<llvm::StringRef, 8> features;
+      featuresStr.split(features, ",", -1, false);
+
+      auto hasFeature = [&features](llvm::StringRef feature) {
+        bool has = false;
```
JohanEngelen, line 135: please add a comment about what behavior is desired.
"+m" and "-m", ordering,... At first I thought an early return on line 138 would work, but I now realize that the ordering and traversing the whole list is important

JohanEngelen: If you want to prevent regression: add a testcase for it in the Lit testsuite.

zyedidia: I have improved the feature parser and added a comment, and added a couple of test cases.

driver/tool.cpp
```diff
@@ -120,13 +121,62 @@ void appendTargetArgsForGcc(std::vector<std::string> &args) {
     }
     return;
 
-  case Triple::riscv64:
-    if (triple.isArch64Bit()) {
-      args.push_back("-march=rv64gc");
-      args.push_back("-mabi=lp64d");
+  case Triple::riscv64:;
```
JohanEngelen, line 124:
```suggestion
  case Triple::riscv64:
```

JohanEngelen:
LGTM

JohanEngelen:
It's OK like this, but you could merge all lit tests into one file, and then use --check-prefix to have different CHECKs per FileCheck invocation.

andreas-schwab:
This probably breaks building druntime.

core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs0, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs1, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs2, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs3, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs4, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs5, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs6, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs7, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs8, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs9, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs10, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs11, 0(a2)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs0, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs1, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs2, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs3, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs4, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs5, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs6, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs7, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs8, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs9, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs10, 0(a0)
        ^
core/thread/osthread.d(1594):1:2: error: instruction requires the following: 'D' (Double-Precision Floating-Point)
        fsd fs11, 0(a0)
        ^
make[2]: *** [runtime/CMakeFiles/druntime-ldc-shared.dir/build.make:346: runtime/objects-shared/core/atomic.o] Error 1

zyedidia:
I think the builder should just be updated to compile with `-mattr=+m,+a,+c,+f,+d,+c -mabi=lp64d`.

andreas-schwab:
When configured for riscv64 the default should always be rv64gc.

zyedidia:
I think that would be reasonable if there were a way to specify the exact RISC-V extensions to use using a flag like `-march`. This is how Clang operates: the default (if no flags are passed) is `rv64imac` for bare-metal targets (`riscv64-unknown-elf`), and `rv64gc` for hosted targets (`riscv64-linux-elf`), but `-march` can be used to override the default and specify `rv64i`, `rv64gc`, or something else.

Currently with LDC, if the default were changed to be `rv64gc`, then to use `rv64im`, the user would have to pass `-mattr=-a,-c,-f,-d,-c`, which is un-ergonomic and strange because it relies on the default and specifies the extensions to remove from the default instead of the ones to include from the base. If we make `-march` work like Clang, then I would be in favor of changing the default to `rv64gc` or `rv64imac`.

PR #4323, opened 2023-02-12
Upgrade frontend & libs to v2.102.1

kinke:


kinke:
Looking pretty good already. The biggest TODO from our side is probably the adaptation to vector comparisons, which now yield a vector mask, not a scalar bool anymore.

kinke:
Oh and looks like the macOS-arm64 CI jobs don't hang anymore for some druntime integration test. :)

PR #4328, opened 2023-02-24
Try to fix tests/sanitizers/fuzz_asan.d flakiness with LLVM 15

kinke:


thewilsonator (approved):

tests/sanitizers/fuzz_asan.d
```diff
@@ -12,7 +12,7 @@ bool FuzzMe(ubyte* data, size_t dataSize)
            data[0] == 'F' &&
            data[1] == 'U' &&
            data[2] == 'Z' &&
-    // CHECK: stack-buffer-overflow
+    // CHECK: ERROR: AddressSanitizer: {{stack-buffer-overflow|stack-use-after-return}}
```
JohanEngelen, line 15: huh? This is quite a strange bug, because you have to explicitly enable checking for stack-use-after-return (fakestack)...

kinke: I take it we're definitely not to blame for this flakiness - it's apparently stack-buffer-overflow most of the time, but can sometimes by the other with LLVM v15. E.g., https://github.com/ldc-developers/ldc/actions/runs/4266350680/jobs/7426765564.

PR #4329, opened 2023-02-25
Specify explicit seed for test, to see if it reproduces on CI server

JohanEngelen:
related to https://github.com/ldc-developers/ldc/pull/4328

The seed comes from this failed CI run: https://github.com/ldc-developers/ldc/actions/runs/4266350680/jobs/7426765564

### DIFF
--- a/tests/sanitizers/fuzz_asan.d
+++ b/tests/sanitizers/fuzz_asan.d
@@ -12,7 +12,7 @@ bool FuzzMe(ubyte* data, size_t dataSize)
            data[0] == 'F' &&
            data[1] == 'U' &&
            data[2] == 'Z' &&
-    // CHECK: ERROR: AddressSanitizer: {{stack-buffer-overflow|stack-use-after-return}}
+    // CHECK: ERROR: AddressSanitizer: stack-buffer-overflow
     // CHECK-NEXT: READ of size 1
     // CHECK-NEXT: #0 {{.*}} in {{.*fuzz_asan6FuzzMe.*}} {{.*}}fuzz_asan.d:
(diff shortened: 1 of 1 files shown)


JohanEngelen:
I tried on Mac but it does not reproduce with this seed, probably because seeds / inputs / instrumentation is slightly different between platforms and thus won't reproduce the testcase nor the failure.

JohanEngelen:
The seed does not trigger the same error (use after return). Very strange ..

kinke:
The new regex is still in place, so you wouldn't know?

JohanEngelen:
It reproduces on the CI server, but I am not able to reproduce in a docker image x86_64 at home, using Ubuntu 18.04 with ldc-latest-ci (`LDC - the LLVM D compiler (1.31.0-git-d89c00c): based on DMD v2.101.2 and LLVM 15.0.7`). CI uses LLVM LLVM 15.0.6, could that matter?

kinke:
I've seen it fail for the `Main` GH Actions workflow too (which uses our LLVM v15.0.7), e.g., https://github.com/ldc-developers/ldc/actions/runs/4263313148/jobs/7419888009 (seed 3337776564). And I think for the Linux aarch64 Cirrus CI job too.

JohanEngelen:
Ah, I now see that stack-use-after-return is automatically _enabled_ with LLVM15 (we set it to "runtime"), I didn't know that.
https://clang.llvm.org/docs/AddressSanitizer.html#stack-use-after-return-uar

JohanEngelen:
Closing, because I think the fix is OK.

PR #4331, opened 2023-02-26
Fix regression with LLVM 13+: some errors in inline assembly don't stop compilation

kinke:
Based on Luís' #4302.


kinke:
We already have a bunch of inline asm diagnostic `fail_compilation` tests, also testing the exit code - so looks like we only missed a subset of the LLVM errors.

thewilsonator (approved):

PR #4334, opened 2023-02-26
Add -fsanitize-address-use-after-return

JohanEngelen:
Make this optional for users, especially on Linux. Because LLVM's default has changed, and using UAR requires runtime support.
relates to https://github.com/ldc-developers/ldc/pull/4328


JohanEngelen:
The option is copied from Clang, see description here: https://clang.llvm.org/docs/AddressSanitizer.html#stack-use-after-return-uar

kinke (approved):

PR #4335, opened 2023-02-26
Add timetrace2txt tool for simple ftime-trace workflow.

JohanEngelen:
It's quite rudimentary, but works and improvements can always be made later ;-)


JohanEngelen:
Running into compilation problems of the tool, because it uses the host/bootstrap compiler to build the tool. Shouldn't we switch to building our tools with the just-built LDC ?

JohanEngelen:
You can also view the tools as an extra testcase. Apparently LDC 1.24 cannot deal with getting `ulong` values from JSON. Pretty bad, but OK...

JohanEngelen:
It's a bit disappointing how difficult it is to write such a simple tool using the standard library when multiple compiler versions are involved :'(

JohanEngelen:
OK to get this in?

thewilsonator (approved):

kinke:
Sure (haven't looked at it). Nice that you got it working for ~every compiler/Phobos. :+1:

PR #4339, opened 2023-03-01
ftime-trace: add tracing of CTFE !

JohanEngelen:
`--ftime-trace` was not yet tracing CTFE, a major oversight... With these changes it now _does_ trace CTFE. The start, and function calls during CTFE. I think that gives the user enough info to pinpoint where time is spent.

Testcase gives this output:

image

@rikkimax

### DIFF
--- a/dmd/dinterpret.d
+++ b/dmd/dinterpret.d
@@ -98,7 +98,7 @@ public Expression ctfeInterpret(Expression e)
 
     auto rgnpos = ctfeGlobals.region.savePos();
 
-    version (LDC)
+    version (IN_LLVM)
     {
         import driver.timetrace, std.format, std.conv;
         auto timeScope = TimeTraceScope(text("CTFE start: ", e.toChars()), e.toChars().to!string, e.loc);
@@ -436,7 +436,7 @@ private Expression interpretFunction(UnionExp* pue, FuncDeclaration fd, InterSta
         printf("\n********\n%s FuncDeclaration::interpret(istate = %p) %s\n", fd.loc.toChars(), istate, fd.toChars());
     }
 
-    version (LDC)
+    version (IN_LLVM)
     {
         import driver.timetrace, std.format, std.conv;
         scope dlg = () {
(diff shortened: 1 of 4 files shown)


JohanEngelen:
about CI errors: Perhaps `dmd-2.102.2-beta.1` miscompiles ?
(it's annoying to test with DMD on macOS, because I have to rebuild LLVM for x86_64 arch......)

JohanEngelen:
> about CI errors: Perhaps `dmd-2.102.2-beta.1` miscompiles ?

Pff, I'm a bit rusty it seems... `version (LDC) ---> version (IN_LLVM)`. doh!

JohanEngelen:
@kinke I'd like to get this in the release, automerge enabled.

JohanEngelen:
@rikkimax This should be available in `ldc-latest-ci` (install.sh), see here: https://github.com/ldc-developers/ldc/releases/tag/CI

rikkimax:
There appears to be something wrong with that build (even without time-trace), its take >3 minutes to build what should take sub-40 seconds normally. Was it not optimized?

Perfetto can no longer work with the resulting trace, when scrolled out. Ah I think I know why, ``CTFE start: [33, 33, 33, 33, 3`` ... giant Unicode table here. Yeahhhh the name field needs to be shrunk. If the name is bigger than what a symbol name can be please truncate. After all how many kilobytes do you need to know its a table?

I've included a screenshot of a problem module, and yeahhhh that tells a much better story.
Capture

JohanEngelen:
> There appears to be something wrong with that build (even without time-trace), its take >3 minutes to build what should take sub-40 seconds normally. Was it not optimized?
> 
> Perfetto can no longer work with the resulting trace, when scrolled out. Ah I think I know why, `CTFE start: [33, 33, 33, 33, 3` ... giant Unicode table here. Yeahhhh the name field needs to be shrunk. If the name is bigger than what a symbol name can be please truncate. After all how many kilobytes do you need to know its a table?

please submit bug reports whenever you find something ;-).  (this one I can remember)

PR #4345, opened 2023-03-12
Upgrade frontend & libs to v2.103.1

kinke:


kinke:
Almost ready on Posix. On Windows, I guess we're hitting some exceeds-max-cmdline-length problem in a .bat generated by ninja.

A bigger thing to tackle still is the new `_d_newclassT` template lowering, and actually using it in LDC. We've customized the old druntime hook with our extra `_d_allocclass` hook, so this probably needs some special-casing for LDC (skipping the init symbol blit). And we'll also need to make sure the GC2Stack pass still works, which might be laborious.

thewilsonator:
Does this also affect GC2Stack optimisations?

kinke:
Sure, as it works with mangled names. After switching to the templated lowering, it's not sufficient to look for `_d_allocclass` functions anymore, but we'll need to handle all `_d_newclassT` instantiations. Plus it isn't called with a `TypeInfo_Class` argument anymore, so we'll need to derive the class traits differently. So probably laborious and thankless to keep it up.

maxhaton:
Does the gc2stack thing actually get used to much effect? I've never really found it particularly useful

dmd/lexer.d
```diff
@@ -123,6 +126,8 @@ class Lexer
         this.tokenizeNewlines = false;
         this.inTokenStringConstant = 0;
         this.lastDocLine = 0;
+        this.eSink = errorSink;
+        assert(errorSink);
```
kinke, line 130: For 32-bit Windows, this assert is hit when compiling the default libs - if building LDC with LTO. Both when using itself as host compiler, and latest v1.32.1. An LTO miscompile, wow.

kinke:
Green now.

I'll open an issue about enabling & codegen'ing the `_d_newclassT` lowering in the future. There's more to come (https://github.com/dlang/dmd/pull/14664), so it probably makes sense to switch when the new lowerings are complete and adapt GC2stack then.

PR #4351, opened 2023-03-15
Fix C assert calls for newlib

kubo39:


driver/main.cpp
```diff
@@ -826,6 +826,8 @@ void registerPredefinedTargetVersions() {
       VersionCondition::addPredefinedGlobalIdent("DRuntime_Use_Libunwind");
     } else if (global.params.isUClibcEnvironment) {
       VersionCondition::addPredefinedGlobalIdent("CRuntime_UClibc");
+    } else if (global.params.isNewlibEnvironment) {
+      VersionCondition::addPredefinedGlobalIdent("CRuntime_Newlib");
```
kinke, line 830: Is newlib a thing on Linux? It's currently only predefined when explicitly targeting bare-metal, see line ~906.

gen/runtime.cpp
```diff
@@ -364,6 +364,7 @@ llvm::Function *getRuntimeFunction(const Loc &loc, llvm::Module &target,
 //                             const char *funcname);
 // uClibc:  void __assert(const char *assertion, const char *filename, int linenumber,
 //                        const char *function);
+// newlib:  void __assert(const char *file, int line, const char *failedexpr)
```
JohanEngelen, line 367: nitpick: can you move this up, below Android? Then it's crystal clear that they have the same signature.

JohanEngelen:
Tangential to this PR, I really think we should have at least one testcase for newlib (`/tests/driver/`). In that testcase, you can test simple things like whether the `version` identifier is indeed set, and whether C assert function is correct, stuff like that.

JohanEngelen:
can you add a testcase for this? (it would also test newlib triple etc.)

JohanEngelen:
lgtm

PR #4366, opened 2023-04-06
Add pragma(musttail)

MrSmith33:
Discussed in https://github.com/ldc-developers/ldc/discussions/4355

Open questions:
- pragma name. Is `LDC_musttail` what we want? Or we should use `musttail`?
- Need to add tests.

### DIFF
--- a/dmd/expression.d
+++ b/dmd/expression.d
@@ -5077,6 +5077,10 @@ extern (C++) final class CallExp : UnaExp
     bool directcall;        // true if a virtual call is devirtualized
     bool inDebugStatement;  /// true if this was in a debug statement
     bool ignoreAttributes;  /// don't enforce attributes (e.g. call @gc function in @nogc code)
+version (IN_LLVM)
+{
+    bool isMustTail;        // If marked with pragma(musttail)
+}
     VarDeclaration vthis2;  // container for multi-context
 
     extern (D) this(const ref Loc loc, Expression e, Expressions* exps)
--- a/dmd/id.d
+++ b/dmd/id.d
@@ -555,6 +555,7 @@ immutable Msgtable[] msgtable =
     { "LDC_global_crt_dtor" },
     { "LDC_extern_weak" },
     { "LDC_profile_instr" },
+    { "musttail" },
 
     // IN_LLVM: LDC-specific traits
     { "targetCPU" },
--- a/dmd/statementsem.d
+++ b/dmd/statementsem.d
@@ -2135,6 +2135,13 @@ else
                 return setError();
             }
         }
+        else if (ps.ident == Id.musttail)
+        {
+            version (IN_LLVM)
+            {
+                pragmaMustTailSemantic(ps);
+            }
+        }
         else if (!global.params.ignoreUnsupportedPragmas)
         {
             ps.error("unrecognized `pragma(%s)`", ps.ident.toChars());
@@ -2151,6 +2158,38 @@ else
             ps._body = ps._body.statementSemantic(sc);
         }
         result = ps._body;
+    }
+
+    version (IN_LLVM)
+    private void pragmaMustTailSemantic(PragmaStatement ps)
+    {
+        if (!ps._body)
+        {
+            ps.error("`pragma(musttail)` must be attached to a return statement");
+            return setError();
+        }
+
+        auto rs = ps._body.isReturnStatement();
+        if (!rs)
+        {
+            ps.error("`pragma(musttail)` must be attached to a return statement");
+            return setError();
+        }
+
+        if (!rs.exp)
+        {
+            ps.error("`pragma(musttail)` must be attached to a return statement returning result of a function call");
+            return setError();
+        }
+
+        auto ce = rs.exp.isCallExp();
+        if (!ce)
+        {
+            ps.error("`pragma(musttail)` must be attached to a return statement returning result of a function call");
+            return setError();
+        }
+
+        ce.isMustTail = true;
     }
 
     override void visit(StaticAssertStatement s)
--- a/gen/tocall.cpp
+++ b/gen/tocall.cpp
@@ -841,7 +841,8 @@ static LLValue *DtoCallableValue(llvm::FunctionType * ft,DValue *fn) {
 
 // FIXME: this function is a mess !
 DValue *DtoCallFunction(const Loc &loc, Type *resulttype, DValue *fnval,
-                        Expressions *arguments, LLValue *sretPointer) {
+                        Expressions *arguments, LLValue *sretPointer,
+                        bool isMustTail) {
   IF_LOG Logger::println("DtoCallFunction()");
   LOG_SCOPE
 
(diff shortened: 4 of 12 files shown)


dmd/expression.d
```diff
@@ -5077,6 +5077,7 @@ extern (C++) final class CallExp : UnaExp
     bool directcall;        // true if a virtual call is devirtualized
     bool inDebugStatement;  /// true if this was in a debug statement
     bool ignoreAttributes;  /// don't enforce attributes (e.g. call @gc function in @nogc code)
+    bool isMustTail;        // If marked with pragma(LDC_musttail)
```
kinke, line 5080: ~All LDC-specific mods in frontend files are to be put inside `version (IN_LLVM)` / `#if IN_LLVM` (for handling future merge conflicts).

kinke:
> Or we should use `musttail`?

I like that better - I assume GDC could implement it too.

dmd/id.d
```diff
@@ -555,6 +555,7 @@ immutable Msgtable[] msgtable =
     { "LDC_global_crt_dtor" },
     { "LDC_extern_weak" },
     { "LDC_profile_instr" },
+    { "LDC_musttail" },
```
MrSmith33, line 558: Should I move it to `// For pragma's` section after renaming to `musttail`? Or leave it in `// IN_LLVM: LDC-specific pragmas`?

dmd/expression.d
MrSmith33: done

dmd/frontend.h
```diff
@@ -7418,6 +7418,7 @@ class CallExp final : public UnaExp
     bool directcall;
     bool inDebugStatement;
     bool ignoreAttributes;
+    bool isMustTail;
```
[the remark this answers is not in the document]
MrSmith33: removed for now

dmd/statementsem.d
```diff
@@ -2135,6 +2135,10 @@ else
                 return setError();
             }
         }
+        else if (ps.ident == Id.musttail)
+        {
+            pragmaMustTailSemantic(ps);
+        }
```
kinke, line 2141: This is a frontend file too (everything in `dmd/`), so `IN_LLVM` again. Note that there's a `IN_LLVM` global/enum somewhere too, so that it's easy to use in `if` expressions.

And `version (IN_LLVM)` for `pragmaMustTailSemantic()` below too.

kinke:
Wrt. a testcase, see e.g. https://github.com/ldc-developers/ldc/blob/master/tests/codegen/array_alloc_gh3041.d for a simple example checking the generated IR. The stuff in `{{…}}` is a regex. See https://llvm.org/docs/CommandGuide/FileCheck.html for the full reference if needed.

dmd/statementsem.d
MrSmith33: What should non-ldc compilers do for this pragma? Will this file be syncronized with dmd at some point?

gen/funcgenstate.cpp
```diff
@@ -103,10 +104,10 @@ FuncGenState::FuncGenState(IrFunction &irFunc, IRState &irs)
     : irFunc(irFunc), scopes(irs), jumpTargets(scopes), switchTargets(),
       irs(irs) {}
 
-LLCallBasePtr FuncGenState::callOrInvoke(llvm::Value *callee,
+LLCallBasePtr FuncGenState::callOrInvoke(const Loc &loc, llvm::Value *callee,
                                          llvm::FunctionType *calleeType,
                                          llvm::ArrayRef<llvm::Value *> args,
-                                         const char *name, bool isNothrow) {
+                                         const char *name, bool isNothrow, bool isMustTail) {
```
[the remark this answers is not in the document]
kinke: Hmm, that probably only works for trivial return value types. Another option: giving `DtoCallFunction()` and optional `std::function` param, which gets the `LLCallBasePtr call` as argument and gets a chance to manipulate it.

dmd/statementsem.d
kinke: Upstreaming our frontend diff is probably not realistic; some of it are hacks, and some others really totally LDC-specific and would just clutter the DMD frontend IMO. If another compiler supported it, we'd upstream a proper semantic for the new pragma though.

JohanEngelen:
@ibuclaw Can GDC implement this too (or does it already have it?) ?

gen/tocall.cpp
```diff
@@ -1065,6 +1066,18 @@ DValue *DtoCallFunction(const Loc &loc, Type *resulttype, DValue *fnval,
       llvm::AttrBuilder(call->getAttributes(), LLAttributeList::FunctionIndex));
 #endif
   call->setAttributes(attrlist);
+  if (isMustTail) {
+    if (auto ci = llvm::dyn_cast<llvm::CallInst>(call)) {
+      ci->setTailCallKind(llvm::CallInst::TCK_MustTail);
+    } else {
+      if (!tf->isnothrow()) {
+        error(loc, "cannot perform tail-call - callee must be nothrow");
```
JohanEngelen, line 1074: can this error not be caught during semantic analysis?

gen/tocall.cpp
```diff
@@ -1065,6 +1066,18 @@ DValue *DtoCallFunction(const Loc &loc, Type *resulttype, DValue *fnval,
...
+      } else {
+        error(loc, "cannot perform tail-call - no code like destructors or scope(exit) should run after the call");
```
JohanEngelen, line 1076: same as above, can this error not be caught during semantic analysis?

thewilsonator:
> I like that better - I assume GDC could implement it too.

cc @ibuclaw

gen/tocall.cpp
MrSmith33: Where in semantic can I check this? I'm not really familiar with the code.

dmd/statementsem.d
maxhaton: Clutter is meaningless versus the asymptotic cost of complexity - it should be upstreamed. LDC is the real D compiler, the codebase should be honest about that

maxhaton: Slightly less abstract: we should get a way upstream for there to no more IN_LLVM and so on other than registering a behaviour (and not "just override a weak symbol")

ibuclaw:
> > > I like that better - I assume GDC could implement it too.
> > 
> > cc @ibuclaw 
> 
> https://github.com/gcc-mirror/gcc/blob/8b5d577caa32145b24acb948117f628a044028be/gcc/tree.h#L753

However, if you think you need musttail, you're probably wrong or mistaken. Compiler will automatically do sibling call optimizations at least at `-O2`.

DmitryOlshansky:
> > However, if you think you need musttail, you're probably wrong or mistaken. Compiler will automatically do sibling call optimizations at least at `-O2`.
> 
> 
> 
> For me the usecase is doing VM dispatch with tailcalls. This should work even without optimizations, otherwise you will get infinite recursion. This is an easy alternative to computed goto, which we also do not have.

Very much this! We do not have computed goto (for the better perhaps) but forced tail call allows for efficient VM implementations.

gen/tocall.cpp
deadalnix: Why does the callee need to be nothrow? Sure, the stack trace will be incomplete, but there is no reason to prevent it.

tests/codegen/musttail_1.d
```diff
@@ -0,0 +1,14 @@
+// Tests successfull musttail application
+
+// RUN: %ldc -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+// CHECK-LABEL: define{{.*}} @{{.*}}foo
+int foo(int x) nothrow
+{
+    // CHECK: musttail call{{.*}} @{{.*}}bar
+    // CHECK-NEXT: ret i32
+    pragma(musttail) return bar(x);
```
deadalnix, line 10: It would be worth adding another test that checks that the tail call also works without the return, when the function returns void.

It would also be beneficial to test for different function signatures.

Last, the failure case with scope exit and destructors would benefit from a test as well.

dmd/statementsem.d
deadalnix: Clearly out of scope for this PR, but yes @maxhaton , you are correct.

PR #4367, opened 2023-04-06
CI: Bump Ubuntu 18 jobs to Ubuntu 20

kinke:
As GitHub Actions eagerly removed support for it (5 years old now).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # LDC master
 
 #### Big news
+- The prebuilt Linux packages are now generated on a Ubuntu 20.04 box, so the min required `glibc` version has been raised from 2.26 to 2.31. (#4367)
 
 #### Platform support
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,14 +100,14 @@ commonSteps: &commonSteps
 
 version: 2
 jobs:
-  Ubuntu-18.04-multilib-rtSanitizers:
+  Ubuntu-20.04-multilib-rtSanitizers:
     <<: *commonSteps
     docker:
-      - image: ubuntu:18.04
+      - image: ubuntu:20.04
     environment:
       - PARALLELISM: 2
       - CI_OS: linux
-      - EXTRA_APT_PACKAGES: llvm-9-dev libclang-common-9-dev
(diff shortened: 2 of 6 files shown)


thewilsonator (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # LDC master
 
 #### Big news
+- The prebuilt Linux packages are now generated on a Ubuntu 20.04 box, so the min required `glibc` version has been raised. (#4367)
```
Geod24, line 4:
```suggestion
- The prebuilt Linux packages are now generated on a Ubuntu 20.04 box, so the min required `glibc` version has been raised from 2.26 to 2.31. (#4367)
```

AFAICS

kinke: Nice, thx for looking it up.

kinke:
Does anyone have a clue wrt. the ASan failures with vanilla LLVM 14 & 15 on Ubuntu 20?
```
error: failed to decompress '.debug_aranges', zlib is not available
```

Both `zlib1g` and `libzstd` are installed in the GH Actions Ubuntu 20 image. The LLVM binaries are totally unchanged. Adding their `bin` directory to PATH didn't fix it either (I assumed maybe they need a special `llvm-symbolizer` executable or so).

Geod24:
> Both zlib1g and libzstd are installed in the GH Actions Ubuntu 20 image.

`-dev` version as well ? Saw this: https://discourse.llvm.org/t/clang-14-warning-cannot-compress-debug-sections-zlib-not-installed-wdebug-compression-unavailable-while-using-address-sanitizer/62506

kinke:
Yep, at least `zlib1g-dev`. But this is interesting, thx for the link:
> At run-time, asan runtime invokes llvm-symbolizer to symbolize the stack trace. llvm-symbolizer is not built with zlib, so it fails to uncompress .debug_* sections, hence the output.

So seems like this has only worked on Ubuntu 18 because that image had an `llvm-symbolizer` with zlib support, but the new image doesn't. And neither does the official LLVM build we use, the thing I tried with PATH. So yeah:

> It does seem unfortunate that the official release wasn’t build against zlib.

PR #4373, opened 2023-04-15
Cirrus CI: Get FreeBSD working again after their LLVM 9 => 15 bump

kinke:


kinke:
According to https://ports.freebsd.org/cgi/ports.cgi?query=llvm&stype=name&sektion=all, they really bumped the default LLVM version from 9 to 15, *and* dropped the `llvm90` package at the same time (only v10+ available). All our existing FreeBSD packages depend on the old LLVM 9. WTF guys.

JohanEngelen:
I don't know about the gdb failures... could it be due to GDB version? Otherwise let's merge this one? (mark lit tests as XFAIL: FreeBSD)

kinke:
Looks like they bumped gdb from v8 to v13 (!), so yeah, we're probably seeing the same gdb v10+ regressions as on Linux.

Our CMake script only outputs the gdb version on Linux, not on other ELF platforms; the gdb dmd-testsuite tests are disabled for the same CMake limitation, so we're only seeing the few gdb lit-tests fail.

So if we cannot downgrade the gdb package, we'll have to disable the failing gdb tests somehow for now - ideally depending on the gdb version.

JohanEngelen:
I took the liberty of "fixing" this with the least amount of effort by marking the tests XFAIL on FreeBSD.

JohanEngelen:
Why is the FreeBSD 12.2 stuck? (some github bug?)

kinke:
> Why is the FreeBSD 12.2 stuck? (some github bug?)

Nope, it's just an orphaned required CI check in the GitHub settings, they are based on the CI job names. So after merging this, the required FreeBSD check needs to be removed, and the new 12.4 one added instead.

JohanEngelen:
Let's merge this?

PR #4382, opened 2023-04-23
Implement @ldc.attributes.noSplitStack

JohanEngelen:


JohanEngelen:
I think the problem is that throwing in __morestack is not supported, because control flow information is not set correctly. See the implementation of __morestack in libgcc: https://github.com/gcc-mirror/gcc/blob/master/libgcc/config/i386/morestack.S

I'll change the test to simply print a message and abort, without stacktrace.

JohanEngelen:
After a lot of trial and error and SSH-ing into the CI systems, I've decided to not include an actual executable run test. The functionality too system dependent.

PR #4387, opened 2023-04-29
Remove some vestigial BUILTINs in func.d?

maxhaton:
I'm flying relatively blind here but spotted these in a diff vs. the frontend tree, it compiled without them.

### DIFF
--- a/dmd/func.d
+++ b/dmd/func.d
@@ -105,33 +105,7 @@ enum BUILTIN : ubyte
     yl2xp1,
     toPrecFloat,
     toPrecDouble,
-    toPrecReal,
-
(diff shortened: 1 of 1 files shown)


kinke:
These are definitely needed as shown by CI - used for CTFEing LDC/LLVM intrinsics.

maxhaton:
Why did it build? Is it done with metaprogramming or something?

kinke:
https://github.com/ldc-developers/ldc/blob/6151310b0ea6081730619c5763718b1b85bba6b1/dmd/builtin.d#L62-L63

maxhaton:
fun

PR #4393, opened 2023-05-09
Fix #4354 - Huge compile slowdown with -g and the new LLVM pass manager

### LINKED ISSUE #4354: 10x slowdown when compiling with -g
At least on Windows, I'm experiencing an extreme slowdown for 1.32.0, as opposed to previous versions, when compiling with `-g`. For a particular library, complation took around 33 s before and 340 s with 1.32.0. Running with `-ftime-trace` shows 222 s spent in "optimizing", as opposed to 650 ms without `-g`.

kinke:
Tested manually with the Symmetry project, and LLVM 14. The `-g` compile times with both legacy and new pass managers are on par again.

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -618,6 +618,8 @@ static PipelineTuningOptions getPipelineTuningOptions(unsigned optLevelVal, unsi
  */
 //Run optimization passes using the new pass manager
 void runOptimizationPasses(llvm::Module *M) {
+  TimeTraceScope timeScope("Optimization passes");
+
   // Create a ModulePassManager to hold and optimize the collection of
   // per-module passes we are about to build.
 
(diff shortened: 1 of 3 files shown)


thewilsonator (approved):

gen/optimizer.cpp
```diff
@@ -660,7 +662,7 @@ void runOptimizationPasses(llvm::Module *M) {
   if (!noVerify) {
     pb.registerPipelineStartEPCallback([&](ModulePassManager &mpm,
                                           OptimizationLevel level) {
-      mpm.addPass(createModuleToFunctionPassAdaptor(VerifierPass()));
+      mpm.addPass(VerifierPass());
```
JohanEngelen, line 665: excellent find!

PR #4397, opened 2023-05-12
Add missing changelog entries

kinke:
As I'll probably release a v1.32.2, mainly because of #4354. **Edit**: Well, and additionally we'd have a FreeBSD package again.


JohanEngelen:
Great thanks

JohanEngelen (approved):

PR #4410, opened 2023-05-21
Port to LLVM 16.0.3

ikeycode:
This has been tested in as far as building `moss` and `boulder` against patched LDC with LLVM16.0.3 in a Serpent OS test environment. The binaries function perfectly and perhaps actually a bit smaller than they were before.

### DIFF
--- a/driver/linker.h
+++ b/driver/linker.h
@@ -42,7 +42,7 @@ bool linkAgainstSharedDefaultLibs();
 /**
  * Returns the -platformlib library names, if specified.
  */
-llvm::Optional<std::vector<std::string>> getExplicitPlatformLibs();
+std::optional<std::vector<std::string>> getExplicitPlatformLibs();
 
 /**
  * Returns the value of -mscrtlib.
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -579,7 +579,9 @@ void initializePasses() {
 #endif
   initializeVectorization(Registry);
   initializeInstCombine(Registry);
+#if LDC_LLVM_VER < 1600
   initializeAggressiveInstCombine(Registry);
+#endif
   initializeIPO(Registry);
 #if LDC_LLVM_VER < 1600
   initializeInstrumentation(Registry);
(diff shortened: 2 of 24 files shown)


ikeycode:
First PR to LDC so I'm expecting a bunch of test failures =)

ikeycode:
My only "real" concern with this PR is there is no real sane way now to `getDefaultCPU` for AArch64, and I'm not entirely convinced `generic` is a great fallback. Perhaps we could put something better there ?

driver/main.cpp
```diff
@@ -579,7 +579,9 @@ void initializePasses() {
 #endif
   initializeVectorization(Registry);
   initializeInstCombine(Registry);
+#if LDC_LLVM_VER < 1600
   initializeAggressiveInstCombine(Registry);
```
JohanEngelen, line 583: Did you check in LLVM history whether this was renamed or replaced with something else?

JohanEngelen:
> First PR to LDC so I'm expecting a bunch of test failures =)

Thanks!
You should add an LLVM16 testcase in this PR to .github/workflows/supported_llvm_versions.yml.

driver/linker.h
```diff
@@ -42,7 +42,11 @@ bool linkAgainstSharedDefaultLibs();
 /**
  * Returns the -platformlib library names, if specified.
  */
+#if LDC_LLVM_VER < 1600
 llvm::Optional<std::vector<std::string>> getExplicitPlatformLibs();
```
JohanEngelen, line 46: I thought that a bunch of these cases could be solved by replacing llvm::Optional with std::optional for all LLVM versions (I think it's remnant from when we did not use C++11?).

ikeycode:
Thanks for the feedback! Working on it now

driver/main.cpp
ikeycode: I also explained this in the commit msg with the refactored PR, but:

 - https://reviews.llvm.org/D137116

ikeycode:
Debugging some cruft with CI missing the llvm-symbolizer..

ikeycode:
Ok so according to the tests @JohanEngelen we can't just switch to std::optional :/

PR #4411, opened 2023-05-22
Enable LLVM 16.0.3

ikeycode:
Each commit in the series explains the rationale behind each decision, but nothing scary really.

The main issue area I have is getting the CI itself to bend to my will, so I could need some help here?

 - gdb is being janky for the DMD tests (complains about `.x` - I saw before a nonjanky gdb was forced?
 - Can't make it "see" FileCheck which breaks the lit tests
 
 Note: due to llvm<16 forcing C++-11 we have to use conditional compilation, otherwise we're trying to
 use C++-17 features ..

### DIFF
--- a/.github/workflows/supported_llvm_versions.yml
+++ b/.github/workflows/supported_llvm_versions.yml
@@ -15,6 +15,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - job_name: Ubuntu 20.04, LLVM 16, latest LDC beta
+            os: ubuntu-20.04
+            host_dc: ldc-beta
+            llvm_version: 16.0.3
           - job_name: Ubuntu 20.04, LLVM 15, latest LDC beta
             os: ubuntu-20.04
             host_dc: ldc-beta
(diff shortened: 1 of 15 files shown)


ikeycode:
Note, 16.0.4 came out after I started this work, but I will update Serpent OS to use latest release and test again there.

.github/workflows/supported_llvm_versions.yml
```diff
@@ -77,7 +81,12 @@ jobs:
           set -eux
           sudo apt-get update
           # Don't use latest gdb v10+ from Ubuntu toolchain PPA with regressions, use official v9
-          sudo apt-get install gdb=9.1-0ubuntu1 llvm
+          version='${{ matrix.llvm_version }}'
+          if [[ "$version" =~ ^1[6-9]\. ]]; then # LLVM 16.0.3+
+            sudo apt-get install gdb llvm llvm-14-tools
```
JohanEngelen, line 86: rather than adding llvm-14-tools, you should copy llvm16's filecheck.cpp to ldc's utils folder

ikeycode: Ok!

ikeycode:
Sorry for letting this go stale! Will get back on to it asap!

ikeycode:
Rebased. I wanted to leave the CI run so you could see the logs. I'm seeing a bunch of errors, chiefly:

`.profdata` file not found issues. Perhaps something wrong with the new `llvm-profdata`? Will investigate
gdb is broken for any dmd/lit tests using it
dcompute seems broken in terms of want LLVM itself wants?

ikeycode:
I just checked, 15 + 16 different significantly re llvm-profdata. Additionally I see a test plugin failure that I need to fix

kinke:
Thanks Ikey for starting this!

ikeycode:
Glad this is in now! Sorry I couldn't do much more on it, kinda hit my domain knowledge with this one :D

kinke:
Yeah, a bit unlucky with that problematic v16 bump - your work would have sufficed for most other LLVM bumps (excl. v15, but that's an entirely different topic). We'll see pretty soon how v17 goes...

PR #4417, opened 2023-06-05
Run C preprocessor

kinke:


driver/cl_options.cpp
```diff
@@ -400,16 +400,18 @@ cl::list<std::string>
...
-    moduleDeps("deps", cl::ValueOptional, cl::ZeroOrMore,
-               cl::value_desc("filename"),
-               cl::desc("Write module dependencies to <filename> (only imports). "
-                        "'-deps' alone prints module dependencies "
-                        "(imports/file/version/debug/lib)"));
+cl::list<std::string> ccSwitches(
+    "Xcc", cl::value_desc("ccflag"), cl::cat(linkingCategory),
+    cl::desc("Pass <ccflag> to GCC/Clang for linking/preprocessing"));
+
+cl::list<std::string> cppSwitches("Xcpp", cl::value_desc("cppflag"),
```
JohanEngelen, line 407: isn't the naming a little confusing? `-Xcc` for linking and preprocessing, `-Xcpp` only for preprocessing. Perhaps name it completely different, like `-Xcc-preprocessor` ?

JohanEngelen: (it's not related to C++, is it?)

kinke: I see these as 'eXternal C Compiler` and `eXternal C PreProcessor`. Note that dmd doesn't forward the `-Xcc` flags. I think it makes sense, e.g., for clang and `-Xcc=-target -Xcc=<myTarget>` - only need to set the cross-compile target once for both linking and preprocessing. `-Xcpp` is mainly intended for all the include paths.

The `-P` and `-Xcpp` distinction might be annoying for dub etc.

kinke:
Microsoft's `cl.exe` is a bit annoying here and doesn't allow suppressing the echo'd source file being preprocessed: https://developercommunity.visualstudio.com/t/request-an-option-to-make-clexe-more-silent-than-n/1359375, https://developercommunity.visualstudio.com/t/allow-having-clexe-not-print-the-compiled-source-f/717761

It prints it to stderr, like all warnings and errors unfortunately, so we can't simply redirect an output stream. This results in dmd-testsuite failures (output checks) and is ugly in general. DMD filters the stderr output in a non-trivial fashion (relying on other helpers in that excluded `dmd/link.d` file): https://github.com/dlang/dmd/blob/7e6cd5e6430956f8220884e98024db768e163cd0/compiler/src/dmd/link.d#L1082-L1122

driver/cl_options.cpp
JohanEngelen: Using "cpp" for C PreProcessor is quite devious :D   I think there is no reason to shorten the option, so then I vote for writing it in full, something like this: `-Xcpreprocessor`.

JohanEngelen: (or just `-P` if that is what DMD does)

kinke: I can live with adopting `-P` - it shouldn't be too long, as I expect some usages for extra include dirs and manual defines.

[`-Xcc` and `-Xcpp` are consistent with the Posix `cc`/`cpp` tools. `c++` and `CXX` (CMake) for C++...]

kinke:
TODO (`compilable/stdcheaders.c` failures):
* FreeBSD x86_64:
  ```
  /usr/include/sys/_types.h(133): Error: alias `stdcheaders.__va_list` recursive alias declaration
  ```
* Linux AArch64:
  ```
  /usr/include/aarch64-linux-gnu/bits/floatn.h(80): Error: illegal combination of type specifiers
  /usr/include/aarch64-linux-gnu/bits/floatn.h(80): Error: illegal type combination
  /tmp/clang/lib/clang/15.0.3/include/tgmath.h(34): Error: named parameter required before `...`
  ```

driver/cl_options.cpp
JohanEngelen: Ah, I had not heard of `cpp` as a program. Then maybe it is only me, and `-Xcpp` makes total sense!

kinke: I'm also unsure myself - I think `-Xcpp` would be ideal (even though we don't invoke `cpp` directly, but `cc -E` - but AFAICT, the flags are the same), but DMD already went ahead with `-P`. These differences are a PITA in dub package files, when using default `ldc2`, not `ldmd2`, requiring to duplicate the flags.

kinke: I've switched to `-P` in the latest commit.

driver/cpreprocessor.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  const auto &triple = *global.params.targetTriple;
+  const bool isMSVC = triple.isWindowsMSVCEnvironment();
+
+#if 0 //ifdef _WIN32
+  // TODO: INCLUDE env var etc.?
+  windows::MsvcEnvironmentScope msvcEnv;
+  if (isMSVC)
+    msvcEnv.setup();
+#endif
```
kinke, line 81: I have some WIP for this TODO, will open a PR after this lands [but not sure if I'm going to finish it].

driver/cpreprocessor.cpp
```diff
@@ -0,0 +1,153 @@
...
+  for (const auto &ccSwitch : opts::ccSwitches) {
+    args.push_back(ccSwitch);
+  }
+  for (const auto &cppSwitch : opts::cppSwitches) {
+    args.push_back(cppSwitch);
+  }
+
+  if (isMSVC) {
+    args.push_back("/nologo");
+    args.push_back("/P"); // preprocess only
```
JohanEngelen, line 103: perhaps for followup, but it'd be nice to put all these hardcoded flags in the ldc2.conf file such that it can be tweaked for a certain platform (or e.g. for cross compilation) without modifying the compiler.

kinke: I'm pretty sure this already covers most use cases. I've tested cross-compiling `compilable/stdcheaders.c` on Linux x86 to Linux aarch64 (to troubleshoot the CI failure), using both the cross-gcc and a generic clang, all working fine.

Cross-compiling C to Windows on a non-Windows guest is probably not easy - would need the Microsoft headers, and there might be no `clang-cl` wrapper available by default on non-Windows systems.

PR #4419, opened 2023-06-09
Set up MSVC environment for C preprocessing on Windows

kinke:


kinke:
I've tested this with VS 2017 and 2019, seems to work fine. My Windows VM only has WinSDK v10, no older ones, so I couldn't test those.

runtime/druntime/src/__builtins.di
```diff
@@ -145,19 +143,18 @@ else version (LDC)
     alias __builtin_huge_valf = __builtin_inff;
     alias __builtin_huge_vall = __builtin_infl;
 
-    alias __builtin_fabs  = llvm_fabs!double;
-    alias __builtin_fabsf = llvm_fabs!float;
-    alias __builtin_fabsl = llvm_fabs!real;
+    alias __builtin_fabs  = imported!"ldc.intrinsics".llvm_fabs!double;
```
JohanEngelen, line 146: why is this extra complication needed?

kinke: These are lazy imports now - the `ldc.intrinsics` modules etc. aren't imported if the aliases aren't ref'd from the module(s) being compiled. The DMD parts use them too, here and in object.d.

PR #4423, opened 2023-06-11
Add support for LLVM 16 and bump LDC-LLVM to v16.0.6 for prebuilt packages

kinke:
Based on #4411.


kinke:
@thewilsonator: The GitHub Actions `Main` workflow is currently using an LDC-LLVM v16.0.6 candidate (from https://github.com/ldc-developers/llvm-project/releases/tag/CI), one with SPIRV included as experimental target. 3 dcompute lit-tests fail, hitting an assertion wrt. `t->toBasetype()->ty == TY::Ttuple || stripAddrSpaces(v->getType()) == DtoPtrToType(t)`, see e.g. https://github.com/ldc-developers/ldc/actions/runs/5485337264/jobs/9994033947.

thewilsonator:
Thanks, Will take a look over the weekend

thewilsonator:
> @thewilsonator: The GitHub Actions `Main` workflow is currently using an LDC-LLVM v16.0.6 candidate (from https://github.com/ldc-developers/llvm-project/releases/tag/CI), one with SPIRV included as experimental target. 3 dcompute lit-tests fail, hitting an assertion wrt. `t->toBasetype()->ty == TY::Ttuple || stripAddrSpaces(v->getType()) == DtoPtrToType(t)`, see e.g. https://github.com/ldc-developers/ldc/actions/runs/5485337264/jobs/9994033947.

Looking at it, that seems to be harmless. For LLVM 16 (where pointers are always opaque) the second part of the `||` should probably be replaced with `isaPointer(v->getType())` anyway since `DtoPtrToType(t) == getVoidPtrType()` always. If it _still_ trips  after that change then we have a problem somewhere. Will do some actual testing tomorrow.

kinke:
We still default to non-opaque pointers with LLVM 16, as we aren't ready yet: https://github.com/ldc-developers/ldc/issues/4261

thewilsonator:
Oh, in that case the cause of that assertion is https://github.com/ldc-developers/ldc/blob/d4f2bed3a45688b32434d429e39c22dea5a9ffda/gen/tollvm.cpp#L589-L590 it should do the LLVM14-15 path and check for `isOpaque`

kinke:
The LLVM 16 bump is a PITA because of:
1. Vanilla LLVM CI job:
  * macOS: Their official releases on GitHub don't have any x86_64 archive anymore, only arm64.
  * Linux x64: They now use primarily Ubuntu 22, with 16.0.0 being the only available Ubuntu-18 archive. I don't want to bump the CI job to Ubuntu 22 (from 20), because its gdb came with regressions for global variables, leading to at least 4 test failures (lit testsuite + DMD testsuite).
2. The apparent miscompile of the frontend itself on Windows and Mac mentioned above in https://github.com/ldc-developers/ldc/pull/4423#issuecomment-1625598506. In the [release log](https://releases.llvm.org/16.0.0/docs/ReleaseNotes.html#changes-to-interprocedural-optimizations), they mention a miscompile with an LTO+PGO combo caused by newly enabled-by-default function specializations; I'm wondering if we're hitting something similar.

kinke:
Wrt. the miscompile when compiling the frontend with LLVM 16, I've tried:
1. Reverting to clang v15 (from v16) for compiling both LLVM and the LDC C++ parts; didn't change anything.
2. Additionally disabling PGO (for the compiler) alone; didn't change anything.
3. Additionally disabling LTO (for the compiler):
   * Makes the Win64 job finally green (Win32 was already green before).
   * Causes a `libc++abi: Pure virtual function called!` error on Mac x64 when trying to compile druntime/Phobos with that fresh LDC compiled with itself!

JohanEngelen:
could the miscompile be because of typed pointers no longer being really supported by LLVM?

kinke:
Your guess is probably as good as mine, but I don't think so.

kinke:
Disabling the new function specializations has fixed the miscompile on Win64 at least; Mac still needs troubleshooting.

kinke:
macOS x64 is 'solved' too now (with LTO+PGO), after disabling the func specializations for the LTO plugin too. 2 important questions remain:
1. What about *arm64* macOS? Linux aarch64 is good. @JohanEngelen: Could you please test the main GHA CI artifact (once available) and check whether you get any crashes or garbage outputs for `fail_compilation/dtor_attributes.d`? (maybe run a few times to really be sure)
2. Looks like the func specializations are highly problematic for us, at least for some platforms. As our generated IR for Linux and macOS x64 don't differ much, I'm a bit skeptical about the problem lying in our court (such as still using typed IR pointers etc.). So I think we should disable it by default in `ldc2.conf` when using LLVM 16 for now (CMake `ADDITIONAL_DEFAULT_LDC_SWITCHES`). I guess for all targets to be on the safe side?

JohanEngelen:
I will test soon. What is the flag to enable/disable the func specialization?

kinke:
> I will test soon.

Great! Nothing to build, just downlaod & extract CI artifact, then try compiling the failing module. The cross-compiled binary is either miscompiled (will crash or produce garbage) or works.

> What is the flag to enable/disable the func specialization?

I'm using `-func-specialization-size-threshold=1000000000`, so that all functions with less than 1e9 instructions aren't candidates. ;) - It defaults to 100 in LLVM 16.

JohanEngelen:
the macos arm64 artifact produces a little garbage indeed...

```
❯ bin/ldc2 ~/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(118): Error: `pure` function `dtor_attributes.test1` cannot call impure destructor `dtor_attributes.Strict.~this`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(113):        generated `Strict.~this` is destructor because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(111):         - HasDtor member
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(103):           member `HasDtor.~this` is declared here
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(118): Error: `@safe` function `dtor_attributes.test1` cannot call `@system` destructor `dtor_attributes.Strict.~this`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(113):        `dtor_attributes.Strict.~this` is declared here
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(113):        generated `Strict.~this` is function because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(111):         - HasDtor member
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(103):           member `HasDtor.~this` is declared here
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(118): Error: `@nogc` function `dtor_attributes.test1` cannot call non-@nogc destructor `dtor_attributes.Strict.~this`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(113):        generated `Strict.~this` is non-@nogc because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(111):         - HasDtor member
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(103):           non-@nogc `HasDtor.~this` is declared here
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(118): Error: destructor `dtor_attributes.Strict.~this` is not `nothrow`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(113):        generated `Strict.~this` is �_��_����_��{��(`@9 because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(111):         - HasDtor member
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(103):           member `HasDtor.~this` is declared here
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(116): Error: function `dtor_attributes.test1` may throw but is marked as `nothrow`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(209): Error: `pure` function `dtor_attributes.test2` cannot call impure destructor `dtor_attributes.StrictClass.~this`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(204):        generated `StrictClass.~this` is destructor because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(203):         - HasDtor member
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(103):           member `HasDtor.~this` is declared here
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(321): Error: `pure` function `dtor_attributes.test3` cannot call impure destructor `dtor_attributes.StrictStructRef.~this`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(316):        generated `StrictStructRef.~this` is destructor because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(310):         - HasDtor structMember
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(103):           structMember `HasDtor.~this` is declared here
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(411): Error: `pure` function `dtor_attributes.test4` cannot call impure destructor `dtor_attributes.StrictNested.~this`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(406):        generated `StrictNested.~this` is destructor because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(403):         - HasDtor[4] arrayMember
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(103):           arrayMember `HasDtor.~this` is declared here
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(509): Error: `pure` function `dtor_attributes.test5` cannot call impure destructor `dtor_attributes.Permissive.~this`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(618): Error: `pure` function `dtor_attributes.test6` cannot call impure destructor `dtor_attributes.HasNestedDtor3.~this`
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(611):        generated `HasNestedDtor3.~this` is destructor because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(613):         - HasNestedDtor2 member3
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(606):        generated `HasNestedDtor2.~this` is `�� because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(608):         - HasNestedDtor1 member2
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(601):        generated `HasNestedDtor1.~this` is `�� because of the following field's destructors:
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(603):         - HasDtor member1
/Users/johan/ldc/ldc/tests/dmd/fail_compilation/dtor_attributes.d(103):           member1 `HasDtor.~this` is declared here
```

kinke:
Alright, thx a lot for testing.

kinke:
[The CircleCI Mac jobs 'fail' because we've run out of credits.]

Issue #4427, opened 2023-06-22
Destructor is not called for AliasSeq (type tuple)


JohanEngelen:
Testcase:
```d
import core.stdc.stdio: printf;
alias AliasSeq(TList...) = TList;

int i = 0;
struct A {
    ~this() {
        i++;
    }
}

void main() {
    {
	    AliasSeq!(A, A) params;
    }
    printf("%d\n", i); // prints 0 !
    assert(i == 2); // fails!
}
```
https://godbolt.org/z/TGMWczh3E

-vcg-ast output shows the problem that the frontend does not add destructors calls like it usually does:
```d
void main()
{
	{
		(A, A) params;
               //  there should be two explicit destructors calls here (~this())
	}
	printf("%d\n", i);
	assert(i == 2);
	return 0;
}
```

kinke:
This was fixed in the frontend, right?

JohanEngelen:
Yep, fixed in LDC 1.35.0

PR #4429, opened 2023-06-22
Call dtor for all VarDeclarations with needsScopeDtor()==true.

JohanEngelen:
Fixes issue #4427

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -876,6 +876,15 @@ void DtoResolveVariable(VarDeclaration *vd) {
  * DECLARATION EXP HELPER
  ******************************************************************************/
 
+void pushVarDtorCleanup(IRState *p, VarDeclaration *vd) {
+  llvm::BasicBlock *beginBB = p->insertBB(llvm::Twine("dtor.") + vd->toChars());
+
+  const auto savedInsertPoint = p->saveInsertPoint();
+  p->ir->SetInsertPoint(beginBB);
+  toElemDtor(vd->edtor);
+  p->funcGen().scopes.pushCleanup(beginBB, p->scopebb());
+}
+
 // TODO: Merge with DtoRawVarDeclaration!
 void DtoVarDeclaration(VarDeclaration *vd) {
   assert(!vd->isDataseg() &&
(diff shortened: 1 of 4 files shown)


JohanEngelen:
This PR brings us on-par with DMD and GDC, but there is more bad news: the simple test case works with DMD and GDC, but slightly more complex and it fails for all https://issues.dlang.org/show_bug.cgi?id=24010.  The destructors are called at the end of their definition, not at the end of their scope.

gen/llvmhelpers.cpp
```diff
@@ -954,6 +963,11 @@ void DtoVarDeclaration(VarDeclaration *vd) {
       toElem(ex->exp);
     }
   }
+
+  if (!vd->isDataseg() && vd->needsScopeDtor()) {
```
kinke, line 967: (There's an assertion at the top wrt. `!vd->isDataseg()`)

gen/llvmhelpers.cpp
```diff
@@ -876,6 +876,15 @@ void DtoResolveVariable(VarDeclaration *vd) {
  * DECLARATION EXP HELPER
  ******************************************************************************/
 
+void pushVarDtorCleanup(IRState *p, VarDeclaration *vd) {
```
kinke, line 879: (This might be nicer as `IRState` method)

kinke (approved):

kinke:
@JohanEngelen: Are you unsure about this (and including it in v1.33)? Just wondering why it isn't merged yet (I was on vacation...).

JohanEngelen:
I was slightly unsure about the implementation, but also that the bug would introduce another bug.
This is now properly fixed upstream: https://github.com/dlang/dmd/pull/15351 
Tuples should now no longer reach this code (if I understand the fix correctly).

JohanEngelen:
@kinke This can be a very disruptive fix. So let's postpone until next release, when also DMD frontend has the additional fix https://github.com/dlang/dmd/pull/15351 . That way, we don't change user code behavior into a new buggy behavior, but instead wait until we have the full proper fix.

JohanEngelen:
I believe this is fixed in 1.35.0 due to upstream fix.

PR #4431, opened 2023-07-06
Fix LDC compilation with LDC 1.37 universal on macOS. Make ExtractDMDSystemLinker.cmake compatible with `-target` and `-arch`

JohanEngelen:
I need this fix when the system D compiler has `-target` in its default linker line

### DIFF
--- a/cmake/Modules/ExtractDMDSystemLinker.cmake
+++ b/cmake/Modules/ExtractDMDSystemLinker.cmake
@@ -64,6 +64,11 @@ separate_arguments(linker_line)
 list(GET linker_line 0 D_LINKER_COMMAND)
 list(REMOVE_AT linker_line 0)
 
+# Fixup known flags with spaces, such as "-target triple" argument, which would be turned
+# into "-target;triple" by `separate_arguments`. Repair by merging into one list item.
+string(REGEX REPLACE ";-target;([A-Za-z0-9_-]+);" ";-target \\1;" linker_line "${linker_line}")
+string(REGEX REPLACE ";-arch;([A-Za-z0-9_-]+);" ";-arch \\1;" linker_line "${linker_line}")
+
 if("${D_COMPILER_ID}" STREQUAL "GDMD")
     # Filter linker arguments for those we know can be safely reused
     set(D_LINKER_ARGS)
--- a/cmake/Modules/ExtractDMDSystemLinker.cmake
+++ b/cmake/Modules/ExtractDMDSystemLinker.cmake
@@ -64,6 +64,9 @@ separate_arguments(linker_line)
 list(GET linker_line 0 D_LINKER_COMMAND)
 list(REMOVE_AT linker_line 0)
 
+# Fixup "-target triple" argument, which would be turned into "-target;triple" by `separate_arguments`. Replace ";" with "=".
+string(REGEX REPLACE ";-target;" ";--target=" linker_line "${linker_line}")
+
 if("${D_COMPILER_ID}" STREQUAL "GDMD")
     # Filter linker arguments for those we know can be safely reused
     set(D_LINKER_ARGS)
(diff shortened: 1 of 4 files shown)


cmake/Modules/ExtractDMDSystemLinker.cmake
```diff
@@ -64,6 +64,9 @@ separate_arguments(linker_line)
 list(GET linker_line 0 D_LINKER_COMMAND)
 list(REMOVE_AT linker_line 0)
 
+# Fixup "-target triple" argument, which would be turned into "-target;triple" by `separate_arguments`. Replace ";" with "=".
+string(REGEX REPLACE ";-target;" ";--target=" linker_line "${linker_line}")
```
kinke, line 68: `linker_line` and `D_LINKER_ARGS` later are CMake lists, so semicolons separating individual args should be fine. IIRC, Apple clang is picky and requires `-target <bla>`, not accepting `--target=bla` - might have been an older version though.

JohanEngelen: The problem is that the D_LINKER_ARGS is passed to CMake, which does something smart: if the argument does not start with `-`, it assumes it is a library and will add `-l` in front of it. Thus `-target some-triple` is converted to `-target -lsome-triple`.

kinke: OMG - yeah I hit that too sometime. That's because we use `target_link_libraries()` for these (with `LDC_LINK_MANUALLY=ON`). There's https://cmake.org/cmake/help/latest/command/target_link_options.html since CMake v3.13, which would presumably fix this and similar ugliness.

JohanEngelen:
I just hit this when trying to build LDC locally with LDC 1.36.0 universal, because our ldc2.conf contains:
```
        "-Xcc=-target",
        "-Xcc=arm64-apple-macos",
```

cmake/Modules/ExtractDMDSystemLinker.cmake
JohanEngelen: thnx for the target_link_options suggestion.
I've been conservative in applying it (not changing it for other linker args that we pass), because some of the linker "flags" in our script are actually linker dependencies (i.e. libraries) such as LDC_LIB==LDCShared.

JohanEngelen:
reverted to the old solution if CI likes that better; perhaps somehow the flags are not passed correctly for some of the CI cases with more complex flag setups

JohanEngelen:
The difference between `target_link_options` and `target_link_libraries` is that the first adds flags at start of commandline, the latter at the end.
There are some other small deltas:

`target_link_options` 
`[108/1132] : && /usr/bin/c++ -DDMDV2 -O3 -DNDEBUG -m64 -Xlinker --export-dynamic -L/opt/hostedtoolcache/dc/dmd-2.107.1-rc.1/x64/dmd2/linux/bin64/../lib64 -Bstatic -lphobos2 -Bdynamic -lpthread -lm -lrt -ldl   obj/ldc2.o -o bin/ldc2  lib/libldc.a  <LLVM libs>   -lz  -lrt  -ldl  -lpthread  -lm  -Wl,--export-dynamic  -ldl && :`

`target_link_libraries`
`[108/1132] : && /usr/bin/c++ -DDMDV2 -O3 -DNDEBUG                                                                                                                                                                obj/ldc2.o -o bin/ldc2  lib/libldc.a  <LLVM libs>  -lz  -lrt  -ldl  -lpthread  -lm  -Wl,--export-dynamic  -m64  -Xlinker  --export-dynamic  -L/opt/hostedtoolcache/dc/dmd-2.107.1-rc.1/x64/dmd2/linux/bin64/../lib64  -Xlinker  -Bstatic  -lphobos2  -Xlinker  -Bdynamic  -lpthread  -lm  -lrt  -ldl  -ldl  -lpthread  -lm && :`

JohanEngelen:
I think I have it working well now. Will clean up after CI confirms.

kinke:
Thx, just in time for v1.37. :)

Issue #4435, opened 2023-07-10
Emit IBT/BTI instructions


ibara:
Hi LDC developers --

OpenBSD moved to an IBT/BTI world. This causes binaries created with LDC to fail, as LDC does not emit the proper instructions for IBT/BTI.

This should be fixable with the addition of a flag that enables these instructions, as LLVM supports IBT/BTI very well.

kinke:
Hi. Is this clang's `-fcf-protection={return,branch,full}`? That would set module flags according to https://github.com/ldc-developers/llvm-project/blob/c12d3509ebf75bb6a16798d5fed91acca86f7899/clang/lib/CodeGen/CodeGenModule.cpp#L771-L783

Is this supposed to be the default setting starting with what OpenBSD version? And OpenBSD only, not for FreeBSD etc.?

ibara:
Hi. Yes. This is -fcf-protection=branch only. This is the default on OpenBSD -current today, which will eventually become OpenBSD 7.4.

OpenBSD only. The BSDs aren't like Linux. One should think of OpenBSD and FreeBSD as differently as you would think of Linux and FreeBSD.

JohanEngelen:
Note:  this should also set a `__CET__` enum value. Or perhaps we should define `__CET_x__` versions  with `x` = 1,2,3.
```
// RUN: %clang -target i386-unknown-unknown -x c -E -dM -o - -fcf-protection=return %s | FileCheck %s --check-prefix=RETURN
// RUN: %clang -target i386-unknown-unknown -x c -E -dM -o - -fcf-protection=branch %s | FileCheck %s --check-prefix=BRANCH
// RUN: %clang -target i386-unknown-unknown -x c -E -dM -o - -fcf-protection=full %s   | FileCheck %s --check-prefix=FULL

// RETURN: #define __CET__ 2
// BRANCH: #define __CET__ 1
// FULL: #define __CET__ 3
void foo() {}
```

JohanEngelen:
https://github.com/llvm/llvm-project/blob/fbe4ff8149a2b656a66cc0c64a38a5302c051c6e/clang/lib/CodeGen/CodeGenModule.cpp#L1015-L1027

JohanEngelen:
> This should be fixable with the addition of a flag that enables these instructions, as LLVM supports IBT/BTI very well.

[...]

PR #4437, opened 2023-07-11
Add -fcf-protection.

JohanEngelen:
Resolves issue #4435

~~I do not know how to cleanly add compiler-defined enum value (e.g. `__CET__ == 1`), so I opted for adding compiler-defined versions instead.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - C files are now automatically preprocessed using the external C compiler (configurable via `-gcc` or the `CC` environment variable, and `-Xcc` for extra flags). Extra preprocessor flags (e.g., include dirs and manual defines) can be added via new command-line option `-P`. (#4417)
   - Windows: If `clang-cl.exe` is on `PATH`, it is preferred over Microsoft's `cl.exe` by default (e.g., to avoid printing the C source file name to stderr during preprocessing).
 - Less pedantic checks for conflicting C(++) function declarations when compiling multiple modules to a single object file ('Error: Function type does not match previously declared function with the same mangled name'). The error now only appears if an object file actually references multiple conflicting functions. (#4420)
+- New command-line option `--fcf-protection`. (#4437)
 
 #### Platform support
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - C files are now automatically preprocessed using the external C compiler (configurable via `-gcc` or the `CC` environment variable, and `-Xcc` for extra flags). Extra preprocessor flags (e.g., include dirs and manual defines) can be added via new command-line option `-P`. (#4417)
   - Windows: If `clang-cl.exe` is on `PATH`, it is preferred over Microsoft's `cl.exe` by default (e.g., to avoid printing the C source file name to stderr during preprocessing).
 - Less pedantic checks for conflicting C(++) function declarations when compiling multiple modules to a single object file ('Error: Function type does not match previously declared function with the same mangled name'). The error now only appears if an object file actually references multiple conflicting functions. (#4420)
-- New command-line option `--fcf-protection`. (#4437)
+- New command-line option `--fcf-protection`, which enables Intel's Control-Flow Enforcement Technology (CET). (#4437)
 
 #### Platform support
 
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -1003,6 +1003,20 @@ void registerPredefinedVersions() {
     VersionCondition::addPredefinedGlobalIdent("LDC_ThreadSanitizer");
   }
 
+  switch (opts::fCFProtection) {
+  case opts::CFProtectionType::Branch:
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - C files are now automatically preprocessed using the external C compiler (configurable via `-gcc` or the `CC` environment variable, and `-Xcc` for extra flags). Extra preprocessor flags (e.g., include dirs and manual defines) can be added via new command-line option `-P`. (#4417)
   - Windows: If `clang-cl.exe` is on `PATH`, it is preferred over Microsoft's `cl.exe` by default (e.g., to avoid printing the C source file name to stderr during preprocessing).
 - Less pedantic checks for conflicting C(++) function declarations when compiling multiple modules to a single object file ('Error: Function type does not match previously declared function with the same mangled name'). The error now only appears if an object file actually references multiple conflicting functions. (#4420)
+- New command-line option `--fcf-protection`. (#4437)
```
thewilsonator, line 9: maybe add a brief explanation of what is does?

driver/main.cpp
```diff
@@ -1003,6 +1003,20 @@ void registerPredefinedVersions() {
     VersionCondition::addPredefinedGlobalIdent("LDC_ThreadSanitizer");
   }
 
+  switch (opts::fCFProtection) {
+  case opts::CFProtectionType::Branch:
+    VersionCondition::addPredefinedGlobalIdent("__CET_1__");
```
thewilsonator, line 1008: if you wanted to add a enum instead of a bunch of version there is `__traits(getInfo, "string_key")`

kinke: `__traits(getTargetInfo, "key")`, might be a good idea. clang's `__CET__` can be used as bitmask, whereas checking for branch-protection with the current `version`s would probably be something ugly like
```D
version (__CET_1__) version = BranchCFProtection;
version (__CET_3__) version = BranchCFProtection;
version (BranchCFProtection) …
```

thewilsonator: `getTargetInfo`, that was what I meant to say. Also it is not immediately obvious what `__CET_n__` refer to. I agree that adding `VersionCondition::addPredefinedGlobalIdent("LDC_BranchCFProtection");` is much better. 
(as an aside, does GCC/GDC have something similar? if so we should try to have a common interface `"D_BranchControlFlowProtection"`? @ibuclaw).

ErnyTech: > as an aside, does GCC/GDC have something similar?

GDC supports CET but I don't think it sets any predefined version

JohanEngelen: Thanks for the help here :)
`__traits(getTargetInfo, "__CET__")` for similarity with C ?  Or without the underscores?
No version needed then, `static if (__traits(getTargetInfo, "key") > 0)` would suffice. I'm thinking that `__CET__` is very rarely needed; the only usecase I can think of is writing assembly (I see in cet.h the bitmasking usecase the Martin mentioned).

ibuclaw: > > as an aside, does GCC/GDC have something similar?
> 
> GDC supports CET but I don't think it sets any predefined version

Correct. The druntime library is compiled with `-fversion=CET` because I've not ported fibers to work with both branch protection or shadow stack.

I gather otherwise the end user shouldn't really care about it.

JohanEngelen:
It turns out that `@naked` functions (with assembly implementation, such as `fiber_switchContext`), actually do get `endbr` added at the start of the function when CET is enabled. So they are not entirely naked, and we don't need to modify those druntime functions :)   So perhaps things just work already without further modification needed.

@ibara Can you try building LDC's runtime with this patch and test the whole LDC test suite on OpenBSD (with CET requirement)?

JohanEngelen:
Should we error when this flag is passed for non-X86 targets?

ErnyTech:
> Should we error when this flag is passed for non-X86 targets?

Yes, Clang currently shows `error: option 'cf-protection=branch' cannot be specified on this target`
Without an error the user might think that the option is valid and the protection applied

ibara:
> It turns out that `@naked` functions (with assembly implementation, such as `fiber_switchContext`), actually do get `endbr` added at the start of the function when CET is enabled. So they are not entirely naked, and we don't need to modify those druntime functions :) So perhaps things just work already without further modification needed.
> 
> @ibara Can you try building LDC's runtime with this patch and test the whole LDC test suite on OpenBSD (with CET requirement)?

Will test this weekend. Thanks!

kinke:
[LLVM 9 needs a fix: `/home/runner/work/ldc/ldc/driver/cl_options_instrumentation.cpp:127:58: error: ‘const class llvm::Triple’ has no member named ‘isX86’`]

JohanEngelen:
all green now

ibara:
Didn't forget about this. Will get you test data when I am home a little later today.

kinke (approved):

PR #4443, opened 2023-07-17
Add pragma(LDC_no_moduleinfo) to all gccbuiltin generated .di files.

### LINKED ISSUE #4442: LDC is trying to reference ModuleInfo from `ldc.gccbuiltins` and friends
CC @JohanEngelen 

`ldc.gccbuiltins_` and other intrinsics modules are effectively a bunch of forward declarations with `pragma(LDC_intrinsic)` which doesn't require ModuleInfo.

I can't get a small reproduction of this, but some module in our codebase, if get imported, generates references to it, which I believe is the compiler trying to ensure some static ctor order, based on its imports.

```
ld.lld: error: undefined symbol: ModuleInfo for ldc.gccbuiltins_x86
>>> referenced by package.d
>>>               ../weka-lib_Dsit.o:(ModuleInfo for weka.lib.bitmanip_ext)

ld.lld: error: undefined symbol: ModuleInfo for ldc.llvmasm
>>> referenced by package.d
>>>               ../weka-lib_Dsit.o:(ModuleInfo for weka.lib.compiler.asm_utils)
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
```

For such modules, `pragma(LDC_no_module_info)` should be used and the compiler should prevent referencing modules that have that `pragma`.

JohanEngelen:
Moduleinfo is not needed, because there are no module ctors, nor unittests, etc. This prevents missing Moduleinfo symbol linking problems for some use cases.

Resolves #4442

### DIFF
--- a/runtime/druntime/src/ldc/llvmasm.di
+++ b/runtime/druntime/src/ldc/llvmasm.di
@@ -1,5 +1,7 @@
 module ldc.llvmasm;
 
+pragma(LDC_no_moduleinfo);
+
 struct __asmtuple_t(T...)
 {
     T v;
--- a/runtime/druntime/src/ldc/profile.di
+++ b/runtime/druntime/src/ldc/profile.di
@@ -16,6 +16,8 @@
  */
 module ldc.profile;
 
+pragma(LDC_no_moduleinfo); // this module is not standard linked with druntime, don't burden the user with requiring linking.
+
 version = HASHED_FUNC_NAMES;
 
 import ldc.intrinsics : LLVM_version;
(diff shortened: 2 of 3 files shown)


ljmf00:
Do we do the check the other way around? I think we only check for `LDC_no_moduleinfo` on creation but not on referencing. I read the code a bit and haven't seen a testcase that proves it.

JohanEngelen:
> I think this can be extended to any `.di` intrinsics.

I'm a bit wary of just blanket removing moduleinfo without needing to do so. But I've removed it from ldc.profile too (that module is intentionally template-only and not linking into druntime).

JohanEngelen:
converted to draft, this actually does not fix issue #4442

JohanEngelen:
why draft, it's simple, this does not fix #4442

PR #4455, opened 2023-07-28
Drop support for LLVM 9 and LLVM 10

kinke:
LLVM 11 is the last version supporting the dynamic-compile functionality, so I thought it might be nice to still kind-of support it.


.github/workflows/supported_llvm_versions.yml
```diff
@@ -19,39 +19,30 @@ jobs:
             os: ubuntu-20.04
             host_dc: ldc-beta
             llvm_version: 15.0.6
-          - job_name: Ubuntu 20.04, LLVM 14, latest LDC beta
```
JohanEngelen, line 22: was this removal intentional? Where is LLVM14 tested on Ubuntu?

kinke: Yes, intentional, see last commit title. It's not like we ever tested all LLVM versions on all platforms. 2 macOS jobs here are standard (mainly to test the `BUILD_SHARED_LIBS=ON|OFF` permutations).

PR #4457, opened 2023-07-29
Cirrus CI: Try switching AArch64 job from container to VM

kinke:


kinke:
Great, this works again, doesn't suffer from the Docker-container issue. Performance seems to be the same; the VM spin-up time not that bad. Based on https://cirrus-ci.org/guide/custom-vms/#custom-compute-engine-vms; quite powerful.

PR #4458, opened 2023-07-29
CI: Switch to custom ninja, trying to schedule `ldc2.o` compilation early

kinke:
See https://github.com/symmetryinvestments/ninja/releases/tag/v1.11.1-sym1.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,6 +79,7 @@ commonSteps: &commonSteps
             -DCMAKE_BUILD_TYPE=Release \
             -DD_COMPILER=$PWD/../host-ldc/bin/ldmd2 \
             -DLDC_LINK_MANUALLY=OFF \
+            -DCMAKE_ADD_CUSTOM_COMMAND_DEPENDS_EXPLICIT_ONLY=ON \
             $EXTRA_CMAKE_FLAGS
           ninja -j$PARALLELISM obj/ldc2.o all ldc2-unittest all-test-runners
           bin/ldc2 -version
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,19 +18,24 @@ commonSteps: &commonSteps
             fi
             apt-get -q update
             apt-get -yq install \
-              git-core cmake $gcc_pkg \
+              git-core $gcc_pkg \
               zlib1g-dev $libcurl_pkg curl gdb python3 python3-pip tzdata unzip zip \
               $EXTRA_APT_PACKAGES
+            # Download & extract CMake
+            curl -fL --retry 3 --max-time 300 -o cmake.tar.gz https://github.com/Kitware/CMake/releases/download/v3.27.1/cmake-3.27.1-linux-x86_64.tar.gz
+            mkdir cmake
+            tar -xf cmake.tar.gz --strip 1 -C cmake
+            rm cmake.tar.gz
             # Download & extract Ninja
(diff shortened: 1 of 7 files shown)


kinke:
E.g., with that custom ninja on my Ubuntu 22 box, building a `RelWithDebInfo` `ldc2` (with LDC v1.30 as host compiler, and `ninja -j24 obj/ldc2.o ldc2`) takes 54 seconds overall. 52.7 secs are required for the `obj/ldc2.o` all-at-once frontend compilation; compiling the 102 .cpp files in parallel takes 22.7 secs overall, and linking the executable 1.5 secs (lld). This results in an overall meager effective parallelism of 9.4x, all caused by the `obj/ldc2.o` monolith.

Scheduling `obj/ldc2.o` first at least makes sure there is no further delay. E.g., when using distro ninja v1.10.1, the overall build time on my box is 68 seconds (=> 7.5x effective parallelism).

JohanEngelen:
When testing a bit, I've noticed that the build of `obj/ldc2.o` somehow has a prerequisite on `libldc.a`. In other words: executing `ninja obj/ldc2.obj` will rebuild `libldc.a` when a `.cpp` file has been modified. The dependency is completely unnecessary, and may explain some parallelism / sequencing of what ninja is currently doing.

kinke:
Hmm, that's not the case for me; in a fresh new CMake build dir:
```bash
$ ninja -j1 obj/ldc2.o
[0/1] Generating obj/ldc2.o
```

JohanEngelen:
In a clean folder on my machine, `ninja obj/ldc2.o` is building all .cpp files.
Details about my build setup:
CMake 3.25.2
ninja 1.11.1
macOS
`cmake -G Ninja -DLDC_INSTALL_LLVM_RUNTIME_LIBS=ON -DRT_SUPPORT_SANITIZERS=ON -DLDC_DYNAMIC_COMPILE=OFF -DCMAKE_BUILD_TYPE="Release" -DCMAKE_INSTALL_PREFIX="~/ldc/install15" -DLLVM_ROOT_DIR="~/llvm/install15" ../ldc`

kinke:
Mhm, confirmed with those CMake flags on my box too (CMake v3.22.1, Ubuntu 22).

JohanEngelen:
Maybe it's because of this in build.ninja:
```
#############################################
# Order-only phony target for ldc2

build cmake_object_order_depends_target_ldc2: phony || cmake_object_order_depends_target_LDCShared obj/ldc2.o
```
which I think means that `cmake_object_order_depends_target_LDCShared` must be build before `obj/ldc2.o`.

kinke:
`-DLDC_LINK_MANUALLY=OFF` makes it work, i.e., not depend on `libldc.a`.

JohanEngelen:
Wow, check out `ninja -t browse`. It opens webbrowser so you can browse the dependency graph, pretty cool.  It indeed lists `lib/libldc.a (order-only)`  for `obj/ldc2.o`.

kinke:
I've played around, and eventually found brand-new `-DCMAKE_ADD_CUSTOM_COMMAND_DEPENDS_EXPLICIT_ONLY=ON`, supported since CMake v3.27 ([docs](https://cmake.org/cmake/help/latest/variable/CMAKE_ADD_CUSTOM_COMMAND_DEPENDS_EXPLICIT_ONLY.html)). That prevents the `ldc2` dependency on `libldc.a` from apparently implicitly trickling down to the custom command for `obj/ldc2.o`. Confirmed locally.

PR #4460, opened 2023-07-30
Do not prefix static libs from pragma(lib) with  `-l`

### LINKED ISSUE #4459: pragma(lib) doesn't work with static libraries
Let's take this example

- app.d

```d
import std.stdio;

pragma(lib, "lib.a");

extern(C) int lib_get_int(); 

void main() {
    writeln(lib_get_int());
}
```

- lib.d

```d
module lib;

export extern(C) int lib_get_int() {
    return 42;
}

```

1. build the static lib: ``ldmd2  -betterC -lib lib.d``
2. build the program: ``ldmd2  -betterC -run app.d``

# Expected:

It should build, link, run and print: ``42``

# What happen instead?:

It doesn't compile:

```
$ ldmd2 -run app.d
/usr/bin/ld: cannot find -llib.a: No such file or directory
```

With DMD it all works as expected

System:

OS: Linux x64
LDC: 

```
LDC - the LLVM D compiler (1.32.2):
  based on DMD v2.102.2 and LLVM 15.0.7
  built with LDC - the LLVM D compiler (1.32.1)
  Default target: x86_64-pc-linux-gnu
  Host CPU: skylake
  http://dlang.org - http://wiki.dlang.org/LDC
```

I suspect the reason is it puts the ``-l`` on the ``.a`` file, it should just pass the path without the ``-l``

ryuukk:
It should fix #4459 

I have issues building ldc on my PC, so i don't know how to test this change other than submit a PR

### DIFF
--- a/gen/declarations.cpp
+++ b/gen/declarations.cpp
@@ -419,17 +419,25 @@ public:
         std::string arg = ("/DEFAULTLIB:\"" + name + "\"").str();
         gIR->addLinkerOption(llvm::StringRef(arg));
       } else {
-        if (name.endswith(".a")) {
-          global.params.linkswitches.push(name);
-        } else {
-          size_t const n = name.size() + 3;
-          char *arg = static_cast<char *>(mem.xmalloc(n));
+        bool isStaticLib = name.endswith(".a");
+
+        size_t const nameLen = name.size();
+        size_t const n = nameLen + 3;
+        char *arg = nullptr;
+
+        if (isStaticLib == false) {
+          arg = static_cast<char *>(mem.xmalloc(n));
           arg[0] = '-';
           arg[1] = 'l';
           memcpy(arg + 2, name.data(), name.size());
           arg[n - 1] = 0;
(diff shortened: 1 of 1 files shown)


gen/declarations.cpp
```diff
@@ -419,12 +419,24 @@ class CodegenVisitor : public Visitor {
...
+        if (isStaticLib == false) {
+          arg = static_cast<char *>(mem.xmalloc(n));
+          arg[0] = '-';
+          arg[1] = 'l';
+          memcpy(arg + 2, name.data(), name.size());
+          arg[n - 1] = 0;
+        } else {
+          arg = static_cast<char *>((mem.xmalloc(nameLen + 1)));
+          memcpy(arg, name.data(), nameLen);
+          arg[nameLen - 1] = 0;
```
kinke, line 437: This is off by one, index should be `nameLen`. - The string needs to be duplicated unfortunately, so that's all good (but ugly).

Issue #4462, opened 2023-07-31
macOS release package does not export all D symbols thus plugins do not work


JohanEngelen:
This is probably related to cross building the package.

I'm testing now by applying `--export-dynamic` to any linker that supports it (so not only on UNIX but for all platforms).

JohanEngelen:
`--export-dynamic` is not supported by the linker that we use for macos arm64 cross build

JohanEngelen:
I think I remember seeing an attempt to use native M1 (arm64) github CI system (using "runs-on: ..."), or was that not tested yet?

kinke:
> I think I remember seeing an attempt to use native M1 (arm64) github CI system (using "runs-on: ..."), or was that not tested yet?

The `macOS arm64` Cirrus CI jobs are native jobs. GitHub Actions have no native AArch64 runners at all, neither Linux nor Mac.

JohanEngelen:
Brute force linking with `-Wl,-exported_symbol;-Wl,'__*'` does work, but gives a ton of warnings like:
`ld: warning: cannot export hidden symbol llvm::ms_demangle::StructorIdentifierNode::~StructorIdentifierNode() from /Users/johan/llvm/install14/lib/libLLVMDemangle.a(MicrosoftDemangleNodes.cpp.o)
`

kinke:
I don't think it's related to cross-building it on an x64 macOS box; especially for the D parts, which basically just get an extra `-mtriple=arm64-apple-macos`. I'd rather check whether `-flto=full` changes the behavior (we also use PGO using an x64 profile, but I don't think that plays a role).

kinke:
Or maybe check `LDC_LINK_MANUALLY=OFF` - it's used by most CI jobs incl. the GHA macOS arm64 cross-build job, meaning that the D host compiler (`ldmd2`) is used for linking, not the CMake C++ compiler. I set DFLAGS for the GHA cross-builds: `DFLAGS=-mtriple=arm64-apple-macos -L-L/Users/runner/work/ldc/build-cross-libs/lib -Xcc=-target -Xcc=arm64-apple-macos -Xcc=-isysroot -Xcc=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk`, to set it up for cross-linking.

kinke:
[...]

PR #4463, opened 2023-08-02
APPLE: export all __* symbols and hide linker warnings.

JohanEngelen:
Resolves issue #4462

This hides many warnings such as: `ld: warning: cannot export hidden symbol llvm::ms_demangle::StructorIdentifierNode::~StructorIdentifierNode() from /Users/johan/llvm/install14/lib/libLLVMDemangle.a(MicrosoftDemangleNodes.cpp.o`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -666,6 +666,8 @@ if(LDC_ENABLE_PLUGINS)
                 append("-disable-linker-strip-dead" DFLAGS_LDC)
             endif()
         endif()
+        message(STATUS "-- Disabling linker warnings (many warnings emitted for symbols that cannot be exported for plugin support)")
+        set(LDC_LINKERFLAG_LIST "${LDC_LINKERFLAG_LIST};-Wl,-exported_symbol;-Wl,'__*';-Wl,-w")
     elseif(UNIX)
         # For plugin support, we need to link with --export-dynamic on Unix.
         # Make sure the linker supports --export-dynamic (on Solaris it is not supported and also not needed).


JohanEngelen:
This is of course quite a brute force fix. However, I don't think we should be maintaining a file with all symbols that need exporting.
This PR is the equivalent of `-Wl,--export-dynamic` that we already use on the Unixes, except that I limited the symbols to `__*`.

kinke:
> However, I don't think we should be maintaining a file with all symbols that need exporting.

Fully agreed! :D

> except that I limited the symbols to `__*`

I.e., all C++ and D symbols, right? Just excluding a few C symbols not starting with an underscore.

---

I'm mildly worried about disabling linker warnings; we might miss some future useful ones this way. As this is apparently only needed for LTO, I'd be okay with adding the extra linker flags for the 2 affected macOS GHA Main jobs only.

JohanEngelen:
It's not only needed for CI, but also for users that build with LTO... :/
I was thinking about trying to detect LTO builds, to limit the impact to such builds.

JohanEngelen:
@kinke Do you know a good way to detect whether a user is doing LTO?  Shall I try to match on some build flags, and hope that's sufficient?

kinke:
*I* use (space-separated) `D_COMPILER_FLAGS` for the official builds, so if they contain `-flto=`, it's an LTO build.

JohanEngelen:
closing in favor of https://github.com/ldc-developers/ldc/pull/4468

JohanEngelen:
closing in favor of https://github.com/ldc-developers/ldc/pull/4468

Issue #4432, opened 2023-07-08
Function pointers don't work on Harvard architechture


dukc:
Compile this with flags `-betterC -mtriple=avr  -mcpu=atmega328p --gcc=avr-gcc --Xcc=-mmcu=atmega328p`

```D
void function() fPtr;

extern(C) void main()
{   fPtr = {fPtr = null;};
}
```

Results in (stack trace omitted): 
```
Invalid bitcast
void ()* bitcast (void () addrspace(1)* @_D5delay4mainUZ9__lambda1FNbNiNfZv to void ()*)
in function main
LLVM ERROR: Broken function found, compilation aborted!
```

Tested version 1.30.0.

kinke:
FWIW, the current problem is that the frontend has this lower bound, but in various codegen sites, we expect `size_t` to match the pointer size, based on: https://github.com/ldc-developers/ldc/blob/cf32bac668540f3dd022493b3ba89b598a485dfb/gen/tollvm.cpp#L334-L349

PR #4465, opened 2023-08-05
Fix function pointers/delegates on Harvard architectures

kinke:
Resolves issue #4432 by setting the address space of such IR pointers.


tests/compilable/funcptr_harvard_gh4432.d
```diff
@@ -0,0 +1,24 @@
...
+{
+    fp();
+    dg();
+}
+
+void bar()
+{
+    foo(() {}, delegate() {});
+
+    FP fp = &bar;
```
JohanEngelen, line 20: Probably good to also add testing of (normal) pointers to variables.
No IR test to validate that the address-space is correct? Or do we assume that is already part of LLVM checking?

kinke: Yeah I was a bit lazy and wasn't sure how far we'd get with this. I've extended the test now, incl. storing and loading a function pointer in static *data*.

dukc:
Great, thanks! I suggest a bit more stress-testing though: assigning a function pointer to a void pointer, pointer arithmetic and casting to/from an integral.

kinke:
> I suggest a bit more stress-testing though: assigning a function pointer to a void pointer, pointer arithmetic and casting to/from an integral.

Heh right, that's exactly what I don't expect to work yet ('not sure how far we'd get with this').

dukc:
> Heh right, that's exactly what I don't expect to work yet ('not sure how far we'd get with this').

Understood :smile: . Function pointers/delegates aren't quite fixed though in that case. Granted, the original issue still is.

PR #4468, opened 2023-08-16
GH Actions: Export ~all symbols from prebuilt LTO'd macOS binaries

### LINKED ISSUE #4462: macOS release package does not export all D symbols thus plugins do not work

kinke:
Fixes #4462 for our official packages.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,9 +42,12 @@ jobs:
             bootstrap_cmake_flags: >-
               -DBUILD_LTO_LIBS=ON
               -DD_COMPILER_FLAGS=-gcc=/usr/bin/c++
+            # https://github.com/ldc-developers/ldc/issues/4462:
+            # When using LTO, we need to explicitly export ~all symbols for plugin support via `ld64 -exported_symbol '__*'`.
+            # Additionally `-w` to suppress resulting linker warnings.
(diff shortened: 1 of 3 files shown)


kinke:
@JohanEngelen: Please let me know if you want to pursue a generic LTO solution or whether that'd be good for now (modulo comments). I think fixing the exports on Mac is the only blocker for beta2.

JohanEngelen:
I think this is OK for now. I guess there are not many people building with LTO anyway.
Can we leave a remark somewhere about the need for these extra flags with LTO ?

kinke:
Feel free to add a further remark wherever you please if you didn't mean a YAML comment which I've just added. [I think the combo of using LTO, *and* needing plugin support, *and* on macOS is probably *really* rare. And I've quickly checked https://github.com/Homebrew/homebrew-core/blob/master/Formula/l/ldc.rb, no LTO being used there.]

JohanEngelen:
> [I think the combo of using LTO, _and_ needing plugin support, _and_ on macOS is probably _really_ rare.]

It does happen though :) Plugins are used at Weka, and people use the Github-built LDC on their macs. So thanks for the fix :)

PR #4390, opened 2023-05-04
Default to rv64gc on riscv64

andreas-schwab:
Fixes #4375

PR #4474, opened 2023-08-25
Default to rv64gc on riscv64 [slightly revised]

kinke:
A slightly revised #4390.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 #### Platform support
 - Supports LLVM 11.0 - 16.0.
-- When nothing else specified now defaults to `-mattr=+m,+a,+f,+d,+c` when targeting hosted 64-bit RISC-V. (#4390)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 #### Platform support
 - Supports LLVM 11.0 - 16.0.
+- When nothing else specified now defaults to `-mattr=+m,+a,+f,+d,+c` when targeting hosted 64-bit RISC-V. (#4390)
```
JohanEngelen, line 10: for me the term "hosted" is new in this context (with "host" and "target" I'm thinking about the "host" system that compiles a program for use on another "target" system).  Is it better to explicitly say for triples with an OS that is not "unknown" or "none" ?

kinke: It was new to me too. - I hope it's somewhat more understandable now.

andreas-schwab: Hosted is the opposite of freestanding.

PR #4476, opened 2023-08-26
Upgrade frontend & libs to v2.105.1

kinke:


kinke:
There's one remaining issue, a segfault on Win32 with `-deps`. I won't be able to debug that one before next week.

kinke:
@JohanEngelen: Please see https://github.com/ldc-developers/ldc/pull/4476/commits/cff6dbdfb3d473d1b4eb41a37c2401056fe5d53c and whether you're fine with the apparently much saner AST now, or whether you'd like to change the testcase to have a somewhat 'beefier' plugin again.

JohanEngelen:
> @JohanEngelen: Please see [cff6dbd](https://github.com/ldc-developers/ldc/commit/cff6dbdfb3d473d1b4eb41a37c2401056fe5d53c) and whether you're fine with the apparently much saner AST now, or whether you'd like to change the testcase to have a somewhat 'beefier' plugin again.

Yep had already seen it and agree with the change. I think it was bad that it the test had this complexity in it, making it so dependent on the actual AST. Because the AST is so unstable, I think it is fine that it is a relatively simple visitor test, proving that the basic visitor case works is OK.

PR #4495, opened 2023-09-16
Windows installer: Add option to add to *system* PATH in admin install mode

### LINKED ISSUE #4494: Windows installer installs to Administrator path when installed from limited user
The Windows installer requires an administrator to install. If the user doing the install is a limited user, this means an administrator types in their password, and it runs under that user.

Once at the end, LDC asks if you want to add LDC to the current user's path. However, what it does is add it to the *Administrator* user's path.

This results in the path not being altered for the user who requested the install.

I propose that the default should be to alter the system path, which is shared by all users. At least the checkbox option should be there to install system-wide path.

kinke:
Resolves #4494.

### DIFF
--- a/packaging/windows_installer.iss
+++ b/packaging/windows_installer.iss
@@ -16,7 +16,7 @@ AppVerName=LDC {#LDCVersion}
 ArchitecturesAllowed=x64
 ; Enable /CURRENTUSER cmdline option to install for current user only, requiring no admin privileges.
 ; This affects the default install dir (override with /DIR="x:\dirname") and the registry root key (HKCU, not HKLM).
-PrivilegesRequiredOverridesAllowed=commandline
+PrivilegesRequiredOverridesAllowed=dialog
 WizardStyle=modern
 DisableProgramGroupPage=yes
 DisableReadyPage=yes
@@ -40,7 +40,8 @@ Name: lib32; Description: "x86 libraries"; Types: full
 
 [Run]
 ; note: not added to PATH for silent installs with /SILENT or /VERYSILENT
-Filename: "{cmd}"; Parameters: "/c echo hello"; Check: not IsInEnvPath; BeforeInstall: AddToEnvPath; Description: "Add to PATH environment variable for current user"; Flags: postinstall skipifsilent runhidden nowait
+Filename: "{cmd}"; Parameters: "/c echo hello"; Check: not IsInUserEnvPath; BeforeInstall: AddToUserEnvPath; Description: "Add to PATH environment variable for current user"; Flags: postinstall skipifsilent runhidden nowait
+Filename: "{cmd}"; Parameters: "/c echo hello"; Check: IsAdminInstallMode and not IsInSystemEnvPath; BeforeInstall: AddToSystemEnvPath; Description: "Add to system PATH environment variable"; Flags: postinstall skipifsilent runhidden nowait unchecked
 Filename: "{app}\README.txt"; Description: "View the README file"; Flags: postinstall shellexec skipifdoesntexist skipifsilent unchecked
 
 [Registry]
@@ -67,13 +68,16 @@ begin
     SendBroadcastNotifyMessage(WM_SETTINGCHANGE, 0, CastStringToInteger(Dummy));
 end;
 
-{ add the target bin dir to user PATH if not already present }
-procedure AddToEnvPath();
+
+{ add the target bin dir to PATH if not already present }
+function AddToEnvPath(const RootKey: Integer; const SubKeyName: string): Boolean;
 var
     Path: string;
     Dir: string;
 begin
-    if not RegQueryStringValue(HKCU, 'Environment', 'Path', Path) then Path := '';
+    result := False;
+
+    if not RegQueryStringValue(RootKey, SubKeyName, 'Path', Path) then Path := '';
 
     Dir := GetTargetBinDir();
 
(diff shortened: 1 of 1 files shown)


schveiguy:
Is there a way I can test (with built artifacts)? Happy to prove it works or not.

kinke:
That'd be awesome, as that's the missing part - I've just blindly added some (Delphi...) code so far.

I believe you should be able to download the `windows-multilib` artifacts from https://github.com/ldc-developers/ldc/actions/runs/6208013760?pr=4495, which should include the .exe Windows installer.

packaging/windows_installer.iss
```diff
@@ -16,7 +16,7 @@ AppVerName=LDC {#LDCVersion}
 ArchitecturesAllowed=x64
 ; Enable /CURRENTUSER cmdline option to install for current user only, requiring no admin privileges.
 ; This affects the default install dir (override with /DIR="x:\dirname") and the registry root key (HKCU, not HKLM).
-PrivilegesRequiredOverridesAllowed=commandline
+PrivilegesRequiredOverridesAllowed=dialog
```
kinke, line 19: According to https://jrsoftware.org/ishelp/index.php?topic=setup_privilegesrequiredoverridesallowed, this should retain the CLI option but also ask whether an admin or local-user-only install should be performed.

packaging/windows_installer.iss
```diff
@@ -40,7 +40,8 @@ Name: lib32; Description: "x86 libraries"; Types: full
 
 [Run]
 ; note: not added to PATH for silent installs with /SILENT or /VERYSILENT
-Filename: "{cmd}"; Parameters: "/c echo hello"; Check: not IsInEnvPath; BeforeInstall: AddToEnvPath; Description: "Add to PATH environment variable for current user"; Flags: postinstall skipifsilent runhidden nowait
+Filename: "{cmd}"; Parameters: "/c echo hello"; Check: not IsInUserEnvPath; BeforeInstall: AddToUserEnvPath; Description: "Add to PATH environment variable for current user"; Flags: postinstall skipifsilent runhidden nowait
+Filename: "{cmd}"; Parameters: "/c echo hello"; Check: IsAdminInstallMode and not IsInSystemEnvPath; BeforeInstall: AddToSystemEnvPath; Description: "Add to system PATH environment variable"; Flags: postinstall skipifsilent runhidden nowait unchecked
```
kinke, line 44: This should be unchecked by default and only visible in admin install mode.

packaging/windows_installer.iss
```diff
@@ -83,21 +87,40 @@ begin
...
 end;
 
-{ remove the target bin dir from user PATH if present }
-procedure RemoveFromEnvPath();
+{ add the target bin dir to system PATH if not already present }
+procedure AddToSystemEnvPath();
+begin
+    if AddToEnvPath(HKLM, 'CurrentControlSet\Control\Session Manager\Environment') then
+    begin
+        Log(Format('Added to user PATH: %s', [GetTargetBinDir()]));
```
schveiguy, line 108: This should probably say added to system path?

schveiguy:
Tested this out:

1. The switch to a dialog instead of a command line switch is great! I tested installing for the local user and it works exactly as designed. Technically this fixes the issue I was having but...
2. The checkbox for adding to the system path doesn't seem to work. I tried both with both boxes checked (local user and system path), and just the system path checked. Neither one adds to system environment variables.

I'll look and see if I can figure out which Key needs setting.

packaging/windows_installer.iss
```diff
@@ -83,21 +87,40 @@ begin
...
         RefreshEnvironment();
     end;
 end;
 
-{ remove the target bin dir from user PATH if present }
-procedure RemoveFromEnvPath();
+{ add the target bin dir to system PATH if not already present }
+procedure AddToSystemEnvPath();
+begin
+    if AddToEnvPath(HKLM, 'CurrentControlSet\Control\Session Manager\Environment') then
```
schveiguy, line 106: I think it's missing the `System` key.

```suggestion
    if AddToEnvPath(HKLM, 'System\CurrentControlSet\Control\Session Manager\Environment') then
```

packaging/windows_installer.iss
```diff
@@ -107,33 +130,67 @@ begin
...
+    begin
+        Log(Format('Removed from user PATH: %s', [GetTargetBinDir()]));
+        RefreshEnvironment();
+    end;
+end;
+
+{ remove the target bin dir from system PATH if present }
+procedure RemoveFromSystemEnvPath();
+begin
+    if RemoveFromEnvPath(HKLM, 'CurrentControlSet\Control\Session Manager\Environment') then
```
schveiguy, line 149:
```suggestion
    if RemoveFromEnvPath(HKLM, 'System\CurrentControlSet\Control\Session Manager\Environment') then
```

packaging/windows_installer.iss
```diff
@@ -107,33 +130,67 @@ begin
...
+{ check if the target bin dir is already in user PATH }
+function IsInUserEnvPath(): Boolean;
+begin
+    result := IsInEnvPath(HKCU, 'Environment');
+end;
+
+{ check if the target bin dir is already in system PATH }
+function IsInSystemEnvPath(): Boolean;
+begin
+    result := IsInEnvPath(HKLM, 'CurrentControlSet\Control\Session Manager\Environment');
```
schveiguy, line 176:
```suggestion
    result := IsInEnvPath(HKLM, 'System\CurrentControlSet\Control\Session Manager\Environment');
```

packaging/windows_installer.iss
kinke: Argh, thx for testing & checking.

schveiguy:
Just tested the updated artifact, and works as designed. Thanks!

schveiguy (approved):

kinke:
Perfect, thank *you*!

PR #4500, opened 2023-09-23
Add initial support for LoongArch64

liushuyu:
This pull request ports the LDC compiler to LoongArch64 CPU architecture.

LoongArch64 is a new CPU architecture and has been supported by LLVM/Clang since LLVM 16.

You can find more information about this architecture here: https://loongson.github.io/LoongArch-Documentation/README-EN.html.

------

The port for D runtime is also in progress at https://github.com/dlang/dmd/pull/15628.

### DIFF
--- a/dmd/cond.d
+++ b/dmd/cond.d
@@ -728,6 +728,10 @@ extern (C++) final class VersionCondition : DVCondition
             case "SPARC_HardFloat":
             case "SPARC_SoftFloat":
             case "SPARC_V8Plus":
+            case "LoongArch32":
+            case "LoongArch64":
+            case "LoongArch_HardFloat":
+            case "LoongArch_SoftFloat":
             case "SystemZ":
             case "SysV3":
             case "SysV4":
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -765,6 +765,14 @@ void registerPredefinedTargetVersions() {
   case llvm::Triple::wasm64:
     VersionCondition::addPredefinedGlobalIdent("WebAssembly");
     break;
+  case llvm::Triple::loongarch32:
+    VersionCondition::addPredefinedGlobalIdent("LoongArch32");
+    registerPredefinedFloatABI("LoongArch_SoftFloat", "LoongArch_HardFloat");
+    break;
+  case llvm::Triple::loongarch64:
+    VersionCondition::addPredefinedGlobalIdent("LoongArch64");
+    registerPredefinedFloatABI("LoongArch_SoftFloat", "LoongArch_HardFloat");
+    break;
   default:
     warning(Loc(), "unknown target CPU architecture: %s",
             triple.getArchName().str().c_str());
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -765,6 +765,7 @@ void registerPredefinedTargetVersions() {
   case llvm::Triple::wasm64:
     VersionCondition::addPredefinedGlobalIdent("WebAssembly");
     break;
+#if LDC_LLVM_VER >= 1600
   case llvm::Triple::loongarch32:
     VersionCondition::addPredefinedGlobalIdent("LoongArch32");
     registerPredefinedFloatABI("LoongArch_SoftFloat", "LoongArch_HardFloat");
@@ -773,6 +774,7 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("LoongArch64");
     registerPredefinedFloatABI("LoongArch_SoftFloat", "LoongArch_HardFloat");
     break;
+#endif // LDC_LLVM_VER >= 1600
   default:
     warning(Loc(), "unknown target CPU architecture: %s",
             triple.getArchName().str().c_str());
--- a/driver/tool.cpp
+++ b/driver/tool.cpp
@@ -96,6 +96,11 @@ void appendTargetArgsForGcc(std::vector<std::string> &args) {
     args.push_back(triple.isArch64Bit() ? "-m64" : "-m32");
     return;
 
+  // LoongArch does not use -m32/-m64 and uses -mabi=.
+  case Triple::loongarch64:
+    args.emplace_back(triple.isArch64Bit() ? "-mabi=lp64d" : "-mabi=ilp32d");
+    return;
+
   // MIPS does not have -m32/-m64 but requires -mabi=.
   case Triple::mips64:
   case Triple::mips64el:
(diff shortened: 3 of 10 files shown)


driver/main.cpp
```diff
@@ -765,6 +765,14 @@ void registerPredefinedTargetVersions() {
   case llvm::Triple::wasm64:
     VersionCondition::addPredefinedGlobalIdent("WebAssembly");
     break;
+  case llvm::Triple::loongarch32:
```
JohanEngelen, line 768: you have the 32-bit version here, but in most other locations you only support 64bit, e.g. in `TargetABI::getTarget() ` and `getTargetCPU`

liushuyu: > you have the 32-bit version here, but in most other locations you only support 64bit, e.g. in `TargetABI::getTarget() ` and `getTargetCPU`

This is because 32-bit LoongArch is still tentative, and there is no real hardware for this specification.
I have added a tentative implementation for LoongArch32 now, but since the 32-bit ABI specification is not stable, I will not include an implementation for it in `TargetABI::getTarget() `.

liushuyu:
New changes since the last push:

- Gated LoongArch support behind LLVM >= 16.0.0 condition. Although the support for LoongArch was added in LLVM 15, the implementation was just a scaffold and was not useful for generating correct machine code.
- Add a missing implementation for the LDC-specific `osthread` facility.

driver/tool.cpp
```diff
@@ -95,7 +95,12 @@ void appendTargetArgsForGcc(std::vector<std::string> &args) {
   case Triple::nvptx64:
     args.push_back(triple.isArch64Bit() ? "-m64" : "-m32");
     return;
-
+#if LDC_LLVM_VER >= 1600
+  // LoongArch does not use -m32/-m64 and uses -mabi=.
+  case Triple::loongarch64:
```
JohanEngelen, line 100: add 32bit here too (otherwise why do the `isArch64Bit` check :))

liushuyu: Ah, sorry, I missed this part.

liushuyu: Wait, I need some clarification. This switch table consists of all the supported 64-bit targets because it is conditioning on the 64-bit triples:

https://github.com/ldc-developers/ldc/blob/ee1a353008363710b140c36887faa1cd46a4df0b/driver/tool.cpp#L87-L89

So, adding 32-bit triple to the case branch is incorrect...?

liushuyu:
All CI tests have successfully passed.

driver/tool.cpp
JohanEngelen: ah, you are correct!

JohanEngelen:
The PR looks good to me.

Tests are missing, perhaps good to have at least one test for the triple and ABI ?

gen/abi/loongarch64.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "gen/irstate.h"
+#include "gen/llvmhelpers.h"
+#include "gen/tollvm.h"
+
+struct LoongArch64TargetABI : TargetABI {
+private:
+  IndirectByvalRewrite indirectByvalRewrite{};
+
+public:
+  auto returnInArg(TypeFunction *tf, bool) -> bool override {
```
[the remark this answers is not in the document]
liushuyu: > nitpick: is there any benefit to this `auto .... -> bool` over `bool ...` ? We don't really use this syntax in LDC's codebase anywhere else.

When I run `clang-tidy` on my changes, it tells me to use this expression instead of `bool func_name(args...)`.
I do notice you have a `.clang-tidy` configuration in the project root. Perhaps that configuration makes `clang-tidy` to add this specific lint? (I am using Clang 16 on my system).

liushuyu:
> Tests are missing, perhaps good to have at least one test for the triple and ABI ?

Will do in a bit.

liushuyu:
> Tests are missing, perhaps good to have at least one test for the triple and ABI ?

Done. Let's see if CI will pick it up.

liushuyu:
CI passed without issue and the new test was picked up by CIs that running with LLVM 16

dmd/cond.d
```diff
@@ -728,6 +728,10 @@ extern (C++) final class VersionCondition : DVCondition
             case "SPARC_HardFloat":
             case "SPARC_SoftFloat":
             case "SPARC_V8Plus":
+            case "LoongArch32":
+            case "LoongArch64":
+            case "LoongArch_HardFloat":
+            case "LoongArch_SoftFloat":
```
kinke, line 734: [We'd get this with the next frontend merge automatically, after your upstream PR has been merged. But is okay (we usually guard LDC-specific frontends mods in `version (IN_LLVM)` for future merge conflicts).]

gen/abi/loongarch64.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+    // pass by reference when > 2*GRLEN
+    return rt->size() > 16;
+  }
+
+  auto passByVal(TypeFunction *, Type *t) -> bool override {
+    if (!t->size()) {
+      return false;
+    }
+    return t->size() > 16;
```
kinke, line 44: The `if` is superfluous. And IIRC, you'd need an `isPOD` check here too, *not* using `byval` for these args to be rewritten by `IndirectByvalRewrite`.

tests/driver/loongarch64_abi.d
```diff
@@ -0,0 +1,6 @@
+// REQUIRES: target_LoongArch
+
+// RUN: %ldc %s -mtriple=loongarch64-unknown-linux-gnu -mattr=+f,+d --gcc=echo > %t && FileCheck %s < %t
+// CHECK: -mabi=lp64d
+
```
kinke, line 5: A little check to make sure the `LoongArch64` version is predefined would be nice, see e.g. https://github.com/ldc-developers/ldc/blob/5a158843b51b309bf13b764643d30431e3ea557e/tests/codegen/avr.d#L5-L6.

liushuyu: Done.

gen/abi/loongarch64.cpp
liushuyu: I am unsure if I have done this correctly, please see the new commits.

kinke: Looks all good. :+1:

kinke (approved):

liushuyu:
Hi,

DMD upstream has merged the LoongArch support for the D runtime. Do you want me to do a merge, or do you want to do it yourself?

JohanEngelen (approved):

JohanEngelen:
not adding it to release notes, because runtime support is lacking and will only come with dlang 2.106

zoujiaqing:
Thank u ;)

PR #4504, opened 2023-09-27
[Dcompute] don't crash on `VarDelaration`s with no type

thewilsonator:
I'm not entirely sure why the VarDeclaration `isUnsigned` has a null type field, but it does.

### DIFF
--- a/tests/semantic/dcompute_enum.d
+++ b/tests/semantic/dcompute_enum.d
@@ -1,0 +1,19 @@
+// RUN: %ldc -o- -mdcompute-targets=cuda-430 %s
+
+@compute(CompileFor.deviceOnly) module dcompute_enum;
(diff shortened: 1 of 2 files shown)


tests/semantic/dcompute_enum.d
```diff
@@ -0,0 +1,19 @@
+// RUN: %ldc -o- -mdcompute-targets=cuda-430 %s
```
thewilsonator, line 1: Hmm, this should be `cuda-350`, `430` is not a valid number, not sure why this is passing, is it not running?

kinke: > PASS: LDC :: semantic/dcompute_enum.d (348 of 355)

Maybe only verified when actually generating code without `-o-`?

Issue #4508, opened 2023-10-09
`-DRT_SUPPORT_SANITIZERS=ON` changes runtime ABI


JohanEngelen:
The runtime ABI should not change for compilation with `-d-version=SupportSanitizers`, yet it does because of: https://github.com/ldc-developers/ldc/blob/a49db8e5d383094cb569057966689bf32ae599d5/runtime/druntime/src/core/thread/fiber.d#L554
and
https://github.com/ldc-developers/ldc/blob/a49db8e5d383094cb569057966689bf32ae599d5/runtime/druntime/src/core/thread/fiber.d#L1169-L1172

PR #4509, opened 2023-10-09
Unconditionally define m_allowMigration field in Fiber

JohanEngelen:
fixes issue #4508


JohanEngelen:
Note that LDC has diverged quite significantly from DMD in this regard.

kinke (approved):

PR #4511, opened 2023-10-10
Fix a regression in gc2stack optimization

### LINKED ISSUE #4510: vector operations are broken with optimization level >= 2
```d
import std.stdio;

void main()
{
  int count;
  foreach (i; 0..34)
  {
    auto flags = new bool[](1);
    if (flags[0] == false) count++;
    flags[] = true;
  }
  writeln(count);
  // expected 34, but LDC with --O2 outputs 1.
}
```

With `--O2` flag, it shows the unexpected result.

```console
(ldc-1.34.0)kubo39@hinoda:~/dev/dlang$ ldc2 -run test.d
34
(ldc-1.34.0)kubo39@hinoda:~/dev/dlang$ ldc2 --O2 -run test.d
1
```

With `--disable-gc2stack` flag, works as expected.

```console
(ldc-1.34.0)kubo39@hinoda:~/dev/dlang$ ldc2 --O2 --disable-gc2stack -run test.d
34
```

kubo39:
This fixes #4510.
LLVM RAII helper did not fit in case, at https://github.com/ldc-developers/ldc/commit/1d969cfccaf7181b5760a0ae7cd6088a36e3de44.

### DIFF
--- a/gen/passes/GarbageCollect2Stack.cpp
+++ b/gen/passes/GarbageCollect2Stack.cpp
@@ -161,16 +161,17 @@ bool ArrayFI::analyze(CallBase *CB, const G2StackAnalysis &A) {
 }
 
 Value* ArrayFI::promote(CallBase *CB, IRBuilder<> &B, const G2StackAnalysis &A) {
+  IRBuilder<> Builder(B.GetInsertBlock(), B.GetInsertPoint());
+
   // If the allocation is of constant size it's best to put it in the
   // entry block, so do so if we're not already there.
   // For dynamically-sized allocations it's best to avoid the overhead
   // of allocating them if possible, so leave those where they are.
   // While we're at it, update statistics too.
-  const IRBuilderBase::InsertPointGuard savedInsertPoint(B);
   if (isa<Constant>(arrSize)) {
     BasicBlock &Entry = CB->getCaller()->getEntryBlock();
-    if (B.GetInsertBlock() != &Entry) {
-      B.SetInsertPoint(&Entry, Entry.begin());
+    if (Builder.GetInsertBlock() != &Entry) {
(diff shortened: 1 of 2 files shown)


gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -261,28 +262,29 @@ bool UntypedMemoryFI::analyze(CallBase *CB, const G2StackAnalysis &A) {
   return true;
 }
 Value* UntypedMemoryFI::promote(CallBase *CB, IRBuilder<> &B, const G2StackAnalysis &A) {
+  IRBuilder<> Builder(B.GetInsertBlock(), B.GetInsertPoint());
```
JohanEngelen, line 265: the diff in this function is not needed.

gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -161,26 +161,27 @@ bool ArrayFI::analyze(CallBase *CB, const G2StackAnalysis &A) {
...
 
   // Convert array size to 32 bits if necessary
-  Value *count = B.CreateIntCast(arrSize, B.getInt32Ty(), false);
+  Value *count = Builder.CreateIntCast(arrSize, Builder.getInt32Ty(), false);
   AllocaInst *alloca =
-      B.CreateAlloca(Ty, count, ".nongc_mem"); // FIXME: align?
+      Builder.CreateAlloca(Ty, count, ".nongc_mem"); // FIXME: align?
 
   if (Initialized) {
     // For now, only zero-init is supported.
```
JohanEngelen, line 187: can you improve the comment on line 190-191 to say that the initialization must be put at the original source variable definition location (explicit mention: because it could be in a loop).

PR #4515, opened 2023-10-21
Upstream FreeBSD-specific patches

arrowd:


JohanEngelen:
Great thanks!

driver/targetmachine.cpp
```diff
@@ -540,6 +540,10 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
       // these OSes.
       // On Android, PIC is default as well.
       relocModel = llvm::Reloc::PIC_;
+    } else if (triple.isOSFreeBSD()) {
+      // We default to PIC code to avoid linking issues on FreeBSD, especially
+      // on aarch64.
+      relocModel = llvm::Reloc::PIC_;
```
JohanEngelen, line 546: should this be for a specific version of FreeBSD and upwards?  Or do all freebsd versions since 9 use PIC ?

arrowd: 9 is long unsupported. That statement is true for 12+ although 12.4 is also going EoL in about a month.

JohanEngelen: OK, and the user can always override it with existing cmdline option

PR #4533, opened 2023-11-25
Fix compilation with LLVM 17.

JohanEngelen:
I did not test anything real yet, but the build succeeds for me and `ldc2 --help` works.
LLVM17 removes support for typed pointers. We don't enable opaque pointers by default on pre-17, so I expect bugs to still linger with `--opaque-pointers=true`.


JohanEngelen:
Looking pretty good!  Only gdb issues because I had to update the tester gdb (gdb 9 not available for ubuntu 22) .....

JohanEngelen:
@kinke  My proposal is to release 1.36 with LLVM16 and with opaque pointers enabled by default. And then release 1.37 with LLVM17.  This way, there is a release with opaque pointers by default (such that users may discover a bug) but users still have the option to disable opaque pointers and check for deltas without changing compiler&LLVM version.

kinke:
Thx for working on this.

Wrt. gdb, we'll eventually have to tackle it and see how to please gdb v10+.

Wrt. opaque pointers, I think the suggestion makes sense. AFAIK, there's sadly no way to overload a pointer comparison, to find all occurrences where we compare Type pointers.

JohanEngelen:
> Wrt. opaque pointers, I think the suggestion makes sense. AFAIK, there's sadly no way to overload a pointer comparison, to find all occurrences where we compare Type pointers.

I checked all getType() comparisons, I think they are all benign with opaque pointers.

JohanEngelen:
I've enabled opaque pointers for LLVM16 too in this PR, to see if LLVM16 starts to fail on something, i.e. there is still work to be done.

JohanEngelen:
Testsuite uncovered this bug: https://github.com/ldc-developers/ldc/issues/4538

cmake/Modules/FindLLVM.cmake
```diff
@@ -32,7 +32,9 @@
 # We also want an user-specified LLVM_ROOT_DIR to take precedence over the
 # system default locations such as /usr/local/bin. Executing find_program()
 # multiples times is the approach recommended in the docs.
-set(llvm_config_names llvm-config-16.0 llvm-config160 llvm-config-16
+set(llvm_config_names llvm-config-18.0 llvm-config180 llvm-config-18
```
kinke, line 35: I've so far tried to avoid future LLVM versions here, as they are bound to fail. For testing/working on supporting the next version, specifying `LLVM_ROOT_DIR` should be simple enough.

driver/main.cpp
```diff
@@ -597,9 +605,11 @@ void initializePasses() {
   initializeGlobalISel(Registry);
   initializeTarget(Registry);
 
+#if LDC_LLVM_VER < 1700
 // Initialize passes not included above
   initializeRewriteSymbolsLegacyPassPass(Registry);
   initializeSjLjEHPreparePass(Registry);
```
kinke, line 611: The latter one seems to exist still at least.

kinke (approved):

cmake/Modules/FindLLVM.cmake
JohanEngelen: OK, I've removed it.

Issue #4501, opened 2023-09-24
Latest xcode results in lots of linker warnings


schveiguy:
A sample for when I build a simple app with ldc 1.34.0  with xcode 15.0 (there are a lot of these):

```
ld: warning: no platform load command found in '/Users/steves/.dub/cache/numbers/~master/build/application-debug-WophZRlX_VEn3wfrspNcCQ/numbers.o', assuming: macOS
ld: warning: no platform load command found in '/Users/steves/ldc2-1.34.0-osx-universal/lib-arm64/libphobos2-ldc.a[2](curl.o)', assuming: macOS
ld: warning: no platform load command found in '/Users/steves/ldc2-1.34.0-osx-universal/lib-arm64/libphobos2-ldc.a[8](zlib.o)', assuming: macOS
ld: warning: no platform load command found in '/Users/steves/ldc2-1.34.0-osx-universal/lib-arm64/libphobos2-ldc.a[9](comparison.o)', assuming: macOS
ld: warning: no platform load command found in '/Users/steves/ldc2-1.34.0-osx-universal/lib-arm64/libphobos2-ldc.a[11](iteration.o)', assuming: macOS
```

It's unfortunate, Apple seems to change the linker rules every release...

kinke:
The apparently common workaround is to use `-L-ld_classic`:
* https://github.com/desktop-app/cmake_helpers/commit/aaedaede705ab908b1bbaf026c59b4379b0d1cf8
* https://github.com/crystal-lang/crystal/issues/13846
* https://projects.blender.org/blender/blender/pulls/110243/files

drpriver:
`-L-w` also suppresses the warnings (though it suppresses all linker warnings).

jacob-carlborg:
Hmm, I thought this would be a thing that LLVM does automatically.

I implemented a fix for a similar warning for DMD a couple of years ago. It was never merged but it shows more or less what needs to be done [1].

[1] https://github.com/dlang/dmd/pull/10476

JohanEngelen:
Fixed in #4534

jacob-carlborg:
Did this not make the 1.36.0 release? I still get these warnings. Actually, with 1.35.0 I get these warnings for both object files from Phobos/druntime and for the object files compiled from my own code. With 1.36.0 I only get warnings for object files from Phobos/druntime.

JohanEngelen:
[...]

PR #4534, opened 2023-11-25
We have to explicitly set the OS version on newer macOSes to support LTO

JohanEngelen:
This also fixes these annoying linker warnings:
`ld: warning: no platform load command found in 'libphobos2-ldc.a[113](core.o)', assuming: macOS`  #4501

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -16,6 +16,7 @@
 #include "driver/targetmachine.h"
 
 #include "dmd/errors.h"
+#include "driver/args.h"
 #include "driver/cl_options.h"
 #include "gen/logger.h"
 #include "llvm/ADT/StringExtras.h"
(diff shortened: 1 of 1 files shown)


JohanEngelen:
reference: https://discourse.llvm.org/t/expressing-the-darwin-macos-target-build-command-in-llvm-ir/74198?u=vtjnash

JohanEngelen:
Should we respect `MACOSX_DEPLOYMENT_TARGET` environment variable for setting macos version? (and use host version if the env variable is not set)

kinke:
According to the GHA CI logs, the default host triple for the macOS jobs is `arm64-apple-darwin21.6.0` / `x86_64-apple-darwin20.6.0`. Do **we** really need to convert that to `macos`, *and* come up with some compatible macOS version?!

JohanEngelen:
> According to the GHA CI logs, the default host triple for the macOS jobs is `arm64-apple-darwin21.6.0` / `x86_64-apple-darwin20.6.0`. Do **we** really need to convert that to `macos`, _and_ come up with some compatible macOS version?!

Yep I guess so.
It fixes all linker issues and warnings on my machine. CI is green now too.

JohanEngelen:
(it's also what Clang does)

driver/targetmachine.cpp
```diff
@@ -440,9 +441,30 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
...
+      // If environment variable MACOSX_DEPLOYMENT_TARGET is not set, then use
+      // host OS version.
+      // https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/cross_development/Configuring/configuring.html
+      std::string envVersion = env::get("MACOSX_DEPLOYMENT_TARGET");
+      if (!envVersion.empty()) {
+        osname += envVersion;
+      } else {
+#if LDC_LLVM_VER >= 1400
+        llvm::VersionTuple OSVersion;
+        triple.getMacOSXVersion(OSVersion);
```
kinke, line 459: At least there's this.

driver/targetmachine.cpp
```diff
@@ -440,9 +441,30 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
...
+        osname += OSVersion.getAsString();
+#else
+        // Hardcode the version, because `getMacOSXVersion` is not available.
+        osname += "10.4";
```
kinke, line 463: I think that's a bit too conservative; we require 10.7 as absolute minimum for native TLS IIRC.

JohanEngelen:
Ready to merge imo.

kinke (approved):

JohanEngelen:
> Does this fix the issue on all Apple platforms (assuming there was an issue) or just on macOS?

I do not know, and do not know how to test it either.

jacob-carlborg:
> I do not know, and do not know how to test it either.

@JohanEngelen cross-compile and use `otool` and look for the appropriate load command:

```
ldc2 -mtriple arm64-apple-ios -c main.d
otool -l main.o
```

Look for a command named `LC_BUILD_VERSION`.

The problem occurs on other Apple platforms as well.

kinke:
When cross-compiling using an explicit target triple, I think the only reasonable way for Apple targets is to include the min OS version there explicitly. E.g., we cross-compile the iOS libs via `<arch>-apple-ios12.0`.

jacob-carlborg:
Oh, the load command is apparently included when an explicit version is set in the target.

But the SDK version is not set. Not sure how important that is:

```
Load command 1
      cmd LC_BUILD_VERSION
  cmdsize 24
 platform 1
    minos 12.0
      sdk n/a
   ntools 0
```

Issue #4538, opened 2023-11-29
Miscompilation with -opaque-pointers=true


JohanEngelen:
Testcase from druntime `rt/lifetime.d` line 2686 miscompiles with `-opaque-pointers=true -O`:
```
static struct S
{
    S* thisptr;
    ~this() { assert(&this == thisptr); thisptr = null;}  // This assertion is triggered
}

void main() {
    S[] test14126 = new S[2048]; // make sure we allocate at least a PAGE
    foreach (ref s; test14126)
    {
        s.thisptr = &s;
    }
}
```

(no need to add this testcase when the bug is fixed, because it is already part of the testsuite when opaque pointers are enabled)

kinke:
FWIW, I've just pushed https://github.com/ldc-developers/llvm-project/tree/ldc-release/17.x (preliminary so far).

kinke:
Also seen with LLVM v15.0.6 on Linux x86_64: https://github.com/ldc-developers/ldc/actions/runs/9052573385/job/24870398659?pr=4657

PR #4540, opened 2023-12-03
Switch to LDC-LLVM v17.0.6

kinke:


kinke:
I currently tend towards using this for the upcoming v1.36.0-beta1 release; the miscompile with LLVM 16 and opaque IR pointers (#4538) is really bad, so the original plan of using LLVM 16 and newly defaulting to opaque pointers is IMO not viable. Any opinions?

JohanEngelen:
OK.
I still hope to find some time to distill the miscompilation to an IR testcase. Maybe there is still some miscompilation lurking in llvm 17...

kinke:
I'll give the beta some extensive tests on the Symmetry codebase then. - LLVM 16 seemed pretty buggy - first the `func-specialization` miscompiles, then the opaque pointers thing. Both of these seem to have been fixed in LLVM 17 now at least.

PR #4549, opened 2023-12-27
Add support for -ffat-lto-objects.

JohanEngelen:


JohanEngelen:
I think Fat LTO is still pretty experimental in LLVM, but clang provides it, so why not us. Does not impact other functionality.

driver/linker-gcc.cpp
```diff
@@ -201,19 +201,24 @@ void ArgsBuilder::addDarwinLTOFlags() {
...
-                 (opts::linker.empty() && isLldDefaultLinker());
     addLTOGoldPluginFlags(!isLld);
   } else if (global.params.targetTriple->isOSDarwin()) {
     addDarwinLTOFlags();
   }
+
+  // Pass Fat LTO option to LLD
+  if (isLld && opts::ltoFatObjects) {
+      addLdFlag("--fat-lto-objects");
+  }
```
kinke, line 221: Hmm, why does LLD need that info? I assume the info is available in the object files themselves.

gen/optimizer.cpp
```diff
@@ -729,7 +729,17 @@ void runOptimizationPasses(llvm::Module *M) {
   OptimizationLevel level = getOptimizationLevel();
 
   if (optLevelVal == 0) {
-    mpm = pb.buildO0DefaultPipeline(level, opts::isUsingLTO() || opts::isUsingThinLTO());
+    mpm = pb.buildO0DefaultPipeline(level, opts::isUsingLTO());
```
kinke, line 732: The removed thin-LTO check looks like a functional change. Not sure how much sense LTO without optimizations makes anyway though...

JohanEngelen: This is not a functional change. `isUsingLTO` is `true` for any type of LTO.

JohanEngelen: it is not needed for this PR, that's true.

driver/linker-gcc.cpp
JohanEngelen: Clang also passes this. I assumed it is needed when making a library.

kinke (approved):

kinke:
Wanna get this into v2.107?

JohanEngelen:
Ah sure, we can put this in. I didn't really test it, but I don't think it hurts anyone by putting it in.

PR #4553, opened 2024-01-06
CI: Always include macOS version in explicit triples

kinke:
Wrt. issue #4501.

### DIFF
--- a/.github/actions/merge-macos/action.yml
+++ b/.github/actions/merge-macos/action.yml
@@ -64,7 +64,7 @@ runs:
             switches = [
                 \"-defaultlib=phobos2-ldc,druntime-ldc\",
                 \"-Xcc=-target\",
-                \"-Xcc=x86_64-apple-macos\",
+                \"-Xcc=x86_64-apple-macos$MACOSX_DEPLOYMENT_TARGET\",
             ];
             lib-dirs = [
                 \"%%ldcbinarypath%%/../lib-x86_64\",
@@ -77,7 +77,7 @@ runs:
             switches = [
                 \"-defaultlib=phobos2-ldc,druntime-ldc\",
                 \"-Xcc=-target\",
-                \"-Xcc=arm64-apple-macos\",
+                \"-Xcc=arm64-apple-macos$MACOSX_DEPLOYMENT_TARGET\",
             ];
             lib-dirs = [
                 \"%%ldcbinarypath%%/../lib-arm64\",
@@ -109,7 +109,7 @@ runs:
       run: |
         set -eux
         echo 'void main() { import std.stdio; writefln("Hello world, %d bits", size_t.sizeof * 8); }' > hello.d
-        for os in macos ios; do
+        for os in macos$MACOSX_DEPLOYMENT_TARGET ios12.0; do
           for arch in x86_64 arm64; do
             triple="$arch-apple-$os"
             ldc2-universal/bin/ldc2 -mtriple="$triple" hello.d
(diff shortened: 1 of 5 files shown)


driver/targetmachine.cpp
```diff
@@ -450,10 +450,10 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
     triple = llvm::Triple(
         llvm::Triple::normalize(llvm::sys::getDefaultTargetTriple()));
 
+    // Apple: translate darwin to macos, apparently like clang
     if (triple.getOS() == llvm::Triple::Darwin) {
-      // We only support OSX, so darwin should really be macosx.
       llvm::SmallString<16> osname;
-      osname += "macosx";
+      osname += "macos";
```
kinke, line 456: I've stripped off the `x`; Apple uses `macos{10.12,11}` in https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary, and Google yields 4-5x more results for `arm64-apple-macos11` than `arm64-apple-macosx11`.

JohanEngelen: tested locally: works.

.github/actions/merge-macos/action.yml
```diff
@@ -63,8 +63,8 @@ runs:
         {
             switches = [
                 \"-defaultlib=phobos2-ldc,druntime-ldc\",
-                \"-Xcc=-target\",
-                \"-Xcc=x86_64-apple-macos\",
+                \"-Xcc=-arch\",
+                \"-Xcc=x86_64\",
```
kinke, line 67: I *think* this is better than specifying a full triple with some hardcoded macOS version.

JohanEngelen:
@jacob-carlborg  Can you verify that this now works for you?  You can grab the package from https://github.com/ldc-developers/ldc/releases/tag/CI  (wait until rebuild with this PR included)

jacob-carlborg:
@JohanEngelen how do I know if this PR is included in the CI release?

jacob-carlborg:
@JohanEngelen Seems to be working. I don't get any warnings and the object files contains the correct sections.

JohanEngelen:
@jacob-carlborg Thanks for confirming the fix. (the 8264f97 binaries contain this PR, Github says they were uploaded ~8hours ago)

.github/actions/merge-macos/action.yml
JohanEngelen: This change leads to a problem on my mac. The compiler does not recognize `-arch` (it wants `-march`).
I'm testing what works best...

JohanEngelen: ah cool... clang wants `-arch` and `c++` wants `-march` ...

JohanEngelen: And everything works when the `target_link_options` version of #4431 is used.

JohanEngelen: (the solution with `-target` works, but runs into an annoying warning about unused `-arch arm64` warning when linking LDC; so trying to find other solution)

jacob-carlborg: Is it possible to get the full triple that was matched and then insert it with the `-target` flag?

Another idea would be to not specify this flag in the config file and instead let LDC get the operating system version at runtime and use the full triple including the version. It's more flexible than a config file.

kinke: Oh right - we could simply have LDC always add an appropriate `-target` option when linking an Apple target. Then we only need the `-isysroot` option in the config file, for iOS targets.

JohanEngelen: #4431 now resolves building LDC with this, so from my side no change is needed (and I think the arch flag is better than triple, because the arch flag can't be wrong here)

Issue #4556, opened 2024-01-08
`"a" is "b"` is true


FeepingCreature:
Repro:

```
extern(C) void printf(const char*, ...);

void main()
{
    if ("a" is "b")
    {
        printf("oops.\n");
    }
}
```

The issue goes away if I switch to `IRBuilder<llvm::NoFolder>`. I suspect this is a side effect of strings being `unnamed_addr`, meaning LLVM not-entirely-unreasonably concludes "well, I can interpret `icmp eq i8* @.str.1, @.str.2` however I feel like". Of course, "a" will never actually be merged with "b", but...

JohanEngelen:
This bug is not related to LLVM or wrong optimization by LLVM. The generated code does not even contain an `icmp` because LDC recognizes that the condition is a compile-time constant and thus is known to be either true or false at compile-time and no branching is needed. (#3134, since 1.19)
The bug arises since 1.25, where for some reason `"a" is "b"` evaluates to `true` (when it is the condition of `if`).

JohanEngelen:
Related LLVM change:  https://reviews.llvm.org/D87123

PR #4559, opened 2024-01-08
Fix if-statement elision on constant true/false condition, for condit…

JohanEngelen:
…ion expressions that are more complex than a simple integer.

Resolves issue #4556


kinke:
Is the Circle CI failure caused by using the distro-LLVM v11.0.0, whereas the vanilla-LLVM job uses v11.1.0?

JohanEngelen:
> Is the Circle CI failure caused by using the distro-LLVM v11.0.0, whereas the vanilla-LLVM job uses v11.1.0?

Yep. The testcase is a bit strange too, let me think about it.

PR #4564, opened 2024-01-24
Strict aliasing fixes

the-horo:


gen/abi/x86-64.cpp
```diff
@@ -259,7 +265,7 @@ void X86_64TargetABI::rewriteArgument(IrFuncTy &fty, IrFuncTyArg &arg,
 }
 
 void X86_64TargetABI::rewriteFunctionType(IrFuncTy &fty) {
-  RegCount &regCount = getRegCount(fty);
+  RegCount regCount = getRegCount(fty);
   regCount = RegCount(); // initialize
```
kinke, line 269: This isn't needed anymore; a simple `RegCount regCount;` can now replace the 2 lines.

the-horo: Oops, I didn't notice it.

kinke (approved):

kinke:
Thanks!

PR #4572, opened 2024-02-02
This line points to empty dir?

denizzzka:
When this ldc2.conf is used this line points to dir which contains empty `ldc/` dir


kinke:
It is usually populated with gcc_builtins modules, see https://github.com/ldc-developers/ldc/blob/cbdce51f41e20b690ce505caabb184c77b08930a/runtime/CMakeLists.txt#L337-L364. Not for a standalone druntime/Phobos build without compiler though (e.g., via `ldc-build-runtime` tool).

PR #4580, opened 2024-02-13
macOS: Don't link lib{pthread,m} as platform libs

kinke:


JohanEngelen:
Is there something I can test for you locally?
(perhaps just the linker-gcc.cpp change?)

kinke:
Hmm, I don't know. I just hate that our CMake build doesn't work out of the box with recent Xcode, requiring that dead-ugly `LDC_LINK_MANUALLY=OFF -DD_COMPILER_FLAGS=-gcc=/usr/bin/c++` workaround.

PR #4582, opened 2024-02-16
support CC envvar to be in the form "command args..."

the-horo:
In case CC is set try to split it by spaces into a command and arguments. This is the same behaviour as in dmd.

### DIFF
--- a/tests/driver/cli_CC_envvar.d
+++ b/tests/driver/cli_CC_envvar.d
@@ -1,0 +1,6 @@
+// UNSUPPORTED: Windows
+
+// RUN: not env CC='cc --linker-arg' %ldc -vv %s | FileCheck %s
+
+// CHECK: Linking with:
+// CHECK-NEXT: --linker-arg
--- a/driver/cpreprocessor.cpp
+++ b/driver/cpreprocessor.cpp
@@ -22,9 +22,11 @@ const char *getPathToImportc_h(const Loc &loc) {
   return cached;
 }
 
-const std::string &getCC(bool isMSVC) {
-  static std::string cached;
-  if (cached.empty()) {
+const std::string &getCC(bool isMSVC,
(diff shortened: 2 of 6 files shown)


kinke:
Out of interest, is that a non-uncommon thing? [I've never seen it before I think.] Or e.g. just some (D-specific?) workaround?

the-horo:
On Gentoo, multilib has a set of predefined flags which you can see at: https://github.com/gentoo/gentoo/blob/master/profiles/arch/amd64/make.defaults#L26-L33. When a package is compiled for multilib, environment variables like CC will be set to "$CC $CFLAGS_abi" which makes ldc fail to link.

Whether this is common for other systems or distributions I don't know.

If you care for more details, the solution currently in Gentoo is resetting CC to a usual value, which I'm trying to move away from since modifying variables like this makes it hard to follow code. I can solve the problem in Gentoo but I thought it's a good enough feature to implement here as well.

kinke:
Alright fine, if it's common on Gentoo at least and not some D kludge (potentially replaceable by `-Xcc` and/or `-L` unless order matters). Thx for the rationale. :+1:

tests/driver/cli_CC_envvar.d
```diff
@@ -0,0 +1,4 @@
+// RUN: not env CC='cc --linker-arg' %ldc -vv %s | FileCheck %s
```
kinke, line 1: Adding `// UNSUPPORTED: Windows` excludes the test on Windows (where linking doesn't go through a C compiler, and CC is unused).

the-horo: I, apparently, failed to see the red cross that meant some tests had failed and have also failed to read the title of the CI email. At least I learned a little of how ldc works on Windows.

PR #4589, opened 2024-03-01
Make compiler inject appropriate -target option when linking Mach-O binaries

kinke:


JohanEngelen:
I'm going on vacation for a week, so can't test this until after 10 March.

This change is a little risky I have noticed, because the linker spits out a warning when it gets both `-arch ...` and `-triple ...` (which did happen for something similar to this change when working on #4431). Not nice to always see a warning when you compile a D program.

kinke:
Don't worry, this is v1.38 stuff. :) - But I think it makes sense, more flexible this way (I hated having to hardcode iOS 12.0 in the config...).

JohanEngelen:
I downloaded the universal build artifact and tested building LDC, and it failed when linking because again the `-target` is not detected correctly (I knew this was going to happen for when -target is at the end of link flags...):
```
-- Host D compiler linker flags: -L/Users/johan/ldc/test_gha/ldc2-746fb353-osx-universal/bin/../lib-arm64;-lphobos2-ldc;-ldruntime-ldc;-Wl,-dead_strip;-lpthread;-lm;-target;arm64-apple-macos14
```
this results in `-larm64-apple-macos14` flag.

I fixed that issue here: https://github.com/ldc-developers/ldc/pull/4603

With that change, building LDC works without issues.

JohanEngelen:
LGTM.

(the user can always override this triple with `-mtriple=`)

PR #4596, opened 2024-03-17
Add memory invariant assumptions for virtual calls to improve devirtualization

JohanEngelen:
Assumptions added:
- class methods do not change the object's vtable pointer
- loads through vtable pointer are invariant (vtables are immutable throughout program life)

Examples of the improvement in devirtualization:
```d
class A {
    void foo();
    void oof();
}

void ggg() {
    A a = new A();
    a.foo();  // was already devirtualized
    a.foo();  // devirtualized by this PR
}

void exacttypeunknown(A a, A b) {
    a.foo();
    a.foo(); // no vtable lookup, PR removes 2 memory reads
}
```

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -778,6 +778,29 @@ public:
     DValue *result =
         DtoCallFunction(e->loc, e->type, fnval, e->arguments, sretPointer);
 
+    // If this is a virtual function call, the object is passed by reference
+    // through the `this` parameter, and therefore the optimizer has to assume
+    // that the vtable field might be overwritten. This prevents optimization of
+    // subsequent virtual calls on the same object. We help the optimizer by
+    // allowing it to assume that the vtable field contents is the same after
+    // the call. Equivalent D code:
+    // ```
+    //  auto saved_vtable = a.__vptr;     // emitted as part of `a.foo()`
+    //  a.foo();
+    //  assume(a.__vptr == saved_vtable); // <-- added assumption
+    // ```
+    // Only emit this extra code from -O2.
+    if (!e->directcall && dfnval && dfnval->vtable && (optLevel() >= 2)) {
+      // Reload vtable ptr. It's the first element so instead of GEP+load we can
+      // do a void* load+bitcast (at this point in the code we don't have easy
+      // access to the type of the class to do a GEP).
+      auto vtable =
+          DtoLoad(dfnval->vtable->getType(),
+                  DtoBitCast(dfnval->vthis, dfnval->vtable->getType()->getPointerTo()));
+      auto cmp = p->ir->CreateICmpEQ(vtable, dfnval->vtable);
+      p->ir->CreateCall(GET_INTRINSIC_DECL(assume), {cmp});
+    }
+
     if (delayedDtorVar) {
       delayedDtorVar->edtor = delayedDtorExp;
       pushVarDtorCleanup(p, delayedDtorVar);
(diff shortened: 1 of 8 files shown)


PetarKirov:
At a language level, I think the `__vptr` should be properly typed. At minimum it should be `immutable void**`. Ideally it should be:

```d
class MyClass {
  immutable MyClass_Vtbl* __vptr;
  object.Object.Monitor __monitor;

  override string toString() => "overriden";
  void foo();
  string bar(float b) const;
}

immutable struct MyClass_Vtbl {
  object.TypeInfo_Class typeInfo;

  // object.Object methods
  string function(MyClass self) toString; // note: first parameter is typed `MyClass`, because the method is overriden
  size_t function(Object self) toHash;
  int function(Object self, Object o) opCmp;
  bool function(Object self, Object o) opEquals;

  // MyClass methods
  void function(MyClass self) foo;
  string function(const MyClasss self, float b) bar;
}
```

JohanEngelen:
A simpler explanation:  the very fact that we construct the class and _write_ the vptr to it, makes the memory location non-immutable throughout program lifetime.  (and then you can also destroy the class and construct a new class in the same location).
Because of that, I feel it is safer to add the assumption of equality after member function calls, rather than marking vptr loads as invariant.load.

kinke:
> Because of that, I feel it is safer to add the assumption of equality after member function calls, rather than marking vptr loads as invariant.load.

Yeah that's exactly what I hoped for at first too, but after reading the `invariant.load` spec:
> If a load instruction tagged with the !invariant.load metadata is executed, the memory location referenced by the load has to contain the same value at all points in the program where the memory location is dereferenceable; otherwise, the behavior is undefined.

I too concluded that we cannot do that; the GC might collect the object at one point, and later reuse the same memory for another object with another vptr, without manual emplace witch craft.

I guess it's worth doing this, so :+1: from me.

gen/toir.cpp
```diff
@@ -773,11 +773,48 @@ class ToElemVisitor : public Visitor {
...
+    //  auto saved_vtable = a.__vptr;     // emitted as part of `a.foo()`,
+    //                                    // except when e->directcall==true for
+    //                                    // final method calls.
+    //  a.foo();
+    //  assume(a.__vptr == saved_vtable); // <-- added assumption
+    // ```
+    // Only emit this extra code from -O2.
+    // This optimization is only valid for D class method calls (not C++).
+    if (dfnval && dfnval->vtable && (optLevel() >= 2) &&
+        dfnval->func->_linkage == LINK::d) {
```
kinke, line 807: This condition is very similar to the one above; might pay off to factor it out once.

PetarKirov:
> because D violates immutable in several places

Can you give some examples of this in `@safe` code? The only one that I can think of (except for object lifetimes, or generally un-`@safe` stuff like `union`s and casting) is `synchronized(obj)`, where `obj` is an `extern (D)` class. However, `synchronized` does not affect class hierarchies without the `__monitor` field, like `extern (C++)` and, hopefully in the not so distant future, these derived from the `ProtoObject` `extern(D)` base class.

> Moreover, we would have to take object lifetime into account (tricky)

I'm pretty sure that most production compiler toolchains must do this already since the beginning of time, as otherwise you can't do stack allocation (reuse the same memory for the locals of `a`, `b` and `c` in `void foo() { a(); b(); c(); }`). Are you saying that LDC can't take advantage of `immutable` for stack allocated POD just because the same addresses will be overridden by a future call frame?

kinke:
One example: immutable globals without static initializer:
```D
@safe:

immutable int bla;

shared static this() { bla = 42; }
```

PetarKirov:
> immutable globals without static initializer

True, though not observable after `_Dmain` is invoked, right? IIRC, one of the previous D releases made it an error to use non-`shared` static constructors to initialize immutable data, exactly because you could observe changes to `immutable` global variables after `_Dmain`.

kinke:
Yeah, not observable regularly after the runtime has been initialized, but still, the memory needs to be mutable, and CRT ctors etc. see/use the initial value.

JohanEngelen:
Here is an example where optimization performs an invalid transformation if we mark the vtable load as `invariant`:
```d
import core.lifetime;

class C {
    this(int i);
    int virtualGetI();
}

auto foo() {
    align(__traits(classInstanceAlignment, C)) byte[__traits(classInstanceSize, C)] buffer;
    C c = emplace!C(buffer[], 42);
    return c.virtualGetI();
}
```

LLVM IR with and without `!invariant.load`:  https://llvm.godbolt.org/z/fvshfq13a

The problem is that the memory `buffer` starts out with zeros, and then later on is filled with class `C` vtable etc. When the vtable ptr load is marked as `!invariant.load`, then we tell the optimizer that the memory never changes, so it assumes it is zeros (`null`) and then that the subsequent dereference is UB (leading to the `unreachable` result).

The solution for this may be to use `llvm.invariant.start`/`end` instead. However that is more codegen work, and where do we add the invariant.start? At the end of statement `C c =...` ?  I tested this before and it didn't quite work (optimization not done), but I'm not 100% sure.

PR #4598, opened 2024-03-19
Add missing `llvm.assume` intrinsic.

JohanEngelen:


JohanEngelen:
FYI:
```
    auto vtable = a.__vptr;
    a.foo();
    llvm_assume(vtable == a.__vptr);
```
works for one part of #4596  ;-)

PR #4599, opened 2024-03-23
tree-wide: port LDC to LLVM 18

liushuyu:
This pull request adds LLVM 18.1.x support to LDC.
LLVM recently switched to a new versioning scheme, where LLVM 18.1.0 was the first stable release of LLVM 18 series.

Aside from the usual API renaming, `zext` can not be used directly now and needs to be utilized through a constant folding mechanism.

Issue #4389, opened 2023-05-02
Fixes needed for newer GDB versions


JohanEngelen:
Tests `debuginfo/{classtypes,print}_gdb.d` fail with newer GDB versions, and are now disabled for FreeBSD. See e.g. https://github.com/ldc-developers/ldc/pull/4373

the-horo:
I'm hitting this on Gentoo as well:
```
LLD 17.0.6
Lit version:  17.0.6dev
-- Testing: 2 tests, 2 workers --
FAIL: LDC :: debuginfo/print_gdb.d (1 of 2)
******************** TEST 'LDC :: debuginfo/print_gdb.d' FAILED ********************
Script:
--
: 'RUN: at line 6';   /home/happy/git/ldc/build-ldc/bin/ldc2  -I/home/happy/git/ldc/tests/debuginfo -g -of=/home/happy/git/ldc/build-ldc/tests/debuginfo/Output/print_gdb.d.tmp /home/happy/git/ldc/tests/debuginfo/print_gdb.d /home/happy/git/ldc/tests/debuginfo/inputs/import_a.d /home/happy/git/ldc/tests/debuginfo/inputs/import_b.d
: 'RUN: at line 7';   sed -e "/^\\/\\/ GDB:/!d" -e "s,// GDB:,," /home/happy/git/ldc/tests/debuginfo/print_gdb.d >/home/happy/git/ldc/build-ldc/tests/debuginfo/Output/print_gdb.d.tmp.gdb
: 'RUN: at line 8';   env LANG=C gdb /home/happy/git/ldc/build-ldc/tests/debuginfo/Output/print_gdb.d.tmp --batch -x /home/happy/git/ldc/build-ldc/tests/debuginfo/Output/print_gdb.d.tmp.gdb >/home/happy/git/ldc/build-ldc/tests/debuginfo/Output/print_gdb.d.tmp.out 2>&1
: 'RUN: at line 9';   FileCheck /home/happy/git/ldc/tests/debuginfo/print_gdb.d -check-prefix=CHECK < /home/happy/git/ldc/build-ldc/tests/debuginfo/Output/print_gdb.d.tmp.out
--
Exit Code: 1

Command Output (stdout):
--
$ ":" "RUN: at line 6"
[...]

PR #4600, opened 2024-03-23
CI: re-organize GitHub Actions

liushuyu:
This pull request re-organizes GitHub Actions, where Linux and macOS tasks are now split.
This is needed because the Linux build environment uses Docker containers to minimize the impact of GitHub Actions adding/removing PPAs from their host VM systems.

In addition, Linux build containers switch to LLVM's upstream APT archive instead of downloading the binary files from GitHub Release.

A LLVM 18.1 build job is also added to the build matrix to support #4599.

### DIFF
--- a/.github/workflows/supported_llvm_versions.yml
+++ b/.github/workflows/supported_llvm_versions.yml
@@ -34,7 +34,7 @@ jobs:
             os: ubuntu-latest
             host_dc: ldc-beta
             llvm_version: 15
-            container: ldc-developers/build-environments:20.04-15
+            container: liushuyu/ldc-build-environments:20.04-15
           - job_name: macOS 11, LLVM 14, latest DMD beta
             os: macos-11
             host_dc: dmd-beta
@@ -44,19 +44,19 @@ jobs:
             os: ubuntu-latest
             host_dc: ldc-beta
             llvm_version: 13
-            container: ldc-developers/build-environments:20.04-13
+            container: liushuyu/ldc-build-environments:20.04-13
             cmake_flags: -DRT_SUPPORT_SANITIZERS=ON
           - job_name: Ubuntu 20.04, LLVM 12, bootstrap LDC
             os: ubuntu-latest
             host_dc: ldc-1.19.0
             llvm_version: 12
-            container: ldc-developers/build-environments:20.04-12
+            container: liushuyu/ldc-build-environments:20.04-12
             cmake_flags: -DBUILD_SHARED_LIBS=ON -DLIB_SUFFIX=64
           - job_name: Ubuntu 20.04, LLVM 11, latest DMD beta
             os: ubuntu-latest
             host_dc: dmd-beta
             llvm_version: 11
-            container: ldc-developers/build-environments:20.04-11
+            container: liushuyu/ldc-build-environments:20.04-11
             cmake_flags: -DBUILD_SHARED_LIBS=OFF -DRT_SUPPORT_SANITIZERS=ON -DLDC_LINK_MANUALLY=ON
     name: ${{ matrix.job_name }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


kinke:
> This is needed because the Linux build environment uses Docker containers to minimize the impact of GitHub Actions adding/removing PPAs from their host VM systems.

Which is a problem because of the LLVM apt repo now? - We've used that apt repo in the past, years ago, and it wasn't reliable IIRC - downtimes, as well as conflicts with the CI runner image from time to time [with whatever CI system we had back then, way before GitHub Actions came around].

The availability of binary LLVM packages from GitHub directly has always been a bit problematic, especially since we cannot use Ubuntu 22 due to debuginfo/gdb regressions (https://github.com/ldc-developers/ldc/issues/4389).

liushuyu:
> Which is a problem because of the LLVM apt repo now? - We've used that apt repo in the past, years ago, and it wasn't reliable IIRC - downtimes, as well as conflicts with the CI runner image from time to time [with whatever CI system we had back then, way before GitHub Actions came around].

We can instead make a prebuilt Docker container image so that apt.llvm.org availability problem will be avoided. Since it's inside a container, it won't be influenced by the host system and can be easily migrated if another CI system is used in the future.

> The availability of binary LLVM packages from GitHub directly has always been a bit problematic, especially since we cannot use Ubuntu 22 due to debuginfo/gdb regressions (#4389).

Has the issue been forwarded to the GDB upstream? If yes, was this issue fixed in a newer GDB release?

kinke:
> We can instead make a prebuilt Docker container image so that apt.llvm.org availability problem will be avoided.

I think that'd be a bit too too much effort. :) - What we have usually done so far is wait a bit for the new LLVM major to mature; then almost every time around there was at least one usable GitHub artifact among the various patch releases. For the recent LLVM versions without Ubuntu 20 packages, we've used macOS arm64 ones as you've seen. I expect one to pop up soon for v18.1.x too.

On rare occasions, we've been temporarily resorting to our LLVM fork and its artifacts. Those are normally only used for the `main.yml` GHA workflow and the LDC CI/release artifacts. The first prebuilt packages based on v18.1.* are available here (optionally with enabled assertions): https://github.com/ldc-developers/llvm-project/releases/tag/CI

The LLVM apt repo dependency aside, I'm not against a container-based approach in principle, but I definitely don't like the current duplication by splitting up the existing job into 2. There's probably a way to set something like `container: ${{ null }}` for the Mac jobs, to keep a single job definition. Possibly of interest: https://stackoverflow.com/questions/76925314/conditional-use-of-container-depending-on-the-group-it-runs-on-for-github-action

> Has the issue been forwarded to the GDB upstream? If yes, was this issue fixed in a newer GDB release?

Not that I know of, mainly because it's not clear whether it isn't our fault. The same tests seem to pass with the DMD compiler on Ubuntu 22, using its default gdb version.

liushuyu:
> I think that'd be a bit too much effort. :) - What we have usually done so far is wait a bit for the new LLVM major to mature; then almost every time around there was at least one usable GitHub artifact among the various patch releases. For the recent LLVM versions without Ubuntu 20 packages, we've used macOS arm64 ones as you've seen. I expect one to pop up soon for v18.1.x too.

I have already prepared a possible solution for making container images: https://github.com/liushuyu/ldc-build-environments.

> The LLVM apt repo dependency aside, I'm not against a container-based approach in principle, but I definitely don't like the current duplication by splitting up the existing job into 2. There's probably a way to set something like container: ${{ null }} for the Mac jobs, to keep a single job definition. Possibly of interest: https://stackoverflow.com/questions/76925314/conditional-use-of-container-depending-on-the-group-it-runs-on-for-github-action

If using prebuilt container images, we can merge the tasks back and skip all the setup steps for Linux.

liushuyu:
Seems like most of this pull request has been superseded. Closing.

PR #4618, opened 2024-04-13
Android: Switch to native ELF TLS & latest LTS NDK, and enable shared druntime/Phobos

### LINKED ISSUE #3918: [Android] Support the lld linker
The BFD linker has been removed since r23, so it isn't going to be an option going forward.

I've tried the following linker script to try and force the two sections together but the script causes the library to not be loadable:

```
SECTIONS
{
    .tdata : { *(.tdata .tdata.* .gnu.linkonce.td.*) }
    .tbss  : { *(.tbss .tbss.* .gnu.linkonce.tb.*) *(.tcommon) }
}
```

I've also tried looking at sections_android to see what would need to be changed to support two non continuous sections but it's a bit out of my field.

https://github.com/ldc-developers/druntime/blob/ldc/src/rt/sections_android.d#L183

It might be possible to just put everything in one section in the compiler?

Tried swapping the sections manually with a program but the program still crashes, there might be more needed to be done.

kinke:
Fixes #3918.

### DIFF
--- a/.github/actions/3-build-cross/action.yml
+++ b/.github/actions/3-build-cross/action.yml
@@ -115,7 +115,7 @@ runs:
         cmakeFlags+=" -DANDROID_USE_LEGACY_TOOLCHAIN_FILE=OFF"
         cmakeFlags+=" -DLDC_LINK_MANUALLY=ON -DD_LINKER_ARGS='-L$PWD/build-cross-libs/lib;-lphobos2-ldc;-ldruntime-ldc'"
 
-        echo "DFLAGS=-mtriple=$triple -fvisibility=hidden -L-L$PWD/build-cross-libs/lib -gcc=$PWD/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/$triple-clang" >> $GITHUB_ENV
+        echo "DFLAGS=-mtriple=$triple -L-L$PWD/build-cross-libs/lib -gcc=$PWD/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/$triple-clang" >> $GITHUB_ENV
         echo "CROSS_TRIPLE=$triple" >> $GITHUB_ENV
         echo "CROSS_CMAKE_FLAGS=$cmakeFlags" >> $GITHUB_ENV
 
--- a/.github/actions/3-build-cross/action.yml
+++ b/.github/actions/3-build-cross/action.yml
@@ -14,10 +14,10 @@ inputs:
     default: false
   android_ndk_version:
     required: false
-    default: r21e
+    default: r26d
   android_api_level:
     required: false
-    default: 21
+    default: 29
 runs:
   using: composite
   steps:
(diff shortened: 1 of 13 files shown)


kinke:
@s-ludwig @PetarKirov: Would you happen to be able and willing to check out the CI artifacts from https://github.com/ldc-developers/ldc/actions/runs/8676824079?pr=4618? E.g., if the bundled dub executable works fine to build some little app natively on Android, I'd expect things to be good (enough TLS coverage in dub and compiler itself).

If this works, we could most likely simply enable shared druntime+Phobos support for Android too, i.e., including the shared .so libs and allowing multiple D binaries in a single process, just like on vanilla Linux.

.github/actions/3-build-cross/action.yml
```diff
@@ -92,30 +92,28 @@ runs:
...
-
         arch='${{ inputs.arch }}'
         apiLevel='${{ inputs.android_api_level }}'
         cmakeFlags="-DTARGET_SYSTEM='Android;Linux;UNIX'"
         if [[ "$arch" == armv7a ]]; then
           triple="$arch-linux-androideabi$apiLevel"
           cmakeFlags+=' -DANDROID_ABI=armeabi-v7a'
         elif [[ "$arch" == aarch64 ]]; then
+          # FIXME: as of NDK rc26d, libc.a has __tls_get_addr, but libc.so only since API level 30 (Android v11)
+          apiLevel=30
```
kinke, line 108: It's a bit weird that the AArch64 `libc.so` got the required `__tls_get_addr` symbol in level 30, not level 29 as for armv7a; the NDK changelog doesn't mention this. Not sure whether this is an oversight in the NDK (it's like this for r26c and r26d), or whether there's a good reason why `__tls_get_addr` requires Android v11+ on AArch64 (unless linking libc statically).

s-ludwig:
> @s-ludwig @PetarKirov: Would you happen to be able and willing to check out the CI artifacts from https://github.com/ldc-developers/ldc/actions/runs/8676824079?pr=4618? E.g., if the bundled dub executable works fine to build some little app natively on Android, I'd expect things to be good (enough TLS coverage in dub and compiler itself).
> 
> If this works, we could most likely simply enable shared druntime+Phobos support for Android too, i.e., including the shared .so libs and allowing multiple D binaries in a single process, just like on vanilla Linux.

We've performed a quick test with a newly built runtime for NDK 25 and while it built successfully, for some reason it now hangs in a `pthread_once` call inside libcrypto:

Screenshot from 2024-04-14 12-52-09

kinke:
THx for checking. So you haven't tested the prebuilt libs, but built them yourself? Using a prebuilt compiler here, or some other existing compiler? The former is required, as the main change is in the used LLVM (not visible in this diff here).

s-ludwig:
I've used the compiler binary from the linux-x86_64 CI build, but generating an android runtime since none is included there. Should I try again using the libraries from the android-aarch64 build?

kinke:
> Should I try again using the libraries from the android-aarch64 build?

Might be worth a try, but I wouldn't expect wonders. Ideally with the same NDK r26d? - No idea how you built the libs exactly; I've dropped the `-fvisibility=hidden` here (for the shared libs) and use the level 29/30 clang for cross-compilation & -linking (and the same level for LDC's `-mtriple`).

Some TLS malfunction resulting in a hanging libcrypto init sounds a bit weird/unlucky though IMO...

kinke:
Shall we get this in for beta1 to gather more feedback, reverting it for the final if problematic?

s-ludwig:
I've just tested again, this time with the latest CI artifacts, using the compiler binary from the linux-x86_64 artifact and the libraries from the android-aarch64 artifact, using NDK 26.3.11579264. Unfortunately, this results in exactly the same hang. I can try to generate a more minimal JNI library and specifically test TLS to hopefully narrow this down a bit.

s-ludwig:
Okay, basic TLS seems to work fine (setting an integer TLS variable, starting a setting/manipulating the same variable concurrently).

kinke:
Oh wow, awesome news, thx!

Issue #4626, opened 2024-04-21
Cross-compilation targets aren't respected when preprocessing C files with `clang-cl` (on Windows).


just-harry:
`clang-cl` doesn't seem to respect the usual MSVC environment, so by default, files are preprocessed with the builtin defines for the host architecture rather than the target architecture.

E.g. with an MSVC environment set-up for compiling for AArch64 from x86-64:
```powershell
PS> "$Env:VSCMD_ARG_HOST_ARCH -> $Env:VSCMD_ARG_TGT_ARCH"
x64 -> arm64
```

using `clang-cl` without a target gives us the defines for x86-64,
```powershell
PS> '' | clang-cl -E -Xclang -dD - | ? {$_ -cmatch 'ARM64|AMD64'}
#define _M_AMD64 100
```

whereas if a target triple is specified we get the expected defines for AArch64,
```powershell
PS> '' | clang-cl --target=aarch64-windows -E -Xclang -dD - | ? {$_ -cmatch 'ARM64|AMD64'}
#define _M_ARM64 1
```

`cl` of course does respect the MSVC environment, so we get the expected defines for AArch64 from it,
```powershell
PS> cl /nologo /EP /PD /Zc:preprocessor (New-TemporaryFile) | ? {$_ -cmatch 'ARM64|AMD64'}
tmp03pm53.tmp
#define _M_ARM64_EXTENSION 80
#define _M_ARM64 1
```

Ideally, LDC would propagate the target to `clang-cl` when using it to preprocess C.

just-harry:
A proposed fix: https://github.com/ldc-developers/ldc/pull/4627

PR #4627, opened 2024-04-21
Propagate the target to `clang-cl` when it's used to preprocess C.

just-harry:
This PR propagates the target to `clang-cl` when it's used to preprocess C, addressing https://github.com/ldc-developers/ldc/issues/4626.

`clang-cl` doesn't seem to support `-mcpu=` nor `-mtune=`, and `-march=` can't be used in their place for AArch64/ARM targets.

So, to ensure the features of LDC's targeted CPU are propagated, every processor feature gets forwarded on via the `-target-feature` option—the downside of this being that it generates a massive command-line: the longest being 11,691 characters just for the target-features when targeting AArch64, which is a non-negligible contribution to Windows' limit of 32,766 UTF-16 code-units for a command-line, but there are only four paths in the command so I think it should be fine.

### DIFF
--- a/driver/cpreprocessor.cpp
+++ b/driver/cpreprocessor.cpp
@@ -4,9 +4,12 @@
 #include "driver/cl_options.h"
 #include "driver/timetrace.h"
 #include "driver/tool.h"
+#include "gen/irstate.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/Program.h"
+#include "llvm/MC/MCSubtargetInfo.h"
+#include "llvm/Target/TargetMachine.h"
 
 namespace {
 const char *getPathToImportc_h(const Loc &loc) {
@@ -119,6 +122,27 @@ FileName runCPreprocessor(FileName csrcfile, const Loc &loc,
       args.push_back("/PD");              // print all macro definitions
       args.push_back("/Zc:preprocessor"); // use the new conforming preprocessor
     } else {
+      // propagate the target to the preprocessor
+      args.push_back("-target");
+      args.push_back(triple.getTriple());
+
+#if LDC_LLVM_VER >= 1800 // getAllProcessorFeatures was introduced in this version
+      // propagate all enabled/disabled features to the preprocessor
+      const auto &subTarget = gTargetMachine->getMCSubtargetInfo();
+      const auto &featureBits = subTarget->getFeatureBits();
+      llvm::SmallString<64> featureString;
+      for (const auto &feature : subTarget->getAllProcessorFeatures()) {
+        args.push_back("-Xclang");
+        args.push_back("-target-feature");
+        args.push_back("-Xclang");
+
+        featureString += featureBits.test(feature.Value) ? '+' : '-';
+        featureString += feature.Key;
+        args.push_back(featureString.str().str());
+        featureString.clear();
+      }
+#endif
+
       // print macro definitions (clang-cl doesn't support /PD - use clang's
       // -dD)
       args.push_back("-Xclang");
--- a/tests/driver/clang-cl-target-forwarding.d
+++ b/tests/driver/clang-cl-target-forwarding.d
@@ -1,0 +1,17 @@
+// On Windows, either clang-cl or cl is used to preprocess C.
+// If clang-cl is used, LDC's target triple and any target features should be
+// passed to clang-cl.
+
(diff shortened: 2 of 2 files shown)


JohanEngelen:
Can you add a testcase in `tests/driver` for this?  (both -target and the features)

About the features: note that features are not passed either for other OS/systems. But I'm OK with passing it here because we know we are passing into clang-cl.

just-harry:
@JohanEngelen, thank you, that's a test-case added now.

I've also wrapped the `target-feature` loop in an `#if LDC_LLVM_VER >= 1800` as that's when `MCSubtargetInfo.getAllProcessorFeatures` was introduced.

kinke:
> About the features: note that features are not passed either for other OS/systems. But I'm OK with passing it here because we know we are passing into clang-cl.

Exactly what I'm thinking.

> the downside of this being that it generates a massive command-line: the longest being 11,691 characters just for the target-features when targeting AArch64, which is a non-negligible contribution to Windows' limit of 32,766 UTF-16 code-units for a command-line

The ridiculous Windows limit is already handled in our `executeToolAndWait()` helper, resorting to a response file.

driver/cpreprocessor.cpp
```diff
@@ -4,9 +4,12 @@
 #include "driver/cl_options.h"
 #include "driver/timetrace.h"
 #include "driver/tool.h"
+#include "gen/irstate.h"
```
kinke, line 7: Seems superfluous.

driver/cpreprocessor.cpp
```diff
@@ -119,6 +122,27 @@ FileName runCPreprocessor(FileName csrcfile, const Loc &loc,
...
+      const auto &featureBits = subTarget->getFeatureBits();
+      llvm::SmallString<64> featureString;
+      for (const auto &feature : subTarget->getAllProcessorFeatures()) {
+        args.push_back("-Xclang");
+        args.push_back("-target-feature");
+        args.push_back("-Xclang");
+
+        featureString += featureBits.test(feature.Value) ? '+' : '-';
+        featureString += feature.Key;
+        args.push_back((std::string)featureString);
```
kinke, line 141: Please no ugly C-style casts; there's a `.str()` method IIRC.

tests/driver/clang-cl-target-forwarding.d
```diff
@@ -0,0 +1,17 @@
+// On Windows, either clang-cl or cl is used to preprocess C.
+// If clang-cl is used, LDC's target triple and any target features should be
+// passed to clang-cl.
+
+// REQUIRES: Windows
```
kinke, line 5: The test additionally requires `target_X86 && target_AArch64`.

driver/cpreprocessor.cpp
just-harry: That's where `gTargetMachine` is coming from

driver/cpreprocessor.cpp
just-harry: Gotcha—wasn't sure if `.str().str()` was better or worse

tests/driver/clang-cl-target-forwarding.d
just-harry: Thanks; they're added now

kinke (approved):

kinke:
Thx mate. What you could do to reduce the command-line ugliness is joining all features into a single argument, like `-Xclang -target-feature -Xclang +sse2,-sse4,…`.

just-harry:
> Thx mate.

:)
> What you could do to reduce the command-line ugliness is joining all features into a single argument, like `-Xclang -target-feature -Xclang +sse2,-sse4,…`.

That's what I tried at first, but sadly it doesn't work:

```powershell
PS> '' | clang-cl /nologo -E -target x86_64-pc-windows-msvc -Xclang -target-feature -Xclang +sse4a -Xclang -dD -Wno-builtin-macro-redefined - | rg SSE4
#define __SSE4A__ 1
```

```powershell
PS> '' | clang-cl /nologo -E -target x86_64-pc-windows-msvc -Xclang -target-feature -Xclang +sse4a,+sse4 -Xclang -dD -Wno-builtin-macro-redefined - | rg SSE4
```

kinke:
Oh right, I see the same behavior on cpp.godbolt.org (https://cpp.godbolt.org/z/3GvqseoTn), wild - it seems to be fine for the generated IR, but not for preprocessing.

Issue #4630, opened 2024-04-22
LDC sets both never-inline and always-inline on inline functions containing DMD-style inline assembly.


just-harry:
If a D function has `pragma(inline, true)` applied to it, or a C function is `inline`, and the function contains DMD-style inline assembly, LDC will end up setting both always-inline and never-inline on the function, causing an error during compilation.

I think it would be better for the inline attribute to be ignored if a function contains DMD-style inline assembly, like DMD does, if only to allow third-party code with the conflicting attributes to be compiled. (Examples of such code include `Int64ShllMod32` in `winnt.h`, which is included by `windows.h`)

just-harry:
A proposed fix: https://github.com/ldc-developers/ldc/pull/4631

kinke:
> If a D function has pragma(inline, true) applied to it, or a C function is inline

Ideally, we'd be able to discriminate these 2 cases, and error out properly for an invalid `pragma(inline, true)` (upstream, in the frontend directly). But if importC translates `inline` to the same AST representation, then that's firstly not really correct, but probably works in most cases, and so I agree that we should just ignore it.

There's a `Dsymbol::isCsymbol()` method that we could use to restrict this 'hack' to functions in C files only.

just-harry:
> > If a D function has pragma(inline, true) applied to it, or a C function is inline
> 
> Ideally, we'd be able to discriminate these 2 cases, and error out properly for an invalid `pragma(inline, true)` (upstream, in the frontend directly). But if importC translates `inline` to the same AST representation, then that's firstly not really correct, but probably works in most cases, and so I agree that we should just ignore it.
> 
> There's a `Dsymbol::isCsymbol()` method that we could use to restrict this 'hack' to functions in C files only.

Mhm, ImportC just sets `FuncDeclaration.inlining` to `PINLINE.always` for `inline` functions (though it doesn't emit a `PragmaDeclaration`).

I've changed https://github.com/ldc-developers/ldc/pull/4631 to ignore always-inline when DMD-style inline assembly is present only for C functions.

PR #4631, opened 2024-04-22
Avoid setting always-inline for functions that contain DMD-style inline assembly.

just-harry:
This PR makes it so the inline attribute is ignored for functions which contain DMD-style inline assembly (as such functions are always made never-inline), addressing issue https://github.com/ldc-developers/ldc/issues/4630.

This is what DMD does, and this makes it possible to compile third-party code which has inline functions with DMD-style inline assembly (e.g. `windows.h`).

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -654,7 +654,22 @@ void DtoDeclareFunction(FuncDeclaration *fdecl, const bool willDefine) {
     irFunc->setNeverInline();
   } else {
     if (fdecl->inlining == PINLINE::always) {
-      irFunc->setAlwaysInline();
+      // If the function contains DMD-style inline assembly.
+      if (fdecl->hasReturnExp & 32) {
+        // The presence of DMD-style inline assembly in a function causes that
+        // function to become never-inline. So, if this function contains DMD-style
+        // inline assembly we'll emit an error as it can't be made always-inline.
+        // However, we'll make an exception for C functions, as the C standard doesn't
+        // actually require that `inline` functions be inlined. So, for C functions we just
+        // ignore the attempt to make it always-inline.
+        if (!fdecl->isCsymbol()) {
+          error(fdecl->loc,
+                "`%s` cannot be `pragma(inline, true)` as it contains DMD-style inline assembly",
+                fdecl->toPrettyChars());
+        }
+      } else {
+        irFunc->setAlwaysInline();
+      }
     } else if (fdecl->inlining == PINLINE::never) {
       irFunc->setNeverInline();
     }
--- a/tests/codegen/dmd_style_asm_in_inline_c_function.c
+++ b/tests/codegen/dmd_style_asm_in_inline_c_function.c
@@ -1,0 +1,15 @@
+// ImportC treats `inline` C functions as though `pragma(inline, true)` was applied to them.
+// The presence of DMD-style inline assembly inhibits inlining,
+// so if `pragma(inline, true)` is also present, an error should occur; but C doesn't
+// require that an `inline` function is actually inlined, so we permit `inline` C functions
+// to contain DMD-style inline assembly,
+
+// REQUIRES: Windows && target_X86
+// RUN: %ldc -mtriple=i686-pc-windows-msvc -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+// CHECK-NOT: alwaysinline
+inline unsigned int hasDMDStyleAsm(unsigned int a) {
+    asm {
+        mov EAX, dword ptr [a];
+    }
+}
--- a/tests/codegen/inputs/inlinables_asm.d
+++ b/tests/codegen/inputs/inlinables_asm.d
@@ -1,14 +1,11 @@
 module inputs.inlinables_asm;
 
 import ldc.attributes;
+import ldc.llvmasm;
 
 extern (C): // simplify mangling for easier function name matching
 
-pragma(inline, true) extern (C) void naked_asm_func()
+pragma(inline, true) extern (C) @naked void naked_asm_func()
 {
-    asm pure nothrow @nogc
-    {
-        naked;
-        nop;
-    }
+	return __asm("nop", "");
 }
--- a/tests/fail_compilation/dmd_style_asm_in_inline_d_function.d
+++ b/tests/fail_compilation/dmd_style_asm_in_inline_d_function.d
@@ -1,0 +1,17 @@
+// The presence of DMD-style inline assembly inhibits inlining,
+// so if `pragma(inline, true)` is also present, some kind of error should occur.
+
+// REQUIRES: target_X86
+// RUN: not %ldc -mtriple=i686-pc-windows-msvc -c %s 2>&1 | FileCheck %s
+
+module dmd_style_asm_in_inline_d_function;
+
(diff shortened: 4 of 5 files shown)


tests/compilable/asm_in_inline_function.d
```diff
@@ -0,0 +1,13 @@
+// DMD ignores any always-inline style pragmas for functions containing inline assembly.
+
+// REQUIRES: Windows && target_X86
+// RUN: %ldc -mtriple=i686-pc-windows-msvc -c %s
```
thewilsonator, line 4: please make this a codegen test and  change this to 
```
RUN: %ldc -mtriple=i686-pc-windows-msvc -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
```
and add 

```
// CHECK-NOT: always-inline
```
(I think I have that right)

just-harry:
I changed this so that the inline attribute is ignored only for C functions which contain DMD-style inline assembly, [per Martin's suggestion](https://github.com/ldc-developers/ldc/issues/4630#issuecomment-2071963380)

tests/codegen/dmd_style_asm_in_inline_c_function.c
```diff
@@ -0,0 +1,15 @@
+// ImportC treats `inline` C functions as though `pragma(inline, true)` was applied to them.
+// The presence of DMD-style inline assembly inhibits inlining,
+// so if `pragma(inline, true)` is also present, an error should occur; but C doesn't
+// require that an `inline` function is actually inlined, so we permit `inline` C functions
+// to contain DMD-style inline assembly,
+
+// REQUIRES: Windows && target_X86
```
JohanEngelen, line 7: the test works on all OS-es (LDC is crosscompiler), so remove `Windows`

tests/fail_compilation/dmd_style_asm_in_inline_d_function.d
```diff
@@ -0,0 +1,15 @@
+// The presence of DMD-style inline assembly inhibits inlining,
+// so if `pragma(inline, true)` is also present, some kind of error should occur.
+
+// REQUIRES: Windows && target_X86
```
JohanEngelen, line 4: the test works on all OS-es (LDC is crosscompiler), so remove `Windows`

tests/codegen/dmd_style_asm_in_inline_c_function.c
just-harry: Done now

tests/fail_compilation/dmd_style_asm_in_inline_d_function.d
just-harry: This one too

tests/codegen/dmd_style_asm_in_inline_c_function.c
kinke: This is a .c file needing `clang-cl` as cross-C-preprocessor, so I don't think it works.

just-harry: Oh, right, I forgot about that style of asm being Microsoft-specific. It's added back now.

tests/fail_compilation/dmd_style_asm_in_inline_d_function.d
```diff
@@ -0,0 +1,15 @@
+// The presence of DMD-style inline assembly inhibits inlining,
+// so if `pragma(inline, true)` is also present, some kind of error should occur.
+
+// REQUIRES: target_X86
+// RUN: not %ldc -mtriple=i686-pc-windows-msvc -c %s 2>&1 | FileCheck %s
+
+// CHECK: {{function can't be never- and always-inline at the same time|noinline.+alwaysinline|alwaysinline.+noinline}}
```
thewilsonator, line 7: is the fact that asm functions are never inlined (i.e.  `pragma(inline, false)` part of that message?

kinke: I don't find it in the spec, but DMD apparently ignores the pragma and never inlines a function with inline asm. I thought that's guaranteed for those functions; I think lots of those in druntime/Phobos rely on arguments in registers and/or their private RBP/RSP. For LDC, there's no technical reason to disable it for non-naked functions, except for these existing usages.

I'm in favor of making this a proper error (for D usages), not depending on our assertions or the LLVM IR verifier. Preferably upstream, but we could add it here in `DtoDeclareFunction()` already in the meantime (via the free-standing `error()` function).

just-harry: That makes sense to me; I've changed this so that it emits an error when attempting to make non-C functions, which contain DMD-style inline assembly, always-inline.

gen/functions.cpp
```diff
@@ -654,7 +654,22 @@ void DtoDeclareFunction(FuncDeclaration *fdecl, const bool willDefine) {
...
+      if (fdecl->hasReturnExp & 32) {
+        // The presence of DMD-style inline assembly in a function causes that
+        // function to become never-inline. So, if this function contains DMD-style
+        // inline assembly we'll emit an error as it can't be made always-inline.
+        // However, we'll make an exception for C functions, as the C standard doesn't
+        // actually require that `inline` functions be inlined. So, for C functions we just
+        // ignore the attempt to make it always-inline.
+        if (!fdecl->isCsymbol()) {
+          error(fdecl->loc,
+                "`%s` cannot be `pragma(inline, true)` as it contains DMD-style inline assembly.",
```
kinke, line 667: [we usually don't terminate error messages with a `.`]

tests/codegen/inputs/inlinables_asm.d
```diff
@@ -1,14 +1,11 @@
...
 
-pragma(inline, true) extern (C) void naked_asm_func()
+pragma(inline, true) extern (C) @naked void naked_asm_func()
 {
-    asm pure nothrow @nogc
-    {
-        naked;
-        nop;
-    }
+	return __asm("nop", "");
```
kinke, line 10: [We might need to emphasize this a bit more on the Wiki etc., but LDC also supports gcc/gdc-style assembly, which doesn't require an extra import and works with gdc too: `asm { "nop"; }`.]

tests/codegen/dmd_style_asm_in_inline_c_function.c
```diff
@@ -0,0 +1,15 @@
...
+// RUN: %ldc -mtriple=i686-pc-windows-msvc -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+// CHECK-NOT: always-inline
```
kinke, line 10: `alwaysinline` without dash, as used in the other test?

gen/functions.cpp
just-harry: Noted; I've got rid of it

tests/codegen/dmd_style_asm_in_inline_c_function.c
just-harry: You're right, thanks for catching that

tests/codegen/inputs/inlinables_asm.d
just-harry: I did actually use the GDC-style when I first changed these tests—but then I thought that since one of the test functions is named `hasLDCStyleAsm` it'd better to use the LDC-specific asm rather than the GDC-style, so I then changed them to use `__asm` 🙃
Will I change it back?

(As a side-note, as an end-user, it's very much appreciated that LDC goes to the effort of supporting all the asm styles of D)

PR #4632, opened 2024-04-22
Fix issue #2257: LLVM does not accept generated inline assembly with variable+offset

just-harry:
This is a bodge.

See issue https://github.com/ldc-developers/ldc/issues/2257 and the comments in the code for more context.
But, in short: where now we generate `4+$0` which would expand to `4+someGlobalVariable`, or `4+16(%ebp)`, or `4+(%ebp)` (invalid), we instead generate `4-8+${0:H}` which would expand to `4-8+8+someGlobalVariable`, or `4-8+24(%ebp)`, or `4-8+8(%ebp)` (valid).

The `H` modifier applies the +8 offset unconditionally: https://github.com/llvm/llvm-project/blob/330d8983d25d08580fc1642fea48b2473f47a9da/llvm/lib/Target/X86/X86AsmPrinter.cpp#L405-L406

### DIFF
--- a/tests/codegen/dmd_style_asm_with_variable_and_offset_memory_reference.d
+++ b/tests/codegen/dmd_style_asm_with_variable_and_offset_memory_reference.d
@@ -1,0 +1,42 @@
+// REQUIRES: target_X86
+
+// RUN: %ldc -mtriple=i686-pc-windows-msvc -O -output-s -of=%t.s %s && FileCheck %s < %t.s
+
+uint getHighHalfWithoutDisplacement(ulong value)
+{
+    asm
+    {
+        // CHECK: movl    4(%esp), %eax
+        mov EAX, dword ptr [value + 4];
+    }
+}
+
+uint getHighHalfWithDisplacement(uint value1, uint value2)
+{
+    asm
+    {
+        // CHECK: movl -2(%ebp), %eax
+        mov EAX, word ptr 2[value1];
+    }
+}
+
+extern(C) __gshared ulong someGlobalVariable;
+
+uint getHighHalfOfGlobal(ulong value)
+{
+    asm
+    {
+        // CHECK: movl    ((4+(-8))+_someGlobalVariable)+8, %eax
+        mov EAX, dword ptr [someGlobalVariable + 4];
+    }
+}
+
+void foo()
+{
+    align(32) uint[4] a;
+    asm pure nothrow @nogc
+    {
+        // CHECK: movl  %ebx, 4(%esp)
+        mov a+4, EBX;
+    }
+}
--- a/gen/asm-x86.h
+++ b/gen/asm-x86.h
@@ -2861,6 +2861,9 @@ struct AsmProcessor {
       }
       operand = &operands[i];
 
+      // This is used only in the case of `operand->cls` being `Opr_Mem`.
+      bool hasConstDisplacement = false;
+
       switch (operand->cls) {
       case Opr_Immediate:
         // for implementing offset:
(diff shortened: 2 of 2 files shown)


tests/codegen/dmd_style_asm_with_variable_and_offset_memory_reference.d
```diff
@@ -0,0 +1,38 @@
...
+// RUN: %ldc -mtriple=i686-pc-windows-msvc -O -output-ll -of=%t.ll %s
+// RUN: FileCheck %s --check-prefix IR < %t.ll
+// RUN: %ldc -mtriple=i686-pc-windows-msvc -O -output-s -of=%t.s %s
+// RUN: FileCheck %s --check-prefix ASM < %t.s
+
+uint getHighHalfWithoutDisplacement(ulong value)
+{
+    asm
+    {
+        // IR: movl 4+-8+${1:H}, %eax
```
JohanEngelen, line 12: I wonder if you really want an IR test for this. If we find a less hacky fix in the future (in IR), then this test would fail / require adjustment.  Shouldn't the ASM test be sufficient?

tests/codegen/dmd_style_asm_with_variable_and_offset_memory_reference.d
```diff
@@ -0,0 +1,38 @@
...
+
+uint getHighHalfOfGlobal(ulong value)
+{
+    asm
+    {
+        // IR: movl 4+-8+${1:H}, %eax
+        // ASM: movl    ((4+(-8))+_someGlobalVariable)+8, %eax
+        mov EAX, dword ptr [someGlobalVariable + 4];
+    }
+}
```
JohanEngelen, line 38: Can you add the testcase of the original bug report #2257 ?

tests/codegen/dmd_style_asm_with_variable_and_offset_memory_reference.d
just-harry: I think the ASM test only should be sufficient, yeah.
I'll remove the IR test.
(Though, `CHECK: movl ((4+(-8))+_someGlobalVariable)+8, %eax` will also fail if a more proper fix is found)

tests/codegen/dmd_style_asm_with_variable_and_offset_memory_reference.d
just-harry: Can do; it's pushed up now

kinke:
For stuff like this, it's often interesting to see which IR clang generates: https://cpp.godbolt.org/z/T154WjToj

```C++
unsigned long long someGlobalVariable;

unsigned getHighHalfOfGlobal(unsigned long long value)
{
    __asm
    {
        mov EAX, dword ptr [someGlobalVariable + 4];
    }
}
```

=>

```LL
%0 = call i32 asm sideeffect inteldialect "mov EAX, dword ptr $1[$$4]", "=&{eax},*m,~{dirflag},~{fpsr},~{flags}"(ptr elementtype(i64) @"?someGlobalVariable@@3_KA")
```

JohanEngelen:
LGTM

nitpick: could add this link to the comment in the code: https://lists.llvm.org/pipermail/llvm-dev/2017-August/116244.html

just-harry:
> nitpick: could add this link to the comment in the code: https://lists.llvm.org/pipermail/llvm-dev/2017-August/116244.html

More context is always good for this sort of thing—it's in the comment now

kinke (approved):

kinke:
Thx, nice work!

PR #4634, opened 2024-04-29
Define CppRuntime_Clang when building for Android

s-ludwig:
With the current switch to NDK 26+ it should now be safe to assume the Clang C++ mangling/runtime.

The background for adding this is that the BindBC bindings currently error out when building for Android, because they require a C++ runtime version in order to potentially perform some manual symbol mangling.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -852,6 +852,7 @@ void registerPredefinedTargetVersions() {
     if (triple.getEnvironment() == llvm::Triple::Android) {
       VersionCondition::addPredefinedGlobalIdent("Android");
       VersionCondition::addPredefinedGlobalIdent("CRuntime_Bionic");
+      VersionCondition::addPredefinedGlobalIdent("CppRuntime_Clang");
     } else if (triple.isMusl()) {
       VersionCondition::addPredefinedGlobalIdent("CRuntime_Musl");
       VersionCondition::addPredefinedGlobalIdent("CppRuntime_Gcc");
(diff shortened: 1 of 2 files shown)


s-ludwig:
> ld.lld: error: duplicate symbol: std::bad_alloc::bad_alloc()
> defined at new_.d
>            new_.o:(std::bad_alloc::bad_alloc()) in archive /home/runner/work/ldc/build-cross-libs/lib/libdruntime-ldc.a
> defined at stdlib_exception.cpp
>            stdlib_exception.cpp.o:(std::bad_alloc::bad_alloc()) in archive /home/runner/work/ldc/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/arm-linux-androideabi/libc++abi.a

How is this supposed to work? It seems like this is always defined on the D side, but for some reason errors out for the Android libc++ - should this be changed in DRuntime to only *declare* the constructor?

kinke:
> How is this supposed to work? It seems like this is always defined on the D side, but for some reason errors out for the Android libc++ - should this be changed in DRuntime to only declare the constructor?

The ctor should most likely be marked `extern(D)`, as done for the other `exception`derived classes in https://github.com/dlang/dmd/blob/b84fa4709f2061b60c2fd398a01964370cbb12c8/druntime/src/core/stdcpp/typeinfo.d#L185-L199.

kinke (approved):

PR #4635, opened 2024-04-29
Add --template-codegen-depth to stop very deep template instantiations.

JohanEngelen:
We've had this at Weka for a long time. It is an escape hatch in case very very deep template instantiation leads to stack exhaustion in the recursive `needsCodegen` function.


kinke:
Is this hack still required at Weka? The `needsCodegen()` stack pressure should have been improved significantly for a while now, in https://github.com/dlang/dmd/pull/14787.

JohanEngelen:
I'll check whether it is needed with newer compilers (weka is currently still at 1.30)

PR #4639, opened 2024-05-03
Fix libunwind backtraces & Make defaulttraceinfo more generic.

JohanEngelen:
Libunwind-based druntime implementation for backtraces results in wrong line numbers (very confusing backtrace). 

1. Make DefaultTraceInfo smart enough to not strictly require execinfo.
2. ~~Stop using libunwind for musl.~~  Fix libunwind implementation

### DIFF
--- a/runtime/druntime/src/core/internal/backtrace/libunwind.d
+++ b/runtime/druntime/src/core/internal/backtrace/libunwind.d
@@ -43,6 +43,7 @@ nothrow:
  * Bindings for libunwind.h
  */
 alias unw_word_t = uintptr_t;
+alias unw_regnum_t = int;
 
 ///
 struct unw_context_t
@@ -86,6 +87,14 @@ int unw_step(unw_cursor_t*);
 int unw_get_proc_info(unw_cursor_t*, unw_proc_info_t*);
 /// Get the name of the current procedure (function)
 int unw_get_proc_name(unw_cursor_t*, char*, size_t, unw_word_t*);
+/// Reads the value of register `reg` in the stack frame identified by cursor `cp` and stores the value in the word pointed to by `valp`.
+int unw_get_reg(unw_cursor_t *cp, unw_regnum_t reg, unw_word_t *valp);
+/// Architecture independent register numbers
+enum
+{
+    UNW_REG_IP = -1, // instruction pointer
+    UNW_REG_SP = -2, // stack pointer
+}
 
 private:
 
--- a/runtime/druntime/src/core/internal/backtrace/libunwind.d
+++ b/runtime/druntime/src/core/internal/backtrace/libunwind.d
@@ -96,6 +96,21 @@ enum
     UNW_REG_SP = -2, // stack pointer
 }
 
+enum
+{
+    UNW_ESUCCESS      = 0,     /* no error */
+    UNW_EUNSPEC       = -6540, /* unspecified (general) error */
+    UNW_ENOMEM        = -6541, /* out of memory */
+    UNW_EBADREG       = -6542, /* bad register number */
+    UNW_EREADONLYREG  = -6543, /* attempt to write read-only register */
+    UNW_ESTOPUNWIND   = -6544, /* stop unwinding */
+    UNW_EINVALIDIP    = -6545, /* invalid IP */
+    UNW_EBADFRAME     = -6546, /* bad frame */
+    UNW_EINVAL        = -6547, /* unsupported operation or bad value */
+    UNW_EBADVERSION   = -6548, /* unwind info has unsupported version */
+    UNW_ENOINFO       = -6549  /* no unwind info found */
+}
+
 private:
 
 // The API between libunwind and llvm-libunwind is almost the same,
--- a/runtime/druntime/src/core/internal/backtrace/libunwind.d
+++ b/runtime/druntime/src/core/internal/backtrace/libunwind.d
@@ -96,21 +96,6 @@ enum
     UNW_REG_SP = -2, // stack pointer
 }
 
-enum
-{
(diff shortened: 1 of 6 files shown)


JohanEngelen:
@Geod24 The libunwind implementation needs a rework. The testcase that shows broken behavior clearly is `dmd/runnable/test17559.d`.
On line 70, `frame.address += pip.start_ip;` is where the backtrace address for symbol:line resolution is stored. But `frame.address` is zero there, so it will be loaded with the function's entry address. Not so useful. Instead we should use something like `unw_get_reg(&cursor, UNW_REG_IP, &ip);` to get the address of the return address (after the function call instruction).

JohanEngelen:
if it works here for LDC CI, I'll submit it upstream.

JohanEngelen:
I first had it working with importC, but the LLVM header of libunwind is pretty small so let's use the existing hard-coded binaries (easy to check that they are correct, and the additions needed are arch independent).

kinke:
Nice that you're hunting down the musl issues! :+1: - How's it looking, many failures still remaining or close to the finish line? Just wrt. v1.38.0 final that I'd like to release this weekend, unless you are close and require a bit more time.

JohanEngelen:
Let's not wait for the musl work (yesterday I ran stdlib tests for the first time, many succeed, but e.g. fiber tests hung and I have not looked into that at all yet). I'm hoping to reach the endpoint with one Github Action that builds using an alpine container (quick to set up in linux), such that we have a fully statically linked musl release.

Geod24:
Indeed, thanks for doing this! I haven't personally needed Musl for a few years so never got around to fixing the more tedious issues, but I do remember backtraces not being working fully...

JohanEngelen:
PR is fully finished. Feel free to *squash* and merge, or postpone until after 1.38 if deemed too risky.

runtime/druntime/src/core/runtime.d
```diff
@@ -901,30 +901,99 @@ else static if (hasExecinfo) private class DefaultTraceInfo : Throwable.TraceInf
...
-                ret = dg( pos, buf );
-                if ( ret )
-                    break;
+                // Poor man solution. Does not show line numbers, but does (potentially) show a backtrace of function names.
+
+                int ret = 0;
+                bool throw_call_found = false;
+                for (size_t i = 0; i < numframes; ++i)
+                {
+                    char[4096] fixbuf = void;
```
kinke, line 975: unused?

runtime/druntime/src/core/internal/backtrace/libunwind.d
```diff
@@ -86,6 +87,14 @@ int unw_step(unw_cursor_t*);
 int unw_get_proc_info(unw_cursor_t*, unw_proc_info_t*);
 /// Get the name of the current procedure (function)
 int unw_get_proc_name(unw_cursor_t*, char*, size_t, unw_word_t*);
+/// Reads the value of register `reg` in the stack frame identified by cursor `cp` and stores the value in the word pointed to by `valp`.
+int unw_get_reg(unw_cursor_t *cp, unw_regnum_t reg, unw_word_t *valp);
```
kinke, line 91: [C-style pointers]

JohanEngelen: hah, nice catch. C copy paste ;)

runtime/druntime/src/core/runtime.d
JohanEngelen: very good catch thanks

JohanEngelen:
> PR is fully finished. Feel free to _squash_ and merge, or postpone until after 1.38 if deemed too risky.

there is still something wrong with backtraces.  One bug I know (need to adjust backtrace address to point to call instruction instead of right after call instruction), but then there is still something buggy giving slightly incorrect backtrace on throwing an exception.

JohanEngelen:
> Nice that you're hunting down the musl issues! 👍 - How's it looking, many failures still remaining or close to the finish line? Just wrt. v1.38.0 final that I'd like to release this weekend, unless you are close and require a bit more time.

To see current status:
https://github.com/ldc-developers/ldc/actions/runs/8976987555/job/24654919738

JohanEngelen:
> > PR is fully finished. Feel free to _squash_ and merge, or postpone until after 1.38 if deemed too risky.
> 
> there is still something wrong with backtraces. One bug I know (need to adjust backtrace address to point to call instruction instead of right after call instruction), but then there is still something buggy giving slightly incorrect backtrace on throwing an exception.

This is fixed now (needed to adjust IP to point to the call instruction, rather than after the call).

runtime/druntime/src/core/runtime.d
```diff
@@ -901,30 +901,98 @@ else static if (hasExecinfo) private class DefaultTraceInfo : Throwable.TraceInf
...
+                    //       inside core.runtime and the object code, so eliminate
+                    //       these for readability.
+                    // They also might depend on build parameters, which would make
+                    // using a fixed number of frames otherwise brittle.
+                    version (LDC) enum BaseExceptionFunctionName = "_d_throw_exception";
+                    else          enum BaseExceptionFunctionName = "_d_throwdwarf";
+                    if (!throw_call_found)
+                    {
+                        throw_call_found = buf == BaseExceptionFunctionName;
+                        continue;
```
kinke, line 987: LGTM. My only worry is this here - no call stack at all if `_d_throw_exception` isn't found. `traceHandlerOpApplyImpl` does it in two passes, stashing the frame names in an array and setting the start index in the 1st pass when encountering the magic symbol name the first time. I guess this could be done here too.

JohanEngelen: I fixed this now

PR #4642, opened 2024-05-04
Bump LDC-LLVM to v18.1.5

kinke:


.github/workflows/main.yml
```diff
@@ -161,17 +161,18 @@ jobs:
...
             extra_cmake_flags: >-
               -DBUILD_LTO_LIBS=ON
               -DD_COMPILER_FLAGS="-O -flto=full -defaultlib=phobos2-ldc-lto,druntime-ldc-lto -L-exported_symbol '-L__*' -L-w"
-              -DEXTRA_CXXFLAGS=-flto=full
-            with_pgo: true
-            llvm_version: 17.0.6 # TODO
+            # FIXME: disabling LTO for C++ parts as well as PGO, out of caution due to
+            #        weird issues for native job experiments in https://github.com/ldc-developers/ldc/pull/4604
+            #  -DEXTRA_CXXFLAGS=-flto=full
+            #with_pgo: true
```
kinke, line 175: @JohanEngelen: This is pretty much shooting in the dark from my side. ;) - The problem is that the current prebuilt macOS arm64 package cannot be used for cross-compiling to Android, as its LDC-LLVM 17 emulates TLS, while druntime expects no TLS emulation.

So no idea whether this might cause random 'pure virtual function called' crashes now with the natively compiled LLVM 18, whether that has already been the case before for our current cross-compiled artifacts, or whether that is related to Xcode 15, or using a native host compiler (with e.g. 64-bit compile-time `real`), ...

What do you think?

JohanEngelen: sorry I have been putting off looking in detail at the pure virtual function called error. I'll see if I can reproduce it this weekend.

PR #4645, opened 2024-05-04
Detect musl in cmake and link with libunwind

JohanEngelen:


JohanEngelen:
@kinke Is it worthwhile letting the CI for cross-compiling druntime/phobos (and ldc-build-runtime probably too?) know about LDC's folder with cmake modules..... Or should I just move the `CHECK_MUSL` macro inside the runtime CMakeLists.txt and be happy?

kinke:
Sigh, if the `pragma(lib, …)` would be supported by the GNU linkers too (not just lld), we could simply embed a reference in a central druntime object file and let the linker handle everything.

If the compiler implicitly added `-lunwind` as 'platform-lib' for musl targets, the druntime integration tests would require less adaptations (then only required when linking via C(++) compiler).

Wrt. detecting musl, another option might be to check the LLVM host triple (`llvm-config --host-target` - AFAICT, would need a new `llvm_set(HOST_TARGET host-target)` in FindLLVM.cmake to set LLVM_HOST_TARGET), checking for a musl environment.

Wrt. integrating that in runtime/CMakeLists.txt, a potential option might be to (ab)use our `TARGET_SYSTEM` variable, adding `Musl` to that list, similar to `Android` (full TARGET_SYSTEM: `Android;Linux;UNIX`). For autodetection as part of a full LDC build, we could use LLVM_HOST_TARGET; for cross-compilation, TARGET_SYSTEM would need to be specified manually (but that's standard - just nice not having to remember to set an extra MUSL_LIBC variable when cross-compiling to a musl target).

JohanEngelen:
> If the compiler implicitly added `-lunwind` as 'platform-lib' for musl targets, the druntime integration tests would require less adaptations (then only required when linking via C(++) compiler).

LDC already does this, but indeed the problem is C/C++ linking.
https://github.com/ldc-developers/ldc/blob/40ad5f7583fe90a85fd675bfc0cfc287d565c94a/driver/linker-gcc.cpp#L684-L686
 
> Wrt. detecting musl, another option might be to check the LLVM host triple (`llvm-config --host-target` - AFAICT, would need a new `llvm_set(HOST_TARGET host-target)` in FindLLVM.cmake to set LLVM_HOST_TARGET), checking for a musl environment.
> 
> Wrt. integrating that in runtime/CMakeLists.txt, a potential option might be to (ab)use our `TARGET_SYSTEM` variable, adding `Musl` to that list, similar to `Android` (full TARGET_SYSTEM: `Android;Linux;UNIX`). For autodetection as part of a full LDC build, we could use LLVM_HOST_TARGET; for cross-compiling the libs, TARGET_SYSTEM would need to be specified manually (but that's standard - just nice not having to remember to additionally set an extra MUSL_LIBC variable when cross-compiling to a musl target).

I also want the autodetection to work for non-cross compile builds, just for building runtime with user-specified options (e.g. ASan stuff); then I cannot use `llvm-config` (although a good idea). So for now I'll stick with the current check but add it as macro to runtime/CMakeLists.txt. And I will add detection of "Musl" string in TARGET_SYSTEM, for user-override / cross-compiling capability.

runtime/druntime/test/common.mak
```diff
@@ -38,7 +40,7 @@ ifeq (,$(findstring ldmd2,$(DMD)))
     endif
 endif
 # LDC: use `-defaultlib=druntime-ldc [-link-defaultlib-shared]` instead of `-defaultlib= -L$(DRUNTIME[_IMPLIB])`
-DFLAGS:=$(MODEL_FLAG) $(PIC) -w -I../../src -I../../import -I$(SRC) -defaultlib=$(if $(findstring ldmd2,$(DMD)),druntime-ldc,) -preview=dip1000 $(if $(findstring $(OS),windows),,-L-lpthread -L-lm $(LINKDL))
+DFLAGS:=$(MODEL_FLAG) $(PIC) -w -I../../src -I../../import -I$(SRC) -defaultlib=$(if $(findstring ldmd2,$(DMD)),druntime-ldc,) -preview=dip1000 $(if $(findstring $(OS),windows),,-L-lpthread -L-lm $(LINKDL) $(LINKUNWIND))
```
kinke, line 43: If the compiler adds it implicitly already, we shouldn't need the extra flag here, and neither in `test/shared/Makefile` below, right? In fact, don't we just a need the `LUNWIND` variable alone, just for manual linking in `test/stdcpp/Makefile`?

runtime/CMakeLists.txt
```diff
@@ -137,6 +151,7 @@ message(STATUS "-- LDC runtime configuration:")
 message(STATUS "--  - Building 32/64-bit libraries (MULTILIB): ${MULTILIB}")
 message(STATUS "--  - Building shared libraries (BUILD_SHARED_LIBS): ${BUILD_SHARED_LIBS}")
 message(STATUS "--  - Building LTO libraries (BUILD_LTO_LIBS): ${BUILD_LTO_LIBS}")
+message(STATUS "--  - Linking test runners with (C_SYSTEM_LIBS): ${C_SYSTEM_LIBS}")
```
kinke, line 154: [I think the more interesting point is that the shared druntime/Phobos libs are linked against these.]

runtime/druntime/test/common.mak
JohanEngelen: I think you are right, I duplicated the LINKDL code. But `-ldl` is also already passed by the compiler implicitly... I Need to test without it to see if indeed it works.

kinke: Yeah that LINKDL ugliness is from upstream, not sure if we'd need it.

PR #4647, opened 2024-05-05
Replace invalid getTypeStoreSize() with getTypeAllocSize()

kinke:


gen/llvmhelpers.cpp
```diff
@@ -1949,7 +1949,7 @@ DValue *makeVarDValue(Type *type, VarDeclaration *vd, llvm::Value *storage) {
       pointeeType = pointeeType->getPointerElementType();
     // Make sure that the type sizes fit - '==' instead of '<=' should probably
     // work as well.
-    assert(getTypeStoreSize(DtoType(type)) <= getTypeStoreSize(pointeeType) &&
+    assert(getTypeAllocSize(DtoType(type)) <= getTypeAllocSize(pointeeType) &&
```
kinke, line 1952: Not 100% sure here, but I think it makes more sense (essentially, comparing the `.sizeof` frontend sizes). It's only an assertion for LLVM < 15, so breaking it wouldn't be the end of the world.

PR #4650, opened 2024-05-07
Add musl libc CI using Alpine Linux container on Ubuntu GHA

JohanEngelen:
Separate PRs to enable musl:
- [x] #4639
- [x] #4648
- [x] #4645 

Upstream PRs: 
- [x]  https://github.com/dlang/dmd/pull/16513
- [x] https://github.com/dlang/dmd/pull/16507
- [x] https://github.com/dlang/dmd/pull/16458

### DIFF
--- a/.github/workflows/alpine_musl.yml
+++ b/.github/workflows/alpine_musl.yml
@@ -21,7 +21,7 @@ jobs:
         uses: jirutka/setup-alpine@v1
         with:
           branch: v3.20
-          packages: ldc git g++ cmake ninja llvm-dev llvm-static libunwind-static libxml2-static zstd-static zlib-static bash grep diffutils make
+          packages: ldc git g++ cmake ninja llvm-dev llvm-static compiler-rt libunwind-static libxml2-static zstd-static zlib-static bash grep diffutils make
 
       - name: Build LDC bootstrap
         shell: alpine.sh {0}
--- a/.github/workflows/alpine_musl.yml
+++ b/.github/workflows/alpine_musl.yml
@@ -1,0 +1,81 @@
+name: Alpine Linux (musl libc)
+on:
+  - pull_request
+  - push
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: Alpine Linux (musl libc)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+          fetch-depth: 50
+
+      - name: Setup latest Alpine Linux
+        uses: jirutka/setup-alpine@v1
+        with:
+          branch: v3.20
+          packages: ldc git g++ cmake ninja llvm-dev llvm-static libunwind-static libxml2-static zstd-static zlib-static bash grep diffutils make
+
+      - name: Build LDC bootstrap
+        shell: alpine.sh {0}
+        run: |
+          ninja --version
+          set -eux
+          mkdir bootstrap
+          cd bootstrap
+          cmake -G Ninja .. \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX="/opt/ldc2" \
+            -DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++ \
+            -DD_COMPILER_FLAGS=-link-defaultlib-shared=false \
+            -DBUILD_SHARED_LIBS=OFF
+          ninja obj/ldc2.o all
+          bin/ldc2 --version
+          cd ..
+
+      # TODO: Add '-DLLVM_IS_SHARED=OFF' when static linking is fully supported
+      - name: Build LDC & LDC D unittests & defaultlib unittest runners
+        shell: alpine.sh {0}
+        run: |
+          set -eux
+          cmake -G Ninja . \
+            -DD_COMPILER=./bootstrap/bin/ldmd2 \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX="/opt/ldc2" \
+            -DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++ \
+            -DD_COMPILER_FLAGS=-link-defaultlib-shared=false \
+            -DBUILD_SHARED_LIBS=OFF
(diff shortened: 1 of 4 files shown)


JohanEngelen:
Is there a good way to disable all `gdb`-related tests?  (the gdb version on Alpine is too new and there are no alpine packages for older gdb versions)

kinke:
Yeah, we have this which should take care of most dmd-testsuite failures: https://github.com/ldc-developers/ldc/blob/40ad5f7583fe90a85fd675bfc0cfc287d565c94a/tests/dmd/tools/d_do_test.d#L1702-L1709

JohanEngelen:
The tester on Alpine Linux with musl is green now :))

JohanEngelen:
Upstream PRs: 
- [ ]  https://github.com/dlang/dmd/pull/16513
- [x] https://github.com/dlang/dmd/pull/16507
- [x] https://github.com/dlang/dmd/pull/16458

cmake/Modules/FindLLVM.cmake
```diff
@@ -85,9 +85,14 @@ else()
        if(LLVM_FIND_QUIETLY)
             set(_quiet_arg ERROR_QUIET)
         endif()
+        if (LLVM_IS_SHARED)
+            set(sharedstatic "--link-shared")
+        else()
+            set(sharedstatic "--link-static")
+        endif()
```
kinke, line 92: Is this stuff really required? Is the alpine llvm package misconfigured?

kinke: Ah, optionally definable manually now, just seen the TODO in the yaml.

.github/workflows/alpine_musl.yml
```diff
@@ -0,0 +1,77 @@
...
+            grep
+            diffutils
+            make
+
+      # TODO: Add '-DLLVM_IS_SHARED=OFF' when static linking is fully supported
+      - name: Build LDC & LDC D unittests & defaultlib unittest runners
+        run: |
+          set -eux
+          export CC=gcc
+          export CXX=g++
```
kinke, line 45: What are the default ones?

.github/workflows/alpine_musl.yml
```diff
@@ -0,0 +1,77 @@
...
+          cmake -G Ninja . \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX="/opt/ldc2" \
+            -DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++ \
+            -DD_COMPILER_FLAGS=-link-defaultlib-shared=false \
+            -DBUILD_SHARED_LIBS=OFF
+          ninja obj/ldc2.o all ldc2-unittest all-test-runners
+          bin/ldc2 --version
+          ldd bin/ldc2
+        shell: alpine.sh {0}
```
kinke, line 55: You can set this once for the whole job: https://docs.github.com/en/actions/using-jobs/setting-default-values-for-jobs#setting-default-values-for-a-specific-job

kinke:
Currently it doesn't find any compiler-rt libs, so `LDC_INSTALL_LLVM_RUNTIME_LIBS_OS` might need tweaking.

cmake/Modules/FindLLVM.cmake
kinke: But judging from the `ldd` output, LLVM is linked statically just fine already, but the other deps aren't (`lib{z,zstd,unwind}` etc.).

kinke: [I'm pretty sure this broke the FreeBSD job, where explicitly using `--link-shared` apparently causes `libzstd` to be dragged in, which isn't installed (so no dependency of the installed official llvm package on that library either) - probably misconfigured by FreeBSD.]

.github/workflows/alpine_musl.yml
```diff
@@ -0,0 +1,77 @@
...
+        if: success() || failure()
+        run: ctest -V -R "dmd-testsuite"
+        shell: alpine.sh {0}
+      - name: Run defaultlib unittests & druntime integration tests
+        if: success() || failure()
+        run: |
+          set -eux
+          excludes=
+          N=$(nproc)
+          ctest -j$N --output-on-failure -E "dmd-testsuite|lit-tests|ldc2-unittest"
```
kinke, line 76: [could be a one-liner as-is, no need for the 2 vars]

.github/workflows/alpine_musl.yml
JohanEngelen: They are empty by default.

cmake/Modules/FindLLVM.cmake
JohanEngelen: I had some trouble in shared/static linking of LLVM. The installation defaults to shared linking on Alpine (llvm-config default output), that's why I added this code here. It's perhaps not needed right now (no static linked build).
Fully static linking is not possible yet, because of the libunwind dependency. libunwind is incompatible with g++ static linking, because of conflicting symbols between libgcc_eh and libunwind. I think the only option is to make the libunwind dependency of druntime optional, and disable it for a fully statically linked LDC.

JohanEngelen: I need to link into this: https://github.com/libunwind/libunwind/issues/405#issuecomment-1209587993

.github/workflows/alpine_musl.yml
kinke: I meant the default compilers chosen by CMake if you don't define these 2 vars.

cmake/Modules/FindLLVM.cmake
kinke: Why do you wanna link fully statically anyway? As preparation for an official musl-x86_64 LDC package? That would be quite a bit more work anyway, and would ideally be integrated in the main.yml workflow then, as that performs the upload to GitHub etc. So I'd suggest deferring that to another PR.

kinke: [Btw a friend recently told me that the Alpine gcc is patched in a way that it doesn't emit the x86_64 cmpxch16b instruction. They noticed a ~20% overall performance improvement of glibc vs. musl because of that.]

JohanEngelen:
Green again!

JohanEngelen:
Any objections before merging?

With this in place, in follow up I intend to work towards statically linked build and uploading a release to GH.

I'll upstream the druntime fix.

PR #4652, opened 2024-05-09
Fix LoongArch support

heiher:
This PR fixes support for LoongArch so that it can run on LoongArch systems.

### DIFF
--- a/gen/abi/loongarch64.cpp
+++ b/gen/abi/loongarch64.cpp
@@ -9,7 +9,7 @@
 //===----------------------------------------------------------------------===//
 //
 // ABI spec:
-// https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
+// https://github.com/loongson/la-abi-specs/blob/release/lapcs.adoc
 //
 //===----------------------------------------------------------------------===//
 
--- a/gen/abi/loongarch64.cpp
+++ b/gen/abi/loongarch64.cpp
@@ -20,9 +20,101 @@
 #include "gen/llvmhelpers.h"
 #include "gen/tollvm.h"
 
+using namespace dmd;
+
+namespace {
+struct Integer2Rewrite : BaseBitcastABIRewrite {
+  LLType *type(Type *t) override {
+    return LLStructType::get(gIR->context(),
+                             {DtoType(Type::tint64), DtoType(Type::tint64)});
+  }
+};
+
+struct FlattenedFields {
+  Type *fields[2];
+  int length = 0; // use -1 to represent "no need to rewrite" condition
+};
+
+FlattenedFields visitStructFields(Type *ty, unsigned baseOffset) {
+  // recursively visit a POD struct to flatten it
+  FlattenedFields result;
+  if (auto ts = ty->isTypeStruct()) {
+    for (auto fi : ts->sym->fields) {
+      auto sub =
+          visitStructFields(fi->type->toBasetype(), baseOffset + fi->offset);
+      if (sub.length == -1 || result.length + sub.length > 2) {
+        result.length = -1;
+        return result;
+      }
+      for (unsigned i = 0; i < (unsigned)sub.length; ++i) {
+        result.fields[result.length++] = sub.fields[i];
+      }
+    }
+    return result;
+  }
+  switch (ty->ty) {
+  case TY::Tcomplex32: // treat it as {float32, float32}
+    result.fields[0] = pointerTo(Type::tfloat32);
+    result.fields[1] = pointerTo(Type::tfloat32);
+    result.length = 2;
+    break;
+  case TY::Tcomplex64: // treat it as {float64, float64}
+    result.fields[0] = pointerTo(Type::tfloat64);
+    result.fields[1] = pointerTo(Type::tfloat64);
+    result.length = 2;
+    break;
+  default:
+    if (ty->size() > 8) {
+      // field larger than GRLEN and FRLEN
+      result.length = -1;
+      break;
+    }
+    result.fields[0] = ty;
+    result.length = 1;
+    break;
+  }
+  return result;
+}
+
+struct HardfloatRewrite : BaseBitcastABIRewrite {
+  LLType *type(Type *ty, const FlattenedFields &flat) {
+    if (flat.length == 1) {
+      return LLStructType::get(gIR->context(), {DtoType(flat.fields[0])},
+                               false);
+    }
+    assert(flat.length == 2);
+    LLType *t[2];
+    for (unsigned i = 0; i < 2; ++i) {
+      t[i] =
+          flat.fields[i]->isfloating()
+              ? DtoType(flat.fields[i])
+              : LLIntegerType::get(gIR->context(), flat.fields[i]->size() * 8);
+    }
+    return LLStructType::get(gIR->context(), {t[0], t[1]}, false);
+  }
+  LLType *type(Type *ty) override {
+    return type(ty, visitStructFields(ty->toBasetype(), 0));
+  }
+};
+} // anonymous namespace
+
 struct LoongArch64TargetABI : TargetABI {
 private:
+  HardfloatRewrite hardfloatRewrite;
   IndirectByvalRewrite indirectByvalRewrite{};
+  Integer2Rewrite integer2Rewrite;
+  IntegerRewrite integerRewrite;
+
+  bool requireHardfloatRewrite(Type *ty) {
+    if (!isPOD(ty) || !ty->toBasetype()->isTypeStruct())
+      return false;
+    auto result = visitStructFields(ty->toBasetype(), 0);
+    if (result.length <= 0)
+      return false;
+    if (result.length == 1)
+      return result.fields[0]->isfloating();
+    return result.fields[0]->isfloating() || result.fields[1]->isfloating();
+  }
 
 public:
   auto returnInArg(TypeFunction *tf, bool) -> bool override {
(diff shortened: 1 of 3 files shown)


gen/abi/loongarch64.cpp
```diff
@@ -66,6 +77,16 @@ struct LoongArch64TargetABI : TargetABI {
       indirectByvalRewrite.applyTo(arg);
       return;
     }
+
+    Type *ty = arg.type->toBasetype();
+    if (isAggregate(ty) && ty->size() && ty->size() <= 16) {
+      if (ty->size() > 8 && DtoAlignment(ty) < 16) {
+        // pass the aggregate as {int64, int64} to avoid wrong alignment
+        integer2Rewrite.applyToIfNotObsolete(arg);
```
kinke, line 85: I'm sure this is better than the status quo, but is this correct for e.g. a `float[3]` too, with size 12?

heiher: It's correct for `float[3]`, but wrong for `{float, float}`. I'll fix it later. Thanks.

heiher: Fixed.

runtime/druntime/src/core/threadasm.S
```diff
@@ -610,6 +611,28 @@ fiber_switchContext:
 
     jr $ra
     .cfi_endproc
+
+/**
+ * When generating any kind of backtrace (gdb, exception handling) for
+ * a function called in a Fiber, we need to tell the unwinder to stop
+ * at our Fiber main entry point, i.e. we need to mark the bottom of
+ * the call stack. This can be done by clearing the link register lr
```
SixWeining, line 619: Nit: `link register lr` -> `return address register $ra`

runtime/druntime/src/core/threadasm.S
```diff
@@ -610,6 +611,28 @@ fiber_switchContext:
...
+ * prior to calling fiber_entryPoint (i.e. in fiber_switchContext) or
+ * using a .cfi_undefined directive for the link register in the
```
SixWeining, line 621: Ditto.

heiher: Done. Thanks.

gen/abi/loongarch64.cpp
```diff
@@ -66,6 +220,18 @@ struct LoongArch64TargetABI : TargetABI {
       indirectByvalRewrite.applyTo(arg);
       return;
     }
+
+    Type *ty = arg.type->toBasetype();
+    if (ty->isintegral() && (ty->ty == TY::Tint32 || ty->ty == TY::Tuns32)) {
+      arg.attrs.addAttribute(LLAttribute::SExt);
```
heiher, line 226: Refer: https://github.com/loongson/la-abi-specs/blob/release/lapcs.adoc
> One exception to the above rule is that in the LP64D ABI, unsigned words, such as those representing unsigned int in [C](https://github.com/loongson/la-abi-specs/blob/release/lapcs.adoc#C-data-types), are stored in general-purpose registers as proper sign extensions of their 32-bit values.

heiher:
@kinke gentle ping.

gen/abi/loongarch64.cpp
```diff
@@ -20,16 +20,155 @@
...
+  };
+  FlattenedField fields[2];
+  int length = 0; // use -1 to represent "no need to rewrite" condition
+};
+
+FlattenedFields visitStructFields(Type *ty, unsigned baseOffset) {
+  // recursively visit a POD struct to flatten it
+  // FIXME: may cause low performance
+  // dmd may cache argtypes in some other architectures as a TypeTuple, but we
+  // need to additionally store field offsets to realign later
```
kinke, line 46: Could you please give an example of such a problematic struct, where the layout doesn't match?

Also note that a union would have overlapping fields. Official ABI docs often don't explicitly detail how unions are to be dealt with.

gen/abi/loongarch64.cpp
```diff
@@ -20,16 +20,155 @@
...
+    result.fields[0].offset = baseOffset;
+    result.length = 1;
+    break;
+  }
+  return result;
+}
+
+bool requireHardfloatRewrite(Type *ty) {
+  if (!ty->toBasetype()->isTypeStruct())
+    return false;
```
kinke, line 92: This excludes complex types and static arrays.

gen/abi/loongarch64.cpp
```diff
@@ -46,11 +188,23 @@ struct LoongArch64TargetABI : TargetABI {
 
   void rewriteFunctionType(IrFuncTy &fty) override {
     if (!fty.ret->byref) {
-      rewriteArgument(fty, *fty.ret);
+      if (!skipReturnValueRewrite(fty)) {
+        if (!fty.ret->byref && isPOD(fty.ret->type) &&
```
kinke, line 192: `!fty.ret->byref` is checked 3 times here; `skipReturnValueRewrite()` checks it too.

gen/abi/loongarch64.cpp
kinke: Please add a source comment, I was pretty sure this has got to be wrong.

gen/abi/loongarch64.cpp
kinke: And I suggest adding the `isPOD` check here (and probably a size check to avoid costly flattening for larger types), instead of doing that at every call site.

gen/abi/loongarch64.cpp
kinke: And stylistically, it looks like you never need the original `ty`, only `ty->toBasetype()`, so flattening it once at the beginning might make sense.

gen/abi/loongarch64.cpp
heiher: The caller has done an `isPOD` check before calling `requireHardfloatRewrite`.

gen/abi/loongarch64.cpp
heiher: For example:

### C

f32x2.c: (https://godbolt.org/z/vK599hrq7)

```c
struct f32x2_s
{
	float a;
	float b;
};

struct f32x2_s f_f32x2_s(struct f32x2_s x)
{
	x.a += 1.0;
	x.b += 1.0;
	return x;
}
```

llvm ir:

```llvm
define dso_local { float, float } @f_f32x2_s(float %0, float %1) local_unnamed_addr #0 !dbg !9 {
    ...
}
```
### D

f32x2.d:

```d
extern(C) struct f32x2_s
{
	float a;
	float b;
}

extern(C) f32x2_s f_f32x2_s(f32x2_s x)
{
	x.a += 1.0;
	x.b += 1.0;
	return x;
}
```

llvm ir **without** this patch:

```llvm
define i64 @f_f32x2_s(i64 %x_arg) local_unnamed_addr #0 {
  ...
}
```

llvm ir **with** this patch:

```llvm
define { float, float } @f_f32x2_s({ float, float } %x_arg) local_unnamed_addr #0 {
  ...
}
```

gen/abi/loongarch64.cpp
heiher: done.

gen/abi/loongarch64.cpp
heiher: I'll add a source comment and describe more here.

For example:

### C

int32.c: (https://godbolt.org/z/vcjErxj76)

```c
int check_int()
{
	int v;
	return v;
}

unsigned int check_uint()
{
	unsigned int v;
	return v;
}
```

llvm ir:

```llvm
define dso_local noundef signext i32 @check_int(i32 noundef returned signext %0) local_unnamed_addr #0 !dbg !9 {
  ...
}

define dso_local noundef signext i32 @check_uint(i32 noundef returned signext %0) local_unnamed_addr #0 !dbg !18 {
  ...
}
```
### D

int32.d:

```d
extern(C) int check_int()
{
	int v;
	return v;
}

extern(C) uint check_uint()
{
	uint v;
	return v;
}
```

llvm ir **without** this patch:

```llvm
define i32 @check_int(i32 returned %v_arg) local_unnamed_addr #0 {
  ...
}

define i32 @check_uint(i32 returned %v_arg) local_unnamed_addr #0 {
  ...
}

```

llvm ir **with** this patch:

```llvm
define signext i32 @check_int(i32 returned signext %v_arg) local_unnamed_addr #0 {
  ...
}

define signext i32 @check_uint(i32 returned signext %v_arg) local_unnamed_addr #0 {
  ...
}
```

heiher (commented):
Thank you.

gen/abi/loongarch64.cpp
kinke: I meant an example where the struct needing a rewrite has a different layout than the rewritten one. Where you need to store the field offsets and do the piecewise memcpy's, instead of a trivial bitcast. This struct here wouldn't need any rewrite.

gen/abi/loongarch64.cpp
kinke: That's what I'm saying, it's ugly to do that at every call site, instead of once here and having the guarantee for the following logic then.

heiher: done

gen/abi/loongarch64.cpp
heiher: The code is copied from RISC-V, and I don't know which struct need to be rewritten to use temporary memory space to realign. Let's simplify it unless there are new tips.

heiher: Thanks for your review and help. I agree that the front-end implementation would be a much better approach. I'm a newbie in D, so this might take some time for me. It would be best if we could move forward with the current patch to make things better, and I will continue to improve it.

gen/abi/loongarch64.cpp
```diff
@@ -20,16 +20,111 @@
...
 
 public:
   auto returnInArg(TypeFunction *tf, bool) -> bool override {
     if (tf->isref()) {
       return false;
     }
     Type *rt = tf->next->toBasetype();
+    if (!rt->size()) {
+      return false;
+    }
```
kinke, line 127: This isn't really required. Every non-POD has a non-empty size (empty structs have a size of 1 - it's only `void`, `noreturn` and 0-length static arrays that have a size of 0).

gen/abi/loongarch64.cpp
```diff
@@ -38,19 +133,30 @@ struct LoongArch64TargetABI : TargetABI {
   }
 
   auto passByVal(TypeFunction *, Type *t) -> bool override {
+    if (!t->size()) {
+      return false;
+    }
```
kinke, line 138: ditto

gen/abi/loongarch64.cpp
```diff
@@ -38,19 +133,30 @@ struct LoongArch64TargetABI : TargetABI {
...
+        // rewrite here because we should not apply this to variadic arguments
+        hardfloatRewrite.applyTo(*fty.ret);
+      } else {
+        rewriteArgument(fty, *fty.ret);
+      }
     }
 
     for (auto arg : fty.args) {
-      if (!arg->byref) {
+      if (!arg->byref && requireHardfloatRewrite(arg->type)) {
```
kinke, line 156: This now means that if `arg->byref` is true, `rewriteArgument()` is called, which is wrong.

gen/abi/loongarch64.cpp
kinke: Note that there's also `TY::Tdchar`, which is a 32-bit integer too.

gen/abi/loongarch64.cpp
kinke: Okay, I did one more review round, just some little nits.

gen/abi/loongarch64.cpp
heiher: Really. The corresponding type `char32_t` in C/C++ is also sign-extended. Thanks.

PR #4657, opened 2024-05-12
Drop support for LLVM < 15, legacy pass manager and non-opaque IR pointers

kinke:
LLVM 14 is the first version supporting both new pass manager and opaque IR pointers.


kinke:
Dropping typed IR pointers seems not really feasible yet:
* LLVM 14: the compiler crashes when compiling druntime, on both macOS and Linux
* LLVM 15+16: unittest `rt.lifetime` fails with `-O` - https://github.com/ldc-developers/ldc/issues/4538

So I guess that'll have to wait until we raise min LLVM to v17.

kinke:
Any opinions? Or concerns wrt. dropping the legacy pass manager support?

As another option, we could drop LLVM 14 too, drop typed IR pointers already, and live with that one test failure for LLVM < 17 (a bit of a PITA, as the same failure causes a druntime integration test to fail too). I don't *think* that LLVM problem would be a huge deal in practice, for non-official packages using such an older LLVM (or our official FreeBSD package, which uses LLVM 15 IIRC). It'd be interesting to see when clang switched to opaque IR pointers by default.

JohanEngelen:
No concerns from my side. Weka is using LLVM14 at the moment, but that's because of using an older LDC for now. The plan is to upgrade to 1.38 first, then upgrade LLVM, and then to later LDC. So no problem if 1.39 does not support LLVM14.

kinke:
This is ready; it's just that the 'required' CI jobs need to adapted after landing this.

I'll try to follow up with some pointer-bitcast cleanup; I have something locally already, but it appears to be misbehaving badly at the moment. ;)

PR #4660, opened 2024-05-18
Upgrade frontend & libs to v2.109

kinke:


driver/cl_options.cpp
```diff
@@ -603,10 +603,31 @@ static cl::opt<bool, true>
...
     cl::desc("Omit generating some runtime information and helper functions"));
 
+static cl::opt<CLIIdentifierTable, true> identifiers(
+    "identifiers", cl::ZeroOrMore, cl::location(global.params.dIdentifierTable),
+    cl::desc("Specify the non-ASCII tables for D identifiers"),
+    cl::values(clEnumValN(CLIIdentifierTable::C99, "c99", "C99"),
+               clEnumValN(CLIIdentifierTable::C11, "c11", "C11"),
+               clEnumValN(CLIIdentifierTable::UAX31, "UAX31", "UAX31"),
+               clEnumValN(CLIIdentifierTable::All, "all",
+                          "All, the least restrictive set, which comes all "
```
JohanEngelen, line 618: typo? "comes" --> "contains"?

kinke: A question for upstream, I was wondering too what this is supposed to mean.

kinke:
Still just a single regression for lit-test `PGO/functions.d`, weirdly on some platforms only (Windows x64 and macOS arm64).

kinke:
AFAICT, the `PGO/functions.d` regression is that not all objects are destructed anymore during the profiled run. E.g., in my Win VM, the `Klass` dtor is run 9 times, while it should be 10. (The test apparently depends on the GC finalizing everything, the class objects aren't destructed explicitly).

The test has also failed on Linux x64 once at least, for a CircleCI job: https://app.circleci.com/pipelines/github/ldc-developers/ldc/4250/workflows/23b08985-2521-4562-ae10-0659fce872cf/jobs/26557

The `-fprofile-instr-generate` IR is identical, so this diverging behavior must be caused by druntime.

kinke:
Ah, expected in v2.109: https://dlang.org/changelog/2.109.0.html#druntime.removenostackcollect

PR #4662, opened 2024-05-18
tests/codegen/casm.c: Move to casm.i to avoid calling the preprocessor

the-horo:
This test may fail on x86 because, when calling the C preprocessor, ldc will append a -m argument based on the target. If the target is x86_64 and the host is x86 the C compiler may fail with:
```
cc1: error: CPU you selected does not support x86-64 instruction set
```

An easy solution for this is to specify i386 instead of x86_64 in the target triple.

I'm not 100% sure if the `-mtriple` argument can be just dropped so I went ahead with a more conservative fix.

### DIFF
--- a/tests/codegen/casm.c
+++ b/tests/codegen/casm.c
@@ -1,6 +1,0 @@
-// REQUIRES: target_X86
-// UNSUPPORTED: Windows
-// RUN: %ldc -mtriple=x86_64-freebsd13 -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
-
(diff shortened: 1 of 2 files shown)


kinke:
We only need this to be a .c to preprocess the `__asm__` to `asm` (done in importc.h`). The cross-compilation complications for the preprocessor is also the only reason why this test is disabled on Windows. So I'd suggest making this an `.i` and replacing the `__asm__` manually to `asm` to avoid the preprocessor and its complications, incl. enabling this test on Windows hosts too then.

FWIW, this test originates from https://github.com/ldc-developers/ldc/blob/master/tests/dmd/fail_compilation/fail24389.c.

PR #4663, opened 2024-05-20
GHA: Get rid of subdirs in .github/actions/

kinke:
I've just recently seen that the YAMLs can be included by filename too.


kinke:
Oh, this is only supported for a job's `uses`: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses

Issue #4670, opened 2024-05-25
WebAssembly: doesn't support emscripten data layout


kassane:
**LDC2 version:** 1.38.0
**EMCC version:** 3.1.59

Based in my analysis https://github.com/kassane/sokol-d/issues/3#issuecomment-2131277347.

I've detected conflicts between targets involving only ldc2 + emcc with LTO enabled on both.
```bash
wasm-ld: warning: Linking two modules of different target triples: '/home/kassane/.cache/zig/p/122083537ec3463ee4b2d520bf86a36b1057ab05abfe712c32bae1048f9b0c6e2a13/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libc.a(llrintl.o at 2994444)' is 'wasm32-unknown-emscripten' whereas 'ld-temp.o' is 'wasm32-unknown-unknown-wasm'
wasm-ld: warning: Linking two modules of different data layouts: '/home/kassane/.cache/zig/p/122083537ec3463ee4b2d520bf86a36b1057ab05abfe712c32bae1048f9b0c6e2a13/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libc.a(frexpf.o at 1903762)' is 'e-m:e-p:32:32-p10:8:8-p20:8:8-i64:64-f128:64-n32:64-S128-ni:1:10:20' whereas 'ld-temp.o' is 'e-m:e-p:32:32-p10:8:8-p20:8:8-i64:64-n32:64-S128-ni:1:10:20'
```
emsc-libc:
`e-m:e-p:32:32-p10:8:8-p20:8:8-i64:64-f128:64-n32:64-S128-ni:1:10:20`
ldc:
`e-m:e-p:32:32-p10:8:8-p20:8:8-i64:64-n32:64-S128-ni:1:10:20`

and **ldc2** no have `wasm32-unknown-emscripten` triple-target.

JohanEngelen:
The warning is fixed by https://github.com/ldc-developers/ldc/pull/4672

The data layout that I get with LDC master and LLVM15, 17, and 18 is the same as emsc-libc:
`e-m:e-p:32:32-p10:8:8-p20:8:8-i64:64-f128:64-n32:64-S128-ni:1:10:20`

PR #4672, opened 2024-05-25
Add support for triple wasm32-unknown-emscripten

JohanEngelen:
fixes https://github.com/ldc-developers/ldc/issues/4670


JohanEngelen:
The `Emscripten` version identifier is already documented: https://dlang.org/spec/version.html

PR #4691, opened 2024-06-30
Remove druntime's libunwind dependency

JohanEngelen:


JohanEngelen:
FYI: this is one of the stumbling blocks when trying to statically link with musl libc

tests/dmd/runnable/test17559.d
```diff
@@ -2,6 +2,7 @@
 // REQUIRED_ARGS(linux freebsd dragonflybsd): -L-export-dynamic
 // LDC (required for Win32 and -O): REQUIRED_ARGS(windows32mscoff): -link-defaultlib-debug
 // LDC (FreeBSD's libexecinfo apparently doesn't like elided frame pointers): REQUIRED_ARGS(freebsd): -link-defaultlib-debug -frame-pointer=all
+// LDC (our _UnwindBacktrace implementation - for musl libc - requires this): REQUIRED_ARGS(linux): -link-defaultlib-debug
```
JohanEngelen, line 5: It'd be worthwhile to figure out what is causing the need for `-link-defaultlib-debug` on all these platforms. Perhaps we can solve this by preventing optimization (with magic UDA) of specific functions in druntime (or remove some other delta between optimized vs debug builds of specific functions).

JohanEngelen: (make new issue for that?)

kinke: Win32 backtraces are a catastrophe AFAIK, but I couldn't care less. For FreeBSD, it's the elided frame pointers as per the comment. For musl, no idea, but NOT testing the release variant for *every* Linux distro would IMO be a test regression.

JohanEngelen: yeah...
add the complexity of musl-specific parameters?

kinke: I don't know, that doesn't sound sexy at least. In the short term, I'd probably remove this test file for musl CI and file an issue, unless you wanna dig into the problem and try to fix it.

kinke:
Please let me know if you wanna include this in v1.39, then I'll wait with the release.

tests/dmd/runnable/test17559.d
JohanEngelen: Yeah I was already digging into it but it will cost more time than simple thing I tried. So let's remove it from CI, and file a separate issue to be solved in time.

JohanEngelen:
> Please let me know if you wanna include this in v1.39, then I'll wait with the release.

It would actually really be nice to get this in yes (minus the test change --> remove tests from musl CI for the time being).

tests/dmd/runnable/test17559.d
JohanEngelen: fixed it by running the tests manually for now

PR #4700, opened 2024-07-12
Make shell configurable in CI scripts, to later enable re-using same …

JohanEngelen:
…scripts for Alpine musl CI.

### DIFF
--- a/.github/actions/2-build-bootstrap/action.yml
+++ b/.github/actions/2-build-bootstrap/action.yml
@@ -5,12 +5,15 @@ inputs:
     default: ''
   arch:
     required: false # Windows only
+  shell:
+    required: false
+    default: bash
 runs:
   using: composite
   steps:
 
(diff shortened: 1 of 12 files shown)


kinke:
God is this ugly. How about simply changing the shell once for the musl job, as already proposed earlier?

kinke:
Seeing that there have been a bunch of explicit `bash` already (for Windows), an alternative might be simply using an Alpine Docker image, instead of that chroot-thing or whatever it is in detail that the GHA action uses.

JohanEngelen:
> God is this ugly. How about simply changing the shell once for the musl job, as already proposed earlier?

how would that work? Because everywhere it is explicitly mentioned, so overrides the default.

kinke:
We can default the Windows job shell to bash too, and explicitly select `cmd` for those Windows-specific steps.

kinke:
Not needed anymore.

PR #4704, opened 2024-07-14
GHA: Default to bash shell for main & supported-LLVM-versions workflows

kinke:
Only overriding it to `cmd`/`pwsh` for Windows-specific steps that need it.


kinke:
Argh: https://github.com/orgs/community/discussions/18597

PR #4705, opened 2024-07-14
GHA: Switch to Docker image for Alpine workflow

kinke:


kinke:
@JohanEngelen: Your call. :)

JohanEngelen:
lol, it's this easy !? I didn't know this is possible.  I don't see any downside to this.

JohanEngelen:
(I guess with this solution, and installing `bash` on Alpine, the other scripts won't need as much adjustment)

kinke:
Yeah exactly, at least we shouldn't need to tweak `shell` anymore with this.

Issue #4646, opened 2024-05-05
Assertion failure when a struct contains a struct which contains bit fields spanning [17, 25) or [33, 57) bits.


just-harry:
A quick test case:

In a C file, define `BitFlags` like:
```c
typedef struct BitFlags {
  unsigned int _00 : 1;
  unsigned int _01 : 1;
  unsigned int _02 : 1;
  unsigned int _03 : 1;
  unsigned int _04 : 1;
  unsigned int _05 : 1;
  unsigned int _06 : 1;
  unsigned int _07 : 1;
  unsigned int _08 : 1;
  unsigned int _09 : 1;
  unsigned int _10 : 1;
  unsigned int _11 : 1;
  unsigned int _12 : 1;
  unsigned int _13 : 1;
  unsigned int _14 : 1;
  unsigned int _15 : 1;
  unsigned int _16 : 1;
} BitFlags;
```

or like:
```c
typedef struct BitFlags {
  unsigned long long _00 : 1;
  unsigned long long _01 : 1;
  unsigned long long _02 : 1;
  unsigned long long _03 : 1;
  unsigned long long _04 : 1;
  unsigned long long _05 : 1;
  unsigned long long _06 : 1;
  unsigned long long _07 : 1;
  unsigned long long _08 : 1;
  unsigned long long _09 : 1;
  unsigned long long _10 : 1;
  unsigned long long _11 : 1;
  unsigned long long _12 : 1;
  unsigned long long _13 : 1;
  unsigned long long _14 : 1;
  unsigned long long _15 : 1;
  unsigned long long _16 : 1;
  unsigned long long _17 : 1;
  unsigned long long _18 : 1;
  unsigned long long _19 : 1;
  unsigned long long _20 : 1;
  unsigned long long _21 : 1;
  unsigned long long _22 : 1;
  unsigned long long _23 : 1;
  unsigned long long _24 : 1;
  unsigned long long _25 : 1;
  unsigned long long _26 : 1;
  unsigned long long _27 : 1;
  unsigned long long _28 : 1;
  unsigned long long _29 : 1;
  unsigned long long _30 : 1;
  unsigned long long _31 : 1;
  unsigned long long _32 : 1;
} BitFlags;
```

then define a struct which contains `BitFlags`, e.g.
```c
typedef struct Foo {
  BitFlags flags;
} Foo;
```

The following failure will occur upon compilation:
```
[...]

PR #4707, opened 2024-07-16
Upgrade frontend & libs to v2.110

kinke:


kinke:
Great, just one problem for new `runnable_cxx/testbitfields.d`, possibly the same as https://github.com/ldc-developers/ldc/issues/4646.

kinke:
Interesting, another bit fields problem remaining, on Linux AArch64 only, working with Apple-clang on arm64 though; probably a frontend problem, needing more special cases:
```
1837: Test 'runnable_cxx/testbitfields.d' failed. The logged output:
1837: /tmp/cirrus-ci-build/../clang/bin/clang++ -c runnable_cxx/extra-files/testbitfields_cpp.cpp -o ../../../build/dmd-testsuite-debug/runnable_cxx/d/testbitfields_cpp.cpp.o -std=c++11
1837: /tmp/build/bin/ldmd2 -conf= -m64 -Irunnable_cxx -g -link-defaultlib-debug -L-lstdc++ -L--no-demangle  -od../../../build/dmd-testsuite-debug/runnable_cxx/d -of../../../build/dmd-testsuite-debug/runnable_cxx/d/testbitfields_0  runnable_cxx/testbitfields.d runnable_cxx/extra-files/testbitfields_importc.c ../../../build/dmd-testsuite-debug/runnable_cxx/d/testbitfields_cpp.cpp.o 
1837: ../../../build/dmd-testsuite-debug/runnable_cxx/d/testbitfields_0  --DRT-testmode=run-main
1837: Struct S16 has different bitfield layout for C and D:
1837:   D: size=4 align=1
1837:   C: size=4 align=4
1837: Struct S18 has different bitfield layout for C and D:
1837:   D: size=9 align=1
1837:   C: size=16 align=8
1837:   char a:
1837:     D: FF000000 00000000 00
1837:     C: FF000000 00000000 00000000 00000000
1837:   char b:
1837:     D: 00000000 00000000 FF
1837:     C: 00000000 00000000 FF000000 00000000
```

kinke:
Oh WTF:
```cpp
struct S16 { int :32; };
struct S18 { char a; long long :0; char b; };

static_assert(alignof(S16) == 1); // 4 on non-Apple AArch64
static_assert(alignof(S18) == 1); // 8 on non-Apple AArch64
```

It works fine for `arm64-apple-macos` and `riscv64-linux-gnu`, x86 etc., but not for `aarch-linux-gnu` and `aarch64-freebsd`. So looks like some AArch64 special case, where anonymous and 0-length bitfields do contribute to the alignment, and Apple went with a simpler/standard approach. I've checked this with clang v18.1.0, and latest stable AArch64 gcc on godbolt.org.

This appears to be handled in https://github.com/dlang/dmd/blob/6c826179ad5eec50b0bc40411acce8f7053e333f/compiler/src/dmd/dsymbolsem.d#L7176-L7193, possibly requiring a new `BitFieldStyle` enum member.

kinke:
Okay, spec is here: https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#1018bit-fields
> The container type contributes to the alignment of the containing aggregate in the same way a plain (not bit-field) member of that type would, **without exception for zero-sized or anonymous bit-fields**.
> 
> Note
> 
> The C++ standard states that an anonymous bit-field is not a member, so it is unclear whether or not an anonymous bit-field of non-zero size should contribute to an aggregate’s alignment. Under this ABI it does.

tim-dlang:
The layout for Linux armv7 is currently also wrong. Interestingly Apple seems to behave the same as Linux for armv7 and only aarch64 is different:
```
for t in aarch64-apple-darwin aarch64-unknown-linux-gnu \
  armv7k-apple-watchos armv7s-apple-ios armv7-unknown-linux-gnueabi
do
  echo ============ $t ============
  clang test.c -Xclang -fdump-record-layouts -c --target="$t" \
    | sed -n "/Dumping AST/,/^$/p"
done
```
```
============ aarch64-apple-darwin ============
*** Dumping AST Record Layout
         0 | struct S18
         0 |   char a
       8:- |   long long 
         8 |   char b
           | [sizeof=9, align=1]

============ aarch64-unknown-linux-gnu ============
*** Dumping AST Record Layout
         0 | struct S18
         0 |   char a
       8:- |   long long 
         8 |   char b
           | [sizeof=16, align=8]

============ armv7k-apple-watchos ============
*** Dumping AST Record Layout
         0 | struct S18
         0 |   char a
       8:- |   long long 
         8 |   char b
           | [sizeof=16, align=8]

============ armv7s-apple-ios ============
*** Dumping AST Record Layout
         0 | struct S18
         0 |   char a
       4:- |   long long 
         4 |   char b
           | [sizeof=8, align=4]

============ armv7-unknown-linux-gnueabi ============
*** Dumping AST Record Layout
         0 | struct S18
         0 |   char a
       8:- |   long long 
         8 |   char b
           | [sizeof=16, align=8]
```

PR #4712, opened 2024-07-27
Bump LDC-LLVM to v18.1.8

kinke:


kinke:
Sigh, that macOS arm64 thing... no improvements with Xcode 16 by the looks of it (beta 3 used for both LLVM and LDC).

The unsupported-stack-probing-method (when enabling LTO for the C++ parts) seems to originate from https://github.com/llvm/llvm-project/blob/3b5b5c1ec4a3095ab096dd780e84d7ab81f3d7ff/llvm/lib/Target/AArch64/AArch64MachineFunctionInfo.cpp#L142, only supporting `inline-asm` apparently.

According to https://stackoverflow.com/questions/78697653/how-do-i-get-a-clang-compiler-invocation-to-generate-debug-information, Apple clang might specify `probe-stack=__chkstk_darwin`.

So I guess what happens is that Apple have modified this in their fork, including patching that check for `inline-asm`. I've tried using `-mno-stack-arg-probe -fno-stack-protector` for the C++ parts then, in the hopes of turning off these stack probes for good, but no, the same error remains.

kinke:
Btw, @ Apple users reading this: is there a way to figure out what LLVM version Apple-clang v16 is based on?

thewilsonator:
Wikipedia says 17.0.6 https://en.wikipedia.org/wiki/Xcode#Xcode_15.0_-_(since_visionOS_support)_2

JohanEngelen:
> Okay, what seems to be working on macOS arm64 is skipping the bootstrap LDC build, using a pinned LDC v1.39 host compiler (based on LLVM 17, as Xcode 16) instead for the PGO-instrumented and final builds. I guess there's a breaking change in LLVM 18 wrt. stack checks on Apple arm64, causing these sporadic 'pure virtual func called' crashes with the C++ parts built with LLVM 17.

Thanks for the hard work of figuring out a solution!!

PR #4716, opened 2024-07-31
cmake: allow building liblphobos with system zlib

liushuyu:
This pull request allows third-party distributors to build Phobos runtime shipped with LDC using the system zlib library.

Many Linux distributions ship their own patches to do this; this pull request tries to avoid that and provides a unified way to achieve this goal.


thewilsonator (approved):

kinke:
Thanks. What's missing is making the compiler link the library, especially/only(?) when linking static Phobos. E.g., by adding it as defaultlib in `ldc2.conf` - something like `-defaultlib=phobos2-ldc,z,druntime-ldc`.

kinke:
Superseded by #4742.

Issue #4714, opened 2024-07-29
Wish: implement %ldccurrentpath% variable for ldc2.conf


denizzzka:
This would allow to make few standalone standard libraries versions and switch them by selecting desired config file containing compiler options and relative paths to all necessary things

(I can implement this, of course)

kinke:
With which semantics?

denizzzka:
Let's say make a `d-stdlibs-freertos` dir containing library files (in `lib/`), imports (in `import/`) and `ldc2.conf` configuration file in root of dir:
```ini
default:
{
    switches = [
        "--mtriple=thumbv7em-unknown-none-eabi",
        "-defaultlib=druntime-ldc,phobos2-ldc"
    ];
    post-switches = [
        "-I%ldccurrentpath%/import/src/",
        "-I%ldccurrentpath%/import/tagged/",
        "-I%ldccurrentpath%/import/phobos/"
    ];
    lib-dirs = [
        "%ldccurrentpath%/lib",
    ];
    rpath = "%ldccurrentpath%/lib";
};
```

Then, using `ldc2 --conf=path/to/d-stdlibs-freertos/ldc2.conf app.d` we can compile `app.d`

kinke:
So the directory containing the ldc2.conf file. I guess that would be fine, but I'd call it `%%ldcconfigpath%%` then, analogous to `%%ldcbinarypath%%`.

PR #4717, opened 2024-07-31
ldc2.conf: %%ldcconfigpath%% support added

denizzzka:
Fix https://github.com/ldc-developers/ldc/issues/4714

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Big news
 - Android: NDK for prebuilt package bumped from r26d to r27. (#4711)
+- ldc2.conf: %%ldcconfigpath%% placeholder added - specifies the directory where current configuration file is located. (#4717)
 
 #### Platform support
 
--- a/driver/configfile.d
+++ b/driver/configfile.d
@@ -18,7 +18,7 @@ import core.stdc.stdio;
 import core.stdc.string;
 
 
-string prepareBinDir(const(char)* binDir)
+string normalizeSlashes(const(char)* binDir)
 {
     immutable len = strlen(binDir);
     auto res = binDir[0 .. len].dup;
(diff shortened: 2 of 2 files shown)


thewilsonator (commented):
Thanks, could you add this to the changelog as well as some documentation.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 #### Big news
 - Android: NDK for prebuilt package bumped from r26d to r27. (#4711)
+- ldc2.conf: %%ldcconfigpath%% placeholder added - specifies the directory where current configuration file is located.
```
thewilsonator, line 5: please also add the PR number to the end (see the above entry for android)

thewilsonator:
> There is no documentation for %%ldcbinarypath%% placeholder, so I don't know where and what to write

Hmm, will wait for others to hopefully give some guidance.

CHANGELOG.md
denizzzka: Done

driver/configfile.d
```diff
@@ -117,8 +150,7 @@ private:
     {
         switches.setDim(0);
         postSwitches.setDim(0);
-
-        immutable dBinDir = prepareBinDir(binDir);
+        const CfgPaths cfgPaths = { cfPath: cfPath, binDir: binDir };
```
kinke, line 153: Hey Dennis, sorry for the delay. - Please convert these paths once here, not for every `replacePlaceholders()` call.

That `prepareBinDir()` is an awful name - apparently its job is to normalize the slashes in the path to forward slashes. No idea why or whether that's necessary, but oh well. If you wanna keep that for the new variable, feel free to, but then we need to rename the function.

Your `baseDir()` addition shouldn't be required; there's a static overloaded `dmd.root.filename.FileName.path()` function that does this already.

denizzzka: Done

kinke (approved):
Thanks!

PR #4720, opened 2024-08-04
Partial revert of d16fba30306b4d8993c0b5e0f22bbbc43b8135e7 to fix #4719

### LINKED ISSUE #4719: Incorrect LLVM type for struct with single align(1) member
This test case triggers the assert: https://github.com/ldc-developers/ldc/blob/48c1dff0466f5231619356ae33ef1a181888a1ff/ir/irtypeaggr.cpp#L43

```d
pragma(msg, TraceBuf.alignof); // prints 1
struct TraceBuf {
    align(1) uint args;
}

void foo() {
    byte[2] fixDescs;
    TraceBuf fixLog;

    auto dlg = delegate() {
        fixDescs[0] = 1;
        fixLog.args = 1;
    };
}
```

This is a minimized testcase, that no longer crashes LDC. The original case ran into the error
`Error: variable ... overlaps previous field. This is an ICE, please file an LDC issue.`

The reason is that the LLVM type is wrong since 1.31.0.

Used to be:
`%example.TraceBuf = type <{ i32 }>`
but now is:
`%example.TraceBuf = type { i32 }`, i.e. alignment is not 1 but 4.

JohanEngelen:


JohanEngelen:
Alternative is preferred: https://github.com/ldc-developers/ldc/pull/4722

PR #4722, opened 2024-08-05
AggrTypeBuilder: mark aggregate as packed when field offset is not naturally aligned, to fix #4719

### LINKED ISSUE #4719: Incorrect LLVM type for struct with single align(1) member

JohanEngelen:
Alternative to https://github.com/ldc-developers/ldc/pull/4720

### DIFF
--- a/tests/codegen/gh4719.d
+++ b/tests/codegen/gh4719.d
@@ -1,0 +1,32 @@
+// RUN: %ldc -c -output-ll -of=%t.ll %s
+
+struct TraceBuf {
+    align(1) uint args;
+}
+
+// Test correct compilation (no error) of the context pointer type for the delegate of `foo`.
+void foo() {
+    byte[2] fixDescs;
+    TraceBuf fixLog;
+
+    auto dlg = delegate() {
+        fixDescs[0] = 1;
+        fixLog.args = 1;
+    };
+}
+
+class TraceClass {
+    align(1)
+    uint args;
+}
+
+// Test correct compilation (no error) of the context pointer type for the delegate of `foo2`.
+void foo2() {
+    byte[2] fixDescs;
+    scope TraceClass fixLog;
+
+    auto dlg = delegate() {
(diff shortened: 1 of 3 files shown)


kinke:
Excellent, I wasn't sure whether you'd expect me to fix this, glad I haven't started yet! :)

tests/codegen/gh4719.d
```diff
@@ -0,0 +1,32 @@
+// RUN: %ldc -c -output-ll -of=%t.ll %s
```
kinke, line 1: Are you planning on adding IR checks?

kinke: Or just a little speedup to skip the machine codegen?

JohanEngelen: Oh this is a good point, I think I should change it to do codegen (and possibly trigger bug further down the line)

kinke: [I've used `tests/compilable/` for pure does-it-compile tests without IR checks.]

kinke (approved):

PR #4731, opened 2024-08-11
Accept pure static libraries without object files as `-lib` inputs

kinke:


kinke:
Not directly related, but turns out we don't actually merge .a libs on Posix as I had assumed (AFAIK, this just works on Windows), the 'merged' .a archive contains the nested .a libs, not their object files, making it unusable.

Issue #4734, opened 2024-08-12
Wrong struct layout (alignment related), followup #4719


JohanEngelen:
This is a followup from #4719.

The layout of struct `Table` (or `Item`) is wrong since LDC 1.31. 

```d
align(1) struct Item {
        KV v;
        uint i;
}
/+
// Also fails:
struct Item {
    align(1):
        KV v;
        uint i;
}+/

struct KV {
        align(1) S* s;
        uint k;
}

struct S {
        Table table;
}

struct Table {
        char a;
        Item v;
}

pragma(msg, Table.sizeof); // "17LU"
pragma(msg, Table.v.offsetof); // "1LU"

void foo(ref Table table, Item a) {
    // should write at offset 1, but LDC >= 1.31 writes at offset 4
    table.v = a;
}
```

In LLVM IR:
```llvm
; LDC 1.30
%example.S = type { %example.Table }
%example.Table = type { i8, %example.Item }
%example.Item = type <{ %example.KV, i32 }>
%example.KV = type <{ %example.S*, i32 }>

; LDC >= 1.31
%example.S = type { %example.Table }
%example.Table = type { i8, %example.Item }
%example.Item = type { %example.KV, i32 }  ; < > missing here
%example.KV = type <{ %example.S*, i32 }>
```

JohanEngelen:
Godbolt:  https://d.godbolt.org/z/o77Kafhr6

kinke:
I guess this is an instance of https://github.com/ldc-developers/ldc/issues/4236, due to `Item` having that **outer** `align(1)`, which just affects the tail padding, without explicitly making all members impicitly (potentially) misaligned.

Does this change things?
```D
struct KV {
        align(4) S* s; // to prevent padding KV to 16 bytes on 64-bit targets
        uint k;
} // note: still an overall alignment of 4 as before, and size of 8/12 bytes

struct Item {
        KV v;
        uint i;
} // now an alignment of 4, and same size 12/16 bytes
```

kinke:
Oh sorry, just seen your
```D
// Also fails:
struct Item {
    align(1):
        KV v;
        uint i;
}
```
now, so aligning the members instead apparently isn't enough.

kinke:
This IR layout seems wrong:
```LL
%example.Table = type { i8, %example.Item }
```

[...]

PR #4736, opened 2024-08-18
Get rid of cycles in DtoType()

kinke:
With the nice side-effect of fixing #4734.

### DIFF
--- a/ir/irtypestruct.cpp
+++ b/ir/irtypestruct.cpp
@@ -11,6 +11,7 @@
 
 #include "dmd/aggregate.h"
 #include "dmd/declaration.h"
+#include "dmd/errors.h"
 #include "dmd/init.h"
 #include "dmd/mtype.h"
 #include "dmd/template.h"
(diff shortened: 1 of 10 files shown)


ir/irtypestruct.cpp
```diff
@@ -91,6 +92,11 @@ IrTypeStruct *IrTypeStruct::get(StructDeclaration *sd) {
     builder.addTailPadding(sd->structsize);
     isaStruct(t->type)->setBody(builder.defaultTypes(), builder.isPacked());
     t->varGEPIndices = builder.varGEPIndices();
+
+    if (getTypeAllocSize(t->type) != sd->structsize) {
+      error(sd->loc, "ICE: struct IR size does not match the frontend size");
+      fatal();
```
JohanEngelen, line 98: I presume this errors for the testcase of #4734 ? (without the fix of this PR)

kinke: Yes, at least without assertions (I've checked the vanilla-LLVM jobs), I think those were hit earlier.

kinke:
Tested at Symmetry (hence the beta2 release - not requiring a fork any longer), looking good.

PR #4738, opened 2024-08-23
Add "Posix" global version conditional for Emscripten targets

ryuukk:
I just had this thought after seeing this: https://forum.dlang.org/post/odqtqqravqlfcqjgfywo@forum.dlang.org

Haven't tested yet, but i submit the PR otherwise i'll forget

And i can't compile ldc from source:

`cmake -G Ninja ../  -DCMAKE_BUILD_TYPE=Release  -DCMAKE_INSTALL_PREFIX=$PWD/../install-ldc`
```
/usr/bin/ld: cannot find -lLLVMSymbolize: No such file or directory
```

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -752,11 +752,25 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("D_HardFloat");
     break;
   case llvm::Triple::wasm32:
+    VersionCondition::addPredefinedGlobalIdent("WebAssembly");
+    if (triple.getOS() == llvm::Triple::Emscripten) {
+      // Emscripten reimplements Posix APIs
+      VersionCondition::addPredefinedGlobalIdent("linux");
+      VersionCondition::addPredefinedGlobalIdent("Posix");
+      VersionCondition::addPredefinedGlobalIdent("CRuntime_Musl");
+      VersionCondition::addPredefinedGlobalIdent("X86_Any");
+      VersionCondition::addPredefinedGlobalIdent("X86");
+    }
+    break;
   case llvm::Triple::wasm64:
     VersionCondition::addPredefinedGlobalIdent("WebAssembly");
     if (triple.getOS() == llvm::Triple::Emscripten) {
       // Emscripten reimplements Posix APIs
+      VersionCondition::addPredefinedGlobalIdent("linux");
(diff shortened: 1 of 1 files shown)


driver/main.cpp
```diff
@@ -752,8 +752,26 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("D_HardFloat");
     break;
   case llvm::Triple::wasm32:
+    VersionCondition::addPredefinedGlobalIdent("WebAssembly");
+    if (triple.getOS() == llvm::Triple::Emscripten) {
+      // Emscripten reimplements Posix APIs
+      VersionCondition::addPredefinedGlobalIdent("linux");
+      VersionCondition::addPredefinedGlobalIdent("Posix");
+      VersionCondition::addPredefinedGlobalIdent("CRuntime_Musl");
```
JohanEngelen, line 760: is this correct? Musl?  that is more than just posix.

driver/main.cpp
```diff
@@ -752,8 +752,26 @@ void registerPredefinedTargetVersions() {
...
+      VersionCondition::addPredefinedGlobalIdent("X86_Any");
```
JohanEngelen, line 761: X86 for sure is wrong

driver/main.cpp
```diff
@@ -752,8 +752,26 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("D_HardFloat");
     break;
   case llvm::Triple::wasm32:
+    VersionCondition::addPredefinedGlobalIdent("WebAssembly");
+    if (triple.getOS() == llvm::Triple::Emscripten) {
+      // Emscripten reimplements Posix APIs
+      VersionCondition::addPredefinedGlobalIdent("linux");
```
JohanEngelen, line 758: `linux` probably also incorrect?

driver/main.cpp
kinke: Hmm interesting. Do you have a link to confirm this? I found https://github.com/emscripten-core/emscripten/pull/13006 which seems to support this.

The proper place to add the musl-compatible predefined versions is https://github.com/ldc-developers/ldc/blob/dd0ff1ee838437972ee05a00b691a5e5354f6943/driver/main.cpp#L921-L923.

kinke:
Superseded by #4750.

PR #4742, opened 2024-08-27
Support building phobos against a system copy of zlib

the-horo:
Many Linux distributions including [Debian](https://wiki.debian.org/UpstreamGuide#No_inclusion_of_third_party_code), [Fedora](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling), and, [Gentoo](https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies), have a policy of unbundling 3rd party dependencies like zlib from software projects. This PR implements support for this so that distributions don't have to implement the same logic individually.

Currently Debian does [remove](https://salsa.debian.org/d-team/ldc/-/blob/93f0e1ea7893eaee408ac9d3c9716d1c31a87bd3/debian/patches/01_no-zlib-embed.patch) the zlib code but they leave the static variants of libphobos with undefined symbols to zlib functions making `-link-defaultlib-shared=false` require the additional `-lz` switch when specified by users. Fedora tried a [patch](https://bugzilla.redhat.com/show_bug.cgi?id=1102856#c8) but it resulted in the build being broken so it has been reverted. Gentoo doesn't do anything currently but I am making this PR instead.

These changes are all behind the `-DPHOBOS_SYSTEM_ZLIB=TRUE` cmake argument so the current code will continue to behave the same and the dependency on zlib is strictly opt-in. A possible unwanted interaction is that, when cross-compiling, ldc2 will try to link `-lz` just like it would when building for the build host but zlib may not exist on the target system. The only solution for this is specifying `-defaultlib= -L-l:libphobos2-ldc.a -L-l:libdruntime-ldc.a` which is a mouthful but it is either this or complicating the linker code more.

An implementation downside of the code is that the phobos config option is placed inside the root CMakeLists.txt because the compiler needs to know whether it needs to links zlib when building statically and, since it is built with a custom_target its target compile arguments can not be modified through something like `target_compile_definitions` after the runtime subdirectory has been included.

What this PR doesn't support is `MULTILIB=ON` and `PHOBOS_SYSTEM_ZLIB=TRUE` simultaneously because cmake, or any other build system that I know of, doesn't allow to specify the ABI when looking for a package. The standard way of compiling for multiple ABIs is to setup a different build directory and configuring cmake for that ABI which is done by setting PKG_CONFIG_PATH to point to the correct libdir and setting environment variables for the toolchain similar to `CC="gcc -m32" CXX="g++ -m32" DMD="ldmd -m32"`. The current cmake code worked for this with only minor changes but I didn't dive in too deeply yet.

I've made plenty of last minute improvements that ended up breaking something so it's fine if this PR waits for the new upcoming release to happen first and then be merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - LLVM for prebuilt packages bumped to v18.1.8 (incl. macOS arm64). (#4712)
 - Android: NDK for prebuilt package bumped from r26d to r27. (#4711)
 - ldc2.conf: %%ldcconfigpath%% placeholder added - specifies the directory where current configuration file is located. (#4717)
+- Add support for building against a system copy of zlib through `-DPHOBOS_SYSTEM_ZLIB=ON`. (#4742)
 
 #### Platform support
 
@@ -923,7 +924,7 @@
 - Misc. debuginfo issues, incl. adaptations to internal LLVM 5.0 changes: (#2315)
   - `ref` parameters and closure parameters declared with wrong address and hence potentially showing garbage.
   - Win64: parameters > 64 bit passed by value showing garbage.
-  - Win64: debuginfos for closure and nested variables now finally available starting with LLVM 5.0. 
+  - Win64: debuginfos for closure and nested variables now finally available starting with LLVM 5.0.
 - LLVM error `Global variable initializer type does not match global variable type!` for `T.init` with explicit initializers for dominated members in nested unions. (#2108)
 - Inconsistent handling of lvalue slicees wrt. visible side-effects of slice lower/upper bound expressions. (#1433)
 - Misc. dcompute issues. (#2195, #2215)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,9 @@ set(LDC_ENABLE_ASSERTIONS "${LLVM_ENABLE_ASSERTIONS}" CACHE BOOL "Enable LDC ass
 # Allow user to specify mimalloc.o location, to be linked with `ldc2` only
 set(ALTERNATIVE_MALLOC_O   ""     CACHE STRING "If specified, adds ALTERNATIVE_MALLOC_O object file to LDC link, to override the CRT malloc.")
 
+# Most linux distributions have a policy of not bundling dependencies like zlib
+set(PHOBOS_SYSTEM_ZLIB OFF CACHE BOOL "Use system zlib instead of Phobos' vendored version")
+
 if(D_VERSION EQUAL 1)
     message(FATAL_ERROR "D version 1 is no longer supported.
 Please consider using D version 2 or checkout the 'd1' git branch for the last version supporting D version 1.")
@@ -290,6 +293,9 @@ if(SANITIZE)
     endif()
 endif()
 append("${SANITIZE_CXXFLAGS}" LDC_CXXFLAGS)
+if(PHOBOS_SYSTEM_ZLIB)
+    append("-DPHOBOS_SYSTEM_ZLIB" LDC_CXXFLAGS)
+endif()
 # LLVM_CXXFLAGS may contain -Werror which causes compile errors with dmd source
 string(REPLACE "-Werror " "" LLVM_CXXFLAGS ${LLVM_CXXFLAGS})
 if (UNIX AND NOT "${LLVM_LDFLAGS}" STREQUAL "")
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -583,6 +583,10 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
   for (const auto &name : defaultLibNames) {
     args.push_back("-l" + name);
   }
+#ifdef PHOBOS_SYSTEM_ZLIB
+  if (!defaultLibNames.empty() && !linkAgainstSharedDefaultLibs())
+      args.push_back("-lz");
+#endif
 
   // libs added via pragma(lib, libname)
   for (auto ls : global.params.linkswitches) {
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -61,6 +61,10 @@ if (RT_SUPPORT_SANITIZERS)
     list(APPEND D_FLAGS -d-version=SupportSanitizers)
 endif()
 
+if(PHOBOS_SYSTEM_ZLIB)
+    message(STATUS "-- Building PHOBOS against system zlib")
+endif()
+
 # Auto-detect TARGET_SYSTEM from host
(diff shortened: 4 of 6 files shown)


the-horo:
I have now seen that https://github.com/ldc-developers/ldc/pull/4716 already existed, but compared to it, this PR correctly handles linking to static phobos and it doesn't affect the druntime libraries at the cost of slightly changing the compiler linker code.

the-horo:
I think it would be beneficial to add a changelog entry and a CI run for this option. Is circleci the preferred runner?

kinke:
Thanks for doing this; LGTM after a superficial glance. Yeah some CI coverage would be great; CircleCI is just kept because it doesn't hurt/get in the way; Cirrus is unfortunately not a sufficient free option anymore, so in essence it's all GitHub's own CI now, they managed to kill all competitors (for open-source projects) unfortunately. You could e.g. select one of the jobs in https://github.com/ldc-developers/ldc/blob/master/.github/workflows/supported_llvm_versions.yml; we use them for testing some CMake permutations too.

runtime/CMakeLists.txt
```diff
@@ -61,6 +61,10 @@ if (RT_SUPPORT_SANITIZERS)
     list(APPEND D_FLAGS -d-version=SupportSanitizers)
 endif()
 
+if(PHOBOS_SYSTEM_ZLIB)
```
kinke, line 64: Ideally, we'd bake this into the `ldc-build-runtime` CMake invocations too, in https://github.com/ldc-developers/ldc/blob/0cc531ee92160407aff4835cac7197c6354ac614/runtime/ldc-build-runtime.d.in#L152-L157.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - LLVM for prebuilt packages bumped to v18.1.8 (incl. macOS arm64). (#4712)
 - Android: NDK for prebuilt package bumped from r26d to r27. (#4711)
 - ldc2.conf: %%ldcconfigpath%% placeholder added - specifies the directory where current configuration file is located. (#4717)
+- Add support for building against a system copy of zlib through `-DPHOBOS_SYSTEM_ZLIB=ON` (#4742)
```
kinke, line 8: When you're revising this PR, please add a trailing full stop here.

CMakeLists.txt
```diff
@@ -149,6 +149,9 @@ set(LDC_ENABLE_ASSERTIONS "${LLVM_ENABLE_ASSERTIONS}" CACHE BOOL "Enable LDC ass
 # Allow user to specify mimalloc.o location, to be linked with `ldc2` only
 set(ALTERNATIVE_MALLOC_O   ""     CACHE STRING "If specified, adds ALTERNATIVE_MALLOC_O object file to LDC link, to override the CRT malloc.")
 
+# Most linux distributions have a policy of not bundling dependencies like zlib
+set(PHOBOS_SYSTEM_ZLIB OFF CACHE BOOL "Use the system copy of zlib when building phobos")
```
kinke, line 153: I'd probably slightly rephrase this, something like 'use system zlib instead of Phobos' vendored version'.

Note that the D bindings have hardcoded zlib version numbers which might be off then: https://github.com/dlang/phobos/blob/f7e523bc3c69506c3e7b4da7e78af93ed8547910/etc/c/zlib.d#L45-L47

The actual API has been pretty stable AFAIK, so I don't think this matters in practice.

the-horo: I've asked  in the Gentoo irc rooms what would are the appropriate version of zlib that we can use given that we hardcode the headers to a certain version and the response was that, given it's zlib, so long as we link against `libz.so.1` it should be fine.

If we want to restrict the version we can do that inside the `find_package` call so if we end up with an incompatible version of zlib in the future we can fail during configuration. For now I think that we shouldn't overly restrict the check unless we know for certain that a zlib version doesn't work.

the-horo:
Rebased onto master to fix the macos xcode failure.

kinke (approved):

kinke:
> I would suggest that instead of linking against -lz only for static builds, you link against -lz always (when the `-DPHOBOS_SYSTEM_ZLIB=TRUE` switch is passed). The cost is almost nothing since libphobos depends on it anyway.

Yeah that should be doable. FWIW, I didn't think anyone would really use the `etc.c.zlib` C bindings directly, but the `std.zlib` wrappers.

schveiguy:
> FWIW, I didn't think anyone would really use the `etc.c.zlib` C bindings directly

std.zlib is not good for streaming.

e.g. in iopipe, I have a `zip` and `unzip` iopipe which compresses or decompresses as you go. std.zip cannot do this.

in other words, std.zip does not expose well enough the capabilities of zlib.

schveiguy:
It's not a matter of opinion or intent. It's a matter of facts. zlib is included with phobos, including linkable symbols from etc.c. This means that a valid D install should be able to link without explicit -lz parameters.

Note that I'm not really in favor of this method of distribution for zlib. It's just, that's what we have now.

Contrast that with std.net.curl, which has no statically linkable symbols. If you use std.net.curl without providing a curl library to run against, your code will fail. That is a very different experience from std.zip.

If we did not include zlib, then users of std.zip would be required to link against -lz. I'm perfectly fine if that's the way we want to go. I've even advocated for it! But what I don't agree with is a state where some compiler distributions have made it so you cannot link applications which build just fine with other distributions of the same compiler. This leads to very puzzling bug reports, without much evidence of why it is happening. That is my experience.

This entire PR is evidence of the requirement! Without this mechanism, you had to add -lz for static builds. Now, the compiler does it for you. Why can't the compiler do this for builds against dynamic phobos?

driver/linker-gcc.cpp
```diff
@@ -583,6 +583,10 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
   for (const auto &name : defaultLibNames) {
     args.push_back("-l" + name);
   }
+#ifdef PHOBOS_SYSTEM_ZLIB
+  if (!defaultLibNames.empty() && !linkAgainstSharedDefaultLibs())
```
kinke, line 587: IIUC, to enable what Steven wants, we'd just remove the `!linkAgainstSharedDefaultLibs()` condition here. Meaning that if the compiler was configured with PHOBOS_SYSTEM_ZLIB, and the user hasn't cleared the `-defaultlib` list, it implicitly adds `-lz` for linking on Posix.

On Windows, we'd actually have to export all zlib symbols from the Phobos DLL in order to enable user code to use the `etc.c.zlib` bindings directly when using `-link-defaultlib-shared`.

the-horo: https://github.com/ldc-developers/ldc/pull/4963

Issue #3901, opened 2022-01-16
CMake error on macOS Monterey, library not found for -lpthread


JohanEngelen:
The first time running cmake on macOS Monterey, I see this error:
```
CMake Error at cmake/Modules/ExtractDMDSystemLinker.cmake:42 (message):
  Failed to link empty D program using
  '/Users/johan/dcompilers/ldc-1.27.1-johan/bin/ldmd2 -wi -link-debuglib':

  ld: library not found for -lpthread

  clang: error: linker command failed with exit code 1 (use -v to see
  invocation)

  Error: /Library/Developer/CommandLineTools/usr/bin/cc failed with status: 1
Call Stack (most recent call first):
  CMakeLists.txt:670 (include)
```

Simply rerunning cmake works and creates a fine build configuration for LDC (`ninja` works fine afterwards).

I've traced the problem down to the following:
There are two `cc` executables on macOS: `/usr/bin/cc` and `/Library/Developer/CommandLineTools/usr/bin/cc`. The first one, `/usr/bin/cc` is usually used by LDC, and correctly finds `libpthread`, so LDC can link executables without problem. The second one, `/Library/Developer/CommandLineTools/usr/bin/cc`, is _not_ able to find `libpthread` and therefore fails linking when invoked by LDC.
In the `execute_process` enviroment of CMake, the `CC` environment variable is set to `/Library/Developer/CommandLineTools/usr/bin/cc`; this is the C compiler detected by CMake (CMake outputs: `-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped`). LDC uses the `CC` variable to find the C compiler, thus the linking problem occurs.
I wonder if this is a problem in CMake, that should detect `/usr/bin/cc` instead of the other one.
I'm using CMake 3.22.1.

kinke:
Thanks, finally some insights for https://github.com/ldc-developers/ldc/pull/3871.

JohanEngelen:
I wonder if we can "fix" this by setting `CC` temporarily to nothing for `execute_process`...

JohanEngelen:
Solutions that work:
- `cmake -G Ninja -DCMAKE_C_COMPILER=/usr/bin/cc ......`
- `cmake -G Ninja -DD_COMPILER_FLAGS=-gcc=/usr/bin/cc .......`

These do **not** work:
- `CC=/usr/bin/cc cmake -G Ninja ......`
- `export CC=/usr/bin/cc`
[...]

PR #4743, opened 2024-08-31
Better support building the runtime separately

the-horo:
My use case for this is compiling the runtime for x86 on x86_64. `-DMULTILIB=ON` is not sufficient as I need to create a separate project in order to use a system copy of zlib like I've described in https://github.com/ldc-developers/ldc/pull/4742#issue-2489471216.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -1,6 +1,5 @@
-project(runtime C ASM)
-
 cmake_minimum_required(VERSION 3.4.3)
+project(runtime C ASM)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR})
 
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -337,7 +337,7 @@ if(LDC_EXE)
                     "$<$<BOOL:${SHARED_LIBS_SUPPORTED}>:${MULTILIB_DIR}>"
                     ${conf_path}.in
                     ${conf_path}
-            DEPENDS ${LDC_EXE} ${LDC_EXE_FULL}
+            DEPENDS ${LDC_EXE_FULL}
         )
         add_custom_command(OUTPUT ${install_conf_path} VERBATIM
             COMMAND ${PROJECT_PARENT_DIR}/tools/add-multilib-section.sh
@@ -347,7 +347,7 @@ if(LDC_EXE)
                     "$<$<BOOL:${SHARED_LIBS_SUPPORTED}>:${MULTILIB_INSTALL_DIR}>"
                     ${install_conf_path}.in
                     ${install_conf_path}
-            DEPENDS ${LDC_EXE} ${LDC_EXE_FULL}
+            DEPENDS ${LDC_EXE_FULL}
         )
         add_custom_target(add-multilib-section ALL
             DEPENDS ${conf_path} ${install_conf_path}
@@ -437,7 +437,9 @@ macro(dc src_files src_basedir d_flags output_basedir emit_bc all_at_once single
         list(APPEND dc_flags -flto=thin --output-bc)
     endif()
 
-    set(dc_deps ${LDC_EXE} ${LDC_EXE_FULL} ${GCCBUILTINS})
+    # dc_deps can only contain paths, otherwise cmake will ignore the dependency.
+    # See: https://github.com/ldc-developers/ldc/pull/4743#issuecomment-2323156173
+    set(dc_deps ${LDC_EXE_FULL} ${GCCBUILTINS})
     if(TARGET add-multilib-section)
         # Make sure the config files are available before invoking LDC.
         set(dc_deps ${dc_deps} add-multilib-section)
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -906,6 +906,7 @@ install(FILES ${GCCBUILTINS} DESTINATION ${INCLUDE_INSTALL_DIR}/ldc)
 #
 # Test targets.
 #
+enable_testing()
 
 # Build the "test runner" executables containing the druntime and Phobos unit
 # tests. They are invoked with the modules to test later.
--- a/runtime/DRuntimeIntegrationTests.cmake
+++ b/runtime/DRuntimeIntegrationTests.cmake
@@ -56,6 +56,8 @@ else()
     list(REMOVE_ITEM testnames uuid)
 endif()
 
+string(REPLACE ";" " " LDMD_CMD "${LDMD_EXE_FULL} ${D_EXTRA_FLAGS}")
+
 foreach(name ${testnames})
     foreach(build debug release)
         set(druntime_path_build ${druntime_path})
@@ -72,7 +74,7 @@ foreach(name ${testnames})
         )
         add_test(NAME ${fullname}
             COMMAND ${GNU_MAKE_BIN} -C ${PROJECT_SOURCE_DIR}/druntime/test/${name}
-                ROOT=${outdir} DMD=${LDMD_EXE_FULL} BUILD=${build}
+                ROOT=${outdir} DMD=${LDMD_CMD} BUILD=${build}
                 DRUNTIME=${druntime_path_build} DRUNTIMESO=${shared_druntime_path_build}
                 SHARED=1 ${cflags_base} ${linkdl}
         )


the-horo:
It looks like cmake is not adding the dependency on `LDC_EXE` properly. The quick fix is to add back `LDC_EXE_FULL` as a dependency when building the runtime but I want to figure out why cmake is misbehaving.

the-horo:
I've reduced it to:
```cmake
cmake_minimum_required(VERSION 3.27)
project("compiler" LANGUAGES CXX)

set(full "${PROJECT_BINARY_DIR}/bin/mycompiler")

if(TRUE)
    # fails
    add_custom_command(OUTPUT "${full}"
	COMMAND g++ -o "${full}" "${PROJECT_SOURCE_DIR}/mycompiler.cpp"
	DEPENDS mycompiler.cpp
    )
    add_custom_target(mycompiler ALL DEPENDS "${full}")
else()
    # works
    add_executable(mycompiler "mycompiler.cpp")
    set_target_properties(mycompiler PROPERTIES
	RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin")
endif()

add_custom_command(OUTPUT foo.o
    COMMAND "${full}" -o foo.o foo.cpp
    DEPENDS mycompiler
    DEPENDS_EXPLICIT_ONLY # this thing breaks
)
add_library(foo STATIC foo.o)
set_target_properties(foo PROPERTIES LINKER_LANGUAGE CXX)

```

the-horo:
This seems to be the intended behavior. From [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html) docs:
> DEPENDS
>
> Specify files on which the command depends. Each argument is converted to a dependency as follows:
>
> 1. If the argument is the name of a target (created by the [add_custom_target()](https://cmake.org/cmake/help/latest/command/add_custom_target.html#command:add_custom_target), [add_executable()](https://cmake.org/cmake/help/latest/command/add_executable.html#command:add_executable), or [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library) command) a target-level dependency is created to make sure the target is built before any target using this custom command. Additionally, if the target is an executable or library, a file-level dependency is created to cause the custom command to re-run whenever the target is recompiled.

This wording clearly states that the dependency of `add_custom_command` aren't for the `add_custom_command` target, they are only transitively applied to _other_ targets that use the output of the command. To put in bluntly:

What we need:
1. generate foo.o from foo.d (depends on LDC)
2. archive foo.o into libruntime.a

What cmake does:
1. generate foo.o from foo.d
2. archive foo.o into libdruntime.a (depends on LDC)

Since this seems to be the intended behavior the `add_custom_command` need to be adjusted to properly specify dependencies. We can either drop `CMAKE_ADD_CUSTOM_COMMAND_DEPENDS_EXPLICIT_ONLY` or specify the dependencies as absolute paths instead of target names. We probably want the latter.

kinke:
Are you aware of https://wiki.dlang.org/Building_LDC_runtime_libraries? The `ldc-build-runtime` tool is just sugar around CMake, but I doubt we really need to support cmdline flags in `LDC_EXE_FULL`; one of the existing, historically named vars like `D_EXTRA_FLAGS` should do.

the-horo:
> Are you aware of https://wiki.dlang.org/Building_LDC_runtime_libraries? The `ldc-build-runtime` tool is just sugar around CMake, but I doubt we really need to support cmdline flags in `LDC_EXE_FULL`; one of the existing, historically named vars like `D_EXTRA_FLAGS` should do.

I was aware of the tool and the fact that it was mostly a wrapper over cmake so I would prefer to use cmake directly. When I first tried to build the runtime for multilib directly the building itself worked but I hit an issue with the test runners and the druntime integration tests. I think I tried to do `-DLDC_EXE_FULL=<ldc2>;-m32` but I didn't try `-DLDC_EXE_FULL=<ldc2> -DD_EXTRA_ARGS=-m32` which does work fine for the test runners.

The only thing that doesn't work are the druntime integration tests so we either have pass `D_EXTRA_FLAGS` to ldmd there too or support that `-DLDMD_EXE_FULL` contain spaces like I originally proposed. Since the rest of the code uses `D_EXTRA_FLAGS` I think it's better to stay consistent.

As a conclusion, is it fine if to keep the non-flag related changes like the addition of `enable_testing()` and to make sure that `D_EXTRA_FLAGS` are respected in `DRuntimeIntegrationTests.cmake`?

the-horo:
I can also improve the code and add support for `-DMULTILIB=ON` for the druntime tests since only the native build is tested currently.

the-horo:
> If we wanted to include the 32-bit druntime integration tests for MULTILIB builds, we'd need to add some test name suffix to make them unique.

The druntime test makefiles already support `ROOT` for the output files so we should only need to add a suffix to the test names and be done with it. We would also need the special case the coverage test but it may also be possible to modify it upstream to use separate copies of the test files instead of modifying the same one.

> > My insistence on allowing LDC_EXE_FULL to contain arguments is because this is how everyone else does it, cmake included. It does accomplish the same thing as the current LDC_EXE_FULL and D_EXTRA_FLAGS, i.e. a command and arguments, but it does so in a less error prone way since it's impossible to forget to take into account the necessary command line arguments.
> 
> What I don't like is that we already have a variable for extra flags, and having to take account cmdline options in a compiler var. I trust this is some common concept on Posix, but I don't really like that extra cognitive overhead and CMake special cases. Maybe coz I grew up with Windows, where such space-separated flags are unthinkable. ;)

Yeah, I can see why you wouldn't want your `Program Files` to be split on windows.

the-horo:
```diff
-set(LDMD_CMD "${LDMD_EXE_FULL} ${D_EXTRA_ARGS}")
+string(REPLACE ";" " " LDMD_CMD "${LDMD_EXE_FULL} ${D_EXTRA_FLAGS}")
```

1. I was using the wrong variable name.
2. `D_EXTRA_FLAGS` is a `;`-delimited list, not space separated.

runtime/DRuntimeIntegrationTests.cmake
```diff
@@ -56,6 +56,8 @@ else()
     list(REMOVE_ITEM testnames uuid)
 endif()
 
+string(REPLACE ";" " " LDMD_CMD "${LDMD_EXE_FULL} ${D_EXTRA_FLAGS}")
```
kinke, line 59: Your use case would be setting `D_EXTRA_FLAGS=-m32`, right? Just to understand. This step should probably be skipped on Windows [but that divergence would be a bit ugly IMO], to avoid spaces in the path and spaces as flag separators.

AFAICT, the multilib case would be designed to be used with the MODEL make variable: https://github.com/ldc-developers/ldc/blob/1f0c36ec6caaad4ce63dad86b73259cc5a0e41e0/runtime/druntime/test/common.mak#L33

The MODEL would make it into DFLAGS: https://github.com/ldc-developers/ldc/blob/1f0c36ec6caaad4ce63dad86b73259cc5a0e41e0/runtime/druntime/test/common.mak#L41

and CFLAGS_BASE (which we override for non-MSVC targets via CMake `cflags_base` though): https://github.com/ldc-developers/ldc/blob/1f0c36ec6caaad4ce63dad86b73259cc5a0e41e0/runtime/druntime/test/common.mak#L34

How do you currently handle the C side for the separate 32-bit build? Setting a CC env var incl. an `-m32` flag?

runtime/CMakeLists.txt
```diff
@@ -437,7 +436,9 @@ macro(dc src_files src_basedir d_flags output_basedir emit_bc all_at_once single
         list(APPEND dc_flags -flto=thin --output-bc)
     endif()
 
-    set(dc_deps ${LDC_EXE} ${LDC_EXE_FULL} ${GCCBUILTINS})
+    # dc_deps can only contain paths, otherwise cmake will ignore the dependency.
+    # See: https://github.com/ldc-developers/ldc/pull/4743#issuecomment-2323156173
+    set(dc_deps ${LDC_EXE_FULL} ${GCCBUILTINS})
```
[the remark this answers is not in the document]
the-horo: > And yet we add a `add-multilib-section` target **name** 3 lines below. According to

> > If the argument is the name of a target (created by the [add_custom_target()](https://cmake.org/cmake/help/latest/command/add_custom_target.html#command:add_custom_target), [add_executable()](https://cmake.org/cmake/help/latest/command/add_executable.html#command:add_executable), or [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library) command) a target-level dependency is created to make sure the target is built before any target using this custom command.
> 
> this should still make sure the target is created before building any dependent. The (additional) path dependency should take care of the 2nd part wrt. rebuilds IIUC:

From looking at `build.ninja`, the `add-multilib-section` target is added as a dependency to the final libraries, like `lib64/libdruntime-ldc-shared.so.109.1`. I think this is fine, as the libraries are compiled with `-conf=` so they don't actually need the config file to be updated. In this case I think the weird cmake behavior does what we want: "update the config file before declaring the library built", it just does it a little bit later in the process.

> > Additionally, if the target is an executable or library, a file-level dependency is created to cause the custom command to re-run whenever the target is recompiled.
> 
> I guess they actually mean 'if the target is an add_executable or add_library target', so that one has to specify the path to the `add_custom_target` executable manually for proper rebuilds.

I'm reading this the other way around:
- if the target is an executable or library, cmake adds a file-path dependency automatically => we don't need to specify a full path as a dependency.
- otherwise, cmake does nothing => we need to specify the full path if we need that dependency to be built before this `add_custom_command` or `add_custom_target`.

runtime/DRuntimeIntegrationTests.cmake
the-horo: > Your use case would be setting `D_EXTRA_FLAGS=-m32`, right? Just to understand. This step should probably be skipped on Windows [but that divergence would be a bit ugly IMO], to avoid spaces in the path and spaces as flag separators.

Yes, only `-m32` for now. Though it is technically possible, in some future, for this to use other `-m<abi>` flags for different architectures.

> AFAICT, the multilib case would be designed to be used with the MODEL make variable:
> 
> https://github.com/ldc-developers/ldc/blob/1f0c36ec6caaad4ce63dad86b73259cc5a0e41e0/runtime/druntime/test/common.mak#L33
> 
> The MODEL would make it into DFLAGS:
> 
> https://github.com/ldc-developers/ldc/blob/1f0c36ec6caaad4ce63dad86b73259cc5a0e41e0/runtime/druntime/test/common.mak#L41
> 
> and CFLAGS_BASE (which we override for non-MSVC targets via CMake `cflags_base` though):
> 
> https://github.com/ldc-developers/ldc/blob/1f0c36ec6caaad4ce63dad86b73259cc5a0e41e0/runtime/druntime/test/common.mak#L34
> 
> How do you currently handle the C side for the separate 32-bit build? Setting a CC env var incl. an `-m32` flag?

These variables are set already by the Gentoo helper functions so I didn't have to interact with them but, if I followed the functions right:
- https://github.com/gentoo/gentoo/blob/e1c6cdf0e713cd916c7880296b3af32834fdb9c0/eclass/multilib.eclass#L544 sets `CC` to contain the `-m` flag.
- https://github.com/gentoo/gentoo/blob/e1c6cdf0e713cd916c7880296b3af32834fdb9c0/eclass/cmake.eclass#L468-L477 example: `set(CMAKE_C_COMPILER "x86_64-pc-linux-gnu-gcc;-m32 -mfpmath=sse")`

Here's what the logs contain:
```
460: x86_64-pc-linux-gnu-gcc -m32 -mfpmath=sse -Wall -Wl,-rpath,/var/tmp/portage/dev-lang/ldc2-1.40.0_beta3-r1/work/ldc-1.40.0-beta3-src_build-abi_x86_32.x86/lib -L/var/tmp/portage/dev-lang/ldc2-1.40.0_beta3-r1/work/ldc-1.40.0-beta3-src_build-abi_x86_32.x86/lib -O3 -o/var/tmp/portage/dev-lang/ldc2-1.40.0_beta3-r1/work/ldc-1.40.0-beta3-src_build-abi_x86_32.x86/druntime-test-shared-release/linkDR  src/linkDR.c /var/tmp/portage/dev-lang/ldc2-1.40.0_beta3-r1/work/ldc-1.40.0-beta3-src_build-abi_x86_32.x86/lib/libdruntime-ldc-shared.so -ldl  -pthread
```
So yeah, setting `CC` seems like what's doing the trick, including for the druntime tests that are called through cmake. Note that there isn't a single `-m32` flag, `-mfpmath=sse` is also present so I would like the solution to allow any combination of flags.

Setting `MODEL_FLAG=${D_EXTRA_FLAGS}` seems like it would do the trick:
- allows for arbitrary flags to be passed through
- doesn't interfere with space-splitting `CC`

But can you explain something for me? Does `gmake` on windows not perform space expansion on variables, regardless of where they appear, or is there some special case only for when the variable appears as the first thing in a command. For example:
```make
main: main.cpp
	$(CC) $(CFLAGS) -o main main.cpp
```
`CFLAGS` gets split by spaces. `CC` is also split on linux. Does windows not split `CC` but splits `CFLAGS`?

kinke: I seem to recall/assume (haven't tested it now!) that on Windows, you are supposed to use explicit quotes, something like `make CC="my cc.exe"`. IIRC, make then uses those quotes as-is, so that `$(CC) -c` gets expanded to `"my cc.exe" -c`. This *should* allow adding the D_EXTRA_FLAGS to `LDMD_CMD` on Windows too. If a poor soul had to split it up into executable and flags, that'd be a PITA of course, as trivial splitting by space wouldn't work of course.

runtime/CMakeLists.txt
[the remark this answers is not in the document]
kinke: > >   I guess they actually mean 'if the target is an add_executable or add_library target', so that one has to specify the path to the add_custom_target executable manually for proper rebuilds.
>
> I'm reading this the other way around:

I'm pretty sure the 'target' they refer to is the dependency being added, i.e., `LDC_EXE` (expands to `ldc2`) here. Which can either be an `add_executable` or an `add_custom_target`, depending on `LDC_LINK_MANUALLY`. So in case the ldc2 executable is a custom target, we additionally add `LDC_EXE_FULL` so that the runtimes stuff is rebuilt whenever the compiler is.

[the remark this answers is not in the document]
the-horo: Ah, yes, I understand it the same way. Maybe I failed to word it right but, in my defense, it is a unintuitive behavior cmake is exhibiting.

> ... so that the runtimes stuff is rebuilt whenever the compiler is.

+ that the compiler is built, if not present, before the runtime build commands try to invoke it.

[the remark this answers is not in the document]
kinke: Fully agreed wrt. unintuitive. Probably related: we had CI issues with the `make` backend, IIRC some targets not waiting for the prerequisites to be finished. Working fine with `ninja` though.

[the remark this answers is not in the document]
kinke: Okay after another re-read, I think I got your point now - that the `LDC_EXE` target name dependency does NOT apply to the custom command getting this dependency directly, but only to the dependents of that custom command. So you're absolutely right, the target-name dependency is insufficient, we need the path dependency to make it apply to the custom command directly, and not just for rebuilds. Thanks for digging!

runtime/DRuntimeIntegrationTests.cmake
the-horo: Amazing! I don't have a windows machine so I trust you on these.

So, what's the final solution we go with?
1. allow `LDMD_CMD` co contain arguments
2. use `MODEL_FLAG` to pass `D_EXTRA_FLAGS`

1 is unintutive on windows because of the comamnd magic, 2 is unintuitive everywhere since the variables don't really seem related.

If the tests are adjusted to also test for 32 bit with `-DMULTILIB=ON` that would require passing -m32/-m64 somewhere. We could use `MODEL` here or just override `MODEL_FLAG` again. I think the final result would look like:

1 + MODEL:
```cmake
# do_druntime_tests invokes the tests with make MODEL=${1}
do_druntime_tests("")
if(MULTILIB)
  do_druntime_tests("${HOST_BITNESS}")
endif()
```

1 + MODEL_FLAG:
```cmake
# do_druntime_tests invokes the tests with make MODEL_FLAG=${1}
do_druntime_tests("")
if(MULTILIB)
  do_druntime_tests("-m${HOST_BITNESS}")
endif()
```

2 + MODEL_FLAG:
```cmake
# do_druntime_tests invokes the tests with make MODEL_FLAG="${D_EXTRA_FLAGS} ${1}"
do_druntime_tests("")
if(MULTILIB)
  do_druntime_tests("-m${HOST_BITNESS}")
endif()
```

the-horo: I have to, sometime, change the dmd makefiles to allow taking flags from the command line instead of relying on hacks like using MODEL_FLAG, as Gentoo policy dictates it. If I get to upstream these changes we should be able to do `make MY_D_FLAGS=${D_EXTRA_FLAGS}` and use either of the model flag logic so maybe it's worth taking this future possibility into consideration.

kinke: For `MULTILIB=ON` builds, we probably have to go with MODEL or MODEL_FLAG, and also account for that in our `cflags_base` override (or handle it in `common.mak`, LDC-specific), so that `-m32` is used for the (64-bit) C[++] compilers too.

For your separate runtime build, I'm not sure, especially if `-m32` might not suffice, `-mabi` etc. potentially coming up, which might diverge for both D and C[++] compilers etc.. AFAIU, for the C side you currently rely on the CC env var defined outside of CMake; IIRC, CMake might tamper with that for invoked child processes (https://github.com/ldc-developers/ldc/issues/3901#issue-1105195277). So I'm really not sure we can marry both use cases (MULTILIB + special separate runtime builds) via a single solution for both.

kinke: The C analogon to `D_EXTRA_FLAGS` is `RT_CFLAGS`. The latter could be included in `cflags_base` and forwarded to `common.mak`. For D, we could add a `DFLAGS_BASE` make variable, populating it with the `D_EXTRA_FLAGS` from CMake, and use them for the final `DFLAGS` in `common.mak`.

This way, we'd simply forward all D_EXTRA_FLAGS and RT_CFLAGS as {D,C}FLAGS_BASE make variables for the druntime integration tests. This could then even be combined with `MULTILIB=ON`, which would just add the `-m{32,64}` flags to both.

And you'd 'have' to split up `CC` again, passing the flags as `RT_CFLAGS` instead. Maybe you don't need a CC env var at all anymore. And there'd be at least one advantage: the CC env var **at ninja build time** isn't required to be still set to the same executable+flags combo as **at cmake-config-time**.

the-horo: > For `MULTILIB=ON` builds, we probably have to go with MODEL or MODEL_FLAG, and also account for that in our `cflags_base` override (or handle it in `common.mak`, LDC-specific), so that `-m32` is used for the (64-bit) C[++] compilers too.
> 
> For your separate runtime build, I'm not sure, especially if `-m32` might not suffice, `-mabi` etc. potentially coming up, which might diverge for both D and C[++] compilers etc.. AFAIU, for the C side you currently rely on the CC env var defined outside of CMake; IIRC, CMake might tamper with that for invoked child processes ([#3901 (comment)](https://github.com/ldc-developers/ldc/issues/3901#issue-1105195277) - **edit**: oh wait, that was for `execute_process` calls at cmake-config-time, not commands run at build-time). So I'm really not sure we can marry both use cases (MULTILIB + special separate runtime builds) via a single solution for both.

Well, with this PR applied I can build the runtime as a separate project and run the tests just fine. `-DMULTILIB=ON` still works the way it did so nothing's lost. If we then want to support `druntime-test-*_32`, so long as the changes are put behind:
```cmake
if(MULTILIB)
   run_druntime_tests_with("CFLAGS_BASE=-m32" "CXXFLAGS_BASE=-m32" "DFLAGS_BASE=-m32")
endif()
```
and the previous code is kept as is, without having any arguments be added to the default abi build, my projects will still work.

> The C analogon to `D_EXTRA_FLAGS` is `RT_CFLAGS`. The latter could be included in `cflags_base` and forwarded to `common.mak`. For D, we could add a `DFLAGS_BASE` make variable, populating it with the `D_EXTRA_FLAGS` from CMake, and use them for the final `DFLAGS` in `common.mak`.
> 
> This way, we'd simply forward all D_EXTRA_FLAGS and RT_CFLAGS as {D,C}FLAGS_BASE make variables for the druntime integration tests. This could then even be combined with `MULTILIB=ON`, which would just add the `-m{32,64}` flags to both.
> 
> And you'd 'have' to split up `CC` again, passing the flags as `RT_CFLAGS` instead. Maybe you don't need a CC env var at all anymore. And there'd be at least one advantage: the CC env var **at ninja build time** isn't required to be still set to the same executable+flags combo as **at cmake-config-time**.

Let me try to summarize our problem:

What we need:
- run druntime tests that require to compile C, C++, and, D sources

What we have:
- `LDMD_EXE_FULL` for our D compiler
- `D_EXTRA_FLAGS` for critical and non-critical D flags. Critical flags I'd call flags that are needed for a successful compilation, like `-m32`, or `-mtriple`, and non-critical I'd call flags like `-O2`.
- `RT_CFLAGS` flags for the C compiler (unspecified if critical or non-critical)
- `CMAKE_C_COMPILER` - a C compiler alongside its critical flags.
- `CMAKE_CXX_COPILER` - a C++ compiler alongside its critical flags.

If the druntime makefile would be well written, it should support `CC` and other compiler vars to be *any command*. I described the case for `CC="gcc -m32"` but I was also given the example where somebody may want to do `CC="zig cc"`. In that case one should be able to do `make CC=${CMAKE_C_COMPILER} CXX=${CMAKE_CXX_COMPILER} DC="${LDMD_EXE_FULL} ${D_EXTRA_FLAGS}"`. Here's a link of what gnu says is supposed to be standard: https://www.gnu.org/prep/standards/html_node/Command-Variables.html#Command-Variables. It doesn't say anything about whether `CC` is supposed to be supported to contain arguments unfortunately.

If we want to use separate variables for passing flags we should be able to do:
```
make CC=${CMAKE_C_COMPILER_EXE} CFLAGS_BASE=${CMAKE_C_COMPILER_ARGS} CXX=${CMAKE_CXX_COMPILER_EXE} CXXFLAGS_BASE=${CMAKE_CXX_COMPILER_ARGS} DMD=${LDMD_EXE_FULL} DFLAGS_BASE=${D_EXTRA_FLAGS}
```

This is a longer but we kinda have to do it if we want to respect the environment that cmake was configured with.

If we want to support stuff like `MODEL=32` the correct approach would be to keep the arguments above and add `MODEL=32` and have the `MODEL` value be inspected inside the makefile. The rules there should look like:
```make
rule: deps
	$(CC) -o rule deps $(INTERNAL_C_FLAGS) $(CFLAGS_BASE)
```
and `INTERNAL_C_FLAGS` should contain stuff like includes, link stuff, and the `MODEL` flag, if specified. This is how gnu says that the makefile should be written, minus the variable name for `CFLAGS_BASE`.

But if we end up needing to modify the makefiles because they don't allow us to specify variables then that change should be made upstream in dmd. Our only responsibility is properly calling make from cmake. This only requires us to respect `CMAKE_C_COMPILER`, `CMAKE_CXX_COMPILER`, and the other critical flags variable. All other things that we want to support should work on top of the variables above. The question is how to respect the variables.

If we require that `CC` incorporate the arguments that would make our code the cleanest and is the friendliest approach for users, so long as `CC` is split up properly. Your test concluded that it works but maybe there are make implementation or versions that don't support this.

The alternative is having dedicated flag variables. This is more code for us to write but probably more portable.

All in all this is pretty messy, having to properly translate variables from cmake and pass them to make. It's already late enough and I can barely process this issue any more but, if we want to be the safest and not risk breaking supposed systems where `CC` can't be expanded properly, we would need to rewrite the makefiles to accept compiler flags in some variable, for all 3 compilers. The fact that we can't do this yet is, as far as I see, a bug in the makefile. If it's a bug then we should fix it and that implies rewriting the druntime test files in dmd. That would also lead to rewriting the rest of the makefiles as the rest are similarly broken.

So fully respecting the cmake configure environment here and not relying on runtime variables is blocked until the dmd makefiles are fixed. So I think it's fine to temporarily keep the current behavior of respecting the `CC` environment variable since we didn't yet have any request that the configure time value be respected. For the makefile fixes and respecting the configure environment, and supporting `-DMULTILIB=ON` it would be best to open separate issues/PR. We don't need to solve everything at once.

the-horo: I'm looking over this again with a fresh mind.

I found out https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html:
> The variables used in implicit rules fall into two classes: those that are names of programs (like CC) and those that contain arguments for the programs (like CFLAGS). (The “name of a program” may also contain some command arguments, but it must start with an actual executable program name.) If a variable value contains more than one argument, separate them with spaces.

So there's no reason not to use `CC` to pass some arguments.

However, I just remembered something from my previous tests of this feature. cmake makes yet another brilliant decision by modifying the `CMAKE_C_COMPILER` that is passed during configure. Inside the cmake code `CMAKE_C_COMPILER` will only be an executable, the flags are stripped and hard embedded inside the `build.ninja` file. This makes it impossible to respect the flags because you can't inspect their values.

We still have the `RT_CFLAGS` variables and the like that we can respect. With a standard makefile it should be as easy as doing `make CFLAGS="${RT_CFLAGS}"` but we have to do with what we have so we would probably need to do `make USER_CFLAGS="${RT_CFLAGS}"` and have `common.mak` add that value to the internal `CFLAGS`.

I have two problems with this approach. Passing important flags in those variables means that we need to implement a variable for each compiler and that increases maintenance. Secondly, the whole point why I'm doing this is because there is already code written by other people who ensures that the standard cmake variables are set accordingly when doing multilib. Having non-standard variables that I need to set defeats the main point of doing all of this.

I think we should just suck up to this shortcoming of cmake and not implement a whole build system on top of cmake for C and C++ sources, i.e. making sure there are correct variables set and they are respected alongside the build and tests.

kinke: > I have two problems with this approach. Passing important flags in those variables means that we need to implement a variable for each compiler and that increases maintenance.

Only C (CFLAGS_BASE used for CXXFLAGS too) and D. Where it's not unlikely that the flags need to be adapted anyway, e.g. between gcc and LDC wrt. `-mattr` or `-mabi`. This was the missing piece for me, **having** to set a weird CC env var isn't really an option to me.

> Secondly, the whole point why I'm doing this is because there is already code written by other people who ensures that the standard cmake variables are set accordingly when doing multilib. Having non-standard variables that I need to set defeats the main point of doing all of this.

IIUC, you/Gentoo wouldn't need `RT_CFLAGS` but can happily live with a CC env var that needs to be persisted for the ctest invocations. Not something I would personally use or recommend, but if you're fine with passing `D_EXTRA_FLAGS` so that they make it to the druntime integration tests as DFLAGS_BASE too, perfect. Then for LDC's `common.mak`, we could simply ditch MODEL and MODEL_FLAGS, and go with {C,D}FLAGS_BASE, which provide enough flexibility I think and trivially enable MULTILIB=ON too.

kinke: I'm not sure we're still on the same page ;) - I can create a PR based on yours here to show what I mean (excl.  MULTILIB=ON support, not in the 1st step). The only compromise I have to make in order to support your external-CC use case is NOT propagating `CMAKE_C_COMPILER` as CC make var in the make cmdline (so that you can still feed it in externally), but that's definitely not the end of the world.

the-horo: > > I have two problems with this approach. Passing important flags in those variables means that we need to implement a variable for each compiler and that increases maintenance.
> 
> Only C (CFLAGS_BASE used for CXXFLAGS too) and D. Where it's not unlikely that the flags need to be adapted anyway, e.g. between gcc and LDC wrt. `-mattr` or `-mabi`. This was the missing piece for me, **having** to set a weird CC env var isn't really an option to me.

Yeah, keeping only `RT_CFLAGS` for both C++ and C arguments is probably fine. This is non-standard which I don't like but, so long as both the standard way of using `CMAKE_C_COMPILER` works, then I have nothing against also support additional methods of passing flags. A problem is that cmake is not good enough to allow us to simply use `CMAKE_C_COMPILER` which makes things unstraightforward.

> > Secondly, the whole point why I'm doing this is because there is already code written by other people who ensures that the standard cmake variables are set accordingly when doing multilib. Having non-standard variables that I need to set defeats the main point of doing all of this.
> 
> IIUC, you/Gentoo wouldn't need `RT_CFLAGS` but can happily live with a CC env var that needs to be persisted for the ctest invocations. Not something I would personally use or recommend, but if you're fine with passing `D_EXTRA_FLAGS` so that they make it to the druntime integration tests as DFLAGS_BASE too, perfect. Then for LDC's `common.mak`, we could simply ditch MODEL and MODEL_FLAGS, and go with {C,D}FLAGS_BASE, which provide enough flexibility I think and trivially enable MULTILIB=ON too.

Yes, if ldc provides simple variables like `RT_CFLAGS` and `RT_LFLAGS` to configure its build, on top of working with the flags cmake is configured with, then the easiest way to pass them along to the druntime tests is having the makefile respect variables like `DFLAGS_BASE` and `CFLAGS_BASE` just like you suggest. For the sake of standardization I would recommend that those variables are, instead, named `DFLAGS` and `CFLAGS` as that makes their intent very clear and is the standard way of handling flags.

This would require the druntime makefiles to be modified, with slight modifications for `CFLAGS_BASE` and more modifications for plainly using `CFLAGS` but it's not impossible. I've looked over some of them and I think some rewrite could be beneficial, especially because of how many times commands are written literally  instead of using something like suffix rules.

This doesn't mean that we need to drop `MODEL` or stuff like `BUILD`, the makefile can still work with them but also support taking `DFLAGS=-m32` on the command line.

All of this would solve the passage of flags, but that still leaves CC to be taken from the environment and, as you mentioned before, relying on the runtime value of the environment variable instead of the value cmake was configured with is a disservice to the users. If I rely on using the `CC` environment variable then that would prevent us from making a change such as invoking the druntime tests with:
```
make CC=${CMAKE_C_COMPILER} CFLAGS_BASE=${RT_CFLAGS} ...
```
because that would shadow the environment variable and, in my case, since I don't pass any flags in `RT_CFLAGS` I would get a miscompilation. This is, as far as I can see, a shortcoming of cmake so I won't insist of changing the ldc cmake files to support my use case. In the worst case I would need to add `-DRT_CFLAGS=$(get_abi_CFLAGS)` which isn't a deal breaker.

the-horo: So, how about we:
1. modify the druntime makefile to support being called with (the variables names can be adjusted):
  ```
  make CFLAGS_BASE=${OUR_CFLAGS} CXXFLAGS_BASE=${OUR_CXXFLAGS} ...
  ```
2. invoke the makefiles from cmake like:
  ```
  ${GNU_MAKE_BIN} CFLAGS=${RT_CFLAGS} CXXFLAGS=${RT_CFLAGS} DFLAGS=${D_EXTRA_FLAGS}
                  CC=${CMAKE_C_COMPILER} CXX=${CMAKE_CXX_COMPILER} DMD=${LDMD_EXE_FULL}
  ```
3. for future `MULTILIB=ON` we would do:
  ```
  ${GNU_MAKE_BIN} CFLAGS="${RT_CFLAGS} -m32" CXXFLAGS="${RT_CFLAGS} -m32" DFLAGS="${D_EXTRA_FLAGS} -m32"
                  CC=${CMAKE_C_COMPILER} CXX=${CMAKE_CXX_COMPILER} DMD=${LDMD_EXE_FULL}
  ```
While I did find that we could pass `-m32` inside of `CC` as far as make is concerned, this will break  on windows as `CC` is also used by ldc to do the linking and the value of the variable is only split on posix, so we can't do that.

Does this sound good enough to open as a separate issue, so as to have an ease discussion around the topic?

kinke: Please see https://github.com/ldc-developers/ldc/pull/4752, might be easier than talking here.

kinke: > This doesn't mean that we need to drop MODEL

I checked, and we cannot just yet:  https://github.com/ldc-developers/ldc/blob/1f0c36ec6caaad4ce63dad86b73259cc5a0e41e0/runtime/druntime/test/profile/Makefile#L44

=> https://github.com/ldc-developers/ldc/blob/master/runtime/druntime/test/profile/myprofilegc.log.linux.32.exp

kinke: > For the sake of standardization I would recommend that those variables are, instead, named DFLAGS and CFLAGS as that makes their intent very clear and is the standard way of handling flags.

AFAIK, it's common to have such CFLAGS_BASE, in order to only set optional extra/critical flags on demand. Specifying a make variable like this (*after* `make`, not as env vars) silently skips all regular assignments of CFLAGS_BASE in `common.mak`. The full list in CFLAGS etc. can then still be joined together in `common.mak`, based on (potentially custom) CFLAGS_BASE.

Similarly, `common.mak` currently also allows one to customize `CXXFLAGS_BASE` - it just defaults to `CFLAGS_BASE` in https://github.com/ldc-developers/ldc/blob/1f0c36ec6caaad4ce63dad86b73259cc5a0e41e0/runtime/druntime/test/common.mak#L56.

kinke: > Yes, if ldc provides simple variables like RT_CFLAGS and RT_LFLAGS to configure its build, on top of working with the flags cmake is configured with, then the easiest way to pass them along to the druntime tests is having the makefile respect variables like DFLAGS_BASE and CFLAGS_BASE just like you suggest.

[The non-super-consistently named CMake variables come from having a fat umbrella CMake build for LDC, where runtime and Phobos are just a subset of all builds. So we ended up with some ugly names for compiler-only D flags, runtime-only D flags etc. etc. They are usually empty, but designed exactly for such target-specific critical flags and/or custom flags.]

> In the worst case I would need to add -DRT_CFLAGS=$(get_abi_CFLAGS) which isn't a deal breaker.

The only potential problem I see with you specifying the C flags explicitly this way is that they would be duplicated when building the runtime libraries (due to Gentoo's CMake hook injecting the critical flags), and potentially also for the make integration tests if CC was still set with flags at test-time. The compiler might complain about some options being specified multiple times.

the-horo: > The only potential problem I see with you specifying the C flags explicitly this way is that they would be duplicated when building the runtime libraries (due to Gentoo's CMake hook injecting the critical flags), and potentially also for the make integration tests if CC was still set with flags at test-time. The compiler might complain about some options being specified multiple times.

I meant that I would pass`-DRT_CFLAGS=-m32` if the cmake code were changed to invoke `gmake` with `CC=${CMAKE_C_COMPILER}` to respect the configure-time value. In that case the environment variable is not in effect so `$(CC)` inside make will be the compiler binary without any arguments so there won't be any duplication with `RT_CFLAGS`.

kinke: Yeah that's the case we can handle. But how about the actual library build?

the-horo: Yeah, in that case the flags would be duplicated. This shouldn't be an issue and, even if it is, I can always just patch the cmake files to use a separate variable for passing abi-related CFLAGS and only respect in when invoking make, like `CC="${CMAKE_C_COMPILER} ${GENTOO_ABI_CFLAGS}"` since this should work on UNIXes.

This is slightly more work for me but, given the choices:
1. don't respect ${CMAKE_C_COMPILER} when calling the makefiles and rely on runtime CC
2. respect ${CMAKE_C_COMPILER}, even if it discards flags

I would prefer for ldc to use 2, because, if it breaks, it would be easier to diagnose even if that entails me possibly carrying one more patch downstream.

PR #4748, opened 2024-09-07
ELF: Emit (most) instantiated symbols in COMDATs

kinke:


gen/tollvm.cpp
```diff
@@ -259,7 +259,13 @@ LLGlobalValue::LinkageTypes DtoLinkageOnly(Dsymbol *sym) {
...
 
 LinkageWithCOMDAT DtoLinkage(Dsymbol *sym) {
-  return {DtoLinkageOnly(sym), needsCOMDAT()};
+  const auto linkage = DtoLinkageOnly(sym);
+  const bool inCOMDAT = needsCOMDAT() ||
+                        // ELF needs some help for ODR linkages:
+                        // https://github.com/ldc-developers/ldc/issues/3589
+                        (linkage == templateLinkage &&
+                         global.params.targetTriple->isOSBinFormatELF());
+  return {linkage, inCOMDAT};
```
ljmf00-wekaio, line 268: I don't think this works, I'll double confirm on weka source code, but we have been getting too much linker issues by applying COMDAT to most of the symbols. e.g.:
```
ld.lld: error: relocation refers to a discarded section: .text._D5mecca3lib10reflection__T2asVAyaa13_6e6f7468726f7720406e6f6763TDFNaNbNiNfZvZQBzFNcMQuZ9__lambda2MFNaNbNiNeZQBs
>>> defined in core/3rdparty/build/mecca/libmecca-notrace.a(.._.._mecca_src_mecca_lib_exception.d.o)
>>> section group signature: _D5mecca3lib10reflection__T2asVAyaa13_6e6f7468726f7720406e6f6763TDFNaNbNiNfZvZQBzFNcMQuZ9__lambda2MFNaNbNiNeZQBs
>>> prevailing definition is in core/3rdparty/build/mecca/libmecca-notrace.a(.._.._mecca_src_mecca_reactor_io_inotify.d.o)
>>> referenced by reflection.d:556 (../core/3rdparty/mecca/src/mecca/lib/reflection.d:556)
>>>               .._.._mecca_src_mecca_lib_exception.d.o:(_D5mecca3lib9exception6ExcBuf__T15constructHelperHTC4coreQBt11AssertErrorTAyaZQBuMFNbNiNeQpmbKQuZQBu) in archive core/3rdparty/build/mecca/libmecca-notrace.a
```

This verifies on symbols with `InternalLinkage` and only after upgrading the compiler (symbols with internal linkage are the ones that are relocations to discarded symbols). I haven't tested with the same LLVM version tho, could be an upstream regression.

---

Bottom line: https://github.com/weka/ldc/commit/e1cca7ad853ce70afe815c00d97fd4ab897d61a7 with `--enable-wekamods=true` yields this linker error, and, without the mods, data culling doesn't work (we start to get duplicate sections, as described in the issue). The `--enable-wekamods=true` works on older versions. So we are on this middle ground where none works unless we only mark specific linkage with comdat.

I'll try to create a test to demonstrate this.

ljmf00-wekaio: Tested, and it works! I see, this works if we guarantee that `templateLinkage` is not ever `InternalLinkage`, which probably is just a derivative of `_odr` attributes? What about `--fvisibility=hidden`, this doesn't influence `templateLinkage`, right? If that's all true, I'm good with this change.

ljmf00-wekaio (approved):

gen/tollvm.cpp
kinke: `templateLinkage` is always an ODR linkage: https://github.com/ldc-developers/ldc/blob/580ff68b6f9f1e952fece1ea015d0cdb2e49bd80/driver/main.cpp#L470-L472

PR #4750, opened 2024-09-08
Emscripten: Predefine versions as for a regular musl Linux platform

kinke:
Based on https://github.com/ldc-developers/ldc/pull/4738.


kinke:
This should pave the way for porting druntime and Phobos to Emscripten, which is probably much simpler with this musl libc interface than the WASI interface directly.

PR #4756, opened 2024-09-20
Add more FP-rounding LLVM intrinsics

### LINKED ISSUE #4754: Missing intrinsic lrint
Someone on Discord needs ``lrint`` to generate ``CVTSS2SI``.

```d
pragma(LDC_intrinsic, "llvm.lrint.i64.f#")
    long llvm_lrint(T)(T val)
        if (__traits(isFloating, T));
```

I was able to find that this is a missing intrinsic. It is not hooked up to core.stdc.math or std.math.

kinke:
Resolves #4754.


kinke:
Oof, once again buggy intrinsic semantics by the looks of it (definitely not 'does what the libm function would do, except for NOT setting errno'), at least when LLVM has to resort to x87 (32-bit x86 not defaulting to SSE etc.). The Win64 `std.mathspecial` failure is probably related to its 64-bit `real`.

kinke:
Turns out the LLVM [docs](https://llvm.org/docs/LangRef.html#llvm-llround-intrinsic) are wrong; there is a difference between `llvm.lround.i64` and  `llvm.llround.i64` after all. In the `-mtriple=x86_64-windows-msvc` case, calling the `lround` vs. `llround` MSVCRT function. This fixed the 32-bit x86 issues too.

JohanEngelen (approved):

JohanEngelen:
> Turns out the LLVM [docs](https://llvm.org/docs/LangRef.html#llvm-llround-intrinsic) are wrong; there is a difference between `llvm.lround.i64` and `llvm.llround.i64` after all. 

Out of interest, what exactly is wrong in the LLVM docs?

kinke:
The syntax section uses `llvm.lround.i64`  exclusively.

PR #4769, opened 2024-10-28
Make ImportC also pick up C files with `-i`

bangbangsheshotmedown:
This was fixed in DMD few weeks ago, shouldn't LDC synchronize the changes automatically?


kinke:
Thx, but this will land when we sync with v2.111 (current DMD master). LDC tracks DMD `stable`.

bangbangsheshotmedown:
Oh i see, thanks

Issue #3606, opened 2020-11-03
__vector << int produce ICE, should be illegal?


p0nce:
The following program crashes LDC 1.24.0 and below too:

```d
import core.simd;
void fun()
{
    int4 A;
    A = A << 8;
}
```

Output with LDC 1.24:
```
#1 with ldc beta
sext source and destination must both be a vector or neither
  %5 = sext i32 8 to <4 x i32>
in function _D7example3funFAAyaZv
LLVM ERROR: Broken function found, compilation aborted!
/opt/compiler-explorer/ldcbeta/bin/ldc2[0x33f03d4]
Compiler returned: 255
```

DMD output is correct:
```
<source>(6): Error: incompatible types for `(A) << (8)`: `__vector(int[4])` and `int`
Compiler returned: 1
```

Interestingly it used to work in LDC 1.4.0 but doesn't since LDC 1.5.0 in Godbolt: https://d.godbolt.org/z/3Eb47n

kinke:
I disagree, this should be legal, because `A << int4(8)` works with LDC. The frontend implicitly casts scalars to vectors for arithmetic ops (`* 2` => `* cast(int4) 2`), but apparently not for shifts.

p0nce:
I have no opinion about this ¯\_(ツ)_/¯

PR #4774, opened 2024-11-07
runtime/jit-rt: revive and migrates the whole thing to OrcJIT v2

liushuyu:
This pull request migrates the dynamic compile system to LLVM OrcJIT v2 (the newer "v2" version from post-LLVM 16 with opaque pointers, not the older "v2" version from the LLVM 12 ~ 15 era). 

The implementation is currently in very rough shape, and I am not entirely sure what I am doing. However, the "Hello World" example in the documentation has started to work now.


runtime/jit-rt/cpp-so/optimizer.cpp
```diff
@@ -11,239 +11,375 @@
...
-#include "llvm/Target/TargetMachine.h"
+#include "valueparser.h"
+#include "utils.h"
 
-#include "llvm/IR/Constants.h"
-#include "llvm/IR/DataLayout.h"
-#include "llvm/IR/LegacyPassManager.h"
-#include "llvm/IR/Module.h"
-#include "llvm/IR/Verifier.h"
+//===-- optimizer.cpp -----------------------------------------------------===//
```
liushuyu, line 17: This is copied almost verbatim from `gen/optimizer.cpp`.

kinke:
Hehe, looks like one's motivated, nice! - In case you haven't seem them, there are quite a few lit-tests for this feature: https://github.com/ldc-developers/ldc/tree/master/tests/dynamiccompile. No idea if anything can still be salvaged from https://github.com/ldc-developers/ldc/pull/3184, Ivan's WIP from 5 (!) years ago by now.

liushuyu:
> Hehe, looks like one's motivated, nice! - In case you haven't seem them, there are quite a few lit-tests for this feature: https://github.com/ldc-developers/ldc/tree/master/tests/dynamiccompile. No idea if anything can still be salvaged from #3184, Ivan's WIP from 5 (!) years ago by now.

I see. I did not know there was another pull request doing the same thing from 5 years ago. Now that I see they made very similar changes before, I realize I duplicated a fair amount of changes (but this seems like a form of cross-validation? Two independent implementations look very similar).

liushuyu:
> there are quite a few lit-tests for this feature: https://github.com/ldc-developers/ldc/tree/master/tests/dynamiccompile.

Now, 100% of those LIT tests are passing (with LLVM 18).

kinke:
Very nice progress! - I've pushed 2 commits to reduce the remaining failures on aarch64.

tests/dynamiccompile/bind_bool.d
```diff
@@ -28,7 +28,7 @@ void main(string[] args)
     settings.optLevel = 3;
     settings.dumpHandler = (DumpStage stage, in char[] str)
     {
-      if (DumpStage.FinalAsm == stage)
+      if (DumpStage.FinalAsm == stage || DumpStage.OptimizedModule == stage)
```
kinke, line 31: Note that this doesn't just output the optimized IR now in case the test fails, but also makes the `indexOf` substring searches now also search in the IR.

liushuyu: > Note that this doesn't just output the optimized IR now too in case the test fails, but also makes the `indexOf` substring searches now also search in the IR.

Yes, LLVM on arm64 uses very clever optimizations, and the constants do not appear in the assembly (because LLVM loads the constant into the register using bit shifts).

kinke:
In my Windows VM, the tests don't hang as they do for CI (with a pushed Win32 fix wrt. `__chkstk`, now on 32-bit Windows too). Instead, some dynamic-compile tests pass (on both 32 and 64-bit Windows), others fail with errors like:

* `dynamiccompile/simd_simple_opt.d`:
  * 64-bit:
    ```
    Dynamic compiler fatal: Symbol not found in jitted code: ".jit_bind" (".jit_bind")
    ```
  * 32-bit:
    ```
    Incorrect number of arguments passed to called function!
      call x86_stdcallcc void @"\01__D3ldc15dynamic_compile__T4bindTPFG16hZG32hTxG16hZQyFQvxQlZ7wrapperFSQCpQCo__TQCaTQByTxQBqZQCnFQClxQCcZ7ContextQCzZQCy"(ptr inreg noalias align 1 %3, ptr byval(%"ldc.dynamic_compile.bind!(ubyte[32] function(ubyte[16]), const(ubyte[16])).bind.Context") align 4 %4) #0
    LLVM ERROR: Broken module found, compilation aborted!
    ```
* `dynamiccompile/simple.d`:
  * 64-bit:
    ```
    Dynamic compiler fatal: Symbol not found in jitted code: "_D6simple3bazFZv" ("_D6simple3bazFZv")
    ```
  * 32-bit:
    ```
    Dynamic compiler fatal: Symbol not found in jitted code: "__D6simple3barFZi" ("__D6simple3barFZi")
    ```

Looks like there's some trouble with the `\1` mangle prefix on Windows, which is used to make LLVM not apply any extra mangling (e.g., extra leading underscore on Win32, and vectorcall/stdcall suffix on Win64).

liushuyu:
> `dynamiccompile/simple.d`:
> 
>     * 64-bit:
>       ```
>       Dynamic compiler fatal: Symbol not found in jitted code: "�_D6simple3bazFZv" ("_D6simple3bazFZv")
>       ```

I have tried to debug this problem on a Windows 11 installation with Visual Studio 2022 (Community Edition).
It seems like either our dynamic compile mechanism setup LLVM in the wrong way, or there is an LLVM bug on Windows: the symbols were not materialized when they should be (LLVM on Unix platforms will materialize the JIT symbols during the lookup, i.e. lazily compile/resolve the symbols).

tests/dynamiccompile/simd_simple_opt.d
```diff
@@ -0,0 +1,67 @@
...
+{
+    const ubyte ascii_a = 'a' - 9 - 1;
+    ubyte16 all_four;
+    all_four[] = 4;
+    ubyte16 invec = loadUnaligned!ubyte16(binary.ptr);
+    ubyte[32] hex;
+
+    ubyte16 masked1 = invec & 0xF;
+    // TODO: (invec >> 4) will cause an LDC internal error
+    ubyte16 masked2 = (invec >> all_four) & 0xF;
```
liushuyu, line 22: We might want to get this problem fixed, as C/C++ compilers could understand `[vector] >> scalar` expression very well.

liushuyu:
On Ubuntu 20.04 aarch64:

```
********************
Unexpectedly Passed Tests (1):
  LDC :: dynamiccompile/throw.d
```

But it's still broken on macOS aarch64 unfortunately

tests/dynamiccompile/simd_simple_opt.d
kinke: This is https://github.com/ldc-developers/ldc/issues/3606#issue-735305825 and should IMO be fixed in the frontend, it somehow doesn't promote scalars to vectors implicitly for shifts.

liushuyu: I have proposed #4780 to fix this issue.

liushuyu:
It seems like I may have discovered several bugs in LLVM that are exclusive to Windows. I have currently worked around them because I am not entirely sure if they are "bugs" or "intended behaviors".

liushuyu:
I have refactored the JIT runtime to use LLVM JIT "easy API" (LLJIT) to try to solve issues on Windows.
And it is mostly working now; the remaining failures are primarily due to API rewrites.

liushuyu:
Done. All the issues have been fixed. Ready for review.

CMakeLists.txt
```diff
@@ -452,20 +452,33 @@ include(HandleLTOPGOBuildOptions)
 #
 # Enable Dynamic compilation if supported for this platform and LLVM version.
 #
-set(LDC_DYNAMIC_COMPILE "AUTO" CACHE STRING "Support dynamic compilation (ON|OFF). Enabled by default; not supported for LLVM >= 12.")
+set(LDC_DYNAMIC_COMPILE "AUTO" CACHE STRING "Support dynamic compilation (ON|OFF). Enabled by default; not supported for LLVM < 18.")
 option(LDC_DYNAMIC_COMPILE_USE_CUSTOM_PASSES "Use custom LDC passes in jit" ON)
 if(LDC_DYNAMIC_COMPILE STREQUAL "AUTO")
-    if(LDC_LLVM_VER LESS 1200)
+    if((LDC_LLVM_VER LESS 2000) AND (LDC_LLVM_VER GREATER 1800))
```
kinke, line 458: Why exclude LLVM 20+? And IMO it's not perfectly clear that LLVM 18(.1) is supported - I'd prefer `if(NOT LDC_LLVM_VER LESS 1800))`, which we use in other places IIRC (for `>=`).

CMakeLists.txt
```diff
@@ -452,20 +452,33 @@ include(HandleLTOPGOBuildOptions)
...
         set(LDC_DYNAMIC_COMPILE ON)
     else()
-        # TODO: port from ORCv1 API (dropped with LLVM 12) to ORCv2 (added with LLVM 7)
         set(LDC_DYNAMIC_COMPILE OFF)
     endif()
 endif()
+# https://llvm.org/docs/JITLink.html for the list of supported platforms
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|x64|amd64|aarch64|arm64|riscv64|loongarch64")
```
kinke, line 465: Not sure how that works when cross-compiling.

gen/dynamiccompile.cpp
```diff
@@ -364,10 +366,20 @@ llvm::Constant *getArrayPtr(llvm::Type *type, llvm::Constant *array) {
...
+static llvm::PointerType *getI8PtrType(llvm::LLVMContext &C) {
+  return llvm::PointerType::get(llvm::IntegerType::get(C, 8), 0);
+}
+
+llvm::Constant *getI8Ptr(llvm::GlobalVariable *val) {
+  assert(nullptr != val);
+  return llvm::ConstantExpr::getBitCast(
+      val, getI8PtrType(val->getContext()));
+}
+
```
kinke, line 378: Is this stuff still needed with opaque IR pointers now?

tests/dynamiccompile/simd_simple_opt.d
kinke: This works now.

CMakeLists.txt
liushuyu: According to https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html#variable:CMAKE_SYSTEM_PROCESSOR, the user should set this variable to the target processor name when cross-compiling.

CMakeLists.txt
liushuyu: Done

gen/dynamiccompile.cpp
liushuyu: Done

liushuyu:
The CI failures are due to network issues; a retry should fix them.

thewilsonator:
>  retry should fix them.

(restarted)

liushuyu:
> > retry should fix them.
> 
> (restarted)

Thanks! All CI tests passed now

thewilsonator:
Did you resolve https://github.com/ldc-developers/ldc/pull/4774#discussion_r1833458713 ?

kinke:
I've pushed a commit merging the two very similar `optimizer.cpp` versions; a mechanical refactoring without any intended functional changes. There's probably more room for follow-up refactorings, such as moving `setRtCompileVar()` to another jitrt-specific module.

liushuyu:
> I've pushed a commit merging the two very similar `optimizer.cpp` versions; a mechanical refactoring without any intended functional changes. There's probably more room for follow-up refactorings, such as moving `setRtCompileVar()` to another jitrt-specific module.

Thanks!

Hopefully, by switching to the more stable easy LLJIT API, we can last longer before another need for refactoring comes.

kinke (approved):

PR #4780, opened 2024-11-12
gen: fix implicit vector cast in binary operations

### LINKED ISSUE #3606: __vector << int produce ICE, should be illegal?

liushuyu:
This pull request fixes implicit vector cast in binary operations. Rather than standard integer extension, vector casts require the RHS (right-hand side) to be a vector of matching size. In this patch, a vector splat operation fills the target vector type with the original value across all the lanes/elements.

This should fix #3606.

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -488,6 +488,26 @@ DValue *DtoNullValue(Type *type, Loc loc) {
  * CASTING HELPERS
  ******************************************************************************/
 
+// Cast a primitive integer type to another primitive integer type
+static inline llvm::Value *DtoCastSimpleInt(llvm::Type *tolltype, Type *dfrom,
+                                            size_t tosz, llvm::Value *rval) {
+  size_t fromsz = size(dfrom);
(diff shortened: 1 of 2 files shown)


kinke:
Alternative frontend fix: https://github.com/ldc-developers/ldc/pull/4781

liushuyu:
> Alternative frontend fix: #4781

Ah I see, then I will close this one.

PR #4783, opened 2024-11-17
Implement -fsanitize-recover for ASan and MSan.

### LINKED ISSUE #4770: Add support for `-fsanitize-recover=address` instrumentation
This is required to avoid the address sanitizer to fail fast on false positives or to report multiple unrelated errors

JohanEngelen:
Resolves #4770

Note that env variable setting `ASAN_OPTIONS=halt_on_error=false` is needed to actually recover from failures (executables built with clang behaves the same).


JohanEngelen:
The windows failure seems unrelated?

kinke:
Yeah, caused by new WinSDK headers in the latest GHA Windows image, and fixed by https://github.com/ldc-developers/ldc/pull/4784.

PR #4772, opened 2024-11-05
Add LLVM 19.1.x support for LDC

liushuyu:
This pull request adds LLVM 19.1.x support for LDC. It is based on previous work by @kinke and @thewilsonator in #4763 and #4735.

Some CI setup is required because it needs 19.1.3 binaries in the forked LLVM repository.

PR #4787, opened 2024-11-25
CircleCI: Bump jobs to Ubuntu 24

kinke:


kinke:
FYI @liushuyu. Looking pretty good, just some sanitizer problems with the distro sanitizer libs, the buggy `codegen/xtensa.d` that you've fixed in the LLVM 19 PR (IIRC), and otherwise just a single gdb-related failure with gdb v15.0, for a missed global again (imported from another module, referenced by identifier only): https://github.com/ldc-developers/ldc/blob/40716bee1df2e202d70304389b4d6e22516ce544/tests/debuginfo/print_gdb.d#L100

liushuyu:
@kinke 

> [Work around possible remaining gdb v15 regression in tests/debuginfo/print_gdb.d](https://github.com/ldc-developers/ldc/pull/4787/commits/1dd76cca6790e7b4557c1a523e7b814e7cb67f23)
[1dd76cc](https://github.com/ldc-developers/ldc/pull/4787/commits/1dd76cca6790e7b4557c1a523e7b814e7cb67f23)

I think I will still open a bug report for GDB, as LLDB (which does not have D support) can resolve this symbol without the namespace:

```
* thread #1, name = 'print_gdb', stop reason = breakpoint 1.1
    frame #0: 0x00005555555552f2 print_gdb`D main at print_gdb.d:103:5
   100  // GDB: p b_Glob
   101  // CHECK: = 99.8
   102
-> 103      cC clsC = new cC;
   104      clsC.classMemberFunc(2);
   105      cC.classStaticFunc(4);
   106      clsC.mixedVal++;
(lldb) p b_Glob
error: print_gdb [0x000000000000002f]: unhandled type tag DW_TAG_module (DW_TAG_module), please file a bug and attach the file at the start of this error message
(float) 99.8799972
(lldb)
```

liushuyu:
Congratulations on making the scripts run on someone else's computer (aka CI)!

I think you can just cherry-pick https://github.com/ldc-developers/ldc/pull/4772/commits/83876511c71f2ae9c58c2e70b579370b791779e1 to make the remainder of the CI tests pass.

kinke:
For posterity: container and VM performances seem more or less identical, and the credit consumption is identical too. The only real downside of the VM is the startup time, something like 30-60 seconds.

liushuyu:
> Thanks, that'd be awesome. As it worked with gdb 9 too, I'm pretty sure it should work without fully qualified name.

@kinke I have opened a bug report on the upstream along with a suggested fix: https://sourceware.org/bugzilla/show_bug.cgi?id=32398

kinke:
Oh wow, so you did basically everything already! :) - I fully trust your detailed report to be enough for the gdb team to come up with a regression test case; I don't build gcc myself or use experimental toolchains, so please leave me out, I won't be able to early-test. But please link your report in #4389, which we'll be able to close once the gdb fix has landed and we can confirm it via CI. It's not an LDC issue per se as you know, just a place where this testsuite-problem is tracked, and remains open, whereas this PR is merged, closed and forgotten. :)

liushuyu:
> Oh wow, so you did basically everything already! :) Thanks! - I fully trust your detailed report to be enough for the gdb team to come up with a regression test case; I don't build gcc myself or use experimental toolchains, so please leave me out, I won't be able to early-test. But please link your report in #4389, which we'll be able to close once the gdb fix has landed and we can confirm it via CI. It's not an LDC issue per se as you know, just a place where this testsuite-problem is tracked, and remains open, whereas this PR is merged, closed and forgotten. :)

Thanks! Will do.

PR #4799, opened 2024-12-08
TargetABI: Add base `rewriteFunctionType()` implementation

kinke:


gen/abi/nvptx.cpp
```diff
@@ -30,12 +30,6 @@ struct NVPTXTargetABI : TargetABI {
     t = t->toBasetype();
     return ((t->ty == TY::Tsarray || t->ty == TY::Tstruct) && size(t) > 64);
   }
-  void rewriteFunctionType(IrFuncTy &fty) override {
-    for (auto arg : fty.args) {
-      if (!arg->byref)
-        rewriteArgument(fty, *arg);
-    }
-  }
```
kinke, line 38: Small functional change: the return value is now potentially rewritten to, as was already done for the spirv ABI.

PR #4804, opened 2024-12-16
add debug info for function parameter types

rainers:
A couple of days ago I added support to mago to evaluate the `__debugOverview`, `__debugExpanded` and `__debugStringview` hooks as free functions instead of struct members (see https://rainers.github.io/visuald/visuald/Debugging.html#customization). Apart from allowing to add debug visualizers for library types without editing the source, I hoped to also have LDC support this feature even though it omits class and struct members in the debug info. Here's a tiny example:

```D
struct String
{
    string _s;
}
string __debugOverview(const String* s)
{
    return s._s;
}
```

Unfortunately, LDC omits the function parameters from the function type, but they are needed for overload resolution. This a small patch to add these. It also ticks off one item of https://github.com/ldc-developers/ldc/issues/1716 as the call stack now displays parameter types and values.

While mago supports passing the argument as a pointer or a reference, the latter doesn't work with LDC as the ref argument is typed as value. Is this removed somewhere in LDC if the ABI passes the value by reference anyway?

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -25,12 +25,10 @@
 #include "driver/cl_options.h"
 #include "driver/ldc-version.h"
 #include "gen/functions.h"
-#include "gen/abi/abi.h"
 #include "gen/irstate.h"
 #include "gen/llvmhelpers.h"
 #include "gen/logger.h"
 #include "gen/optimizer.h"
-#include "gen/pragma.h"
 #include "gen/tollvm.h"
 #include "ir/irfunction.h"
 #include "ir/irfuncty.h"
@@ -722,59 +720,51 @@ DIType DIBuilder::CreateAArrayType(TypeAArray *type) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-DISubroutineType DIBuilder::CreateFunctionType(Type *type, FuncDeclaration* fd) {
+DISubroutineType DIBuilder::CreateFunctionType(Type *type,
+                                               FuncDeclaration *fd) {
   TypeFunction *t = type->isTypeFunction();
   assert(t);
 
-  Type *retType = t->next;
   llvm::SmallVector<LLMetadata *, 8> params;
-
-  auto irFunc = fd ? getIrFunc(fd, false) : nullptr;
-  if (irFunc) {
-    auto push_arg = [&](IrFuncTyArg *arg) {
-      if (!arg)
-        return;
-      auto ditype = CreateTypeDescription(arg->type);
-      if (arg->byref) {
-        ditype = DBuilder.createPointerType(ditype, target.ptrsize * 8, 0, llvm::None,
-            processDIName((type->toPrettyChars(true) + llvm::StringRef("*")).str()));
+  auto pushParam = [&](Type *type, bool isRef) {
+    auto ditype = CreateTypeDescription(type);
+    if (isRef) {
+      if (!ditype) { // void or noreturn
+        ditype = CreateTypeDescription(Type::tuns8);
       }
-      params.push_back(ditype);
-    };
-    // mimmicking what happens in DtoFunctionType
-    IrFuncTy &irFty = irFunc->irFty;
-    push_arg(irFty.ret);
-    push_arg(irFty.arg_sret);
-    push_arg(irFty.arg_this);
-    push_arg(irFty.arg_nest);
-    push_arg(irFty.arg_objcSelector);
-    push_arg(irFty.arg_arguments);
+      ditype = DBuilder.createReferenceType(llvm::dwarf::DW_TAG_reference_type,
+                                            ditype, target.ptrsize * 8);
+    }
+    params.emplace_back(ditype);
+  };
 
-    TargetABI *abi = DtoIsIntrinsic(fd) ? TargetABI::getIntrinsic() : gABI;
-    if (irFty.arg_sret && irFty.arg_this && abi->passThisBeforeSret(t)) {
-      std::swap(params[0], params[1]);
+  // the first 'param' is the return value
+  pushParam(t->next, t->isref());
+
+  // then the implicit 'this'/context pointer
+  if (fd) {
+    DIType pointeeType = nullptr;
+    if (auto parentAggregate = fd->isThis()) {
+      pointeeType = CreateCompositeType(parentAggregate->type);
+    } else if (fd->isNested()) {
+      pointeeType = CreateTypeDescription(Type::tuns8); // cannot use void
     }
 
-    const size_t numExplicitLLArgs = irFty.args.size();
-    for (size_t i = 0; i < numExplicitLLArgs; i++) {
-      push_arg(irFty.args[i]);
-    }
-  } else {
-    // incomplete information without actual code gen
-    params.emplace_back(CreateTypeDescription(t->next)); // return type
-    auto len = t->parameterList.length();
-    for (size_t i = 0; i < len; i++) {
-      auto ptype = t->parameterList[i]->type;
-      auto ditype = CreateTypeDescription(ptype);
-      if (t->parameterList[i]->isReference()) {
-        auto name = processDIName(
-            (ptype->toPrettyChars(true) + llvm::StringRef("*")).str());
-        ditype = DBuilder.createPointerType(
-            ditype, target.ptrsize * 8, 0, llvm::None, name);
-      }
+    if (pointeeType) {
+      DIType ditype = DBuilder.createReferenceType(
+          llvm::dwarf::DW_TAG_pointer_type, pointeeType, target.ptrsize * 8);
+      ditype = DBuilder.createObjectPointerType(ditype);
       params.emplace_back(ditype);
     }
   }
+
+  // and finally the formal parameters
+  const auto len = t->parameterList.length();
+  for (size_t i = 0; i < len; i++) {
+    const auto param = t->parameterList[i];
+    pushParam(param->type, param->isReference());
+  }
+
   auto paramsArray = DBuilder.getOrCreateTypeArray(params);
   return DBuilder.createSubroutineType(paramsArray, DIFlags::FlagZero, 0);
 }
@@ -852,7 +842,7 @@ DIType DIBuilder::CreateTypeDescription(Type *t, bool voidToUbyte) {
                                       llvm::None, processDIName(name));
   }
   if (auto tf = t->isTypeFunction())
-    return CreateFunctionType(tf, nullptr);
+    return CreateFunctionType(tf);
   if (auto td = t->isTypeDelegate())
     return CreateDelegateType(td);
 
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -25,10 +25,12 @@
 #include "driver/cl_options.h"
 #include "driver/ldc-version.h"
 #include "gen/functions.h"
+#include "gen/abi/abi.h"
 #include "gen/irstate.h"
 #include "gen/llvmhelpers.h"
 #include "gen/logger.h"
 #include "gen/optimizer.h"
+#include "gen/pragma.h"
 #include "gen/tollvm.h"
 #include "ir/irfunction.h"
 #include "ir/irfuncty.h"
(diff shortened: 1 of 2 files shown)


gen/dibuilder.cpp
```diff
@@ -809,6 +810,8 @@ DIType DIBuilder::CreateTypeDescription(Type *t, bool voidToUbyte) {
     return CreateFunctionType(tf);
   if (auto td = t->isTypeDelegate())
     return CreateDelegateType(td);
+  if (auto tt = t->isTypeTuple())
+    return DBuilder.createUnspecifiedType(processDIName("tuple")); // TODO
```
rainers, line 814: I had to avoid crashing the compiler on tuples when importing std.typecons here. For some reason, tuples work elsewhere in LDC. but not dmd. Support for this should be added anyway, but I would postpone this to another PR.

gen/dibuilder.cpp
```diff
@@ -725,11 +725,12 @@ DISubroutineType DIBuilder::CreateFunctionType(Type *type) {
...
 
   Type *retType = t->next;
-
-  // Create "dummy" subroutine type for the return type
-  LLMetadata *params = {CreateTypeDescription(retType)};
+  llvm::SmallVector<LLMetadata *, 8> params;
+  params.emplace_back(CreateTypeDescription(retType));
+  if (auto parameters = t->parameterList.parameters)
+    for (auto &p : *parameters)
+      params.emplace_back(CreateTypeDescription(p->type));
```
kinke, line 732: This is too simplistic - we need to match the IR parameters, which includes sret pointer, this pointer etc., and might skip zero-sized params for some ABIs etc.

And the problem with the `ref` that you mentioned is probably caused by this as well - it doesn't respect the `Parameter::storageClass`, only uses its `type`.

gen/dibuilder.cpp
```diff
@@ -725,11 +725,12 @@ DISubroutineType DIBuilder::CreateFunctionType(Type *type) {
   assert(t);
 
   Type *retType = t->next;
-
-  // Create "dummy" subroutine type for the return type
-  LLMetadata *params = {CreateTypeDescription(retType)};
+  llvm::SmallVector<LLMetadata *, 8> params;
+  params.emplace_back(CreateTypeDescription(retType));
+  if (auto parameters = t->parameterList.parameters)
```
kinke, line 730: [These are the raw formal params, where tuples aren't expanded. `t->parameterList` has a `.length()` and indexing operator for C++, which expands tuples automatically.]

gen/dibuilder.cpp
kinke: FWIW, `IrFuncTy` has all the info about the IR params: https://github.com/ldc-developers/ldc/blob/master/ir/irfuncty.h

rainers: I was suspecting it was to easy ;) Thanks for the link to IrFuncTy, I have updated it now with the info found there. This now also works with ref parameters in mago.

gen/dibuilder.cpp
rainers: No longer necessary when using the expanded parameter list.

gen/dibuilder.cpp
```diff
@@ -720,16 +722,60 @@ DIType DIBuilder::CreateAArrayType(TypeAArray *type) {
...
+
+  auto irFunc = fd ? getIrFunc(fd, false) : nullptr;
+  if (irFunc) {
+    auto push_arg = [&](IrFuncTyArg *arg) {
+      if (!arg)
+        return;
+      auto ditype = CreateTypeDescription(arg->type);
+      if (arg->byref) {
+        ditype = DBuilder.createPointerType(ditype, target.ptrsize * 8, 0, llvm::None,
+            processDIName((type->toPrettyChars(true) + llvm::StringRef("*")).str()));
```
kinke, line 740: `arg->byref` might be set for non-D-refs. I think the only robust way would be to use `IrFuncTyArg::parametersIdx` and look up the AST param.

Also note that there's a `DBuilder.createReferenceType()`; used in e.g. `DIBuilder::EmitLocalVariable()`. But a struct `this` needs to be represented as a pointer, not a ref, see comment in that function...

gen/dibuilder.cpp
```diff
@@ -720,16 +722,60 @@ DIType DIBuilder::CreateAArrayType(TypeAArray *type) {
...
+    push_arg(irFty.ret);
+    push_arg(irFty.arg_sret);
+    push_arg(irFty.arg_this);
+    push_arg(irFty.arg_nest);
+    push_arg(irFty.arg_objcSelector);
+    push_arg(irFty.arg_arguments);
+
+    TargetABI *abi = DtoIsIntrinsic(fd) ? TargetABI::getIntrinsic() : gABI;
+    if (irFty.arg_sret && irFty.arg_this && abi->passThisBeforeSret(t)) {
+      std::swap(params[0], params[1]);
```
kinke, line 755: `params[1]` and `params[2]` - 0 is the return value.

gen/dibuilder.cpp
```diff
@@ -720,16 +722,60 @@ DIType DIBuilder::CreateAArrayType(TypeAArray *type) {
...
+    // mimmicking what happens in DtoFunctionType
+    IrFuncTy &irFty = irFunc->irFty;
+    push_arg(irFty.ret);
+    push_arg(irFty.arg_sret);
+    push_arg(irFty.arg_this);
+    push_arg(irFty.arg_nest);
+    push_arg(irFty.arg_objcSelector);
+    push_arg(irFty.arg_arguments);
+
+    TargetABI *abi = DtoIsIntrinsic(fd) ? TargetABI::getIntrinsic() : gABI;
```
kinke, line 753: `DtoIsIntrinsic()` should segfault with `fd == nullptr`.

kinke: Oh, we're in a `if (irFunc)` block, all good.

gen/dibuilder.cpp
```diff
@@ -720,16 +722,60 @@ DIType DIBuilder::CreateAArrayType(TypeAArray *type) {
...
+      push_arg(irFty.args[i]);
+    }
+  } else {
+    // incomplete information without actual code gen
+    params.emplace_back(CreateTypeDescription(t->next)); // return type
+    auto len = t->parameterList.length();
+    for (size_t i = 0; i < len; i++) {
+      auto ptype = t->parameterList[i]->type;
+      auto ditype = CreateTypeDescription(ptype);
+      if (t->parameterList[i]->isReference()) {
```
kinke, line 769: Note that iterating over a `parameterList` is expensive IIRC, as tuples are expanded on the fly every time (from the beginning/first param). So keeping a ref to `t->parameterList[i]` probably worth it.

In general, not sure if it's worth using the formal params as fallback in this case where we don't have an IrFunc.

gen/dibuilder.cpp
kinke: > `arg->byref` might be set for non-D-refs. I think the only robust way would be to use `IrFuncTyArg::parametersIdx` and look up the AST param.

It's totally unclear to me what these new param DIs are used for (as *inside* a function, we do have DIs for the params, but for their variable storage in memory, just like for locals, so not tied to the raw IR param). Maybe `IrFuncTyArg::byref` is better than the frontend ref-ness indeed - with a likely exception for `byval` params though (`IrFuncTyArg::isByVal()`) I guess.

kinke: Interesting: https://cpp.godbolt.org/z/x83nbd4Eh

```LL
; Large foo(Large l, Large &, NonPOD) { return l; }
define linkonce_odr dso_local void @S::foo(Large, Large&, NonPOD)(ptr dead_on_unwind noalias writable sret(%struct.Large) align 4 %agg.result, ptr noundef nonnull align 1 dereferenceable(1) %this, ptr noundef byval(%struct.Large) align 8 %l, ptr noundef nonnull align 4 dereferenceable(32) %0, ptr noundef %1) comdat align 2 !dbg !46 {

…

!46 = distinct !DISubprogram(name: "foo", linkageName: "S::foo(Large, Large&, NonPOD)", scope: !25, file: !4, line: 11, type: !28, scopeLine: 11, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !0, declaration: !27, retainedNodes: !23)

!28 = !DISubroutineType(types: !29)
!29 = !{!30, !36, !30, !37, !3}

!3 = distinct !DICompositeType(tag: DW_TAG_structure_type, name: "NonPOD", file: !4, line: 5, size: 32, flags: DIFlagTypePassByReference | DIFlagNonTrivial, elements: !5, identifier: "_ZTS6NonPOD")
!30 = distinct !DICompositeType(tag: DW_TAG_structure_type, name: "Large", file: !4, line: 1, size: 256, flags: DIFlagTypePassByValue, elements: !31, identifier: "_ZTS5Large")
!36 = !DIDerivedType(tag: DW_TAG_pointer_type, baseType: !25, size: 64, flags: DIFlagArtificial | DIFlagObjectPointer)
!37 = !DIDerivedType(tag: DW_TAG_reference_type, baseType: !30, size: 64)
```

So clang uses the frontend signature and types, only adding the implicit `this` pointer (with extra flags) - no sret pointer. And the types and ref-ness is determined from the AST alone as well - the IR function has 5 *pointer* IR params, but the function-type-DI only has 4 - implicit `this` (pointer), `Large l` (value), `Large&` (ref) and `NonPOD` (value).

kinke: I've had a go at reworking this accordingly and pushed a commit.

rainers: > It's totally unclear to me what these new param DIs are used for

The VS debugger seems to need this information to populate the parameter types and values in the call stack.

Mago uses the function type to find the correct overload of the `__debug*` functions and verifies the signature so that it can call the fucntion in the debugged process. Indeed, this could also check debug info for local variables (not sure you can easily distinguish between function arguments and locals), but it seems more complicated.

kinke: > Thanks @kinke , that makes it a bit simpler again, i.e. no dependency on the IR.

Yeah as I brought you on the wrong (IR) tack, I figured I better 'fix' it again. :)

> It doesn't show some hidden parameters (e.g. _arguments), but I'm not sure if that's good or bad.

Same here.

> The new version works with mago, too. I had some troubles also getting the struct members to work, because
> 
>     * sometimes the struct type is returned when querying IDiaSymbol instead of a pointer to the struct.
> 
>     * there seem to be conflicting debug information regarding the calling convention depending on how the symbol is accessed.
> 
> 
> I don't see how this is caused by the emitted debug symbols, as it looks correct in a dump of the debug symbols. I have added some workarounds to mago.

clang uses an abundance of flags (`DIFlagTypePassByReference | DIFlagNonTrivial`, `DIFlagTypePassByValue`), that might be required/helpful in some circumstances. And I haven't quickly found a way to add the extra 'this' pointer flags, so I left that as a TODO comment.

gen/dibuilder.cpp
```diff
@@ -720,16 +720,53 @@ DIType DIBuilder::CreateAArrayType(TypeAArray *type) {
...
-  LLMetadata *params = {CreateTypeDescription(retType)};
-  auto paramsArray = DBuilder.getOrCreateTypeArray(params);
+  // the first 'param' is the return value
+  pushParam(t->next, t->isref());
+
+  // then the implicit 'this'/context pointer
+  if (fd) {
+    Type *pointeeType = nullptr;
+    if (auto parentAggregate = fd->isThis()) {
+      pointeeType = parentAggregate->type;
```
kinke, line 748: My bad - this works for structs, but not for classes.

kinke: Fix amended, incl. 'object-pointer' and 'artificial' flags.

kinke (approved):

gen/dibuilder.cpp
rainers: Thanks and sorry for the delay. The previous version already worked for me for classes, too.

kinke: Np at all, my bad anyway, plus I didn't get to it sooner either. - FWIW, the previous version emitted an additional indirection for classes, so maybe you can get rid of a workaround in mago now.

PR #4812, opened 2025-01-03
Add configuration placeholder %%ldcversion%%

d-musique:
Hi, this introduces a placeholder for `ldc2.conf` which is set to the version of LDC.

The rationale is: when following the "Cross-compiling with LDC" on a linux system, according to section "Default libraries", it is required to download prebuilt runtimes of the version matching the host compiler.

Since the host compiler is subject to system updates, I might find myself in a situation of version mismatch if I forget to update target libraries.

Favoring versioned directories may work as a preventive measure against such cases.

```
    lib-dirs = ["/opt/ldc2-%%ldcversion%%-windows-multilib/lib64"];
```

### DIFF
--- a/driver/configfile.d
+++ b/driver/configfile.d
@@ -110,11 +110,18 @@ struct CfgPaths
     }
 }
 
+extern(C++, "ldc")
+{
+    extern __gshared const(const(char) *) ldc_version;
+}
+
 string replacePlaceholders(string str, CfgPaths cfgPaths)
 {
+    const dVersion = ldc_version[0..strlen(ldc_version)];
     return str
         .replace("%%ldcbinarypath%%", cfgPaths.ldcBinaryDir)
-        .replace("%%ldcconfigpath%%", cfgPaths.cfgBaseDir);
+        .replace("%%ldcconfigpath%%", cfgPaths.cfgBaseDir)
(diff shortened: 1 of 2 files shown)


JohanEngelen:
clang does something similar with library folders. Indeed it would help a lot to easily support multiple compiler versions on the same system.

driver/configfile.d
```diff
@@ -110,11 +110,18 @@ struct CfgPaths
     }
 }
 
+extern(C++, "ldc")
+{
+    extern __gshared const(const(char) *) ldc_version;
```
JohanEngelen, line 115: This is not needed. Please use `global.ldc_version` which is already a dstring type  (`import dmd.globals`).

driver/configfile.d
```diff
@@ -110,11 +110,18 @@ struct CfgPaths
...
+}
+
 string replacePlaceholders(string str, CfgPaths cfgPaths)
 {
+    const dVersion = ldc_version[0..strlen(ldc_version)];
```
JohanEngelen, line 120: As mentioned above, please use `dmd.globals.ldc_version`.

JohanEngelen (commented):
Can you please submit a followup PR that fixes my comments? Thanks.

PR #4818, opened 2025-01-12
NFC: don't generate moduleinfo for sanitizer support in runtime

JohanEngelen:
Just a small cleanup


JohanEngelen:
unittesting fails when module info is removed.

PR #4821, opened 2025-01-12
Fix ASan for zero-sized variables passed as function argument.

### LINKED ISSUE #4816: ASAN fails on extern(C) structs (zero sized structs) when passing stack variable to class member
```d
class Foo
{
    this(Bar bar) { this.bar = bar; }
    Bar bar;
}

extern(C) struct Bar {}

void main()
{
	auto bar = Bar();
    auto foo = new Foo(bar);
}
```

With `ldc2 -fsanitize=address %`

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==27==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x55e3bea07532 bp 0x7ffdd7f55120 sp 0x7ffdd7f55100 T0)
==27==The signal is caused by a READ memory access.
==27==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
    #0 0x55e3bea07532 in D main /sandbox/onlineapp.d:12:5
    #1 0x55e3beab788c in _D2rt6dmain212_d_run_main2UAAamPUQgZiZ6runAllMFZv (/tmp/onlineapp-433482+0xce88c) (BuildId: 8e61071a375c929d9dfe0882c992f5960dc2fae5)
    #2 0x55e3beab7786 in _d_run_main2 (/tmp/onlineapp-433482+0xce786) (BuildId: 8e61071a375c929d9dfe0882c992f5960dc2fae5)
    #3 0x55e3beab75dc in _d_run_main (/tmp/onlineapp-433482+0xce5dc) (BuildId: 8e61071a375c929d9dfe0882c992f5960dc2fae5)
    #4 0x55e3bea07591 in main /dlang/ldc-1.38.0/bin/../import/core/internal/entrypoint.d:42:17
    #5 0x7f3f5704c1c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 08134323d00289185684a4cd177d202f39c2a5f3)
    #6 0x7f3f5704c28a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 08134323d00289185684a4cd177d202f39c2a5f3)
    #7 0x55e3bea073c4 in _start (/tmp/onlineapp-433482+0x1e3c4) (BuildId: 8e61071a375c929d9dfe0882c992f5960dc2fae5)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /sandbox/onlineapp.d:12:5 in D main
==27==ABORTING
Error: /tmp/onlineapp-433482 failed with status: 1
```

JohanEngelen:
Fixes #4816

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -895,10 +895,12 @@ void DtoVarDeclaration(VarDeclaration *vd) {
 
     Type *type = isSpecialRefVar(vd) ? pointerTo(vd->type) : vd->type;
 
+    // We also allocate a variable for zero-sized variables, because they are technically not `null` when loaded.
+    // The x86_64 ABI "loads" zero-sized function arguments, and without an allocation ASan will report an error (Github #4816).
     llvm::Value *allocainst;
     bool isRealAlloca = false;
     LLType *lltype = DtoType(type); // void for noreturn
-    if (lltype->isVoidTy() || gDataLayout->getTypeSizeInBits(lltype) == 0) {
+    if (lltype->isVoidTy()) {
       allocainst = getNullPtr();
     } else if (type != vd->type) {
       allocainst = DtoAlloca(type, vd->toChars());
--- a/tests/sanitizers/asan_gh4816.d
+++ b/tests/sanitizers/asan_gh4816.d
@@ -1,0 +1,20 @@
+// Test extern(C) structs (zero sized structs) and zero-sized arrays. Github #4816
+
+// REQUIRES: ASan
+
+// RUN: %ldc -g -fsanitize=address %s -of=%t%exe
+// RUN: %t%exe
+
+auto foo(void[0] bar) { }
+
+extern(C) struct S {}
+auto foo(S s) { }
+
+void main()
+{
+    void[0] bar;
+    foo(bar);
+
+    S s;
+    foo(s);
+}


JohanEngelen:
@ljmf00 On mac (arm64) there is no problem, so I was only able to verify the testcase on godbolt.

gen/llvmhelpers.cpp
```diff
@@ -895,10 +895,12 @@ void DtoVarDeclaration(VarDeclaration *vd) {
 
     Type *type = isSpecialRefVar(vd) ? pointerTo(vd->type) : vd->type;
 
+    // We also allocate a variable for zero-sized variables, because they are technically not `null` when loaded.
+    // The x86_64 ABI "loads" zero-sized function arguments, and without an allocation ASan will report an error (Github #4816).
     llvm::Value *allocainst;
     bool isRealAlloca = false;
     LLType *lltype = DtoType(type); // void for noreturn
-    if (lltype->isVoidTy() || gDataLayout->getTypeSizeInBits(lltype) == 0) {
```
ljmf00-wekaio, line 901: doesn't this bloat the stack by one byte? Understand the actual load is zero so it would be optimised, but does it on debug builds? I guess its totally fine tho, would be cool if D explicitly specifies, whether or not, for zero-sized types, the enforcement of unique address per instance within the same scope. I guess this happens with C, no?

ljmf00-wekaio (approved):

gen/llvmhelpers.cpp
JohanEngelen: > doesn't this bloat the stack by one byte? 

No it does not.  By the way, this new IR emission is already the behavior for zero-sized function parameters inside the function, where we allocate stack space for local storage of the parameter (in case someone writes to it).

PR #4826, opened 2025-01-18
GHA: Integrate Alpine musl job in main workflow

kinke:
Incl. prebuilt package etc.


JohanEngelen:
cool that you are working on this!

kinke:
Okay, should be ready now, after lots of trial & error runs, as usual for these CI MRs. I guess adding a musl AArch64 job as follow-up would be pretty trivial.

Some notes:
* Preliminary runs with Alpine v3.21 and LLVM 19 exhibited LTO problems IIRC. So sticking with v3.20 / LLVM 17 for now.
* I couldn't find an Alpine package with *static* LLD libs. So no `LDC_WITH_LLD` / `-link-internally` support.
* The bundled `ldc-prof{data,gen}` tools aren't linked with `-static`, so depend on musl .so. That would require more CMake fiddling.
* `std.net.curl` support depends on loading libcurl dynamically, so doesn't work (at runtime) with fully static executables. So bundled `dub` and `reggae` depend on musl .so as well (not linked with `-Xcc=-static`).
  * The bundled `ldc-build-runtime` tool is fully static, so won't be able to download the source zip automatically. Again, would require more CMake fiddling.

JohanEngelen:
Very cool! Let's fix the static linking later.

runtime/druntime/test/shared/src/finalize.d
```diff
@@ -58,8 +58,20 @@ void main(string[] args)
     auto r = Runtime.unloadLibrary(h);
     if (!r)
         assert(0);
-    if (finalizeCounter != 4)
-        assert(0);
+    version (darwin)
```
JohanEngelen, line 61: nitpick, replace `darwin` with `OSX` because it is more common throughout the source code.

.github/workflows/main.yml
```diff
@@ -53,6 +53,27 @@ jobs:
...
+            extra_cmake_flags: >-
+              -DBUILD_LTO_LIBS=ON
+              -DLLVM_IS_SHARED=OFF
+              -DLDC_ENABLE_PLUGINS=OFF
+              -DLDC_DYNAMIC_COMPILE=OFF
+              -DLDC_INSTALL_LTOPLUGIN=OFF
+              -DCOMPILER_RT_LIBDIR_OS=linux
+              -DLDC_INSTALL_LLVM_RUNTIME_LIBS_ARCH=x86_64
+              -DTEST_COMPILER_RT_LIBRARIES="profile;lsan;asan;msan;fuzzer"
+              -DCMAKE_EXE_LINKER_FLAGS=-static-libstdc++
```
JohanEngelen, line 71: perhaps it is as simple as `-DCMAKE_EXE_LINKER_FLAGS="-static-libstdc++;-static"` to get fully statically linked ldc_profgen (cpp-only executables) ?  I'm OK to leave as is for now.

JohanEngelen (approved):

kinke:
Btw, I was surprised that I could link & run a Phobos-hello-world on my Ubuntu 24 box - with the bundled compiler, druntime and Phobos targeting a musl triple. No idea if that's just lucky or whether musl is a strict compatible subset of glibc API-wise.

runtime/druntime/test/shared/src/finalize.d
kinke: Yeah I know it's deprecated (unfortunately), but it's more convenient than the usual https://github.com/ldc-developers/ldc/blob/9de21fdd8ff0a30adcda5b05b9c64c55801353ff/runtime/druntime/src/core/demangle.d#L15-L22 which would be required for the other Apple platforms too. There are 3 existing `version (darwin)` usages for this `shared` druntime integration test, so I added the 4th one, out of laziness of course. ;)

.github/workflows/main.yml
kinke: Yes that would include the 2 pure C++ LLVM tools, but unfortunately also the static Phobos test runner, where the `std.net.curl` unittests then fail...

kinke: And `D_LINKER_ARGS` apply to all D executables built via `build_d_executable()` (now for `LDC_LINK_MANUALLY=OFF` too, where it was previously ignored). And that unfortunately also includes the `ldc-build-runtime` tool.

So I think we need a dedicated CMake option for fully-static linking (on Posix), adding `-static` for suited executables only (no `ldc-build-runtime`, but the PGO LLVM tools).

kinke: Okay I've pushed a commit to that effect, and now all but dub/reggae/ldc-build-runtime executables are fully static.

PR #4835, opened 2025-02-09
run on and build for Windows on ARM64

rainers:
This allows to build and run the runtime with ldc-build-runtime and option `-march=arm64` on Windows on ARM. The compiler used is the one for x64 (it shows some floating-point inaccuracies in the x86-emulator).

All druntime unittests pass but the two disabled ones involving chained exceptions in fibers.

In phobos a std.process test fails (but also for emulated x64) and a tests in std.outbuffer indicates that there is still an issue with va_arg handling. Without a curl build, the respective tests fail, too.

I haven't yet managed to run the compiler test suite on the ARM machine.

### DIFF
--- a/dmd/root/longdouble.d
+++ b/dmd/root/longdouble.d
@@ -342,7 +342,7 @@ void ld_setull(longdouble_soft* pthis, ulong d)
     {
         // emulator accuracy not good enough when running on Windows on ARM,
         // so avoid chopping off small numbers
-        version(all)
+        version(LDC)
         {
             if (!(d & (1L << 63)))
             {
--- a/dmd/root/longdouble.d
+++ b/dmd/root/longdouble.d
@@ -342,17 +342,14 @@ void ld_setull(longdouble_soft* pthis, ulong d)
     {
         // emulator accuracy not good enough when running on Windows on ARM,
         // so avoid chopping off small numbers
-        version(LDC)
+        if (!(d & (1L << 63)))
         {
-            if (!(d & (1L << 63)))
+            asm nothrow @nogc pure @trusted
             {
-                asm nothrow @nogc pure @trusted
-                {
-                    fild qword ptr d;
-                }
-                mixin(fstp_parg!("pthis"));
-                return;
+                fild qword ptr d;
             }
+            mixin(fstp_parg!("pthis"));
+            return;
         }
         d ^= (1L << 63);
         auto pTwoPow63 = &twoPow63;
--- a/dmd/root/longdouble.d
+++ b/dmd/root/longdouble.d
@@ -338,9 +338,23 @@ void ld_setll(longdouble_soft* pthis, long d)
 
 void ld_setull(longdouble_soft* pthis, ulong d)
 {
-    d ^= (1L << 63);
     version(AsmX86)
     {
+        // emulator accuracy not good enough when running on Windows on ARM,
+        // so avoid chopping off small numbers
+        version(all)
+        {
+            if (!(d & (1L << 63)))
+            {
+                asm nothrow @nogc pure @trusted
+                {
+                    fild qword ptr d;
+                }
+                mixin(fstp_parg!("pthis"));
+                return;
+            }
+        }
+        d ^= (1L << 63);
         auto pTwoPow63 = &twoPow63;
         mixin(fld_parg!("pTwoPow63"));
         asm nothrow @nogc pure @trusted
--- a/runtime/druntime/src/core/internal/vararg/aarch64.d
+++ b/runtime/druntime/src/core/internal/vararg/aarch64.d
@@ -14,7 +14,7 @@ module core.internal.vararg.aarch64;
 
 version (AArch64):
 
-// Darwin uses a simpler varargs implementation
+// Darwin and Windows use a simpler varargs implementation
 version (OSX) {}
 else version (iOS) {}
 else version (TVOS) {}
--- a/runtime/druntime/src/core/internal/vararg/aarch64.d
+++ b/runtime/druntime/src/core/internal/vararg/aarch64.d
@@ -19,6 +19,7 @@ version (OSX) {}
 else version (iOS) {}
 else version (TVOS) {}
 else version (WatchOS) {}
+else version (CRuntime_Microsoft) {}
 else:
 
 import core.stdc.stdarg : alignUp;
--- a/runtime/druntime/src/core/thread/fiber.d
+++ b/runtime/druntime/src/core/thread/fiber.d
@@ -373,18 +373,10 @@ private
                 stp x27, x28, [sp, #-16]!;
                 stp x29, x31, [sp, #-16]!; // fp,0
 
-                mov x19, v8.d[0];
-                mov x20, v9.d[0];
-                stp x19, x20, [sp, #-16]!;
-                mov x19, v10.d[0];
-                mov x20, v11.d[0];
-                stp x19, x20, [sp, #-16]!;
-                mov x19, v12.d[0];
-                mov x20, v13.d[0];
-                stp x19, x20, [sp, #-16]!;
-                mov x19, v14.d[0];
-                mov x20, v15.d[0];
-                stp x19, x20, [sp, #-16]!;
+                stp d8, d9, [sp, #-16]!;
+                stp d10, d11, [sp, #-16]!;
+                stp d12, d13, [sp, #-16]!;
+                stp d14, d15, [sp, #-16]!;
 
                 ldr x19, [x18];
                 ldr x20, [x18, #8];
@@ -405,18 +397,10 @@ private
                 str x19, [x18];
 
                 // load saved state from new stack
-                ldp x19, x20, [sp], #16;
-                mov v15.d[0], x20;
-                mov v14.d[0], x19;
-                ldp x19, x20, [sp], #16;
-                mov v13.d[0], x20;
-                mov v12.d[0], x19;
-                ldp x19, x20, [sp], #16;
-                mov v11.d[0], x20;
-                mov v10.d[0], x19;
-                ldp x19, x20, [sp], #16;
-                mov v9.d[0], x20;
-                mov v8.d[0], x19;
+                ldp d14, d15, [sp], #16;
+                ldp d12, d13, [sp], #16;
+                ldp d10, d11, [sp], #16;
+                ldp d8, d9, [sp], #16;
 
                 ldp x29, x31, [sp], #16; // fp,0
                 ldp x27, x28, [sp], #16;
--- a/runtime/druntime/src/core/thread/fiber.d
+++ b/runtime/druntime/src/core/thread/fiber.d
@@ -363,59 +363,59 @@ private
         else version (AsmAArch64_Windows)
         {
             pragma(LDC_never_inline);
-            __asm(
+            asm pure nothrow @nogc
+            {
                 `// save current stack state (similar to posix version in threadasm.S)
-                stp x19, x20, [sp, #-16]!
-                stp x21, x22, [sp, #-16]!
-                stp x23, x24, [sp, #-16]!
-                stp x25, x26, [sp, #-16]!
-                stp x27, x28, [sp, #-16]!
-                stp fp,  lr,  [sp, #-16]!
-                mov x19, sp               // no need to scan FP registers, so snapshot sp here
+                stp x19, x20, [sp, #-16]!;
+                stp x21, x22, [sp, #-16]!;
+                stp x23, x24, [sp, #-16]!;
+                stp x25, x26, [sp, #-16]!;
+                stp x27, x28, [sp, #-16]!;
+                stp fp,  lr,  [sp, #-16]!;
+                mov x19, sp;               // no need to scan FP registers, so snapshot sp here
 
-                stp d8,  d9,  [sp, #-16]!
-                stp d10, d11, [sp, #-16]!
-                stp d12, d13, [sp, #-16]!
-                stp d14, d15, [sp, #-16]!
+                stp d8,  d9,  [sp, #-16]!;
+                stp d10, d11, [sp, #-16]!;
+                stp d12, d13, [sp, #-16]!;
+                stp d14, d15, [sp, #-16]!;
 
-                ldr x20, [x18, #8]        // read stack range from TEB
-                ldr x21, [x18, #16]
-                stp x20, x21, [sp, #-16]!
+                ldr x20, [x18, #8];        // read stack range from TEB
+                ldr x21, [x18, #16];
+                stp x20, x21, [sp, #-16]!;
 
-                ldr x20, [x18, #0x1478]   // read Deallocation Stack
-                ldr w21, [x18, #0x1748]   // read GuaranteedStackBytes
-                stp x20, x21, [sp, #-16]!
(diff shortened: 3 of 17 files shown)


runtime/druntime/src/core/thread/fiber.d
```diff
@@ -64,6 +64,9 @@ version (Windows)
     import core.stdc.stdlib : malloc, free;
     import core.sys.windows.winbase;
     import core.sys.windows.winnt;
+
+    version (LDC)
+        version = LDC_Windows;
```
rainers, line 69: Version LDC_Windows was never set in this but used below. Is this deliberate or was it lost in the thread/fiber refatorings?

JohanEngelen: can you put this in a separate PR?  (related to x86, could make bughunting a little easier in the future,thanks!)

rainers: see https://github.com/ldc-developers/ldc/pull/4836

kinke:
> All druntime unittests pass but the two disabled ones involving chained exceptions in fibers.
> 
> In phobos a std.process test fails (but also for emulated x64) and a tests in std.outbuffer indicates that there is still an issue with va_arg handling. Without a curl build, the respective tests fail, too.

Awesome! I'd have expected much more needed tweaks to get so far already.

Wrt. `va_arg` - `core.stdc.stdarg` has a `version(Win64)` block *before* `version(AAPCS64)`: https://github.com/ldc-developers/ldc/blob/45670c2948b9a9a71400c89843a3d40a90a06385/runtime/druntime/src/core/stdc/stdarg.d#L181

runtime/druntime/src/core/thread/fiber.d
```diff
@@ -352,6 +360,74 @@ private
                 "~{xmm6},~{xmm7},~{xmm8},~{xmm9},~{xmm10},~{xmm11},~{xmm12},~{xmm13},~{xmm14},~{xmm15}"
             );
         }
+        else version (AsmAArch64_Windows)
+        {
+            pragma(LDC_never_inline);
+            asm nothrow @nogc {
+            "
```
kinke, line 367: This looks similar to the Posix version in https://github.com/ldc-developers/ldc/blob/45670c2948b9a9a71400c89843a3d40a90a06385/runtime/druntime/src/core/threadasm.S#L805-L886 (which AFAICT uses `D8` to access the lower 64-bits of `V8`). Have you tried that one?

rainers:
> Awesome! I'd have expected much more needed tweaks to get so far already.

I was surprised aswell :)
 
> Wrt. `va_arg` - `core.stdc.stdarg` has a `version(Win64)` block _before_ `version(AAPCS64)`:

Turns out it was about va_list, that seems to be a pointer only, as on Darwin. Seems fixed now.

runtime/druntime/src/core/thread/fiber.d
rainers: Indeed, knowing that snippet would have made my life easier. I'm a newbie to AArch64 assembly. I suspect building the .S file on Windows is a problem, so keeping the inline assembly (made a bit nicer if possible) is ok?

kinke: No idea about the MSVC ARM toolchain support for assembly, but I suppose a `clang-cl` assembly driver would work. But sure okay to keep it; might be worth emphasizing that the x18 stuff is Windows-specific (similar to `GS` on Windows x86_64), the rest identical to Posix.

rainers:
Finally figured out what was wrong with the fiber code. All unittests pass now for druntime and phobos (curl not tested, one std.process test fails both for arm64 and x64).

runtime/druntime/src/core/thread/fiber.d
```diff
@@ -352,6 +360,63 @@ private
...
+
+                ldp fp,  lr,  [sp], #16;
+                ldp x27, x28, [sp], #16;
+                ldp x25, x26, [sp], #16;
+                ldp x23, x24, [sp], #16;
+                ldp x21, x22, [sp], #16;
+                ldp x19, x20, [sp], #16;
+
+                ret;
+            " : ; }
```
kinke, line 418: Just wondering, was the colon required?

You don't specify any clobbers (the list would be huge!), but as this function is never inlined, that shouldn't be a problem.

runtime/druntime/src/ldc/eh_msvc.d
```diff
@@ -540,6 +540,34 @@ void msvc_eh_terminate() nothrow @naked
             "~{memory},~{flags},~{rbp},~{rsp},~{rax},~{rbx},~{rdx}"
         );
     }
+    else version (AArch64)
+    {
+        __asm(`
```
kinke, line 545: Regular GDC-style `asm { <string>}` would be preferred. I only use LDC-specific `__asm` when it returns a value (instead of writing the value to a dummy variable and then returning that variable, making a one-liner 3 lines long).

runtime/druntime/src/ldc/eh_msvc.d
```diff
@@ -540,6 +540,34 @@ void msvc_eh_terminate() nothrow @naked
...
+            mov w0, #0;   // return EXCEPTION_CONTINUE_SEARCH
+            ldp fp,lr,[sp],#16;
+            ldp x19,x20,[sp],#0x10;
+            autibsp; // uses lr,sp and x19 as input to hash
+            ret;
+
+        1:
+            ldp fp,lr,[sp],#16;
+            ret`,
+            ""
```
kinke, line 568: No clobbers again, but this function is never called directly (and so possibly inlined) AFAIK.

runtime/druntime/src/core/internal/vararg/aarch64.d
```diff
@@ -19,6 +19,7 @@ version (OSX) {}
 else version (iOS) {}
 else version (TVOS) {}
 else version (WatchOS) {}
+else version (CRuntime_Microsoft) {}
```
kinke, line 22: [would be nice if you could tweak the comment above wherever adding this MSVCRT exclusion]

kinke:
Oh and please add a changelog.md entry. Rest LGTM, thanks!

runtime/druntime/src/core/thread/fiber.d
rainers: IIRC ldc complained about a syntax error without the colon.

runtime/druntime/src/core/thread/fiber.d
```diff
@@ -1611,6 +1676,41 @@ private:
...
+            push( 0x00000000_00000000 );                            // X21
+            push( 0x00000000_00000000 );                            // X22
+            push( 0x00000000_00000000 );                            // X23
+            push( 0x00000000_00000000 );                            // X24
+            push( 0x00000000_00000000 );                            // X25
+            push( 0x00000000_00000000 );                            // X26
+            push( 0x00000000_00000000 );                            // X27
+            push( 0x00000000_00000000 );                            // X28
+            push( cast(size_t) &fiber_entryPoint );                 // X30 (lr)
+            push( cast(size_t) nextfp );                            // X29 (fp)
```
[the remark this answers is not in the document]
rainers: The two are correct, as the `stp` instruction reads the second register from an offset 8. The other registers above are actually reversed pairwise. Maybe it would be nicer to reverse them in the `stp`/`ldp` instructions.

runtime/druntime/src/ldc/eh_msvc.d
rainers: I can try that, I moved this from the section below that was used before version LDC_Windows was set.

runtime/druntime/src/ldc/eh_msvc.d
rainers: Yeah, it's only passed as a terminate handler.

kinke:
[`dmd.vsoptions` will need some extensions for AArch64 support too, it currently just features a `bool x64` argument to differentiate between x86 and x64.]

dmd/root/longdouble.d
```diff
@@ -338,9 +338,23 @@ void ld_setll(longdouble_soft* pthis, long d)
 
 void ld_setull(longdouble_soft* pthis, ulong d)
 {
-    d ^= (1L << 63);
     version(AsmX86)
     {
+        // emulator accuracy not good enough when running on Windows on ARM,
+        // so avoid chopping off small numbers
+        version(LDC)
```
kinke, line 345: So you changed this from `all` to `LDC`? Because `longdouble_soft` is unused with a DMD host compiler?

runtime/druntime/src/core/thread/fiber.d
```diff
@@ -352,6 +360,63 @@ private
                 "~{xmm6},~{xmm7},~{xmm8},~{xmm9},~{xmm10},~{xmm11},~{xmm12},~{xmm13},~{xmm14},~{xmm15}"
             );
         }
+        else version (AsmAArch64_Windows)
+        {
+            pragma(LDC_never_inline);
+            __asm(
```
kinke, line 366: And you've switched from standard GDC-style asm to LDC-specific `__asm` here?

dmd/root/longdouble.d
rainers: Hmm, for some reason I was deluded into thinking that dmd wasn't affected because it doesn't build for ARM. But running dmd on the ARM system to build x86 executable could be useful, too. Compiling
```D
pragma(msg, cast(double)cast(ulong)314);
```
yields `0.0`, though. I'll remove the version completely.

runtime/druntime/src/core/thread/fiber.d
rainers: Hmm, still confused which one is which ;) I was just adopting to what was above that, but obviously in the wrong place. I'll revert and change it in eh_msvc.d instead.

runtime/druntime/src/ldc/eh_msvc.d
```diff
@@ -542,27 +542,28 @@ void msvc_eh_terminate() nothrow @naked
...
+            ret;
 
         1:
-            ldp fp,lr,[sp],#16
-            ret`,
-            "~{memory},~{x0},~{x19},~{x20}"
-        );
+            ldp fp,lr,[sp],#16;
+            ret;`
+            : : : "x0", "x19", "x20";
```
kinke, line 565: The semicolons shouldn't be needed, line breaks should do - but if you prefer the semicolons, absolutely fine.

`"memory"` as clobber would be supported too, but I guess you don't touch any memory here, just registers.

kinke (approved):

kinke:
As a follow-up, we could build and bundle the ARM64 libraries with the windows-multilib installation package, incl. an `ldc2.conf` tweak. So that cross-compiling via `-mtriple=arm64-windows-msvc` works out of the box.

runtime/druntime/src/ldc/eh_msvc.d
rainers: Actually, I liked it better without ;) LDC just kept complaining about missing semicolons, but maybe this was just the one after the string.

rainers:
> As a follow-up, we could build and bundle the ARM64 libraries with the windows-multilib installation package, incl. an `ldc2.conf` tweak. So that cross-compiling via -mtriple=arm64-windows-msvc works out of the box (edit: with an installed ARM64 MSVC toolchain).

I added the entry in ldc2.conf yesterday: when compiling with -march=arm64 the triple came out as aarch64-pc-windows-msvc. With the ARM64 environment set up with vcvarsamd64.cmd it then linked without issues.

Is there a way to cross-compile and run the tests in a CI on WoA?

kinke:
Btw, out of interest - how slow is the x64 emulator on the ARM box? Comparable to Apple's Rosetta 2?

runtime/druntime/src/ldc/eh_msvc.d
kinke: Yeah most likely just that one; in GDC-style `asm`, the block needs statements closed with semicolons, each statement starting with an asm string literal, then possibly colons and the constraints.

rainers:
> Btw, out of interest - how slow is the x64 emulator on the ARM box? Comparable to Apple's Rosetta 2?

No real measurements so far, but the  emulator performance seems pretty good. When compiling for ARM64EC, you can also mix Intel and ARM binaries in a single process, or even within one binary. Marshaling between native and emulated code causes some notable slowdowns, though.

rainers:
I just tested Visual D on ARM (VS doesn't support x86_64 extensions as it is not compiled as ARM64EC), but unfortunately it crashes when loading a D source file. Not analyzed yet in detail, but I suspect an issue with std.concurrency used in communication with dmdserver.xe .

kinke:
> VS doesn't support x86_64 extensions as it is not compiled as ARM64EC

Oh wow. This is *exactly* what I thought this ARM64EC would be for, keeping up support for all the existing x86 plugins. lol

rainers:
> but unfortunately it crashes when loading a D source file.

Turned out stack scanning was missing the proper top of stack. Fixed now and Visual D runs (only very little testing done). The debugger extension doesn't yet work, though, but that's a C++ DLL.

rainers:
> Oh wow. This is _exactly_ what I thought this ARM64EC would be for, keeping up support for all the existing x86 plugins. lol

Yeah, I was pretty surprised, too. Not sure if there are any downsides to ARM64EC but a slightly different ABI. I doubt there are any performance issues if you just build for ARM64 anyway, but it might keep people from porting to ARM to begin with.

kinke:
2 incompatible ABIs for a 'new' platform - IMO a recipe for disaster. Especially if the x86 emulation performance isn't too bad, and AFAICT, all ARM64EC would enable is better performance by having some parts of a process in native code. E.g., they could still provide an x86 VS build as install option on ARM, to keep the x86 plugins working. If Microsoft themselves aren't using ARM64EC for stuff like VS (and Office - no idea if they do), then it's dead on arrival to me.

Apple has shown that transitioning from Intel to ARM can be quite smooth with a built-in emulator, no need for an extra ABI to be able to mix architectures in a single process.

---

Feel free to merge when ready. [You could bump the Phobos submodule here and add the PR refs to the changelog entry.]

denizzzka:
> If Microsoft themselves aren't using ARM64EC for stuff like VS (and Office - no idea if they do)

According to [this article](http://www.emulators.com/docs/abc_arm64ec_explained.htm) - they do

p0nce:
Here is my understanding about what can host what in Windows arm64.

423104004-4c0f6233-ae2b-4045-92a1-ad1ca0f885ea-1

If you are not making a host, but just want to make a plugin, wrapping an arm64 binary inside "arm64X" is probably a linker trick and perhaps doesn't need arm64ec, not sure.

p0nce:
> Btw, out of interest - how slow is the x64 emulator on the ARM box? Comparable to Apple's Rosetta 2?

@kinke making a single quick measurements using LDC 1.41-beta1 using simple video generation

This program only exercices canvasity (relatively slow) rasterizer and a single RGB => YUV conversion, still it should be in the ballpark for optimized code, the 4k Y4M is piped into ffmpeg arm64 which doesn't seem to be a bottleneck.

x86_64 runtime: 48 sec
arm64 runtime: 30 sec

that is a 0.625 ratio, very similar to Rosetta 2. But this is only a single measure.

PR #4840, opened 2025-02-18
ppc64(le): Add an option to use IEEE long double ABI on Linux [2]

kinke:
Based on #4833, with my proposed changes.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -681,16 +681,10 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("PPC64");
     registerPredefinedFloatABI("PPC_SoftFloat", "PPC_HardFloat");
     if (triple.getOS() == llvm::Triple::Linux) {
-      VersionCondition::addPredefinedGlobalIdent(
-          triple.getArch() == llvm::Triple::ppc64 ? "ELFv1" : "ELFv2");
-      if (opts::mABI == "ieeelongdouble") {
-        if (triple.getEnvironment() == llvm::Triple::GNU) {
-          // Only GLibc needs this for IEEELongDouble
-          VersionCondition::addPredefinedGlobalIdent("D_PPCUseIEEE128");
-        } else {
-          warning(Loc(), "float ABI 'ieeelongdouble' is not supported by the target system");
-        }
-      }
+      const llvm::SmallVector<llvm::StringRef> features{};
+      const std::string abi = getABI(triple, features);
+      VersionCondition::addPredefinedGlobalIdent(abi == "elfv1" ? "ELFv1"
+                                                                : "ELFv2");
     }
     break;
   case llvm::Triple::arm:
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -683,6 +683,10 @@ void registerPredefinedTargetVersions() {
     if (triple.getOS() == llvm::Triple::Linux) {
       VersionCondition::addPredefinedGlobalIdent(
           triple.getArch() == llvm::Triple::ppc64 ? "ELFv1" : "ELFv2");
+      if (opts::mABI == "ieeelongdouble" && triple.isOSGlibc()) {
+        // Only GLibc needs this for IEEELongDouble
+        VersionCondition::addPredefinedGlobalIdent("D_PPCUseIEEE128");
+      }
     }
     break;
   case llvm::Triple::arm:
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -683,9 +683,13 @@ void registerPredefinedTargetVersions() {
     if (triple.getOS() == llvm::Triple::Linux) {
       VersionCondition::addPredefinedGlobalIdent(
           triple.getArch() == llvm::Triple::ppc64 ? "ELFv1" : "ELFv2");
-      if (opts::mABI == "ieeelongdouble" && triple.isOSGlibc()) {
-        // Only GLibc needs this for IEEELongDouble
-        VersionCondition::addPredefinedGlobalIdent("D_PPCUseIEEE128");
+      if (opts::mABI == "ieeelongdouble") {
+        if (triple.getEnvironment() == llvm::Triple::GNU) {
+          // Only GLibc needs this for IEEELongDouble
+          VersionCondition::addPredefinedGlobalIdent("D_PPCUseIEEE128");
+        } else {
+          warning(Loc(), "float ABI 'ieeelongdouble' is not supported by the target system");
+        }
       }
     }
     break;
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -22,6 +22,7 @@
 #include "dmd/statement.h"
 #include "dmd/target.h"
 #include "dmd/template.h"
+#include "driver/cl_options.h"
 #include "driver/cl_options_instrumentation.h"
 #include "driver/timetrace.h"
 #include "gen/abi/abi.h"
@@ -403,6 +404,23 @@ void addModuleFlags(llvm::Module &m) {
   if (opts::fCFProtection == opts::CFProtectionType::Branch ||
       opts::fCFProtection == opts::CFProtectionType::Full) {
     m.setModuleFlag(ModuleMinFlag, "cf-protection-branch", ConstantOneMetadata);
+  }
+
+  // Target specific flags
+  const auto ModuleErrFlag = llvm::Module::Error;
+  switch (global.params.targetTriple->getArch()) {
+  case llvm::Triple::ppc64:
+  case llvm::Triple::ppc64le:
+    if (opts::mABI == "ieeelongdouble") {
+      const auto ConstantIEEE128String = llvm::MDString::get(gIR->context(), "ieeequad");
+      m.setModuleFlag(ModuleErrFlag, "float-abi", ConstantIEEE128String);
+    } else {
+      const auto ConstantIBM128String = llvm::MDString::get(gIR->context(), "doubledouble");
+      m.setModuleFlag(ModuleErrFlag, "float-abi", ConstantIBM128String);
+    }
+    break;
+  default:
+    break;
   }
 }
 
--- a/gen/modules.cpp
+++ b/gen/modules.cpp
@@ -411,7 +411,7 @@ void addModuleFlags(llvm::Module &m) {
   switch (global.params.targetTriple->getArch()) {
   case llvm::Triple::ppc64:
   case llvm::Triple::ppc64le:
-    if (opts::mABI == "ieeelongdouble") {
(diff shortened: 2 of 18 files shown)


gen/target.cpp
```diff
@@ -84,14 +87,28 @@ llvm::Type *getRealType(const llvm::Triple &triple) {
...
     }
-    // dual-ABI complications: PPC only has IEEE 128-bit quad precision on
-    // Linux, IBM double-double is available on both AIX and Linux.
-    return triple.isOSLinux() && realTypeEncoding == RealTypeEncoding::IEEEQuad
-               ? LLType::getFP128Ty(ctx)
-               : LLType::getPPC_FP128Ty(ctx);
+#if defined(__linux__) && defined(__powerpc64__)
+    // for a PowerPC64 Linux build:
+    // default to the C++ host compiler's `long double` ABI
+    switch (std::numeric_limits<long double>::digits) {
```
liushuyu, line 99: I would use:

```suggestion
    switch (__LDBL_MANT_DIG__) {
```

This one is defined by the compiler directly and I have tested to contain correct information.

gen/target.cpp
```diff
@@ -46,9 +47,11 @@ llvm::Type *getRealType(const llvm::Triple &triple) {
 
   auto &ctx = getGlobalContext();
 
-  // If user specified double precision, use it unconditionally.
-  if (realTypeEncoding == RealTypeEncoding::Double) {
+  // If user specified double or quad precision, use it unconditionally.
+  if (realPrecision == RealPrecision::Double) {
     return LLType::getDoubleTy(ctx);
+  } else if (realPrecision == RealPrecision::Quad) {
+    return LLType::getFP128Ty(ctx);
```
liushuyu, line 54: Have you tested if this will blow up x86 builds?

kinke: Android x86_64 is already using it, so I guess that's fine. - This is an advanced option (hence hidden), for pros only - druntime and Phobos need to match too etc. (no problem for simple betterC stuff in wasm, just need to compile everything with the override). If LLVM can't cope with this format for a target's codegen, it'll error out. [I'm a fan of giving users full control and letting them explore the compiler/LLVM limits.]

gen/target.cpp
kinke: Fine by me. - We should probably only forward the host precision when targeting a *Linux* ppc64 target, not e.g. default to IEEE quad when cross-compiling to a ppc64 FreeBSD target.

gen/target.cpp
kinke: [Plus there are some compiler assumptions wrt. `real` precision, in existing `TargetABI` implementations - as the `real` precision wasn't configurable so far.]

liushuyu: > [I'm a fan of giving users full control and letting them explore the compiler/LLVM limits.]

I see. Then that's fine by me. [Although LLVM is quite fragile when operating beyond its comfort zone]

JohanEngelen: `RealPrecision::DoubleDouble` should be handled here too (i.e. the user has set it explicitly)

gen/target.cpp
```diff
@@ -64,9 +85,34 @@ llvm::Type *getRealType(const llvm::Triple &triple) {
...
+    switch (std::numeric_limits<long double>::digits) {
+    case 113:
+      return LLType::getFP128Ty(ctx);
+    case 106:
+      return LLType::getPPC_FP128Ty(ctx);
+    case 53:
+      return LLType::getDoubleTy(ctx);
+    default:
+      llvm_unreachable("Unexpected host C++ 'long double' precision for a "
+                       "PowerPC64 target!");
```
JohanEngelen, line 108: perhaps print the actual `__LDBL_MANT_DIG__` number in the diagnostic? (will help troubleshooting when this error triggers)

gen/target.cpp
JohanEngelen: when changed to the compiler macro, could this be #if sequence such that failure happens during compilation rather than at runtime?

gen/target.cpp
liushuyu: > `RealPrecision::DoubleDouble` should be handled here too (i.e. the user has set it explicitly)

Right, and for all other architectures, it should throw an error.

gen/target.cpp
```diff
@@ -64,9 +85,34 @@ llvm::Type *getRealType(const llvm::Triple &triple) {
...
+    }
```
liushuyu, line 109: > perhaps print the actual __LDBL_MANT_DIG__ number in the diagnostic? (will help troubleshooting when this error triggers)

@JohanEngelen  Like this?

```suggestion
#if __LDBL_MANT_DIG__ == 113
      return LLType::getFP128Ty(ctx);
#elif  __LDBL_MANT_DIG__ == 106
      return LLType::getPPC_FP128Ty(ctx);
#elif __LDBL_MANT_DIG__ == 53
      return LLType::getDoubleTy(ctx);
#else
#error "Unexpected host C++ 'long double' precision for a PowerPC64 target!"
#endif
```

gen/target.cpp
kinke: > `RealPrecision::DoubleDouble` should be handled here too (i.e. the user has set it explicitly)

*That* I wanted to avoid though. I originally only wanted to expose the 2 standard IEEE formats as overrides, but unfortunately we need a way to make a native-PPC64 build with default IEEE quad ABI still cross-compile to the old doubledouble ABI. (I still don't wanna expose x87 if we can help it - although admittedly it could be useful for MSVC targets, to unlock the x87 `real`, which the compiler itself could use as `real_t` - bye-bye `dmd.root.longdouble`...)

But yeah, we can bail out as a compromise if this option is used for non-ppc64 targets.

kinke:
> should the D mangling scheme also be extended to differentiate the types?

I'd say D doesn't need to, assuming PPC will remain the only architecture  (well, AFAIK) with such a `long double` ABI change. [And users now starting to use `-real-precision` should IMO remain an uncommon special case, for wasm and the like, not needing the nicety of undefined symbols.]

driver/main.cpp
```diff
@@ -681,8 +681,10 @@ void registerPredefinedTargetVersions() {
     VersionCondition::addPredefinedGlobalIdent("PPC64");
     registerPredefinedFloatABI("PPC_SoftFloat", "PPC_HardFloat");
     if (triple.getOS() == llvm::Triple::Linux) {
-      VersionCondition::addPredefinedGlobalIdent(
-          triple.getArch() == llvm::Triple::ppc64 ? "ELFv1" : "ELFv2");
+      const llvm::SmallVector<llvm::StringRef> features{};
+      const std::string abi = getABI(triple, features);
+      VersionCondition::addPredefinedGlobalIdent(abi == "elfv1" ? "ELFv1"
+                                                                : "ELFv2");
```
kinke, line 687: According to https://github.com/dlang/dmd/blob/e082ce247afa6cf41c552ec57db2e95c3f7c0dac/druntime/src/etc/valgrind/valgrind.h#L154-L159, little-endian uses v2, big-endian v1.

gen/modules.cpp
```diff
@@ -404,6 +404,23 @@ void addModuleFlags(llvm::Module &m) {
...
+  const auto ModuleErrFlag = llvm::Module::Error;
+  switch (global.params.targetTriple->getArch()) {
+  case llvm::Triple::ppc64:
+  case llvm::Triple::ppc64le:
+    if (target.RealProperties.mant_dig == 113) {
+      const auto ConstantIEEE128String = llvm::MDString::get(gIR->context(), "ieeequad");
+      m.setModuleFlag(ModuleErrFlag, "float-abi", ConstantIEEE128String);
+    } else if (target.RealProperties.mant_dig == 106) {
+      const auto ConstantIBM128String = llvm::MDString::get(gIR->context(), "doubledouble");
+      m.setModuleFlag(ModuleErrFlag, "float-abi", ConstantIBM128String);
```
kinke, line 418: I take it this is just informational, or is this supported by some toolchains?

gen/target.cpp
```diff
@@ -197,6 +244,18 @@ void Target::_init(const Param &params) {
...
+        CTFloat::parse("0x1.fffffffffffff7ffffffffffff8p1023", isOutOfRange);
+    RealProperties.min_normal = CTFloat::parse("0x1p-969", isOutOfRange);
+    RealProperties.epsilon =
+        CTFloat::parse("0x0.000000000000000000000000008p-969", isOutOfRange);
+    RealProperties.dig = 31;
+    RealProperties.mant_dig = 106;
+    RealProperties.max_exp = 1024;
+    RealProperties.min_exp = -968;
+    RealProperties.max_10_exp = 308;
+    RealProperties.min_10_exp = -291;
```
kinke, line 258: I was surprised by this being quite a bit greater than `double.min_10_exp` (-307). How did you get these numbers? Querying GDC or the C++ compiler?

liushuyu: > I was surprised by this being quite a bit greater than `double.min_10_exp` (-307). How did you get these numbers? Querying GDC or the C++ compiler?

Yes, the numbers are obtained from GCC/Clang. IBM stated due to the peculiarity of the type, those values are best-effort approximations.

gen/modules.cpp
liushuyu: > I take it this is just informational, or is this supported by some toolchains?

I see Clang emit this information, I am guessing this is to avoid mixing compilation units with different float ABIs when doing LTO.

driver/main.cpp
liushuyu: > According to https://github.com/dlang/dmd/blob/e082ce247afa6cf41c552ec57db2e95c3f7c0dac/druntime/src/etc/valgrind/valgrind.h#L154-L159, little-endian uses v2, big-endian v1.

That's a bad detection: big-endian ppc can use ELFv2, and little-endian ppc can also use ELFv1: https://godbolt.org/z/s5eceeqxs.

kinke:
@liushuyu: Are you able to/have you already tested this on ppc64le, checking whether the results haven't suffered since your original PR?

gen/target.cpp
JohanEngelen: if that works, then that's great yeah :)

liushuyu:
> @liushuyu: Are you able to/have you already tested this on ppc64le, checking whether the results haven't suffered since your original PR?

I have tested your changes on a POWER 9 system and it seems okay.
Some parts of `phobos` are failing, but these are also failing in my PR (also these are not compiler issue).

kinke:
Perfect, thx for testing! So I guess we're set and ready to merge? I'm happy with the overall changes.

liushuyu:
> Perfect, thx for testing! So I guess we're set and ready to merge? I'm happy with the overall changes.

I think it's ready for the merge.

PR #4843, opened 2025-02-21
Preliminary LLVM 20 support

liushuyu:
This pull request adds basic LLVM 20 support for LDC.

LLVM 20 changed a lot of APIs again, especially how passes are structured.

### DIFF
--- a/driver/cl_options_sanitizers.h
+++ b/driver/cl_options_sanitizers.h
@@ -14,7 +14,11 @@
 #pragma once
 
 #include "driver/cl_helpers.h"
+#if LDC_LLVM_VER >= 2000
+#include "llvm/Transforms/Utils/Instrumentation.h"
+#else
 #include "llvm/Transforms/Instrumentation.h"
+#endif
 
 class FuncDeclaration;
 namespace llvm {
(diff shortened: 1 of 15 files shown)


the-horo (changes requested):
In `gen_gccbuiltins.cpp` you can get away with retroactively applying `const` instead of the `LLVM_20_CONST` conditional.

One more change that github doesn't let me submit:
```diff
-    return TableGenMain(argv[0], emit);
+    // Since LLVM-20 `records` needs to be const&
+    return TableGenMain(argv[0], [](raw_ostream &os, auto &records){
+      return emit(os, records);
+    });
```

kinke:
First [LDC-LLVM 20](https://github.com/ldc-developers/llvm-project/tree/ldc-release/20.x) release-candidate binaries are available: https://github.com/ldc-developers/llvm-project/releases/tag/CI

@rainers: FYI, incl. natively-built Windows-on-ARM64 packages, as GitHub Actions now has such runners: https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/. [And they seem significantly faster than the x86_64 runners.]

PR #4846, opened 2025-02-26
CI: Bundle arm64 libraries with Windows multilib package

kinke:


kinke:
FYI @rainers

rainers:
Nice. I think it's failing because cmake builds and runs test executables during the configuration phase that only run on an ARM64 system. Can this be disabled?

rainers:
Not sure what the difference is, but for me, it works when using
```
"c:\Program Files\Microsoft Visual Studio\2022\Preview\vc\Auxiliary\Build\vcvarsamd64_arm64.bat"
```
instead of
```
"c:\Program Files\Microsoft Visual Studio\2022\Preview\Common7\Tools\VsDevCmd.bat" -arch=amd64
```

kinke:
Thx, yeah I guess we might have hit https://developercommunity.visualstudio.com/t/vsdevcmd-creates-invalid-ARM64-build-con/10090339.

kinke:
> BTW: I recently also tried the prebuilt libcurl from https://curl.se/windows/, that worked for the phobos unittests, too. I don't remember why we have to compile libcurl ourselves...

Did you test it on x64 or arm64? Great, I don't think they had usable official Windows builds when I last checked, years ago. I don't really remember why we built it ourselves either, but I think one reason was to have a static .lib and the .exp, to enable linking it statically (while still exporting those symbols, so that Phobos can load them from the executable). Looking at the `lib` dir in the official package, it has 14 `.a` libs for MinGW, unusable for MSVC.

And for 32-bit x86, we might have had to tweak the compile flags to make it work with the MinGW-based libs (disabling `/GS` etc.), but most likely just for the static lib again.

I've checked https://vcpkg.io/en/package/curl.html yesterday (wanting to use some prebuilt libcurl too), but looking at the 'Features' tab with all those dependencies made me quickly close the browser tab again... ;)

rainers:
Sorry for the delay. I tested libcurl successfully on arm64. I guess you are right about the static library build, maybe they even had mingw runtime dependencies back then.

PR #4849, opened 2025-02-26
driver/linker-gcc.cpp: Don't pass -rpath to linker on wasm targets

the-horo:
wasm-ld and lld::wasm don't understand the -rpath argument.

### DIFF
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -475,7 +475,7 @@ void ArgsBuilder::addCppStdlibLinkFlags(const llvm::Triple &triple) {
 void ArgsBuilder::addObjcStdlibLinkFlags(const llvm::Triple &triple) {
   if (linkNoObjc)
     return;
-    
+
   args.push_back("-lobjc");
 }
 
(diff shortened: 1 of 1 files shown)


driver/linker-gcc.cpp
```diff
@@ -613,7 +613,8 @@ void ArgsBuilder::build(llvm::StringRef outputPath,
   }
 
   // -rpath if linking against shared default libs or ldc-jit
-  if (linkAgainstSharedDefaultLibs() || opts::enableDynamicCompile) {
+  if ((linkAgainstSharedDefaultLibs() || opts::enableDynamicCompile)
+      && !triple.isOSBinFormatWasm()) { // wasm-ld doesn't recognize -rpath
```
kinke, line 617: Really needed? wasm will most likely never support shared druntime/Phobos, nor dynamic-compile.

the-horo: Of course I could add it back but it seems more reasonable to ignore `-link-defaultlib-shared`, in my opinion at least

PR #4850, opened 2025-02-26
CI: Prevent hanging druntime/Phobos unit/integration tests

kinke:
I've seen them hanging yesterday for Linux aarch64 GHA, still running that step after 17 mins.


kinke:
Should we hardcode this in CMake instead? It's mainly only the druntime integration tests which might run for a longer time; something generous like 2 mins 'ought to be enough for everyone' perhaps, even slow emulators?

PR #4853, opened 2025-02-28
Bump mimalloc and bundled dlang tools

kinke:


kinke:
Hmm, TLS trouble on macOS with mimalloc v1.8.9 (no v1.8.8 available), which they've worked on according to their changelog.

packaging/mimalloc_version
```diff
@@ -1 +1 @@
-v1.8.7
\ No newline at end of file
+v1.8.9
```
liushuyu, line 1:
```suggestion
v1.9.3
```

Maybe let's try a newer version released in March 2025 to see if the macOS situation has been resolved?

kinke: Same problem unfortunately:
```
Aborting from rt/sections_elf_shared.d(200) Could not determine TLS range
```

What's new is a compile error for Alpine:
```
In file included from /__w/ldc/mimalloc/src/prim/unix/prim.c:36,
                 from /__w/ldc/mimalloc/src/prim/prim.c:25,
                 from /__w/ldc/mimalloc/src/static.c:38:
/usr/include/sys/prctl.h:88:8: error: redefinition of 'struct prctl_mm_map'
   88 | struct prctl_mm_map {
      |        ^~~~~~~~~~~~
In file included from /__w/ldc/mimalloc/src/prim/unix/prim.c:34:
/usr/include/linux/prctl.h:134:8: note: originally defined here
  134 | struct prctl_mm_map {
      |        ^~~~~~~~~~~~
```

liushuyu: It seems like `mimalloc` breaks LDC after this `mimalloc` commit:
```
commit f126b503822addadbdfcc82f940c5a086d3f75dc (HEAD)
Author: Daan <daanl@outlook.com>
Date:   Sun Oct 27 21:10:46 2024 -0700

    update comments, set constructor priority to 101 on macOS

 src/prim/osx/alloc-override-zone.c |  4 ++--
 src/prim/prim.c                    | 10 +---------
 2 files changed, 3 insertions(+), 11 deletions(-)
```

kinke: That v1.8.8 change seems likely; same issue with latest v1.9.7. I guess the problem is that we use `malloc_size()` to get the size of the TLS block on Darwin: https://github.com/ldc-developers/ldc/blob/c719c8503d50db8fe90b795703200d6ac889fea1/runtime/druntime/src/rt/sections_darwin_64.d#L105

mimalloc overrides that symbol. With that mimalloc change in v1.8.8, the CRT constructor priority was *decreased* from 0 (absolutely highest) to 101 [while our hook for druntime registration is of default lowest priority 65535]. So I guess the TLS block isn't allocated with mimalloc's `malloc` anymore (overridden later), so that our query for its size (overridden at that point) yields 0.

kinke: Reverting their priority to 0 in the 2nd commit seems to make it work. No idea whether that's safe to do for our use case, linking mimalloc statically into the ldc2 executable.

PR #4856, opened 2025-03-01
ldc2.conf: Implement ~= concatenating operator [2]

kinke:


kinke:
After countless CI attempts, it looks like the combination of an older host LDC (v1.35.0 with LLVM 16, or v1.37.0 with LLVM 17 - the macos-13 Xcode is LLVM-16 based), incl. skipping the bootstrap compiler build (as for macOS arm64 already), *and* disabling mimalloc (apparently causing the dreaded `libc++abi: Pure virtual function called!`), *and* disabling LTO (the new `config_diag.d` test command otherwise still crashed during unwinding with LDC v1.37.0; couldn't get LTO to work with v1.35.0, still some `liblto.dylib` issue) makes it finally work. Mad.

kinke:
Okay, with LDC v1.35.0 as host/bootstrap compiler, LTO can be restored without breaking `config_diag.d`. [But had to exclude druntime+Phobos from LTO; they were prebuilt with a macos triple without OS version...]

the-horo:
From my little testing it was way easier to make the bug disappear. Either:
- don't build the D parts with lto
- don't build the C++ parts with lto
- don't pass `-defaultlib=phobos2-ldc-lto,druntime-ldc-lto`
- use `-DCOMPILE_D_MODULES_SEPARATELY=ON`
- use `-DLDC_LINK_MANUALLY=ON`

I did stop testing once I saw that `config_diag.d` worked though.

kinke:
Yeah macOS CI is a bit of a black box, very fragile, and last time we had a similar problem (on arm64 back then), nobody could reproduce all the issues (`pure virtual func` (mostly sporadic, not deterministic!), unwinding issues in the compiler)  locally. Probably as it depends on so many different factors - used Xcode version, the host D compiler (incl. whether it's a universal or native macOS package due to https://github.com/ldc-developers/ldc/pull/4857), and now even mimalloc, which was really surprising.

the-horo:
I've tried [-flto=thin](https://github.com/the-horo/ldc/commit/d00d037fb05eb2850d9ad27ad30f979ab1904aef) and it seems fine.

If the environment is so poor I would rather drop optimizations and go with a minimalist flag set that doesn't produce broken executables but not like I have any setup to run this on and, from your comment, it seems that not even if you have a device can you actually debug these issues.

kinke:
> I've tried [-flto=thin](https://github.com/the-horo/ldc/commit/d00d037fb05eb2850d9ad27ad30f979ab1904aef) and it seems fine.

Oh wow, that's interesting.

> If the environment is so poor I would rather drop optimizations and go with a minimalist flag set that doesn't produce broken executables

So far noone has complained, so I guess they are working and sufficiently (CI-)tested. But yeah, it's definitely not a comfortable feeling. As dropping LTO breaks the build (with mimalloc at least), dropping optimizations wasn't really an option so far. But the C++/D interop LTO is IMO quite important, to inline little one-liners etc., so I'd like to keep that. I don't think PGO was ever a problem.

I'm going to test mimalloc once more; disabling that for both macOS builds (no problems on arm64 so far) is IMO an option on the table, to be on the safer side.

I'm also wondering if there were some breaking macOS-only changes in recent LLVM versions, making mixing bitcode objects of different clang/LDC host compiler LLVM versions problematic.

kinke:
Alright, with LDC v1.35.0 and restored LTO, mimalloc can be restored too without breaking `config_diag.d`. So matching the host/bootstrap LDC LLVM version with the Xcode LLVM version seems to behave as expected - v1.35.0 with its LLVM 16 is the only version that works with and without (full) LTO, and with/without mimalloc. We already matched the host compiler LLVM versions on macOS arm64 (but can use LDC v1.39.0 with macOS-specific LLVM 17 there, compatible with the `macos-15` Xcode 16).

kinke:
@the-horo: What do you think about the 3rd commit here ('somewhat simplify driver/configfile.d')? I've worked on that before getting an idea of where you wanna go with that config directory. In the case of multiple .conf files and multiple consecutive `ConfigFile::readConfig()` calls, your original code makes sense, looking up merged `ArraySetting` instances and checking whether the current file only appends flags without overwriting them. With the existing single config file support, that's not needed though.

the-horo:
It seems alright. On its own the changes look good and I like what you did. Sure, it would require me to do some more changes when (if) the directory option is merged but I don't think they would be so many that your few changes here really affect it. I think discussing about it would take more time then implementing it. Passing the array as a `ref` parameter to `findArraySetting` sounds like it would be enough to support it.

When I first wrote my changes I actually wanted to combine the two array merging code paths (the one in `readArraySetting` and in `applyArray`) but I didn't come up with it anything. Your changes do what I wanted so no complaining here ;)

PR #4862, opened 2025-03-03
CI: Bump Alpine image to v3.21

kinke:
Incl. enabling C++ LTO for the compiler, as well as PGO (although I'm not sure PGO really works; >1k warnings wrt. changed function control flows... **edit**: oh well, looks like those are for the `ldc2-unittest` object file only).

### DIFF
--- a/tests/dmd/fail_compilation/needspkgmod.d
+++ b/tests/dmd/fail_compilation/needspkgmod.d
@@ -9,7 +9,7 @@
 /*
 TEST_OUTPUT:
 ----
-$r:.+_D7imports9pkgmod3133mod3barFZv.*$
+$r:.+(_D7imports9pkgmod3133mod3barFZv|imports\.pkgmod313\.mod\.bar).*$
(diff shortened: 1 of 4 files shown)


tests/dmd/fail_compilation/needspkgmod.d
```diff
@@ -9,7 +9,7 @@
 /*
 TEST_OUTPUT:
 ----
-$r:.+_D7imports9pkgmod3133mod3barFZv.*$
+$r:.+(_D7imports9pkgmod3133mod3barFZv|imports\.pkgmod313\.mod\.bar).*$
```
the-horo, line 12: Note that I fixed this upstream differently: https://github.com/dlang/dmd/pull/16833/files

kinke: Oh wow - I'm having a hard time understanding why this works on Windows, `link.exe` surely has no `--no-demangle` option.

PR #4864, opened 2025-03-05
runtime: Link libatomic on 64-bit RISC-V and LoongArch

### LINKED ISSUE #4847: Linker errors with `__atomic_*` symbols on LoongArch64/RISC-V due to missing `-latomic` for non-inlined atomic operations
When building LDC 1.40.0 on Arch Linux for Loong64 (LoongArch 64bit) and archriscv (RISC-V 64bit), the test suite fails during linking `runtime/druntime-test-runner-shared` `runtime/druntime-test-runner` and `runtime/druntime-test-runner-debug` with multiple undefined references to atomic operations:

```
/usr/bin/ld: undefined reference to `__atomic_compare_exchange_16'
/usr/bin/ld: undefined reference to `__atomic_store_16'
/usr/bin/ld: undefined reference to `__atomic_load_16'
```

## Analysis
The errors occur during linking of the druntime test runner. These atomic operations require linking with `-latomic` on these architectures. The issue appears to be architecture-specific, as loong64 and ricv64‘s compiler may not be able to inline all atomic operations supported by x86_64, falling back to the implementation in `libatomic`.

E.g. LoongArch64's instructions only support 32/64-bit atomic operations while lacking native 128-bit (16 byte, the error above) atomic support.

## Workaround
Adding `-DLD_FLAGS="-Wl,--no-as-needed -latomic -Wl,--as-needed"` to the CMake command resolves the issue. This forces the linker to include libatomic when needed.

## Possible Solution
I think above may need to be solved in ldc instead of downstream linux distros. The build configuration here might not depend on architecture-specific features or low level toolchain implementations. May we use the `-latomic` linker parameter by default to build this test?

## Detailed Failed Log of Check Process
[ldc-3:1.40.0-2-loong64-check.log](https://pastebin.com/74ETuvG3)

kinke:
Fixes #4847.


runtime/CMakeLists.txt
```diff
@@ -141,6 +141,10 @@ if("${C_SYSTEM_LIBS}" STREQUAL "AUTO")
         else()
             set(C_SYSTEM_LIBS m pthread)
         endif()
+        # some archs need libatomic for 128-bit CAS druntime unittests
+        if(CMAKE_SYSTEM_PROCESSOR MATCHES "riscv64|loongarch64")
+            set(C_SYSTEM_LIBS atomic S{C_SYSTEM_LIBS})
```
wszqkzqk, line 146: Typo: `S{C_SYSTEM_LIBS}` -> `${C_SYSTEM_LIBS}`

kinke: No, this is to *pre*pend libatomic to the existing `C_SYSTEM_LIBS` libs. Just in case libatomic depends on some of the other libs (which definitely don't depend on libatomic).

wszqkzqk: But it caused compile error:

```log
/usr/bin/ld: cannot find -lS{C_SYSTEM_LIBS}: No such file or directory

```

kinke: Oh sorry, saw the character diff just now...

denizzzka:
I had a similar problem on RISC-V, Ubuntu 24.10 :
```
FAILED: runtime/druntime-test-runner 
: && /usr/bin/cc -DHAVE_UNISTD_H -O3 -DNDEBUG  runtime/objects-unittest/test_runner.o -o runtime/druntime-test-runner  -Wl,--whole-archive,/mnt/shared/ldc_whole_riscv/lib/libdruntime-ldc-unittest.a,--no-whole-archive  -lm  -lpthread  -lrt  -ldl && :
/usr/bin/ld: /mnt/shared/ldc_whole_riscv/lib/libdruntime-ldc-unittest.a(atomic.o): in function `core.atomic.__unittest_L984_C24()':
atomic.d:(.text._D4core6atomic19__unittest_L984_C24FNaNbNfZv+0x4ae): undefined reference to `__atomic_compare_exchange_16'
/usr/bin/ld: atomic.d:(.text._D4core6atomic19__unittest_L984_C24FNaNbNfZv+0x4e0): undefined reference to `__atomic_compare_exchange_16'
/usr/bin/ld: atomic.d:(.text._D4core6atomic19__unittest_L984_C24FNaNbNfZv+0x516): undefined reference to `__atomic_compare_exchange_16'
/usr/bin/ld: atomic.d:(.text._D4core6atomic19__unittest_L984_C24FNaNbNfZv+0x562): undefined reference to `__atomic_compare_exchange_16'
```

This patch fixes it. Thanks!

PR #4867, opened 2025-03-06
Adds native support for Fiber context switching on RISC-V

denizzzka:
This code is based on mine [earlier implementation](https://github.com/denizzzka/dfruntime/blob/dfruntime/runtime/druntime/src/core/thread/fiber/switch_context_riscv.S) for `rv32`, so both platforms should be supported: `rv32` and `rv64` with and without compressed instructions and with and without `HardFloat`.

For now it is tested only on `rv64` with `HardFloat` (QEMU, Ubuntu 24.10, `ldc2` from `~master`)

Implemented using separate file because I see potential for growth of this assembly code (I want to implement embedded ISA subset and maybe something vendor-specific), and also to put an end to ~~vicious~~ scaring me  practice of putting assembly code for completely different platforms into one huge .S file

Path to the new asm file is the same as in the current DMD `~master` for `core.thread.fiber` package

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -231,7 +231,11 @@ endif()
 # druntime ASM parts
 set(DRUNTIME_ASM)
 if("${TARGET_SYSTEM}" MATCHES "UNIX")
-    list(APPEND DRUNTIME_ASM ${RUNTIME_DIR}/src/core/thread/fiber/switch_context_asm.S ${RUNTIME_DIR}/src/ldc/eh_asm.S)
+    list(APPEND DRUNTIME_ASM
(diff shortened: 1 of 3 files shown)


denizzzka:
@wszqkzqk Hi! Maybe you are also interested in this?

kinke (approved):

kinke:
For posterity: I've 'pushed' for this upstream in https://github.com/dlang/dmd/pull/20964, but turns out GDC doesn't use that upstream `core/thread/fiber/switch_context_asm.S` file anymore, but has its own set of asm files apparently. And the .S file only handles archs that DMD cannot target, so it isn't really used by DMD either. So this separate file here for RISC-V and LDC alone seems okay to me for now.

So thx Dennis for bearing with me.

PR #4868, opened 2025-03-06
Merge upstream v2.110.0

kinke:


kinke:
Still waiting for https://code.dlang.org/packages/dub to get updated to v1.39.0, so that reggae can use it (https://github.com/atilaneves/reggae/pull/315). @s-ludwig: That'd be nice. :)

WebFreak001:
updated

PR #4869, opened 2025-03-08
CI: adds riscv64 builds

denizzzka:
Current druntime `~master` [has broken](https://github.com/dlang/dmd/pull/16695#issuecomment-2358432942) for platforms that use Posix context switching. it seems, for LDC such platform is only riscv64.

Maybe this is a good reason to add CI for that platform

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -93,6 +93,21 @@ jobs:
               -DEXTRA_CXXFLAGS=-flto=full
             with_pgo: true
 
+          - job_name: Linux riscv64
+            os: ubuntu-20.04-riscv64
+            container_image: ubuntu:20.04
+            arch: riscv64
+            base_cmake_flags: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -93,6 +93,21 @@ jobs:
               -DEXTRA_CXXFLAGS=-flto=full
             with_pgo: true
 
+          - job_name: Linux riscv64
+            os: ubuntu-20.04-riscv64
```
kinke, line 97: If only it was that easy! :) - The only available runners are https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories.

denizzzka: Yeah, it's a shame it's like that

denizzzka:
Somehow I don't want to go through the QEMU

Ok, let's put it off until another time

PR #4876, opened 2025-03-13
[WIP] Upgrade frontend & libs to v2.111

kinke:


kinke:
Switching to an upstream branch, for simpler potential collaboration on remaining points: https://github.com/ldc-developers/ldc/pull/4877

Issue #4374, opened 2023-04-15
ICE: overlapping initializers for struct literal


dkorpel:
I ran into this when re-ordering fields in a union:
```D
union U
{
    int x;
    int[1] arr;

    U setAll(int x)
    {
        arr[0] = x;
        return this;
    }

    enum zero = U.init.setAll(0);
}

void main()
{
    auto u = U.zero;
}
```

```
app.d(17): Error: ICE: overlapping initializers for struct literal
```

It works with dmd, and when the array comes first in the union, it also works with ldc.

kinke:
A frontend issue IMO (IIRC, I've simplified the LDC implementation in this regard not too long ago when implementing the bitfields) - it really sets both initializers, for absolutely no good reason AFAICT. Here's the `-vcg-ast` output for 3 such enums:
```
enum zero = U.init.setAll(0); // U(0, [0])
enum U foo = {x: 0};          // U(0, );
enum U bla = {arr: [0]};      // U(, [0])
```

kinke:
Oh right, I think I've simplified it (this is a v1.31 regression) because the user doing this explicitly has been an error for a while:
```D
enum U bla = {x: 0, arr: 0}; // Error: overlapping initialization for field `x` and `arr`
```

kinke:
`U.init.setAll(1)` is lowered to a `U(0, [1])`... and using `arr[] = x;` in the ctor, instead of setting the single element explicitly, makes the problematic `x` initializer vanish.

kinke:
There seems to be a related regression in v2.111.0, surfacing when compiling the `std.json` unittests in #4877 - for the `std.json.JsonValue.emptyObject` literal. With turned-on `-vv` output, we see the nonsensical overlapping initializers for the `union Store` field initializer:
```
StructLiteralExp::toElem: JSONValue(Store(null, , , , Object(false, null, ), ), JSONType.object) @ JSONValue
```

The 1st `null` initializer for the `string Store.str` field is wrong; we just need the `Object(…)` literal (which is even larger in size, and contains a further nested union, with a single initializer luckily...).

kinke:
There's an according upstream issue: https://github.com/dlang/dmd/issues/20675

PR #4877, opened 2025-03-14
Upgrade frontend & libs to v2.111

kinke:


kinke:
Okay, this isn't in a too bad state, especially considering the many accumulated changes for v2.111; most tests are passing, only about 10 failures remaining. We *could* work on the rest in parallel, by creating PRs against this `merge-2.111` branch.

TODO:
* I haven't resolved most merge conflicts for the druntime integration test Makefiles yet, which have been significantly reworked upstream:
  ```
  $ git grep "<<<< HEAD"
  runtime/druntime/test/common.mak:<<<<<<< HEAD
  runtime/druntime/test/shared/Makefile:<<<<<<< HEAD
  runtime/druntime/test/shared/Makefile:<<<<<<< HEAD
  runtime/druntime/test/shared/Makefile:<<<<<<< HEAD
  runtime/druntime/test/stdcpp/Makefile:<<<<<<< HEAD
  ```
  @the-horo: Maybe you wanna have a stab at it, or perhaps even did some LDC porting already? You know everything about the new system. I hope most LDC-specifics in these files are guarded by checking whether `DMD` contains `ldmd2`, or at least provide comments wrt. what was added/changed.
* timetrace - some things were upstreamed and now 'conflict' (yeah, even a couple of warnings wrt. duplicate C++ symbol definitions...). Maybe @JohanEngelen or @thewilsonator wanna consolidate these?
* apparent CTFE regression wrt. overlapping initializers for unions: https://github.com/ldc-developers/ldc/issues/4374#issuecomment-2723072893. I'll try to fix that in the DMD frontend, don't wanna have to re-deal with that crap in LDC codegen.
* apparent frontend regression wrt. lvalue-ness of indexed *cast* AA (lit-test `compilable/gh3162.d`)
* adapt LDC codegen to removed `_d_newThrowable` (`runnable/test19317.d`)
* add support for placement new (`runnable/placenew.d`)
* reggae can't be compiled with v2.111, apparently due to a Phobos `std.getopt` regression, wrt. `const(string)[]` IIRC (=> `const` is the problem)
* probably have to minimally adapt a unittest assertion in `std.experimental.allocator.building_blocks.allocator_list` for macOS arm64
* probably have to disable new `runnable/exe1.c`; seems to test edge cases for *compiling* C only (former DMC backend tests or something...)

the-horo:
> @the-horo: Maybe you wanna have a stab at it, or perhaps even did some LDC porting already? You know everything about the new system. I hope most LDC-specifics in these files are guarded by checking whether DMD contains ldmd2, or at least provide comments wrt. what was added/changed.

Sure. When I rewrote the makefiles upstream I did think *somewhat* of ldc but I didn't test anything. I'm a bit busy atm but I'll see what I can do

the-horo:
> * timetrace - some things were upstreamed and now 'conflict' (yeah, even a couple of warnings wrt. duplicate C++ symbol definitions...). Maybe @JohanEngelen or @thewilsonator wanna consolidate these?

https://github.com/ldc-developers/ldc/pull/4880 (those warning are errors on my system)

the-horo:
> I haven't resolved most merge conflicts for the druntime integration test Makefiles yet, which have been significantly reworked upstream:
> 
> ```
> $ git grep "<<<< HEAD"
> runtime/druntime/test/common.mak:<<<<<<< HEAD
> runtime/druntime/test/shared/Makefile:<<<<<<< HEAD
> runtime/druntime/test/shared/Makefile:<<<<<<< HEAD
> runtime/druntime/test/shared/Makefile:<<<<<<< HEAD
> runtime/druntime/test/stdcpp/Makefile:<<<<<<< HEAD
> ```
> 
> @the-horo: Maybe you wanna have a stab at it, or perhaps even did some LDC porting already? You know everything about the new system. I hope most LDC-specifics in these files are guarded by checking whether `DMD` contains `ldmd2`, or at least provide comments wrt. what was added/changed.

https://github.com/the-horo/ldc/tree/refs/heads/pr-2.111. It worked fine for my one test locally on linux but the other environments might still have issues.

kinke:
Thanks a lot Andrei!

the-horo:
Alright, some progress has been made. The tests are running but some of them are failing. `importc_compare` fails on every platform but there are some other failures as well.

the-horo:
My action plan right now is:
1. isolate and fix the newly failing tests
2. go through the logs and check that no tests started being accidentally skipped
3. run the tests linking druntime both statically and dynamically when `-DBUILD_SHARED_LIBS=BOTH`

kinke:
I suggest you open a PR with that branch; we can remove your 1st commit from https://github.com/ldc-developers/ldc/pull/4880 once the timetrace mess has been dealt with (too bad you're getting errors instead of warnings, but oh well). Then we can discuss the druntime integration tests stuff there. And please don't worry too much about remaining failures; once I see them in the CI logs, I can help out as well. Resolving the merge conflicts was already a great help.

> 3. run the tests linking druntime both statically and dynamically when -DBUILD_SHARED_LIBS=BOTH

Oh, cool stuff! But please keep that separate, in another PR.

kinke:
FYI, I'll look into the timetrace stuff now; just to prevent duplicate efforts.

runtime/druntime/test/exceptions/src/assert_fail.d
```diff
@@ -539,21 +539,19 @@ void testDestruction()
 
     version (LDC)
     {
-        version (D_Optimized) {} else
+        version (D_Optimized)  enum allowFailure = true;
```
the-horo, line 542: I got this error as well and it's pretty much affecting all lot of the tests by inlining/removing functions which mess up the hardcoded line information. In this particular case it is failing because it optimizes away the `clobberStack` function which leaves those pointers on the stack.

I think it would be better if we did, in the makefile:
```make
override DFLAGS += -O0
```

kinke: Oh, thx for the investigation! Well in that case we should probably just try to avoid that `clobberStack` call getting optimized away, which we can selectively via attributes, without bluntly disabling all optimizations for this subtest.

the-horo: Well, like I said, I have 1 other test fail bc a function call got optimized away leading to a different stacktrace, but maybe there won't be many more compared to these 2

kinke: [I hadn't seen these other failures locally for `exceptions` because OS wasn't set.]

runtime/druntime/test/exceptions/Makefile
```diff
@@ -106,7 +106,9 @@ $(ROOT)/static_dtor.done: private negate=!
 $(ROOT)/future_message.done: stderr_exp="exception I have a custom message. exception exception "
 $(ROOT)/catch_in_finally.done: stderr_exp="success."
 $(ROOT)/rt_trap_exceptions.done: stderr_exp="object.Exception@src/rt_trap_exceptions.d(12): this will abort"
+ifneq ($(BUILD)-$(IN_LDC),release-1) # disable 2nd expression for LDC in BUILD=release mode
 $(ROOT)/rt_trap_exceptions.done: stderr_exp2="src/rt_trap_exceptions.d:8 main"
+endif
```
the-horo, line 111: I don't think conditioning the test on `BUILD` is a wise idea. If a user passes `-DD_FLAGS="-O -release -mcpu=native"` then this test would start failing even with `BUILD=debug`. This is why is recommended overwriting the flags by appending `-O0`

kinke: I'd say let's not over-engineer these Makefiles, supporting everything that a user might specify, especially an explicit `-O`, with our CMake script creating pairs of debug/release versions.

FWIW, I've hackily tried adding `-O0` during the tests, to `extra_dflags`, but that came too late, was overridden by `-O` from BUILD=release later.

kinke: FWIW, I don't really understand why this 2nd expression works in debug mode in the first place - that line calls `_d_run_main()` from druntime.

the-horo: > FWIW, I've hackily tried adding -O0 during local tests, to extra_dflags, but that came too late, was overridden by -O from BUILD=release later.

Yup, the user `DFLAGS` are meant to apply after project-specific flags, this is true with pretty much all build systems. That's why I said to override `DFLAGS` rather then `extra_dflags`.

kinke: See the latest force push, this is now handled by preventing the tail call via UDA. This is now finally enough for all backtrace tests on glibc Linux x86_64; musl, Mac and BSDs apparently need some more (maybe frame pointers).

the-horo: Still unsure if this is a better solution that passing `-O0` but so long as flags passed by the user don't make the tests fail I'm happy

kinke:
Looks like there's a single remaining failure - lit-test `sanitizers/asan_fiber.d` with `RT_SUPPORT_SANITIZERS=ON`. `core.thread.fiber` has been split up into a package; I've checked the (messy) LDC-specific diff before and after, and couldn't spot an obvious problem.

kinke:
Thx for checking! Yeah saw that diff too and thought this was just an old workaround (for a compiler warning); looks like it should be restored, but I'm almost certain it won't fix the problem at hand.

kinke:
> lit-test sanitizers/asan_fiber.d with RT_SUPPORT_SANITIZERS=ON

Little correction: on macOS arm64 at least; Linux x86_64 is fine (tested by 2 CI jobs). So might be Darwin and/or AArch64 specific.

kinke:
@JohanEngelen: Do you think you'll find some time to have a look at the remaining issue? Otherwise I'll disable that test for now, to move on.

PR #4880, opened 2025-03-17
dmd/timetrace.d: Add namespace to C++ functions

the-horo:
Add namespacing to the dmd functions to avoid symbol conflicts between them and the similar functions from driver/timetrace.d.

### DIFF
--- a/dmd/timetrace.d
+++ b/dmd/timetrace.d
@@ -34,7 +34,7 @@ TimeTraceProfiler* timeTraceProfiler = null;
  *   timeGranularityUs = minimum event size in microseconds
  *   processName = name of this executable
  */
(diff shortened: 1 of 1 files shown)


kinke:
This is most likely just a bunch of copy&paste during upstreaming, so the conflicting functions most likely being totally identically, or almost identical. It just needs to be cleaned up. Putting the new upstream free functions in a `dmd` C++ namespace would make sense, but shouldn't be required.

kinke:
I've added these in https://github.com/ldc-developers/ldc/pull/4884, before realizing that this doesn't work upstream with their oldest-supported bootstrap compilers: https://github.com/dlang/dmd/pull/21034

PR #4883, opened 2025-03-17
Druntime test merge fix

the-horo:
Fix the druntime tests merge conflicts as well as resolve some issues with the test code. Some changes are specific to LDC and should remain here, some are bugs that should be fixed upstream but currently they're all in separate commits and will be split and squashed later.


.github/actions/1-setup/action.yml
```diff
@@ -155,6 +155,13 @@ runs:
       if: runner.os == 'Windows'
       shell: cmd
       run: make --version
+    - name: 'macOS: Install a more recent GNU make'
+      if: runner.os == 'macOS'
+      shell: bash
+      run: |
+        brew install make
+        echo "PATH=/usr/local/opt/make/libexec/gnubin:$PATH" >> $GITHUB_ENV
+        make --version
```
kinke, line 164: Won't work on the M1 runners (at least last time I checked): https://github.com/dlang/dmd/pull/17035/files#r1997643588

runtime/druntime/test/common.mak
```diff
@@ -1,5 +1,5 @@
 # set explicitly in the make cmdline in druntime/Makefile (`test/%/.run` rule):
-ifneq (,$(findstring ldmd2,$(DMD)))
+ifdef IN_LDC
```
kinke, line 2: Mm, much better, I like! :+1:

.github/actions/1-setup/action.yml
the-horo: I was sure that that PATH was going to break eventually. I suppose we don't need the PATH modifications since we're checking for `gmake` first in https://github.com/ldc-developers/ldc/blob/dddcd98b358f3f7eebb993014c90e476486ce6f9/runtime/DRuntimeIntegrationTests.cmake#L4

kinke: Right, the `macOS arm64` job seems happy as-is, so we can remove it. :+1:

runtime/DRuntimeIntegrationTests.cmake
```diff
@@ -82,12 +76,9 @@ foreach(name ${testnames})
                 ROOT=${outdir} DMD=${LDMD_EXE_FULL} BUILD=${build} SHARED=1
                 CC=${CMAKE_C_COMPILER} CXX=${CMAKE_CXX_COMPILER}
                 DRUNTIME=${druntime_path_build} DRUNTIMESO=${shared_druntime_path_build}
-                CFLAGS_BASE=${cflags_base} DFLAGS_BASE=${dflags_base} ${linkdl}
+                CFLAGS=${cflags_base} CXXFLAGS=${cflags_base} DFLAGS=${dflags_base} ${linkdl}
```
kinke, line 79: I wonder if we can really override `DFLAGS` with the (optional) base flags - don't we lose the implicit `[-g -debug | -O -release] -w` as default DFLAGS this way?

runtime/druntime/test/common.mak
```diff
@@ -186,9 +157,17 @@ extra_ldlibs.d += -L$(druntime_for_linking)
 
 extra_ldflags.d += -defaultlib=
 extra_ldlibs.d += $(d_platform_libs)
+else # IN_LDC
+    extra_ldflags.d += -defaultlib=druntime-ldc
+    extra_ldflags.d += $(if $(LINK_SHARED),-link-defaultlib-shared)
```
kinke, line 162: `-link-defaultlib-shared` isn't a pure link-only flag, but also affects some compile flag defaults for Windows target (e.g., defaulting to `-dllimport=defaultLibsOnly` when not using `-fvisibility`...). So needs to be specified when compiling too (`extra_dflags`).

runtime/druntime/test/profile/Makefile
```diff
@@ -30,8 +33,7 @@ $(ROOT)/profile.done: $(ROOT)/%.done: $(ROOT)/%$(DOTEXE)
 	sort $(ROOT)/mytrace.def -o $(ROOT)/mytrace.def
 	(sort mytrace.def.exp | $(DIFF) - $(ROOT)/mytrace.def) || (sort mytrace.releaseopt.def.exp | $(DIFF) - $(ROOT)/mytrace.def)
 	@touch $@
-# LDC: enable assertions for BUILD=RELEASE (=> `-O -release`)
-$(ROOT)/profile$(DOTEXE): extra_dflags += -profile $(if $(findstring ldmd2,$(DMD)),-check=assert=on,)
```
kinke, line 34: I don't see this `-check=assert=on` re-added.

I don't recall whether this was actually a problem, or whether the problem was that some test failure wasn't caught because all asserts are disabled in `BUILD=RELEASE` mode, which implies `-release`, disabling all asserts.

IIRC, some tests use manual assert helpers to make sure the tests actually test anything in BUILD=RELEASE mode. I guess it would just make sense to gt rid of the `-release`, upstream.

runtime/druntime/test/shared/Makefile
```diff
@@ -30,21 +34,16 @@ endif
...
 for_linking := $(if $(filter windows,$(OS)),.lib,$(DOTDLL))
->>>>>>> origin/dmd-rewrite-stable
 
 ifeq (windows,$(OS)) # extra tests on Windows
 
-ifneq (,$(findstring ldmd2,$(DMD)))
+ifdef IN_LDC
     # LDC: required for executables to implicitly dllimport DLL data symbols
-    DFLAGS+=-dllimport=all
+    extra_ldflags.d += -dllimport=all
```
kinke, line 46: This is pure compile flag only, doesn't affect linking at all.

kinke: Btw for DMD this is done in `common.mak:153`. I guess it doesn't matter where it's defined, this `shared/Makefile` is IIRC the only one setting `LINK_SHARED`.

runtime/druntime/test/shared/Makefile
```diff
@@ -55,38 +54,23 @@ PATH := $(dir $(DRUNTIMESO));$(PATH)
...
-
 # LDC: this test is designed for .exe & .dll with separate druntimes
-ifneq (,$(findstring ldmd2,$(DMD)))
-$(ROOT)/dllgc$(DOTEXE): DFLAGS+=-link-defaultlib-shared=false -dllimport=none
-endif
-
-$(ROOT)/dllgc$(DOTEXE): $(SRC)/dllgc.d
-	$(QUIET)$(DMD) $(DFLAGS) -version=DLL -shared -of$(ROOT)/dllgc$(DOTDLL) $<
-	$(QUIET)$(DMD) $(DFLAGS) -of$@ $<
+$(ROOT)/dllgc$(DOTEXE): private extra_ldflags.d += -link-defaultlib-shared=false -dllimport=none
```
kinke, line 58: required for compilation too, not just linking

runtime/druntime/test/shared/Makefile
```diff
@@ -55,38 +54,23 @@ PATH := $(dir $(DRUNTIMESO));$(PATH)
...
-	$(QUIET)$(DMD) $(DFLAGS) -shared -L/EXPORT:gc_setProxy -L/EXPORT:gc_clrProxy -version=DLL -of$(ROOT)/dll_gc_proxy_teardown$(DOTDLL) $<
-	$(QUIET)$(DMD) $(DFLAGS) -shared -L/EXPORT:gc_setProxy -L/EXPORT:gc_clrProxy -version=DLL -version=NoUnload -of$(ROOT)/dll_gc_proxy_teardown_nounload$(DOTDLL) $<
-	$(QUIET)$(DMD) $(DFLAGS) -of$(ROOT)/load_dll_gc_proxy_teardown$(DOTEXE) $<
-	$(QUIET)$(DMD) $(DFLAGS) -version=NoUnload -of$(ROOT)/load_dll_gc_proxy_teardown_nounload$(DOTEXE) $<
-	$(QUIET)$(ROOT)/load_dll_gc_proxy_teardown$(DOTEXE)
-	$(QUIET)$(ROOT)/load_dll_gc_proxy_teardown_nounload$(DOTEXE)
-endif
+$(OBJDIR)/dll_gc_proxy_teardown%$(DOTDLL): private extra_ldflags.d += -L/EXPORT:gc_setProxy -L/EXPORT:gc_clrProxy
+$(OBJDIR)/dll_gc_proxy_teardown%$(DOTDLL): private extra_dflags += -version=DLL
+$(OBJDIR)/dll_gc_proxy_teardown%$(DOTEXE): private extra_ldflags.d += -link-defaultlib-shared=false -dllimport=none
```
kinke, line 63: dflags/ldflags need tweaking again

runtime/druntime/test/profile/Makefile
kinke: Ah great, I see you added this now for all tests - via `-check=assert`, that's why I haven't found it earlier.

runtime/DRuntimeIntegrationTests.cmake
the-horo: This is intended behavior. If one of those flags (notably `-g`) is required by the tests then that should be specified in the makefile, outside of `DFLAGS`. If we want to test multiple configurations (like `LINK_SHARED=1` or `DFLAGS=-m32` or even `DFLAGS="-O -release"`) then these flags should be specified explicitly.

If we want to keep testing both sets of flags we could reintroduce the `DFLAGS_BASE` variables to preserve the defaults from the makefile or we could pass `D_EXTRA_FLAGS` in `DMD` which would be appropriate if that cmake variable is meant for `-m32` or `-mtriple` like arguments.

runtime/druntime/test/shared/Makefile
the-horo: If you remember I do want to also start testing with the shared druntime when building both static&dynamic runtime libs so I want to support that flag being passed on the CLI so I'd rather have the logic for it in `common.mak`

the-horo:
> It might be worth reconsidering splitting the D flags into compile- and link-flags - there's no harm in including link-flags in compiles too, just a bit longer/more explicit command lines in the make output (normally hidden).

`DFLAGS` and `extra_dflags` do apply to both compilation and linking, on top of most of the tests doing compilation+linking in one compiler invocation (`stdcpp` is the only one that does manual compilation because of the need to link the C++ stdlib).

I would rather not change this design because I purposefully made it follow the way other build systems do it, and if they do it this way I assume that there is a reason for it. So, until I fully understand this reason I would rather go with the possibly sub-optimal design.

runtime/DRuntimeIntegrationTests.cmake
the-horo: We can also start passing `LDFLAGS=${LD_FLAGS}` but, if those flags are only meant for C sources, we would also need `LDFLAGS.d=""` since `common.mak` autofills the latter based on the former. There's also `d_platform_libs` which is used by `stdcpp`, maybe we could set that to `C_SYSTEM_LIBS`?

runtime/druntime/test/common.mak
the-horo: The `dmd` branch adds `-dllimport` to `ldflags`, should that be changed as well?

runtime/druntime/test/gc/invariant.d
```diff
@@ -0,0 +1,3 @@
+// ldc fails to link a file named invariant.d
+// https://github.com/ldc-developers/ldc/issues/4879
+module ldc_rename_me;
```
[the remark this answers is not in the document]
the-horo: I usually chose to rebase since I have do it eventually and it doesn't take me that long to drop a commit.

the-horo:
I've marked which commits should be upstreamed.

The failure which I gathered are:
- windows `druntime-test-exceptions-debug` > `winstack`
- windows `druntime-test-shared` > `loadDR`
- alpine `druntime-test-shared` > `loadDR`
- macos `druntime-test-shared` > `load_linkdep`
- most platforms `importc_compare`

----
The `loadDR` test on windows logs:
```
C:/Program Files/LLVM/bin/clang-cl.exe  /FoD:/a/ldc/build/runtime/druntime-test-shared-debug/       src/loadDR.c    /FeD:/a/ldc/build/runtime/druntime-test-shared-debug/loadDR.exe
Testing loadDR
D:/a/ldc/build/runtime/druntime-test-shared-debug/loadDR.exe D:/a/ldc/build/lib/druntime-ldc-debug-shared.lib
Assertion failed: h != NULL, file src/loadDR.c, line 11
```
- ```
  D:/a/ldc/build/runtime/druntime-test-shared-debug/loadDR.exe D:/a/ldc/build/lib/druntime-ldc-debug-shared.lib
  ```

Shouldn't the library path on windows by the `.dll` instead of the `.lib`?

the-horo:
Yes! The old makefile changed the suffix from `.lib` to `.dll`: https://github.com/ldc-developers/ldc/blob/61c11b7dafa24d9c67011d03a0b63d3287791c10/runtime/druntime/test/shared/Makefile#L82
But I changed that as I assumed the `DRUNTIMESO` would be the dll: https://github.com/ldc-developers/ldc/blob/e60ead6a20c691f9aa57232e8596e8553f92d009/runtime/druntime/test/shared/Makefile#L149
 which it is within dmd but not within ldc.

runtime/DRuntimeIntegrationTests.cmake
kinke: > If we want to keep testing both sets of flags we could reintroduce the DFLAGS_BASE variables to preserve the defaults from the makefile or we could pass D_EXTRA_FLAGS in DMD which would be appropriate if that cmake variable is meant for -m32 or -mtriple like arguments.

We run every test in both debug/release variants, only differing in `BUILD`, which only affects temp output dir and these D/C/CXX flags (and `BUILD=debug` enabling one extra subtest on Windows - `winstack` for exceptions). So if we specify/override the flags anyway in the make cmdline, we don't test both debug/release variants, but basically run the tests twice with identical config, except for the output dir - neither an optimized nor a debug variant.

The aim of these `{D,C}FLAGS_BASE` was to propagate the optional extra args specified by the user in the CMake invocation (`D_EXTRA_FLAGS` and `RT_CFLAGS`). These can be anything really, but setting up cross-compilation is probably one of the more common usages.

So I don't really care whether those are propagated in `DMD`/ `CC`/`CXX` or in the former extra `…_BASE` variables - but what's important is that we run each test in proper debug/release configs.

runtime/druntime/test/common.mak
kinke: Yes.

runtime/druntime/test/shared/Makefile
kinke: Absolutely fine by me.

kinke:
> DFLAGS and extra_dflags do apply to both compilation and linking

Ah okay, that is fine then.

kinke:
> I changed that as I assumed the DRUNTIMESO would be the dll

IIRC, upstream changed that not too long ago. I went with DRUNTIMESO as import .lib because that's needed much more often than the DLL (so no need for variable `druntime_for_linking`). Anyway, switching to the DLL should be a matter of using `TARGET_FILE` here: https://github.com/ldc-developers/ldc/blob/61c11b7dafa24d9c67011d03a0b63d3287791c10/runtime/DRuntimeIntegrationTests.cmake#L31

runtime/DRuntimeIntegrationTests.cmake
the-horo: > and BUILD=debug enabling one extra subtest on Windows

I though I got rid of `BUILD`-condition tests, seems not.

> which only affects temp output dir and these D/C/CXX flags

I thought the BUILD argument was most importantly for picking which druntime variant is used, so passing `-link-defaultlib-debug`. Upstream, this flag does only modifies the compilation flags, but it also changes the output dir like you said which changes the druntime library that is used in tests, leading to an equivalent functionality to passing `-link-defaultlib-debug`.

> So I don't really care whether those are propagated in DMD/ CC/CXX or in the former extra …_BASE variables - but what's important is that we run each test in proper debug/release configs.

So, do you specifically want compiling the test executables with `-O -realease` / `-g` or is it enough to test the two different druntime builds?

the-horo:
I've looked a little bit more:

* [X] windows `druntime-test-exceptions-debug` > `winstack`
   Fixed by passing -g


* [X] windows `druntime-test-shared` > `loadDR`
Works correctly now that the `.dll` is passed on the command line

* [ ] alpine `druntime-test-shared` > `loadDR`
Before:
```
1808: /usr/bin/cc -Wall -Wl,-rpath,/__w/ldc/build/lib -g -o/__w/ldc/build/runtime/druntime-test-shared-debug/loadDR src/loadDR.c -ldl  -pthread
1808: Testing loadDR
1808: /__w/ldc/build/runtime/druntime-test-shared-debug/loadDR /__w/ldc/build/lib/libdruntime-ldc-debug-shared.so
```
Now:
```
1840: /usr/bin/clang        src/loadDR.c  -ldl  -o /__w/ldc/build/runtime/druntime-test-shared-debug/loadDR
1840: Testing loadDR
1840: /__w/ldc/build/runtime/druntime-test-shared-debug/loadDR /__w/ldc/build/lib/libdruntime-ldc-debug-shared.so.111.0
1840: Aborting from rt/sections_elf_shared.d(966) Failed to get library handle.gmake: *** [../common.mak:182: /__w/ldc/build/runtime/druntime-test-shared-debug/loadDR.done] Aborted (core dumped)
```

* [ ] macos `druntime-test-shared` > `load_linkdep`
This test should be skipped, using `OS` before including `common.mak` leads to it being run
https://github.com/ldc-developers/ldc/blob/8069d9a695558a5e80bdb84c14e06a01ab437c6e/runtime/druntime/test/shared/Makefile#L16-L22

A new test started failing:
* [ ] windows `druntime-test-shared` > `dll_gc_proxy_teardown`
It seems to be caused by flags not being properly passed.

runtime/druntime/test/shared/Makefile
```diff
@@ -55,38 +59,24 @@ PATH := $(dir $(DRUNTIMESO));$(PATH)
...
-	$(QUIET)$(DMD) $(DFLAGS) -shared -L/EXPORT:gc_setProxy -L/EXPORT:gc_clrProxy -version=DLL -version=NoUnload -of$(ROOT)/dll_gc_proxy_teardown_nounload$(DOTDLL) $<
-	$(QUIET)$(DMD) $(DFLAGS) -of$(ROOT)/load_dll_gc_proxy_teardown$(DOTEXE) $<
-	$(QUIET)$(DMD) $(DFLAGS) -version=NoUnload -of$(ROOT)/load_dll_gc_proxy_teardown_nounload$(DOTEXE) $<
-	$(QUIET)$(ROOT)/load_dll_gc_proxy_teardown$(DOTEXE)
-	$(QUIET)$(ROOT)/load_dll_gc_proxy_teardown_nounload$(DOTEXE)
-endif
+dgpt = dll_gc_proxy_teardown dll_gc_proxy_teardown_nounload
+$(dgpt:%=$(OBJDIR)/%$(DOTDLL)): private extra_ldflags.d += -L/EXPORT:gc_setProxy -L/EXPORT:gc_clrProxy
+$(dgpt:%=$(OBJDIR)/%$(DOTDLL)): private extra_dflags += -version=DLL
+$(dgpt:%=$(OBJDIR)/%$(DOTEXE)): private extra_dflags += -link-defaultlib-shared=false -dllimport=none
```
kinke, line 69: these 2 aren't just for the executables, but the DLLs too

runtime/druntime/test/shared/Makefile
```diff
@@ -179,12 +147,14 @@ $(ROOT)/linkD$(DOTEXE): private extra_ldlibs += $(abs_root)/lib$(for_linking)
...
 $(ROOT)/linkDR$(DOTEXE): $(DRUNTIME_DEP) utils.h
 $(ROOT)/linkDR$(DOTEXE): private extra_ldlibs += $(ldl) $(druntime_for_linking)
+ifneq ($(OS),windows)
+$(ROOT)/linkDR$(DOTEXE): private extra_ldflags += -Wl,-rpath,$(dir $(druntime_for_linking))
+endif
 
 $(ROOT)/loadDR.done: $(ROOT)/lib$(DOTDLL) $(DRUNTIMESO)
 $(ROOT)/loadDR.done: private run_args = $(DRUNTIMESO)
 $(ROOT)/loadDR$(DOTEXE): utils.h
 $(ROOT)/loadDR$(DOTEXE): private extra_ldlibs += $(ldl)
```
kinke, line 157: I *guess* Alpine/musl somehow needs the rpath too (as for `linkDR`, and what we previously enforced in CMake, `CFLAGS_BASE`) - the assertion/abort comes from being unable to load a dependency of either druntime or the `lib.so`.

runtime/druntime/test/shared/Makefile
```diff
@@ -30,21 +39,16 @@ endif
...
 for_linking := $(if $(filter windows,$(OS)),.lib,$(DOTDLL))
->>>>>>> origin/dmd-rewrite-stable
 
 ifeq (windows,$(OS)) # extra tests on Windows
 
-ifneq (,$(findstring ldmd2,$(DMD)))
+ifdef IN_LDC
     # LDC: required for executables to implicitly dllimport DLL data symbols
-    DFLAGS+=-dllimport=all
+    extra_dflags += -dllimport=all
```
kinke, line 51: This should now be superfluous, as `common.mak` now sets it for LDC too.

runtime/druntime/test/shared/Makefile
the-horo: You've already added a check like this to upstream's `exceptions` test but I think it will be easier if we keep the musl detection in cmake (since we already have a snippet for that) and just pass `IS_MUSL=1` to the makefiles.

kinke: Okay setting the rpath worked. - Are you going to upstream that `IS_MUSL` then?

kinke:
Thanks buddy!

the-horo:
There are still a few more tests failing which I accidentally overlooked but they should only be a few, I will go through them again

PR #4888, opened 2025-03-20
Revise unwind-table emission, enabling them for most popular targets

kinke:
Based on the clang 19 logic (but not 100% accurate).

There's an interesting special case - Darwin on arm64 apparently uses *synchronous* tables.

### DIFF
--- a/gen/abi/aarch64.cpp
+++ b/gen/abi/aarch64.cpp
@@ -57,6 +57,10 @@ private:
 public:
   AArch64TargetABI() {}
 
+  llvm::UWTableKind defaultUnwindTableKind() override {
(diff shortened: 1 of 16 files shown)


.github/actions/4d-test-libs/action.yml
```diff
@@ -24,8 +24,6 @@ runs:
           excludes+='|^std.internal.math.gammafunction'
           # FIXME: failing unittest(s) with enabled optimizations
           excludes+='|^std.math.exponential(-shared)?$'
-          # FIXME: subtest rt_trap_exceptions fails
-          excludes+='|^druntime-test-exceptions-debug$'
```
kinke, line 28: Damn, I hoped this would fix this subtest, but it doesn't.

Issue #4694, opened 2024-07-03
musl libc requires `-link-defaultlib-debug -frame-pointer=all` to show line information in backtrace


JohanEngelen:
See `tests/dmd/runnable/test17559.d` and `tests/dmd/runnable/test19086.d` failure without those flags.

kinke:
Fixed in https://github.com/ldc-developers/ldc/pull/4889.

PR #4889, opened 2025-03-22
Keep frame pointers by default with -O for some targets

### LINKED ISSUE #4594: Re-enable frame pointers for optimized code
See:
https://issues.dlang.org/show_bug.cgi?id=24439
https://news.ycombinator.com/item?id=39731824
https://www.brendangregg.com/blog/2024-03-17/the-return-of-the-frame-pointers.html
and
https://github.com/ldc-developers/ldc/issues/2222

Code change will be simply to adjust this:
https://github.com/ldc-developers/ldc/blob/3eb31901b3d253b3f23b8599111d930994696ef7/gen/functions.cpp#L507-L508

kinke:
Primarily to fix (druntime) backtraces with optimized code for those targets, incl. prebuilt release druntime/Phobos.

Fixes #4594 and #4694.

### DIFF
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -432,8 +432,6 @@ void applyTargetMachineAttributes(llvm::Function &func,
   opts::setFunctionAttributes(cpu, features, func);
   if (opts::fFastMath) // -ffast-math[=true] overrides -enable-unsafe-fp-math
     func.addFnAttr("unsafe-fp-math", "true");
-  if (!func.hasFnAttribute("frame-pointer")) // not explicitly set by user
-    func.addFnAttr("frame-pointer", isOptimizationEnabled() ? "none" : "all");
 }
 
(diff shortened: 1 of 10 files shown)


gen/functions.cpp
```diff
@@ -448,6 +446,23 @@ void applyXRayAttributes(FuncDeclaration &fdecl, llvm::Function &func) {
   }
 }
 
+// Keep frame pointers by default with enabled optimizations?
+// The logic is very loosely based on clang's
+// `useFramePointerForTargetByDefault()`, as well as backtrace test results for
+// druntime-test-exceptions-release.
```
kinke, line 452: See https://github.com/llvm/llvm-project/blob/cd708029e0b2869e80abe31ddb175f7c35361f90/clang/lib/Driver/ToolChains/CommonArgs.cpp#L75-L154.

Issue #4899, opened 2025-04-02
ldmd2 Segmentation fault: 11 for macOS 15.4 arm64


calvin2021y:
try v1.40.1 and v1.40.1-beta1, crash immediate.

ldc2 work.

```sh
./ldmd2 -v
Segmentation fault: 11
```

remy-moueza:
I have the same issue.  A compilation with ldc2 works.  However the compiled executable still fails with a segmentation fault. 

I have downloaded a nightly build to check if that issue was fixed.  I still encounter segmentation faults. 
I also have XCode installed and freshly udpated. 

I tried to run it through `lldb` to see if I could get a lead on what could be wrong with my set up. I did not understand much, but here is what I get, it might be helpful :  

`hello.d`:
```d
module hello;

void main() {
}
```

```sh
> ~/dev/conf/ldc2-06b2e42a-osx-arm64/bin/ldc2 hello.d
> ./hello
fish: Job 1, './hello' terminated by signal SIGSEGV (Address boundary error)
> lldb ./hello
(lldb) target create "./hello"
Current executable set to '/tmp/hello' (arm64).
(lldb) run
Process 30568 launched: '/tmp/hello' (arm64)
Process 30568 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x3c1f99f5530)
    frame #0: 0x000000018ab20bf4 libsystem_pthread.dylib`pthread_getspecific + 4
libsystem_pthread.dylib`pthread_getspecific:
->  0x18ab20bf4 <+4>: ldr    x0, [x8, x0, lsl #3]
    0x18ab20bf8 <+8>: ret

libsystem_pthread.dylib`pthread_setspecific:
    0x18ab20bfc <+0>: sub    x8, x0, #0xa
    0x18ab20c00 <+4>: cmp    x8, #0x2f5
Target 0: (hello) stopped.
(lldb)
```

Maybe the Mac OS 15.4 pthread dynamic library is incompatible with the one used to compile ldc2, or XCode libraries are causing trouble.  I haven't tried to remove XCode yet.

```
> lldb -v
lldb-1700.0.9.46
Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)

> clang -v
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: arm64-apple-darwin24.4.0
Thread model: posix
[...]

PR #4901, opened 2025-04-07
Switch FreeBSD CI to LLVM 19

arrowd:
The default LLVM version on FreeBSD Ports was switched to 19 a while ago. Current LDC bootstraps gets build against 15, which forces the port to pull in two LLVM versions, 15 and 19, at the same time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,14 @@ commonSteps: &commonSteps
           fi
     - checkout
     - run:
+        name: Merge PR with target branch
+        command: |
+          set -ux
+          if [ -n "${CIRCLE_PR_NUMBER:-}" ]; then
+            git fetch origin "+refs/pull/$CIRCLE_PR_NUMBER/merge:"
+            git checkout -f FETCH_HEAD
+          fi
+    - run:
         name: Checkout git submodules
         command: git submodule update --init
     - run:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -587,6 +587,12 @@ set_target_properties(
     # Use a custom .props file to set up Visual D (import paths, predefined versions...).
     VS_USER_PROPS "${PROJECT_SOURCE_DIR}/cmake/VisualD.props"
 )
+if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+    # FreeBSD LLVM port links to zstd, but does not convey this information via CMake
+    # Workaround it here until it is fixed in the port
+    find_package(zstd)
+    list(APPEND LLVM_LIBRARIES "$<TARGET_LINKER_FILE:zstd::libzstd_shared>")
+endif()
 # LDFLAGS should actually be in target property LINK_FLAGS, but this works, and gets around linking problems
 target_link_libraries(${LDC_LIB} ${LLVM_LIBRARIES} ${LLVM_LDFLAGS})
 if(WIN32)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -587,12 +587,6 @@ set_target_properties(
     # Use a custom .props file to set up Visual D (import paths, predefined versions...).
     VS_USER_PROPS "${PROJECT_SOURCE_DIR}/cmake/VisualD.props"
 )
-if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
-    # FreeBSD LLVM port links to zstd, but does not convey this information via CMake
-    # Workaround it here until it is fixed in the port
-    find_package(zstd)
-    list(APPEND LLVM_LIBRARIES "$<TARGET_LINKER_FILE:zstd::libzstd_shared>")
-endif()
 # LDFLAGS should actually be in target property LINK_FLAGS, but this works, and gets around linking problems
 target_link_libraries(${LDC_LIB} ${LLVM_LIBRARIES} ${LLVM_LDFLAGS})
 if(WIN32)
@@ -626,15 +620,20 @@ if(LDC_WITH_LLD)
     else()
         set(LDC_LINKERFLAG_LIST -lLLVMSymbolize ${LDC_LINKERFLAG_LIST})
     endif()
-    set(LLD_MACHO lldMachO)
     if(MSVC)
-        list(APPEND LDC_LINKERFLAG_LIST lldMinGW.lib lldCOFF.lib lldELF.lib ${LLD_MACHO}.lib lldWasm.lib lldCommon.lib)
+        list(APPEND LDC_LINKERFLAG_LIST lldMinGW.lib lldCOFF.lib lldELF.lib lldMachO.lib lldWasm.lib lldCommon.lib)
     else()
-        set(LDC_LINKERFLAG_LIST -llldMinGW -llldCOFF -llldELF -l${LLD_MACHO} -llldWasm -llldCommon ${LDC_LINKERFLAG_LIST})
+        set(LDC_LINKERFLAG_LIST -llldMinGW -llldCOFF -llldELF -llldMachO -llldWasm -llldCommon ${LDC_LINKERFLAG_LIST})
     endif()
     if(APPLE)
         # LLD 13.0.0 on Mac needs libxar
         list(APPEND LDC_LINKERFLAG_LIST -lxar)
+    endif()
+    if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+        # FreeBSD LLVM port links to zstd, but does not convey this information via CMake
+        # Workaround it here until it is fixed in the port
+        find_package(zstd)
+        list(APPEND LDC_LINKERFLAG_LIST "$<TARGET_LINKER_FILE:zstd::libzstd_shared>")
     endif()
 endif()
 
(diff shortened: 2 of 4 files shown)


kinke:
I've tried LLVM 19: https://github.com/ldc-developers/ldc/pull/4865/commits/b884ed4b6799c122ca3d6f100a02b6a782fa558c

Then LLVM 18: https://github.com/ldc-developers/ldc/pull/4865/commits/184d7e02834d69c67d71217a00bbaf6647e21987

I finally gave up because of the lit-test failures and reverted to 15.

arrowd:
Yes, I now bumped into the same test failures. I'll look into them more.

arrowd:
Ok, this is really a mess, unfortunately. There is so much logic trying to guess proper filepaths and names on different OSes. It is fragile and FreeBSD is just an example of this.

I'm not experienced enough to quickly devise a different approach, so I'm just leaving this PR as it is hoping someone will pick it up.

arrowd:
All right, now that I made the CI green I noticed that I partly redid what #4665 does.

I think we should first land this PR, as it isn't as invasive as the first one and then try to overhaul the logic for searching the runtime libraries.

.circleci/config.yml
```diff
@@ -97,7 +97,7 @@ jobs:
       - CI_OS: linux
       - LLVM_MAJOR: 18
       - HOST_LDC_VERSION: 1.39.0
-      - EXTRA_CMAKE_FLAGS: "-DMULTILIB=ON -DRT_SUPPORT_SANITIZERS=ON -DBUILD_LTO_LIBS=ON"
+      - EXTRA_CMAKE_FLAGS: "-DMULTILIB=ON -DRT_SUPPORT_SANITIZERS=ON -DBUILD_LTO_LIBS=ON -DCOMPILER_RT_LIBDIR_OS=linux"
```
kinke, line 100: Why is this `DCOMPILER_RT_LIBDIR_OS` now needed everywhere? I'm afraid that's unacceptable.

CMakeLists.txt
```diff
@@ -826,9 +832,15 @@ elseif(UNIX)
     set(COMPILER_RT_LIBDIR_OS_DEFAULT "x86_64-unknown-linux-gnu")
     set(COMPILER_RT_LIBDIR_OS "${COMPILER_RT_LIBDIR_OS_DEFAULT}"   CACHE STRING "Non-Mac Posix: OS used as directory name for the compiler-rt source libraries, e.g., 'freebsd'.")
     set(COMPILER_RT_LIBDIR "${COMPILER_RT_LIBDIR}/${COMPILER_RT_LIBDIR_OS}")
+    if(NOT COMPILER_RT_LIBDIR_OS STREQUAL COMPILER_RT_LIBDIR_OS_DEFAULT)
+        set(WANT_COMPILER_RT_LIBDIR_CONFIG TRUE)
```
kinke, line 836: Is this used anywhere?

CMakeLists.txt
```diff
@@ -826,9 +832,15 @@ elseif(UNIX)
...
+    endif()
 elseif(WIN32)
     set(COMPILER_RT_LIBDIR "${COMPILER_RT_LIBDIR}/windows")
 endif()
+if(NOT EXISTS ${COMPILER_RT_LIBDIR}/)
+    message(SEND_ERROR "Computed compiler-rt dir to be ${COMPILER_RT_LIBDIR}, but it does not exist!")
+endif()
```
kinke, line 843: The compiler-rt libs aren't mandatory.

CMakeLists.txt
kinke: Oh I see now, this controls whether the dir makes it to the config file. I guess it doesn't hurt to always add it here in this branch - if the dir exists. When using `LDC_INSTALL_LLVM_RUNTIME_LIBS`, it's disabled later anyway.

.circleci/config.yml
the-horo: I guess we could change it to `linux` here if it's the more common value https://github.com/ldc-developers/ldc/blob/06b2e42a1b8436faae2b7976a1d41a635df116d5/CMakeLists.txt#L826-L827

CMakeLists.txt
kinke: [Yeah I very much like that in principle, forwarding CMake options to the compiler, to put less burden on the hardcoded compiler magic.]

CMakeLists.txt
kinke: But a warning would be appreciated.

.circleci/config.yml
arrowd: > Why is this DCOMPILER_RT_LIBDIR_OS now needed everywhere?

I made it a hard error if the directory does not exist

> I'm afraid that's unacceptable.

Why? It serves as a safebelt for a user. Without this it took me a while to figure out why profiling tests are failing on FreeBSD - turns out that profiling functionality is silently disabled if the compiler can't locate the runtime dir properly.

kinke: > > Why is this DCOMPILER_RT_LIBDIR_OS now needed everywhere?
>
> I made it a hard error if the directory does not exist

Ah alright, so it really just worked accidentally, and the default value should be `linux` then.

> > I'm afraid that's unacceptable.
> 
> Why?

Because I thought then every user/maintainer needs to follow suit and set it now.

> Without this it took me a while to figure out why profiling tests are failing on FreeBSD - turns out that profiling functionality is silently disabled if the compiler can't locate the runtime dir properly.

Understood. Ideally, the according lit-tests would actually be disabled ('unsupported') if the compiler-rt libs aren't available, e.g.: https://github.com/ldc-developers/ldc/blob/06b2e42a1b8436faae2b7976a1d41a635df116d5/tests/PGO/lit.local.cfg#L2-L3

arrowd:
- Made `linux` the default.
- Made the hard error a warning.

It feels like macOS failure isn't my fault?

CMakeLists.txt
```diff
@@ -707,7 +713,7 @@ build_d_executable(
     "${DFLAGS_BUILD_TYPE} ${DFLAGS_LDC}"
     "${ALTERNATIVE_MALLOC_O};${LDC_LINKERFLAG_LIST};${FULLY_STATIC_LDFLAG}"
     "${FE_RES}"
-    "${LDC_LIB}"
+    "${LDC_LIB};${FREEBSD_LDC_DEPS}"
```
kinke, line 716: You don't set `FREEBSD_LDC_DEPS` anymore.

kinke:
> It feels like macOS failure isn't my fault?

That's almost certainly https://github.com/ldc-developers/ldc/issues/4899, now that GHA bumped their macos-15 image to v15.4. Thx Apple, greatly appreciated!

kinke:
> Made linux the default.

I guess that still needs investigating - a 'normal' LLVM build (with in-tree compiler-rt sources) seems to default to the full triple. As said above, that's what our LLVM builds require, and the FreeBSD port too. So it *looks* like the majority (?) of Linux distros went with `linux`, but that diverges from the LLVM default.

the-horo:
> I guess that still needs investigating - a 'normal' LLVM build (with in-tree compiler-rt sources) seems to default to the full triple. As said above, that's what our LLVM builds require, and the FreeBSD port too. So it _looks_ like the majority (?) of Linux distros went with `linux`, but that diverges from the LLVM default.

The jist is that when building `compiler-rt` separately you end up, by default, with `/linux` and when building it alongside `llvm` you end up with `/<triple>`. Relevant llvm cmake code for this is:

https://github.com/llvm/llvm-project/blob/5c8ba28c751bf0e7cb5c3ac9d223acba4cefc739/compiler-rt/cmake/base-config-ix.cmake#L106-L118

https://github.com/llvm/llvm-project/blob/5c8ba28c751bf0e7cb5c3ac9d223acba4cefc739/compiler-rt/cmake/Modules/CompilerRTUtils.cmake#L512-L519

With `LLVM_ENABLE_PER_TARGET_RUNTIME_DIR` being set at https://github.com/llvm/llvm-project/blob/5c8ba28c751bf0e7cb5c3ac9d223acba4cefc739/llvm/CMakeLists.txt#L959-L965

I guess I should have wrote these in #4665. It would have saved me for re-checking the llvm logic.

kinke:
Thx a lot Andrei. - So to me this is clearly the way to go then:
> we could also support both, preferring the full triple and checking if the dir exists; if it doesn't, trying to fall back to linux if that dir exists

(Doesn't have to be in this PR, can create one myself.)

arrowd:
Anything required from me to move this forward?

kinke:
At most a rebase after #4665 has landed, then you shouldn't have to deal with compiler-rt here anymore.

kinke:
Yeah, beta1 coming up this week. - If you wanna make the bootstrapping of the LDC FreeBSD port more independent, I suggest using GDC via its `gdmd` wrapper as host compiler.

arrowd:
It was the other way around - we switched from `gdmd` to binary LDC bootstraps because GDC 12+ also needs a bootstrap compiler. So using LDC bootstraps is lesser of two evils and we're actually depending on you guys.

kinke:
Mhm. The official FreeBSD package from GitHub is the only prebuilt package which doesn't link LLVM statically. I assumed the majority of those users are fine with a slim package, having the LLVM port installed already or quickly installing it if needed.

If static linking is feasible with the LLVM port, and it would make things easier for you guys, we could IMO switch.

kinke:
For us it wouldn't be any easier/harder. As long as you won't be needing bleeding-edge LLVM version support, you should be good then - as you probably know, it can take a bit of time until a new LLVM version is supported here in master, and some more time until a release with prebuilt packages is available.

PR #4903, opened 2025-04-11
CMake: Improve default value of COMPILER_RT_LIBDIR_OS on non-Mac Posix

kinke:
By preferring the LLVM host target triple (instead of hardcoding a Linux x86_64 glibc one), and falling back to the OS name alone if that directory exists.


.github/workflows/main.yml
```diff
@@ -251,6 +248,8 @@ jobs:
             os: android
             arch: armv7a
             android_x86_arch: i686
+            extra_cmake_flags: >-
+              -DCOMPILER_RT_LIBDIR_OS=linux
```
kinke, line 252: Our Android LLVM binaries have been cross-compiled on Linux, so not sure if it should always be `linux` (not `android`) on Android in general (as CMake special case), or whether this is just a side-effect of cross-compiling LLVM incl. compiler-rt.

the-horo: Using `linux` for android is intended, cross-compiling notwithstanding https://github.com/llvm/llvm-project/blob/5c8ba28c751bf0e7cb5c3ac9d223acba4cefc739/compiler-rt/cmake/base-config-ix.cmake#L98-L102

kinke: Ah nice, thx for checking!

driver/linker-gcc.cpp
```diff
@@ -269,26 +270,35 @@ std::string getCompilerRTLibFilename(const llvm::Twine &name,
...
-      (llvm::Twine("clang/") + llvm_major_version + "/lib/" + OSName + "/" +
-       name)
-          .str();
+  LLSmallVector<llvm::StringRef, 2> osNameCandidates;
+  if (triple.isOSDarwin()) {
+    osNameCandidates.emplace_back("darwin");
+  } else if (triple.isAndroid()) {
+    osNameCandidates.emplace_back("linux");
+  } else {
+    osNameCandidates.emplace_back(triple.str());
```
kinke, line 286: This seems to be required for the FreeBSD 19 port at least.

kinke:
Relevant changes cherry-picked to #4665.

PR #4912, opened 2025-04-21
GHA macOS: Enable bootstrap stage again

### LINKED ISSUE #4899: ldmd2 Segmentation fault: 11 for macOS 15.4 arm64

kinke:
And the latest stable LDC release as first host compiler.

Fixes #4899 for good.

### DIFF
--- a/.github/actions/1-setup/action.yml
+++ b/.github/actions/1-setup/action.yml
@@ -60,6 +60,9 @@ runs:
         mkdir clang
         tar -xf clang.tar.xz --strip 1 -C clang
         rm clang.tar.xz
+        # also upgrade CMake to v4
+        brew install cmake
+        cmake --version
     - name: 'Windows: Install clang v19.1.3 from GitHub'
       if: runner.os == 'Windows'
       shell: bash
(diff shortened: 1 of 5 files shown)


kinke:
Okay, looks like we could re-enable the bootstrap stage on macOS - at the cost of disabling mimalloc *and* C++ LTO for the arm64 package (no costs for the x86_64 one). That setup survived 15 consecutive CI job runs on each arch, and in my previous attempts last time, I've only seen very few CI job runs which would NOT show at least one 'Pure virtual function called!' error (as they are sporadic, not deterministic).

The only real improvement (when mixing LLVM versions) from last time is that I haven't seen any sporadic compiler crashes (during unwinding) for `driver/config_diag.d` anymore (probably fixed by #4888).

kinke:
Okay, a promising approach on arm64 is using Homebrew clang v19 for the final LDC build, instead of Apple clang - that enables C++ LTO (no more errors wrt. unsupported stack probe), and the mixed D & C++ LTO seems to work around the 'pure virtual function called' errors with mimalloc, as for x86_64. But I've only seen 2 consecutive green CI jobs so far, no stress-testing, so might still be lucky only.

PR #4913, opened 2025-04-24
Templatize _d_arraysetcapacity hook

aneiosi:
I added the changes of this pull from `dmd` [#21143](https://github.com/dlang/dmd/pull/21143)

I also added `version (D_BetterC)` checks to `core.internal.dassert.d`, `core.lifetime.d` and `core.internal.array.construction.d` to bypass the `try catch` blocks present in those files.

The overall aim is to make the `core` library entirely useable with `betterC` so as to have viable support for WebAssembly.

If these changes get approved, more will follow :)


kinke:
Some more context specific to the try-catch workarounds for `-betterC` / `-fno-exceptions`: https://github.com/dlang/dmd/issues/19970 - these shouldn't be necessary in the first place, if the compiler had better support for this no-exceptions mode.

And there's a more specific predefined version to use in workarounds - `D_Exceptions` (not set when compiling with `-fno-exceptions`, one of the `-betterC` effects). Used e.g. here: https://github.com/dlang/dmd/blob/bdd45686de74cad5986694810a360b4deae18f3d/druntime/src/core/exception.d#L12-L20

aneiosi:
> We'll get that change automatically when that hits DMD `stable`; I don't see a good reason to cherry-pick this early.

OK, my bad, then :)

> I _hate_ `-betterC`, and attempts to uglify druntime/Phobos code (dummy-templatizing without good reason etc.) to make it work somewhat 'better' (and so making the whole mess even worse), crippled as it is. - For wasm specifically, I think the way to go is to make parts of druntime compile for emscripten, and use that subset as sort of a minimal druntime for now / in the future. wasm doesn't have to remain `-betterC`-crippled, that's just a hack for now.

I think this comment is kinda funny because I'm at the exact opposite end of the spectrum. I think it's `emscipten` that is a hack and I'm very intentionally trying to avoid it. My goal is to compile for WASI which I think is the correct middle ground between native and browser environments. Humans are diverse, i guess :)

kinke:
I don't think we're on separate ends of the spectrum at all - targeting WASI directly would still require dropping the `-betterC` hack, and link a proper druntime. I just think emscripten would be *much* simpler, because they wrap WASI in a libc interface AFAIU, which druntime is based on.

If you haven't seen it already: a WASI port was already attempted some time ago:
* https://forum.dlang.org/thread/fgxyeisrgrfhnduhcwbi@forum.dlang.org
* https://github.com/skoppe/druntime/commit/bffaff347d5bc600599eefa3dbaaa982ae15442e

PR #4918, opened 2025-04-29
Merge upstream stable

kinke:


kinke:
Unfortunately there's apparently not going to be a v2.111.1 upstream - stable is already at v2.112.

The 2 'test' commits so become eager cherry-picks of https://github.com/dlang/dmd/pull/21343 and https://github.com/dlang/dmd/pull/21350, which fix v2.111 regressions showing up with LDC on a Symmetry codebase.

PR #4922, opened 2025-05-02
CI: Link libxml2 statically for the prebuilt Linux packages

### LINKED ISSUE #4921: arch linux: libxml2.so.2 => not found
OS: Arch Linux x86_64
libxml2 2.14.2-2

Latest arch update broke ldc

```
$ ldd dlang/ldc-1.41.0-beta1/bin/ldmd2
	linux-vdso.so.1 (0x000072452b780000)
	libm.so.6 => /usr/lib/libm.so.6 (0x000072452af08000)
	libz.so.1 => /usr/lib/libz.so.1 (0x000072452b738000)
	libxml2.so.2 => not found
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x000072452aedb000)
	libc.so.6 => /usr/lib/libc.so.6 (0x000072452aceb000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x000072452b782000)

```

https://gitlab.archlinux.org/archlinux/packaging/packages/libxml2/-/issues/4

That suggest it's an issue with how ldc was built

EDIT:

to future readers who need a workaround asap, install:  https://archlinux.org/packages/extra/x86_64/libxml2-legacy/

kinke:
Resolves #4921.


kinke:
Oof, more complicated because of the deps (libzstd didn't have any). Still need to check the impact on the final `ldc2` executable size; libicu isn't tiny.

kinke:
Ouch, at least on AArch64 it's way worse than expected! `ldc2` goes from 138 (latest master build) to 200 MB - with enabled LLVM/LDC assertions, so the relative increase would be even higher for a tagged release build.

kinke:
Very similar for x86_64: 140 => 201 MB.

kinke:
The libxml2 dependency will be disabled for the next prebuilt LLVM binaries; rationale: https://github.com/ldc-developers/ldc/issues/4921#issuecomment-2848631174

PR #4911, opened 2025-04-19
Add support for LLVM 20 (modulo dynamic-compile regressions)

kinke:
Includes #4843.

PR #4924, opened 2025-05-09
Preliminary support for LLVM 20 RTSan

### LINKED ISSUE #4755: Add RTSan support
> [!IMPORTANT]
> Requires LLVM 20 support!

Detect violations of unsafe system library calls (`malloc`, `free`, `pthread_mutex_lock`) in real time.

official-doc: https://clang.llvm.org/docs/RealtimeSanitizer.html
based ref.: https://github.com/realtime-sanitizer/rtsan

liushuyu:
This pull request adds basic RTSan (real-time sanitizer) support for D developers to check if their applications are real-time safe (i.e. suitable to use in a real-time context).

Needs #4911. Fix #4755.


runtime/druntime/src/rt/dmain2.d
```diff
@@ -70,6 +70,15 @@ extern (C) void thread_joinAll();
 extern (C) UnitTestResult runModuleUnitTests();
 extern (C) void _d_initMonoTime() @nogc nothrow;
 
+version (SupportSanitizers)
+{
+    import ldc.attributes;
+
+    @weak extern (C) void __rtsan_disable() @nogc nothrow {}
```
JohanEngelen, line 77: Have a look at module ldc.sanitizers_optionally_linked. I think that's more robust / works on more platforms probably?

cjappl:
One of the authors of rtsan here, was just linked this PR! 

Just shout if you have any questions. Excited to see it go into D

cjappl:
An overall note before I look any further - it may be simpler to break this up into:

1. Updating to llvm20
2. Adding rtsan

Seems like there are two major steps being done in this single PR. Breaking them up would make it easier to review and roll back upon an issue

liushuyu:
Hi, thank you for your review!

> An overall note before I look any further - it may be simpler to break this up into:
> 
>     1. Updating to llvm20
> 
>     2. Adding rtsan
> 
> 
> Seems like there are two major steps being done in this single PR. Breaking them up would make it easier to review and roll back upon an issue

This is also why this is a draft. I am currently waiting on LLVM upstream to fix a LLVM 20 regression that breaks LDC in a very bad way (see #4911). This pull request will be rebased when that LLVM 20 upgrade pull request lands.

runtime/druntime/src/rt/dmain2.d
liushuyu: Done.

PR #4926, opened 2025-05-11
macOS: Fix #3901 - CMake quirks on macOS wrt. CMAKE_C[XX]_COMPILER

### LINKED ISSUE #3901: CMake error on macOS Monterey, library not found for -lpthread

kinke:
See https://cmake.org/cmake/help/latest/policy/CMP0132.html.

### DIFF
--- a/.github/actions/helper-build-ldc/action.yml
+++ b/.github/actions/helper-build-ldc/action.yml
@@ -36,8 +36,6 @@ runs:
           -DLLVM_ROOT_DIR="$PWD/../${{ inputs.llvm_dir }}" \
           -DD_COMPILER='${{ inputs.host_dc }}' \
           -DLDC_LINK_MANUALLY=OFF \
-          ${{ runner.os == 'macOS' && '-DCMAKE_C_COMPILER=/usr/bin/cc' || '' }} \
-          ${{ runner.os == 'macOS' && '-DCMAKE_CXX_COMPILER=/usr/bin/c++' || '' }} \
           ${{ inputs.specify_install_dir == 'true' && '-DCMAKE_INSTALL_PREFIX="$installDir"' || '' }} \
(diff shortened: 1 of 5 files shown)


JohanEngelen:
This indeed fixed the issue on my macbook where the first CMake invoke would give an error, and invoking it again would fix it. Thanks!

Issue #4895, opened 2025-03-29
Exception stack traces on macOS/iOS are in a bad state


s-ludwig:
At least since the change to use "atos" for resolving stack frames, the state of printing call stacks is pretty disastrous on macOS, regardless of the `-g`  switch:

1. Printing an exception stack trace will freeze the calling thread for an extended period of time (can be a few hundred ms), which means that a process that logs/prints a larger number of exceptions can easily become unusable or would be a very easy DDoS target.

2. For x86-64 builds in particular, the stack unwinding functionality appears to be broken and only the topmost stack frame is being displayed.

3. Resolving file names and line numbers doesn't seem to work at all.

4. Stack traces generated for system crash reports usually stop at the first `extern (D)` function (I'm not sure whether this might be a simple issue of `--frame-pointer`, I'll have to check whether that makes a difference)

5. On some systems `atos` isn't available at all

In comparison, dmd appears to work fine (stack traces with symbols working regardless and file/line number working correctly when passing `-g`.

This has been an issue that wanted to dig deeper into for a while now, but never really got to it, so I just wanted to at least get this written down to maybe start a discussion.

kinke:
This can hardly be the general case, as macOS CI includes backtrace tests and is happy on both x86_64 and arm64. So at least some aspects of this seem specific to your setup.

Are you talking about unoptimized `-g` code, or optimized builds incl. `-O`? In the latter case, the missing frame pointers might be a problem (not just in your code, but druntime/Phobos too, unless using `-link-defaultlib-debug`). But those are going to be kept by default for Apple targets with the next version (#4889).

The `atos`-based implementation has been upstreamed, but DMD's druntime most likely doesn't use it because it probably still prevents the linker from stripping the `__debug_line` section (which we did before LDC v1.31 by hacking LLVM, before the `atos`-based implementation was merged in #4291).

Are you talking about distributed binaries, where e.g. the .dSYM files might be missing? [Note that I don't have any Apple stuff and cannot test anything myself.] Ref: https://github.com/ldc-developers/ldc/issues/4006

s-ludwig:
I've tested with and without `-g` and with and without optimizations. It's very well possible that there are some system dependent factors here. I'll try to summarize what I got here:

Snippet used for testing:
```D
void foo(int i)
{
	if (i > 0)
		foo(i - 1);
	else assert(false);
}

void bar()
{
	foo(4);
}

void main()
{
	bar();
}
```

[...]

PR #4930, opened 2025-05-15
Remove atos for backtrace generation.

LunaTheFoxgirl:
Using atos is not portable, given that:
1. It's a developer tool that may or may not exist on a given system
2. Said tool relies on being able to execute external applications at runtime; something the hardened runtime heavily restricts
3. Apple does not support this usecase.

As an alternative a weak `rt_dwarfSymbolicate` symbol has been added, allowing dub libraries to implement this functionality instead so that the developer can opt-in to this non-portable behaviour.

Down the line I can investigate making a naïve dSYM lookup system, but without OS help -- that Apple will reject on AppStore review -- full coverage will be unlikely.

### DIFF
--- a/runtime/druntime/src/core/internal/backtrace/dwarf.d
+++ b/runtime/druntime/src/core/internal/backtrace/dwarf.d
@@ -244,11 +244,8 @@ int processCallstack(Location[] locations, const(ubyte)[] debugLineSectionData,
 {
     if (debugLineSectionData)
         resolveAddresses(debugLineSectionData, locations, baseAddress);
-    version (Darwin)
-    {
-        if (!debugLineSectionData)
-            resolveAddressesWithAtos(locations);
-    }
+    else
+        resolveAddressesWithDladdr(locations);
 
     TraceInfoBuffer buffer;
     foreach (idx, const ref loc; locations)
@@ -267,138 +264,31 @@ int processCallstack(Location[] locations, const(ubyte)[] debugLineSectionData,
     return 0;
 }
 
-version (Darwin) {
-    /**
-     * Resolve the addresses of `locations` using `atos` (executable that ships with XCode)
-     *
-     * Spawns a child process that calls `atos`. Communication is through stdin/stdout pipes.
-     *
-     * After this function successfully completes, `locations` will contain
-     * file / lines informations.
-     *
-     * The lifetime of the `Location` data surpases function return (strndup is used).
-     *
-     * Params:
-     * locations = The locations to resolve
-     */
-    private void resolveAddressesWithAtos(Location[] locations) @nogc nothrow
-    {
-        import core.stdc.stdio : fclose, fflush, fgets, fprintf, printf, snprintf;
-        import core.stdc.stdlib : exit;
-        import core.sys.posix.stdio : fdopen;
-        import core.sys.posix.unistd : close, dup2, execlp, fork, getpid, pipe;
-        // Create in/out pipes to communicate with the forked exec
-        int[2] dummy_pipes; // these dummy pipes are there to prevent funny issues when stdin/stdout is closed and pipe returns id 0 or 1
-        int[2] pipes_to_atos;
-        int[2] pipes_from_atos;
-        if ( pipe(dummy_pipes) < 0 || pipe(pipes_to_atos) < 0 || pipe(pipes_from_atos) < 0 ) {
-            printf("some pipe creation error!\n");
-            return;
+/**
+ * Resolve the addresses of `locations` using dladdr.
+ *
+ * Params:
+ *   locations = The locations to resolve
+*/
+void resolveAddressesWithDladdr(ref Location[] locations) @nogc nothrow {
+    import core.demangle : demangle;
+    import core.sys.posix.dlfcn : dladdr, Dl_info;
+    import core.stdc.string : strlen;
+    foreach(ref Location location; locations) {
+        
+        Dl_info info;
+        dladdr(location.address, &info);
+        location.line = -1;
+        
+        // Strip off path to image.
+        if (info.dli_fname) {
+            location.file = info.dli_fname[0..strlen(info.dli_fname)];
         }
-        close(dummy_pipes[0]);
-        close(dummy_pipes[1]);
-        auto write_to_atos = pipes_to_atos[1];
-        auto read_from_atos = pipes_from_atos[0];
-        auto atos_stdin = pipes_to_atos[0];
-        auto atos_stdout = pipes_from_atos[1];
-        auto self_pid = cast(int) getpid();
-        // Spawn a child process that calls atos, reads/writes from the pipes, and then exits.
-        auto child_id = fork();
-        if (child_id == -1)
-        {
-            printf("some fork error!\n");
-            return;
+
+        // Demangle D symbols.
+        if (info.dli_sname) {
+            location.procedure = info.dli_sname[0..strlen(info.dli_sname)];
         }
-        else if (child_id == 0)
-        {
-            // We are in the child process, spawn atos and link pipes
-            // Close unused read/write ends of pipes
-            close(write_to_atos);
-            close(read_from_atos);
-            // Link pipes to stdin/stdout
-            dup2(atos_stdin, 0);
-            close(atos_stdin);
-            dup2(atos_stdout, 1);
-            close(atos_stdout);
-            char[10] pid_str;
-            snprintf(pid_str.ptr, pid_str.sizeof, "%d", cast(int) self_pid);
-            const(char)* atos_executable = "atos";
-            const(char)* atos_p_arg = "-p";
-            const(char)* atos_fullpath_arg = "-fullPath";
-            execlp(atos_executable, atos_executable, atos_fullpath_arg, atos_p_arg, pid_str.ptr, null);
-            // If exec returns, an error occurred, need to exit the forked process here.
-            printf("some exec error!\n");
-            exit(0);
-        }
-        // Parent process just continues from here.
-        // Close unused pipes
-        close(atos_stdin);
-        close(atos_stdout);
-        auto to_atos = fdopen(write_to_atos, "w");
-        auto from_atos = fdopen(read_from_atos, "r");
-        // buffer for atos reading. Note that symbol names can be super large...
-        static char[16 * 1024] read_buffer = void;
-        char* status_ptr = null;
-        foreach (ref loc; locations)
-        {
-            fprintf(to_atos, "%p\n", loc.address);
-            fflush(to_atos);
-            read_buffer[0] = '\0';
-            status_ptr = fgets(read_buffer.ptr, read_buffer.sizeof, from_atos);
-            if (!status_ptr)
-                break;
-            Location parsed_loc = parseAtosLine(read_buffer.ptr);
-            if (parsed_loc.line != -1)
-            {
-                // Only update the file:line info, keep the procedure name as found before (preserving the standard truncation).
-                loc.file = parsed_loc.file;
-                loc.line = parsed_loc.line;
-            }
-        }
-        if (!status_ptr)
-            printf("\nDid not succeed in using 'atos' for extra debug information.\n");
-        fclose(to_atos);
-        fclose(from_atos);
-        close(write_to_atos);
-        close(read_from_atos);
-    }
-    private Location parseAtosLine(char* buffer) @nogc nothrow
-    {
-        // The line from `atos` is in one of these formats:
-        // myfunction (in library.dylib) (sourcefile.c:17)
-        // myfunction (in library.dylib) + 0x1fe
-        // myfunction (in library.dylib) + 15
-        // 0xdeadbeef (in library.dylib) + 0x1fe
-        // 0xdeadbeef (in library.dylib) + 15
-        // 0xdeadbeef (in library.dylib)
-        // 0xdeadbeef
-        import core.stdc.stdlib : atoi;
-        import core.stdc.string : strchr, strstr;
-        import core.sys.posix.string : strndup;
-        Location loc;
-        if (!buffer)
-            return loc;
-        if (buffer[0] == '0' && buffer[1] == 'x')
-            // no named symbol found
-            return loc;
-        const symbolname_end = strstr(buffer, " (in ");
-        if (!symbolname_end)
-            return loc;
-        const symbolname_size = symbolname_end - buffer;
-        loc.procedure = strndup(buffer, symbolname_size)[0..symbolname_size];
-        const filename_start = strstr(symbolname_end, ") (") + 3;
-        if (cast(size_t)filename_start < 4)
-            return loc;
-        const colon_location = strchr(filename_start, ':');
-        if (!colon_location)
-            return loc;
-        const filename_size = colon_location - filename_start;
-        loc.file = strndup(filename_start, filename_size)[0..filename_size];
-        const final_paren = strchr(colon_location+1, ')');
-        if (!final_paren)
-            return loc;
-        loc.line = atoi(colon_location+1);
-        return loc;
     }
 }
 
--- a/runtime/druntime/src/core/internal/backtrace/dwarf.d
+++ b/runtime/druntime/src/core/internal/backtrace/dwarf.d
@@ -69,6 +69,7 @@ else
 
 import core.internal.container.array;
 import core.stdc.string : strlen, memcpy;
+import core.attribute : weak;
 
 //debug = DwarfDebugMachine;
 debug(DwarfDebugMachine) import core.stdc.stdio : printf;
@@ -237,6 +238,24 @@ struct TraceInfoBuffer
     }
 }
 
+/**
+ * A weakly linked hook which can be implemented by external libraries
+ * to extend the symbolication capabilites when debug info is missing.
+*/
+extern(C) void rt_dwarfSymbolicate(ref Location[] locations) @weak
+{
+    import core.sys.posix.dlfcn : dladdr, Dl_info;
+    Dl_info info;
+    foreach(ref Location location; locations)
+    {
+        if (dladdr(location.address, &info) == 0)
+            continue;
+        
+        location.file = info.dli_fname[0..strlen(info.dli_fname)];
+        location.procedure = info.dli_sname[0..strlen(info.dli_sname)];
+    }
+}
+
 private:
 
 int processCallstack(Location[] locations, const(ubyte)[] debugLineSectionData,
@@ -244,8 +263,7 @@ int processCallstack(Location[] locations, const(ubyte)[] debugLineSectionData,
 {
     if (debugLineSectionData)
         resolveAddresses(debugLineSectionData, locations, baseAddress);
-    else
-        resolveAddressesWithDladdr(locations);
+    else rt_dwarfSymbolicate(locations);
 
     TraceInfoBuffer buffer;
     foreach (idx, const ref loc; locations)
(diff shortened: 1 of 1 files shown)


runtime/druntime/src/core/internal/backtrace/dwarf.d
```diff
@@ -237,18 +238,32 @@ struct TraceInfoBuffer
...
+    Dl_info info;
+    foreach(ref Location location; locations)
+    {
+        if (dladdr(location.address, &info) == 0)
+            continue;
+        
+        location.file = info.dli_fname[0..strlen(info.dli_fname)];
+        location.procedure = info.dli_sname[0..strlen(info.dli_sname)];
+    }
+}
```
kinke, line 257: This is more or less already handled earlier: https://github.com/ldc-developers/ldc/blob/53ffdf11c9770a2b43ad36372122959bb3ce5997/runtime/druntime/src/core/runtime.d#L925-L948

LunaTheFoxgirl: Ah good to know, then this can just be a no-op stub

kinke:
> Do we need to have a function that processes the entire array at once? Why not a function that works on a single stack frame?

That gives greatest flexibility and reduces the need for some state/cache somewhere. The implementation will most likely have to read files etc., so processing the whole batch at once makes sense.

runtime/druntime/src/core/internal/backtrace/dwarf.d
kinke: Right. The hook's job is to populate the `file` and `line` fields of the `Location`s, if possible. So the name could reflect that.

LunaTheFoxgirl:
dladdr has been removed and a cleanup hook added. Also moved them be run before and after the rest of the symbol resolution.

LunaTheFoxgirl:
> I did not test this, but it looks like a regression to me (no more line info). I'm OK with making the `atos` thing opt-in, but I don't understand why it must be removed. Basically again we are left with backtraces without line info? Having to resort to a (yet non-existing) external implementations is very bad imo.

As said in #4895, having good stack traces is important, but those stack traces should not come at the cost of it being practically impossible to use D to release stuff on the AppStore. At this point D is already capable of being used for that, and at least in my business it's a part of my future strategy to release i(Pad)OS versions of my software.

runtime/druntime/src/core/internal/backtrace/dwarf.d
LunaTheFoxgirl: I've called it symbolicate since it may, on macOS also end up handling symbolication of Objective-C which requires another kind of roundtrip not related to DWARF. So in general it's supposed to be a hook that pre-symbolicates stuff (including the name of the procedure if relevant). Then letting the default implementation handle anything left over.

JohanEngelen:
> While yes, we could make the runtime only compile in atos stuff in debug mode, that does also mean there's extra friction for developers using DLang to develop apps; 

What is that extra friction? Do people put debug builds in the app store?

> and also that it's likely to be useless anyways on Darwin derived OSes that don't have atos.

This is not a valid argument to make things worse for macOS.

How about providing a separate small library with `rt_dwarfSymbolicate` overrides that call into `atos`, so that people can at least have the old (better ;)) behavior by adding `-lsomelib` to the compiler invoke?  (something similar to the `ldc_rt.dso.obj` lib on windows)

LunaTheFoxgirl:
> > While yes, we could make the runtime only compile in atos stuff in debug mode, that does also mean there's extra friction for developers using DLang to develop apps;
> 
> What is that extra friction? Do people put debug builds in the app store?
> 
> > and also that it's likely to be useless anyways on Darwin derived OSes that don't have atos.
> 
> This is not a valid argument to make things worse for macOS.
> 
> How about providing a separate small library with `rt_dwarfSymbolicate` overrides that call into `atos`, so that people can at least have the old (better ;)) behavior by adding `-lsomelib` to the compiler invoke? (something similar to the `ldc_rt.dso.obj` lib on windows)

This is not only about app store, it also affects things such as making builds with the hardened runtime. Which can end up making your life a headache if you want to sign and notarize your apps as well. Over all, I don't agree that this is the correct approach here.

And yeah, some may want to put release builds with debug info, for example, onto the app store. and also may not know to pass in extra LDC specific flags to link to the non-debug version of druntime, etc. While those can be somewhat solved by making dub more intelligent on the matter, it's just overall a massive ugly hack and your application freezing for a couple of seconds while atos runs every time you generate a stack trace is a little excessive.

LunaTheFoxgirl:
Also as a side-note, it's the default behaviour to generate debug symbols for releases with i(Pad)OS apps and the like, xcode does it for you on compile when using swift or Objective-C; so it is over-all expected that you will in fact have the debug symbols there even if it's technically in release mode. So if we for example, just detected whether the auto-generated dSYMs were there (which apple also forcefully does for you once you publish an app) instead of tying it to release-debug; then atos would still possibly be run. Additionally having strings in druntime relating to atos might be enough to trigger a review rejection.

schveiguy:
My opinion is that the easy experience for D should have stack traces with file/line. You need them when you are learning D, so that is what should be the default.

I'm OK with making D easier to release on the app store, and I'm hoping there's a way we can make it easy to do (it's OK to require some extra effort for this, with documentation). But it would be bad for the first experience with D to be unreadable stack traces.

LunaTheFoxgirl:
> My opinion is that the easy experience for D should have stack traces with file/line. You need them when you are learning D, so that is what should be the default.
> 
> 
> 
> I'm OK with making D easier to release on the app store, and I'm hoping there's a way we can make it easy to do (it's OK to require some extra effort for this, with documentation). But it would be bad for the first experience with D to be unreadable stack traces.

As I said in the other thread. Besides providing these hooks. Making a new utility that embeds dSYMs into the dwarf section might be the way to go. Means LDC avoids ugly hacks in its source tree that just makes it more difficult for existing D users to get their work done. My main goal right now is to ensure we get the ugly hacks rooted out that breaks production software for businesses like mine or Auburn Sounds who rely on D and LDC to make a living.

kinke:
Yeah my main concern is exactly that - the default experience *on a dev box* at least should be stack traces with resolved source Locs. Ideally with an acceptable runtime overhead etc., but that's secondary.

So a solution that depends on libatos (dynamically, i.e., copes with it not being available) would be fine to me as well - we depend on poorly documented stuff on Darwin already, that whole TLS disaster with macOS 15.4 was caused by Apple removing an API in macOS 10.15, and Jacob having to re-implement those finicky details in upstream druntime again. This stuff breaking was just a question of time I guess, and might happen again anytime.

rikkimax:
One option here might be to add a dedicated object file that is linked automatically for executables on Posix. Pass a switch and it won't do this.

That object file can contain the atos stuff, giving the desired default.

However, this is unnecessary if ``dlopen`` approach can ship.

LunaTheFoxgirl:
> One option here might be to add a dedicated object file that is linked automatically for executables on Posix. Pass a switch and it won't do this.
> 
> 
> 
> That object file can contain the atos stuff, giving the desired default.
> 
> 
> 
> However, this is unnecessary if ``dlopen`` approach can ship.

I'm not sure it can, Apple also scans strings. They'd easily find out that you are trying to load atos or that you are at some point in the application's life cycle.

If you try to hide this kind of stuff from them, they might revoke your dev and signing access.

dnadlinger:
Note that `compiler-rt` itself, in the sanitizer runtime, also shells out to `atos` on macOS for symbolizing. Has this caused any concrete issues with D projects being submitted to the App Store? Without any receipts to that end, I'm not sure whether there is something to fix here.

Of course, a more elegant solution could always be nice, but might be quite a bit of engineering effort.

LunaTheFoxgirl:
> Note that `compiler-rt` itself, in [the sanitizer runtime](https://github.com/llvm/llvm-project/blob/efa28338d858e1ea2bf705d50a0404bc602c8fe1/compiler-rt/lib/sanitizer_common/sanitizer_symbolizer_mac.cpp), also shells out to `atos` on macOS for symbolizing. Has this caused any concrete issues with D projects being submitted to the App Store? Without any receipts to that end, I'm not sure whether there is something to fix here (though of course the implementation might have issues; #4895).
> 
> Using private frameworks directy is another issue, but we are not doing that. Of course, a more elegant solution could always be nice, but might be quite a bit of engineering effort.

compiler-rt and the ASAN are not compiled in when you make release versions, so there it's irrelevant. The problem is that private APIs shouldn't make their way into production software; while you're developing and testing locally with self signed certs, it doesn't matter as much, even if you may end up with your application crashing due to using these things with no backtrace. At least you know what you're getting into there.

LunaTheFoxgirl:
I'll get back to this soonish; have some other stuff happening which means I need to focus on some other things.
Instead of going the atos route I'm thinking of adding so that a dSYM file is attempted to be located and loaded instead; since dSYM files are just mach-o files that only contains DWARF sections.

LunaTheFoxgirl:
Update: I got it working!
```
luna@feixiao n_test % cat test.d
import std.stdio;
import core.sys.darwin.dlfcn;

void main() {
	myFunction();
}


void myFunction() {
	throw new Exception("Test");
}
luna@feixiao n_test % rm test test.o; ldc2-dev -g test.d; dsymutil test; ./test
object.Exception@test.d(10): Test
----------------
test.d:10 void test.myFunction() [0x10057c993]
test.d:5 _Dmain [0x10057c8fb]
```

runtime/druntime/src/core/internal/macho/dl.d
```diff
@@ -0,0 +1,228 @@
...
+    /**
+        Returns the shared object with the given index.
+    */
+    static SharedObject fromIndex(uint idx) {
+        return SharedObject(
+            cast(mach_header_64*)_dyld_get_image_header(idx),
+            _dyld_get_image_vmaddr_slide(idx),
+            _dyld_get_image_name(idx)
+        );
+    }
```
jacob-carlborg, line 75: It looks like the `_dyld_get_image_` functions are discouraged from being used. Can we use `dladdr` to get the mach_header instead? (not sure how to get the slide).

LunaTheFoxgirl: These specific functions are still documented as usable by Apple in their man pages in Tahoe; so I don't think that's a problem? Alternatively yeah; we'd have to calculate the slide ourselves to account for address space randomisation; which would require a lot more work.

The functions which Apple does not support are the get_image functions that take a mach header, to be specific. The ones that take an image index are still available.

runtime/druntime/src/core/internal/macho/dl.d
```diff
@@ -0,0 +1,228 @@
...
+        foreach(i; 0.._dyld_image_count) {
+            if (int result = dg(SharedObject.fromIndex(i)))
+                return result;
+        }
+        return 0;
+    }
+}
+
+/**
+    A loeaded mach-o binary.
```
the-horo: typo

runtime/druntime/src/core/internal/macho/dl.d
jacob-carlborg: Yeah, I know. I was thinking about this comment: https://github.com/apple-oss-distributions/dyld/blob/93bd81f9d7fcf004fcebcb66ec78983882b41e71/include/mach-o/dyld.h#L46-L52. We already calculate some form of slide: https://github.com/LunaTheFoxgirl/ldc/blob/4e1fa51ef35d77606d86686376a2332ceee93a73/runtime/druntime/src/rt/sections_darwin_64.d#L139.

LunaTheFoxgirl:
> Perhaps we could get some tests?

I want to, but I have no idea where to slot in `dsymutil` in the makefiles in the runtime tests. They're quite confusingly set up.

the-horo:
> > Perhaps we could get some tests?
> 
> I want to, but I have no idea where to slot in `dsymutil` in the makefiles in the runtime tests. They're quite confusingly set up.

The closest example I can think of is https://github.com/ldc-developers/ldc/blob/46bbe8b47f7a69aff651e913d80dd846e1d6f613/runtime/druntime/test/exceptions/Makefile#L125-L141

If it helps you your example can be roughly translated as:
```
diff --git a/runtime/druntime/test/dsymutil/Makefile b/runtime/druntime/test/dsymutil/Makefile
new file mode 100644
index 0000000000..9059819c2b
--- /dev/null
+++ b/runtime/druntime/test/dsymutil/Makefile
@@ -0,0 +1,49 @@
+ifdef IN_LDC
+# Include this for OS
+include ../../../../dmd/osmodel.mak
+endif
+
+ifeq ($(OS),osx)
+# Only define the tests on macos
+endif
+TESTS := my_example
+
+
+# common.mak picks up TESTS and defines default build rules
+include ../common.mak
+
+# Adds -g to all built executables
+$(OBJDIR)/%$(DOTEXE): private extra_dflags += -g
+
+# Tell make how to produce the .dSYM from an object file
+$(OBJDIR)/%$(DOTEXE).dSYM: $(OBJDIR)/%$(DOTEXE)
+	dsymutil $<
+
+# Defines how to run the tests, they require the executable to be build and dsymutil to have been run.
+# And force use this rule instead of the one created by common.mak
+$(TESTS:%=$(OBJDIR)/%.done): $(OBJDIR)/%.done: $(OBJDIR)/%$(DOTEXE) $(OBJDIR)/%$(DOTEXE).dSYM
+	@echo Testing $*
+
+# Run the program and capture stderr
+# Print an error or something if the program didn't fail
+	if $(TIMELIMIT)$< 2>$(OBJDIR)/$*.stderr; then \
+		echo "Program completed unexpectedly. It should have failed" ; \
+		exit 1 ; \
+	fi
+
+# Check the stderr for a pattern
+	if ! grep -q "^object.Exception@src/my_example.d(10): Test$$" $(OBJDIR)/$*.stderr; then \
+		echo "The stderr is not alright" ; \
+		cat $(OBJDIR)/$*.stderr ; \
+		exit 1 ; \
+	fi
+
+# Or check for an exact file match, using sed to remove unportable address and other stuff
+# If you do this add `%.expected` to the target for make to properly track the dependency
+	if ! sed \
+		"s|^.*/src/|src/|g; s/\[0x[0-9a-f]*\]/\[ADDR\]/g; s/scope //g; s/Nl//g" $(OBJDIR)/$*.stderr \
+		| diff -q $*.expected -; then \
+		echo "The stderr did not match $*.expected exactly" ; \
+		cat $(OBJDIR)/$*.stderr ; \
+		exit 1 ; \
+	fi
diff --git a/runtime/druntime/test/dsymutil/my_example.expected b/runtime/druntime/test/dsymutil/my_example.expected
new file mode 100644
index 0000000000..acc96ac820
--- /dev/null
+++ b/runtime/druntime/test/dsymutil/my_example.expected
@@ -0,0 +1,4 @@
+object.Exception@src/my_example.d(10): Test
+----------------
+src/my_example.d:10 void test.myFunction() [ADDR]
+src/my_example.d:5 _Dmain [ADDR]
\ No newline at end of file
diff --git a/runtime/druntime/test/dsymutil/src/my_example.d b/runtime/druntime/test/dsymutil/src/my_example.d
new file mode 100644
index 0000000000..30e8600295
--- /dev/null
+++ b/runtime/druntime/test/dsymutil/src/my_example.d
@@ -0,0 +1,11 @@
+import std.stdio;
+import core.sys.darwin.dlfcn;
+
+void main() {
+	myFunction();
+}
+
+
+void myFunction() {
+	throw new Exception("Test");
+}

```

I'm not on a mac so I couldn't test it at all but you can try using this template.

the-horo:
worked for me: https://github.com/the-horo/ldc/actions/runs/16126084384

runtime/druntime/src/core/internal/macho/dl.d
```diff
@@ -0,0 +1,228 @@
...
+
+        Params:
+            path = Path to the object to load.
+        
+        Returns:
+            A SharedObject instance matching the given path,
+            or an empty handle on failure.
+    */
+    static SharedObject fromFile(const(char)* path) {
+        if (auto hndl = dlopen(path, 0))
```
jacob-carlborg, line 110: When can this fail while the code below works?

LunaTheFoxgirl: It's a fallback to support `-preserve-dwarf-line-section`.

It will fail on dSYM files specifically because they are given a special mach-o ID which dlopen won't open.

This is why there's manual mach-o verification further down in the function to ensure we're actually loading a DSYM mach-o object file, and not something random.

jacob-carlborg (approved):
LGTM. The style guide is not entirely followed, but I'll let someone else worry about that.

LunaTheFoxgirl:
> Are there any plans to support reading the debug info from object files as well?

This would be far more involved work. For now I think just using dSYM is the most sensible way.
Especially if we add support in dub for automating this step.

JohanEngelen:
I still think we should let other tools do all the heavy lifting, rather than implement something (far inferior?) ourselves and pray it won't break with Apple's next OS release...

https://wiki.dwarfstd.org/Apple%27s_%22Lazy%22_DWARF_Scheme.md is a nice read by the way, it mentions the many things that are needed to make this work well.

LunaTheFoxgirl:
> Does this detect stale dSYM files?  (I don't see where it does that check)

It does not; but would be relatively trivial to add.

LunaTheFoxgirl:
> I still think we should let other tools do all the heavy lifting, rather than implement something (far inferior?) ourselves and pray it won't break with Apple's next OS release...
> 
> 
> 
> https://wiki.dwarfstd.org/Apple%27s_%22Lazy%22_DWARF_Scheme.md is a nice read by the way, it mentions the many things that are needed to make this work well.

Those tools are only available on macOS. This is the only portable way we can do this.

LunaTheFoxgirl:
> > > I still think we should let other tools do all the heavy lifting, rather than implement something (far inferior?) ourselves and pray it won't break with Apple's next OS release...
> > > https://wiki.dwarfstd.org/Apple%27s_%22Lazy%22_DWARF_Scheme.md is a nice read by the way, it mentions the many things that are needed to make this work well.
> > 
> > 
> > Those tools are only available on macOS. This is the only portable way we can do this.
> 
> It's not the only way, alternatives debated before.
> 
> > Additionally due to the sanity checks I've put in place; if Apple breaks something it shouldn't lead to crashes.
> 
> Be aware of what you are signing up to. If things break, who is going to fix it? Will we again have years (!) of no debug info on macOS?
> 
> > This is far more preferable than backtraces only working on macOS, and the backtraces causing the application to freeze for a few seconds; which in debugging situations like on iPadOS can cause the application to be killed by the os for being unresponsive.
> 
> Preference is subjective, and I expect the vast majority of Apple LDC users to use it for macOS, not on iPadOS etc. I expect there to again be reports of "debug info not working", time will tell.

Mainly me, my business is gearing up to release software for macOS, iOS, etc.
I practically volunteered to help maintain this stuff; so if it breaks again do ping me.

LunaTheFoxgirl:
@kinke Could this get a review?

kinke:
Sorry, I don't have much spare time lately, plus I hate dedicating any time to Apple specifics.

From https://github.com/ldc-developers/ldc/pull/4930#issuecomment-2887324233:
> Yeah my main concern is exactly that - the default experience on a dev box at least should be stack traces with resolved source Locs.

The changes to the druntime integration tests, adding a manual dsymutil invocation, clearly show that that'd be lost. Note that there's https://github.com/ldc-developers/ldc/issues/4006, which was never followed up on.

LunaTheFoxgirl:
> Sorry, I don't have much spare time lately, plus I hate dedicating any time to Apple specifics.
> 
> 
> 
> From https://github.com/ldc-developers/ldc/pull/4930#issuecomment-2887324233:
> 
> > Yeah my main concern is exactly that - the default experience on a dev box at least should be stack traces with resolved source Locs.
> 
> 
> 
> The changes to the druntime integration tests, adding a manual dsymutil invocation, clearly show that that'd be lost. Note that there's https://github.com/ldc-developers/ldc/issues/4006, which was never followed up on.

Having to run dsymutil is the case for every other natively compiled language on the platform, xcode just does it for you. So I don't think it's a big issue and is something that should be fixed in dub.

PR #4931, opened 2025-05-15
CI: Add native Windows arm64 job/package

kinke:


kinke:
Current blocker: `ldc2.exe` doesn't link, 3 misaligned relocations apparently. With `-link-internally`:
```
lld-link: error: misaligned ldr/str offset

lld-link: error: misaligned ldr/str offset

lld-link: error: misaligned ldr/str offset

Error: linking with LLD failed
```

Surprisingly better diagnostics with the MS linker:
```
ldc.lib(main.cpp.obj) : error LNK2048: relocation PAGEOFFSET_12L targeting '?global@@3UGlobal@@A' (0774A1B8) is invalid for the instruction (3D800100 at RVA 0023822C) at section 0x16 offset 0xAB0, due to bad alignment of offset to target (1B8); expected to be 16 bytes aligned
ldc.lib(main.cpp.obj) : error LNK2048: relocation PAGEOFFSET_12L targeting '?global@@3UGlobal@@A' (0774A1B8) is invalid for the instruction (3D811500 at RVA 0023850C) at section 0x16 offset 0xD90, due to bad alignment of offset to target (608); expected to be 16 bytes aligned
ldc.lib(main.cpp.obj) : error LNK2048: relocation PAGEOFFSET_12L targeting '?global@@3UGlobal@@A' (0774A1B8) is invalid for the instruction (3D80E900 at RVA 0023855C) at section 0x16 offset 0xDE0, due to bad alignment of offset to target (558); expected to be 16 bytes aligned
LINK : fatal error LNK1165: link failed because of fixup errors
```

I.e., coming from C++, and wrt. the `Global global` variable (defined in D). I wasn't really expecting anything with an alignment of 16 on WoA64 here in that struct (`real_t` is a 64-bit double).

PR #4933, opened 2025-05-17
Run C preprocessor in C11 mode

kinke:
Basically an eager cherry-pick of
* https://github.com/dlang/dmd/pull/21371
* https://github.com/dlang/dmd/pull/21370
* https://github.com/dlang/dmd/pull/21372
* https://github.com/dlang/dmd/pull/21391

PR #4934, opened 2025-05-18
Windows: Disable clang headers when C-preprocessing with clang-cl

kinke:
As they apparently use unsupported types like `__int128`, `__bf16` etc., even if preprocessing with `/std:c11`. Use the Microsoft headers instead.


kinke:
[Oh okay, depends on #4933 for `__declspec(_Noreturn)` support.]

PR #4935, opened 2025-05-18
CI: Bump LDC-LLVM to v20.1.5, and disable dynamic-compile feature by default with LLVM 20

kinke:


kinke:
I guess that's better than sticking to LLVM 19, even when losing the dynamic-compile feature for the prebuilt packages - as the LLVM 20 bump fixes #4921 and restores `-preserve-dwarf-line-section` for Apple targets. I didn't dare using the latter by default, as I'm unsure whether #3864 might still occur in that case.

PR #4945, opened 2025-06-06
[dcompute] update `@kernel` UDA, include launch dimension

thewilsonator:
This PR does not update the OpenCL nor CUDA backends to their corresponding attributes (`reqd_work_group_size(X, Y, Z)`. This is required for the Vulkan compute backend to function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
   - macOS: Fix weird linker error when running CMake the first time. (#3901, #4926)
   - Reworked integration of the LLVM compiler-rt libraries. Package maintainers may want to see [docs/compiler_rt.md](https://github.com/ldc-developers/ldc/blob/master/docs/compiler_rt.md). (#4665)
   - Somewhat simplify separate compiler and runtime builds, incl. cross-compiling LDC itself. (#4872)
+- **Breaking change for dcompute**: The special `@kernel` UDA is now a function and _**requires**_ parentheses as in `@kernel() void foo(){}`. Optionally you can provide launch dimensions, `@kernel([2,4,8])`, to specify to the compute runtime how the kernel is intended to be launched. 
 
 #### Platform support
 - Supports LLVM 15 - 20.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # LDC master
 
 #### Big news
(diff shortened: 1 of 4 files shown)


kinke:
Do you wanna include this in v1.41.0 final? Just asking because I was planning on releasing over the weekend.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
   - macOS: Fix weird linker error when running CMake the first time. (#3901, #4926)
   - Reworked integration of the LLVM compiler-rt libraries. Package maintainers may want to see [docs/compiler_rt.md](https://github.com/ldc-developers/ldc/blob/master/docs/compiler_rt.md). (#4665)
   - Somewhat simplify separate compiler and runtime builds, incl. cross-compiling LDC itself. (#4872)
+- **Breaking change for dcompute**: The special `@kernel` UDA is now a function and _**requires**_ parentheses as in `@kernel() void foo(){}`. Optionally you can provide launch dimensions, `@kernel([2,4,8])`, to specify to the compute runtime how the kernel is intended to be launched.
```
JohanEngelen: Is it possible to avoid this breakage by some hackery ?  (supporting `@kernel, @kernel(), and @kernel([2,4,8])`

thewilsonator: Not that I could come up with:
You can't rely on UFCS because then `@kernel` refers to the function symbol, not its value. 
You can't template it because you can't have a template resolve to two different symbols with the same name.

JohanEngelen:
@thewilsonator I'm not so happy with this. If someone writes `@kernel` it is accepted without error or warning. Also, it seems more complex than needed? Simply making `kernel` a public struct should work syntax-wise, and then you have to deal with types/value in LDC code. What I mean is: we should make it easy for the user, not necessarily easy for the LDC implementer.

https://d.godbolt.org/z/fvbx8Ece1

JohanEngelen:
> Hmm, I had not considered a public struct, but I don't think it will work because then `@kernel` will be an `alias` symbol reference to the struct, not an instance of it. For the same reason that a function doesn't work because it refers to the symbol not a return value of it.

But you can handle both in LDC's code no?  One case for the alias symbol (or whatever) in case the parens are missing, and then you apply the default constants, and another case (struct instance) when user used parens.

```d
if (auto sle = e->isStructLiteralExp()) { ... } 
else if (auto exp = e->isAliasWhateverSymbolExp()) { ... } 
```

PR #4949, opened 2025-06-07
Upgrade frontend & libs to v2.112

kinke:


kinke:
> Instead test21382 fails (because LDC does not emit member functions).

Interestingly, it seems to work on 32-bit Windows though. I had to double-check, thinking that the test muss somehow be skipped on Win32, but no, it seems to be run and work. We do set the DI parent scope of member function definitions to the struct/class type (in IR), we 'just' don't explicitly add any functions as members when defining the parent aggregate type in IR (as clang does).

PR #4950, opened 2025-06-08
Fix a bunch of errors and deprecation for LLVM 21

thewilsonator:
Best reviewed per commit.

I am unsure what the correct value to use for the `llvm::LocationSize` in the case we don't know the size.

There are still a bunch of deprecations from using `InsertPosition` e.g. from `gIR->topallocapoint()` instead of `BasicBlock::iterators` for insertion.

### DIFF
--- a/gen/abi/generic.h
+++ b/gen/abi/generic.h
@@ -260,7 +260,11 @@ struct IndirectByvalRewrite : ABIRewrite {
     auto &attrs = arg.attrs;
     attrs.clear();
     attrs.addAttribute(LLAttribute::NoAlias);
+#if LDC_LLVM_VER >= 2100
+    attrs.addCapturesAttr(llvm::CaptureInfo::none());
+#else
     attrs.addAttribute(LLAttribute::NoCapture);
+#endif
     if (auto alignment = DtoAlignment(arg.type))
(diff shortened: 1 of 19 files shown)


thewilsonator:
Also it seems that `CallGraphNode::removeCallEdgeFor` was removed. I'm not sure how that functionality is supposed to be replaced.

gen/abi/generic.h
```diff
@@ -260,7 +260,9 @@ struct IndirectByvalRewrite : ABIRewrite {
     auto &attrs = arg.attrs;
     attrs.clear();
     attrs.addAttribute(LLAttribute::NoAlias);
+#if LDC_LLVM_VER < 2100
     attrs.addAttribute(LLAttribute::NoCapture);
+#endif
```
kubo39, line 267: add `attrs.addCapturesAttr(llvm::CaptureInfo::none());` if LDC_LLVM_VER >= 2100?

thewilsonator: Thanks! Fixed this.

thewilsonator:
Ping

thewilsonator:
I will merge this tomorrow if there are no objections.

PR #4965, opened 2025-08-08
Use templated lowering for `_d_newarray*`

Albert24GG:
I am not sure why, but the dynamic array creation still relies on the old hooks, despite having templated versions of `_d_newarray*` that the expressions are lowered to in `expressionsem`. The old hooks are soon going to be removed, because they were only used by AA hooks but now that is no longer the case.

I did not remove [`DtoNewDynArray`](https://github.com/ldc-developers/ldc/blob/9db6d953149f4698bf017e0ac42f188f58e303a4/gen/arrays.cpp#L616) since it is still used for [ArrayLiteralExp](https://github.com/ldc-developers/ldc/blob/9db6d953149f4698bf017e0ac42f188f58e303a4/gen/toir.cpp#L2391),at least until the templated version of `_d_arrayliteralTX` will be integrated.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -1578,15 +1578,9 @@ public:
       // get dim
       assert(e->arguments);
       assert(e->arguments->length >= 1);
-      if (e->arguments->length == 1) {
-        DValue *sz = toElem((*e->arguments)[0]);
-        // allocate & init
-        result = DtoNewDynArray(e->loc, e->newtype, sz, true);
-      } else {
-        assert(e->lowering);
-        LLValue *pair = DtoRVal(e->lowering);
-        result = new DSliceValue(e->type, pair);
-      }
+      assert(e->lowering);
+      LLValue *pair = DtoRVal(e->lowering);
+      result = new DSliceValue(e->type, pair);
     }
     // new static array
     else if (ntype->ty == TY::Tsarray) {


kinke:
Thx! I don't really recall whether the GC2Stack regressions were the reason, but they are definitely a problem. - We have an extra LLVM pass that looks for GC allocations that don't escape their scope (after inlining & optimizing), and convert them to stack allocations instead. This was all based on the old non-templated druntime hooks, based on TypeInfos etc.

So for instance, this now fails: https://github.com/ldc-developers/ldc/blob/9db6d953149f4698bf017e0ac42f188f58e303a4/tests/codegen/gc2stack.d#L10-L18

as the GC2Stack pass doesn't eliminate the superfluos GC allocation anymore when switching to the templated lowering.

kinke:
Once everything GC-related is templated (or at least the stuff that makes sense), the GC2Stack would need to be overhauled, incl. accounting for the new unstable mangled names and populating the TypeInfo LLVM meta data (that we emit in IR for the later GC2Stack pass) differently etc. That's probably a bigger task, and I think only makes sense tackling once all hooks have been revised.

In the meantime, here for array GC allocations, I went the conservative route of using the old non-templated hook to keep GC2Stack working, as the druntime helpers are still available.

PR #4968, opened 2025-08-16
fiber/switch_context_riscv.S: disable executable stack

the-horo:
Like it is done for the other assembly files, add the GNU-stack note.

### DIFF
--- a/runtime/druntime/src/core/thread/fiber/switch_context_riscv.S
+++ b/runtime/druntime/src/core/thread/fiber/switch_context_riscv.S
@@ -6,7 +6,7 @@
  * Authors:   Denis Feklushkin
  */
 
-#if (__linux__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __DragonFly__) && __ELF__
+#if __ELF__
 /*
  * Mark the resulting object file as not requiring execution permissions on
  * stack memory. The absence of this section would mark the whole resulting
--- a/runtime/druntime/src/core/thread/fiber/switch_context_asm.S
+++ b/runtime/druntime/src/core/thread/fiber/switch_context_asm.S
@@ -13,7 +13,7 @@
  *          http://www.boost.org/LICENSE_1_0.txt)
  */
 
-#if (__linux__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __DragonFly__) && __ELF__
+#if __ELF__
 /*
  * Mark the resulting object file as not requiring execution permissions on
  * stack memory. The absence of this section would mark the whole resulting
--- a/runtime/druntime/src/ldc/eh_asm.S
+++ b/runtime/druntime/src/ldc/eh_asm.S
@@ -6,6 +6,7 @@
  * License:   <a href="http://www.boost.org/LICENSE_1_0.txt">Boost License 1.0</a>.
  */
 
+#if __ELF__
 /*
  * Mark the resulting object file as not requiring execution
  * permissions on stack memory. The absence of this section would mark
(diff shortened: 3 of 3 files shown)


the-horo:
I hope that this doesn't break anything, but I have no way of testing riscv

thewilsonator (approved):

thewilsonator:
cc @denizzzka

runtime/druntime/src/core/thread/fiber/switch_context_riscv.S
```diff
@@ -6,6 +6,18 @@
  * Authors:   Denis Feklushkin
  */
 
+#if (__linux__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __DragonFly__) && __ELF__
+/*
+ * Mark the resulting object file as not requiring execution permissions on
+ * stack memory. The absence of this section would mark the whole resulting
+ * library as requiring an executable stack, making it impossible to
+ * dynamically load druntime on several Linux platforms where this is
```
denizzzka: > on several Linux platforms

On several *nix platforms?

runtime/druntime/src/core/thread/fiber/switch_context_riscv.S
```diff
@@ -6,6 +6,18 @@
...
+ * forbidden due to security policies.
+ */
+.section .note.GNU-stack,"",%progbits
+#endif
+
+
 #if defined(__riscv)
```
denizzzka, line 21: I am not sure, maybe it is better to insert proposed `.section .note.GNU-stack,"",%progbits` after this line?

denizzzka:
I checked this PR on QEMU - druntime was built and tests passed

runtime/druntime/src/core/thread/fiber/switch_context_riscv.S
the-horo: I've copied the lines from https://github.com/ldc-developers/ldc/blob/9db6d953149f4698bf017e0ac42f188f58e303a4/runtime/druntime/src/core/thread/fiber/switch_context_asm.S#L16-L25

I don't object to changing both of these files though

runtime/druntime/src/core/thread/fiber/switch_context_riscv.S
the-horo: As in, after the riscv? That wouldn't work because this file is always compiled and linked into druntime regardless if the arch is riscv or not.

runtime/druntime/src/core/thread/fiber/switch_context_riscv.S
denizzzka: I think it's worth doing because it looks comment is outdated compared to the code. Maybe just remove "Linux" word here?

runtime/druntime/src/core/thread/fiber/switch_context_riscv.S
```diff
@@ -6,6 +6,18 @@
  * Authors:   Denis Feklushkin
  */
 
+#if (__linux__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __DragonFly__) && __ELF__
```
JohanEngelen: why is the `#if` needed? Seems better to include this section for all targets.

the-horo: Removed the `#if` here and in the other two files

JohanEngelen: Not sure about the other files. This functionality is perhaps only available for ELF and GCC `ld`  (I read that `lld` ignores this section and defaults to `noexecstack` for all object files), so the `#if ELF` may make sense in the other files that are used for broader range of targets including non-ELF ?

the-horo: Let's see if `#if __ELF__` is enough. I've added it to all three files since they are all part of the build, regardless of target. The riscv-specificity comes from the code being in an `#if defined(__riscv)`

kinke (approved):

PR #4971, opened 2025-08-18
[druntime-test]: Fix glibc-aarch64 skipped tests

the-horo:
The cmake files already check the triple for musl and that is a more reliable check compared to looking for Alpine's apk.

### DIFF
--- a/runtime/DRuntimeIntegrationTests.cmake
+++ b/runtime/DRuntimeIntegrationTests.cmake
@@ -57,7 +57,7 @@ else()
     list(REMOVE_ITEM testnames uuid)
 endif()
 
-set(musl "")
+set(musl "IS_MUSL=0")
 if(TARGET_SYSTEM MATCHES "musl")
     set(musl "IS_MUSL=1")
 endif()
--- a/runtime/druntime/test/exceptions/Makefile
+++ b/runtime/druntime/test/exceptions/Makefile
@@ -3,6 +3,15 @@ ifdef IN_LDC
 include ../../../../dmd/osmodel.mak
 endif
 
+ifndef IS_MUSL # LDC defines it externally
+    ifeq ($(OS),linux)
+        # FIXME: detect musl libc robustly; just checking Alpine Linux' apk tool for now
+        ifeq (1,$(shell which apk >/dev/null 2>&1 && echo 1))
+            IS_MUSL := 1
+        endif
+    endif
+endif
+
 TESTS=stderr_msg unittest_assert invalid_memory_operation static_dtor \
       future_message refcounted rt_trap_exceptions_drt catch_in_finally \
       message_with_null
--- a/runtime/druntime/test/importc_compare/Makefile
+++ b/runtime/druntime/test/importc_compare/Makefile
@@ -5,12 +5,19 @@ endif
 
 TESTS := importc_compare
 
+ifndef IS_MUSL # LDC defines it externally
+    ifeq ($(OS),linux)
+        # FIXME: detect musl libc robustly; just checking Alpine Linux' apk tool for now
+        ifeq (1,$(shell which apk >/dev/null 2>&1 && echo 1))
+            IS_MUSL := 1
+        endif
+    endif
+endif
+
 # FIXME: fails on Alpine v3.21 with conflicting struct declarations in the C headers:
 # /usr/include/asm-generic/fcntl.h(195): Error: struct `importc_includes.flock` conflicts with struct `importc_includes.flock` at /usr/include/fcntl.h(24)
-ifeq ($(OS),linux)
-    ifeq (1,$(IS_MUSL))
-        TESTS :=
-    endif
+ifeq ($(IS_MUSL),1)
+    TESTS :=
 endif
 
 # FIXME: fails on macOS arm64, e.g., due to unsupported `_Float16`


runtime/druntime/test/exceptions/src/memoryerror_stackoverflow.d
```diff
@@ -1,16 +1,19 @@
 import etc.linux.memoryerror;
+import core.volatile;
 
 pragma(inline, false):
 
 void f(ref ubyte[1024] buf)
 {
     ubyte[1024] cpy = buf;
+    volatileStore(&cpy[0], 1);
```
kinke, line 9: This is one option, another would be the `@optStrategy("none")` UDA (applied to the whole module, like the `pragma(inline, false)` above, which is enough for DMD).

runtime/druntime/test/importc_compare/src/importc_includes.c
```diff
@@ -110,7 +110,9 @@
 #include <linux/io_uring.h>
 #include <linux/eventpoll.h>
 #endif
+#if !(defined(__linux__) && defined(__aarch64__)) // /usr/include/bits/math-vector.h(162): Error: undefined identifier `__Float32x4_t`
```
kinke: I guess you haven't by chance spotted some macro that we could define (e.g., in `importc.h`) , to avoid these apparent type extensions? Not being able to include `math.h` on AArch64 is pretty bad IMO.

runtime/druntime/test/exceptions/Makefile
```diff
@@ -65,6 +58,23 @@ endif
 ifeq ($(OS),windows)
     TESTS+=winstack
 endif
+
+ifdef IN_LDC
+    ifeq ($(OS)-$(ARCH),linux-aarch64)
+        # This hits the "should never happen" default branch in the
+        # rt.dwarfeh._d_throw_exception:399 unwind switch statement.
+        TESTS := $(filter-out memoryerror_null_call,$(TESTS))
```
kinke, line 76: Interesting; would be good to check which unexpected value we're getting.

runtime/druntime/src/core/sys/posix/sys/types.d
```diff
@@ -103,7 +103,20 @@ version (linux)
...
     alias uint      gid_t;
     alias uint      mode_t;
-    alias ulong_t   nlink_t;
+
+  version (X86_64)
+    alias ulong nlink_t;
+  else version (S390)
+    alias size_t nlink_t;
+  else version (PPC)
+    alias size_t nlink_t;
```
kinke, line 112: I guess this would be needed for `PPC64` too, as that would currently use `uint`. **Edit**: Or *only* PPC64, the default `uint` being fine for 32-bit PPC.

runtime/druntime/test/exceptions/Makefile
the-horo:
```
w0             0xf7fe7fb8          4160651192
```
ok...

runtime/druntime/src/core/sys/posix/sys/types.d
the-horo: I though PPC was for all PPC, not just 32-bit, my bad

runtime/druntime/test/exceptions/src/memoryerror_stackoverflow.d
the-horo: Yeah, that's a possibility as well. Upstream already merged the volatile approach so I guess it's not worth bothering with this anymore: https://github.com/dlang/dmd/pull/21742

runtime/druntime/test/importc_compare/src/importc_includes.c
the-horo: Last time I touched compilable/stdcheaders which suffered from the same issue I didn't find anything. Maybe it's worth for me to take another look, since it's so easy to hit

runtime/druntime/test/exceptions/Makefile
the-horo: At least it's consistent, I get it on every run

runtime/druntime/test/importc_compare/src/importc_includes.c
the-horo: The last time I touched compilable/stdcheaders I don't remember having found anything useful. Though maybe I'll try to take a look again

runtime/druntime/test/exceptions/Makefile
kinke: Oh WTF.

runtime/druntime/test/importc_compare/src/importc_includes.c
kinke: I have a RPi 5 now, and with vanilla Raspbian 12 (incl. gcc-12), `math.h` can be included just fine. I don't have a `/usr/include/bits` dir there...

the-horo: So, the include-flow (for me) is like:
```c
// math.h

#include <bits/libc-header-start.h> /* defines __GNUC_PREREQ */
// ...
#include <bits/math-vector.h>
```

```c
// bits/math-vector.h

#if __GNUC_PREREQ(9, 0)
#  define __ADVSIMD_VEC_MATH_SUPPORTED
typedef __Float32x4_t __f32x4_t;
typedef __Float64x2_t __f64x2_t;
#elif __glibc_clang_prereq(8, 0)
#  define __ADVSIMD_VEC_MATH_SUPPORTED
typedef __attribute__ ((__neon_vector_type__ (4))) float __f32x4_t;
typedef __attribute__ ((__neon_vector_type__ (2))) double __f64x2_t;
#endif
```

Getting out of this I think is only possible with either:
1. `#undef __GNUC__` or `#undef __GNUC_MINOR__` (this would make the `__GNUC_PREREQ` fail)
2. define `__Float32x4_t`

`features.h` defines `__GNUC_PREREQ` as:
```c
#if defined __GNUC__ && defined __GNUC_MINOR__
# define __GNUC_PREREQ(maj, min) \
        ((__GNUC__ << 16) + __GNUC_MINOR__ >= ((maj) << 16) + (min))
#else
# define __GNUC_PREREQ(maj, min) 0
#endif
```

kinke: Mhm; in my case, `/usr/include/aarch64-linux-gnu/bits/math-vector.h` boils down to a:
```C
#include <bits/libm-simd-decl-stubs.h>
```

A potential avenue to explore would be to define `__Float32x4_t` and `__Float64x2_t` in `__importc_builtins.di` - as aliases to the D vector types. That might work with gcc then; I can't think of how we'd be able to support clang's `__attribute__ ((__neon_vector_type__ (4))) float` - unless we'd go ahead and try to fake an older gcc/clang version (< v9/v8 apparently) on Linux AArch64 only.

the-horo: I think 2 is safer because 1 can affect a lot of things. With:
```diff
diff --git i/runtime/druntime/src/importc.h w/runtime/druntime/src/importc.h
index d60fcfff75..f8a2fd4c1d 100644
--- i/runtime/druntime/src/importc.h
+++ w/runtime/druntime/src/importc.h
@@ -187,8 +187,19 @@ typedef unsigned long long __uint64_t;
 #define _Float128 long double
 #define __float128 long double
 #endif
+
+#ifdef __aarch64__
+// Fix math.h include
+typedef __attribute__ ((__neon_vector_type__ (4))) float __Float32x4_t;
+typedef __attribute__ ((__neon_vector_type__ (2))) double __Float64x2_t;
+
+typedef struct {} __SVBool_t;
+typedef struct {} __SVFloat32_t;
+typedef struct {} __SVFloat64_t;
 #endif

+#endif // __linux__
+
 #if __APPLE__
 #undef __SIZEOF_INT128__
 #endif
```
`dmd-testsuite` and `druntime-test` passed for me. I'm not 100% sold on those empty struct `typedef` but those types needed to be defined.

the-horo: > Mhm; in my case, `/usr/include/aarch64-linux-gnu/bits/math-vector.h` boils down to a:
> 
> ```c
> #include <bits/libm-simd-decl-stubs.h>
> ```
> 
> A potential avenue to explore would be to define `__Float32x4_t` and `__Float64x2_t` in `__importc_builtins.di` - as aliases to the D vector types. That might work with gcc then; I can't think of how we'd be able to support clang's `__attribute__ ((__neon_vector_type__ (4))) float` - unless we'd go ahead and try to fake an older gcc/clang version (< v9/v8 apparently) on Linux AArch64 only.

I wish github would show comments without requiring me to manually refresh the page. Anyway, yeah, if we could map those types to something that makes sense it would be great but the testsuite seems to not use them so my placeholder definitions were enough to get it to stop erroring. I prefer this to trying to force an older gcc/clang.

kinke: Okay that's pretty good news! So if that clang typedef 'works' for importC, I *guess* it strips all `__attribute__` stuff there and treats it as a scalar float/double. `alias __Float32x4_t = __vector(float[4])` (in `__importc_builtins.di`, either in general or just for AArch64) would probably be correct, and `alias __Float64x2_t = __vector(double[2])`. No idea about the 3 `__SV*` types.

**Edit**: According to [godbolt](https://godbolt.org/z/xhnr5Pa5a), I guess an empty struct dummy is okay:
```
<source>:3:1: error: SVE type '__SVBool_t' does not have a fixed size
    3 | __SVBool_t a; 
<source>:4:1: error: SVE type '__SVFloat32_t' does not have a fixed size
    4 | __SVFloat32_t b;
<source>:5:1: error: SVE type '__SVFloat64_t' does not have a fixed size
    5 | __SVFloat64_t c;
```

the-horo: The `__Float??x?` aliases seem to work:
```
dlang: arm64-desk-s /tmp/x # cat d.d
import importc_math;
import std.math;
void main () {
	__f64x2_t x;
	x[0] = PI * 0;
	x[1] = PI * 0.5;

	auto r = _ZGVnN2v_cos(x);
	import std.stdio;
	writeln(r);
}
dlang: arm64-desk-s /tmp/x # cat importc_math.c
#include <math.h>
dlang: arm64-desk-s /tmp/x # ~/build/bin/ldc2 -run d.d
[1, 6.12323e-17]
```

The only issue that I hit was that the vector types are not indexable in C:
```
dlang: arm64-desk-s /tmp/x # cat c.c
#include <math.h>
#include <stdio.h>
int main () {
	__f32x4_t x;
	x[0] = M_PI;

	__f32x4_t r = _ZGVnN4v_cosf(x);

	printf("%f\n", r[0]);
}
dlang: arm64-desk-s /tmp/x # ~/build/bin/ldc2 -o- c.c
c.c(5): Error: can only `*` a pointer, not a `__vector(float[4])`
 x[0] = M_PI;
  ^
c.c(9): Error: can only `*` a pointer, not a `__vector(float[4])`
 printf("%f\n", r[0]);
                 ^
```
I don't personally have any issue with this though as I only value importC, not compileC.

kinke: > I don't personally have any issue with this though as I only value importC, not compileC.

Exactly my thoughts. Just nice that one can resort to C function prototypes with vector types, passing our `float4`/`double2` types directly. And in case someone was to try to use the SVE types (or functions taking thiose etc.), a dummy empty struct should yield compile errors right away - definitely better than something accidentally compiling, but totally failing at runtime.

the-horo: I'm afraid that the SVE types are allowed by the compiler and their usage does lead to runtime errors.

kinke: Yeah well but I guess noone feeds such a function with a default-initialized SVE type, so should get a compile error when trying to pass in some vector. Alternatively, a forward-declared struct might work better, assuming that math.h still works then.

runtime/druntime/test/exceptions/Makefile
```diff
@@ -65,6 +58,33 @@ endif
...
+        #
+        # The failure is caused by the backtrace call in
+        # core.runtime.DefaultTraceInfo.this
+        ifeq ($(BUILD_SHARED_LIBS),ON)
+            TESTS := $(filter-out memoryerror_%,$(TESTS))
+        endif
+        $(ROOT)/memoryerror_%: private extra_ldflags.d += -link-defaultlib-shared=false
+    endif
+endif
+
```
kinke, line 96: This is the most controversial part. Whenever something only failed for (some) CI systems, I've tried to remove the test on that CI platform only - which is doable for compiler and druntime/Phobos unittests, but the druntime integration tests with their Makefiles are unfortunately a PITA in that regard. Just so that the tests are all included in a normal test run.

It might pay off to investigate further, incl. checking whether we can disable based on the `BUILD` var (debug/release).

the-horo: > which is doable for compiler and druntime/Phobos unittests, but the druntime integration tests with their Makefiles are unfortunately a PITA in that regard

Adding something akin to:
```make
TESTS := $(filter-out $(LDC_SKIP_DRUNTIME_TESTS),$(TESTS))
```
in `common.mak`, alongside setting that environment variable when `ctest -E 'ldc2-unittest|lit-tests|dmd-testsuite'` is run should work. I can't test right now though.

> It might pay off to investigate further, incl. checking whether we can disable based on the BUILD var (debug/release).

Yeah, I agree that we shouldn't skip tests on all runners simply because 1 misbehaves. Yet debugging in online CI is a pretty miserable experience

runtime/DRuntimeIntegrationTests.cmake
```diff
@@ -57,7 +57,7 @@ else()
     list(REMOVE_ITEM testnames uuid)
 endif()
 
-set(musl "")
+set(musl "IS_MUSL=0")
```
the-horo, line 60:
```suggestion
set(musl "IS_MUSL=")
```

make doesn't treat `0` as false in an `ifndef`

kinke: What I *want* is that we skip the upstream musl-detection, in all cases, so for non-musl targets too. So I hope the explicit `0` means that it's defined, and false.

the-horo: Oops, I saw `ifndef IS_MUSL` and my mind jumped to conclusions. Yeah, setting it to 0 here is the right call

Issue #4972, opened 2025-08-24
OSX Arm: OSX-Universal distribution needs support to the arm64-ios-simulator triple


MrcSnm:
So, a little bit of history:

I was trying to build my project for iOS. It did worked when I was using x86_64. But that actually happened because when one uses the libraries for lib-ios-x86_64 is already inferred as a simulator since actual x86_64 iOS phones does not exists.

The same doesn't happen for the arm64 host. When one tries to build to lib-ios-arm64, they're building for the real device.

For actually building to the iOS simulator, one needs the triple `arm64-apple-ios12.0-simulator`. Unfortunately, the simulator triple is more important than I wanted, since without it, it causes the following linking error:

`ld: building for 'iOS-simulator', but linking in object file (/Users/rdgsnm/Documents/HipremeEngine/build/appleos/D/static/libdruntime-ldc.a[4](bitop.o)) built for 'iOS'`

kinke:
This was reported in #4517, and fixed in #4584 for the bundled druntime/Phobos libs.

kinke:
So there's an arm64 simulator too? Wow, Apple really knows how to break balls.

MrcSnm:
Yeah I basically needed another apple device for testing on iOS since on x64 was really slow, but looks like they managed to make that nasty stuff too

kinke:
I was expecting the non-simulator arm64 iOS binaries to work on arm64 macOS too, inside some simulator container/VM, but oh well.

How did you try to build the runtime libs yourself? According to https://github.com/ldc-developers/ldc/blob/9db6d953149f4698bf017e0ac42f188f58e303a4/.github/actions/5a-ios/action.yml#L25-L33 that should look something like:
```
ldc-build-runtime --dFlags="-mtriple=arm64-apple-ios12.0-simulator" CMAKE_SYSTEM_NAME=iOS CMAKE_OSX_SYSROOT="/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk" CMAKE_OSX_ARCHITECTURES="arm64" CMAKE_OSX_DEPLOYMENT_TARGET="12.0"
```

kinke:
I'd hate to add yet another set of libs for that crappy Apple ecosystem, they are really busting my balls badly this year. If that's what you meant; wrt. `--targetSystem`, you most likely simply shouldn't have specified it.

MrcSnm:
[...]

PR #4974, opened 2025-08-25
Bundle arm64-ios-simulator libraries with arm64/universal macOS packages

MrcSnm:
Fixes the issue #4972 for the simulator


kinke:
Thx for trying! You'll need those libs though, adding that as an arm64-specific extra step in https://github.com/ldc-developers/ldc/blob/master/.github/actions/5a-ios/action.yml.

.github/actions/merge-macos/action.yml
```diff
@@ -91,6 +91,17 @@ runs:
...
+        {
+            switches ~= [
+                "-Xcc=-isysroot",
+                "-Xcc=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk",
+            ],
+            lib-dirs = [
+                "%%ldcbinarypath%%/../lib-ios-arm64-simulator",
+            ],
+            rpath = "%%ldcbinarypath%%/../lib-ios-arm64-simulator";
+        };
```
kinke: And this section needs to come *after* the `arm64-apple-ios` one below. And you should allow for some iOS version in the triple.

MrcSnm: Is that like arm64-apple-ios*-simulator ?

kinke: `.*`, this is a regex, see the header comment in some generated `ldc2.conf` (also wrt. why the simulator section needs to come after the normal iOS one).

MrcSnm:
@kinke Can you bump the test runner? Alpine for some reason hang up and stopped my thing on running

MrcSnm:
@kinke Got all green and tested the binaries also with my engine to build on the iOS and it worked for me

PR #4978, opened 2025-08-30
ldc2.conf dir: Slightly shuffle order and file names, and make `ldc-build-runtime --installWithSuffix` install a .conf file

kinke:
The proposal is to have:
* `00-docs.conf`
* `30-compiler.conf`: 'compiler-specific defaults'
  * The main goal is to add the implicit `-I` post-switch(es) (=> min priority as post-switches) for the bundled druntime/Phobos imports.
  * Might contain LLVM-version-specific hacks, like the `-func-specialization-size-threshold=1000000000` for LLVM 16.
  * Also drags in druntime and Phobos by default (so a normal (pre-)switch, overrideable in a later config file or by the user directly) for linking via `-defaultlib=phobos2-ldc,druntime-ldc`.
* 5X: This is the group for target-specific settings - mainly `lib-dirs` and according `rpath`, as well as potential extra switches (for cross-compilation and/or to default to `-link-defaultlib-shared[=false]` if only one shared/static version of the libs is available). So by far the most interesting group.
  * `50-target-default.conf`: default target settings, applying to *all* targets by default (as safe fallback for all the weird triples one might use for the actual default/native target)
  * `55-target-…`: cross-compilation targets, all overriding the `lib-dirs` (and `rpath` if supported by the target) inherited from the default target, and potentially appending extra switches required for cross-compilation/linking (`-Xcc`, `-gcc`, `-defaultlib` etc.).
* Optionally a `70-compiler-rt.conf` to append a `lib-dir` (in most cases to a directory provided by the system-default LLVM, which might support a few targets). Any `lib-dir` in the target-specific 5X group takes precedence.

### DIFF
--- a/.github/actions/5a-ios/action.yml
+++ b/.github/actions/5a-ios/action.yml
@@ -32,6 +32,7 @@ runs:
           CMAKE_OSX_DEPLOYMENT_TARGET="$deployment_target" \
           BUILD_LTO_LIBS=ON
 
+        # overwrite the generated .conf file (need to patch triple regex and append cross-compile flags)
         cat > installed/etc/ldc2.conf/55-target-ios-$arch.conf <<EOF
         "${arch}-apple-ios":
         {
--- a/.github/actions/5a-ios/action.yml
+++ b/.github/actions/5a-ios/action.yml
@@ -32,16 +32,17 @@ runs:
           CMAKE_OSX_DEPLOYMENT_TARGET="$deployment_target" \
           BUILD_LTO_LIBS=ON
 
-        section="\"$arch-apple-ios\":
+        cat > installed/etc/ldc2.conf/55-target-ios-$arch.conf <<EOF
+        "${arch}-apple-ios":
         {
             switches ~= [
-                \"-Xcc=-isysroot\",
-                \"-Xcc=$sysroot\",
+                "-Xcc=-isysroot",
+                "-Xcc=${sysroot}",
             ];
             lib-dirs = [
-                \"%%ldcbinarypath%%/../lib-ios-$arch\",
+                "%%ldcbinarypath%%/../lib-ios-${arch}",
             ];
-            rpath = \"%%ldcbinarypath%%/../lib-ios-$arch\";
-        };"
-        echo "$section" >> installed/etc/ldc2.conf/31-ldc-runtime-lib-ios-$arch.conf
+            rpath = "%%ldcbinarypath%%/../lib-ios-${arch}";
+        };
+        EOF
         cat installed/etc/ldc2.conf/*
--- a/.github/actions/5a-ios/action.yml
+++ b/.github/actions/5a-ios/action.yml
@@ -22,7 +22,7 @@ runs:
             sysroot='/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk'
         fi
 
-        installed/bin/ldc-build-runtime --ninja \
+        install/bin/ldc-build-runtime --ninja \
           --dFlags="-mtriple=$triple" \
           --ldcSrcDir="$PWD/ldc" \
           --installWithSuffix="-ios-$arch" \
@@ -33,7 +33,7 @@ runs:
           BUILD_LTO_LIBS=ON
 
         # overwrite the generated .conf file (need to patch triple regex and append cross-compile flags)
-        cat > installed/etc/ldc2.conf/55-target-ios-$arch.conf <<EOF
+        cat > install/etc/ldc2.conf/55-target-ios-$arch.conf <<EOF
         "${arch}-apple-ios":
         {
             switches ~= [
@@ -46,4 +46,4 @@ runs:
             rpath = "%%ldcbinarypath%%/../lib-ios-${arch}";
         };
         EOF
-        cat installed/etc/ldc2.conf/*
+        cat install/etc/ldc2.conf/*
--- a/.github/actions/5a-ios/action.yml
+++ b/.github/actions/5a-ios/action.yml
@@ -26,24 +26,13 @@ runs:
           --dFlags="-mtriple=$triple" \
           --ldcSrcDir="$PWD/ldc" \
           --installWithSuffix="-ios-$arch" \
+          RT_CONF_TRIPLE_REGEX="$arch-apple-ios" \
           CMAKE_SYSTEM_NAME=iOS \
           CMAKE_OSX_SYSROOT="$sysroot" \
           CMAKE_OSX_ARCHITECTURES="$arch" \
           CMAKE_OSX_DEPLOYMENT_TARGET="$deployment_target" \
           BUILD_LTO_LIBS=ON
 
-        # overwrite the generated .conf file (need to patch triple regex and append cross-compile flags)
-        cat > install/etc/ldc2.conf/55-target-ios-$arch.conf <<EOF
-        "${arch}-apple-ios":
-        {
-            switches ~= [
-                "-Xcc=-isysroot",
-                "-Xcc=${sysroot}",
-            ];
-            lib-dirs = [
-                "%%ldcbinarypath%%/../lib-ios-${arch}",
-            ];
-            rpath = "%%ldcbinarypath%%/../lib-ios-${arch}";
-        };
-        EOF
+        # append cross-compile flags to generated .conf file
+        sed -i "" "s|^{\$|{\n    switches ~= [\n        \"-Xcc=-isysroot\",\n        \"-Xcc=${sysroot}\",\n    ];|" install/etc/ldc2.conf/55-target-ios-$arch.conf
         cat install/etc/ldc2.conf/*
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -280,7 +280,18 @@ if(SHARED_LIBS_SUPPORTED)
     set(install_rpath "${install_libdir}")
 endif()
 
-makeConfSection(NAME "50-target-default"
+# LIB_SUFFIX is (ab)used by ldc-build-runtime and set to the optional `--installWithSuffix`.
+# If set and non-empty, change the name of the generated .conf file.
+set(conf_base_name "50-target-default")
+if(NOT "${LIB_SUFFIX}" STREQUAL "")
+    if(LIB_SUFFIX MATCHES "^-")
+        set(conf_base_name "55-target${LIB_SUFFIX}")
+    else()
+        set(conf_base_name "55-target-${LIB_SUFFIX}")
+    endif()
+endif()
+
+makeConfSection(NAME "${conf_base_name}"
     SECTION "default"
 
     BUILD
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -283,8 +283,10 @@ endif()
 # This is e.g. set from `ldc-build-runtime --installWithSuffix`.
 set(RT_CONF_BASE_NAME "50-target-default" CACHE STRING "Base name of the generated runtime .conf file")
 
+set(RT_CONF_TRIPLE_REGEX "default" CACHE STRING "Regex for the target triple in the generated runtime .conf file")
+
 makeConfSection(NAME ${RT_CONF_BASE_NAME}
-    SECTION "default"
+    SECTION ${RT_CONF_TRIPLE_REGEX}
 
     BUILD
     SWITCHES ${switches}
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -280,7 +280,7 @@ if(SHARED_LIBS_SUPPORTED)
     set(install_rpath "${install_libdir}")
 endif()
 
-makeConfSection(NAME "30-ldc-runtime-lib${LIB_SUFFIX}"
+makeConfSection(NAME "50-target-default"
     SECTION "default"
 
     BUILD
@@ -304,7 +304,7 @@ if(MULTILIB AND NOT "${TARGET_SYSTEM}" MATCHES "APPLE")
         set(install_rpath "${install_libdir}")
     endif()
 
-    set(name "31-ldc-runtime-lib${MULTILIB_SUFFIX}")
+    set(name "55-target-m${MULTILIB_SUFFIX}")
     set(sectionPlaceholder "LDC_CONF_MULTILIB_TRIPLE_REGEX")
     makeConfSection(NAME "${name}"
         SECTION "${sectionPlaceholder}"
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -280,18 +280,10 @@ if(SHARED_LIBS_SUPPORTED)
     set(install_rpath "${install_libdir}")
 endif()
 
-# LIB_SUFFIX is (ab)used by ldc-build-runtime and set to the optional `--installWithSuffix`.
-# If set and non-empty, change the name of the generated .conf file.
-set(conf_base_name "50-target-default")
-if(NOT "${LIB_SUFFIX}" STREQUAL "")
-    if(LIB_SUFFIX MATCHES "^-")
-        set(conf_base_name "55-target${LIB_SUFFIX}")
-    else()
-        set(conf_base_name "55-target-${LIB_SUFFIX}")
-    endif()
-endif()
+# This is e.g. set from `ldc-build-runtime --installWithSuffix`.
+set(RT_CONF_BASE_NAME "50-target-default" CACHE STRING "Base name of the generated runtime .conf file")
 
-makeConfSection(NAME "${conf_base_name}"
+makeConfSection(NAME ${RT_CONF_BASE_NAME}
     SECTION "default"
 
     BUILD
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -67,11 +67,12 @@ packaging_steps_template: &PACKAGING_STEPS_TEMPLATE
     cd ..
     perl -pi -e s?$PWD/install/?%%ldcbinarypath%%/../?g install/etc/ldc2.conf/*
     if [[ "$CI_OS" == "freebsd" ]]; then
-      cat > install/etc/ldc2.conf/35-ldc-default-CC.conf <<EOF
+      cat > install/etc/ldc2.conf/31-compiler-default-CC.conf <<EOF
+    // Default to the clang version matching LDC's LLVM version.
     "default":
     {
         switches ~= [ "-gcc=${CC}" ];
-    }
+    };
     EOF
     fi
     cp $CIRRUS_WORKING_DIR/{LICENSE,packaging/README} install
--- a/.github/actions/3-build-cross/action.yml
+++ b/.github/actions/3-build-cross/action.yml
@@ -144,7 +144,6 @@ runs:
           --installWithSuffix "" \
           CMAKE_INSTALL_PREFIX="$PWD/install" \
           INCLUDE_INSTALL_DIR="$PWD/install/import" \
-          CONF_INST_DIR="$PWD/install/etc" \
           "${flags[@]}"
 
     - name: Cross-compile LDC executables
--- a/.github/actions/5-install/action.yml
+++ b/.github/actions/5-install/action.yml
@@ -41,24 +41,6 @@ runs:
           cp ldc/packaging/README install/
         fi
 
-    - name: Make portable
-      shell: bash
-      run: |
-        set -eux
-        cd ..
-        absInstallDir="$PWD/install"
-        if [[ '${{ runner.os }}' == Windows ]]; then
-          # /d/a/1/install => D:/a/1/install
-          absInstallDir=$(cygpath --mixed "$absInstallDir")
-        fi
-        confs=( install/etc/ldc2.conf/* )
-        perl -pi -e "s|$absInstallDir/|%%ldcbinarypath%%/../|g" "${confs[@]}"
-        cat "${confs[@]}"
-
-    - name: Rename the installation dir to test portability
-      shell: bash
-      run: mv ../install ../installed
-
     - name: 'Windows: Copy curl & MinGW-w64-based libs'
       if: runner.os == 'Windows'
       shell: bash
@@ -66,7 +48,7 @@ runs:
         set -eux
         cd ..
 
-        cp libcurl/ldc2/* installed/bin/
+        cp libcurl/ldc2/* install/bin/
 
         curl -fL --retry 3 --max-time 60 -o mingw-w64-libs.7z \
           https://github.com/ldc-developers/mingw-w64-libs/releases/download/v8.0.0/mingw-w64-libs-v8.0.0.7z
@@ -80,4 +62,4 @@ runs:
         else
           model=64
         fi
-        cp -r "lib$model" ../installed/lib/mingw
+        cp -r "lib$model" ../install/lib/mingw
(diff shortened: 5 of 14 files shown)


runtime/CMakeLists.txt
```diff
@@ -304,7 +304,7 @@ if(MULTILIB AND NOT "${TARGET_SYSTEM}" MATCHES "APPLE")
         set(install_rpath "${install_libdir}")
     endif()
 
-    set(name "31-ldc-runtime-lib${MULTILIB_SUFFIX}")
+    set(name "55-target-m${MULTILIB_SUFFIX}")
```
the-horo, line 312: I think that keeping the config files (here and in CI) be predictably named would help in the future with `ldc-build-runtime`. I think that the current CI setup should be an approximation of what we would want `ldc-build-runtime` to be able to do. Since CI currently does `ldc-build-runtime --installWithSuffix=-ios-arm64` I think we should name the config files `55-target-ios-arm64` as that would make a possible future transition smoother. Maybe it's worth using the `target-lib` form as a runtime built with `LIB_SUFFIX=""` would correspond to a config file named `55-target` which looks weird.

---

Somewhat unrelated to this but we should also consider config switches that don't really have something to do with the runtime build, but are still necessary. Like the `-Xcc=${sysroot}` on macos and a supposed `-gcc=aarch64-unknown-linux-gnu` on linux. I think it would be best to put these settings in a separate file (maybe `55-target-ios-arm64-env`) in order to allow `ldc-build-runtime` to freely install the D portion of the config file, which we should be able to fully generate.

The ideal setup I picture would then be:
- the official releases come with `55-target-ios-arm64` and `55-target-ios-arm64-env`. The users can modify `-ios-arm64-env` however they wish, in case *somebody* has broken their setup or they are using an older compiler
- during CI, we would call `ldc-build-runtime --installWithSuffix=-ios-arm64` and that would build and install the libraries, as well as install `55-target-ios-arm64`. We would then manually write `55-target-ios-arm64-env` with the same settings that we currently add now
- if a user wanted to build another copy of the runtime, they would follow the same steps as CI, i.e. `ldc-build-runtime --installWithSuffix=<target>` and, if they need to, write a `55-target-<target>-env` file, preferably following documentation that we make available.

kinke: > I think that keeping the config files (here and in CI) be predictably named would help in the future with ldc-build-runtime

For this specific line, i.e., the multilib extra .conf, I went with `55-target-m32.conf` (derived from `-m32`), as I guess we don't have enough info at CMake time, and I also didn't want the `fill-multilib-triple.sh` script to *also* come up with a suited logic to additionally rename the file based on the normalized triple.

---

Yeah wrt. `ldc-build-runtime`, I haven't thought that through yet. So far, one would have to rename the generated `50-target-default.conf` to some `55-….conf`, and potentially add the extra switches that were presumably already passed in `--dflags` (with the exception of `-mtriple`, we'd have to filter that one out). An optional CMake var (set externally) for the config base name would make things a bit easier, and that would be the `installWithSuffix` option of that tool then. **Edit**: Hmm or we just do the rename as part of `ldc-build-runtime --installWithSuffix`, leaving CMake alone.

There's one scenario currently where we don't have a solution: the 50 default target only comes with static or shared druntime/Phobos, and so appends a `-link-default-shared[=false]` switch. Now a 55 override target comes with both variants - but it cannot re-enable both sets without completely overriding all `switches`, as it's a boolean switch, no way to undefine it...

kinke: > and potentially add the extra switches that were presumably already passed in `--dflags`

Okay no that definitely doesn't cut it, neither additionally using the `--cflags`, after a look at the CI usages for cross-compiling to iOS and Android - both are mainly configured via CMake, either by specifying CMAKE_OSX_SYSROOT etc. or even a `CMAKE_TOOLCHAIN_FILE` on Android. Plus we might have e.g. dflags that should only apply to the runtime builds, not all user builds later on too.

So yeah I guess adding the cross-switches remains a manual step. I don't like a pair of generated + manual `.env` files per such target though, and prefer in-place editing that file then, so that we'd normally have a single `55-target-…` file per distinct cross-compilation target.

https://github.com/ldc-developers/ldc/pull/4974 is already going to be a slight challenge - we'd have two iOS arm64 targets for the universal package, a native iOS one and the iOS-simulator one. And as it's apparently just a `-simulator` environment addition to the triple, we have a ordering problem which should otherwise not be the case in the 55 group - the simulator one needs to come after the normal iOS one, and it inherits the normal section. Hmm, maybe we should try to make the normal iOS one stricter, with something like `arm64-apple-ios[^-]*$`, so that it doesn't apply to the `-simulator` one...

kinke: Oh and wrt. the generated 55 .conf file - one major problem is figuring out a matching triple regex, which will almost certainly have to come from the user.

kinke:
I'm just playing around here with the 2nd commit for now, the `ldc-build-runtime` one - will be extracted to a follow-up PR if we wanna go that route. We should probably have an explicit CMake var for the .conf base name (or do the rename in the ldc-build-runtime tool, but then the CMake-install might overwrite an existing `50-target-default.conf`...), and have the tool do the triple regex patching, by having the user specify it in the ldc-build-runtime cmdline.

runtime/CMakeLists.txt
the-horo: > I went with 55-target-m32.conf (derived from -m32), as I guess we don't have enough info at CMake time

I don't like this. It is inconsistent with how the default built gets it's name (derived from `LIB_SUFFIX`), I think using `MULTILIB_SUFFIX` here would be more appropriate. In fact, I would rather have `ldc-build-runtime` be able to build and install the `-m32` libraries and nuke all the multilib logic out of cmake.

> and potentially add the extra switches that were presumably already passed in --dflags (with the exception of -mtriple, we'd have to filter that one out)

No, build flags should not end up in the installed image.

> And as it's apparently just a -simulator environment addition to the triple, we have a ordering problem which should otherwise not be the case in the 55 group - the simulator one needs to come after the normal iOS one, and it inherits the normal section. Hmm, maybe we should try to make the normal iOS one stricter [...]

If we want to put everything in a `55` group then those triples should be disjoint. Yes, making one stricter is the only sane solution I can think of.

> There's one scenario currently where we don't have a solution: the 50 default target only comes with static or shared druntime/Phobos, and so appends a -link-default-shared[=false] switch.

We can just add `-link-defaultlib-shared=false` in a `BUILD_SHARED_LIBS=BOTH` build. This is the current behavior and a user can always overwrite it on the command line with `-link-defaultlib-shared`.

> Oh and wrt. the generated 55 .conf file - one major problem is figuring out a matching triple regex, which will almost certainly have to come from the user.

How about this: `ldc-build-runtime -installWithSuffix=foo -configTriple=my-foo-target <build_flags>` Would install a `55-target-foo` config that contains:
```
my-foo-target:
{
    lib-dirs = [ "/install/libfoo" ]
    rpath = "/insta/foo"
}
```

This should work with any build system that properly configures a C compiler and sets `CC` to it, regardless of any switches that would be needed by the `foo` target. Internally, the cmake build would have a `CONFIG_TRIPLE` option and would change the filename for the configuration file to `55` instead of `50`, and change the `"default"` triple to `my-foo-target`. `ldc-build-runtime` then just needs to `ninja install` and that would be all.

What would be missing is the out-of-the-box support for the target. Adding it would require to find the `-Xcc` switches for the target and embed them in the config file, which starts to sound like we're trying to implement a mini build system in the config files, one that just hardcodes what `CC` and args should be. I don't think this would be a bad idea straight out, so long as we keep the switches to a small list and the user can overwrite them on the command line, in case we miss a more exotic setup or in case some external entity decides that it would be a good idea to break previous setups.

What I want to highlight is that those two config files have entirely different jobs. There's the `target-foo` for the druntime & phobos side, with a suitable libdir and possible switches, that should basically always be correct, users shouldn't need changing those. But there's also the `target-foo` for the: you're trying to cross compile a program, outside of a build system which (should) have the appropriate settings for configuring a C compiler (and maybe even a D one), said C compiler being a dependency for linking your binary, so we will try to pick some defaults that should match what you probably want to happen.

> I don't like a pair of generated + manual .env files per such target though, and prefer in-place editing that file then, so that we'd normally have a single 55-target-… file per distinct cross-compilation target.

For the reasons above, and the fact that implementing a diff analysis to figure out what can be changed in the config file (think of a build that didn't have any `switches` followed by a build for the same target that now contains `switches`) sounds way more complicated than splitting the parts that solely depend on how the runtime was built, that can be easily picked up in the cmake files, from the portions that have to do with how a user uses the built runtime; I think it would be better to have separately named files, though maybe not so close in name (maybe a `70-env-ios-simulator` or something). I also want `ldc-build-runtime` to keep being a thin layout around cmake so I don't want it to be required to support manual installation or moving of files or having to edit their contents too much, I would prefer the cmake build to be able to do as much as it's needed for a minimal setup.

kinke: > > I went with 55-target-m32.conf (derived from -m32), as I guess we don't have enough info at CMake time

> I don't like this. It is inconsistent with how the default built gets it's name (derived from LIB_SUFFIX), I think using MULTILIB_SUFFIX here would be more appropriate.

Maybe take a 2nd look at the diff? You had `31-ldc-runtime-lib${MULTILIB_SUFFIX}`, I changed it to `55-target-m${MULTILIB_SUFFIX}`.

---

Let me read the rest of your post later in more detail. I was also thinking about a `-confTripleRegex` option for the ldc-build-runtime tool, to let it do the patching (and make that required when using a non-empty `-installWithSuffix` option, as the `"default"` would otherwise break the real default target!). And maybe even reinstantiate an option to include default switches; no, I still don't like the .env file idea, because I want everything target-specific in a single file, not multiple files with identical triple regexes. Sure, patching a file is harder than adding a new one, but if ldc-build-runtime takes care of it with an extra cmdline switch, it should be fine for the user.

---

But most of this discussion is for the 2nd step, which I wanna extract to a different PR, to keep this one simple. I just went ahead for exploration purposes. Too bad I can't push the 1st commit to a remote branch and change this existing PR with comments to that branch; that would allow me to push the rest as another branch targeting the first one...

kinke: > What would be missing is the out-of-the-box support for the target. Adding it would require to find the -Xcc switches for the target and embed them in the config file, which starts to sound like we're trying to implement a mini build system in the config files, one that just hardcodes what CC and args should be. I don't think this would be a bad idea straight out, so long as we keep the switches to a small list and the user can overwrite them on the command line, in case we miss a more exotic setup or in case some external entity decides that it would be a good idea to break previous setups.

We're not just trying, this is the way the official packages have worked for years, and what people are accustomed to. No need for a build system to prepare anything.

For cross-linking to Windows, no extra switches are required at all, regardless of the host platform (thanks to `-link-internally` being the default on non-Windows hosts, and the bundled MinGW-based import libs, so that no official WinSDK + MSVCRT libs are required). Sure, importC isn't handled out of the box.

For cross-linking to other Apple targets on a Mac, we need to pass the according Xcode sysroot alone, that's 2 `-Xcc` switches. Probably works for importC too.

For cross-linking to non-Apple Posix targets (incl. Android), you usually need an according C cross-compiler, and so a simple `-gcc` switch in the LDC config. Should work for importC too.

kinke: > >  There's one scenario currently where we don't have a solution: the 50 default target only comes with static or shared druntime/Phobos, and so appends a -link-default-shared[=false] switch.
> 
> We can just add -link-defaultlib-shared=false in a BUILD_SHARED_LIBS=BOTH build. This is the current behavior and a user can always overwrite it on the command line with -link-defaultlib-shared.

Right, good point.

kinke: > I don't think this would be a bad idea straight out, so long as we keep the switches to a small list and the user can overwrite them on the command line

The very few expected switches should be overrideable (by the user or a build system); `-Xcc` switches depending on whether the C compiler allows them to be overridden. E.g., Apple clang's `-isysroot` seems overrideable just fine, that's currently used in #4974 (with 2 matching iOS sections for the simulator target, and so 2 `-isysroot`).

~~The only real problem OTOH would be a `-gcc` in the config, to be overridden by a CC env var set up by the external build system (or user) - the build system would have to set `-gcc` explicitly to that CC (as the explicit cmdline option takes precedence over the env var), and in case it'd be more than just the path to that C compiler (so including space-separated flags that you've added support for), it'd have to pass those separately as `-Xcc`.~~ **Edit**: Nope - an empty `-gcc=` activates the CC env var, so this is also doable easily.

kinke:
Rough sketch for a future `ldc-add-target` tool (or so), extending a local LDC installation by some cross-compilation target based on the prebuilt official packages, without building anything yourself (that's related `ldc-build-runtime`):

Tasks for a `ldc-add-target windows-arm64` example (`<target name>` being `windows-arm64`):
1. Download the according official release package (matching the LDC version, like `ldc-build-runtime` optionally downloading the associated source package) containing the libs for the desired target:  [ldc2-1.41.0-windows-multilib.7z ](https://github.com/ldc-developers/ldc/releases/download/v1.41.0/ldc2-1.41.0-windows-multilib.7z)
2. Unpack it and move the desired libs subdir to the local LDC installation, renaming to `lib-<target name>`.
3. Move the matching 5X config file from extracted `etc/ldc2.conf/` to the local LDC installation's `etc/ldc2.conf/`, renaming to `55-target-<target name>.conf` and patching the triple regex.
4. Optionally allow appending extra switches provided in the command-line to the config file, such as a `-gcc=aarch64-linux-gnu-gcc` for the `linux-aarch64` target.

Would need a mapping for each supported target, such as:
```yaml
windows-arm64:
    artifact_suffix: "windows-multilib.7z"
    artifact_lib_dir: "lib-windows-arm64"  # would be "lib64" for the windows-x86_64 target
    artifact_conf_file: "55-target-windows-arm64.conf"  # would be "50-target-default.conf" for the windows-x86_64 target
    conf_triple_regex: "(aarch|arm)64-.*-windows-msvc"
```

Then all a user would additionally need is the according C cross-compiler for non-Apple Posix targets (`apt install gcc-aarch64-linux-gnu`).

runtime/CMakeLists.txt
the-horo: > Maybe take a 2nd look at the diff? You had 31-ldc-runtime-lib${MULTILIB_SUFFIX}, I changed it to 55-target-m${MULTILIB_SUFFIX}.

I meant this as in keeping the multilib name exactly like one that would be produced by a cross-build, so `55-target-32` rather than `55-target-m32`. It's not that big of a deal to use different names though, I just think it would be nice to decide on how the config files would be named and keep the scheme, regardless of how the the runtime build and the config generation happens. I won't insist if you don't care about it.

> We're not just trying, this is the way the official packages have worked for years, and what people are accustomed to. No need for a build system to prepare anything.

Sorry, I didn't mean my point there to be read as asking whether such support should be implemented because, like you said, the current setup has been working like this on macos and it cannot be broken. Note that on other systems we still have the freedom to choose whether to configure CC in the config file, I'm still trying to go through the arguments to see if one setup would be more appropriate...

the-horo:
I've asked around a meson dev what he think of the changes and the short summary is: it would be great if this sort of CC detection didn't overwrite the `CC` from the environment. Suggestions to achieve this included trying to look for `<triple>-gcc` instead of just `cc` when `CC` is unset, or adding another key to the config file. I think the former sounds like a better idea.

runtime/CMakeLists.txt
kinke: > I meant this as in keeping the multilib name exactly like one that would be produced by a cross-build, so `55-target-32` rather than `55-target-m32`.

Well, the scheme I'd suggest would be `55-target-linux-x86` (or `55-target-freebsd-x86` etc.). But yeah I don't really care about this multilib config name, since the `-m32` support is IMO just legacy - x86 hardly needed nowadays on Posix, plus I think more recent archs don't support that switch anymore (e.g., no `-m32` for AArch64 compilers - `aarch64-unknown-linux-gnu-g++: error: unrecognized command-line option '-m32'` on godbolt).

kinke:
Yeah in essence I think the only controversial point here is the `-gcc`, and whether/where to add it to the config. As this remains a manual step, not done as part of `ldc-build-runtime`, people are free to do as they please. My perspective here is solely on the official packages (and so affecting `ldc-add-target/runtime`), where the `-gcc` would only come in for the Android x86[_64] cross-targets [which didn't have any config sections so far; I was too lazy, and noone complained] - and yeah, the FreeBSD package also sets `-gcc` as you've seen, but only to make sure that the same-LLVM-versioned clang is used (IIRC, that was required for compiler-rt support for CI at one point, but might not be needed anymore after your improvements there).

And my stance there wrt. `-gcc` is to provide a sane default, so that `-mtriple=…` alone suffices for cross-linking. So while I obviously have no idea where the guys installed their Android NDK, I think it's good that we still default to the expected clang wrapper (same name on Windows and Linux) at least. Otherwise I guess on Windows you'd get an error saying that it cannot find `gcc` for linking; `aarch64-linux-android30-clang` is much better in that regard - might be in PATH, yay, or otherwise provides much better context for what's required (and could be fixed once by the user specifying the full path in the config). And IIRC, those Android clang wrappers are just bash/cmd.exe scripts, no need to provide a `-target …` switch for some generic clang.

So similarly, when invoking `ldc-add-target linux-aarch64` to add that cross-compilation target, I wanna default to an `aarch64-linux-gnu-gcc`, not a generic `gcc`. But that's just an `ldc-add-target` detail (not in the official native Linux-aarch64 package's conf).

LDC package maintainers for distros such as you are free to do what they want, tweaking the config files to their system, and maybe even patching the `ldc-add-target` tool accordingly.

The only problematic aspect of defaulting to a `-gcc` is, as you mentioned, that it overrides a CC set up by some build system. Which can be avoided by passing `-gcc=` in the cmdline, something the build system could do. Adding something inbetween, like a `-defaultCC`, which would NOT override CC, isn't really something I'd like, just for this rather theoretical problem - the main 'build systems' for D (dub and reggae) don't set CC.

the-horo:
> > I very much dislike the idea of making ldc not respect CC and I much more prefer something akin to
> 
> Okay, making the compiler choose a somewhat sane default cross-CC when explicitly providing an `-mtriple` option is definitely on the table. We'd have to come up with some rules, like doing that for non-Apple/Windows targets only, and maybe stripping the vendor automatically (or simply assuming the user won't specify it for those targets). **Edit**: And probably including some gcc vs. clang logic, so that we use clang for Android (and FreeBSD?).

Alright then, how do we want the list of lookups to be like?

- Arch uses `aarch64-linux-gnu-gcc`: https://archlinux.org/packages/extra/x86_64/aarch64-linux-gnu-gcc/files/
- Debian uses `aarch64-linux-gnu-gcc`: https://packages.debian.org/bullseye/amd64/gcc-aarch64-linux-gnu/filelist
- Gentoo uses `aarch64-unknown-linux-gnu-gcc`

clang here doesn't come with a `aarch64-unknown-linux-gnu-clang` script and I can't find anything on debian and arch as well so I assume it similarly doesn't exist there.

The other OSes I don't know about.

kinke:
Thx for checking. - Oof, so Gentoo does include the nonsensical vendor. I'd go with NOT tampering with the vendor then, so that it works on Gentoo when using `-mtriple=aarch64-unknown-linux-gnu`, and otherwise using the `-mtriple=aarch64-linux-gnu`.

And wrt. clang vs. gcc, I'd only use clang where we know it's the only or default supported compiler. I.e., Android for recent NDKs, and maybe FreeBSD if it comes with these wrappers; no idea where else clang is the default by now (=> hope for PRs from users if they care about it).

kinke:
Okay Andrei, ready now from my side. Please let me know if you can live with this, or what you'd like changed.

.github/actions/5a-ios/action.yml
```diff
@@ -22,26 +22,17 @@ runs:
...
-            ];
-            lib-dirs = [
-                \"%%ldcbinarypath%%/../lib-ios-$arch\",
-            ];
-            rpath = \"%%ldcbinarypath%%/../lib-ios-$arch\";
-        };"
-        echo "$section" >> installed/etc/ldc2.conf/31-ldc-runtime-lib-ios-$arch.conf
-        cat installed/etc/ldc2.conf/*
+        # append cross-compile flags to generated .conf file
+        sed -i "" "s|^{\$|{\n    switches ~= [\n        \"-Xcc=-isysroot\",\n        \"-Xcc=${sysroot}\",\n    ];|" install/etc/ldc2.conf/55-target-ios-$arch.conf
```
the-horo, line 37: I think this would look better by using a `cat <<EOF`

runtime/CMakeLists.txt
```diff
@@ -280,8 +280,13 @@ if(SHARED_LIBS_SUPPORTED)
...
 endif()
 
-makeConfSection(NAME "30-ldc-runtime-lib${LIB_SUFFIX}"
-    SECTION "default"
+# This is e.g. set from `ldc-build-runtime --installWithSuffix`.
+set(RT_CONF_BASE_NAME "50-target-default" CACHE STRING "Base name of the generated runtime .conf file")
+
+set(RT_CONF_TRIPLE_REGEX "default" CACHE STRING "Regex for the target triple in the generated runtime .conf file")
+
+makeConfSection(NAME ${RT_CONF_BASE_NAME}
```
the-horo, line 288: Do we want to add `-link-defaultlib-shared=false` with `-DBUILD_SHARED_LIBS=BOTH` like we discussed previously?

kinke: My thinking was that it isn't the end of the world, if the `50-target-default.conf` specifies `-link-defaultlib-shared` (when coming with shared libs only), and a cross-compilation target would inherit that default when coming with both shared/static sets. Might even be more intuitive for some users, used to that behavior for the default target; no strong opinion from my side.

the-horo: Yeah, we can leave it as is then, I too don't have a strong preference towards this.

the-horo:
> Okay Andrei, ready now from my side. Please let me know if you can live with this, or what you'd like changed.

Everything looks great. The only nit that I can struggle to find is that we started installing a `.conf` file, which comes with the section triple as `default`, which may mess up setups in which users were creating the config manually, and now their settings would be overwritten. But you did mention this in the Changelog so, if you don't consider this too much of a downside, I won't comment on it either.

the-horo:
> > The only nit that I can struggle to find is that we started installing a .conf file, which comes with the section triple as default, which may mess up setups in which users were creating the config manually, and now their settings would be overwritten.
> 
> Thx for the review! How do you propose we improve on that nit? We could bail out in ldc-build-runtime, if a non-empty install suffix was specified without also specifying RT_CONF_TRIPLE_REGEX; as I can't think of why anyone would want to install a `55-target-<suffix>` with a `default` triple.

Yes, something like that. If the users wants to install the runtime but doesn't pass a `RT_CONF_TRIPLE_REGEX` (or a `--conf-triple-regex` options if we decide to promote the switch) we can just fail early. This way, even if some setups may start failing, users will be made aware of the change and be able to decide how to fix it: either let the `55-target-<suffix>` file be installed (which is probably what most want) by passing the config triple; or specifically disable installing the config file with `-DRT_CONF_DIR=`. So, just don't silently allow the possibly incomplete config file to be installed, either keep its installation opt in, or install it by default but fail early if the triple is not specified.

And, if someone really wants the `default` triple they can still pass `-DRT_CONF_TRIPLE_REGEX=default`.

kinke:
Yeah okay then I'll add that bail-out soon-ish, makes sense to me too; the expanded help text is probably too buried by itself.

PR #4979, opened 2025-09-06
[LLVM 21] fix calls to `llvm.lifetime.start`

thewilsonator:
This mostly fixes compilation for LLVM21, except for some breakage with regards to moving `llvm.lifetime.start` to taking only and `AllocaInst` (or a `Poison`).

This both produces a call to `@llvm.lifetime.start.p0(ptr captures(none) %0)` and `%0` in this case is an `Argument` not an `AllocaInst`. 

I'm trying to figure out where those calls are coming from. Is it the ABI messing things up somehow? I don't know what is special about those functions.
```
% git grep getLLVMLifetimeStartFn
variable_lifetime.cpp:  irs.CreateCallOrInvoke(getLLVMLifetimeStartFn(),
variable_lifetime.cpp:llvm::Function *LocalVariableLifetimeAnnotator::getLLVMLifetimeStartFn() {
variable_lifetime.h:  llvm::Function *getLLVMLifetimeStartFn();
```
there is only the one use of it and 
```
% git grep lifetime_start
variable_lifetime.cpp:      &irs.module, llvm::Intrinsic::lifetime_start, allocaType);
```

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -858,6 +858,64 @@ void DtoResolveVariable(VarDeclaration *vd) {
   }
 }
 
+namespace {
+bool eval(VarDeclaration *vd, Expression *e)  {
+  
+  auto ve = e->isVarExp();
+  if (!ve) {
+    return false;
+  }
+  VarDeclaration *v = nullptr;
+  if (ve)
+    v = ve->var->isVarDeclaration();
+  return vd == v;
+}
+bool walk(VarDeclaration *vd, CommaExp *ce) {
+  IF_LOG Logger::println("ce = %s", ce->toChars());
+  if (auto ce2 = ce->e2->isCommaExp()) {
+    if (walk(vd, ce2))
+      return true;
+  }
+  if (auto ce1 = ce->e1->isCommaExp()) {
+    if (walk(vd, ce1))
+      return true;
+  }
+  if (eval(vd, ce->e2))
+    return true;
+  if (eval(vd, ce->e1))
+    return true;
+  return false;
+}
+bool varIsSret(VarDeclaration *vd, IrFunction *f) {
+  if (!f->sretArg)
+    return false;
+  auto fd = f->decl;
+  auto rets = fd->returns;
+  if (!rets)
+    return false;
+#if LDC_LLVM_VER >= 1800
+  #define startswith starts_with
+#endif
+  llvm::StringRef name = vd->ident->toChars();
+  if (name.startswith("__tmpfordtor") ||name.startswith("__sl")) {
+    return true;
+  }
+#if LDC_LLVM_VER >= 1800
+  #undef startswith
+#endif
+  for (d_size_t i = 0; i < rets->length; i++) {
+    auto rs = (*rets)[i];
+    Expression *e = rs->exp;
+    CommaExp *ce = e->isCommaExp();
+    if (ce) {
+      if (walk(vd, ce)) return true;
+    }
+    if (eval(vd, e))
+      return true;
+  }
+  return false;
+}
+}
 /******************************************************************************
  * DECLARATION EXP HELPER
  ******************************************************************************/
@@ -906,29 +964,22 @@ void DtoVarDeclaration(VarDeclaration *vd) {
 
     // We also allocate a variable for zero-sized variables, because they are technically not `null` when loaded.
     // The x86_64 ABI "loads" zero-sized function arguments, and without an allocation ASan will report an error (Github #4816).
-    llvm::Value *allocainst;
-    bool isRealAlloca = false;
     LLType *lltype = DtoType(type); // void for noreturn
     if (lltype->isVoidTy()) {
-      allocainst = getNullPtr();
-    } else if (type != vd->type) {
-      allocainst = DtoAlloca(type, vd->toChars());
-      isRealAlloca = true;
-    } else {
-      allocainst = DtoAlloca(vd, vd->toChars());
-      isRealAlloca = true;
+      irLocal->value = getNullPtr();
     }
-
-    irLocal->value = allocainst;
-
-    if (!lltype->isVoidTy())
+    else {
+      auto allocainst = type != vd->type ? DtoAlloca(type, vd->toChars())
+                                         : DtoAlloca(vd, vd->toChars());
+      irLocal->value = allocainst;
+      
       gIR->DBuilder.EmitLocalVariable(allocainst, vd);
-
-    // Lifetime annotation is only valid on alloca.
-    if (isRealAlloca) {
+      
       // The lifetime of a stack variable starts from the point it is declared
-      gIR->funcGen().localVariableLifetimeAnnotator.addLocalVariable(
-          allocainst, DtoConstUlong(size(type)));
+      if (!vd->isParameter() && !varIsSret(vd, gIR->func())) {
+        gIR->funcGen().localVariableLifetimeAnnotator.addLocalVariable(
+               allocainst, DtoConstUlong(size(type)));
+      }
     }
   }
 
--- a/gen/variable_lifetime.cpp
+++ b/gen/variable_lifetime.cpp
@@ -37,7 +37,7 @@ LocalVariableLifetimeAnnotator::LocalVariableLifetimeAnnotator(IRState &irs)
 
 void LocalVariableLifetimeAnnotator::pushScope() { scopes.emplace_back(); }
 
-void LocalVariableLifetimeAnnotator::addLocalVariable(llvm::Value *address,
+void LocalVariableLifetimeAnnotator::addLocalVariable(llvm::AllocaInst *address,
                                                       llvm::Value *size) {
   assert(address);
   assert(size);
@@ -52,8 +52,13 @@ void LocalVariableLifetimeAnnotator::addLocalVariable(llvm::Value *address,
   scopes.back().variables.emplace_back(size, address);
 
   // Emit lifetime start
-  irs.CreateCallOrInvoke(getLLVMLifetimeStartFn(), {size, address}, "",
-                         true /*nothrow*/);
+  irs.CreateCallOrInvoke(getLLVMLifetimeStartFn(),
+#if LDC_LLVM_VER >= 2100
+                         {address},
+#else
+                         {size, address},
+#endif
+                         "", true /*nothrow*/);
 }
 
 // Emits end-of-lifetime annotation for all variables in current scope.
@@ -67,8 +72,14 @@ void LocalVariableLifetimeAnnotator::popScope() {
 
     assert(address);
 
-    irs.CreateCallOrInvoke(getLLVMLifetimeEndFn(), {size, address}, "",
-                           true /*nothrow*/);
+    irs.CreateCallOrInvoke(getLLVMLifetimeEndFn(),
+#if LDC_LLVM_VER >= 2100
+                           {address},
+#else
+                           {size, address},
+#endif
+                           "", true /*nothrow*/);
+
   }
   scopes.pop_back();
 }
--- a/gen/variable_lifetime.h
+++ b/gen/variable_lifetime.h
@@ -21,12 +21,13 @@ namespace llvm {
 class Function;
 class Type;
 class Value;
+class AllocaInst;
 }
 struct IRState;
 
 struct LocalVariableLifetimeAnnotator {
   struct LocalVariableScope {
-    std::vector<std::pair<llvm::Value *, llvm::Value *>> variables;
+    std::vector<std::pair<llvm::Value *, llvm::AllocaInst *>> variables;
   };
   /// Stack of scopes, each scope can have multiple variables.
   std::vector<LocalVariableScope> scopes;
@@ -52,5 +53,5 @@ public:
   void popScope();
 
   /// Register a new local variable for lifetime annotation.
-  void addLocalVariable(llvm::Value *address, llvm::Value *size);
+  void addLocalVariable(llvm::AllocaInst *address, llvm::Value *size);
 };


gen/llvmhelpers.cpp
```diff
@@ -897,30 +897,28 @@ void DtoVarDeclaration(VarDeclaration *vd) {
...
       allocainst = DtoAlloca(type, vd->toChars());
-      isRealAlloca = true;
     } else {
       allocainst = DtoAlloca(vd, vd->toChars());
-      isRealAlloca = true;
     }
 
     irLocal->value = allocainst;
 
     if (!lltype->isVoidTy())
```
JohanEngelen: this if statement can go too

thewilsonator: done

JohanEngelen:
Did you check whether those lifetime calls are there for LLVM 20?  (in the exact same functions?)  Would help to get a standalone smaller testcase btw. Something like this: https://d.godbolt.org/z/d6f9763KE

thewilsonator:
reduced to 
```d
struct GcPolicy { }

alias CodepointSet = InversionList!GcPolicy;

struct InversionList(SP)
{
    typeof(this) opBinary(string op, U)(U )
    {
        return this;
    }

    CowArray!SP data;
}

struct CowArray(SP)
{
    this(this) { }
}

CodepointSet memoizeExpr(string expr)()
{
    if (__ctfe)
        return mixin(expr);
    alias T = typeof(mixin(expr));
    T slot;
    return slot; // This NVRO is broken
}

void wordCharacter() {
    memoizeExpr!"unicode.A | unicode.M";
}

struct unicode
{
    static opDispatch(string name)()
    {
        return loadAny(name);
    }

    static loadAny(Set=CodepointSet, C)(C)
    {
        Set set;
        return set;
    }
}
```
it generates a broken call
```llvm
; Function Attrs: uwtable(sync)
define weak_odr void @_D7package__T11memoizeExprVAyaa21_756e69636f64652e41207c20756e69636f64652e4dZQCmFNaNbNiNfZSQDl__T13InversionListTSQEi8GcPolicyZQBe(ptr noalias sret(%"package.InversionList!(GcPolicy).InversionList") align 1 %.sret_arg) #0 {
  %slot = alloca %"package.InversionList!(GcPolicy).InversionList", align 1
  %__copytmp3 = alloca %"package.InversionList!(GcPolicy).InversionList", align 1
  call void @llvm.lifetime.start.p0(ptr captures(none) %slot) #2
  call void @llvm.memset.p0.i64(ptr align 1 %slot, i8 0, i64 1, i1 false)
  call void @llvm.lifetime.start.p0(ptr captures(none) %.sret_arg) #2 ; <<<<<< HERE
  call void @llvm.memcpy.p0.p0.i64(ptr align 1 %.sret_arg, ptr align 1 %slot, i64 1, i1 false)
  call void @_D7package__T13InversionListTSQBc8GcPolicyZQBe15__fieldPostblitMFNaNbNiNeZv(ptr nonnull %.sret_arg) #0
  ret void
}
```

gen/llvmhelpers.cpp
```diff
@@ -906,27 +964,25 @@ void DtoVarDeclaration(VarDeclaration *vd) {
...
 
-    if (!lltype->isVoidTy())
-      gIR->DBuilder.EmitLocalVariable(allocainst, vd);
+    gIR->DBuilder.EmitLocalVariable(allocainst, vd);
 
-    // Lifetime annotation is only valid on alloca.
-    if (isRealAlloca) {
-      // The lifetime of a stack variable starts from the point it is declared
+    // The lifetime of a stack variable starts from the point it is declared
+    if (!vd->isParameter() && !varIsSret(vd, gIR->func())) {
```
JohanEngelen: why not simply an `allocainst->isAllocaInst()` ?

JohanEngelen: the problem is that `isRealAlloca` is apparently not what is says it is. Seems to me that the best way to fix that is to replace it with an actual check whether `allocainst` is a real alloca instruction.

thewilsonator: > why not simply an allocainst->isAllocaInst() ?

Because `isa<llvm::AllocaInst>(allocainst)` does not work, it is _always_ true because it _is_ statically an `AllocaInst`, but gets transformed into an `Argument` as an `sret` sometime later (quite how I don't know), and any call to the lifetime intrinsics are not removed in the process.

thewilsonator:
I realised what you were fixing in that commit, which was the `noreturn` issue (`runnable/noreturn1.d`). I have subsequently removed the early return and it works now.

gen/llvmhelpers.cpp
```diff
@@ -906,27 +964,25 @@ void DtoVarDeclaration(VarDeclaration *vd) {
...
     // We also allocate a variable for zero-sized variables, because they are technically not `null` when loaded.
     // The x86_64 ABI "loads" zero-sized function arguments, and without an allocation ASan will report an error (Github #4816).
-    llvm::Value *allocainst;
-    bool isRealAlloca = false;
     LLType *lltype = DtoType(type); // void for noreturn
     if (lltype->isVoidTy()) {
-      allocainst = getNullPtr();
-    } else if (type != vd->type) {
+      irLocal->value = getNullPtr();
+    }
```
kinke, line 970: The missing `else` here compared to my commit now means that this line above has no more effect for IR-void-typed vars; `irLocal->value` is immediately overwritten below with a dummy void-alloca (probably promoted to some dummy i8 storage). And now gets DI infos.

**Edit**: And additionally caused you to come up with a weird logic for the lifetime-start annotation apparently. So please just try my commit.

gen/llvmhelpers.cpp
JohanEngelen: but then, isn't that an LLVM bug?

thewilsonator: I don't know, I guess it depends where that transformation occurs, if it occurs in LDC then no, if it occurs in LLVM then maybe if there is no ability to control it from LDC.

gen/llvmhelpers.cpp
thewilsonator: Noted wr.t. DI infos, fixed that. Nope:
```c++
LLType *lltype = DtoType(type); // void for noreturn
if (lltype->isVoidTy()) {
  irLocal->value = getNullPtr();
} else {
  auto allocainst = type != vd->type ? DtoAlloca(type, vd->toChars())
                                     : DtoAlloca(vd, vd->toChars());
  irLocal->value = allocainst;
      
  gIR->DBuilder.EmitLocalVariable(allocainst, vd);
      
   // The lifetime of a stack variable starts from the point it is declared
    gIR->funcGen().localVariableLifetimeAnnotator.addLocalVariable(
           allocainst, DtoConstUlong(size(type)));
}
```
still fails with 
```
llvm.lifetime.start can only be used on alloca or poison
  call void @llvm.lifetime.start.p0(ptr captures(none) %0) #4
```

thewilsonator: Adding back in the `if`
```c++
 if (!vd->isParameter() && !varIsSret(vd, gIR->func())) {
  gIR->funcGen().localVariableLifetimeAnnotator.addLocalVariable(
             allocainst, DtoConstUlong(size(type)));
}
```
then causes it to pass.

thewilsonator: alternately if (AFAIU) the entire point of these lifetime annotations is to aid debugging, then we could restrict them to only user variables (i.e. ig more all compiler generated ones beginning with `__`, e.g. `__copytemp` which in addition to `__tmpfordtor` and `__sl` seems to cover all the bases).

kinke: Well IIRC, Phobos couldn't be compiled successfully in #4990 after cherry-picking an earlier version of this PR's commit, but applying both linked commits I linked made it work with LLVM 21 and older versions at least. I think I understood what you were after with the earlier unrelated refactoring (combining the DI and lifetime-start conditions to non-IR-void-typed vars only, without the `realAlloca` helper var and a 2nd void check), that made sense; it's just that the early return instead of an `else` branch caused skipping some stuff at the end of the function.

kinke: Also note that the ~~sret~~ NRVO case is handled just above in an earlier `if`: https://github.com/ldc-developers/ldc/blob/b5f7bb0713d9b753e6d627e8f0dd8aa8a31b840b/gen/llvmhelpers.cpp#L887-L900

That's where the lval is set to the sret argument instead of an alloca. And where no lifetime-start annotation is added.

kinke: Okay, but as your interesting reduced test case shows, that's not enough. In-place construction can replace a temporary alloca later with an sret argument, as the lvalue of a variable: https://github.com/ldc-developers/ldc/blob/b5f7bb0713d9b753e6d627e8f0dd8aa8a31b840b/gen/toir.cpp#L3002-L3015

According to the `-vv` log, this is what happens in the last `memoizeExpr` return statement:
```
ExpStatement::toIR(): <source>(15)
* D to dwarf stoppoint at line 15, column 5
* DeclarationExp::toElem: (CodepointSet slot = 0;) | T=void
* * DtoDeclarationExp: slot
* * * VarDeclaration
* * * DtoVarDeclaration(vdtype = CodepointSet)
* * * * llvm value for decl:   %2 = alloca %example.CodepointSet, align 1
* * * * expression initializer
* * * * AssignExp::toElem: slot = 0 | (CodepointSet)(CodepointSet = int)
* * * * * VarExp::toElem: slot @ CodepointSet
* * * * * * DtoSymbolAddress ('slot' of type 'CodepointSet')
* * * * * * * a normal variable
* * * * * attempting in-place construction
* * * * * * aborted due to different base types without modifiers
* * * * * IntegerExp::toElem: 0 @ int
* * * * * * IntegerExp::toConstElem: 0 @ int
* * * * * * * value = i32 0
* * * * * performing aggregate zero initialization
ReturnStatement::toIR(): <source>(16)
* D to dwarf stoppoint at line 16, column 5
* attempting in-place construction
* * success, in-place-constructing temporary
* * CommaExp::toElem: (CodepointSet __copytmp3 = (__copytmp3 = slot).this(this)();) , __copytmp3 @ CodepointSet
* * * DeclarationExp::toElem: (CodepointSet __copytmp3 = (__copytmp3 = slot).this(this)();) | T=void
* * * * DtoDeclarationExp: __copytmp3
* * * * * VarDeclaration
* * * * * DtoVarDeclaration(vdtype = CodepointSet)
* * * * * * llvm value for decl:   %3 = alloca %example.CodepointSet, align 1
* * * * * * expression initializer
* * * * * * CallExp::toElem: (__copytmp3 = slot).this(this)() @ void
* * * * * * * DotVarExp::toElem: (__copytmp3 = slot).this(this) @ void()
* * * * * * * * AssignExp::toElem: __copytmp3 = slot | (CodepointSet)(CodepointSet = CodepointSet)
* * * * * * * * * VarExp::toElem: __copytmp3 @ CodepointSet
* * * * * * * * * * DtoSymbolAddress ('__copytmp3' of type 'CodepointSet')
* * * * * * * * * * * a normal variable
* * * * * * * * * attempting in-place construction
* * * * * * * * * VarExp::toElem: slot @ CodepointSet
* * * * * * * * * * DtoSymbolAddress ('slot' of type 'CodepointSet')
* * * * * * * * * * * a normal variable
* * * * * * * * * performing normal assignment (rhs has lvalue elems = 1)
* * * * * * * * * DtoAssign()
* * * * * * * DtoCallFunction()
* * * * * * * * Building type: void()
* * * * * * * * * DtoFunctionType(void())
* * * * * * * * * * x86-64 ABI: Transforming argument types
* * * * * * * * * * Final function type: void ()
* * * * * * * * doing normal arguments
* * * * * * * * Arguments so far: (1)
* * * * * * * * *   %3 = alloca %example.CodepointSet, align 1
* * * * * * * * Function type: void()
* * * VarExp::toElem: __copytmp3 @ CodepointSet
* * * * DtoSymbolAddress ('__copytmp3' of type 'CodepointSet')
* * * * * a normal variable
```

So `__copytmp3` first gets an alloca, and then that lval is IR-replaced with the sret argument later, incl. for the lifetime-start call. Your diagnosis might be right wrt. `__copytmp3` apparently not being the frontend-NRVO variable directly; the return expression in the `if (__ctfe)` block might break it (as the NRVO var must normally be returned in *every* return statement of the function).

**Edit**: But there are always going to be remaining non-NRVO cases. So for these lifetime annotations, we might have to skip this in-place construction of sret values from non-NRVO temporaries (**edit**: oh well, the extra memcpy and different address might break code depending on RVO though...). Or scan the existing uses, to remove the lifetime intrinsic calls (possibly missing the lifetime end calls if they are pushed later on...), before replacing all uses.

**Edit2**: I think what would probably be best is to *pre*-set the lval (i.e., `getIrLocal(vd, true)->value`) of the temporary to the sret argument in `toInPlaceConstruction()`, before evaluating the comma-expression. [Incl. emitting DI there, as `DtoVarDeclaration()` wouldn't do it anymore due to `isIrLocalCreated(vd)`.] Instead of replacing all IR uses *after* a normal evaluation.

gen/llvmhelpers.cpp
thewilsonator: LDC does a `replaceAllUsesWith` on it later, see https://github.com/ldc-developers/ldc/pull/4979#discussion_r2499801544

gen/llvmhelpers.cpp
thewilsonator: >  Your diagnosis might be right wrt. __copytmp3 apparently not being the frontend-NRVO variable directly; the return expression in the if (__ctfe) block might break it (as the NRVO var must normally be returned in every return statement of the function).

Yeah, an `if(false) return ...;` also breaks NRVO.  I opened a DMD issue for this.

The removal of `isRealAlloca` was mostly to try to use the type system to fix/make sure I wasn't doing anything stupid with it, but it seems that the  `replaceAllUsesWith` is what is causing it to change type from an `AllocaInst` to `Argument`. Thank you for finding that, I thought I was going mad!

It seems to me that the easiest solution then would be to strip the calls to `@llvm.lifetime` prior to the `replaceAllUsesWith`, though I don't quite understand what you are trying to say in your `Edit2`.

Anyway I've added two of the reductions to the test suite (there is another one with `__tmpfordtor` that I ned to re-run). I'm going away for the weekend, so if you feel like implementing any of your suggestions, please do.

JohanEngelen: good find!

kinke: > so if you feel like implementing any of your suggestions, please do.

Okay, I've tried: https://github.com/ldc-developers/ldc/pull/5015. I'm pretty sure I've added that replace-all-uses-with hack, so only fair to clean it up now. :)

thewilsonator:
Superseded by https://github.com/ldc-developers/ldc/pull/5015 #5017

PR #4983, opened 2025-09-13
driver: Pick a better default C cross compiler on non-Apple Posix

the-horo:
When cross-compiling on a non-Apple Posix build machine pick the C cross compiler (when not specified through `$CC` or `-gcc=`) from:
- `<triple>-gcc`
- `<triple>-clang`
- `clang --target=<triple>`

This should be a better UX when cross-compiling with simpler build systems like `dub` or even without one at all, offering a out-of-the-box functional setup, on most systems.

The advantage to performing this check in the compiler code as opposed to hardcoding a default value in the config file is that:
1. Multiple programs can be searched, instead of just one
2. The `$CC` value continues to be respected, which keeps more advanced cross-compiling setups working (for example Meson)

----

A sample implementation of https://github.com/ldc-developers/ldc/pull/4978#issuecomment-3242341965, I haven't rigorously tested this but I'm interested in comments on the approach/fallback rules.

### DIFF
--- a/driver/cpreprocessor.cpp
+++ b/driver/cpreprocessor.cpp
@@ -30,33 +30,6 @@ const char *getPathToImportc_h(Loc loc) {
   return cached;
 }
 
-const std::string &getCC(bool isMSVC,
-                         std::vector<std::string> &additional_args) {
-  static std::string cached_cc;
-  static std::vector<std::string> cached_args;
-  if (cached_cc.empty()) {
-    std::string fallback = "cc";
-    if (isMSVC) {
-#ifdef _WIN32
-      // by default, prefer clang-cl.exe (if in PATH) over cl.exe
-      // (e.g., no echoing of source filename being preprocessed to stderr)
-      auto found = llvm::sys::findProgramByName("clang-cl.exe");
-      if (found) {
-        fallback = found.get();
-      } else {
-        fallback = "cl.exe";
-      }
-#else
-      fallback = "clang-cl";
-#endif
-    }
-    cached_cc = getGcc(cached_args, fallback.c_str());
-  }
-
-  additional_args.insert(additional_args.end(), cached_args.cbegin(), cached_args.cend());
-  return cached_cc;
-}
-
 FileName getOutputPath(Loc loc, const char *csrcfile) {
   llvm::SmallString<64> buffer;
 
@@ -96,7 +69,7 @@ FileName runCPreprocessor(FileName csrcfile, Loc loc, OutBuffer &defines) {
   FileName ipath = getOutputPath(loc, csrcfile.toChars());
 
   std::vector<std::string> args;
-  const std::string &cc = getCC(isMSVC, args);
+  const std::string &cc = getCC(args);
 
   args.push_back(isMSVC ? "/std:c11" : "-std=c11");
 
--- a/driver/tool.cpp
+++ b/driver/tool.cpp
@@ -109,10 +109,10 @@ llvm::SmallVector<std::string, 2> findCCFallback() {
   }
 
   if (isNativeBuild) {
-    error(Loc(), "Could not find C compiler `%s`", choices[0][0].c_str());
+    error(Loc(), "could not find C compiler `%s`", choices[0][0].c_str());
   } else {
-    error(Loc(), "Could not find a C cross-compiler for this cross-compilation.");
-    tip("Make sure you have a compiler like `%s` installed and set $CC or pass -gcc accordingly", choices[0][0].c_str());
+    error(Loc(), "could not find a C cross-compiler for this cross-compilation");
+    tip("make sure you have a compiler like `%s` installed and set $CC or pass -gcc accordingly", choices[0][0].c_str());
   }
 
   fatal();
--- a/driver/tool.cpp
+++ b/driver/tool.cpp
@@ -21,6 +21,12 @@
 #include "llvm/Support/Path.h"
 #include "llvm/Target/TargetMachine.h"
 
+#if LDC_LLVM_VER >= 1600
+#include "llvm/TargetParser/Host.h"
+#else
+#include "llvm/Support/Host.h"
+#endif
+
 #ifdef _WIN32
 #include <Windows.h>
 #endif
@@ -52,6 +58,69 @@ static std::string findProgramByName(llvm::StringRef name) {
 
 //////////////////////////////////////////////////////////////////////////////
 
+namespace {
+llvm::SmallVector<std::string, 2> findCCFallback() {
+  const auto &triple = *global.params.targetTriple;
+  llvm::Triple nativeTriple{ llvm::sys::getDefaultTargetTriple() };
+  const auto isNativeBuild =
+    opts::mTargetTriple.empty() || triple.isCompatibleWith(nativeTriple);
+
+  llvm::SmallVector<llvm::SmallVector<std::string, 2>, 3> choices;
+  if (triple.isWindowsMSVCEnvironment()) {
+#ifdef _WIN32
+    // by default, prefer clang-cl.exe (if in PATH) over cl.exe
+    // (e.g., no echoing of source filename being preprocessed to stderr)
+    choices = {
+      { "clang-cl.exe" },
+      { "cl.exe" },
+    };
+#else
+    choices = { { "clang-cl" } };
+#endif
+  } else if (isNativeBuild) {
+    choices = { { "cc" } };
+  } else if (triple.isOSDarwin()) {
+    // Cross-compiling for Apple is most probably done on an Apple machine
+    choices = { { "cc" } };
+  } else {
+    const auto tripleString = triple.getTriple();
+    choices = {
+      { tripleString + "-gcc" },
+      { tripleString + "-clang" },
+      { "clang", "--target=" + tripleString },
+    };
+  }
+
+
+  const auto verbose = global.params.v.verbose;
+  const auto logCross = verbose && !isNativeBuild;
+  if (logCross) message("Trying to find a C cross-compiler");
+
+  for (auto &choice : choices) {
+    auto fullPath = findProgramByName(choice[0]);
+    if (fullPath.empty()) {
+      if (logCross) message("Did not find C cross-compiler: `%s`", choice[0].c_str());
+      continue;
+    }
+
+    if (logCross) message("Found C cross-compiler: `%s`", fullPath.c_str());
+    choice[0] = fullPath;
+    return choice;
+  }
+
+  if (isNativeBuild) {
+    error(Loc(), "Could not find C compiler `%s`", choices[0][0].c_str());
+  } else {
+    error(Loc(), "Could not find a C cross-compiler for this cross-compilation.");
+    tip("Make sure you have a compiler like `%s` installed and set $CC or pass -gcc accordingly", choices[0][0].c_str());
+  }
+
+  fatal();
+}
+} // anonymous namespace
+
+////////////////////////////////////////////////////////////////////////////////
+
 std::string getProgram(const char *fallbackName,
                        const llvm::cl::opt<std::string> *opt,
                        const char *envVar) {
@@ -76,19 +145,24 @@ std::string getProgram(const char *fallbackName,
 
 ////////////////////////////////////////////////////////////////////////////////
 
-std::string getGcc(std::vector<std::string> &additional_args,
-                   const char *fallback) {
+namespace {
+std::string getCCImpl(std::vector<std::string> &additional_args) {
+  if (!gcc.empty())
+    return getProgram(nullptr, &gcc);
+
+  std::string cc = env::get("CC");
+  if (cc.empty()) {
+    auto fallback = findCCFallback();
+    additional_args.insert(additional_args.end(), fallback.begin() + 1, fallback.end());
+    return fallback[0];
+  }
+
 #ifdef _WIN32
   // spaces in $CC are to be expected on Windows
   // (e.g., `C:\Program Files\LLVM\bin\clang-cl.exe`)
-  return getProgram(fallback, &gcc, "CC");
+  return getProgram(cc.c_str(), &gcc);
 #else
   // Posix: in case $CC contains spaces split it into a command and arguments
-  std::string cc = env::get("CC");
-  if (cc.empty())
-    return getProgram(fallback, &gcc);
-
-  // $CC is set so fallback doesn't matter anymore.
   if (cc.find(' ') == cc.npos)
     return getProgram(cc.c_str(), &gcc);
 
@@ -102,6 +176,18 @@ std::string getGcc(std::vector<std::string> &additional_args,
     additional_args.emplace_back(args[i].str());
   return getProgram(args[0].str().c_str(), &gcc);
 #endif
+}
+} // anonymous namespace
+
+std::string getCC(std::vector<std::string> &additional_args) {
+  static std::string cachedResult;
+  static std::vector<std::string> cachedArgs;
+
+  if (cachedResult.empty())
+    cachedResult = getCCImpl(cachedArgs);
+
+  additional_args.insert(additional_args.end(), cachedArgs.begin(), cachedArgs.end());
+  return cachedResult;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/tests/driver/cross_cc_fallback.d
+++ b/tests/driver/cross_cc_fallback.d
@@ -1,0 +1,8 @@
+// RUN: env PATH= CC= not %ldc -mtriple aarch64-unknown-linux-gnut64 -v -defaultlib= %s 2>&1 | FileCheck %s
+
+// CHECK-DAG: aarch64-unknown-linux-gnut64-gcc
+// CHECK-DAG: aarch64-unknown-linux-gnut64-clang
+
+// UNSUPPORTED: Windows
+
+extern(C) void main () {}
--- a/.github/actions/5a-android-x86/action.yml
+++ b/.github/actions/5a-android-x86/action.yml
@@ -30,5 +30,4 @@ runs:
           "${flags[@]}" \
           ANDROID_ABI="$abi" # override the one in CROSS_CMAKE_FLAGS
 
-        # TODO: append default -gcc switch (x86_64-linux-android30-clang, i686-linux-android29-clang)
         cat install/etc/ldc2.conf/55-target-android-$arch.conf
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 - ldc2.conf can now be a directory. All the files inside it, ordered naturally, will be concatenated and treated like a big config. (#4954)
   - Running `ldc-build-runtime --installWithSuffix` now includes installing a target-specific .conf file to that directory. (#4978)
 - **Breaking change for ldc2.conf cmake generation**: The `cmake` build process now generates the `ldc2.conf` and `ldc2_install.conf` as directories. `ldc2*.conf.in` and `ADDITIONAL_DEFAULT_LDC_SWITCHES` have been removed, if you need to add switches check out `makeConfSection` in `LdcConfig.cmake`. (#4954)
+- When cross-compiling, the fallback value for the (cross) C compiler will be picked based on some heuristics.
+  The old behavior was to default to `cc`.
+  As an example, when cross-compiling for `aarch64-linux-gnu` the compilers that are checked are:
+  - `aarch64-linux-gnu-gcc`
+  - `aarch64-linux-gnu-clang`
+  - `clang --target=aarch64-linux-gnu`
 
 #### Platform support
 
--- a/driver/linker-gcc.cpp
+++ b/driver/linker-gcc.cpp
@@ -844,7 +844,7 @@ int linkObjToBinaryGcc(llvm::StringRef outputPath,
     tool = getProgram("wasm-ld", &opts::linker);
   } else {
     argsBuilder = std::make_unique<ArgsBuilder>();
-    tool = getGcc(argsBuilder->args);
+    tool = getCC(argsBuilder->args);
   }
 
   // build arguments
--- a/driver/toobj.cpp
+++ b/driver/toobj.cpp
@@ -156,7 +156,7 @@ void codegenModule(llvm::TargetMachine &Target, llvm::Module &m,
 static void assemble(const std::string &asmpath, const std::string &objpath) {
   std::vector<std::string> args;
   std::string gcc;
-  gcc = getGcc(args);
+  gcc = getCC(args);
 
   args.push_back("-O3");
   args.push_back("-c");
--- a/driver/tool.h
+++ b/driver/tool.h
@@ -25,8 +25,7 @@ namespace opts {
 extern llvm::cl::opt<std::string> linker;
 }
 
-std::string getGcc(std::vector<std::string> &additional_args,
-		   const char *fallback = "cc");
+std::string getCC(std::vector<std::string> &additional_args);
 void appendTargetArgsForGcc(std::vector<std::string> &args);
 
 std::string getProgram(const char *fallbackName,


the-horo:
I have no idea how this fails with `LLVM-15` but works with all others, given that the C++ compiler is the same
```
FAILED: [code=1] CMakeFiles/LDCShared.dir/driver/tool.cpp.o 
/usr/bin/c++ -DLDC_ENABLE_PLUGINS -DLDC_JITRT_USE_JITLINK -DLDC_LLVM_SUPPORTED_TARGET_AArch64=1 -DLDC_LLVM_SUPPORTED_TARGET_AMDGPU=1 -DLDC_LLVM_SUPPORTED_TARGET_ARM=1 -DLDC_LLVM_SUPPORTED_TARGET_AVR=1 -DLDC_LLVM_SUPPORTED_TARGET_BPF=1 -DLDC_LLVM_SUPPORTED_TARGET_Hexagon=1 -DLDC_LLVM_SUPPORTED_TARGET_Lanai=1 -DLDC_LLVM_SUPPORTED_TARGET_M68k=1 -DLDC_LLVM_SUPPORTED_TARGET_MSP430=1 -DLDC_LLVM_SUPPORTED_TARGET_Mips=1 -DLDC_LLVM_SUPPORTED_TARGET_NVPTX=1 -DLDC_LLVM_SUPPORTED_TARGET_PowerPC=1 -DLDC_LLVM_SUPPORTED_TARGET_RISCV=1 -DLDC_LLVM_SUPPORTED_TARGET_Sparc=1 -DLDC_LLVM_SUPPORTED_TARGET_SystemZ=1 -DLDC_LLVM_SUPPORTED_TARGET_VE=1 -DLDC_LLVM_SUPPORTED_TARGET_WebAssembly=1 -DLDC_LLVM_SUPPORTED_TARGET_X86=1 -DLDC_LLVM_SUPPORTED_TARGET_XCore=1 -I/home/runner/work/ldc/ldc/. -I/home/runner/work/ldc/ldc/dmd -DDMDV2 -O3 -DNDEBUG -I/usr/lib/llvm-15/include -std=c++14   -fno-exceptions -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fno-rtti  -Wall -Wextra -Wno-unused-parameter -Wno-comment -Wno-missing-field-initializers -Wno-non-virtual-dtor -Wno-pedantic -DLDC_POSIX  -DIN_LLVM -DOPAQUE_VTBLS "-DLDC_INSTALL_PREFIX=R\"(/usr/local)\"" -DLDC_LLVM_VER=1500 "-DLDC_LIBDIR_SUFFIX=R\"(64)\"" -DLDC_HOST_DigitalMars=1 -DLDC_HOST_FE_VER=2111 "-DLDC_LLVM_LIBDIR=R\"(/usr/lib/llvm-15/lib)\"" -DNDEBUG -MD -MT CMakeFiles/LDCShared.dir/driver/tool.cpp.o -MF CMakeFiles/LDCShared.dir/driver/tool.cpp.o.d -o CMakeFiles/LDCShared.dir/driver/tool.cpp.o -c /home/runner/work/ldc/ldc/driver/tool.cpp
/home/runner/work/ldc/ldc/driver/tool.cpp: In function ‘llvm::SmallVector<std::__cxx11::basic_string<char>, 2> {anonymous}::findCCFallback()’:
/home/runner/work/ldc/ldc/driver/tool.cpp:68:14: error: missing template arguments before ‘choices’
   68 |   std::array choices = {
      |              ^~~~~~~
/home/runner/work/ldc/ldc/driver/tool.cpp:77:22: error: ‘choices’ was not declared in this scope; did you mean ‘choice’?
   77 |   for (auto choice : choices) {
      |                      ^~~~~~~
      |                      choice
/home/runner/work/ldc/ldc/driver/tool.cpp:90:97: error: ‘choices’ was not declared in this scope
   90 |   tip("Make sure you have a compiler like `%s` installed and set $CC or pass -gcc accordingly", choices[0][0].c_str());
      |                                                                                                 ^~~~~~~
```

driver/tool.cpp
```diff
@@ -52,6 +53,46 @@ static std::string findProgramByName(llvm::StringRef name) {
 
 //////////////////////////////////////////////////////////////////////////////
 
+namespace {
+llvm::SmallVector<std::string, 2> findCCFallback() {
+  // Non-cross build => use "cc"
+  if (opts::mTargetTriple.empty())
+    return { "cc" };
+
+#if defined(__APPLE__) || !defined(LDC_POSIX)
```
kinke: This shouldn't be based on the host, but the target. E.g., when cross-compiling to Linux on Windows, we should still try to look up the according cross-gcc/clang.

I suggest we don't do anything (i.e., just using `cc`) for Apple targets. AFAIK, there's no *official* way to cross-compile to Apple targets on non-Apple hosts. And we already forward the triple to Apple clang in https://github.com/ldc-developers/ldc/blob/83362a71810a4c2d7ddcf7fec030aa2ca5bc7467/driver/tool.cpp#L115-L120

For wasm, we don't go through the C compiler for linking, using `wasm-ld` (or builtin lld via `-link-internally`) directly instead. So most likely doesn't need anything at the moment either. [We could default to Emscripten's `emcc` in the future.]

kinke: And for Windows (where we don't link through the C compiler either), the logic is currently here: https://github.com/ldc-developers/ldc/blob/83362a71810a4c2d7ddcf7fec030aa2ca5bc7467/driver/cpreprocessor.cpp#L40-L51

driver/tool.cpp
```diff
@@ -52,6 +53,46 @@ static std::string findProgramByName(llvm::StringRef name) {
...
+      continue;
+    }
+
+    if (verbose) message("Found C cross-compiler: `%s`", fullPath.c_str());
+    choice[0] = fullPath;
+    return choice;
+  }
+
+  error(Loc(), "Could not find a C cross-compiler for this cross-compilation.");
+  tip("Make sure you have a compiler like `%s` installed and set $CC or pass -gcc accordingly", choices[0][0].c_str());
```
kinke: I guess this is better than defaulting to a `cc`. Though not every `-mtriple` necessarily means that we're actually cross-compiling; e.g., one might target an older FreeBSD OS version than the default one via `-mtriple`. But at least for FreeBSD, the logic will probably use the `clang --target=…` option, which might just work.

driver/tool.cpp
the-horo: > This shouldn't be based on the host, but the target. E.g., when cross-compiling to Linux on Windows, we should still try to look up the according cross-gcc/clang.

Oh, it seems we're using different terms to refer to things:
- the build machine, I use build, you seem to be using host
- the machine that will run the produced exe, I use host, you use target

Not an issue per se, just something to note in case we're confused about the other's words.

> I suggest we don't do anything (i.e., just using `cc`) for Apple targets. AFAIK, there's no _official_ way to cross-compile to Apple targets on non-Apple hosts. And we already forward the triple to Apple clang in

Just a possibility but we could move the `-target` argument in the default `CC` logic. This would allow someone to, technically, overwrite it by passing a different `-gcc`. I'm not familiar with Apple stuff but from your description this sounds pointless because a user will always have to pass the `-target` themselves so the current approach of the compiler doing it sounds better.

> And for Windows (where we don't link through the C compiler either), the logic is currently here:

Speaking strictly about this code, I think this can be safely moved to the default `CC` logic that I've added.

I haven't done windows cross-compilation before so I've tried to play around a little bit:
```
[happy@happypc] /tmp/x % cat a.d
import core.stdc.stdio;

extern(C) void mainCRTStartup () {
	puts("Hello, world!");
}
[happy@happypc] /tmp/x % ldc2 -mtriple=x86_64-windows-msvc  -L-L/usr/lib/mingw64-toolchain/x86_64-w64-mingw32/lib -defaultlib= a.d -v -platformlib=libkernel32.a,libuser32.a,libgdi32.a,libwinspool.a,libshell32.a,libole32.a,liboleaut32.a,libuuid.a,libcomdlg32.a,libadvapi32.a -L/subsystem:console -mscrtlib=libmsvcrt.a | grep lld-link
/usr/lib/llvm/20/bin/lld-link /NOLOGO /OPT:REF /OPT:ICF /DEFAULTLIB:libmsvcrt.a /OUT:a.exe a.obj /LIBPATH:/usr/lib/mingw64-toolchain/x86_64-w64-mingw32/lib /subsystem:console /LIBPATH:/usr/lib/ldc2/1.40/lib64 /LIBPATH:/usr/lib/ldc2/1.40/lib64/mingw libkernel32.a libuser32.a libgdi32.a libwinspool.a libshell32.a libole32.a liboleaut32.a libuuid.a libcomdlg32.a libadvapi32.a
[happy@happypc] /tmp/x % WINEDEBUG=-all wine a.exe
Hello, world!
```

which is a bit of a mouthful and it seems to be caused by my `mingw` installation having the libraries called like `libkernel32.a` instead of the expected `kernel32.lib`:
```
[happy@happypc] /tmp/x % ldc2 -mtriple=x86_64-windows-msvc  -L-L/usr/lib/mingw64-toolchain/x86_64-w64-mingw32/lib -defaultlib= a.d -v -L/subsystem:console | grep lld-link
/usr/lib/llvm/20/bin/lld-link /NOLOGO /OPT:REF /OPT:ICF /DEFAULTLIB:vcruntime140 /OUT:a.exe a.obj /LIBPATH:/usr/lib/mingw64-toolchain/x86_64-w64-mingw32/lib /subsystem:console /LIBPATH:/usr/lib/ldc2/1.40/lib64 /LIBPATH:/usr/lib/ldc2/1.40/lib64/mingw kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib oldnames.lib legacy_stdio_definitions.lib
lld-link: error: could not open 'kernel32.lib': No such file or directory
lld-link: error: could not open 'user32.lib': No such file or directory
lld-link: error: could not open 'gdi32.lib': No such file or directory
lld-link: error: could not open 'winspool.lib': No such file or directory
lld-link: error: could not open 'shell32.lib': No such file or directory
lld-link: error: could not open 'ole32.lib': No such file or directory
lld-link: error: could not open 'oleaut32.lib': No such file or directory
lld-link: error: could not open 'uuid.lib': No such file or directory
lld-link: error: could not open 'comdlg32.lib': No such file or directory
lld-link: error: could not open 'advapi32.lib': No such file or directory
lld-link: error: could not open 'oldnames.lib': No such file or directory
lld-link: error: could not open 'legacy_stdio_definitions.lib': No such file or directory
Error: /usr/lib/llvm/20/bin/lld-link failed with status: 1
[happy@happypc] /tmp/x % ls /usr/lib/mingw64-toolchain/x86_64-w64-mingw32/lib/libkernel32.a
/usr/lib/mingw64-toolchain/x86_64-w64-mingw32/lib/libkernel32.a
```

I also have installed a windows cross-gcc:
```
[happy@happypc] /tmp/x % x86_64-w64-mingw32-gcc -v
Using built-in specs.
COLLECT_GCC=x86_64-w64-mingw32-gcc
COLLECT_LTO_WRAPPER=/usr/lib/mingw64-toolchain/libexec/gcc/x86_64-w64-mingw32/15/lto-wrapper
Target: x86_64-w64-mingw32
Configured with: /var/tmp/portage/dev-util/mingw64-toolchain-13.0.0/work/gcc/configure --build=x86_64-pc-linux-gnu --target=x86_64-w64-mingw32 --docdir=/.skip --infodir=/.skip --mandir=/.skip MAKEINFO=: --prefix=/usr/lib/mingw64-toolchain --host=x86_64-pc-linux-gnu --disable-bootstrap --disable-cet --disable-gcov --disable-gomp --disable-libquadmath --disable-libsanitizer --disable-libssp --disable-libvtv --disable-shared --disable-werror --with-gcc-major-version-only --with-system-zlib --without-isl --without-zstd --enable-languages=c,c++ --enable-threads=posix --with-sysroot=/usr/lib/mingw64-toolchain --with-build-sysroot=/var/tmp/portage/dev-util/mingw64-toolchain-13.0.0/temp/root/usr/lib/mingw64-toolchain
Thread model: posix
Supported LTO compression algorithms: zlib
gcc version 15.1.0 (GCC)
[happy@happypc] /tmp/x % cat a.d
extern(C) void puts(const char *);

extern(C) void main () {
	puts("Hello, world!");
}
[happy@happypc] /tmp/x % ldc2 -mtriple=x86_64-w64-mingw32  -defaultlib= a.d -v -gcc=x86_64-w64-mingw32-gcc | grep gcc
/usr/bin/x86_64-w64-mingw32-gcc a.o -o a.exe -L/usr/lib/ldc2/1.40/lib64 -Wl,-rpath,/usr/lib/ldc2/1.40/lib64 -Wl,--gc-sections -lws2_32 -m64
[happy@happypc] /tmp/x % WINEDEBUG=-all wine a.exe
Hello, world!
```

One issue that I've found is that druntime seem to be lacking the bindings for `mingw`:
```
[happy@happypc] /tmp/x % cat a.d
import core.stdc.stdio;

extern(C) void main () {
	puts("Hello, world!");
}
[happy@happypc] /tmp/x % ldc2 -mtriple=x86_64-w64-mingw32  -defaultlib= a.d -gcc=x86_64-w64-mingw32-gcc
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(186): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(188): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(194): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(196): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(215): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(217): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(221): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(223): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(243): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/wchar_.d(252): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1204): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1209): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1215): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1219): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1221): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1221): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1224): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1226): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1234): Error: undefined identifier `FILE`
/usr/lib/ldc2/1.40/include/d/core/stdc/stdio.d(1240): Error: undefined identifier `FILE`
```

kinke: Wrt. Windows, note that druntime isn't compatible with MinGW (last time I checked) - and that's why the official Windows packages ship MinGW-**based** libraries, where all MinGW-specifics have been patched out: https://github.com/ldc-developers/mingw-w64-libs

> Speaking strictly about this code, I think this can be safely moved to the default CC logic that I've added.

Yeah; just note that on Windows, the default `cl.exe` (Microsoft's C(++) compiler) might not be in PATH yet at that point, since we need to set up environment variables for the MS tools (`MsvcEnvironmentScope`).

driver/tool.cpp
the-horo: Yeah, but now I'm wondering about things like `dub`. Wouldn't passing a redundant `--arch x86_64-pc-linux-gnu` switch take the cross-compilation branch and change the C compiler default from `cc` to `x86_64-pc-linux-gnu-gcc`. This doesn't sound like an acceptable side-effect.

Maybe it's worth to still try `cc` last and print the tip only with `-v`.

driver/tool.cpp
```diff
@@ -52,6 +53,60 @@ static std::string findProgramByName(llvm::StringRef name) {
...
+  const auto &triple = *global.params.targetTriple;
+  // TODO: find a better way to determine this
+  const auto isNativeBuild = opts::mTargetTriple.empty();
+
+  llvm::SmallVector<llvm::SmallVector<std::string, 2>, 3> choices;
+  if (triple.isWindowsMSVCEnvironment()) {
+#ifdef _WIN32
+    choices = {
+      { "clang-cl.exe" },
+      { "cl.exe" },
```
kinke, line 75: Okay, this should actually work, because of the preceding `windows::MsvcEnvironmentScope` in `runCPreprocessor()` - `cl.exe` should be in PATH by then, so that the later `findProgramByName()` succeeds.

driver/tool.cpp
```diff
@@ -52,6 +53,60 @@ static std::string findProgramByName(llvm::StringRef name) {
...
+llvm::SmallVector<std::string, 2> findCCFallback() {
+  const auto &triple = *global.params.targetTriple;
+  // TODO: find a better way to determine this
+  const auto isNativeBuild = opts::mTargetTriple.empty();
+
+  llvm::SmallVector<llvm::SmallVector<std::string, 2>, 3> choices;
+  if (triple.isWindowsMSVCEnvironment()) {
+#ifdef _WIN32
+    choices = {
+      { "clang-cl.exe" },
```
kinke, line 74: Please copy over the comment as to why we prefer clang-cl (no stderr output breaking most importC tests, and on Windows we only need the C compiler for preprocessing, not for linking).

driver/cpreprocessor.cpp
```diff
@@ -30,33 +30,6 @@ const char *getPathToImportc_h(Loc loc) {
...
-      }
-#else
-      fallback = "clang-cl";
-#endif
-    }
-    cached_cc = getGcc(cached_args, fallback.c_str());
-  }
-
-  additional_args.insert(additional_args.end(), cached_args.cbegin(), cached_args.cend());
-  return cached_cc;
```
kinke, line 57: We use a cache here, as we might be preprocessing multiple files during a run (contrary to an at most single linker run). So moving the cache to `getCC()` would be nice.

tests/driver/cross_cc_fallback.d
```diff
@@ -0,0 +1,8 @@
+// RUN: env PATH= CC= not %ldc -mtriple aarch64-unknown-linux-gnu -v -defaultlib= %s 2>&1 | FileCheck %s
+
+// CHECK-DAG: aarch64-unknown-linux-gnu-gcc
+// CHECK-DAG: aarch64-unknown-linux-gnu-clang
+
+// UNSUPPORTED: Windows, Darwin
```
kinke: AFAICT, the only problem is that there's no `env` tool on Windows. Assuming it exists on Mac, this test should work there too.

driver/tool.cpp
kinke: I'm not so much worried about `dub --target` - I doubt many people know about this and use it, and it's only supported by LDC. I'm rather mildly 'worried' about using `-mtriple` for non-cross compiles, such as to specify a non-default OS/env version, or some subarch - cases in which we likely don't find a corresponding C compiler.

So in your current version, people would have to specify `-gcc` to make it work in these rather exotic cases, after seeing the error and hints. I think that's definitely better than a failing `cc` run (the linker errors aren't always very good) - if it does fail. :) A viable trade-off might be to fall back to `cc`, but emitting the hints (without explicit `-v`).

tests/driver/cross_cc_fallback.d
the-horo: That's right, the `Darwin` exclusion was accidentally left from when the detection involved the build system arch, not the `--mtriple` arch.

driver/tool.cpp
the-horo: Yes, I wasn't sure if I wanted to move it inside `findCCFallback` because it looked like that setup included environment needed for compilation & preprocessing and, at the end of the function, they would be rollback.

driver/tool.cpp
the-horo: Ok, I've changed the native build detection to use [Triple::isCompatibleWith](https://llvm.org/doxygen/classllvm_1_1Triple.html#ae86f9f850769a4896b883b0da1322a51) and that solves the issue for pointless `--mtriple` as well as for different OS versions (e.g. `freebsd` vs `freebsd16`). It won't work with a different vendor, though maybe that's fine as we can just let people use/not use the vendor. If their C cross-compiler is `aarch64-linux-gnu-gcc` just specify `aarch64-linux-gnu` and if gcc is `aarch64-unknown-linux-gnu-gcc` specify `aarch64-unknown-linux-gnu`, so just go with whatever the distro defaults are. I didn't get to test `aarch64-linux-gnu` since my cross compilers are all vendor aware so I just hope that it works.

kinke: Hmm, I hate it when you need to look at the actual implementation to see the semantics - no wonder, with that convoluted logic. ;) So AFAICT, for Android targets, a difference in the environment version (`android21` vs. `android29` - their 'API levels') would be considered 'compatible' - but the NDK provides separate clang wrappers for each level, basically just injecting a `--target=…` [**edit**: for `clang` in the same dir as the wrapper]. No idea if that's a real problem, but e.g. native TLS was only added in API levels 29/30, so I think it might.

kinke: Hmm, Android's NDK seems a bit weird - there's e.g. no `cc` in `android-ndk-r27c/toolchains/llvm/prebuilt/linux-x86_64/bin`, but a `clang`, and all the wrappers like `aarch64-linux-android21-clang`. Maybe we should special-case Android analogous to Darwin, using `clang` by default (like `cc` on Darwin), and always specifying `--target=…` in `appendTargetArgsForGcc()`.

[**Edit**: We don't support older gcc-based NDKs anymore - we require native TLS nowadays, which requires more recent clang-based NDKs.]

the-horo: Wouldn't the `clang --target` case still be handled by the default cross target logic? If you compile for `aarch64-linux-android21` or `aarch64-linux-android29` on a linux build machine you would still hit the `aarch64-linux-android??-clang` or the `clang --target aarch64-linux-android??` fallbacks. Since the compatibility check is between the build arch and the target arch we would be hitting conflicts only if one were to cross compile for a different android host, from an android build machine.

kinke: Yep, sorry for not being more specific - I was referring to running LDC natively on Android, and 'cross'-compiling to a non-default API level (such as a newer one). I know that some guys do really run LDC on Android. :)

**Edit**: And btw, ridiculously, Google don't provide native NDKs. So AFAIK, people use apps like Termux to get a Linux terminal with a package manager. So there might be no real default C compiler environment on native Android. But I'd suspect any candidate would have to be similar to an NDK at least.

the-horo: Alright, I get you. So what default do we want? `clang --target=<android_triple>`, regardless of native/cross build? Or, like you said, make it the same as Apple and move the `--target` switch out of the cc fallback and into `appendTargetArgsForGcc`.

kinke: I don't really know. The danger with generic `clang --target=…` is that one might be using a non-Android one, such as the `C:\Program Files\LLVM\bin\clang.exe` on Windows, where the error message is presumably pretty bad, compared to us warning/erroring about not finding a corresponding `…-androidNN-clang`. And I suspect that risk is considerably higher than running into the mentioned problem on native Android. So maybe just keep things as they are. :)

driver/tool.cpp
```diff
@@ -20,6 +20,13 @@
 #include "llvm/Support/MemoryBuffer.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Target/TargetMachine.h"
+#include <string>
```
kinke: I doubt we really need this explicitly. - System includes come last, after the LLVM imports.

driver/tool.cpp
kinke: I think it's fine as-is in that regard. But what's more problematic is that native Windows builds would now have `-v` output and potential errors mentioning a 'C cross-compiler', while that's not the case in this `#ifdef _WIN32` block. So analogous to the Posix-native-build case, we'd need an early-return / common helper function / or simply using `choices` for the Posix-native-build case too, and then adapting the strings based on `isNativeBuild`.

driver/tool.cpp
kinke: Yeah I'm satisfied with the are-triples-compatible check.

kinke:
Looks pretty ready to me, just 2 nits. - If this lands after https://github.com/ldc-developers/ldc/pull/4978, you could remove the FIXME comment for the Android x86[_64] .confs, no need to inject `-gcc` anymore.

kinke:
Cheers, LGTM, modulo error strings starting in uppercase. Let me fix the changelog merge conflict and convert to lowercase.

PR #4987, opened 2025-09-17
Fix #4986 : wrong type for return value

### LINKED ISSUE #4986: Another ICE
I reduced it to this
```d
struct Test
{
    this(ref inout typeof(this) rhs) inout pure {}
    const(char)[] toString() const pure => null;
    alias toString this;
}

const(char)[] test(ref Test t)
{
    return t;
}
```

```
Function return type does not match operand type of return inst!
  ret ptr %2, !dbg !150940
COMPILED : { i64, ptr }LLVM error : Broken module found, compilation aborted!
Exception Code: 0xC000001D
 #0 0x00007ff621b28ea6 (C:\dev\D\ldc2-1.41.0-windows-multilib\bin\ldc2.exe+0x488ea6)
 #1 0x00007ff6252289b1 (C:\dev\D\ldc2-1.41.0-windows-multilib\bin\ldc2.exe+0x3b889b1)
...
```

thewilsonator:


thewilsonator:
This is good to go now.

cc @TurkeyMan

TurkeyMan:
I thought last time we discussed this, it was resolved that it was a DMD bug? There's a ticket in the DMD bug tracker somewhere to deal with issues something like this iirc?

kinke:
Please revert; this is a hack around a symptom.

PR #4995, opened 2025-10-20
[irstate] add addrspace to string literals

thewilsonator:
This is required for Vulkan support where these live in a different address space than the default

### DIFF
--- a/gen/irstate.h
+++ b/gen/irstate.h
@@ -27,6 +27,7 @@
 #include <set>
 #include <sstream>
 #include <vector>
+#include <optional>
 
 namespace llvm {
 class LLVMContext;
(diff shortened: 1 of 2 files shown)


thewilsonator:
`lit` seems to fail to be installed from home-brew on macOS, but that `Vanilla LLVM / Ubuntu 24.04, LLVM 15` failure I don't understand at all, `#include <optional>` is right there, and it compiles on the `Ubuntu rolling x64 shared-libs-only` test, so I didn't typo anything.

gen/irstate.h
```diff
@@ -27,6 +27,7 @@
 #include <set>
 #include <sstream>
 #include <vector>
+#include <optional>
```
kubo39, line 30: Looking at `gen/dcompute/druntime.h`, this should be like:

```cpp
#if LDC_LLVM_VER < 1700
#include "llvm/ADT/Optional.h"
#else
#include <optional>
namespace llvm {
template <typename T> using Optional = std::optional<T>;
}
#endif
```

kubo39: or, maybe?

```
#if LDC_LLVM_VER >= 1600
#include <optional>
#endif
```

PR #4996, opened 2025-10-20
Fix #4993: allow to cast between fat values

### LINKED ISSUE #4993: Internal Compiler Error: Invalid struct cast from Foo to ubyte[1]
```D
struct Foo {}
void main() {
    Foo f;
    auto x = cast(ubyte[Foo.sizeof])f;
}
```
Successfully compiles on all versions of DMD available on run.dlang.io, but fails with LDC 1.41.0 with an ICE. Occurs both with and without fields in Foo.

The same issue occurs when trying to cast a T to a T[1]:
```D
struct Foo {}
void main() {
    Foo f;
    auto x = cast(Foo[1])f;
}
```

kubo39:


gen/llvmhelpers.cpp
```diff
@@ -673,6 +673,14 @@ DValue *DtoCastStruct(Loc loc, DValue *val, Type *to) {
     return new DLValue(to, lval);
   }
 
+  // Bugzlla 3133: A cast between fat values is possible only when the sizes match.
```
thewilsonator: I'd probably link to #4993 rather than a bugzilla here

tests/compilable/gh4993.d
```diff
@@ -0,0 +1,16 @@
+// RUN: %ldc -c %s
+
```
thewilsonator, line 2: I'd also link to #4993 here if you are going to call the file this and not `gh4993`

gen/llvmhelpers.cpp
kubo39: thanks, updated.

tests/compilable/gh4993.d
kubo39: renamed.

PR #5011, opened 2025-10-30
Add `-fdebug-prefix-map` command line argument

yanok:
Debug info stores the path to the directory where the source is located and that causes problem with getting reproducible build results if ran from different work directories.

Traditionally people solve that by controlling the build environment, such that sources are always at the fixed location. But this can sometimes be undesirable (for example with bazel, where each build action runs in a sandbox with a hash in the path name), so clang supports the `-fdebug-prefix-map` option to control this.

This PR implement the same option in LDC.

### DIFF
--- a/gen/dibuilder.cpp
+++ b/gen/dibuilder.cpp
@@ -218,6 +218,15 @@ void DIBuilder::SetValue(Loc loc, llvm::Value *value,
                                    IR->scopebb());
 }
 
+std::string DIBuilder::remapDIPath(llvm::StringRef path) {
+  for (const auto &[from, to] : opts::debugPrefixMap) {
+    if (path.starts_with(from)) {
+      return to + path.substr(from.size()).str();
+    }
+  }
+  return std::string(path);
+}
+
 DIFile DIBuilder::CreateFile(const char *filename) {
   if (!filename)
     filename = IR->dmodule->srcfile.toChars();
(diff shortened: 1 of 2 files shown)


gen/dibuilder.cpp
```diff
@@ -218,6 +218,15 @@ void DIBuilder::SetValue(Loc loc, llvm::Value *value,
                                    IR->scopebb());
 }
 
+std::string DIBuilder::remapDIPath(llvm::StringRef path) {
+  for (const auto &[from, to] : opts::debugPrefixMap) {
+    if (path.starts_with(from)) {
```
JohanEngelen, line 223: `starts_with` looks dangerous, what if I map "some/path="other/thing", does that mean that "some/path_more" also matches and is remapped?

JohanEngelen: (could be checked in a test case)

JohanEngelen:
Don't forget to add testcase(s)

gen/dibuilder.cpp
yanok: Yes, but this is what clang does. I guess in the most (only?) use cases this is going to be some "root" path, so we don't expect sources from `root_some` to be passed to the compiler. Cautious people can do `some/path/=other/thing/`. It's easy to enforce this in the code, but I don't want to diverge from clang...

PR #5014, opened 2025-11-04
Fix #4779 - C++ classes need TypeInfo

### LINKED ISSUE #4779: C++ classes need TypeInfo
I've finally got around to using C++ classes in -betterC cross the shared library boundary.

As of ldc 1.38.0, what I've found is that the TypeInfo/Interface do not get emitted (expected) from within the shared library.
However, on the executable side that is full D, it'll error given that it doesn't exist.
I have not had this as an issue for any other TypeInfo.

I was not able to use my usual tricks via inline assembly to emit the data due to LLVM's type checking (I really wish I had a way to disable that for a symbol).
In the end, the workaround I came up with was to template the entire class hierarchy (not good).

On that note, wouldn't the external path switch have helped here? Since if it was external, it may not have been compiled by the D compiler.

rikkimax:
I have literally no idea if this is even going to work.

I only managed to find the location that needed changing thanks to GH's web copilot (and promptly burned through an entire month of free credits lol).

### DIFF
--- a/gen/typinf.cpp
+++ b/gen/typinf.cpp
@@ -506,8 +506,19 @@ class DeclareOrDefineVisitor : public Visitor {
     auto cd = decl->tinfo->isTypeClass()->sym;
     DtoResolveClass(cd);
 
+    auto irclass = getIrAggr(cd, true);
     IrGlobal *irg = getIrGlobal(decl, true);
-    irg->value = getIrAggr(cd)->getClassInfoSymbol();
+    
+    auto ti = irclass->getClassInfoSymbol();
+    irg->value = ti;
+    
+    // check if the definition can be elided
+    if (irclass->suppressTypeInfo()) {
+        return;
+    }
+    
+    irclass->getClassInfoSymbol(true);
+    ti->setLinkage(TYPEINFO_LINKAGE_TYPE); // override
   }
 
   // Build all other TypeInfos.
--- a/ir/iraggr.cpp
+++ b/ir/iraggr.cpp
@@ -24,6 +24,7 @@
 #include "gen/mangling.h"
 #include "gen/pragma.h"
 #include "gen/tollvm.h"
+#include "gen/linkage.h"
 #include "ir/irdsymbol.h"
 #include "ir/irtypeclass.h"
 #include "ir/irtypestruct.h"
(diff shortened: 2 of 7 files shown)


rikkimax:
Ok yeah I'm lost.
```
ld: error: duplicate symbol: core.internal.gc.impl.proto.gc.ProtoGC.__interfaceInfos
>>> defined at cast_.d
>>>            runtime/objects-debug-shared/core/internal/cast_.o:(core.internal.gc.impl.proto.gc.ProtoGC.__interfaceInfos)
>>> defined at gc.d
>>>            runtime/objects-debug-shared/core/internal/gc/impl/proto/gc.o:(.data.rel.ro._D4core8internal2gc4impl5protoQo7ProtoGC16__interfaceInfosZ+0x0)
```

rikkimax:
Added the interfaces.

Also checked that the macos failure isn't caused by me, it's in other PR's.

rikkimax:
> You're now switching to the emit-into-every-referencing-object-file TI emission scheme for _all_ classes, and all symbols they reference. Not restricting it to C++ classes, nor as the hacky 'solution' this is for interoperating with separate betterC binaries not exporting any TypeInfos. So this would affect OO code emission quite considerably, and badly as in bad code bloat.

I can gate it at the end, although I am hoping by not gating it that it'll maybe find some cases with the help of the D classes in testsuite that I won't think of.

And yes, I am aware of a testsuite change for D classes that's easily undone.

> Have you considered making your betterC libraries a bit more fine-grained, disabling all the rest except for leaving TypeInfos enabled? I don't expect it would work out of the box, presumably needing some LDC adaptations, but it might be a better solution for the special problem in #4779.

That would be a valid solution for *me*, and so is the one I ended up with by templating classes.

However the point of ProjectSidero is three fold:
1. To discover under-the-hood issues
2. Fix issues found
3. Learn about standard library creation

By implementing what amounts to a standard library, I can verify if it's a problem worth solving.
If I can do it without a fix reasonably well, so can others.

In this case I have an example of where somebody tried to mix full D and -betterC together with C++ classes and failed due to linker issues. And that is Andrei with dmd.common.

Given that I could confirm that it's a problem worth solving, and I have Andrei as an example, I am confident this needs fixing.

rikkimax:
Now that I am thinking about gating beyond simply keeping it an option for later:

TypeInfo is really only needed for C++ classes in the following cases (that I can think of):

- Child class
- typeid
- new, due to templated hooks this is now typeid problem
- AA's which come under new
- Reference by another TypeInfo

If it were gated upon reference by another TypeInfo and typeid, that may make it PAYG enough to remove the code bloat issues. Would also work for structs.

I suspect implementing it may be more than what I can do, though ;)

PR #5022, opened 2025-11-12
[LLVM22] Add test case for `lifetime.start` not appearing with `sret`

thewilsonator:


calvin2021y:
add `LDC_LLVM_VER`  to keep the old code for old LLVM versions will fix this?

kinke:
> I think your fix https://github.com/ldc-developers/ldc/pull/5015 was either wrong for old LLVM versions or it was already broken for old LLVM versions.

I very much doubt so - we surely have many many RVO cases as part of the testsuite.

I'm not sure an extra test case (for NRVO and RVO) for these lifetime annotations are required. I mean you seem to have hit it quite immediately with LLVM master, without extra test case. And if you really want to cover this LLVM 22 limitation to allocas, you don't really have to check for elided lifetime annotations either - we'd most likely get LLVM errors/assertions [**edit**: well, before #5015] as soon as there are (N)RVO cases in an existing `-femit-local-var-lifetime` test file, without IR checks.

thewilsonator:
> I'm not sure an extra test case (for NRVO and RVO) for these lifetime annotations are required.

Well these tests fail on old LLVMs so...

kinke:
The actual problem is this:
```D
extern(C): // this!

struct S {
    this(this) {}
}

S sretReturn_NRVO() {
    S ret;
    return ret;
}

S sretReturn_RVO() {
    return S();
}
```

This last compiled with LDC v1.27.1 according to godbolt, v1.28.1 already failed. The problem is that with `extern(C)`, we don't use sret, but return a pointer (to the stack-allocated temporary!) directly, which is utterly wrong. Coupled with the `IndirectByvalRewrite`, we then end up adding the noalias and nocapture attributes to the return value, which LLVM complains about.

As C doesn't have non-PODs but the frontend happily accepts such signatures apparently, we'll probably have to deal with such non-POD return types as we do for `extern(D)`, using an `sret` parameter.

calvin2021y:
Do we really need `non-POD struct types` for extern(C)?

PR #4990, opened 2025-10-02
Bump LDC-LLVM to v21.1.2

kinke:

PR #5026, opened 2025-12-03
[LLVM21] Support LLVM version 21.1

d-musique:
Define `LLVM_version` to allow building with llvm 21.1


d-musique:
>but see #4990

Ok I'm closing, given that someone is working on this.
I have built, and bootstrapped ldc successfully with LLVM 21.1.7.
However I then experienced some crashes which happen exclusively under -O, something also mentioned the other PR.

PR #5032, opened 2025-12-16
Revise DtoDeclarationExp()

kinke:
After comparing this with DMD's `Dsymbol_toElem()`, incl. checking the `VarDeclaration` overloads of the codegen visitors.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -211,15 +211,6 @@ static void write_struct_literal(Loc loc, LLValue *mem, StructDeclaration *sd,
 }
 
 namespace {
-void pushVarDtorCleanup(IRState *p, VarDeclaration *vd) {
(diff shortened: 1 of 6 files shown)


gen/toir.cpp
```diff
@@ -318,13 +309,7 @@ class ToElemVisitor : public Visitor {
     auto &PGO = gIR->funcGen().pgo;
     PGO.setCurrentStmt(e);
 
-    result = DtoDeclarationExp(e->declaration);
-
-    if (auto vd = e->declaration->isVarDeclaration()) {
-      if (!vd->isDataseg() && vd->needsScopeDtor()) {
-        pushVarDtorCleanup(p, vd);
```
kinke, line 325: This check was insufficient, as it turned out during intermediate refactorings - IIRC, missing temporaries inside an outer `align(N)` declaration (as `e->declaration`).

Issue #4294, opened 2023-01-03
`multibyteMulAdd` "is already defined" with `-flto=full` on i686


CyberShadow:
Trying to port btdu to i686 I ran into this. `ldc-build-runtime` is invoked with `--dFlags="-flto=full"` and later `ldc2` with `-flto=full` and `-static`. This happens (but does not happen on x86_64):

```
/usr/bin/i686-linux-gnu-gcc btdu-static-i686.o -static -o btdu-static-i686 -Wl,-plugin,/tmp/ldc2-1.30.0-linux-x86_64/lib/LLVMgold-ldc.so -Wl,-plugin-opt=mcpu=pentium4 -Wl,-plugin-opt=O3 -Wl,-plugin-opt=-function-sections -Wl,-plugin-opt=-data-sections -Lrelease -l:libtermcap.a -l:libncursesw.a -l:libtinfo.a -l:libz.a -L/tmp/ldc-build-runtime.tmp/lib -lphobos2-ldc -ldruntime-ldc -Wl,--gc-sections -lrt -ldl -lpthread -lm -m32 
/usr/lib/gcc-cross/i686-linux-gnu/8/../../../../i686-linux-gnu/bin/ld: error: LLVM gold plugin: <inline asm>:876:1: symbol '.L_D3std8internal4math10biguintx86__T15multibyteMulAddVai43ZQxFNaNbNfAkxAkkkZk_L1' is already defined
.L_D3std8internal4math10biguintx86__T15multibyteMulAddVai43ZQxFNaNbNfAkxAkkkZk_L1:      mull 36(%esp)
^

collect2: error: ld returned 1 exit status
Error: /usr/bin/i686-linux-gnu-gcc failed with status: 1
DOCKER=podman ci/build-docker.sh i686 x86_64 aarch64  58.45s user 12.43s system 77% cpu 1:31.21 total
```

Let me know if you need a full reproducer.

JohanEngelen:
> Let me know if you need a full reproducer.

yes please

CyberShadow:
OK, here it is:

```console
$ git clone -b ldc4294 https://github.com/CyberShadow/btdu
$ cd btdu
$ DOCKER=podman ci/build-docker.sh i686
```

Sorry it's not a minimal test case!

CyberShadow:
Here it is, reduced it to 3 files (one D file):

https://github.com/cybershadow/btdu/tree/ldc4294-v2

Run `build.sh` to reproduce.

PR #5036, opened 2025-12-20
Fix DMD-style asm labels for LTO with template functions

CyberShadow:
LLM disclosure: this entire PR is AI-generated. It might be OK, or it might be slop. The fix looks sensible on the surface, and it passes the test suite, however, so does most slop in general. The bug report was open for a couple years now and no one was picking it up, so off the chance that this is actually OK, it seems worth a shot.

----

When template functions containing DMD-style inline assembly are instantiated in multiple modules and linked with LTO, duplicate symbol errors can occur because the asm labels have identical names.

This fix assigns a unique ID to each function containing inline asm, based on a deterministic hash of the template instantiation module name. The ID is appended to label names, ensuring uniqueness across compilation units.

Fixes: https://github.com/ldc-developers/ldc/issues/4294

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dmd/declaration.h
+++ b/dmd/declaration.h
@@ -565,6 +565,11 @@ public:
     // Whether to emit instrumentation code if -fprofile-instr-generate is specified,
     // the value is set with pragma(LDC_profile_instr, true|false)
     bool emitInstrumentation;
+
+    // Unique ID for asm labels in this function, used to prevent duplicate
+    // symbol errors when LTO merges multiple instantiations of template functions.
+    // See: https://github.com/ldc-developers/ldc/issues/4294
+    unsigned asmLabelId;
 #endif
 
     VarDeclaration *vresult;            // result variable for out contracts
--- a/dmd/func.d
+++ b/dmd/func.d
@@ -240,6 +240,11 @@ version (IN_LLVM)
     // Whether to emit instrumentation code if -fprofile-instr-generate is specified,
     // the value is set with pragma(LDC_profile_instr, true|false)
     bool emitInstrumentation = true;
+
+    // Unique ID for asm labels in this function, used to prevent duplicate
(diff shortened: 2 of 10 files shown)


tests/linking/inputs/asm_lto_template.d
```diff
@@ -0,0 +1,36 @@
+// Template with naked function containing inline asm labels.
```
JohanEngelen: FYI: you can use `split-file` to avoid having to manually create separate test files, i.e. you can keep all needed test source in a single file. See for example: https://github.com/ldc-developers/ldc/blob/5fa7914fdfd023a0822814aedd20c865ab2ab89c/tests/PGO/sample_based.d#L3 or https://github.com/ldc-developers/ldc/blob/5fa7914fdfd023a0822814aedd20c865ab2ab89c/tests/driver/config_diag.d#L1

CyberShadow: Thanks, let me know if you prefer that style and I'll refactor.

JohanEngelen: Prefer: yes. Needed: no. Use your time well ;-)

CyberShadow: Thank you for the kind consideration. Fortunately, it's not my time at stake - it's the bot's time, and it's on payroll anyway :grin: Done.

CyberShadow:
> OK, looks like to fix those we need to stop using module-level asm and move to embedding x86 asm in LLVM IR functions. This would be the more general / comprehensive fix for this bug too, and seems to be what clang does for C/C++, but it's a much bigger lift.

Well, here it is: https://github.com/ldc-developers/ldc/pull/5041

PR #5039, opened 2025-12-22
Add `-fdebug-prefix-map` and follow clang's behavior for debuginfo file/dir names

JohanEngelen:
https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fdebug-prefix-map

This is required to have reproducible builds without remapping the source code directory to a fixed location (for example in Bazel, where the source code for every build action will end up in a sandbox directory with some hash in the path name).

Try to keep the implementation as close as possible to what clang does.

Originally implemented by Ilya, modified and expanded by Johan.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -224,7 +224,7 @@ cl::opt<bool> emitDwarfDebugInfo(
     cl::desc("Emit DWARF debuginfo (instead of CodeView) for MSVC targets"));
 
 // Prefix map for filenames in DWARF debuginfo
-std::map<std::string, std::string> debugPrefixMap;
+llvm::SmallVector<std::pair<std::string, std::string>, 0> debugPrefixMap;
 
 struct DwarfPrefixParser : public cl::parser<std::string> {
   explicit DwarfPrefixParser(cl::Option &O) : cl::parser<std::string>(O) {}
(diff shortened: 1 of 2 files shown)


JohanEngelen:
I guess the test timeouts mean that indeed this now takes a lot of compile time to generate file info. Will implement a caching mechanism

JohanEngelen:
@kinke Can you review and give your OK for this change? Appreciated!

gen/dibuilder.cpp
```diff
@@ -218,27 +219,80 @@ void DIBuilder::SetValue(Loc loc, llvm::Value *value,
...
 
+std::string DIBuilder::remapDIPath(llvm::StringRef path) {
+#if LDC_LLVM_VER >= 1800
+#define startswith starts_with
+#endif
+
+  for (const auto &[from, to] : opts::debugPrefixMap) {
+    if (path.startswith(from)) {
+      return to + path.substr(from.size()).str();
+    }
```
kinke: clang uses `llvm::sys::path::replace_path_prefix()`, which seems to accept different casing and separators on Windows. Was this manual, string-exact approach intentional?

JohanEngelen: ah, indeed, let's use that too then.

gen/dibuilder.cpp
```diff
@@ -220,20 +220,11 @@ void DIBuilder::SetValue(Loc loc, llvm::Value *value,
...
-      return to + path.substr(from.size()).str();
-    }
-  }
-  return std::string(path);
-
-#if LDC_LLVM_VER >= 1800
-#undef startswith
-#endif
+  llvm::SmallString<256> P = path;
+  for (auto &[from, to] : llvm::reverse(opts::debugPrefixMap))
```
kinke, line 224: (clang's `reverse` makes sense because their `debugPrefixMap` is a vector of string-pairs, not a `std::map`)

**Edit**: I guess that's done so that later `-fdebug-prefix-map` cmdline args can override earlier ones.

kinke: I've pushed a commit to mimic clang there too.

JohanEngelen: thanks!

PR #5041, opened 2025-12-23
Emit naked functions as LLVM IR with inline asm instead of module asm

### LINKED ISSUE #4294: `multibyteMulAdd` "is already defined" with `-flto=full` on i686

CyberShadow:
LLM disclosure: this entire PR is AI-generated. It might be OK, or it might be slop. The fix looks sensible on the surface, and it passes the test suite, however, so does most slop in general. Caveat emptor.

This is the more comprehensive alternative to #5036. Instead of hacking the front-end to track and deduplicate label names in assembly so that they don't collide in LTO, this changes LDC's handling of naked asm functions to use a completely different approach - LLVM IR with inline asm blocks. I understand this is what C++ does, and is how clang avoids the problem we were seeing.

LLM output follows:

----

## Summary

This PR fixes issue #4294 where LTO linking fails with "symbol already defined" errors for naked template functions. The fix changes how naked functions are emitted: instead of generating module-level assembly, we now emit proper LLVM IR functions with inline assembly.

## Problem

When compiling code that uses naked template functions (like `std.internal.math.biguintx86`) with LTO (`-flto=full` or `-flto=thin`), linking fails with errors like:

```
error: symbol '.L_D3std8internal4math10biguintx86__T15multibyteMulAddVai43ZQxFNaNbNfAkxAkkkZk_L1' is already defined
```

### Root Cause

The previous implementation emitted naked functions as **module-level assembly** using `gIR->module.appendModuleInlineAsm()`. This approach has a fundamental problem with LTO:

1. During normal (non-LTO) linking, COMDAT deduplication works correctly - the linker sees multiple definitions of template instantiations and keeps only one
2. During LTO, LLVM first **concatenates all module-level assembly** from all compilation units into a single block
3. This concatenation happens **before** COMDAT deduplication can occur
4. Result: duplicate label definitions cause "symbol already defined" errors

### Why This Affects Specific Code

- **i686 only**: `biguintx86.d` only exists for x86 (32-bit). x86_64 uses `biguintnoasm.d` (pure D)
- **Templates**: The issue requires the same template to be instantiated in multiple compilation units
- **Labels**: Naked functions with assembly labels (like `L1:`, `L_unrolled:`) trigger the duplicate symbol errors

## Solution

Emit naked functions as **LLVM IR functions with inline assembly** instead of module-level assembly:

```cpp
// Create LLVM function with proper linkage
llvm::Function *func = llvm::Function::Create(funcType, linkage, mangle, &module);

// Set naked attribute - LLVM won't generate prologue/epilogue
func->addFnAttr(llvm::Attribute::Naked);

// Prevent optimizations that might clone the function (would duplicate labels)
func->addFnAttr(llvm::Attribute::OptimizeNone);
func->addFnAttr(llvm::Attribute::NoInline);

// For templates, set up COMDAT for proper deduplication
if (fd->isInstantiated()) {
    func->setComdat(module.getOrInsertComdat(mangle));
}

// Create inline asm call with the function body
llvm::InlineAsm *inlineAsm = llvm::InlineAsm::get(asmFuncType, asmBody, "", true, false, AD_ATT);
gIR->ir->CreateCall(inlineAsm);
```

### Key Changes

1. **LLVM IR emission**: Naked functions are now proper LLVM functions with the `naked` attribute
2. **COMDAT groups**: Template instantiations use COMDAT for proper deduplication during LTO
3. **Optimization prevention**: `OptimizeNone` and `NoInline` attributes prevent LLVM from cloning the function (which would duplicate labels)
4. **Label consistency**: Label definitions use `printLabelName()` to match the quoted format used in jump instructions

### New Tests Added

| Test | Description |
|------|-------------|
| `tests/linking/asm_labels_lto.d` | Verifies LTO linking works with naked template functions |
| `tests/codegen/naked_asm_output.d` | Verifies correct assembly output (no prologue/epilogue, correct labels, COMDAT) |
| `tests/codegen/naked_asm_corner_cases.d` | Tests corner cases: push/pop, forward/backward jumps, multiple control flow paths |

### Test Results

- **All lit tests pass** (337 pass, 1 pre-existing failure unrelated to this change)
- **All druntime unit tests pass**
- **All phobos unit tests pass**
- **Phobos library builds successfully** with `-O3 -release` (includes `biguintx86.d`)
- **Manual LTO tests pass** (both full and thin LTO)

### Verified Properties

1. **No prologue/epilogue overhead** - Assembly output verified
2. **Correct label generation** - Labels like `.LfuncName_labelName:`
3. **COMDAT for templates** - Section uses `"axG"` with comdat group
4. **Proper LLVM IR attributes** - `naked noinline nounwind optnone`
5. **Stack manipulation works** - push/pop instructions verified at runtime
6. **Control flow works** - Forward/backward jumps verified at runtime

## Files Changed

| File | Changes |
|------|---------|
| `gen/naked.cpp` | Refactored `DtoDefineNakedFunction` to emit LLVM IR instead of module asm |
| `tests/linking/asm_labels_lto.d` | New test for LTO linking |
| `tests/codegen/naked_asm_output.d` | New test for assembly output verification |
| `tests/codegen/naked_asm_corner_cases.d` | New test for corner cases |

### Before (Module Assembly)
```asm
    .section  .text.funcName,"axG",@progbits,funcName,comdat
    .weak     funcName
    .type     funcName,@function
funcName:
    ; inline asm body
    .size     funcName, .-funcName
```

### After (LLVM IR with Inline Asm)
```llvm
define i32 @funcName() #0 comdat {
entry:
  call void asm sideeffect "...", ""()
  unreachable
}

attributes #0 = { naked noinline nounwind optnone }
```

### Why `OptimizeNone` and `NoInline`?

During optimization passes (especially at `-O1` and above), LLVM may clone or duplicate functions. For naked functions with assembly labels, this would create duplicate label definitions. These attributes prevent such transformations.

### Why `unreachable` terminator?

Naked functions handle their own return via assembly (`ret` instruction). The `unreachable` terminator tells LLVM not to generate any return sequence - the inline asm handles it.

## Compatibility

- **No ABI changes**: The generated machine code is identical
- **No behavior changes**: Functions work exactly as before
- **Backwards compatible**: Existing code continues to work

## Limitations

- Tested on x86_64 Linux; Windows and macOS may need additional verification
- The i686 scenario (original bug) requires a 32-bit toolchain to fully verify

## Related Issues

- Fixes #4294
- Related to btdu build failures with LTO

## Checklist

- [x] Code compiles without warnings
- [x] All existing tests pass
- [x] New tests added for the fix
- [x] Assembly output verified (no overhead)
- [x] Runtime behavior verified
- [x] LTO linking verified (both full and thin)
- [x] Phobos builds successfully with optimizations


kinke:
Alright, I got back to this and have revised the tests too now. I'm happy with it; I hope you are too after the rework, Vladimir :) - if not, please shout!

Btw, IIRC, I checked whether we can enable LTO for LDC's own Win32 build (months ago, via temp commits checking CI), testing both thin and full LTO, but we still hit exactly the same problems as before.

kinke (approved):

gen/asm-x86.h
```diff
@@ -2340,57 +2340,15 @@ struct AsmProcessor {
                   AsmCode *asmcode, AsmArgMode mode = Mode_Input) {
     using namespace dmd;
 
-    if (sc->func->isNaked()) {
-      switch (type) {
-      case Arg_Integer:
-        if (e->type->isUnsigned()) {
-          insnTemplate << "$" << e->toUInteger();
-        } else {
+    if (type == Arg_Integer) {
```
kinke, line 2343: Doing this now for non-naked asm too, to avoid an integer input for the asm expression - simply use the literal in the template string directly, for readability and simplicity.

gen/irstate.h
```diff
@@ -94,9 +94,6 @@ struct IRAsmBlock {
   std::set<std::string> clobs;
   size_t outputcount;
 
-  // stores the labels within the asm block
-  std::vector<Identifier *> internalLabels;
```
kinke, line 98: Collecting them now in `FuncDeclaration.asmLabels` instead, for the whole function, means that *non*-naked asm functions also don't need the extra complicated forwarding code when jumping to an asm label in another `asm` block.

PR #5045, opened 2026-01-19
Set lambda linkage to LinkOnceODR (was Internal)

JohanEngelen:
This reverts https://github.com/ldc-developers/ldc/pull/4415. The mangling of lambdas has been fixed (and is now stable and cannot clash with other modules anymore), so there is no longer a need for internal linkage. This allows merging of identical lambdas again (multiple instantiations).

This also fixes an instantiation bug, where the lambda's emission is sometimes culled because the frontend knows it is already instantiated in another module, but because of internal linkage it cannot resolve the symbol during linking (thus resulting in a missing symbol during linking).

### DIFF
--- a/gen/tollvm.cpp
+++ b/gen/tollvm.cpp
@@ -228,6 +228,10 @@ LLGlobalValue::LinkageTypes DtoLinkageOnly(Dsymbol *sym) {
     if ((decl->storage_class & STCstatic) && decl->isCsymbol())
       return LLGlobalValue::InternalLinkage;
 
+  /* Function (incl. delegate) literals are emitted into each referencing
+   * compilation unit, so use linkonce_odr for all lambdas and all global
+   * variables they define.
+   */
   auto potentialLambda = sym;
   if (auto vd = sym->isVarDeclaration())
     if (vd->isDataseg())


JohanEngelen:
@ljmf00 Perhaps you can dustmite the `libwtracer.a` build, to create a testcase? I have not been able to create a testcase by some ideas that I had (imported modules, ...)
Details: compile with LDC v1.38.0-weka9 (nota bene, -weka10 has the issue fixed; or with LDC v.1.38.0 vanilla), `libwtracer.a` contains a reference (declare, does not define) to symbol `_D3std4conv__T6toImplTiTkZQmFkZ20__dgliteral_L561_C17MFNaNbNfZC6object9Throwable`. Phobos binary does not export that symbol with LDC v1.38.0-weka9 (because internal linkage), so it should have been emitted into `libwtracer.a`.

gen/tollvm.cpp
kinke: Please revert this comment too.

JohanEngelen: Done

kinke: I meant to the version before #4415 - pushed an according commit.

PR #5046, opened 2026-01-21
Merge upstream stable

kinke:


kinke:
Damn, the null-pointer druntime integration tests which recently regressed for CircleCI now also seem to have regressed with GHA.

PR #5052, opened 2026-01-30
Try to fix recent segfault-handler test CI regressions on Linux

kinke:
Based on https://github.com/ldc-developers/ldc/pull/5041#issuecomment-3689617009.


kinke:
Interesting, looks like Claude was on the right track, pretty impressive! Thanks @CyberShadow.

PR #5054, opened 2026-02-03
PowerPC64: Generalize inherit-real-ABI-from-C++-host-compiler logic to all PowerPC64 hosts, not just Linux

### LINKED ISSUE #4959: Unable to build LDC 1.41 on FreeBSD/PowerPC64
Attempting to build LDC 1.41 on FreeBSD 14.1 on PowerPC64(BE) using GDC, from GCC v14.2.  

Get the below error:

`std/internal/math/gammafunction.d(262): Error: static assert:  "missing MAXGAMMA for other real types"
    static assert(0, "missing MAXGAMMA for other real types");
    ^`

All/Any help is appreciated.

kinke:
Should fix #4959 for FreeBSD.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 #### Platform support
 - Supports LLVM 15 - 21.
+- PowerPC64: `real` now matches the C++ host compiler's `long double` when compiling for the native target, not just on Linux hosts. (#5054)
 
 #### Bug fixes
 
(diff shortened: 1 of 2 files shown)


kinke:
I guess it's safer to restrict this logic to the same OS - so that cross-compiles (to PPC64 with different OS) all default to the `doubledouble` type, like non-PPC64 LDC builds targeting PPC64.

PR #5066, opened 2026-02-17
Avoid undefined RTInfoImpl symbols at link-time

kinke:
At Symmetry, we have come across 2 such undefined symbols for a huge shared library - for TypeInfos for structs defined in .di files.

See https://github.com/dlang/dmd/pull/21812#discussion_r2804764715 and following for more context.

### DIFF
--- a/ir/irstruct.cpp
+++ b/ir/irstruct.cpp
@@ -9,6 +9,7 @@
 
 #include "dmd/errors.h"
 #include "dmd/mangle.h"
+#include "dmd/module.h"
 #include "dmd/mtype.h"
 #include "dmd/template.h"
 #include "gen/irstate.h"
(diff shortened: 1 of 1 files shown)


ir/irstruct.cpp
```diff
@@ -186,6 +187,23 @@ LLConstant *IrStruct::getTypeInfoInit() {
   // immutable(void)* m_RTInfo
   if (!isOpaque && sd->getRTInfo) {
     b.push(toConstElem(sd->getRTInfo, gIR));
+
+    // Make sure the pointed-to RTInfoImpl global is defined in a root module
+    // (e.g., not guaranteed for TypeInfos for structs defined in .di files,
+    // causing undefined RTInfoImpl symbols at link-time).
+    if (global.params.linkonceTemplates == LinkonceTemplates::no) {
```
kinke, line 194: With `-linkonce-templates`, the template instance is codegen'd in every referencing object file.

ir/irstruct.cpp
```diff
@@ -186,6 +187,23 @@ LLConstant *IrStruct::getTypeInfoInit() {
...
+      if (auto soe = sd->getRTInfo->isSymOffExp()) {
```
kinke, line 195: Weirdly, I've seen that `sd->getRTInfo` can be a `NullExp` - not sure how that can happen, as `__traits(getPointerBitmap, T)` should never return a null-slice.

kinke: Ah wait, that's from the `RTInfo` template, not the `RTInfoImpl` one - so the `NullExp` happens if the struct doesn't contain any pointers.

ir/irstruct.cpp
```diff
@@ -186,6 +187,23 @@ LLConstant *IrStruct::getTypeInfoInit() {
...
+        assert(soe->offset == 0);
+        auto rtInfoImplVar = soe->var->isVarDeclaration();
+        auto mod = rtInfoImplVar->isInstantiated()->inst->memberOf;
+        if (!mod || !mod->isRoot()) {
+          // The RTInfoImpl template instance won't be codegen'd in this
+          // compiler run - define the variable manually in the current object
+          // file.
+          getIrGlobal(rtInfoImplVar)->getValue(/*define=*/true);
+        }
```
kinke, line 205: Note that in case `mod` is a root module, `needsCodegen()` culling might still prevent the variable from being defined. But I haven't come across that yet.

kinke: Correction: I have seen `needsCodegen()` culling leading to undefined `RTInfoImpl` symbols now, with `-allinst` at least.

Not too surprising, since emitting a TypeInfo for a non-sema3 struct involves a new `RTInfoImpl` instantiation (as part of `semanticTypeInfoMembers()`), extending that template-instance-graph (used for the `needsCodegen()` logic) - **at the codegen stage**, not in the earlier regular semantic passes. E.g., the (primary) `RTInfoImpl` template instance might have been added to an already-codegen'd module, with `needsCodegen()` returning false at that time - and another later codegen'd module emitting a TypeInfo, instantiating a new sibling instance, which would have made `needsCodegen()` return true.

Issue #4383, opened 2023-04-27
Android binaries from releases page are broken with `stack corruption detected (-fstack-protector)`


truboxl:
Downloaded from https://github.com/ldc-developers/ldc/releases/tag/v1.32.1

https://github.com/ldc-developers/ldc/releases/download/v1.32.1/ldc2-1.32.1-android-aarch64.tar.xz
```
--------- beginning of crash
04-27 14:41:12.102 18232 18232 F libc    : stack corruption detected (-fstack-protector)
04-27 14:41:12.103 18232 18232 F libc    : stack corruption detected (-fstack-protector)
04-27 14:41:12.104 18232 18232 F libc    : Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 18232 (ldc2), pid 18232 (ldc2)
04-27 14:41:12.106 20700 20769 E Netd    : getNetworkForDns: getNetId from enterpriseCtrl is netid 0
04-27 14:41:12.177 18236 18236 I crash_dump64: obtaining output fd from tombstoned, type: kDebuggerdTombstone
04-27 14:41:12.181  3897  3897 I tombstoned: received crash request for pid 18232
04-27 14:41:12.184 18236 18236 I crash_dump64: performing dump of process 18232 (target tid = 18232)
04-27 14:41:12.186 18236 18236 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
04-27 14:41:12.187 18236 18236 F DEBUG   : Build fingerprint: 'samsung/a30sdx/a30s:11/RP1A.200720.012/A307GNDXS5CWC3:user/release-keys'
04-27 14:41:12.187 18236 18236 F DEBUG   : Revision: '3'
04-27 14:41:12.187 18236 18236 F DEBUG   : ABI: 'arm64'
04-27 14:41:12.189 18236 18236 F DEBUG   : Timestamp: 2023-04-27 14:41:12+0800
04-27 14:41:12.189 18236 18236 F DEBUG   : pid: 18232, tid: 18232, name: ldc2  >>> ./ldc2 <<<
04-27 14:41:12.189 18236 18236 F DEBUG   : uid: 10574
04-27 14:41:12.189 18236 18236 F DEBUG   : signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
04-27 14:41:12.189 18236 18236 F DEBUG   : Abort message: 'stack corruption detected (-fstack-protector)'
[...]

PR #5070, opened 2026-02-22
Android: Don't register LLVM signal handlers

kinke:
A little test wrt. #4383, based on https://github.com/termux/termux-packages/pull/28586#issuecomment-3941836499.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -296,6 +296,7 @@ jobs:
             android_x86_arch: x86_64
             extra_cmake_flags: >-
               -DLDC_INSTALL_LLVM_RUNTIME_LIBS_ARCH=aarch64-android
+              -DCMAKE_BUILD_TYPE=RelWithDebInfo
 
     name: ${{ matrix.job_name }}
     runs-on: ${{ matrix.host_os }}


robertkirkman:
I was able to successfully compile this in GitHub Actions by changing the Ubuntu version to 24.04 and trying until it worked:

https://github.com/robertkirkman/ldc/actions/runs/22288993543/job/64472748363

and, when I test this one, it looks like it works just slightly more than the [1.41.0 release](https://github.com/ldc-developers/ldc/releases/tag/v1.41.0)! It prints the messages `Error: No source files` and `LDC - the LLVM D compiler`, but the other one doesn't.

however, unfortunately, when I try to compile anything with it, it still prints `stack corruption detected (-fstack-protector) Aborted`, so the same problem might still be happening, and it might be just a coincidence that it prints slightly more messages.

```
~/ldc2-8005ae55-android-aarch64/bin $ ./ldc2 --version
LDC - the LLVM D compiler (1.42.0-git-8005ae5):
  based on DMD v2.112.1 and LLVM 21.1.8
  built with LDC - the LLVM D compiler (1.42.0-git-8005ae5)
  Default target: aarch64-none-linux-android29
stack corruption detected (-fstack-protector)
Aborted                    ./ldc2 --version
~/ldc2-8005ae55-android-aarch64/bin $ cat > test.d << 'EOF'
>      void main()
     {
          import std.stdio : writefln;
          "Hello".writefln!"%s, World!";
     }
> EOF
~/ldc2-8005ae55-android-aarch64/bin $ ./ldc2 test.d
stack corruption detected (-fstack-protector)
Aborted                    ./ldc2 test.d
~/ldc2-8005ae55-android-aarch64/bin $ ./ldc2
Error: No source files
~/ldc2-8005ae55-android-aarch64/bin $ gdb --args ./ldc2 test.d
GNU gdb (GDB) 16.3
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/
...
stack corruption detected (-fstack-protector)

Program received signal SIGABRT, Aborted.
0x0000007fbb5a7994 in abort () from /apex/com.android.runtime/lib64/bionic/libc.so
(gdb) bt
#0  0x0000007fbb5a7994 in abort () from /apex/com.android.runtime/lib64/bionic/libc.so
#1  0x0000007fbb5bc5ec in __stack_chk_fail () from /apex/com.android.runtime/lib64/bionic/libc.so
#2  0x000000555a1c4a18 in RegisterHandlers() ()
#3  0x0000000000000000 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) 
```

robertkirkman:
`ulimit -s` shows `8192` in both Termux and ADB shell on my device, yes.

robertkirkman:
There's a `libllvm` package in Termux at version 21.1.8 that we try to use as much as possible with LLVM-based compilers, so I might try compiling ldc with non-cross-compilation (inside Termux) and the prebuilt Termux `libllvm` package by attempting to follow these steps: https://wiki.dlang.org/Building_LDC_from_source#Building_LDC_from_source to check if it's possible for it to work that way. If it does, then it would suggest the problem is just in the LLVM build settings of the CI in github.com/ldc-developers/ldc , but if it doesn't, then something else might be going on.

PR #5072, opened 2026-02-23
Android: Don't build LDC with NDK's implicit `-fstack-protector-strong`

kinke:
Another test wrt. #4383, incl. new LLVM Android builds without `-fstack-protector-strong`.

### DIFF
--- a/.github/actions/helper-build-ldc/action.yml
+++ b/.github/actions/helper-build-ldc/action.yml
@@ -40,7 +40,7 @@ runs:
           ${{ inputs.specify_install_dir == 'true' && '-DINCLUDE_INSTALL_DIR="$installDir/import"' || '' }} \
           ${{ inputs.cmake_flags }}
 
-        ninja obj/ldc2.o ${{ inputs.build_targets }}
+        ninja -v obj/ldc2.o ${{ inputs.build_targets }}
 
     # Windows: invoke CMake & ninja in MSVC env
     - if: runner.os == 'Windows'


robertkirkman:
Thanks! I'll test it ASAP

kinke:
I'll now try without alternate signal stack on Android: https://github.com/ldc-developers/llvm-project/commit/e7130c48e18acba0cb125af1b6be9bf476b3ca93

kinke:
That doesn't change anything either, giving up.

kinke:
Not just yet; trying a no-op LLVM `RegisterHandlers()` on Android: https://github.com/ldc-developers/llvm-project/commit/2846ebab28eb4aee529f2dedf87f97f0e9c87b3e

FWIW, I've also glanced at a few files in https://github.com/termux/termux-packages/tree/master/packages/libllvm, and I couldn't spot any interesting patch or CMake build option for Termux' LLVM.

kinke:
The crash for `ldc2 --version` might be the most interesting/simple one. This: https://github.com/ldc-developers/ldc/blob/2e33c9c4873ef557d5359ce0eaf8cd7631db3fe4/driver/main.cpp#L138-L156 only manages to print the default target triple; then the stack corruption gets reported with `-fstack-protector-strong`, or a segfault/sigbus happens without that switch.

kinke:
I think I'm really out of ideas now. Still getting sigbus for `ldc2 --version` and when compiling a hello-world - but the emitted object file (emitted successfully before the crash) is absolutely fine; I can link it manually with Termux's `cc` (clang 21.1.8 targeting API level 30) using bundled druntime and Phobos, and that hello-world then runs fine too.

PR #5085, opened 2026-03-08
CI: Remove obsolete test exclusion for supported-LLVM-versions workflow

kinke:


.github/workflows/supported_llvm_versions.yml
```diff
@@ -166,10 +166,6 @@ jobs:
           excludes='dmd-testsuite|lit-tests|ldc2-unittest'
           if [[ '${{ runner.os }}' == macOS ]]; then
             N=$(sysctl -n hw.logicalcpu)
-            if [[ '${{ matrix.os }}' == macos-14 ]]; then
-              # FIXME: crashes frequently with enabled optimizations on M1 runners
-              excludes+='|^std.internal.math.gammafunction(-shared)?$'
-            fi
```
kinke, line 172: This module works fine for the Main GHA workflow nowadays, incl. on macOS arm64, so should work here too.

PR #5088, opened 2026-03-09
llvm23 3 files commit

gnavdev28:


driver/cl_options.cpp
```diff
@@ -788,12 +788,19 @@ static cl::extrahelp
 /// The clashing LLVM options are suffixed with "llvm-" and hidden from the
 /// -help output.
 void createClashingOptions() {
+#if LDC_LLVM_VER >= 2300
+  auto &map = cl::getRegisteredOptions();
+#else
   llvm::StringMap<cl::Option *> &map = cl::getRegisteredOptions();
-
+#endif
```
thewilsonator: Just use `auto &map = cl::getRegisteredOptions();` for all versions and remove the preprocessor conditional compilation. The total diff should then be 
```diff
-  llvm::StringMap<cl::Option *> &map = cl::getRegisteredOptions();
+  auto &map = cl::getRegisteredOptions();
```

driver/cl_options.cpp
```diff
@@ -1016,13 +1023,18 @@ void hideLLVMOptions() {
                                                "trap-func",
                                                "W"};
 
-  llvm::StringMap<cl::Option *> &map = cl::getRegisteredOptions();
+  auto &map = cl::getRegisteredOptions();
```
thewilsonator, line 1022: as was done here

driver/cl_options.cpp
```diff
@@ -1032,6 +1044,5 @@ void hideLLVMOptions() {
       map.erase(it);
     }
   }
-}
-
-} // namespace opts
+ }
+}
```
thewilsonator: remove this portion of the diff

driver/cl_options.cpp
```diff
@@ -1016,13 +1019,18 @@ void hideLLVMOptions() {
...
   for (const auto name : hiddenOptions) {
     // Check if option exists first for resilience against LLVM changes
     // between versions.
     auto it = map.find(name);
     if (it != map.end()) {
+#if LDC_LLVM_VER >= 2300
       it->second->setHiddenFlag(cl::Hidden);
+#else
+      it->getValue()->setHiddenFlag(cl::Hidden);
+#endif
```
thewilsonator: this does not change the compilation path for LLVM 23, why is it necessary?

driver/cl_options.cpp
```diff
@@ -788,12 +788,15 @@ static cl::extrahelp
 /// The clashing LLVM options are suffixed with "llvm-" and hidden from the
 /// -help output.
 void createClashingOptions() {
-  llvm::StringMap<cl::Option *> &map = cl::getRegisteredOptions();
-
+  auto &map = cl::getRegisteredOptions();
   auto renameAndHide = [&map](const char *from, const char *to) {
     auto i = map.find(from);
     if (i != map.end()) {
+#if LDC_LLVM_VER >= 2300
```
gulugulubing, line 795: The exact threshold here needs to be checked. I found that the signature of cl::getRegisteredOptions has changed since LLVM 22.1.0
<https://github.com/llvm/llvm-project/commit/ecc8a955af76458e14fe231447e42bb08d75246b>

runtime/druntime/src/ldc/intrinsics.di
```diff
@@ -28,6 +28,7 @@ else version (LDC_LLVM_1901) enum LLVM_version = 1901;
 else version (LDC_LLVM_2001) enum LLVM_version = 2001;
 else version (LDC_LLVM_2101) enum LLVM_version = 2101;
 else version (LDC_LLVM_2200) enum LLVM_version = 2200;
```
gulugulubing, line 30: Related to the above review, if we use llvm 22.1.0, we need a version 2201 here.

PR #5098, opened 2026-03-14
Fix #5095: LDC fails to build on Windows with DMD/MSVC due to real_t confusion

### LINKED ISSUE #5095: LDC fails to build on Windows with DMD/MSVC due to `real_t` confusion
```d
module dmd.root.longdouble;

version (CRuntime_Microsoft)
{
    static if (real.sizeof > 8)
        alias longdouble = real; /// <<< this is selected
    else
        alias longdouble = longdouble_soft;
}
else
    alias longdouble = real;

// longdouble_soft needed when building the backend with
// Visual C or the frontend with LDC on Windows
version (CRuntime_Microsoft):
/// ... implementation of `longdouble_soft`
```
```c++
///longdouble.h
#if !_MSC_VER // has native 10 byte doubles
#include <stdio.h>
typedef long double longdouble;
typedef volatile long double volatile_longdouble;

// also used from within C code, so use a #define rather than a template
// template<typename T> longdouble ldouble(T x) { return (longdouble) x; }
#define ldouble(x) ((longdouble)(x))

...

#else

#include <float.h>
#include <limits>

struct longdouble_soft;

// implemented in longdouble.d
double ld_read(const longdouble_soft* const ld);
long long ld_readll(const longdouble_soft* const ld);
/// declaration of longdouble_soft
...

size_t ld_sprint(char* str, size_t size, int fmt, longdouble_soft x);

//////////////////////////////////////////////
typedef longdouble_soft longdouble;
```

results in D thinking `longdouble = real` and C++ thinking `longdouble = longdouble_soft`.
Setting the line marked `<<<` to be 
    `alias longdouble = longdouble_soft;` 
results in 
    `*res = core.math.yl2x(*x, *y);`
in `root/ctfloat.d` being used with `longdouble_soft` which fails to compile.
`enum yl2x_supported = __traits(compiles, core.math.yl2x(1.0L, 2.0L));` should have a check for `is(real_t == real)`.

thewilsonator:
N.B. not tested

cc @gnavdev28

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,9 @@ set(ALTERNATIVE_MALLOC_O   ""     CACHE STRING "If specified, adds ALTERNATIVE_M
 # Most linux distributions have a policy of not bundling dependencies like zlib
 set(PHOBOS_SYSTEM_ZLIB OFF CACHE BOOL "Use system zlib instead of Phobos' vendored version")
 
+# Force use of softfloat for compiler
+option(LDC_FORCE_SOFTFLOAT_REAL "Force LDC to be built with 80-bit real in software" OFF)
+
 if(D_VERSION EQUAL 1)
     message(FATAL_ERROR "D version 1 is no longer supported.
 Please consider using D version 2 or checkout the 'd1' git branch for the last version supporting D version 1.")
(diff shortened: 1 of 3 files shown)


gnavdev28:
this worked, thanks

PR #5104, opened 2026-03-25
Android: Add ImportC definitions

eakmyr:
This PR fixes broken ImportC on Android targets. Tested on sqlite3, zlib, lwIP.


thewilsonator (commented):
please file this upstream to DMD and once that is accepted with review (e.g. `__builtin_memset` and friend have recently beed added to DMD, though maybe that was for `APPLE` I forget, and #define __alignof__(x) sizeof(x)` looks suspect) cherrypick this back to LDC if you need to.

PR #5111, opened 2026-03-31
druntime: Implement va_arg for WebAssembly

IDONTUSEGH:
It uses 'pragma(LDC_va_arg)' to emit the LLVM instruction for scalars and pointers (matches how it's implemented in tocall.cpp), else it falls back to manual pointer arithmetic.

I'm not 100% sure if that's correct.

### DIFF
--- a/runtime/druntime/src/core/stdc/stdarg.d
+++ b/runtime/druntime/src/core/stdc/stdarg.d
@@ -180,6 +180,24 @@ else version (DigitalMars)
  */
 version (GNU)
     T va_arg(T)(ref va_list ap); // intrinsic
+else version (WebAssembly){
+    pragma(LDC_va_arg)
+    T ldc_va_arg(T)(ref va_list ap);
+    T va_arg(T)(ref va_list ap)
+    {
+        static if (__traits(isScalar, T) || is(T == U*, U))
+        {
+            return ldc_va_arg!T(ap);
+        }
+        else
+        {
+            ap = ap.alignUp!(T.alignof);
+            auto p = cast(T*) ap;
+            ap += T.sizeof.alignUp;
+            return *p;
+        }
+    }
+}
 else
 T va_arg(T)(ref va_list ap)
 {


runtime/druntime/src/core/stdc/stdarg.d
```diff
@@ -180,6 +180,24 @@ else version (DigitalMars)
...
 version (GNU)
     T va_arg(T)(ref va_list ap); // intrinsic
+else version (WebAssembly){
+    pragma(LDC_va_arg)
+    T ldc_va_arg(T)(ref va_list ap);
+    T va_arg(T)(ref va_list ap)
+    {
+        static if (__traits(isScalar, T) || is(T == U*, U))
+        {
+            return ldc_va_arg!T(ap);
```
kinke, line 190: We don't use this LLVM intrinsic anywhere else, it's basically useless and only works for trivial types (if at all), where the implementation is trivial anyway (your `else` branch most likely).

After glancing at https://github.com/ldc-developers/ldc/blob/master/gen/abi/wasm.cpp and the last paragraph in https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md#function-arguments-and-return-values, I think the `else` branch might really suffice.

runtime/druntime/src/core/stdc/stdarg.d
```diff
@@ -180,6 +180,24 @@ else version (DigitalMars)
...
+        }
+        else
+        {
+            ap = ap.alignUp!(T.alignof);
+            auto p = cast(T*) ap;
+            ap += T.sizeof.alignUp;
```
kinke, line 196: I don't think the size needs to be aligned up here; this will be taken care of by the `alignUp!(T.alignof)` for the next argument. If I understand the spec paragraph correctly:
> Arguments are arranged in order in the buffer, with each argument placed at the lowest appropriately aligned address after the previous argument.

runtime/druntime/src/core/stdc/stdarg.d
kinke: So if the `else` branch suffices, I'd say incorporate it in the generic `va_arg` template below, as for the other archs.

QuantumSegfault:
I handled this as part of dlang/dmd#23294

JohanEngelen:
> I handled this as part of [dlang/dmd#23294](https://github.com/dlang/dmd/pull/23294)

so this PR can be closed?

QuantumSegfault:
> > I handled this as part of [dlang/dmd#23294](https://github.com/dlang/dmd/pull/23294)
> 
> so this PR can be closed?

Should be able to. Unless I did mine incorrectly (somebody should check; I did some very basic comparisons against Clang output for wasm32 and wasm64, but by no means comprehensive).

kinke:
Yep thx, the upstream implementation seems reasonable (and would otherwise need to be fixed there).

Issue #3504, opened 2020-07-11
LLVM error ("unwind edges out of a funclet pad...") - only on Windows


dewf:
**System**: Windows 10 Pro 1909
**Architecture**:  x86_64-pc-windows-msvc
**LLVM**: 10.0.0 (via LDC 1.22.0)
**LDC**: 1.22.0

**Issue description:**
"_Unwind edges out of a funclet pad must have the same unwind dest_"
Discovered when attempting to build Apache Thrift - DMD (2.092.1) works fine, LDC2 does not. Reduced test case below. Oh, and the issue is ONLY on Windows - no issues on Mac/Linux.

**Test case(s), stack traces and other relevant information:**
```
class Condition
{
    void notifyAll()
    {
    }
}

struct QueueState {
    Condition connClosed;
}

struct VariantN()
{
    ptrdiff_t function() fptr;
    this()()
    {
    }
    ~this()
    {
        (() => fptr())();
    }
}

alias Variant = VariantN!();

void worker(QueueState queueState)
{
    scope (exit)       synchronized         queueState.connClosed.notifyAll;
    Variant connectionContext;
}
```

**Full error:**
```
Unwind edges out of a funclet pad must have the same unwind dest
  %8 = cleanuppad within none []
  cleanupret from %8 unwind to caller
  invoke x86_vectorcallcc void %17(%combined.Condition* nonnull %18) [ "funclet"(token %8) ]
          to label %19 unwind label %33
in function _D8combined6workerFSQs10QueueStateZv
LLVM ERROR: Broken function found, compilation aborted!
```

kinke:
Thx for the reduced testcase, boiling further down to
```D
void foo();

void bar()
{
    try { foo(); }
    finally
    {
        try { foo(); } finally {}
    }
}
```

@rainers: Do you have time to have a look or too busy with other stuff?

kinke:
AFAICT, the problem is that the cleanup scope is only pushed after emitting the finally block - so the nested try-finally has no idea about the outer `finally` cleanup scope it should unwind to: https://github.com/ldc-developers/ldc/blob/4e36492f7bb40f0a4fb6ede4c88e44bb4f2f1a8e/gen/statements.cpp#L820-L833

rainers:
> @rainers: Do you have time to have a look or too busy with other stuff?

I hardly remember any of this stuff ;-) And yes, I'm rather busy with other stuff, not sure I can give input on this soon.

rainers:
[...]

PR #5113, opened 2026-04-07
Upgrade frontend & libs to v2.113

kinke:


kinke:
On Win64, the first hurdle/regression is pretty wild - the compiler quickly ends up with random garbage alignments (incl. for `real` - alignments over 1000 etc.). Going down that rabbit hole led to the observation that the MSVC++ POD logic is even more curious - not just any ctor makes a struct a non-POD, but apparently also any private (or non-public) *data* symbols (private functions don't trigger this). Which we have in the frontend for the `structalign_t` struct - which we treat as POD on the D side, but C++ wants a non-POD because of the private fields.

So any existing LDC miscompiles the new compiler sources on Win64. We can work around this by adding a ctor to the D `structalign_t` definition for now.

**Edit**: Filed https://github.com/ldc-developers/ldc/issues/5139.

kinke:
The next problem on Win64 is that we're hitting #3504 now when compiling Phobos' `std.parallelism` (not touched itself).

kinke:
On macOS arm64, the single remaining `std.uuid` problem is https://github.com/dlang/phobos/pull/10862#discussion_r3225803119.

For Linux AArch64, it's the usual annoying x87-centric upstream tests for `std.internal.math.gammafunction` that need to be adapted for quadruple precision. And probably some separate AArch64-specific thingy as well (NaN sign propagation?), as the unittests fails on *musl* aarch64 too, with its double-precision real.

The other Posix targets are all green.

kinke:
We can work around (some/most instances of) #3504 for now by defaulting to new `-foptimize-nothrow` (DMD's `-nothrow-optimizations`) for MSVC targets. Are we good with that CLI option name?

rainers:
> We can work around (some/most instances of) #3504 for now by defaulting to new `-foptimize-nothrow` (DMD's `-nothrow-optimizations`) for MSVC targets. Are we good with that CLI option name?

Sounds good to me. I have yet to understand why cleanup in case of Errors is suddenly necessary, while it was never guaranteed before (and only happened to work partially). Conflating Exceptions and Errors was a questionable decision anyway.

kinke:
@ibuclaw: Have you already selected a CLI option name for DMD's `-nothrow-optimizations`? I went with `-foptimize-nothrow` for now.

kinke:
> On macOS arm64, the single remaining `std.uuid` problem is [dlang/phobos#10862 (comment)](https://github.com/dlang/phobos/pull/10862#discussion_r3225803119).

That's the only remaining blocker. If you have such a device, a diagnosis/fix would be appreciated, as doing that via CI would be a pain.

ibuclaw:
> @ibuclaw: Have you already selected a CLI option name for DMD's `-nothrow-optimizations`? I went with `-foptimize-nothrow` for now.

No. There's no tests for it, and I'm not sure when it might be the preferred behaviour.

QuantumSegfault:
> > On macOS arm64, the single remaining `std.uuid` problem is [dlang/phobos#10862 (comment)](https://github.com/dlang/phobos/pull/10862#discussion_r3225803119).
> 
> That's the only remaining blocker. If you have such a device, a diagnosis/fix would be appreciated, as doing that via CI would be a pain.

I've got an Apple Silicon Mac.

I believe I've figured it out. It comes down to the fact that `MonotonicUUIDsFactory` uses `StopWatch`, which internallu stores the `timeElapsed` in system ticks. But the `machTicksPerSecond` differs between arm64 and x86_64.

With `-march=x86-64` (and running via Rosetta)
```
Sys ticks per second: 1000000000
Input hnsecs: 17577131251230003
Sys ticks: 1757713125123000300
Output hnsecs: 17577131251230003
```

As expected. However compiling/running a native arm64 version
```
Sys ticks per second: 24000000
Input hnsecs: 17577131251230003
Sys ticks: 42185115002952007
Output hnsecs: 17577131251230002
```

kinke:
Ah, nice finding, thx!

kinke:
Does it work with `d += dur!"hnsecs"(4);` (instead of 3) perhaps? The 3 hnsecs are about 7.2 ticks of that 24-MHz-sysclock, so probably truncated to 2 hnsecs, while the test apparently requires at least 2.5 hnsecs according to the comment.

**Edit**: CI says yes.

denizzzka:
@kinke seems `subMsecsPart` value need to be changed for MacOS if it have less accuracy of monotonic clock. Especially, 10_000 value should be changed

RFC prescribes to use available sub-millisecond accuracy

kinke:
Yeah, I guess that UUID factory should be based on system ticks, not hnsecs. The 24 MHz of that macOS *arm64* clock aren't any coarser than the 10 MHz hnsecs granularity.

denizzzka:
> Yeah, I guess that UUID factory should be based on system ticks, not hnsecs.

In fact, that's how it's done, it's just that the variable name `const float hnsecs` in `v7Timestamp_method3()` is unfortunate - it just so happened that on all the systems I tried these values ​​were the same and I thought that this is a general rule everywhere

> The 24 MHz of that macOS arm64 clock aren't any coarser than the 10 MHz hnsecs granularity.

We only have access to `hnsecs` in D, not more precise values is available. So, perhaps the problem is in the place where `hnsecs` are read into `StopWatch.peek()` result?

denizzzka:
> In fact, that's how it's done

Almost so, but not quite.

~~But why is it done this way? Because we don't have direct access to the monotonic clock, only through the `StartStop` meter results.~~ I remember that I didn't like that it is need to specify additional argument `ClockType` for `MonoTime`

In any case, this should not have affected the result if MacOS's monotonic clock accuracy is higher.

kinke:
I'm not sure how critical this is in practice. The test is green when adding 4 instead of 3 hnsecs (https://github.com/ldc-developers/phobos/commit/12d555a6d026147180a19fa4a9531050adef3aa2), is that an acceptable solution/workaround for now?

denizzzka:
> is that an acceptable solution/workaround for now?

I think this masks a time value conversion issue somewhere else in druntime or Phobos: if MacOS have highter accuracy (more than 40x!) of monotonic clock so it shouldn't broke code at this place by this way

I would agree if `sp.hnsecs` was 1 or 3 [here](https://github.com/dlang/phobos/pull/10862#pullrequestreview-4271579301), but not zero

kinke:
AFAICT, the truncation happens in the `MonotonicUUIDsFactory(Duration, bool)` ctor used by the unittest, where you initialize the Stopwatch via `setTimeElapsed()`. The 3 hnsecs there are converted to MonoTime ticks, most likely rounded/truncated to 7 (=> about 2.8 hnsecs). And then `peek()`ing the non-started Stopwatch probably truncates those internal 2.8 hnsecs to 2, for `Duration`'s hnsecs precision, which you then use for the sub-ms part.

So when using `core.time.MonoTime` directly instead of a Stopwatch, you could use the ticks instead (which have 2.4x higher resolution on macOS arm64 than `Duration`s hnsecs).

denizzzka:
@kinke ok, agree

I'll try to make new PR to fix this by moving from StartStop to MonoTime.
For now, if you need to do everything quickly, I suggest temporarily disabling v7 method 3 completely for MacOS.

denizzzka:
@QuantumSegfault just to ensure: MacOS Arm64 provides less monotone ticks per second?

QuantumSegfault:
> @QuantumSegfault just to ensure: MacOS Arm64 provides less monotone ticks per second?

As I posted before:

x86_64 (on Apple Silicon over Rosetta): `1000000000`
Native arm64: `24000000`

So yes, less on ARM than on x86.

denizzzka:
Fixes UUID v7 Method 3 processing for MacOS on arm64: https://github.com/dlang/phobos/pull/11065

kinke:
Wrt. `-foptimize-nothrow` (the pre-2.113 behavior):

> I have yet to understand why cleanup in case of Errors is suddenly necessary, while it was never guaranteed before (and only happened to work partially).

> There's no tests for it, and I'm not sure when it might be the preferred behaviour.

Yeah I'm skeptical too; I guess some benchmarks would be in order, to gauge the effect on runtime and code size. So I'm fully open to change the default to enabled (not just on Windows), and making the optimization opt-out instead.

JohanEngelen:
Thanks for all this work @kinke  !!!!!

QuantumSegfault:
Now that this is done, what would be the best way to get some of my changes from DMD `master` here?

Should I open some PRs applying/cherry-picking the changes?

kinke:
You should find the (rewritten) commits of yours in branch `dmd-rewrite-master` (last synced yesterday or so), and could cherry-pick those to new PRs here. **Edit**: https://github.com/ldc-developers/ldc/compare/master...dmd-rewrite-master

See https://github.com/ldc-developers/ldc/blob/master/tools/update-dmd-rewrite.sh for details if interested in the rewrite/sync with DMD upstream.

QuantumSegfault:
> You should find the (rewritten) commits of yours in branch `dmd-rewrite-master` (last synced yesterday or so), and could cherry-pick those to new PRs here.
> 
> See https://github.com/ldc-developers/ldc/blob/master/tools/update-dmd-rewrite.sh for details if interested in the rewrite/sync with DMD upstream.

Perfect! Thank you.

How often does `dmd-rewrite-master` get synced?

kinke:
> How often does dmd-rewrite-master get synced?

I run it manually from time to time (in a Ubuntu 22 container, doesn't work on Ubuntu 24 natively). It needs user interaction from time to time, when new files are added upstream and we need to check whether we need them, so it's not perfectly automatable.

So for now, just ping me whenever you need a new sync.

PR #5115, opened 2026-04-18
ldc: Add support for GNU/Hurd

yelninei:
This adds support for GNU/Hurd to ldc

Depends on 
https://github.com/dlang/dmd/pull/22829
https://github.com/dlang/phobos/pull/10980

There are some inconsistencies as the TARGET_SYSTEM/CMAKE_SYSTEM_NAME is GNU while in D the Hurd name is used everywhere

I was able to build an ldc that wasx  able to rebuild itself from either my gdc and dmd versions on x86_64 Hurd.
Have not tried the tests but atleast in dmd almost all work.

I have not tried 32bit version as my attempt to build llvm there failed.


gen/target.cpp
```diff
@@ -149,6 +149,8 @@ void Target::_init(const Param &params) {
     os = OS_DragonFlyBSD;
   } else if (triple.isOSSolaris()) {
     os = OS_Solaris;
+  else if (triple.isOSHurd()) {
+    os = OS_Hurd;
```
yelninei, line 153: The OS_Hurd variable is from target.h which is not updated in dmd.

In DMD i only regenerated the frontend.h

kinke:
Thx, looks pretty good after a quick glance. I'll take a proper look once we're at D v2.113.

kinke:
(The upstream support was added in v2.114, so this will have to wait.)

PR #5117, opened 2026-04-20
Fix #5114 - Avoid same-symbol qualifier-difference case falling into wrong cast branch

### LINKED ISSUE #5114: Segfault caused by interface containing `const` function with pre/post-condition
As of LDC 1.42.0 (new regression in this version), tested on macOS 14.5 (AArch64), trying to compile a file that contains an interface like this:
```d
interface I{
	void fn() const
	out(; true);
}
```
Causes LDC2 to segfault with a trace like this:
```
0  ldc2                     0x00000001042c6338 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 56
1  ldc2                     0x00000001042c404c llvm::sys::RunSignalHandlers() + 64
2  ldc2                     0x00000001042c6e2c SignalHandler(int, __siginfo*, void*) + 356
3  libsystem_platform.dylib 0x000000018b007584 _sigtramp + 56
4  ldc2                     0x00000001048379a8 ObjCState::getProtocol(InterfaceDeclaration*) + 236
5  ldc2                     0x00000001048379a8 ObjCState::getProtocol(InterfaceDeclaration*) + 236
6  ldc2                     0x000000010483a668 ObjCState::deref(ClassDeclaration*, llvm::Type*) + 56
7  ldc2                     0x000000010480388c DtoDynamicCastInterface(Loc, DValue*, Type*) + 396
8  ldc2                     0x000000010480353c DtoCastClass(Loc, DValue*, Type*) + 1160
9  ldc2                     0x00000001048622bc ImplicitArgumentsBuilder::addContext(bool) + 328
10 ldc2                     0x0000000104860a1c DtoCallFunction(Loc, Type*, DValue*, Array<Expression*>*, llvm::Value*, bool) + 1144
11 ldc2                     0x0000000104876bf0 ToElemVisitor::call(IRState*, CallExp*, llvm::Value*) + 1076
12 ldc2                     0x000000010486a5cc ToElemVisitor::visit(CallExp*) + 28
13 ldc2                     0x000000010485b430 ToIRVisitor::visit(ExpStatement*) + 400
14 ldc2                     0x000000010485b5ec ToIRVisitor::visit(CompoundStatement*) + 260
15 ldc2                     0x0000000104855668 ToIRVisitor::visit(LabelStatement*) + 1304
16 ldc2                     0x000000010485b5ec ToIRVisitor::visit(CompoundStatement*) + 260
17 ldc2                     0x000000010481adc4 DtoDefineFunction(FuncDeclaration*, bool) + 5696
18 ldc2                     0x000000010480a984 CodegenVisitor::visit(InterfaceDeclaration*) + 396
19 ldc2                     0x000000010482bc40 codegenModule(IRState*, Module*) + 3288
20 ldc2                     0x00000001048ad510 ldc::CodeGenerator::emit(Module*) + 220
21 ldc2                     0x00000001048cd5d8 codegenModules(Array<Module*>&) + 944
22 ldc2                     0x00000001047cecd4 mars_tryMain(Param&, Array<char const*>&) + 1684
23 ldc2                     0x00000001048cc2d0 cppmain() + 6332
24 ldc2                     0x00000001044babac _D2rt6dmain212_d_run_main2UAAamPUQgZiZ6runAllMFZv + 128
25 ldc2                     0x00000001044ba8d0 _d_run_main2 + 484
26 ldc2                     0x00000001044ba6d0 _d_run_main + 144
27 ldc2                     0x00000001048ca9bc main + 732
28 dyld                     0x000000018ac4e0e0 start + 2360
Segmentation fault: 11
```

Pre-conditions and post-conditions cause the same segfault. The return type, parameters of the function, and contents of the pre/post-condition seem to be irrelevant. This issue does not occur with `abstract` functions in a `class` (or functions in an `abstract class`). The segfault does not occur if `const` *or* the pre/post-condition are removed.

gulugulubing:
It is related to #5114. The root cause is  the const(I) -> I contract-context conversion always called DtoCastClass, then due to I.isBaseOf(I, ...) return false, the program accidently fall into the  DtoDynamicCastInterface which is ObjC-only and ends in unreachable for non-ObjC.

So I added an early guard for same-symbol qualifier-difference case there.

The patch is only in tocall.cpp, while objcgen.cpp and classes.cpp are modified just because of removing trailing blank.

### DIFF
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -279,7 +279,7 @@ DValue *DtoCastClass(Loc loc, DValue *val, Type *_to) {
 
   // Qualifier-only casts are semantic no-ops; avoid dynamic cast routing.
   if (dmd::equivalent(from, to)) {
-    Logger::println("qualifier-only cast");
+    IF_LOG Logger::println("qualifier-only cast");
     return new DImValue(_to, DtoRVal(val));
   }
 
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -335,7 +335,7 @@ bool DtoIsObjcLinkage(Type *_to) {
     DtoResolveClass(to->sym);
     return to->sym->classKind == ClassKind::objc;
   }
-  
+
   return false;
 }
 
@@ -391,7 +391,7 @@ DValue *DtoDynamicCastInterface(Loc loc, DValue *val, Type *_to) {
   // In this case we want to call the Objective-C runtime to first
   // get a Class object from the `id`.
   // Then check if class_conformsToProtocol returns true,
-  // if it does, then we can cast and return the casted value, 
+  // if it does, then we can cast and return the casted value,
   // otherwise return null.
   if (DtoIsObjcLinkage(_to)) {
     llvm::Function *getClassFunc =
@@ -400,10 +400,10 @@ DValue *DtoDynamicCastInterface(Loc loc, DValue *val, Type *_to) {
     llvm::Function *kindOfProtocolFunc =
       getRuntimeFunction(loc, gIR->module, "class_conformsToProtocol");
 
-    // id -> Class 
+    // id -> Class
     LLValue *obj = DtoRVal(val);
     LLValue *objClass = gIR->CreateCallOrInvoke(getClassFunc, obj);
-    
+
     // Get prototype_t handle
     LLValue *protoTy = getNullPtr();
     if (auto ifhndl = _to->isClassHandle()->isInterfaceDeclaration()) {
@@ -413,7 +413,7 @@ DValue *DtoDynamicCastInterface(Loc loc, DValue *val, Type *_to) {
     // Class && kindOfProtocolFunc(Class) ? id : null
     LLValue *ret = gIR->ir->CreateSelect(
       gIR->CreateCallOrInvoke(kindOfProtocolFunc, objClass, protoTy),
-      obj, 
+      obj,
       getNullPtr()
     );
     return new DImValue(_to, ret);
--- a/gen/classes.cpp
+++ b/gen/classes.cpp
@@ -277,6 +277,12 @@ DValue *DtoCastClass(Loc loc, DValue *val, Type *_to) {
   Type *from = val->type->toBasetype();
   TypeClass *fc = static_cast<TypeClass *>(from);
 
+  // Qualifier-only casts are semantic no-ops; avoid dynamic cast routing.
+  if (dmd::equivalent(from, to)) {
+    Logger::println("qualifier-only cast");
+    return new DImValue(_to, DtoRVal(val));
+  }
+
   // copy DMD logic:
   // if to isBaseOf from with offset:   (to ? to + offset : null)
   // else if from is C++ and to is C++:  to
--- a/gen/tocall.cpp
+++ b/gen/tocall.cpp
@@ -727,14 +727,29 @@ private:
           //  class pointer
           Type *thistype = gIR->func()->decl->vthis->type;
           if (thistype != iface->type) {
-            DImValue *dthis = new DImValue(thistype, DtoLoad(DtoType(thistype),thisptrLval));
-            thisptrLval = DtoAllocaDump(DtoCastClass(loc, dthis, iface->type));
+            auto thisVal = DtoLoad(DtoType(thistype), thisptrLval);
+            auto thisTypeClass = thistype->toBasetype()->isTypeClass();
+            auto ifaceTypeClass = iface->type->toBasetype()->isTypeClass();
+            const bool sameInterfaceSymbol =
+                thisTypeClass && ifaceTypeClass &&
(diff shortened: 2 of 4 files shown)


thewilsonator (commented):
please remove the whitespace changes

gen/tocall.cpp
```diff
@@ -727,14 +727,29 @@ class ImplicitArgumentsBuilder {
...
+                thisTypeClass->sym == ifaceTypeClass->sym;
+
+            if (sameInterfaceSymbol) {
+              // Qualifier-only interface casts don't require dynamic cast routing.
+              thisptrLval = DtoAllocaDump(
+                  new DImValue(iface->type,
+                               DtoBitCast(thisVal, DtoType(iface->type))));
+            } else {
+              DImValue dthis(thistype, thisVal);
+              thisptrLval = DtoAllocaDump(DtoCastClass(loc, &dthis, iface->type));
```
kinke: I'd say it's better to handle an unnecessary cast in `DtoCastClass()`, there might be more such cases. `Type::equivalent(Type*)` might be enough to discard the qualifiers, without having to check the interface `sym`.

gulugulubing: Yeah, it's much better, thanks.

gulugulubing:
> please remove the whitespace changes

OK, is it accpetable:
For files with actual content changes, I kept the trailing whitespace cleanup (since we're already modifying these files)
For files with only whitespace changes, I reverted them completely to avoid unnecessary diffs.

gen/tocall.cpp
```diff
@@ -727,14 +727,15 @@ class ImplicitArgumentsBuilder {
           //  class pointer
           Type *thistype = gIR->func()->decl->vthis->type;
           if (thistype != iface->type) {
-            DImValue *dthis = new DImValue(thistype, DtoLoad(DtoType(thistype),thisptrLval));
-            thisptrLval = DtoAllocaDump(DtoCastClass(loc, dthis, iface->type));
+            auto thisVal = DtoLoad(DtoType(thistype), thisptrLval);
+            DImValue dthis(thistype, thisVal);
+            thisptrLval = DtoAllocaDump(DtoCastClass(loc, &dthis, iface->type));
```
JohanEngelen: looks like no change was made here, except for the dangerous-looking change from heap allocated dthis to stack allocated. Possibly use-after-scope. Revert this change?

tests/codegen/gh5114.d
```diff
@@ -0,0 +1,79 @@
+module tests.codegen.gh5114;
+
+// During interface contract context setup, a qualifier-only cast for the same
+// interface symbol (e.g., `const(I)` -> `I`) must be handled as a repaint/
+// bitcast and must not route through dynamic interface cast lowering.
+//
+// True dynamic interface casts (different interface symbols) are still valid
+// and are covered below.
+//
+// RUN: %ldc -c %s
```
JohanEngelen: nitpick: line 11 already tests compilation, so you can remove line 10

gen/tocall.cpp
gulugulubing: Does the original heap allocated dthis cause potential leak-like lifetime?

JohanEngelen: that's for you to prove ;)
A mem leak is much more benign than a use-after-scope bug.

gulugulubing: In the original heap version, there are no deallocation in the caller and callee, so I think there is a mem leak.

And use-after-scope does not happen, because DtoCastClass only consumes the pointer immediately and does not store it.

At last, I am a little confused here: do you mean the original code intentionlly choose mem-leak to avoid use-after-scope?

gulugulubing: In classes.cpp, there is similar stack allocation:
```d
    // get class ptr
    LLValue *v = DtoRVal(val);
    // cast to size_t
    v = gIR->ir->CreatePtrToInt(v, DtoSize_t(), "");
    // cast to the final int type
    DImValue im(Type::tsize_t, v);
    return DtoCastInt(loc, &im, _to);
```
So I am not sure which is best practise here.

gulugulubing: I just realized whether heap or stack allocation is not too much related to this issue, so I reverted it first.

gen/classes.cpp
```diff
@@ -277,6 +277,12 @@ DValue *DtoCastClass(Loc loc, DValue *val, Type *_to) {
   Type *from = val->type->toBasetype();
   TypeClass *fc = static_cast<TypeClass *>(from);
 
+  // Qualifier-only casts are semantic no-ops; avoid dynamic cast routing.
+  if (dmd::equivalent(from, to)) {
+    Logger::println("qualifier-only cast");
```
thewilsonator: this should be guarded by `IF_LOG` as in 
```c++
IF_LOG Logger::println("qualifier-only cast");
```

thewilsonator (commented):
otherwise looks good.

thewilsonator (approved):

PR #5118, opened 2026-04-25
Initial Metal backend support to dcompute

asindarov:


gen/dcompute/targetMetal.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+    // std::string targTripleStr = is64 ? SPIR_TARGETTRIPLE64
+    //                                   : SPIR_TARGETTRIPLE32;
+    #if LDC_LLVM_VER >= 2100
+        llvm::Triple targTriple = llvm::Triple(tripleString);
+    #else
+        std::string targTriple = tripleString;
+    #endif
+        _ir->module.setTargetTriple(targTriple);
+
```
thewilsonator:
```
/home/runner/work/ldc/ldc/gen/dcompute/targetMetal.cpp:58:37: error: cannot convert ‘std::string’ {aka ‘std::__cxx11::basic_string<char>’} to ‘llvm::Triple’
   58 |         _ir->module.setTargetTriple(targTriple);
```

gulugulubing:
In cmakelist, LDC_LLVM_VER was removed. Try LLVM_VERSION_MAJOR.

gen/dcompute/targetMetal.cpp
```diff
@@ -0,0 +1,203 @@
...
+            "-v16:16:16-v24:32:32-v32:32:32-v48:64:64-v64:64:64-v96:128:128-v128:128:128-v192:256:256-v256:256:256-"
+            "v512:512:512-v1024:1024:1024-n8:16:32";
+
+        _ir->module.setDataLayout(is64 ? dataLayout: /* TODO: need to find 32-bit data layout */dataLayout);
+        _ir->dcomputetarget = this;
+  }
+
+  void addMetadata() override {
+    llvm::NamedMDNode *airVersion = _ir->module.getOrInsertNamedMetadata("air.version");
+    llvm::Metadata *major = llvm::ConstantAsMetadata::get(llvm::ConstantInt::get(llvm::IntegerType::get(ctx, 32), 2));
```
thewilsonator: add a method to `TargetMetal` that is 
```c++
llvm::Metadata *metaInt(int a) {
  return llvm::ConstantAsMetadata::get(llvm::ConstantInt::get(llvm::IntegerType::get(ctx, 32), a));
}
```
and then use it throughout. might want to do the same for string literals.

gen/dcompute/targetMetal.cpp
```diff
@@ -0,0 +1,203 @@
...
+//
+// This file is distributed under the BSD-style LDC license. See the LICENSE
+// file for details.
+//
+//===----------------------------------------------------------------------===//
+#if LDC_LLVM_SUPPORTED_TARGET_AArch64
+
+#include "declaration.h"
+#include "gen/dcompute/druntime.h"
+#include "mtype.h"
```
thewilsonator, line 15: `dmd/mtype.h` and sort the includes. Also are all of them required?

gen/abi/metal.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "dmd/identifier.h"
+#include "dmd/nspace.h"
+#include "gen/abi/abi.h"
+#include "gen/abi/generic.h"
+#include "gen/dcompute/druntime.h"
+#include "gen/logger.h"
+#include "gen/uda.h"
+#include "ir/irfuncty.h"
+#include "gen/dcompute/abi-rewrites.h"
+#include "mtype.h"
```
thewilsonator: `dmd/mtype.h`, also sort the includes.

gen/abi/abi.cpp
```diff
@@ -22,8 +22,10 @@
 #include "gen/llvmhelpers.h"
 #include "gen/logger.h"
 #include "gen/tollvm.h"
+#include "dmd/globals.h"
 #include "ir/irfunction.h"
 #include "ir/irfuncty.h"
+#include "llvm/TargetParser/Triple.h"
```
thewilsonator: are these includes needed?

driver/dcomputecodegenerator.cpp
```diff
@@ -11,6 +11,7 @@
 #include "driver/cl_options.h"
 #include "driver/cl_helpers.h"
 #include "dmd/errors.h"
+#include "dmd/globals.h"
```
thewilsonator: is this include needed?

gen/abi/abi.cpp
```diff
@@ -286,6 +288,17 @@ TargetABI *TargetABI::getTarget() {
...
+  case llvm::Triple::UnknownArch:
+    if (global.params.targetTriple->getArchName() == "air64") {
+      return createMetalABI();
+    }
+
+    warning(Loc(),
+            "unknown target ABI, falling back to generic implementation. C/C++ "
+            "interop will almost certainly NOT work.");
+    return new TargetABI;
+
```
thewilsonator: this is the same as the code below, just fall through and add a comment `//fallthrough`

gen/abi/targets.h
```diff
@@ -40,3 +40,5 @@ TargetABI *getX86TargetABI();
 TargetABI *getLoongArch64TargetABI();
 
 TargetABI *getWasmTargetABI();
+
+auto createMetalABI() -> TargetABI*;
```
thewilsonator: please write this as `TargetABI* createMetalABI();`

gen/abi/metal.cpp
```diff
@@ -0,0 +1,67 @@
...
+                pointerRewite.applyTo(arg);
+            }
+
+        if (ty->isScalar()) {
+          metalScalarRewrite.applyTo(arg);
+        }
+    }
+};
+
+auto createMetalABI() -> TargetABI* { return new MetalABI(); }
```
thewilsonator: ditto

gen/dcompute/druntime.h
```diff
@@ -38,4 +38,4 @@ struct DcomputePointer {
     return LLPointerType::get(getGlobalContext(), as);
   }
 };
-std::optional<DcomputePointer> toDcomputePointer(StructDeclaration *sd);
+auto toDcomputePointer(StructDeclaration *sd) -> std::optional<DcomputePointer>;
```
thewilsonator: ditto

gen/dcompute/target.h
```diff
@@ -58,6 +58,10 @@ class DComputeTarget {
 DComputeTarget *createCUDATarget(llvm::LLVMContext &c, int sm);
 #endif
 
+#if LDC_LLVM_SUPPORTED_TARGET_AArch64
+auto createMetalTarget(llvm::LLVMContext &c, int version) -> DComputeTarget *;
```
thewilsonator: ditto

gen/dcompute/targetMetal.cpp
```diff
@@ -0,0 +1,203 @@
...
+      : DComputeTarget(
+            c, version, ID::Metal, "metal", "air", createMetalABI(),
+
+            // DCompute Order: [Private, Global, Shared, Constant, Generic]
+            // AIR equivalents: Private=0, Device/Global=1, Threadgroup/Shared=3, Constant=2
+            {{0, 1, 3, 2, 0}}) {
+    const bool is64 = global.params.targetTriple->isArch64Bit();
+
+    _ir = new IRState("dcomputeTargetMetal", ctx);
+    // TODO: need to find 32-bit triple
```
thewilsonator: apple has shipped 64-bit stuff for a very long time, does metal even do a 32-bit mode? Dcompute is only in 32-bit if the host is built for 32-bit, and Apple hosts have been 64-bit for a very long time.

gen/dcompute/targetMetal.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+    argumentMetadata.push_back(llvm::MDString::get(ctx, "air.arg_type_name"));
+    // TODO: check if using char needed instead of int8 as in ocl target implementation
+    argumentMetadata.push_back(llvm::MDString::get(ctx, basicTypeToString(type)));
+
+    argumentMetadata.push_back(llvm::MDString::get(ctx, "air.arg_name"));
+    argumentMetadata.push_back(llvm::MDString::get(ctx, vd->ident->toChars()));
+  }
+
+  auto basicTypeToString(Type *t) -> std::string {
```
thewilsonator, line 199: this doesn't appear to be used, also its `#includes` can be removed.

runtime/druntime/src/ldc/dcompute.d
```diff
@@ -14,6 +14,7 @@ enum ReflectTarget : uint
     Host = 0,
     OpenCL = 1,
     CUDA = 2,
+    Metal = 3,
```
thewilsonator, line 17: all the rest of the changes in this file are not necessary

gen/dcompute/targetMetal.cpp
```diff
@@ -0,0 +1,203 @@
...
+      ss << t->toChars();
+    }
+
+    return ss.str();
+  }
+
+};
+} // anonymous namespace.
+
+auto createMetalTarget(llvm::LLVMContext &c, int version) -> DComputeTarget * {
```
thewilsonator: `DComputeTarget *createMetalTarget(llvm::LLVMContext &c, int version)`

gen/dcompute/targetMetal.cpp
asindarov: It is used while filling up the metadata information about the type name. 
```
argumentMetadata.push_back(
      metaString("air.arg_type_name"));
argumentMetadata.push_back(
      metaString(basicTypeToString(type)));
```

thewilsonator: oh, OK.

gen/abi/abi.cpp
asindarov: yes, I have added `// fallthrough` comment

gen/dcompute/abi-rewrites.h
```diff
@@ -32,3 +33,19 @@ struct DComputePointerRewrite : ABIRewrite {
     return ptr->toLLVMType(true);
   }
 };
+
+struct DcomputeMetalScalarRewrite : ABIRewrite {
```
thewilsonator, line 36: Is this now unused?

gen/dcompute/abi-rewrites.h
```diff
@@ -14,6 +14,7 @@
 #pragma once
 
 #include "gen/abi/generic.h"
+#include "gen/dcompute/druntime.h"
```
thewilsonator: is this needed?

driver/dcomputecodegenerator.cpp
```diff
@@ -59,6 +67,7 @@ DComputeCodeGenManager::createComputeTarget(const std::string &s) {
 #endif
   }
 
+
```
thewilsonator: stray whitespace change

runtime/druntime/src/ldc/dcompute.d
```diff
@@ -24,7 +25,6 @@ enum ReflectTarget : uint
  * valid values of _version are for OpenCL 100 110 120 200 210
  * and for CUDA are x*100 + y*10 for x any valid values of sm x.y
  * use 0 as a wildcard to match any version.
-
```
thewilsonator, line 27: ditto, also could you update the documentation here to specify what are the valid versions of Metal (e.g. 400)?

thewilsonator (commented):
Looking good. Needs some tests.

runtime/druntime/src/ldc/dcompute.d
```diff
@@ -24,7 +24,9 @@ enum ReflectTarget : uint
  * arguments MUST be compiletime constants
  * valid values of _version are for OpenCL 100 110 120 200 210
  * and for CUDA are x*100 + y*10 for x any valid values of sm x.y
+ * and for Metal is 4.0.0
```
thewilsonator: This bit of documentation refers to the integer parameter passed to a function, e.g. 400 for metal 4.0.0. This should also be tested

gen/dcompute/abi-rewrites.h
asindarov: it still used in `abi/metal.cpp` file:
```
struct MetalABI : TargetABI {
    DComputePointerRewrite pointerRewite;
    DcomputeMetalScalarRewrite metalScalarRewrite;
.....
```

It is used to make scalar objects turn into a pointer in constant memory address space. Although it turns out scalars can also be stored in device level address space which will be `Global` in the dcompute term as I understand

thewilsonator: scalar arguments to kernels are just passed as parameters

driver/toobj.cpp
```diff
@@ -73,6 +72,57 @@ void codegenModule(llvm::TargetMachine &Target, llvm::Module &m,
...
+        fatal();
+      }
+      
+      llvm::WriteBitcodeToFile(m, out.os());
+
+      out.keep();
+
+      // Terminate upon errors during the LLVM passes.
+      if (global.errors || global.warnings) {
+        Logger::println("Aborting because of errors/warnings during LLVM passes");
```
thewilsonator: this should be an error not a logger print

driver/toobj.cpp
```diff
@@ -468,5 +518,5 @@ void writeModule(llvm::Module *m, const char *filename) {
     if (useIR2ObjCache) {
       cache::cacheObjectFile(filename, moduleHash);
     }
-  }
```
thewilsonator, line 471: whitespace

dmd/globals.h
```diff
@@ -393,6 +393,7 @@ const DString ll_ext = "ll";
 const DString mlir_ext = "mlir";
 const DString bc_ext = "bc";
 const DString s_ext = "s";
+const DString metallib_ext = "metallib";
```
thewilsonator: this seems unused

tests/codegen/dcompute_metal_argument_metadata.d
```diff
@@ -0,0 +1,33 @@
+// REQUIRES: target_AArch64
+// REQUIRES: llvm20
```
JohanEngelen: this should be an "atleast" requirement

tests/codegen/dcompute_metal_argument_metadata.d
```diff
@@ -0,0 +1,33 @@
...
+
+// RUN: %ldc -c -mdcompute-targets=metal-400 -output-ll -of=%t.ll %s
+// RUN: FileCheck %s --check-prefix=AIR < kernels_metal400_64.air
```
JohanEngelen: I don't quite see how this filename is generated, but i guess it is based on the name of the module?  Better to give it a unique name (across tests) so there can never be a weird race condition between tests executed in parallel.

tests/codegen/dcompute_metal_argument_metadata.d
asindarov: Unfortunately it should be `REQUIRES` because currently there are outstanding issues with the new changes in llvm versions after `llvm20`. Therefore current changes work only with `llvm20`.

asindarov: the issue was in newer llvm versions  some attributes got changed. [Here](https://discourse.llvm.org/t/rfc-improvements-to-capture-tracking/81420) you can find `nocapture` got changed to `captures`. Apple llvm bitcode writer does not know `captures` attribute's existence, it expects `nocapture` instead.

tests/codegen/dcompute_metal_argument_metadata.d
asindarov: Good catch, thanks for the review. I have used default file name that would be generated. I now have changed it using `--mdcompute-file-prefix=<prefix>`

tests/codegen/dcompute_metal_argument_metadata.d
JohanEngelen: i don't understand. Which LLVM bitcode writer?
Upon changing the attribute, ofcourse the bitcode writer is also updated to work with that?

driver/toobj.cpp
```diff
@@ -73,6 +107,63 @@ void codegenModule(llvm::TargetMachine &Target, llvm::Module &m,
 #endif
   }
 
+#ifdef LDC_LLVM_SUPPORTED_TARGET_AArch64
+  if (cb == ComputeBackend::METAL) {
+#ifdef __APPLE__
```
JohanEngelen: This ifdef would mean LDC cannot produce metal bitcode output on other platforms, and instead will silently not create any output.

I think this code should be included on any LDC build (it is a cross-compiler after all). `xcrun` would not be found, perhaps make that a warning instead of a fatal error. Generating the object file could still be useful, no?

driver/toobj.cpp
```diff
@@ -73,6 +107,63 @@ void codegenModule(llvm::TargetMachine &Target, llvm::Module &m,
...
+    std::vector<std::string> args = {
+      xcrunpath.get(), "-sdk", "macosx", "metallib", filename, "-o", metallibOutPath.c_str()
+    };
+
+    std::string errorMsg;
+
+    int status =  executeToolAndWait(Loc(), args[0], args);
+
+    if (status < 0) {
+      error(Loc(), "program received signal %d (%s)", -status,
```
JohanEngelen: "program" can be very ambiguous here. Why not call it "xcrun" ?

tests/codegen/dcompute_metal_argument_metadata.d
JohanEngelen: I guess you mean that we have to match the test with the available xcrun version on the tester. Can we rewrite the test to not rely on running xcrun? Just use the outputted object file (that is passed to xcrun) instead?

JohanEngelen: (we don't need to test xcrun, we need to test our output that goes into xcrun, right?)

asindarov: Sorry I have not provided more context on this. In the upstream LLVM there is no official Apple Metal backend target, I reverse engineered the necessary metadata that Apple Metal Shading Language compiler (which is old fork of LLVM with adjustments by Apple) expects for the kernel functions. The pipeline is as follows:
```
DCompute kernel code ->  LLVM IR bitcode with reverse engineered metadata ->
                         xcrun -sdk macosx metallib .... -> 
                         generates <filename>.metallib binary file
```

We use that `<filename>.metallib` binary file to execute kernels at the dcompute host code part as follows:
```
...
        auto program = Program.fromFile(device, "./kernels_metal400_64.metallib");

        Program.globalProgram = program;
        
        auto deviceX = device.makeBuffer!float(x);
        auto deviceY = device.makeBuffer!float(y);
        auto deviceRes = device.makeBuffer!float(res);

        auto queue = Queue(device);

        queue.enqueue!(saxpy)
                ([N,1,1],[1,1,1])
                (deviceRes, alpha, deviceX, deviceY, N);

        queue.finish();
...
```

The bitcode issue is that `xcrun` uses Apple toolchain which is closed source LLVM fork. Current changes offer basic support. What I am thinking to do afterwards is to explore to write custom bitcode writer as done in Julia compiler (https://github.com/JuliaLLVM/llvm-downgrade/blob/main/src/BitcodeWriter140.cpp). They mostly copy & pasted older llvm bitcode writer and adjusted a bit for their codegen. They also did similar changes to emit correct metadata as in this PR. (https://github.com/JuliaGPU/GPUCompiler.jl/blob/main/src/metal.jl)

JohanEngelen: It's just that this is quite flaky... and this test will start to fail when the system xcode version changes.

asindarov: nope, it will not. In the test I only tested what would be generated as LLVM IR from ldc compiler not the output of `xcrun`.

asindarov: all the metadata is added at the ldc side and llvm bitcode is emitted and sent over to xcrun afterwards. Tests in this PR only tested the emitted metadata from ldc side only. But once `xcrun -sdk macosx metallib` compiler changes what it expects from the Apple Metal Shading language compiler (closed old fork of LLVM), only then we might have to adjust the metal target metadata codegen part in ldc and these tests.

driver/toobj.cpp
asindarov: I have renamed it to "xcrun"

driver/toobj.cpp
asindarov: Yes, agreed. I have removed the `#if` directive and replaced fatal error with warning.

tests/codegen/dcompute_metal_argument_metadata.d
asindarov: Now that I thought about it more, I think you were right that I should've tested our output that goes into `xcrun`. I have updated the requirement as suggested.

asindarov:
Hello @thewilsonator @JohanEngelen. I wanted to provide an architectural update on how the final device code generation is now working.

After researching how other Metal GPU backends (Julia's Metal.jl, Mojo compiler) handle LLVM bitcode mismatch problem, I have adjusted the compilation pipeline slightly that resembles how Julia compiler does generate Metal GPU device code. Because there is no backwards-compatibility on LLVM bitcode writers, current LLVM bitcode writers cannot emit Apple-specific LLVM bitcode directly from LDC. To resolve this Julia compiler maintainers created a utility tool that downgrades newer version of LLVM emitted bitcode into older LLVM bitcode that Apple toolchain expects. I have created wrapper library to use that as done in Julia compiler ecosystem i.e: 
* [LLVMDowngrader_jll.jl](https://github.com/JuliaBinaryWrappers/LLVMDowngrader_jll.jl).
* D alternative I created:  [llvm_downgrader](https://github.com/asindarov/llvm_downgrader). 
* [llvm-downgrade](https://github.com/JuliaLLVM/llvm-downgrade) utility created by Julia compiler maintainers.

Compilation pipeline now works as follows:

1. **Frontend**: LDC compiler emits LLVM IR with Apple-specific metadata.
2. **Downgrade**: DCompute library takes the IR and passes it through `llvm_downgrader` library function I created (similar to how Julia's Metal.jl does [here](https://github.com/JuliaGPU/Metal.jl/blob/1323fcd5bf808bf7dbf1cae9d48159f1a059bf16/bin/metallib-as#L84-L95)) which downgrades given LLVM IR bitcode into older LLVM IR bitcode. Currently it supports lowering from LLVM version 22.
3. **Linking**: Finally, we invoke `xcrun metallib` linker to emit `<filepath>.metallib` binary. It is done at [here](https://github.com/asindarov/dcompute/blob/584160a531640fadfb6653b1d4f5c504daf5a9c3/source/dcompute/driver/metal/program.d#L68C5-L87C6)


Here is how high-level compilation pipeline looks:
```text
        [ D Device Code ]
               │
               ▼ (LDC Compiler)
[ LLVM IR + Apple-specific Metadata ]
               │
               ▼ (llvm_downgrader at the DCompute side)
[ Apple-compatible LLVM Bitcode (AIR) ]
               │
               ▼ (xcrun metallib executed at the DCompute side)
      [ .metallib Binary ]
```

[Here](https://github.com/asindarov/dcompute/blob/584160a531640fadfb6653b1d4f5c504daf5a9c3/source/dcompute/driver/metal/program.d#L42-L52) is the link to the linker integration part in my DCompute PR for the host side runtime support.

PR #5121, opened 2026-04-29
use `getTerminatorOrNull` for  Windows EH with LLVM23

gnavdev28:
This PR contains two related fixes for building and running LDC with LLVM trunk on Windows:

**1. MSVC Build Fix:**
- Fixed compilation errors in `cl_options-llvm.cpp` and `llvmhelpers.cpp` to ensure compatibility with LLVM trunk.

**2. cloneBlocks Segfault Fix:**
- Fixed an `0xC0000005` Access Violation in `cloneBlocks` when compiling code with `try-finally` cleanups.
- Added a defensive check `term->isTerminator()` in `ms-cxx-helper.cpp` to prevent crashing when the frontend generates a basic block without a proper terminator (e.g., ending abruptly with a `call` to `__dtor`).

### DIFF
--- a/gen/ms-cxx-helper.cpp
+++ b/gen/ms-cxx-helper.cpp
@@ -96,7 +96,7 @@ void cloneBlocks(const std::vector<llvm::BasicBlock *> &srcblocks,
   for (auto bb : srcblocks) {
     llvm::Function *F = bb->getParent();
     auto nbb = llvm::BasicBlock::Create(bb->getContext(), bb->getName());
-
+    // Loop over all instructions, and copy them over.
     for (auto &II : *bb) {
       llvm::Instruction *Inst = &II;
       llvm::Instruction *newInst = nullptr;
@@ -117,7 +117,7 @@ void cloneBlocks(const std::vector<llvm::BasicBlock *> &srcblocks,
         newInst = Inst->clone();
 
       newInst->insertInto(nbb, nbb->end());
-      VMap[Inst] = newInst;
+      VMap[Inst] = newInst; // Add instruction map to value.
       if (unwindTo)
         if (auto dest = getUnwindDest(Inst))
           VMap[dest] = unwindTo;
--- a/gen/ms-cxx-helper.cpp
+++ b/gen/ms-cxx-helper.cpp
@@ -83,7 +83,6 @@ void cloneBlocks(const std::vector<llvm::BasicBlock *> &srcblocks,
                  llvm::BasicBlock *continueWith, llvm::BasicBlock *unwindTo,
                  llvm::Value *funclet) {
   llvm::ValueToValueMapTy VMap;
-  
   // map the terminal branch to the new target
   if (continueWith) {
     if (auto term = srcblocks.back()->getTerminator()) {
@@ -119,7 +118,6 @@ void cloneBlocks(const std::vector<llvm::BasicBlock *> &srcblocks,
 
       newInst->insertInto(nbb, nbb->end());
       VMap[Inst] = newInst;
-      
       if (unwindTo)
         if (auto dest = getUnwindDest(Inst))
           VMap[dest] = unwindTo;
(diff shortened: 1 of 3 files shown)


driver/cl_options-llvm.cpp
```diff
@@ -72,7 +72,7 @@ std::string getFeaturesStr() {
 
 void setFunctionAttributes(StringRef cpu, StringRef features,
                            Function &function) {
-  return codegen::setFunctionAttributes(cpu, features, function);
+  return codegen::setFunctionAttributes(function, cpu, "", features);
```
thewilsonator: see https://github.com/ldc-developers/ldc/pull/5120/changes#diff-fb3fcb2c35281341f5bc3e3455019cecc149f960823ba626e3f0f94ae8de8c60R75-R77

gen/llvmhelpers.cpp
```diff
@@ -1364,7 +1364,7 @@ void printLabelName(std::ostream &target, const char *func_mangle,
                     const char *label_name) {
   // note: quotes needed for Unicode
   target << '"'
-         << gTargetMachine->getMCAsmInfo()->getPrivateGlobalPrefix().str()
+         << gTargetMachine->getMCAsmInfo()->getPrivateLabelPrefix().str()
```
thewilsonator, line 1368: see that other PR

kinke:
Where did you hit this problem - when building druntime/Phobos already, or for an unrelated code base? In the first case, it might be an LLVM trunk regression/change (e.g., not returning null if there aren't any successors - which could still be fixed for LLVM 23, it wasn't even branched off yet). In the 2nd case, a testcase is needed.

thewilsonator:
When building druntime/phobos.

Alas I don't have a windows system, I'll ask on the LLVM discord and see what the consensus is.

thewilsonator:
> getTerminator() now has a precondition that a terminator exists. You can use getTerminatorOrNull() if that doesn't hold

kinke:
Oof; thx for asking. We have more usages incl. null checks.

thewilsonator (approved):

gen/ms-cxx-helper.cpp
```diff
@@ -84,10 +84,16 @@ void cloneBlocks(const std::vector<llvm::BasicBlock *> &srcblocks,
...
-      if (auto succ = term->getSuccessor(0))
-        VMap[succ] = continueWith;
+  if (continueWith) {
+#if LLVM_VERSION_MAJOR >= 23
+    auto term = srcblocks.back()->getTerminatorOrNull();
+#else
+    auto term = srcblocks.back()->getTerminator();
+#endif
+    if (term && term->getNumSuccessors() > 0) {
+      VMap[term->getSuccessor(0)] = continueWith;
```
thewilsonator: @gnavdev28 does 
```c++
if (term)
  if (auto succ = term->getSuccessor(0))
        VMap[succ] = continueWith;
```
work here? Would be less changes and less calls.

thewilsonator (approved):
Otherwise looks great.

kinke (dismissed):
As said, there are more usages.

gen/trycatchfinally.cpp
```diff
@@ -331,7 +331,11 @@ llvm::BasicBlock::iterator getTerminatorPos(llvm::BasicBlock *bb) {
 }
 #else
 llvm::Instruction *getTerminatorPos(llvm::BasicBlock *bb) {
+#if LLVM_VERSION_MAJOR >= 23
+  return bb->getTerminatorOrNull();
+#else
   return bb->getTerminator();
+#endif
```
thewilsonator: this one is not needed to be updated because it is in the `#else` branch of `#if LLVM_VERSION_MAJOR >= 19`

gnavdev28: tks for clarifying

gen/passes/GarbageCollect2Stack.cpp
```diff
@@ -611,7 +611,11 @@ static bool mayBeUsedAfterRealloc(Instruction *Def, BasicBlock::iterator Alloc,
 
     // All instructions after the starting point in this block have been
     // accounted for. Look for successors to add to the work list.
+#if LLVM_VERSION_MAJOR >= 23
+    auto *Term = B->getTerminatorOrNull();
+#else
     auto *Term = B->getTerminator();
+#endif
     unsigned SuccCount = Term->getNumSuccessors();
```
kinke, line 619: Here (and in other sites) the breaking change for the existing API would have actually made sense, because we don't check for null afterwards, assuming that LLVM trunk introduced an appropriate assertion.

PR #5124, opened 2026-05-04
update-dmd-rewrite.sh: Deal with frontend headers having moved to compiler/include/

kinke:
By leaving them as-is in our `dmd/` directory, so that the headers stay next to the D modules (locality is nice for e.g. frontend diffs), and our mods to the headers don't cause any conflicts.


kinke:
Already used for the latest sync some mins ago (with manual dry-run tests beforehand obviously); headers are still there: https://github.com/ldc-developers/ldc/blob/dmd-rewrite-master/dmd/aggregate.h

And the original commit (https://github.com/dlang/dmd/commit/02c1d220cec65e1322d714345a566a1233aa88d4) isn't in our history (empty remaining patch after git-rewrites).

Issue #4982, opened 2025-09-09
Look into using LLVM's new raw byte type for arrays of `void`, `clang` compatibility


thewilsonator:
https://blog.llvm.org/posts/2025-08-29-gsoc-byte-type/

Also we might need to do this for C/C++ interoperability for function signatures for `[[un]signed] char` and `std::byte`, for e.g. cross-language LTO.

gulugulubing:
https://github.com/llvm/llvm-project/pull/186888
https://github.com/llvm/llvm-project/pull/188030
https://github.com/llvm/llvm-project/pull/194074

It seems that the project is still at developping phrase in llvm.
We could do some preparation like rewriting the abi when interoperability, but to achieve the kind of truly impactful use cases, we will likely need to wait for the upstream to flesh out more of the core logic.

PR #5125, opened 2026-05-04
Initial support to bytetype

gulugulubing:
This is relate to #4982. 
ABI of aarch64 is simpler than x86-64(and I only have a macbook), so I just implement byte type on aarch64 first.

### Status
-  AArch64 ABI rewrite (ubyte/char → b8 + zeroext)
-  Call-site: `bitcast i8 → b8` (piggybacking the original addExplicitArguments)
-  Return-site: `bitcast b8 → i8`
-  Callee entry: `bitcast b8 → i8`
-  Callee return: `bitcast i8 → b8`

### Gate
`-fc-interop-llvm-byte` (default off)

### Known Limitations
- LLVM upstream work for byte types is still in progress 
- Clang does not yet emit b8 for unsigned char
- LTO testing uses hand-written .ll for C side

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -616,6 +616,14 @@ cl::opt<bool>
     fSplitStack("fsplit-stack", cl::ZeroOrMore,
                 cl::desc("Use segmented stack (see Clang documentation)"));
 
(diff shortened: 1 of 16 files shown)


thewilsonator:
Thanks for looking into this

PR #5127, opened 2026-05-11
Add pre semantics plugin hook

ljmf00-wekaio:
Like `runSemanticAnalysis`, we should have something like `runPreSemanticAnalysis`. One of the motivations for this is to help referencing AST nodes that represent user code that was deleted by semantics, e.g. constant folding.

CC @JohanEngelen

### DIFF
--- a/driver/plugins.cpp
+++ b/driver/plugins.cpp
@@ -43,11 +43,15 @@ cl::list<std::string> pluginFiles("plugin", cl::CommaSeparated,
 
 struct SemaPlugin {
   llvm::sys::DynamicLibrary library;
+  void (*runPreSemanticAnalysis)(Module *);
   void (*runSemanticAnalysis)(Module *);
 
   SemaPlugin(const llvm::sys::DynamicLibrary &library,
+             void (*runPreSemanticAnalysis)(Module *),
              void (*runSemanticAnalysis)(Module *))
-      : library(library), runSemanticAnalysis(runSemanticAnalysis) {}
+      : library(library),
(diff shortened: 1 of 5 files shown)


driver/plugins.cpp
```diff
@@ -117,10 +126,17 @@ void registerAllPluginsWithPassBuilder(llvm::PassBuilder &PB) {
...
+    if (plugin.runPreSemanticAnalysis)
+      plugin.runPreSemanticAnalysis(m);
+  }
+}
+
 void runAllSemanticAnalysisPlugins(Module *m) {
   for (auto &plugin : sema_plugins) {
-    assert(plugin.runSemanticAnalysis);
-    plugin.runSemanticAnalysis(m);
+    if (plugin.runSemanticAnalysis)
```
JohanEngelen, line 138: nitpick: cleaner to early exit when this is null?

ljmf00-wekaio: The idea here is to optionally require those functions in the plugin, as now we have multiple of them, but I guess we can keep the old behavior and only make `runPreSemanticAnalysis` optional.

What if a user only wants to have a plugin for `runPreSemanticAnalysis`? I can check for both in both places (if they have at least one).

JohanEngelen: That's not what i meant. I agree with the change that both are optional.

I just meant hoisting the if-statement out of the loop :)

JohanEngelen:
It's a good addition I think, although it is a workaround around a fundamental frontend issue (information loss after sema).

Issue #5100, opened 2026-03-18
Compiler segfaults in `DComputeSemanticAnalyser` on indirect function calls


Codingisinmyblud:
When compiling `@compute` code that contains an indirect function call (such as a function pointer or delegate call), LDC crashes with a segmentation fault during the DCompute semantic validation pass.

The bug occurs in [gen/semantic-dcompute.cpp](https://github.com/ldc-developers/ldc/blob/master/gen/semantic-dcompute.cpp) within the DComputeSemanticAnalyser visitor.

The visitor assumes that every `CallExp` is a direct function call and unconditionally dereferences `e->f`. For indirect function calls, `e->f` is a `nullptr.`

In instances such as [isNonComputeCallExpValid](https://github.com/ldc-developers/ldc/blob/master/gen/semantic-dcompute.cpp#L43), the code immediately dereferences f without checking if it exists.

THis can also be seen in [visit(CallExp *e)](https://github.com/ldc-developers/ldc/blob/master/gen/semantic-dcompute.cpp#L214), in `if (e->f->ident == Id::criticalenter)` the visitor unconditionally assumes e->f is a valid pointer.

A suggested fix would be to check `if (e->f != nullptr) ` before accessing its members in `visit(CallExp *e)`  and `isNonComputeCallExpVaild`

badnikhil:
this issue is fixed #5143

PR #5131, opened 2026-05-18
Fix #5116 - Skip the sematics checks of compiler-generated functions and template instances in dcompute

### LINKED ISSUE #5116: Can't compile shared static array in dcompute
Does current ldc's dcompute support shared memory? 
I tried to compile this:
```d
@compute(CompileFor.deviceOnly) module shared_static_array_kernel;

import ldc.dcompute;

private enum N = 16u;

@kernel()
void k(GlobalPointer!float g) {
    Shared!(float[N]) tile;
    tile[0] = g[0];
}
```
but it crashed:
/Users/qiugaofei/dlang/ldc/runtime/druntime/src/core/internal/array/equality.d(42): Error: can only call functions from other `@compute` modules in `@compute` code
    return __equals(lhs[], rhs[]);
                   ^
Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
0  libLLVM.dylib            0x0000000112d0db48 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 56
1  libLLVM.dylib            0x0000000112d0e328 SignalHandler(int, __siginfo*, void*) + 272
2  libsystem_platform.dylib 0x000000019cb64624 _sigtramp + 56
3  ldc2                     0x00000001043566a4 RecursiveWalker::visit(CallExp*) + 40
4  ldc2                     0x00000001043566a4 RecursiveWalker::visit(CallExp*) + 40
5  ldc2                     0x0000000104356088 RecursiveWalker::visit(CompoundStatement*) + 132
6  ldc2                     0x0000000104356088 RecursiveWalker::visit(CompoundStatement*) + 132
7  ldc2                     0x0000000104355338 RecursiveWalker::visit(AttribDeclaration*) + 132
8  ldc2                     0x000000010439206c dcomputeSemanticAnalysis(Module*) + 144
9  ldc2                     0x000000010406f314 extraLDCSpecificSemanticAnalysis(Array<Module*>&) + 80
10 ldc2                     0x00000001041bcd28 mars_tryMain(Param&, Array<char const*>&) + 4308
11 ldc2                     0x00000001043fafe0 cppmain() + 6764
12 ldc2                     0x000000010455e4f0 _D2rt6dmain212_d_run_main2UAAamPUQgZiZ6runAllMFZv + 120
13 ldc2                     0x000000010455e29c _d_run_main2 + 364
14 ldc2                     0x000000010455e114 _d_run_main + 144
15 ldc2                     0x00000001043f951c main + 596
16 dyld                     0x000000019c78ab98 start + 6076

gulugulubing:
It is related to #5116 and #5100.

`Shared!(float[N]) tile;` is actually a struct with static array, DMD's semantic analysis automatically generates support functions for such structs, and those functions drag in template instantiations from `non-@compute` modules, which  triggers both the error message and the crash of compiler.

### DIFF
--- a/gen/semantic-dcompute.cpp
+++ b/gen/semantic-dcompute.cpp
@@ -236,7 +236,13 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
     }
       
     Module *m = e->f->getModule();
-    if ((m == nullptr || (hasComputeAttr(m) == DComputeCompileFor::hostOnly)) &&
+    // Template-instantiated functions are cross-module by nature: the template
+    // declaration and the instantiated function live in different modules.
+    // getModule() returns the *declaration* module, which says nothing about
+    // whether the generated code can run on GPU. Skip the module check for them.
+    const bool isTemplateFunc = e->f->isInstantiated() != nullptr;
+    if (!isTemplateFunc &&
+        (m == nullptr || (hasComputeAttr(m) == DComputeCompileFor::hostOnly)) &&
         !isNonComputeCallExpVaild(e)) {
       error(e->loc, "can only call functions from other `@compute` modules in "
                     "`@compute` code");
@@ -247,16 +253,6 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
   void visit(FuncDeclaration *fd) override {
     if (getKernelAttr(fd) && fd->vthis) {
       error(fd->loc, "`@kernel` functions must not require `this`");
-      stop = true;
-      return;
-    }
-
-    // Skip compiler-generated struct support functions (e.g. __xopEquals,
-    // __xopCmp, postblit, destructor). Their bodies may reference non-@compute
-    // templates (such as __equals for static array comparison) that are outside
-    // of user control. They are only codegenerated if actually referenced by
-    // user code, at which point the codegen layer will report any issues.
-    if (fd->isGenerated()) {
       stop = true;
       return;
     }
@@ -275,20 +271,6 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
     // as they contain unsupported global variables.
     if (ti->tempdecl == Type::rtinfo || ti->tempdecl == Type::rtinfoImpl) {
       stop = true;
-      return;
-    }
-
-    // Template instantiations for templates declared in non-@compute modules
-    // (e.g. __equals and isEqual from core.internal.array.equality) are
-    // created as a side effect of compiler-generated support functions. They
-    // contain calls back into their declaring (non-@compute) module, which
-    // would produce spurious errors. Skip them.
-    if (ti->tempdecl) {
-      Module *m = ti->tempdecl->getModule();
-      if (m && hasComputeAttr(m) == DComputeCompileFor::hostOnly) {
-        stop = true;
-        return;
-      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


gen/semantic-dcompute.cpp
```diff
@@ -246,6 +251,16 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
       return;
     }
 
+    // Skip compiler-generated struct support functions (e.g. __xopEquals,
+    // __xopCmp, postblit, destructor). Their bodies may reference non-@compute
+    // templates (such as __equals for static array comparison) that are outside
```
thewilsonator: `__equals`  is a template which is specifically allowed, any non-template code it generates will still cause problems though.

gulugulubing: Are you suggesting that __equals (and its generated code, like isEqual) should be included in `isNonComputeCallExpVaild` similar to dcReflect, rather than using the module-based check in visit(TemplateInstance *)?

gulugulubing: Or should we make __equals really work on GPU?

thewilsonator: well it should either fall back on default element-wise comparison, or use a user-supplied `opEquals` both of which should be doable.

gulugulubing: Sorry if I've been going in circles here--With the current patch (skipping the semantic check for these
generated template instances), __equals should actually be able to run on GPU.

Are you suggesting that I should implement the new lowering for == on GPU targets, so that == doesn't rely on __equals from the CPU runtime at all)?

thewilsonator: > Are you suggesting that __equals (and its generated code, like isEqual) should be included in isNonComputeCallExpVaild similar to dcReflect, rather than using the module-based check in visit(TemplateInstance *)?

No, templates should already work across host/device (assuming that they are leaf level function or only call other templates. If `__equals` does something like call `memset` then that obviously will not work.

Why specifically does it not currently work? is it because of `memset`.

gulugulubing: > Why specifically does it not currently work? is it because of `memset`.

No, it's not about `memset`
The problem isn't that __equals does anything GPU-incompatible — at first, I didn't quite figure it out either.  The problem is how it gets instantiated. These template instances are added as members of the `@compute` module. Then `dcomputeSemanticAnalysis` walks them and either:
- Crashes on the function pointer cast `(cast(PureType)&isEqual)() inside __equals` because e->f is null for indirect calls.
- Reports spurious errors on the nested __equals calls , because __eqauls is not tagged with `@compute` then `dcomputeSemanticAnalysis` triggers `Error: can only call functions from other @compute modules in @compute code`.

gen/semantic-dcompute.cpp
```diff
@@ -246,6 +251,16 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
...
+    // of user control. They are only codegenerated if actually referenced by
+    // user code, at which point the codegen layer will report any issues.
+    if (fd->isGenerated()) {
+      stop = true;
+      return;
+    }
+
```
gulugulubing: I just find this guard is redundant.

gen/semantic-dcompute.cpp
gulugulubing, line 246: Is here a better place to fix the issue by skipping  template-instantiated functions? Because When the compiler instantiates a template, it copies the function body into the instantiating scope, but the internal field that `getModule()` reads is **never updated** — it still references the module where the template was originally declared.

gulugulubing:
Closed. #5142 fixed the similar problem.

PR #5136, opened 2026-06-09
CI: Work around curl.se having dropped official 32-bit Windows builds

kinke:
They've apparently removed all published 32-bit packages too, not just stopped putting out new 32-bit builds.

Resort to the latest 32-bit DLL bundled with LDC v1.42.0, but use the latest `curl-ca-bundle.crt` from the official x86_64 build.

### DIFF
--- a/.github/actions/1-setup/action.yml
+++ b/.github/actions/1-setup/action.yml
@@ -173,10 +173,6 @@ runs:
         cd ..
 
         url='https://curl.se/windows/latest.cgi?p=win64-mingw.zip'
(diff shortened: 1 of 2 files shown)


JohanEngelen:
You could use this URL too: https://web.archive.org/web/20260106010715/https://curl.se/windows/dl-8.15.0_5/curl-8.15.0_5-win32-mingw.zip

kinke:
Right, but I guess GitHub handles the traffic better than the archive.

PR #5140, opened 2026-06-11
Implement native device code embedding for DCompute (PTX and SPIR-V)

badnikhil:
## Overview                                                                                                                                                                            
  This PR introduces Native device code Embedding for DCompute, enabling device kernels to be natively baked directly into the host executable's `.rodata` section.                                                                                          
                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                         
  ## Key Changes                                                                                                                                                                         
                                                                                                                                                                                         
 ### 1. Native device code Injection (`target.cpp`, `main.cpp`, `codegenerator.cpp`)                                                                                                            
  - Implemented a callback hook (`dcomputeHook_`) in `CodeGenerator` that triggers right before `writeAndFreeLLModule`.                                                                  
  - Used `llvm::GlobalAlias` to intercept and resolve `extern(C)` payload references in the host module, seamlessly pointing them to the actual PTX binary string embedded as an LLVM    
`GlobalVariable`.   
- OpenCL SPIR-V embedding now dynamically assigns the correct `spv` identifier instead of hardcoding `ptx` to prevent naming collisions.Additionally, explicitly enforces a 4-byte memory alignment (`align 4`) on the embedded SPIR-V payload to prevent unaligned memory access crashes on OpenCL drivers.                                                                                                                                                                

 ### 2. ABI Resolution & Codegen Execution Reordering (`main.cpp`)
  - **Bug Fixed:** Previously, if host modules were emitted before device modules, `DtoResolveFunction` would erroneously resolve the kernel function using the host ABI instead of the  
target's `PTX_Kernel` calling convention.
  - **Solution:** Reordered the generation pipeline to ensure `dccg.emit(mod)` executes strictly *before* `cg.emit(m)`, guaranteeing kernels are resolved with the correct device ABI.   
   - **File Type Filtering:** Whitelisted `FileType::c` so that ImportC source modules are safely permitted through the codegen phase alongside standard `.d` files.                      
   
  ### 3. Global State Restoration (`dcomputecodegenerator.cpp`)
  - **Bug Fixed:** Device compilation overrode critical global state (`gIR`, `gTargetMachine`, and `gABI`) with device targets, which subsequently poisoned the host compilation and     
triggered LLVM Verifier failures.
  - **Solution:** `DComputeCodeGenManager::emit` now securely caches `oldGIR`, `oldGTargetMachine`, and `oldGABI` before generating device code, and seamlessly restores the host state  before handing execution back.
### 4. LLVM Context & Type Leakage Fixes
 - **Double Free Fix:** Resolved a double-free bug that occurred during LLVM context teardown by ensuring DCompute targets are always flushed correctly.
  - **Host Codegen Isolation:** Fixed an issue where DCompute types were leaking into host codegen, ensuring device-specific types do not accidentally pollute host IR generation.
  ### 5. CLI Toggles
  
 - **-fembed-dcompute** : Added a new command-line option (defaults to  true ). Users can pass  -fembed-dcompute=false  to bypass the new embedding behavior and fall back exclusively to  emitting standalone  .ptx  /  .spv  files to the disk. (Safeguarded for LLVM builds lacking DCompute support like Android).
  

  ## Testing
  -  Added  tests/codegen/dcompute_dual_targets.d  to verify safe injection and prevent naming collisions when simultaneously generating  -mdcompute-targets=cuda-350,ocl-220 .             
  - Added  tests/codegen/dcompute_no_embed.d  to strictly verify the  -fembed-dcompute=false  fallback logic using  FileCheck .

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -1283,7 +1283,7 @@ void codegenModules(Modules &modules) {
     for (d_size_t i = modules.length; i-- > 0;) {
       Module *const m = modules[i];
 
-      if (m->filetype == FileType::dhdr)
+      if (m->filetype != FileType::d)
         continue;
 
       if (global.params.v.verbose)
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -1283,7 +1283,7 @@ void codegenModules(Modules &modules) {
     for (d_size_t i = modules.length; i-- > 0;) {
       Module *const m = modules[i];
 
-      if (m->filetype != FileType::d)
+      if (m->filetype != FileType::d && m->filetype != FileType::c)
         continue;
 
       if (global.params.v.verbose)
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -1218,6 +1218,7 @@ void codegenModules(Modules &modules) {
   if (global.params.obj && !modules.empty()) {
     dmd::TimeTraceScope timeScope(TimeTraceEventType::codegenGlobal);
 
+    DComputeCodeGenManager dccg(getGlobalContext());
 #if LDC_MLIR_ENABLED
     mlir::MLIRContext mlircontext;
     ldc::CodeGenerator cg(getGlobalContext(), mlircontext,
@@ -1225,7 +1226,6 @@ void codegenModules(Modules &modules) {
 #else
     ldc::CodeGenerator cg(getGlobalContext(), global.params.oneobj);
 #endif
-    DComputeCodeGenManager dccg(getGlobalContext());
     std::vector<Module *> computeModules;
     // When inlining is enabled, we are calling semantic3 on function
     // declarations, which may _add_ members to the first module in the modules
@@ -1241,23 +1241,7 @@ void codegenModules(Modules &modules) {
       if (m->filetype == FileType::dhdr)
         continue;
 
-      if (global.params.v.verbose)
-        message("code      %s", m->toChars());
-
       const auto atCompute = hasComputeAttr(m);
-      if (atCompute == DComputeCompileFor::hostOnly ||
-          atCompute == DComputeCompileFor::hostAndDevice) {
-        dmd::TimeTraceScope timeScope(
-            TimeTraceEventType::codegenModule,
-            (llvm::Twine("Codegen: module ") + m->toChars()).str().c_str(),
-            m->toChars(), m->loc);
-#if LDC_MLIR_ENABLED
-        if (global.params.output_mlir == OUTPUTFLAGset)
-          cg.emitMLIR(m);
-        else
-#endif
-          cg.emit(m);
-      }
       if (atCompute != DComputeCompileFor::hostOnly) {
         computeModules.push_back(m);
         if (atCompute == DComputeCompileFor::deviceOnly) {
@@ -1271,10 +1255,9 @@ void codegenModules(Modules &modules) {
           }
         }
       }
-      if (global.errors)
-        fatal();
     }
 
+    bool dcomputeWritten = false;
     if (!computeModules.empty()) {
       dmd::TimeTraceScope timeScope("Codegen DCompute device modules");
       for (auto &mod : computeModules) {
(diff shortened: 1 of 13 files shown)


gen/dcompute/target.cpp
```diff
@@ -64,6 +71,47 @@ void DComputeTarget::writeModule() {
 
   ::writeModule(&_ir->module, path);
 
+  if (hostModule) {
+    auto bufferOrErr = llvm::MemoryBuffer::getFile(path);
+    if (bufferOrErr) {
```
thewilsonator: these two should be early returns.

gen/semantic-dcompute.cpp
```diff
@@ -219,17 +219,19 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
     //    _d_criticalexit( & __critsec105);   |
     // So we intercept it with the CallExp ----
 
-    if (e->f->ident == Id::criticalenter) {
+    if (e->f && e->f->ident == Id::criticalenter) {
```
thewilsonator: please split the changes in this file out into a separate PR, with a test they can be merged straight away.

thewilsonator (commented):
needs tests, and `codegen/dcompute_host_and_device.d` fails

badnikhil:
> needs tests, and `codegen/dcompute_host_and_device.d` fails
```
 for (auto exp : *attrs) {
    unsigned prevErrors = global.startGagging();
    auto e = ctfeInterpret(exp);
    if (global.endGagging(prevErrors))
      continue;
```
this introduces address spaces so 
` load ptr, ptr`
is now 
`load ptr addrspace(3)`
Updating the test and also  checks for the embedded ptx

badnikhil:
@thewilsonator please check the latest commits.. `lsan_memleak.d`  test detected memory leaks.. also fixed it

tests/codegen/dcompute_host_and_device.d
```diff
@@ -22,40 +22,43 @@ void main(string[] args)
     foo(global_x);
 }
 
+// LL: __dcompute_ptx_internal_
+// LL: __dcompute_ptx_{{.*}} ={{.*}}alias
```
thewilsonator, line 26: what is the `alias` there for?

badnikhil: The alias is   to prevent duplicating the PTX string in the final binary. 

Since the NVPTX backend bundles all  @compute  modules into a single, unified PTX string, we  emit that string exactly once as a private constant ( __dcompute_ptx_internal_... ). Then, we loop through the modules and create an alias for each one that points back to that shared string, ensuring the host can resolve the symbols without bloating the executable.

driver/main.cpp
```diff
@@ -1286,8 +1270,47 @@ void codegenModules(Modules &modules) {
...
+        }
+      });
+      IrTypeStruct::resetDComputeTypes();
+    }
+
+    bool hasHostModules = false;
+    for (d_size_t i = modules.length; i-- > 0;) {
+      Module *const m = modules[i];
+
+      if (m->filetype == FileType::dhdr)
```
thewilsonator: this should be `m->filetype != FileType::d`

gen/dcompute/target.cpp
```diff
@@ -63,9 +70,53 @@ void DComputeTarget::writeModule() {
...
+
+  if (!hostModule)
+    return;
+
+  auto bufferOrErr = llvm::MemoryBuffer::getFile(path);
+  if (!bufferOrErr)
+    return;
+
+  llvm::StringRef ptxString = bufferOrErr.get()->getBuffer();
+  llvm::Constant *ptxConst = llvm::ConstantDataArray::getString(
```
thewilsonator: rename these to `deviceString` etc

gen/dcompute/target.cpp
```diff
@@ -63,9 +70,53 @@ void DComputeTarget::writeModule() {
...
+      ctx, ptxString, true);
+
+  std::string internalName = "__dcompute_" + std::string(binSuffix) + "_internal_" + std::string(short_name) + std::to_string(tversion) + "_" + opts::dcomputeFilePrefix;
```
thewilsonator: the logic here would be easier to understand...

gen/dcompute/target.cpp
```diff
@@ -63,9 +70,53 @@ void DComputeTarget::writeModule() {
...
+      true, // isConstant
+      llvm::GlobalValue::PrivateLinkage,
+      ptxConst,
+      internalName);
+  gv->setAlignment(llvm::Align(4));
+
+  for (auto *m : modules) {
+    std::string modName = m->toPrettyChars();
+    std::replace(modName.begin(), modName.end(), '.', '_');
+    std::string symName = "__dcompute_" + std::string(binSuffix) + "_" + std::string(short_name) + std::to_string(tversion) + "_" + modName;
```
thewilsonator: ...if this were `std::string symName = internalName + ...;

gen/dcompute/target.h
```diff
@@ -38,15 +39,20 @@ class DComputeTarget {
   std::array<int, 5> mapping;
 
   IRState *_ir;
+  std::vector<::Module *> modules;
```
thewilsonator, line 42: do we need a vector of frontend modules _per target_? They are all the same. Make this static?

badnikhil: This is because  DComputeCodeGenManager::emit(Module *m)  iterates over all targets and calls  target->emit(m)  for each one. If  modules  is  static  and shared across all targets,    
  compiling with two targets (e.g., CUDA and OpenCL) would cause  target->emit(m)  to push the exact same module into the static vector twice, resulting in duplicates.

thewilsonator (approved):

Issue #5138, opened 2026-06-10
CI: Replace dead Cirrus CI


kinke:
They were bought by OpenAI and shut down on June 1st: https://cirruslabs.org

We mainly need a successor for the FreeBSD CI artifacts. https://github.com/cross-platform-actions/action is probably the first candidate, but no idea if we can use our reusable custom actions in `.github/actions/` nowadays.

PR #5141, opened 2026-06-11
FreeBSD: add powerpc64 support

pkubaj:
Use ELFv2 on powerpc64 - FreeBSD switched to ELFv2 in 13.0-RELEASE.

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -143,7 +143,7 @@ const char *getABI(const llvm::Triple &triple, const llvm::SmallVectorImpl<llvm:
   case llvm::Triple::mips64el:
     return "n32";
   case llvm::Triple::ppc64:
-    return "elfv1";
+    return triple.isOSFreeBSD() ? "elfv2" : "elfv1";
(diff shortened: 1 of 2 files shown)


kinke:
Thx for the work! Cirrus is unfortunately dead: https://github.com/ldc-developers/ldc/issues/5138

pkubaj:
I updated the PR to only contain the first commit. The CI-related one will have a new PR once you decide on the successor.

kinke:
> FreeBSD switched to ELFv2 in 13.0-RELEASE

Do you think we should check the OS major then (`triple.getOSMajorVersion()`), or do we not expect anyone to target FreeBSD < 13 these days?

pkubaj:
12.x went EOL 31.12.2023. Apart from using ELFv1 on powerpc64, it was also limited to GCC 4.2 as a base compiler. It always had issues with LLVM from ports, because LLVM has assumed for many years that FreeBSD uses libc++, because that's what amd64 had been doing. So no, there's no point in supporting ELFv1 on FreeBSD.

PR #5142, opened 2026-06-12
DCompute: Skip host-side template instantiations during semantic analysis

badnikhil:
When a `@compute` device module instantiates a template that resides in a normal, host-only module (like `std.stdio` or internal runtime `TypeInfo` templates), the DCompute semantic  visitor previously attempted to fully analyze those instantiations. Because host-side templates frequently contain constructs that are explicitly banned in `@compute` code (such as global variables, classes, or unsupported function calls), this resulted in spurious compilation errors that prevented otherwise valid code from compiling.
  
 **Fix**
    We now check the module where the template is declared (`ti->tempdecl->getModule()`). If the declaring module does not have a `@compute` attribute (meaning it resolves to `CompileFor.
  hostOnly`), we explicitly skip DCompute semantic analysis for that specific instantiation. 
  **Tests**
    - Added a new compilable test case (`tests/compilable/dcompute_host_template_test.d`) which imports a mock host-only module and instantiates a template containing an unsupported      
  global variable.
    - Verified that it correctly bypasses semantic analysis and compiles successfully.

### DIFF
--- a/tests/compilable/dcompute_host_template_test.d
+++ b/tests/compilable/dcompute_host_template_test.d
@@ -1,4 +1,6 @@
 // RUN: %ldc -c -mdcompute-targets=cuda-350 -I%S %s
+// RUN: not %ldc -o- -mdcompute-targets=cuda-350 -verrors=0 -I%S -d-version=Fail %s 2>&1 | FileCheck %s
+
 @compute(CompileFor.deviceOnly) module dcompute_host_template_test;
 import ldc.dcompute;
 import inputs.host_template_module;
@@ -8,4 +10,9 @@ void kernel() {
     // This should compile successfully because DCompute semantic analysis
     // should skip over host-only template instantiations.
     alias X = HostTemplate!int;
+
+    version(Fail) {
+        // CHECK: dcompute_host_template_test.d([[@LINE+1]]): Error: can only call functions from other `@compute` modules in `@compute` code
+        X.doHostThings();
+    }
 }
--- a/tests/compilable/dcompute_host_template_test.d
+++ b/tests/compilable/dcompute_host_template_test.d
@@ -1,0 +1,11 @@
+// RUN: %ldc -c -mdcompute-targets=cuda-350 -I%S %s
+@compute(CompileFor.deviceOnly) module dcompute_host_template_test;
+import ldc.dcompute;
+import inputs.host_template_module;
+
+void kernel() {
+    // Instantiate a template that resides in a hostOnly module.
+    // This should compile successfully because DCompute semantic analysis
+    // should skip over host-only template instantiations.
+    alias X = HostTemplate!int;
+}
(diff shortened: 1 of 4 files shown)


badnikhil:
@thewilsonator i can't see any option to re run the failing check. Please re run it it'll probably run successfully. It failed because of some GitHub side issues.

tests/compilable/dcompute_host_template_test.d
```diff
@@ -0,0 +1,11 @@
+// RUN: %ldc -c -mdcompute-targets=cuda-350 -I%S %s
+@compute(CompileFor.deviceOnly) module dcompute_host_template_test;
+import ldc.dcompute;
+import inputs.host_template_module;
+
+void kernel() {
+    // Instantiate a template that resides in a hostOnly module.
+    // This should compile successfully because DCompute semantic analysis
+    // should skip over host-only template instantiations.
+    alias X = HostTemplate!int;
```
thewilsonator, line 12: this should include a semantic test where `doHostThings` is called and it should fail with not-a-linker error

badnikhil: added a commit!

tests/compilable/dcompute_host_template_test.d
```diff
@@ -0,0 +1,18 @@
+// RUN: %ldc -c -mdcompute-targets=cuda-350 -I%S %s
+// RUN: not %ldc -o- -mdcompute-targets=cuda-350 -verrors=0 -I%S -d-version=Fail %s 2>&1 | FileCheck %s
```
thewilsonator: add a `--check-prefix=FAIL` to the `FileCheck` command here...

tests/compilable/dcompute_host_template_test.d
```diff
@@ -0,0 +1,18 @@
...
+import inputs.host_template_module;
+
+void kernel() {
+    // Instantiate a template that resides in a hostOnly module.
+    // This should compile successfully because DCompute semantic analysis
+    // should skip over host-only template instantiations.
+    alias X = HostTemplate!int;
+
+    version(Fail) {
+        // CHECK: dcompute_host_template_test.d([[@LINE+1]]): Error: can only call functions from other `@compute` modules in `@compute` code
```
thewilsonator: ...and change this `//CHECK:` to `//FAIL:`

thewilsonator (approved):

gulugulubing:
Does it mean operations like `float[N] == float[N]` are not allowed in dcompute? Otherwise DMD's frontend lowers float[N] == float[N] to a call to __equals from core.internal.array.equality, which  is a template declared in a host  module.

badnikhil:
> Does it mean operations like `float[N] == float[N]` are not allowed in dcompute? Otherwise DMD's frontend lowers float[N] == float[N] to a call to __equals from core.internal.array.equality, which is a template declared in a host module.

That is correct. Operations that implicitly lower to host-only runtime templates or hooks (like  __equals  from  core.internal.array.equality ) are inherently not supported for execution on the device in DCompute..      
  
This PR doesn't introduce this limitation; rather, it fixes how the compiler handles the template instantiation step itself. By  handling host-only template instantiations, we prevent the compiler from failing or crashing during semantic analysis on unused templates. If a kernel actually attempts to call the lowered  __equals  (e.g., by doing  float[N] ==  float[N] ), it will appropriately fail later in the pipeline because the function definition hasn't been emitted for the device.

PR #5143, opened 2026-06-12
DCompute: Fix semantic compiler crash on indirect calls 

badnikhil:
Indirect calls, such as those made through function pointers and delegates, are explicitly unsupported in DCompute code. When the `DComputeSemanticAnalyser` attempted to visit a  `CallExp` resulting from an indirect call, the compiler would crash with a segmentation fault due to a null pointer dereference on `e->f->ident` (because `e->f` is null for indirect calls). 
  
**Fix**
    Added an explicit check for `!e->f` at the very beginning of the `CallExp` visitor. Instead of silently ignoring it or crashing, we now correctly throw a compilation error (`"function  pointers and delegates are not allowed in @compute code"`) and halt further analysis. By moving this check to the top, we also cleaned up redundant null checks lower in the function.   
  
**Tests**
    - Folded a test case for indirect calls into the existing semantic failure suite (`tests/semantic/dcompute.d`).
    - Verified via `FileCheck` that the compiler successfully rejects the code and emits the correct error message instead of crashing.

### DIFF
--- a/gen/semantic-dcompute.cpp
+++ b/gen/semantic-dcompute.cpp
@@ -266,7 +266,6 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
     // as they contain unsupported global variables.
     if (ti->tempdecl == Type::rtinfo || ti->tempdecl == Type::rtinfoImpl) {
       stop = true;
-      return;
     }
   }
 
--- a/gen/semantic-dcompute.cpp
+++ b/gen/semantic-dcompute.cpp
@@ -212,6 +212,12 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
     }
   }
   void visit(CallExp *e) override {
+    if (!e->f) {
+      error(e->loc, "function pointers and delegates are not allowed in `@compute` code");
+      stop = true;
+      return;
+    }
+
     // SynchronizedStatement is lowered to
     //    Critsec __critsec105; // 105 == line number
     //    _d_criticalenter(& __critsec105); <--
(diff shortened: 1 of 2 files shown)


thewilsonator (approved):

gen/semantic-dcompute.cpp
```diff
@@ -260,6 +266,7 @@ struct DComputeSemanticAnalyser : public StoppableVisitor {
     // as they contain unsupported global variables.
     if (ti->tempdecl == Type::rtinfo || ti->tempdecl == Type::rtinfoImpl) {
       stop = true;
+      return;
```
JohanEngelen: no error message here?

thewilsonator: Correct, this is for RT generated by kernel stuff, not used/referenced by it.

JohanEngelen: ah now i see more context around the diff. Why make any change to this function at all? The return does not do anything here.

badnikhil: my most recent 2 PRs in ldc are opened to seperate some changes from my most recent 3rd PR #5140 . Here,another if statement was added b ( so the `return` ) It got sepertaed to (#5142  ). `return` can be removed too its not a problem actually but it will be added again by that PR [here](https://github.com/ldc-developers/ldc/pull/5142/changes#diff-de13c158b367eb65832807c3ccc82b984bea4f0e84421118a843f3a53ffa74ac)..

PR #5144, opened 2026-06-15
Fix #5135: Prevent CondExp from dropping address spaces

### LINKED ISSUE #5135: dcompute: ternary ?: on GlobalPointer load uses generic ptr, reads fail
For `@compute, (cond) ? input[i] : 0 with GlobalPointer!(const(T))` can produce a load from generic ptr instead of ptr addrspace(1). Runtime result is wrong.  

Root cause: `visit(CondExp)` in `gen/toir.cpp` stores **pointers to branch lvalues** in a `condtmp` alloca(which is generic pointer) and reloads through `DSpecialRefValue::getRVal()`, which uses generic pointers and drops the device address space. 

Minimal repoducer(I built it on my own developping Metal backend,  so I am not sure it will happen on any other backend of dcompute):
```d
@compute(CompileFor.deviceOnly) module repro;
import ldc.dcompute;
import dcompute.std.index;

enum size_t N = 4;

@kernel()
void loadViaTernary(GlobalPointer!(const(int)) input, GlobalPointer!int output) {
    size_t idx = GlobalIndex.x;
    int val = (idx < N) ? input[idx] : 0;  // broken
    output[idx] = val;
}
```

So I am thinking could you fix this by add a guard: if condExp is not used as lvalue, just store the rvalue in `condtmp`, but it could only fix `val = (c) ? input[i] : 0`, not `val = (c) ? input[i] : input[j]`.

anushkagupta200615-jpg:
This fixes an issue in the DCompute backend where ternary condition expressions dropped device address spaces. R-values now use local allocas for the value, and L-values natively preserve pointer types through LLVM PHI nodes.

Fixes #5135.

This PR fixes a bug in the DCompute backend where ternary condition expressions `(c) ? e1 : e2` stripped device address spaces (e.g., `addrspace(1)`) from `GlobalPointer` memory references. 

**Root Cause:**
Previously, `visit(CondExp)` indiscriminately wrapped the branches inside a generic address space temporary (`condtmp` pointer), which forced `DSpecialRefValue::getRVal()` to reload the branches using `getOpaquePtrType()`. This stripped out the inner device address space.

**Solution:**
The code has been updated to distinctly handle ternary expressions depending on their evaluation context:
1. **L-value context (`e->isLvalue()`):** Entirely bypasses intermediate allocation wrappers and `DSpecialRefValue`. Instead, it fetches the actual device pointers using `makeLValue` and naturally preserves their exact LLVM type (including `addrspace`) by linking them into an LLVM `PHINode` at the branch merge point. 
2. **R-value context (`!e->isLvalue()`):** Creates a standard `alloca` for the target value type (rather than an `alloca` of a pointer), evaluates the condition branches as standard R-values, and stores them appropriately. This prevents any inner structs (like `GlobalPointer`) from having their address spaces stripped from memory.

This fully resolves `val = (c) ? input[i] : 0` as well as the related `val = (c) ? input[i] : input[j]` bug.

### DIFF
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -2173,7 +2173,7 @@ public:
       llvm::Type *ptrType = nullptr;
       if (u_val) ptrType = u_val->getType();
       else if (v_val) ptrType = v_val->getType();
-      else ptrType = DtoPtrToType(dtype);
+      else ptrType = getOpaquePtrType();
 
       if (u_bb && v_bb) {
         llvm::PHINode *phi = p->ir->CreatePHI(ptrType, 2, "condtmp");
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -2179,6 +2179,8 @@ public:
         ptrType = v_val->getType();
       }
 
+      LLValue *val = nullptr;
+
       if (u_bb && v_bb) {
         llvm::PHINode *phi = p->ir->CreatePHI(ptrType, 2, "condtmp");
         
@@ -2194,21 +2196,29 @@ public:
         }
         phi->addIncoming(v_inc, v_bb);
         
-        result = new DLValue(e->type, phi);
+        val = phi;
       } else if (u_bb) {
         LLValue *u_inc = u_val ? u_val : llvm::UndefValue::get(ptrType);
         if (u_inc->getType() != ptrType) {
           u_inc = p->ir->CreatePointerBitCastOrAddrSpaceCast(u_inc, ptrType);
         }
-        result = new DLValue(e->type, u_inc);
+        val = u_inc;
       } else if (v_bb) {
         LLValue *v_inc = v_val ? v_val : llvm::UndefValue::get(ptrType);
         if (v_inc->getType() != ptrType) {
           v_inc = p->ir->CreatePointerBitCastOrAddrSpaceCast(v_inc, ptrType);
         }
-        result = new DLValue(e->type, v_inc);
+        val = v_inc;
       } else {
-        result = new DLValue(e->type, llvm::UndefValue::get(ptrType));
+        val = llvm::UndefValue::get(ptrType);
+      }
+
+      if (auto du = u ? u->isDDcomputeLVal() : nullptr) {
+        result = new DDcomputeLValue(e->type, du->lltype, val);
+      } else if (auto dv = v ? v->isDDcomputeLVal() : nullptr) {
+        result = new DDcomputeLValue(e->type, dv->lltype, val);
+      } else {
+        result = new DLValue(e->type, val);
       }
     } else if (retPtr) {
       result = new DLValue(e->type, retPtr);
--- a/gen/toir.cpp
+++ b/gen/toir.cpp
@@ -2170,19 +2170,26 @@ public:
 
     p->ir->SetInsertPoint(condend);
     if (isLvalue) {
-      if (u_bb && v_bb && u_val && v_val) {
-        llvm::Type *phiType = u_val->getType();
-        llvm::PHINode *phi = p->ir->CreatePHI(phiType, 2, "condtmp");
-        phi->addIncoming(u_val, u_bb);
-        if (v_val->getType() != phiType) {
-          v_val = DtoBitCast(v_val, phiType);
+      llvm::Type *ptrType = nullptr;
+      if (u_val) ptrType = u_val->getType();
+      else if (v_val) ptrType = v_val->getType();
+      else ptrType = DtoPtrToType(dtype);
+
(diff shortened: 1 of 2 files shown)


JohanEngelen:
Needs a test case.

> This fully resolves `val = (c) ? input[i] : 0` as well as the related `val = (c) ? input[i] : input[j]` bug.

You could use this :)

gen/toir.cpp
```diff
@@ -2131,26 +2133,67 @@ class ToElemVisitor : public Visitor {
...
 
+    LLValue *u_val = nullptr;
+    llvm::BasicBlock *u_bb = nullptr;
     p->ir->SetInsertPoint(condtrue);
     PGO.emitCounterIncrement(e);
     DValue *u = toElem(e->e1);
-    if (retPtr && u->type->toBasetype()->ty != TY::Tnoreturn) {
-      LLValue *lval = makeLValue(e->loc, u);
-      DtoStore(lval, retPtr);
+    if (u->type->toBasetype()->ty != TY::Tnoreturn) {
```
gulugulubing: `if (u && u->type->toBasetype()->ty != TY::Tnoreturn) {` may fix the ci.

gen/toir.cpp
```diff
@@ -2131,26 +2133,67 @@ class ToElemVisitor : public Visitor {
...
     createBranch(condend, p->scopebb());
 
+    LLValue *v_val = nullptr;
+    llvm::BasicBlock *v_bb = nullptr;
     p->ir->SetInsertPoint(condfalse);
     DValue *v = toElem(e->e2);
-    if (retPtr && v->type->toBasetype()->ty != TY::Tnoreturn) {
-      LLValue *lval = makeLValue(e->loc, v);
-      DtoStore(lval, retPtr);
+    if (v->type->toBasetype()->ty != TY::Tnoreturn) {
```
gulugulubing: Similarly, `if (v && v->type->toBasetype()->ty != TY::Tnoreturn)`

gulugulubing (commented):
Thanks for working on this fix!

badnikhil:
@anushkagupta200615-jpg Add if ` (u && ...)` / `if (v && ...) `null checks (per gulugulubing's review) and harden the PHI path.

gen/toir.cpp
```diff
@@ -2131,26 +2133,67 @@ class ToElemVisitor : public Visitor {
...
+      llvm::Type *ptrType = nullptr;
+      if (u_val) ptrType = u_val->getType();
+      else if (v_val) ptrType = v_val->getType();
+      else ptrType = getOpaquePtrType();
+
+      if (u_bb && v_bb) {
+        llvm::PHINode *phi = p->ir->CreatePHI(ptrType, 2, "condtmp");
+        phi->addIncoming(u_val ? u_val : llvm::UndefValue::get(ptrType), u_bb);
+        LLValue *v_inc = v_val ? v_val : llvm::UndefValue::get(ptrType);
+        if (v_inc->getType() != ptrType) {
```
badnikhil: With opaque pointers these types only differ by address space, and bitcast across address spaces is invalid (needs addrspacecast). Since preserving the address space is the whole point of this PR, unifying via bitcast looks like it would either assert or drop the space again . what's the intended behavior when the two branches have different address spaces?

gulugulubing:
The CI failures are caused by DtoRVal() crashing on memory-only types in the rvalue path. I've submitted a fix PR to your branch: https://github.com/anushkagupta200615-jpg/ldc/pull/1. Merging it should solve this problem.

gulugulubing:
Hi @anushkagupta200615-jpg, I've been looking into the CI failures(I reported #5135 originally). The rvalue path was straightforward to fix, but the remaining lvalue failure (ternary_lval_assign test) is more subtle than expected.

badnikhil:
@gulugulubing the failing test is unsupported for LLVM18 ,  couldnt reproduce it:) took some effort to figure out the exact problem .. Also i am not very much sure about the side effects(may or may not be)  @JohanEngelen @thewilsonator please have a look.

gen/toir.cpp
```diff
@@ -2131,26 +2133,96 @@ class ToElemVisitor : public Visitor {
...
+        ptrType = u_val->getType();
+      } else if (u_val && !v_val) {
+        ptrType = u_val->getType();
+      } else if (v_val && !u_val) {
+        ptrType = v_val->getType();
+      }
+
+      LLValue *val = nullptr;
+
+      if (u_bb && v_bb) {
```
thewilsonator, line 2194: these three branches seem largely duplicates of each other. It would be good to factor out the commonalities.

thewilsonator (approved):

PR #5146, opened 2026-06-15
[regression] Restore `cold`/`noreturn` attributes on runtime error functions

kubo39:
The D runtime error functions (_d_assert, _d_arraybounds* etc.) lost their `cold` and `noreturn` attributes in https://github.com/ldc-developers/ldc/commit/705a60e8131b0e6c5c47fb0afa219c40deb5adc8, which rewrote the attribute table in gen/runtime.cpp to lambdas: the `cold` builder was never applied and the `noreturn` builder was missing. Restore both attributes and add a regression test.


thewilsonator (approved):
Thanks.

Issue #4991, opened 2025-10-05
Array ops not as optimized as it could be


rikkimax:
Following: https://salykova.github.io/gemm-cpu
I decided to have a play with array ops specifically for the kernel:

```d
void kernel4(ref float[16] o, ref const float[16] a, ref const float[16] b, ref const float[16] c) {
    o[] = b[] * a[];
    o[] += c[];
}
```

For all of these examples you need ``-mattr=+avx512vl``

Right now the codegen for this is abysmal it is nearly 90 lines full of ``vmovss`` and ``vmulss``.
Compare it to explicit simd types:

```d
void kernel1(ref float16 o, ref const float16 a, ref const float16 b, ref const float16 c) {
    o = (b * a) + c;
}
```

```
_D7example7kernel1FKNhG16fKxNhQiKxQgKxQkZv:
.Lfunc_begin6:
        .cfi_startproc
        .loc    1 34 5 prologue_end
        vmovaps (%rdx), %ymm0
        vmovaps 32(%rdx), %ymm1
        vmovaps (%rsi), %ymm2
        vmovaps 32(%rsi), %ymm3
        vfmadd213ps     32(%rcx), %ymm1, %ymm3
        vfmadd213ps     (%rcx), %ymm0, %ymm2
        vmovaps %ymm2, (%rdi)
        vmovaps %ymm3, 32(%rdi)
        .loc    1 35 1
        vzeroupper
        retq
```

Quite different.

With some modifications to core.internal.array.operations we can get the codegen for kernel4 to be:

```
0000000000000000 <_D4main7kernel4FKG16fKxG16fKxQgKxQkZv>:
   0:   62 d1 7c 48 28 00       vmovaps (%r8),%zmm0
   6:   62 f1 7c 48 59 02       vmulps (%rdx),%zmm0,%zmm0
   c:   62 f1 7c 48 29 01       vmovaps %zmm0,(%rcx)
  12:   62 d1 7c 48 58 01       vaddps (%r9),%zmm0,%zmm0
  18:   62 f1 7c 48 29 01       vmovaps %zmm0,(%rcx)
  1e:   c5 f8 77                vzeroupper
  21:   c3                      ret
```

The modifications required are:

Turn on the SIMD helper code (gated by DigitalMars).
Switch ``regsz`` from 16 to 64.

Swap out store+load, for their more generic pointer casts. ``*(cast(vec*)p) = val;`` ``return *cast(vec*)p;``

[...]

Issue #5145, opened 2026-06-15
Implement @restrict for slices using llvm.assume["separate_storage"...]


JohanEngelen:
Just saw this in LLVM LangRef that I think can be used to implement restrict-like behavior for slice function parameters, without changing the function signature.

https://github.com/llvm/llvm-project/blob/9ffa02dbaa065dee20ab4d604a3eed042f0b3cc4/llvm/docs/LangRef.rst#L3188-L3190

So something like:
```d
void foo(@restrict int[] a, @restrict int[] b) {
```

```llvm
define void @void example.foo(int[], int[])({ i64, ptr } %a_arg, { i64, ptr } %b_arg) {
; Current prologue:
  %a = alloca { i64, ptr }, align 8
  %b = alloca { i64, ptr }, align 8
  store { i64, ptr } %a_arg, ptr %a, align 8
  store { i64, ptr } %b_arg, ptr %b, align 8
; NEW:
  %a_ptr = getelementptr inbounds { i64, ptr }, ptr %a, i32 0, i32 1
  %b_ptr = getelementptr inbounds { i64, ptr }, ptr %b, i32 0, i32 1
  call void @llvm.assume(i1 true) ["separate_storage"(ptr %a_ptr, ptr %b_ptr)]
  ...
```

First step is to show with `llc` that indeed this result in the desired optimizations.

kinke:
Nice!

kinke:
Implemented in #5148.

PR #5148, opened 2026-06-17
Enable @restrict to slices via llvm.assume["separate_storage"] (#5145)

gulugulubing:
Try to implement #5145: For slice params marked `@restrict`, emit pairwise llvm.assume calls with "separate_storage" operand bundles in the function prologue.  Also prevent the invalid noalias attribute from being applied to non-pointer
parameter types (slices, static arrays, structs, scalars).

### DIFF
--- a/tests/codegen/restrict_slice.d
+++ b/tests/codegen/restrict_slice.d
@@ -22,10 +22,19 @@ void threeSlices(@restrict int[] a, @restrict int[] b, @restrict int[] c) { }
 // CHECK-NOT: separate_storage
 void singleSlice(@restrict int[] a) { }
 
-// Mix: restrict pointer (noalias attr) + restrict slice (separate_storage)
+// Mix: restrict pointer (noalias attr) + restrict slices (separate_storage)
 // CHECK-LABEL: define {{.*}}@{{.*}}mixed
 // CHECK-SAME: ptr noalias %p_arg
+// CHECK-NOT: noalias %a_arg
+// CHECK-NOT: noalias %b_arg
 // CHECK: separate_storage
-void mixed(@restrict int[] a, @restrict int* p) {
+void mixed(@restrict int[] a, @restrict int[] b, @restrict int* p) {
     a[0] = *p;
+    b[0] = *p;
 }
+
+// Restrict on static array (non-pointer type) → should not crash
+// CHECK-LABEL: define {{.*}}@{{.*}}staticArray
+// CHECK-NOT: noalias
+// CHECK-NOT: separate_storage
+void staticArray(@restrict int[4] arr) {}
--- a/tests/codegen/restrict_slice.d
+++ b/tests/codegen/restrict_slice.d
@@ -2,11 +2,13 @@
 
 import ldc.attributes;
 
-// Two restrict slices → one assume with separate_storage
+// Two restrict slices → one pairwise assume with separate_storage
 // CHECK-LABEL: define {{.*}}@{{.*}}twoSlices
 // CHECK: getelementptr inbounds {{.*}} i32 0, i32 1
+// CHECK: [[A:%.restrict.ptr[0-9]*]] = load ptr, ptr
 // CHECK: getelementptr inbounds {{.*}} i32 0, i32 1
-// CHECK: separate_storage
+// CHECK: [[B:%.restrict.ptr[0-9]*]] = load ptr, ptr
+// CHECK: separate_storage{{.*}}ptr [[A]], ptr [[B]]
 void twoSlices(@restrict int[] a, @restrict int[] b) {
     a[0] = 1;
     b[0] = 2;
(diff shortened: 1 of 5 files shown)


tests/codegen/restrict_slice.d
```diff
@@ -0,0 +1,34 @@
+// RUN: %ldc -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+import ldc.attributes;
+
+// Two restrict slices → one assume with separate_storage
+// CHECK-LABEL: define {{.*}}@{{.*}}twoSlices
+// CHECK: getelementptr inbounds {{.*}} i32 0, i32 1
+// CHECK: getelementptr inbounds {{.*}} i32 0, i32 1
+// CHECK: separate_storage
+void twoSlices(@restrict int[] a, @restrict int[] b) {
```
JohanEngelen, line 12: also add a test with one `@restrict` slice and the other normal: then there should not be separate_storage emitted.

tests/codegen/restrict_slice.d
```diff
@@ -0,0 +1,34 @@
+// RUN: %ldc -output-ll -of=%t.ll %s && FileCheck %s < %t.ll
+
+import ldc.attributes;
+
+// Two restrict slices → one assume with separate_storage
+// CHECK-LABEL: define {{.*}}@{{.*}}twoSlices
+// CHECK: getelementptr inbounds {{.*}} i32 0, i32 1
+// CHECK: getelementptr inbounds {{.*}} i32 0, i32 1
+// CHECK: separate_storage
```
JohanEngelen: i think this check should be more strict. You can check that the separate storage was applied to the two getelementptr variables.

JohanEngelen:
would be cool if you can include a test that shows that the desired optimization is performed.

tests/codegen/restrict_slice.d
```diff
@@ -0,0 +1,34 @@
...
+
+// Single restrict slice → no assume generated
+// CHECK-LABEL: define {{.*}}@{{.*}}singleSlice
+// CHECK-NOT: separate_storage
+void singleSlice(@restrict int[] a) { }
+
+// Mix: restrict pointer (noalias attr) + restrict slices (separate_storage)
+// CHECK-LABEL: define {{.*}}@{{.*}}mixed
+// CHECK-SAME: ptr noalias %p_arg
+// CHECK-NOT: noalias %a_arg
```
JohanEngelen, line 37: CHECKs get priority over CHECK-NOTs, so this checks that _after_ `ptr noalias %p_arg` there is no `noalias %a_arg`. However in the function signature in IR it will appear before. So the check does not detect the wrong IR. It's generally best to duplicate CHECK-NOTs a few times, in this case before and after the CHECK-SAME is probably best.

gulugulubing:
> would be cool if you can include a test that shows that the desired optimization is performed.

OK, I'll try it.

thewilsonator (approved):

kinke:
Thx! Unfortunately this doesn't work for tuple parameters, as used for the array ops in https://github.com/ldc-developers/ldc/blob/7e9dae9ea17b641a61762775569e25e9b9902bc8/runtime/druntime/src/core/internal/array/operations.d#L36 Found while trying to fix #4991 via:
```diff
diff --git a/runtime/druntime/src/core/internal/array/operations.d b/runtime/druntime/src/core/internal/array/operations.d
index 7e5b5f43e9..fa994bb53c 100644
--- a/runtime/druntime/src/core/internal/array/operations.d
+++ b/runtime/druntime/src/core/internal/array/operations.d
@@ -12,6 +12,8 @@ import core.internal.traits : Filter, staticMap, Unqual;
 version (GNU) version = GNU_OR_LDC;
 version (LDC) version = GNU_OR_LDC;
 
+version (LDC) import ldc.attributes : restrict;
+
 /**
  * Perform array (vector) operations and store the result in `res`.  Operand
  * types and operations are passed as template arguments in Reverse Polish
@@ -33,7 +35,7 @@ version (LDC) version = GNU_OR_LDC;
  *
  * Returns: the slice containing the result
  */
-T[] arrayOp(T : T[], Args...)(T[] res, Filter!(isType, Args) args) @trusted
+T[] arrayOp(T : T[], Args...)(@restrict T[] res, @restrict Filter!(isType, Args) args) @trusted
 {
     alias scalarizedExp = staticMap!(toElementType, Args);
     alias check = typeCheck!(true, T, scalarizedExp); // must support all scalar ops
```

**Edit**: It compiles fine, but the `@restrict` for the `args` tuple seems ignored.

JohanEngelen:
This deserves a release note!

gulugulubing:
It seems dmd/typesem.d strips tuple's userAttributes when flatting the tuple. Since this is inpurely frontend code, not LDC-specific, should I fix it upstream to dlang/dmd? Or keep it as an LDC-local patch ?

JohanEngelen:
> It seems dmd/typesem.d strips tuple's userAttributes when flatting the tuple. Since this is inpurely frontend code, not LDC-specific, should I fix it upstream to dlang/dmd? Or keep it as an LDC-local patch ?

maybe ask first on the forums if it is OK to propagate UDAs to tuple members. If not, then it could be a local LDC patch to propagate only `@restrict` to tuple members.

gulugulubing:
> maybe ask first on the forums if it is OK to propagate UDAs to tuple members. If not, then it could be a local LDC patch to propagate only `@restrict` to tuple members.

It's confirmed by the forums: As for the compiler's "tuples", it makes sense to propagate UDAs to memebers.

PR #5149, opened 2026-06-19
Fix/Emit array comparison/equality hooks for device code

badnikhil:
Array comparison/equality (`==` `!=` `<` `>` `<=` `>=`) on arrays/slices in `@compute` device code lowers to the druntime hooks `core.internal.array.comparison.__cmp` and `core.internal.array.equality.__equals`
  (plus helpers `isEqual`/`at`, and `core.internal.string.dstrcmp` for `char`). These live in host-only modules and were dropped during dcompute semantic analysis and codegen, ending up as hollow `declare`-only
  stubs — unresolved at device link/JIT (and a compiler crash before #5142). This makes them emit real `ptx_device`/SPIR-V bodies.
  ## Background ##
  - Before , the dcompute semantic walk descended into these host-only templates and segfaulted.
  - #5142 stopped the crash by skipping *all* host-side template instantiations — but that also skipped the comparison/equality hooks, which are device-legal and meant to work, leaving them as hollow stubs.  
## FIX ##
  Adds `isDeviceArrayComparisonHook()` in  (`gen/uda`) recognizing the hooks(`__cmp`/`__equals`/`isEqual`/`at` by module, `dstrcmp` by symbol), and:
 - allows the kernel's call into them during dcompute semantic analysis (`gen/semantic-dcompute.cpp`), and
  - exempts them from the codegen host-only-template skip so their real device bodies are emitted (`gen/declarations.cpp`).
  -  ` _d_arraybounds_index`  its illegal (on device) because it throws RangeError. so never emit in device code.


  The semantic walk still doesn't descend into the hook bodies; codegen emits the device-reachable instances . Compiler-only — no druntime changes.

 ## Further  Required Fix
.The memcmp-able fast path (`int[]`/`char[] ==` lowers to an inline `memcmp`,  and `char[] <` reaches `memcmp` via `dstrcmp` But still references a device `memcmp`, which druntime/dcompute doesn't provide (`core.stdc.string.memcmp` is  only an `extern(C)` libc binding with no body). That's a separate  gap and is not addressed here.

The tests are generated with Claude code ( Opus 4.8 - xhigh)

### DIFF
--- a/gen/uda.cpp
+++ b/gen/uda.cpp
@@ -16,6 +16,7 @@
 #include "ir/irvar.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringSwitch.h"
+#include <cstring>
 #include <stdio.h>
 
 using namespace dmd;
@@ -642,6 +643,28 @@ extern "C" DComputeCompileFor hasComputeAttr(Dsymbol *sym) {
   checkStructElems(sle, {Type::tint32});
 
   return static_cast<DComputeCompileFor>(1 + (*sle->elements)[0]->toInteger());
+}
+
+bool isDeviceArrayComparisonHook(Dsymbol *sym) {
+  if (!sym)
+    return false;
+  Module *m = sym->getModule();
+  if (!m)
+    return false;
+  // These druntime modules are host-only, but their templates (`__cmp`,
+  // `__equals` and helpers like `isEqual`/`at`) are the lowering hooks emitted
+  // for array `<`/`==` in device code, so they must not be skipped.
+  const char *fqn = m->toPrettyChars();
+  if (!strcmp(fqn, "core.internal.array.comparison") ||
+      !strcmp(fqn, "core.internal.array.equality"))
+    return true;
+  // From core.internal.string only `dstrcmp` is a device-legal hook helper (the
+  // `char`/string `__cmp` specialization calls it). The rest of that module
+  // (e.g. signedToTempString) contains device-illegal code, so don't exempt the
+  // whole module.
+  if (!strcmp(fqn, "core.internal.string"))
+    return sym->ident && !strcmp(sym->ident->toChars(), "dstrcmp");
+  return false;
 }
 
 /// Returns whether `sym` has the `@ldc.dcompute._kernel()` UDA applied.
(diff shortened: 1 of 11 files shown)


gen/uda.cpp
```diff
@@ -644,6 +645,28 @@ extern "C" DComputeCompileFor hasComputeAttr(Dsymbol *sym) {
...
+  if (!strcmp(fqn, "core.internal.array.comparison") ||
+      !strcmp(fqn, "core.internal.array.equality"))
+    return true;
+  // From core.internal.string only `dstrcmp` is a device-legal hook helper (the
+  // `char`/string `__cmp` specialization calls it). The rest of that module
+  // (e.g. signedToTempString) contains device-illegal code, so don't exempt the
+  // whole module.
+  if (!strcmp(fqn, "core.internal.string"))
+    return sym->ident && !strcmp(sym->ident->toChars(), "dstrcmp");
+  return false;
```
thewilsonator: check identifiers rather than using `strcmp`, see other parts of this file as an example of how to do this

badnikhil: done

gen/uda.cpp
```diff
@@ -644,6 +644,36 @@ extern "C" DComputeCompileFor hasComputeAttr(Dsymbol *sym) {
...
+    return false;
+  // These druntime modules are host-only, but their templates (`__cmp`,
+  // `__equals` and helpers like `isEqual`/`at`) are the lowering hooks emitted
+  // for array `<`/`==` in device code, so they must not be skipped.
+  Identifier *const internal = Identifier::idPool("internal");
+  if (md->packages.length == 3 && md->packages.ptr[0] == Id::core &&
+      md->packages.ptr[1] == internal &&
+      md->packages.ptr[2] == Identifier::idPool("array") &&
+      (md->id == Identifier::idPool("comparison") ||
+       md->id == Identifier::idPool("equality")))
```
thewilsonator: these `idPool` should be added to `dmd/id.h`/`dmd/id.d` so they are not called repeatedly.

gen/uda.cpp
```diff
@@ -644,6 +644,36 @@ extern "C" DComputeCompileFor hasComputeAttr(Dsymbol *sym) {
...
+    return true;
+  // From core.internal.string only `dstrcmp` is a device-legal hook helper (the
+  // `char`/string `__cmp` specialization calls it). The rest of that module
+  // (e.g. signedToTempString) contains device-illegal code, so don't exempt the
+  // whole module.
+  if (md->packages.length == 2 && md->packages.ptr[0] == Id::core &&
+      md->packages.ptr[1] == internal &&
+      md->id == Identifier::idPool("string"))
+    return sym->ident == Identifier::idPool("dstrcmp");
```
thewilsonator: ditto

badnikhil: @thewilsonator done

thewilsonator (approved):

PR #5154, opened 2026-06-22
Add noalias/allocsize attributes to GC allocation runtime functions

kubo39:
Mark the returned pointer of _d_allocmemory, _d_allocmemoryT and _d_newclass/_d_allocclass with noalias, and give _d_allocmemory an allocsize(0).
This helps alias analysis and enables llvm's objectsize-based optimizations.


thewilsonator:
Looks good,
```
The following tests FAILED:
	1716 - druntime-test-exceptions-release (Failed)
```
not sure why though.

JohanEngelen:
Are we sure this works when doing something like:

```
new
delete (GC collection)
new
delete (GC collection)
new
delete (GC collection)
new
```

where all `new`s might actually return the same memory address. Will LLVM's optimizer make the right decisions?

A `delete` means that we have no references any more to the item, so in D code, the `noalias` must be correct because any `new` call will return a pointer that is indeed not aliasing any _existing_ memory reference any more.
However, in the LLVM IR optimizer, this "temporal" connection might be gone after reordering of code. I did not think this through myself yet.

kubo39:
Looks like it's because of SimplifyDRuntimeCalls.

PR #5156, opened 2026-06-22
DCompute: Inline memcmp for device array equality

badnikhil:
## Problem
  The comparison/equality hooks added in #5149  lower array `==`/`<` to a `memcmp` call (`DtoArrayEqCmp_memcmp`).That works on the host, but device  targets have no libc to provide `memcmp`, and there is no `llvm.memcmp` intrinsic to fall back on ,so the call has nothing to bind to on the device.
  
  ## Explanation of the Fix

   - Added emitInlineMemcmp(), which emits a small loop (cond/body/diff/inc/end blocks) that walks the two buffers byte-by-byte and returns an i32 following C memcmp convention: 0 when all sizeInBytes bytes are
  equal, otherwise the unsigned difference of the first mismatching byte.
  - callMemcmp() now branches on gIR->dcomputetarget: device code gets the inline loop, host code keeps the existing memcmp runtime call. The byte-size computation (numElements * elementSize) is shared by both
  paths.
  - Modified the PHI in DtoArrayEqCmp_memcmp: because callMemcmp can now emit extra basic blocks, the merge phi takes its incoming value from the block we actually end up in (memcmpResultBB = GetInsertBlock()) rather
  than assuming the result is still produced in memcmpBB.

### DIFF
--- a/gen/arrays.cpp
+++ b/gen/arrays.cpp
@@ -679,18 +679,73 @@ DSliceValue *DtoAppendDCharToUnicodeString(Loc loc, DValue *arr,
 
 ////////////////////////////////////////////////////////////////////////////////
 namespace {
-// Create a call instruction to memcmp.
-llvm::CallInst *callMemcmp(Loc loc, IRState &irs, LLValue *l_ptr,
-                           LLValue *r_ptr, LLValue *numElements, LLType *elemty) {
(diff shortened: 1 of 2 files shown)


thewilsonator (commented):
test?

thewilsonator (approved):

PR #5157, opened 2026-06-23
Add `llvm_is_fpclass` (`llvm.is.fpclass`) to `ldc.intrinsics`

QuantumSegfault:
Adds `llvm_is_fpclass` to `ldc.intrinsics` (for scalars only).

https://llvm.org/docs/LangRef.html#llvm-is-fpclass-intrinsic

### DIFF
--- a/runtime/druntime/src/ldc/intrinsics.di
+++ b/runtime/druntime/src/ldc/intrinsics.di
@@ -378,9 +378,9 @@ pragma(LDC_intrinsic, "llvm.maximum.f#")
     T llvm_maximum(T)(T vala, T valb)
         if (__traits(isFloating, T));
 
-/// The ‘llvm.is.fpclass’ intrinsic returns a boolean value or vector of
-/// boolean values depending on whether the first argument satisfies the
-/// test specified by the second argument.
+/// The ‘llvm.is.fpclass’ intrinsic returns a boolean value depending on
+// whether the first argument satisfies the test specified by the second
+// argument.
 ///
 /// The function checks if `op` belongs to any of the floating-point classes
 /// specified by `test`.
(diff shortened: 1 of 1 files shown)


QuantumSegfault:
Requesting review

@thewilsonator @JohanEngelen

runtime/druntime/src/ldc/intrinsics.di
```diff
@@ -378,6 +378,32 @@ pragma(LDC_intrinsic, "llvm.maximum.f#")
     T llvm_maximum(T)(T vala, T valb)
         if (__traits(isFloating, T));
 
+/// The ‘llvm.is.fpclass’ intrinsic returns a boolean value or vector of
+/// boolean values depending on whether the first argument satisfies the
```
thewilsonator: this doesn't look to handle the vector return case (though I'm not sure we can or need to).

thewilsonator (approved):

runtime/druntime/src/ldc/intrinsics.di
QuantumSegfault: No. I copied those docs from the LangRef.

Boolean vector output is for when you have a vector input.

thewilsonator: then please update the doc to either remove reference to the vector case or add a note that it is not (yet) supported.

QuantumSegfault: Done (vector mention removed).

Something to re-evaluate if/when support for vector intrinsics gets added.

thewilsonator: FWIW we have intrinsic support for vectors, but not for vector of bools.

PR #5162, opened 2026-06-28
Add support for Wasm EH

QuantumSegfault:
Adds codegen and DRuntime support for WebAssembly exception handling.


QuantumSegfault:
Some IR comparisons for reference

C++/Clang: https://godbolt.org/z/heMqqGvbb
D/LDC (this PR): https://godbolt.org/z/vMYWYvPE5

QuantumSegfault:
@kinke 

Maybe @rainers?

It works, but maybe one of you has a better idea on how to organize the modifications to [‎`gen/trycatchfinally.cpp`‎](https://github.com/ldc-developers/ldc/pull/5162/changes#diff-a2d5c1832f0a59af174250418bdcc4816bc725405fd1cbc93c36b3efa0fc92d4).

Wasm EH follows MSVC style `catchpad`/`catchswitch`/`catchret` & `cleanpad`/`cleanupret`. However there is only one `catchswitch` per `catchpad`, after which an Itanium-landingpad-style comparison chain needs to be emitted to determine the correct `catch` block (if any).

Right now I'm treating it more like MSVC, and just dumping everything in `emitCatchBodies`...though this requires exposing some `TryCatchFinallyScope` internals...

rainers:
Sorry, it's been so long since I worked on the EH, I don't remember any details. I doubt I can be of any help here.

QuantumSegfault:
> Sorry, it's been so long since I worked on the EH, I don't remember any details. I doubt I can be of any help here.

That's unfortunate...

Do you know of anyone else who might be more familiar with the EH code at this point (MSVC especially)?

rainers:
I suspect that would only be @kinke.

kinke:
Cool, thanks for tackling this! - Unfortunately I'm not really familiar with the MSVC EH stuff either, as Rainer thankfully did all the work. ;) It has just come up again though, as related #3504 becomes more pressing, being a blocker for the D v2.113 merge in https://github.com/ldc-developers/ldc/pull/5113#issuecomment-4479376515. I really hate that LLVM doesn't take care of this for us (accepting the generic Itanium EH IR representation for all targets and lowering it appropriately as needed).

QuantumSegfault:
> Cool, thanks for tackling this! - Unfortunately I'm not really familiar with the MSVC EH stuff either, as Rainer thankfully did all the work. ;) It has just come up again though, as related #3504 becomes more pressing, being a blocker for the D v2.113 merge in [#5113 (comment)](https://github.com/ldc-developers/ldc/pull/5113#issuecomment-4479376515). I really hate that LLVM doesn't take care of this for us (accepting the generic Itanium EH IR representation for all targets and lowering it appropriately as needed).

Okay, then I guess we just have to consider what we have, and ignore prior rationale.

On one hand, doing what I did works, and makes it fairly easy to keep it inline with any fixes to MSVC EH (which will most likely impact Wasm EH as well).

On the other hand, it's kind of ugly from a code cleanliness point of view, cause I'm jamming what should probably go into `emitLandingPad` into `emitWasmCatchBodies`. Then again, MSVC EH seems to be using `emitLandingPad` to manage cleanup scopes really, not true landing pads. Try to fit in into there seems like it would be more difficult.

I mean, technically we could always refactor it later, right? But when's "later"?

I guess I'll just mark this "Ready for review" and we'll go from there.

QuantumSegfault:
So, what's the word on this?

Are we waiting on some other PR, or does it just still need to be reviewed?

rainers:
I took a cursory look at the changes and they seem pretty solid. AFAICT they don't change anything outside of WASM, so probably low risk to merge.
I wonder whether you can run the test suite for WASM. If yes, why are there no tests enabled that didn't pass before?

QuantumSegfault:
> I took a cursory look at the changes and they seem pretty solid. AFAICT they don't change anything outside of WASM, so probably low risk to merge. I wonder whether you can run the test suite for WASM. If yes, why are there no tests enabled that didn't pass before?

Technically I do make a couple of (what should be) NFC changes; making `getOrCreateEhSelectorSlot`, and making it and `getOrCreateEhPtrSlot` public (exposing it from `TryCatchFinallyScope` to `TryCatchScope`). But other than that, yeah, all the changes should otherwise be to Wasm only.


Which testsuite? There are various tests (unittests) for DRuntime that can now pass, but that can't really be run in LDC's current state (upstream changes need to come down, and for nearly full test-suite passing [-threads], we need this and GC).

Unless you mean the main compiler testsuite? Most tests won't run be run for Wasm, as it's not the default/native triple (in any CI build; or any vanilla LLVM distribution). Do you want a `codegen` test to be added?

rainers:
I never tried wasm, but as there are wasm32/wasm64 targets for years I thought it was ready to use with some restrictions. Just ran my very first -betterC wasm program...

But indeed, the D runtime doesn't seem ready yet. I suspect you could run 99% of the dmd/ldc test suite with the "manual" GC, though.

rainers (approved):

QuantumSegfault:
> I never tried wasm, but as there are wasm32/wasm64 targets for years I thought it was ready to use with some restrictions. Just ran my very first -betterC wasm program...
> 
> But indeed, the D runtime doesn't seem ready yet. I suspect you could run 99% of the dmd/ldc test suite with the "manual" GC, though.

Not sure about 99%, but most of the testsuite can pass, yeah.

The problem is the "manual" GC itself doesn't actually pass tests. Doesn't handle block info and finalizers and all that properly.

PR #5168, opened 2026-07-04
Enable @restrict for tuple parameters (follow-up to #5148)

gulugulubing:
Follow-up to #5148. The fix is a one-line change in dmd/typesem.d: when expanding a tuple parameter into individual elements, propagate `userAttribDecl` from the outer parameter to each element.

Also submitted upstream: https://github.com/dlang/dmd/pull/23358.


kinke:
Thanks! Something went wrong during the rebase though, there are still the 7 commits from #5148. We only need the latest 2.

gulugulubing:
> Thanks! Something went wrong during the rebase though, there are still the 7 commits from #5148. We only need the latest 2.

OK. Fixed the git history.

PR #5170, opened 2026-07-06
Fix #5144 on LLVM 23

QuantumSegfault:
Use `getTerminatorOrNull` instead of `getTerminator` on LLVM 23 (after llvm/llvm-project#189416).


QuantumSegfault:
@thewilsonator

PR #5173, opened 2026-07-06
druntime: Fix `getStackBottom` on Wasm

QuantumSegfault:
Stack grows down on Wasm, not up.

Also, it's the address of `__stack_high` that marks the location, not the value.


QuantumSegfault:
@thewilsonator

thewilsonator (approved):

PR #5175, opened 2026-07-06
Add predefined global dcompute versions for cuda, ocl, metal

asindarov:
Hello, This PR adds `LDC_DCompute_CUDA`, `LDC_DCompute_OCL`, `LDC_DCompute_Metal` corresponding to what users specifies for `mdcompute-targets` command line option.

### DIFF
--- a/driver/cl_options.cpp
+++ b/driver/cl_options.cpp
@@ -751,7 +751,7 @@ cl::opt<unsigned>
                    cl::desc("Warn for stack size bigger than the given number"),
                    cl::value_desc("threshold"));
 
-#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX
+#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX || LDC_LLVM_SUPPORTED_TARGET_AArch64
 cl::list<std::string>
     dcomputeTargets("mdcompute-targets", cl::CommaSeparated,
                     cl::desc("Generates code for the specified DCompute target"
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -993,7 +993,7 @@ void registerPredefinedVersions() {
 
     #if LDC_LLVM_SUPPORTED_TARGET_NVPTX
       if (targetInfo.starts_with("metal")) {
-          VersionCondition::addPredefinedGlobalIdent("LDC_DCompute_METAL");
+          VersionCondition::addPredefinedGlobalIdent("LDC_DCompute_Metal");
       }
     #endif
   }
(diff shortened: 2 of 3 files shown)


driver/main.cpp
```diff
@@ -975,6 +976,28 @@ void registerPredefinedVersions() {
...
+
+    #if LDC_LLVM_SUPPORTED_TARGET_SPIRV
+      if (targetInfo.starts_with("ocl")) {
+            VersionCondition::addPredefinedGlobalIdent("LDC_DCompute_OCL");
+      }
+    #endif
+
+    #if LDC_LLVM_SUPPORTED_TARGET_NVPTX
+      if (targetInfo.starts_with("metal")) {
+          VersionCondition::addPredefinedGlobalIdent("LDC_DCompute_METAL");
```
JohanEngelen: I think "Metal" is not commonly capitalized, so let's not do it here either.

asindarov: fair, I updated it

driver/main.cpp
```diff
@@ -975,6 +976,28 @@ void registerPredefinedVersions() {
   }
 #endif
 
+  for(auto& dcomputeTarget: dcomputeTargets) {
+    auto targetInfo = llvm::StringRef(dcomputeTarget);
+    
+    #if LDC_LLVM_SUPPORTED_TARGET_NVPTX
```
JohanEngelen: instead of these `#if`, move the whole loop into the `#if` on line 972

asindarov: Done

thewilsonator (approved):

driver/main.cpp
```diff
@@ -969,7 +970,23 @@ void registerPredefinedVersions() {
...
+
+    if (targetInfo.starts_with("ocl")) {
+      VersionCondition::addPredefinedGlobalIdent("LDC_DCompute_OCL");
+    }
+
+    if (targetInfo.starts_with("metal")) {
+      VersionCondition::addPredefinedGlobalIdent("LDC_DCompute_Metal");
+    }
+  }
+
```
thewilsonator, line 989: I think you are missing an `#endif` here, also see the CI failures.

asindarov: Fixed

driver/cl_options.cpp
```diff
@@ -751,7 +751,7 @@ cl::opt<unsigned>
                    cl::desc("Warn for stack size bigger than the given number"),
                    cl::value_desc("threshold"));
 
-#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX
+#if LDC_LLVM_SUPPORTED_TARGET_SPIRV || LDC_LLVM_SUPPORTED_TARGET_NVPTX || LDC_LLVM_SUPPORTED_TARGET_AArch64
```
JohanEngelen, line 754: nitpick: AArch64 is pretty much always supported as target, so might as well get rid of this conditional compilation altogether.

asindarov: Unfortunately as I recall CI was failing because AArch64 was not included in the `#if`, erroring `dcomputeTargets` is not defined

PR #5178, opened 2026-07-10
Add `WasmPointersSpill` pass to force potential pointers onto stack

QuantumSegfault:
This PR adds an LLVM pass to run after all other optimization (but before LTO). It searches for any IR vreg that MAY contain a pointer (is either definitely a LLVM pointer, or is transitively used by a `inttoptr`) that crosses any sort of `call` (any of which might trigger a GC collection), and `store volatile`s them to a new alloca slot right after it's def if needed.

This forces the potential GC pointer onto the stack, so it can be seen during GC collection on Wasm. Unlike native platforms, Wasm has opaque `locals` and a value stack that cannot be spilled or introspected (yet). Spilling such pointers makes sure that they will be seen during the GC's sweep of the normal stack (in Wasm linear memory).

Alternative to #5174


QuantumSegfault:
@kinke 
@JohanEngelen 

Here's the other idea.

It actually ended up being not as complicated as I thought (unless I'm missing something; which I probably am).

This way we don't have to play whack-a-mole trying to track down pointers and make sure they are `alloca` and pinned across the front-end.

It's probably a little over-aggressive/convservative, but I think it turned out nicely. Any pointers, any contributors to `inttoptr` get marked as being "PotentialPointers". So this helps catches D-side casts and unions that get optimized...smuggling pointers in float, int, etc. But it's only needed across calls (which I currently assume ALL calls might trigger GC; including intrinsics...). So it ends up spilling a reasonably minimal amount.

QuantumSegfault:
Probably needs some more work to handle stores/loads to `alloca`...when a `load` is reached backwards from `inttoptr` need to continue checking through all stores into said `alloca`.

QuantumSegfault:
> An LLVM pass like the last one might be a way forward for now if you don't wanna wait.

Seems you were right @kinke. Just took me a bit to figure out how to recover the information.

QuantumSegfault:
Right. That should do it.

Should be reasonably optimized, but conservative when it comes to things that MAY be pointers.

QuantumSegfault:
Okay, this needs a little bit more tweaking to account for `invoke` and fix `lifetime.end` placement in certain cases...need to create more BB to split edges.

QuantumSegfault:
Okay. Ready. Now it works on DRuntime.

JohanEngelen:
I think it'd help if you could add some codegen tests. That will make it easier to discover issues than tracking down a failing druntime/phobos test.

QuantumSegfault:
> I think it'd help if you could add some codegen tests. That will make it easier to discover issues than tracking down a failing druntime/phobos test.

Working on it. Good suggestion...helped me find another little issue.

I'll try to get some decent coverage, but I won't be able to think of EVERYTHING.

Try to get tests at least for some of the issues I ran into along the way.

gen/optimizer.cpp
```diff
@@ -511,6 +525,10 @@ void runOptimizationPasses(llvm::Module *M, llvm::TargetMachine *TM) {
 
   pb.registerOptimizerLastEPCallback(addStripExternalsPass);
 
+  if (TM->getTargetTriple().isWasm()) {
+    pb.registerOptimizerLastEPCallback(addWasmPointersSpillPass);
```
kinke, line 543: It might pay off to only add the pass if the GC might actually be used, i.e., not adding it with `-betterC` and/or when targeting naked wasm without an OS (~only for WASI and emscripten).

QuantumSegfault: Yeah, we can condition this on `global.params.useGC`.

I also want to, down the line, pass through `@nogc` info via LLVM metadata so we can also filter per function.

Just still trying to iron out all the edge cases with this first.

QuantumSegfault:
Ok, I still have more tests to write, but I'm pretty confident in what I have now.

The big pain has been placing the `lifetime.end` markers ideally (and we DO want/need them, to help mitigate the stack/memory overhead of this).

QuantumSegfault:
Requesting (re)review

@JohanEngelen

QuantumSegfault:
Some code-size diffs:

On `master` (6c745f9)
```
1979112  druntime-test-runner
8971262  druntime-test-runner-debug

3169714  libdruntime-ldc-debug.a
14489024 libdruntime-ldc-unittest-debug.a
3146712  libdruntime-ldc-unittest.a
1291652  libdruntime-ldc.a
```

With this PR merged in:
```
2177264  druntime-test-runner
9268808  druntime-test-runner-debug

3235860  libdruntime-ldc-debug.a
14904210 libdruntime-ldc-unittest-debug.a
3410360  libdruntime-ldc-unittest.a
1373972  libdruntime-ldc.a
```

So not ideal. +2 - 10%.

-------

There are more oppurtunities to elide spills when dealing with intrinsics (which can't trigger GC), and when dealing with `@nogc` functions, so hopefully we can mitigate some of this in the future. But...this is the cost of correctness.

kinke:
Thx for the numbers! How about adding an opt-out switch? I'm thinking about cases where one might design a wasm application to be used with disabled GC only, still wanting to have full WASI druntime (and no `@nogc` restrictions), but wanting to avoid this spilling because the GC won't ever run.

tests/codegen/wasm_spill_pointers_pass_nogc.d
```diff
@@ -0,0 +1,20 @@
+// REQUIRES: target_WebAssembly, link_WebAssembly
```
kinke: Creating IR/LL alone doesn't require an enabled LLVM wasm backend, nor linkability. So no `REQUIRES` needed, same for the other test.

QuantumSegfault:
Done. Anything else?

kinke (approved):
Thanks for all this! I can't really afford reviewing pass and the huge test in detail, will just trust you on that.

QuantumSegfault:
> Thanks for all this! I can't really afford reviewing pass and the huge test in detail, will just trust you on that.

Well, then here's hoping I didn't miss anything! 

I don't normally do this kind of stuff (nor have I taken CS courses that might have taught compiler fundamentals). But working on and with LLVM (mainly Wasm instruction selection) has familiarized me enough with the basics to make this easy enough.

I don't forsee there being any more problems with lifetime and spill placement; that's pretty well tested (unless...it's possible to somehow create a `callbr` with D> but there is an assertion that will trip, so that's covered). But `PotentialPointers` may need to be refined further. I've tried to make it err on the side of being conservative; same with finding calls that might trigger GC; currently all calls, including intrinsics are considered doing so. So there SHOULDN'T be any underspilling (though passing/storing pointers in non-pointer arguments/returns/globals is UB as far as I know, so assumptions I've made can cause issues if anybody tries doing that).

There is one thing I deliberately omitted and that's spilling incoming arguments in the function/callee. The assumption being that either D code is calling other D code and the caller will handle the spill. Or it's C code calling into D code, in which it still shouldn't be a problem because `GC.addRoot` should be called before handing GC pointers of to C land anyway.

----------

Worst case this can be incrementally refined later.

kinke:
> Worst case this can be incrementally refined later.

Yeah exactly.

> nor have I taken CS courses that might have taught compiler fundamentals

Welcome to the club! :)

QuantumSegfault:
How come you didn't squash? Now there's 32 ugly commits scattered about.

kinke:
A quick glance at the commit list seemed okay, to document the evolution. [I did notice the last 'stray-merge-relict' conflict, but oh well.]

PR #5183, opened 2026-07-13
CI: Move Cirrus FreeBSD job to GitHub Actions

### LINKED ISSUE #5138: CI: Replace dead Cirrus CI

kinke:
Resolves #5138. The FreeBSD image version was raised from 13.5 to 14.4 - 13 is EOL by now, and the 13.5 image had weird segfaults with libcurl (via `std.net.curl`) in the new CI job.

As a side-effect, the prebuilt FreeBSD ldc2 executable is now also PGO'd and includes mimalloc.


.github/workflows/main.yml
```diff
@@ -142,14 +142,39 @@ jobs:
...
+              -DEXTRA_CXXFLAGS=-flto=full
+            with_pgo: true
+
     name: ${{ matrix.job_name }}
     runs-on: ${{ matrix.os }}
     container:
       image: ${{ matrix.container_image }}
     timeout-minutes: 90
     env:
+      GHA_BASH_SHELL: ${{ matrix.vm_os && 'cpa.sh {0}' || 'bash' }}
```
kinke, line 165: Thx @jacob-carlborg, the recently added `cpa.sh` shell wrapper is a life-saver, allowing to use most of our internal composite actions (via `shell: ${{ env.GHA_BASH_SHELL }}` for all run-steps expecting a bash shell). :+1:

jacob-carlborg: Thanks, that's great to hear.

kinke:
There's one weirdness so far - every (dlang) app loading libcurl dynamically seems to segfault at program shutdown. This includes all 4 `std.net.curl[-shared][-debug]` unittest runs, as well as running the just-built dub to build reggae (after downloading the dub dub-package via libcurl and successfully building and copying the reggae executable). Using `ubuntu-latest` as host, and the FreeBSD 13.5 image (the same FreeBSD version we've been using until not too long ago for Cirrus CI). Smells emulation-related to me…

**Edit**: Oh, the FreeBSD 14.4 image seems to work fine.

PR #5185, opened 2026-07-14
Fix Wasm `getStackTop` on older LLVM

QuantumSegfault:
Fixes `getStackTop` to work on older versions of LLVM.

`llvm_stackaddress` was introduced in LLVM 22. Use `llvm_stacksave` instead when needed.

This also fixes the intrinsic names/mangling of `llvm_stacksave` and `llvm_stackrestore` after their adjustment in LLVM 18.

### DIFF
--- a/runtime/druntime/src/core/thread/osthread.d
+++ b/runtime/druntime/src/core/thread/osthread.d
@@ -1688,7 +1688,7 @@ in (fn)
             }}
             asm pure nothrow @nogc {
                 ("sd $gp, %0") : "=m" (regs[8]);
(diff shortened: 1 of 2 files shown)


QuantumSegfault:
@thewilsonator

QuantumSegfault:
Actually... looks like https://github.com/ldc-developers/ldc/commit/b7e7e2f36f216164c71a3b03a173d258046e0e54 dropped support for LLVM older than 18.

Would it be okay then to drop it here as well? Doing so would remove most of the complexity.

kinke:
Yeah sure, absolutely.

thewilsonator (approved):

PR #5186, opened 2026-07-16
Cherry-pick upstream changes for WASI

QuantumSegfault:
Cherry-picks all of my WASI related changes thus far from `dmd-rewrite-master`.

422ab4fab4fbbef524d2e6208e647075bee8ce6d isn't a pure cherry-pick, as I had to backport the threading changes (upstream is currently refactoring `osthread.d`).


QuantumSegfault:
@thewilsonator 
@kinke

runtime/druntime/src/core/sys/wasi/posix/time.d
```diff
@@ -0,0 +1,37 @@
...
+
+alias long suseconds_t;
+
+struct timeval
+{
+    time_t      tv_sec;
+    suseconds_t tv_usec;
+}
+
+pragma(mangle, muslRedirTime64Mangle!("clock_getres", "__clock_getres_time64"))
```
kinke: This seems off - `muslRedirTime64Mangle` is an identity-function on non-musl.

runtime/druntime/src/rt/profilegc.d
```diff
@@ -13,6 +13,12 @@
 
 module rt.profilegc;
 
+// On Wasm, it increases codesize noticably.
+// Meanwhile, ldc (ldmd2) does not support `-profile-gc`, so there's no point
+// keeping it around on Wasm targets
+version (WASI) {}
```
kinke: Hmm, code size of what? If unused, the object file shouldn't be linked from static druntime.

(We might be able to add `-profile=gc` support nowadays, as I think I saw that this is now handled in the frontend, not the glue layer anymore.)

runtime/druntime/src/core/internal/abort.d
```diff
@@ -41,6 +41,94 @@ void abort(scope string msg, scope string filename = __FILE__, size_t line = __L
...
+        extern(C) struct streams_stream_error_t {
+            ubyte tag;
+
+            union Val {
+                streams_own_error_t     last_operation_failed;
+            }
+            Val val;
+        }
+
+        pragma(mangle, "streams_output_stream_drop_own")
```
JohanEngelen, line 85: why are these pragma mangles needed? Doesn't `extern(C)` already give you the exact same mangle as function identifier?

kinke: IIRC, that doesn't hold for *nested* funcs - they always get a D mangle, just the calling convention is C then.

runtime/druntime/src/rt/profilegc.d
QuantumSegfault: The final binary once statically linked.

The `static ~this` will always get run. And it drags in some things that programs might never use:
- qsort
- fopen
- fclose
- fprintf

Whatever the exact cause, it does have an effect:
```d
import core.stdc.stdio;

// Without profilegc: 684K	
// With profilegc: 844K
void main() {
    printf("Hello, WASI!\n");
}
```

To be fair though, this is at -O3 instead of -Oz. If everything were compiled at -Oz (particularly `wasi-libc`, the difference wouldn't be as large). Or more complex files that do interact with files would mask it as well.

kinke: Hmm. There's https://github.com/ldc-developers/ldc/blob/702471d5d82e7c4d9c2fb784cbc487919c87d5f7/runtime/druntime/src/core/internal/array/utils.d#L24-L37
which drags in this `rt.profilegc` module AFAICT. Maybe this function should be a dummy-template instead, so that the `TraceHook` instantiations instantiate it as needed instead of baking this wrapper into prebuilt druntime.

**Edit**: Or moving this function wrapper to the `rt.profilegc` module (=> same object file in static druntime).

runtime/druntime/src/core/sys/wasi/posix/time.d
QuantumSegfault: Well, on one hand, being a musl fork:

https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/include/time.h#L171

----------

On the other, in practice, `wasi-libc` is a new enough fork targeting a new enough arch that `_REDIR_TIME64` is never set. What do you want to do?

runtime/druntime/src/rt/profilegc.d
QuantumSegfault: That's not the issue. It's the module destructors. Leaving `accumulate` has no visible impact. It's the two destructors. Mainly the second one (the first contributes to ~8K; the rest is the second and deps).

runtime/druntime/src/core/sys/wasi/posix/time.d
kinke: Oh, but we don't predefine `CRuntime_Musl` for WASI AFAICT, just for Emscripten targets?

runtime/druntime/src/rt/profilegc.d
kinke: I'm pretty sure that is the issue, becausing nothing else imports that module (client code cannot even), so the module dtor shouldn't matter as long as nothing references the object file.

runtime/druntime/src/core/sys/wasi/posix/time.d
QuantumSegfault: Correct. Because `wasi-libc` != `musl`

runtime/druntime/src/rt/profilegc.d
QuantumSegfault: Oh...wow, you're right. Removing `accumulate` did it.

I didn't realize entire modules would be GC'd like that.

runtime/druntime/src/core/sys/wasi/posix/time.d
kinke: So why use the mangle logic if it has no effect?

runtime/druntime/src/rt/profilegc.d
kinke: Okay!

[This is how linkers work with static libs, only pulling object files out of static libs if they provide a symbol to be resolved.]

QuantumSegfault: Or even better: just move `accumulatePure` inside the `D_ProfileGC` block. That way it's never used if never defined

runtime/druntime/src/core/sys/wasi/posix/time.d
QuantumSegfault: Really no good reason. I'll remove them and add a comment explaining that it's not needed (despite the original C header having it).

runtime/druntime/src/rt/profilegc.d
kinke: > Or even better: just move accumulatePure instead the D_ProfileGC block. That way it's never used if never defined

If you meant moving the non-templated function into the `version (D_ProfileGC)` block: no, that won't work, since prebuilt druntime itself is built without `-profile=gc`. It's only these templates which might get instantiated in client code (**edit**: as soon as we add support for it), and they need the fake-pure wrapper. Currently taken from the prebuilt `core.internal.array.utils` module, which therefore currently depends on the real `accumulate` symbol from the `rt.profilegc` module, dragging this object file into the binary, with its module dtors.

runtime/druntime/src/core/sys/wasi/posix/time.d
kinke: > despite the original C header having it

Ah okay, now I see - I was wondering why you bothered with this factually dead code, thinking it's probably a copy&paste relict.

runtime/druntime/src/rt/profilegc.d
QuantumSegfault: Right...

Well then it seems the best solution would be yeah...move `accumulatePure` into `profilegc.d`, and make the existing `accumulatePure` an alias to `externDFunc!(...)`

kinke: Yep, exactly. Then the prebuilt `core.internal.array.utils` module doesn't depend on `rt.profilegc` anymore, only any template instantiations in user code compiled with `-profile=gc`.

This should bring down the executable sizes in general, not just wasm. [This is most likely a 'regression' from when these templates were added, moving the `-profile=gc` stuff from the glue layer to the frontend, by lowering to these templates.]

QuantumSegfault: Hmmm...

https://github.com/dlang/dmd/pull/22859/

Stuff has already been refactored upstream. Looks like it shouldn't be a problem any more?

Nothing else in `profile_gc.d` is used elsewhere except the hooks, which are only imported when `D_ProfileGC` is active. So the linker should be able to strip `profile_gc`, and thus drop `rt.profilegc` as well?

Or will it still be a problem?

runtime/druntime/src/core/sys/wasi/posix/time.d
QuantumSegfault: dlang/dmd#23407

runtime/druntime/src/rt/profilegc.d
kinke: Mhm, yes, I think your analysis is correct, and so this should be fine in master/v2.114. We should probably still resolve it for v2.113/DMD stable too, mainly because it's a simple fix. I can create an upstream PR too over the weekend, just let me know if you wanna do it yourself.

So for this PR here, I'd say just remove this WASI-specific exclusion of the module. Oh, that's upstream I guess - in which case that would be part of the DMD PR moving the function, and so we'd get it removed with the next `dmd-rewrite-stable` merge anyway. So no blocker for this PR here. **Edit**: Nope, your stuff is in master, not stable... :)

Good that you observed this heavy hidden object-file dependency in druntime, and we went to the bottom of it. :)

QuantumSegfault: Oh, I can do that real quick. I already have it prototyped against LDC. Just a quick shuffle. I'll open a PR shortly.

And then I guess a separate PR to remove the `version (WASI)` once that gets merged.

QuantumSegfault: https://github.com/dlang/dmd/pull/23408

QuantumSegfault:
https://github.com/ldc-developers/ldc/pull/5186/commits/e5150d1a0763f1f9cf8e8e18acc6d4bf87358aa5 is a big new one (since you initially reviewed).

runtime/druntime/src/rt/profilegc.d
QuantumSegfault: Alright. That's all in now. Plus some extra.

QuantumSegfault:
All the prior feedback has been addressed.

And now (along with everything else merged into LDC), I can easily build the runtime & testrunners for WASIp1 & WASIp2 and run the testrunner (in `wasmtime`) and they pass.

runtime/druntime/src/core/math.d
```diff
@@ -343,6 +343,11 @@ else
     real ldexp(real n, int exp);     /* intrinsic */ /// ditto
 }
 
+// Wasm is always cross-compiled, so we are subject to
+// https://wiki.dlang.org/Cross-compiling_with_LDC#Limitations
+// (precision issues with `real`)
```
kinke, line 348: In that regard, it might be interesting to use a Linux AArch64 GitHub Actions runner for CI, with native 128-bit `real` - if the SDK and wasmtime support that host.

kinke (approved):

runtime/druntime/src/core/math.d
QuantumSegfault: They do! You want to do that?

Though I don't think that changes anything here.

------

Interesting that AArch64 Linux supports that but not ARM64 Mac...

PR #5187, opened 2026-07-16
CMake/build tweaks for WASI

QuantumSegfault:
Adds `ldc2.conf` entries for WASI targets to make using them less cumbersome.

Also fixes the `runtime` build so `druntime-ldc` and `druntime-test-runner` can be built for WASI


QuantumSegfault:
I had to make the one WASI test into just a general Wasm test.

I think it would be better to eventually (perhaps once Phobos can be built) how to run whole test-suite with the WASI triples (as a dedicated CI target system). WASI targets now depend on `wasi-sdk` to provide an appropriate Clang as a linker driver (though, you can opt out of it by passing a bunch of flags to counteract the defaults).

thewilsonator (approved):

PR #5188, opened 2026-07-16
druntime: Move `callWithStackShell` implementation to `LDC` block

QuantumSegfault:
`callWithStackShell` contains a `LDC` version block here that's expected to handle all supported architectures. Move the Wasm implementation (which is LDC specific) into there.

See dlang/dmd#23402


QuantumSegfault:
@thewilsonator

thewilsonator (approved):

PR #5191, opened 2026-07-17
CI: Fix update-dmd-rewrite workflow

kinke:
The GHA Ubuntu 22 image apparently has a too recent git; we need git v2.34 to get the correct/existing hashes for the rewritten commits.

So run in a vanilla Ubuntu 22 container instead.


kinke:
Alright - a manual run has worked, updating `dmd-rewrite-master`: https://github.com/ldc-developers/ldc/actions/runs/29604231833/job/87963482169

Just need to check whether the scheduled runs at midnight work too.

PR #5195, opened 2026-07-18
Fix #5192 - enable version(Apple) on iOS, tvOS, watchOS, and visionOS

### LINKED ISSUE #5192: `version(Apple)` is not enabled on iOS
`version(Apple)` is not enabled when building for iOS; which is incorrect according to [the spec](https://dlang.org/spec/version.html#predefined-versions).

Potentially an upstream issue, but from a cursory look it appears that dmd doesn't even to account for the existence of iOS, and yet `version(iOS)` works fine for LDC.

niy-ati:
## Summary
- Enables `version(Apple)` for `iOS`, `TvOS`, and `WatchOS` triples (same as `MacOSX`), matching the [spec](https://dlang.org/spec/version.html#predefined-versions).
- Adds `XROS` handling with `version(VisionOS)` and `version(Apple)`.
- Adds `tests/compilable/gh5192.d`.

Fixes https://github.com/ldc-developers/ldc/issues/5192

## Test plan
- [x] Code review against existing `MacOSX` / `iOS` version setup in `driver/main.cpp`.
- [x] New lit test (requires `target_AArch64`) compiles for ios/tvos/watchos/xros triples with `version(Apple)`.
- [ ] CI

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -897,18 +897,28 @@ void registerPredefinedTargetVersions() {
     break;
   case llvm::Triple::IOS:
     VersionCondition::addPredefinedGlobalIdent("iOS");
+    VersionCondition::addPredefinedGlobalIdent("Apple");
     VersionCondition::addPredefinedGlobalIdent("Posix");
     VersionCondition::addPredefinedGlobalIdent("CppRuntime_LLVM");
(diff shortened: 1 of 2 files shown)


kinke (approved):
Thanks!

PR #5196, opened 2026-07-18
fix issue 5134 windows response files with spaces in paths

### LINKED ISSUE #5134: Compiling programs that use libraries fails on Windows.
As of LDC 1.42, I am no longer able to compile programs that use libraries on Windows. If I try to compile any program that uses a library, I get the following error:

```
Error: unrecognized file extension lib
-vcolumns
Error C:\Program Files\LDC 1.42\bin\ldc2.exe failed with exit code 1.
```

Here is a complete, verbose log: 

[build_log.txt](https://github.com/user-attachments/files/28287448/build_log.txt)

I don't have any other D compilers installed on my system and I am using the dub executable bundled with LDC.

### Reproduction

1. Install LDC 1.42 toolchain on Windows using the multilib installer provided on the releases page.
2. Create a new vibe.d project with `dub init hello -t vibe.d` (or locate any program that uses a library).
3. Try compiling with `dub build`

### Workaround
Downgrade to LDC 1.41.

niy-ati:
fixes the windows bug where dub builds fail if the user profile path has spaces.

dub writes each compiler flag on its own line and quotes paths that contain spaces. llvm response file parsing could stick that quoted path to the next line so a .lib argument became something like lib plus vcolumns. ldc then reported unrecognized file extension lib.

ldc2 now expands response files with the same dmd compatible parser ldmd already uses.

fixes #5134

### DIFF
--- a/tests/driver/gh5134_rsp_spaces.d
+++ b/tests/driver/gh5134_rsp_spaces.d
@@ -1,0 +1,13 @@
+// https://github.com/ldc-developers/ldc/issues/5134
+// Dub writes LF-only response files on Windows; LLVM's Windows tokenizer
+// expects CRLF, which used to glue the next flag onto a quoted .lib path
+// (Error: unrecognized file extension lib).
+//
+// REQUIRES: Windows
+// RUN: mkdir "%t dir with spaces"
+// RUN: echo void main(){} > "%t dir with spaces/t.d"
+// RUN: python %S/inputs/write_lf_rsp.py %t.rsp "\"%t dir with spaces/t.d\"" -c -o- -vcolumns
+// RUN: %ldc @%t.rsp
+// RUN: python %S/inputs/write_lf_rsp.py %t2.rsp "\"%t dir with spaces/dummy.lib\"" -vcolumns
+// RUN: not %ldc @%t2.rsp 2>&1 | FileCheck %s
+// CHECK-NOT: unrecognized file extension
--- a/tests/driver/inputs/write_lf_rsp.py
+++ b/tests/driver/inputs/write_lf_rsp.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+"""Write a response file with LF-only line endings (dub-style on Windows)."""
+import sys
+
+
+def main():
+    if len(sys.argv) < 3:
+        sys.stderr.write("usage: write_lf_rsp.py <path> <line>...\n")
+        return 1
+    path = sys.argv[1]
+    content = "\n".join(sys.argv[2:]) + "\n"
+    with open(path, "wb") as f:
+        f.write(content.encode("utf-8"))
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())


kinke:
It'd be good to check whether the testcase fails without the proposed fix, by extracting that as a first commit and letting CI verify it. **Edit**: And only later, after the CI run (which would get interrupted/cancelled otherwise), pushing the proposed fix as 2nd commit.

niy-ati:
Good point @kinke , I'll split this into two commits:
1. test-only: add gh5134_rsp_spaces.d (Windows regression for #5134)
2. fix: switch ldc2 to the DMD-compatible response expander on Windows (+ link response.cpp)

I'll push the test commit first so CI can show it failing on Windows, then push the fix as a second commit.

kinke:
Yeah so the testcase doesn't fail without the fix. I don't think we need to switch to the DMD response-file parser (fine for ldmd2 as we currently do, but ldc2's response-file behavior is the default LLVM behavior, which is surely more standard than DMD's). Maybe the new dub emits `\n` newlines on Windows, while LLVM might be expecting Windows-standard `\r\n` newlines, or something like that.

niy-ati:
Thanks @kinke ,agreed on both points.

The lit test reproduces a simplified one-arg-per-line @file case, but it doesn’t match what dub actually emits on Windows, so CI passing without the fix isn’t surprising. I’ll rework the test first so it fails on Windows without a fix (likely CRLF line endings and/or a dub-like layout).
On the fix side: I’ll drop the DMD `response_expand` approach for ldc2 and instead look at keeping LLVM’s response file handling, e.g. normalizing newlines before tokenization if that’s where the glueup happens.

niy-ati:
@kinke reworked per your notes:
- dropped the DMD response_expand switch for ldc2
- kept LLVM response-file parsing and normalize bare LF to CRLF on Windows before expansion (matches what dub emits)
- lit test is LF-only / one-arg-per-line like the dub case

all checks are green now, including Windows. happy to adjust if you want something different.

kinke:
I haven't seen a CI run for the first commit, the testcase only, proving that it now really fails on Windows, and so confirming my *guess*. So we first need to identify what the problem really is, via a testcase that actually fails. And then we'd need to check what happened between LDC v1.41 (which apparently worked fine) and LDC v1.42, e.g., whether it is LLVM that changed and needs CRLF line-endings, or whether it's dub that now emits LF instead of CRLF. Once we know these things, we can decide whether to handle it in LDC via the proposed normalization (if this is really an LLVM change/regression), or fix it in dub instead.

niy-ati:
@kinke thanks ,force-pushed so the branch is **test-only** again (`8bd50db`). Waiting on Windows CI to confirm the lit test fails without a fix.

In parallel I'll check LDC 1.41 (LLVM 20) vs 1.42 (LLVM 21) and what dub actually writes for `@` file newlines on Windows, then report back before putting any fix commit back on the branch.

kinke:
Sounds good. Also note that Windows CI here is currently using LLVM 22, so if this was an LLVM 21 regression (and fixed in LLVM 22 again), we wouldn't be able to reproduce it via CI.

niy-ati:
good point , I'll treat CI on LLVM 22 as inconclusive for an LLVM 21-only regression and reproduce on a 1.42 (LLVM 21) Windows build locally.

niy-ati:
@kinke local results on Windows:

Reproduced the issue with a spaced LOCALAPPDATA (dub package cache under a
path with spaces). Exact error:
  Error: unrecognized file extension lib
  -vcolumns

Matrix:
- dub 1.40 (bundled with LDC 1.41) + LDC 1.41 (LLVM 20) → OK
- dub 1.40 + LDC 1.42 (LLVM 21.1.8) → OK
- dub 1.41 (bundled with LDC 1.42) + LDC 1.41 → FAIL
- dub 1.41 + LDC 1.42 ->FAIL

So this is not an LLVM 20->21 regression. Both dub versions write LF-only
response files (`escapeArgs(...).join("\n")`). Converting a failing rsp
LF→CRLF still fails.

The difference is the `-I` line when the path has spaces:
- dub 1.40: "-IC:\...\source"
- dub 1.41: "-IC:\...\source\"

The trailing `\` before the closing quote breaks Windows tokenization, so
later args get glued (hence `lib` + `-vcolumns`). Editing only that line to
drop the trailing backslash (LF kept) makes LDC 1.42 accept the same rsp.

Conclusion: this looks like a dub 1.41 quoting/path issue on Windows, not
something to fix by switching ldc2 to DMD's response parser or by LF->CRLF
normalization in LDC.

kinke:
Okay thanks, interesting findings! So yep, very much looks like a dub regression.

niy-ati:
agreed. I'll open this on dlang/dub (trailing `\` before the
closing quote on quoted `-I` paths in Windows response files)

kinke:
For posterity, this should have fixed it: https://github.com/dlang/dub/pull/3135

PR #5199, opened 2026-07-19
WASI CI

QuantumSegfault:
Adds CI runs to build DRuntime for WASIp1 and WASIp2 and run the `druntime-test-runner`.

### DIFF
--- a/.github/actions/5a-wasi/action.yml
+++ b/.github/actions/5a-wasi/action.yml
@@ -1,0 +1,41 @@
+name: "WASI: Cross-compile & libraries & copy to install dir"
+inputs:
+  os:
+    required: true
+  arch:
+    required: true
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      run: |
+        set -eux
+        cd ..
+
+        os='${{ inputs.os }}'
+        arch='${{ inputs.arch }}'
+        triple="$arch-unknown-$os"
+
+        if [ "$os" = 'wasi' ]; then
+          # legacy `wasm32-wasi` triple
+          toolchain_suffix=''
+        else
+          # newer `wasm32-wasipX` triples
+          toolchain_suffix=-`echo "$os" | cut -c 5-`
+        fi
+
+        # Normally set by 3-build-cross
+        echo "CROSS_TRIPLE=$triple" >> $GITHUB_ENV
+
+        # use bootstrap-ldc, which is guaranteed to be native
+        bootstrap-ldc/bin/ldc-build-runtime \
+          --ninja \
+          --dFlags="-mtriple=$triple" \
+          --ldcSrcDir="$PWD/ldc" \
+          --installWithSuffix="-$os-$arch" \
+          --testrunners \
+          CMAKE_INSTALL_PREFIX="$PWD/install" \
+          CONF_INST_DIR= \
+          PHOBOS2_DIR= \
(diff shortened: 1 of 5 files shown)


.github/actions/5a-wasi/action.yml
```diff
@@ -0,0 +1,41 @@
...
+        bootstrap-ldc/bin/ldc-build-runtime \
+          --ninja \
+          --dFlags="-mtriple=$triple" \
+          --ldcSrcDir="$PWD/ldc" \
+          --installWithSuffix="-$os-$arch" \
+          --testrunners \
+          CMAKE_INSTALL_PREFIX="$PWD/install" \
+          CONF_INST_DIR= \
+          PHOBOS2_DIR= \
+          CMAKE_TOOLCHAIN_FILE="wasi-sdk/share/cmake/wasi-sdk${toolchain_suffix}.cmake" \
```
kinke:
```suggestion
          CMAKE_TOOLCHAIN_FILE="$PWD/wasi-sdk/share/cmake/wasi-sdk${toolchain_suffix}.cmake" \
```

QuantumSegfault:
https://github.com/ldc-developers/ldc/pull/5186#discussion_r3611458913
> In that regard, it might be interesting to use a Linux AArch64 GitHub Actions runner for CI, with native 128-bit real - if the SDK and wasmtime support that host.

Do you want to try that? I guess it might help some Phobos things dealing with `real` be a little more correct???

I don't see there being much difference in the end, as the issue is still present on many common dev setups (i.e. anything other than AArch Linux...and maybe AArch Windows). This just makes anything built/baked into DRuntime or Phobos more precise.

QuantumSegfault:
I'd rather not install yet ANOTHER version of LLVM.

I checked with LLVM 20 and LLVM `master` (now 24) and they are both fine (test runner works)

Could somebody check LLVM 21, and LLVM 22 locally please? Having some artifacts to compare would be useful.

QuantumSegfault:
I'm still investigating but it's a bug present on both LLVM 21 and 22, but not 20, and seems to has been fixed for what's going to be 23.

kinke:
> This just makes anything built/baked into DRuntime or Phobos more precise.

And that'd be the main goal, providing a little archive with prebuilt druntime(+phobos?) for WASI as CI artifact (incl. release artifact), which doesn't suffer from an insufficient compile-time real. As well as not having to disable tests, since that is exactly what we want, showing that the cross-compiled libs on hosts with non-quadruple `real` are problematic and not production-ready.

.github/workflows/main.yml
```diff
@@ -14,8 +14,8 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  LLVM_VERSION: 22.1.8
-  CLANG_MAJOR: 22
+  LLVM_VERSION: 20.1.8
```
kinke, line 17: The used LLVM builds are from our fork, where we don't have a v20.1.8, but v20.1.5 as latest v20 release: https://github.com/ldc-developers/llvm-project/releases/tag/ldc-v20.1.5

So also no need to change the casing of the `os` variable, nor the `llvm-` filename prefix.

QuantumSegfault: Yeah. Once I realized that I stopped trying to abusing the CI and tried those forks locally. So it's not just the CI. It's a consistent issue.

I started `git bisect ` to figure out what fixes it after `22` and see maybe if it's something we can cherrypick in our fork or something.

QuantumSegfault:
Closing in favor of #5207

QuantumSegfault:
XD

I said closing but forgot to actually press Close

PR #5201, opened 2026-07-19
Fix missing D_HardFloat/D_SoftFloat on RISC-V, wasm, MSP430, Xtensa

### LINKED ISSUE #5200: RISC-V, WebAssembly, MSP430, and Xtensa do not predefine D_HardFloat / D_SoftFloat
## Summary

`registerPredefinedTargetVersions()` in `driver/main.cpp` never sets `D_HardFloat` or `D_SoftFloat` for RISC-V, WebAssembly, MSP430, or Xtensa. Other architectures either call `registerPredefinedFloatABI` or set one of those versions explicitly.

Per the [D spec predefined versions](https://dlang.org/spec/version.html#predefined-versions), targets should expose whether hardware floating point is available via `D_HardFloat` / `D_SoftFloat`.

## Location

`driver/main.cpp` , these cases only register the arch ident:

```cpp
case llvm::Triple::msp430:
  VersionCondition::addPredefinedGlobalIdent("MSP430");
  break;
case llvm::Triple::riscv32:
  VersionCondition::addPredefinedGlobalIdent("RISCV32");
  break;
case llvm::Triple::riscv64:
  VersionCondition::addPredefinedGlobalIdent("RISCV64");
  break;
case llvm::Triple::wasm32:
case llvm::Triple::wasm64:
  VersionCondition::addPredefinedGlobalIdent("WebAssembly");
  break;
case llvm::Triple::xtensa:
  VersionCondition::addPredefinedGlobalIdent("Xtensa");
  break;
```

Contrast with LoongArch (same function), which does call `registerPredefinedFloatABI` and therefore also defines `D_HardFloat` / `D_SoftFloat`. NVPTX, SystemZ, and AVR set one of them directly. ARM/PPC/MIPS/SPARC go through `registerPredefinedFloatABI` as well.

`registerPredefinedFloatABI` is documented to define `D_HardFloat` or `D_SoftFloat` from the selected float ABI.

## Impact

Phobos `std.math.hardware.FloatingPointControl` is wrapped in `version (D_HardFloat)`, but already contains full `RISCV_Any` implementations inside that block. On RISC-V those paths never compile in, so the API is unavailable even though the arch-specific code exists.

Same gap for wasm / MSP430 / Xtensa for any `version (D_HardFloat)` / `version (D_SoftFloat)` code.

## How to reproduce

With an LDC build that includes the relevant LLVM targets:

```d
version (D_HardFloat) pragma(msg, "HF");
else version (D_SoftFloat) pragma(msg, "SF");
else pragma(msg, "NEITHER");
void main() {}
```

```
ldc2 -mtriple=riscv64-unknown-linux-gnu -c -o- test.d
ldc2 -mtriple=wasm32-unknown-unknown-wasm -c -o- test.d
ldc2 -mtriple=msp430-unknown-elf -c -o- test.d
ldc2 -mtriple=xtensa-esp32-elf -c -o- test.d
```

**Actual:** `NEITHER`  
**Expected:** `HF` or `SF` as appropriate for the target float ABI (e.g. typical `riscv64-unknown-linux-gnu` -> `HF`; MSP430 -> `SF`; wasm with f32/f64 -> `HF`).

Control:

```
ldc2 -mtriple=loongarch64-unknown-linux-gnu -c -o- test.d
```

-> `HF` (and `tests/driver/loongarch64_abi.d` already asserts `version (D_HardFloat)`).

## Suggested fix

For these arches, set `D_HardFloat` / `D_SoftFloat` the same way other targets do — either call `registerPredefinedFloatABI(...)` where arch-specific soft/hard idents exist, or at least add the `D_*Float` ident from `floatABI` (as NVPTX/SystemZ/AVR do). Add a small lit test similar to `tests/driver/loongarch64_abi.d` for at least `riscv64`.

niy-ati:
## Summary

`registerPredefinedTargetVersions()` did not set `D_HardFloat` / `D_SoftFloat` for RISC-V, WebAssembly, MSP430, or Xtensa.

- MSP430 → `D_SoftFloat` (same as AVR; no hardware FPU)
- WebAssembly → `D_HardFloat` (f32/f64)
- RISC-V / Xtensa → follow `-float-abi` (default hard → `D_HardFloat`)

Adds lit tests for riscv64, wasm32, and msp430.

Fixes #5200

## Test plan

- [ ] CI lit: `tests/driver/riscv64_float_versions.d`, `wasm_float_versions.d`, `msp430_float_versions.d`
- [ ] Manual: `version (D_HardFloat)` / `version (D_SoftFloat)` with the triples from #5200

### DIFF
--- a/tests/driver/msp430_float_versions.d
+++ b/tests/driver/msp430_float_versions.d
@@ -1,0 +1,6 @@
+// REQUIRES: target_MSP430
+
+// RUN: %ldc -c -o- %s -mtriple=msp430-unknown-elf
+
+version (D_SoftFloat) {} else static assert(0);
+version (D_HardFloat) static assert(0);
--- a/tests/driver/msp430_float_versions.d
+++ b/tests/driver/msp430_float_versions.d
@@ -1,5 +1,3 @@
-// REQUIRES: target_MSP430
-
 // RUN: %ldc -c -o- %s -mtriple=msp430-unknown-elf
 
 version (D_SoftFloat) {} else static assert(0);
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -706,12 +706,18 @@ void registerPredefinedTargetVersions() {
     break;
   case llvm::Triple::msp430:
     VersionCondition::addPredefinedGlobalIdent("MSP430");
+    // No hardware FPU; match AVR.
+    VersionCondition::addPredefinedGlobalIdent("D_SoftFloat");
     break;
   case llvm::Triple::riscv32:
     VersionCondition::addPredefinedGlobalIdent("RISCV32");
+    VersionCondition::addPredefinedGlobalIdent(
+        floatABI == FloatABI::Soft ? "D_SoftFloat" : "D_HardFloat");
     break;
   case llvm::Triple::riscv64:
     VersionCondition::addPredefinedGlobalIdent("RISCV64");
+    VersionCondition::addPredefinedGlobalIdent(
+        floatABI == FloatABI::Soft ? "D_SoftFloat" : "D_HardFloat");
     break;
   case llvm::Triple::sparc:
     // FIXME: Detect SPARC v8+ (SPARC_V8Plus).
@@ -739,6 +745,7 @@ void registerPredefinedTargetVersions() {
   case llvm::Triple::wasm32:
   case llvm::Triple::wasm64:
     VersionCondition::addPredefinedGlobalIdent("WebAssembly");
+    VersionCondition::addPredefinedGlobalIdent("D_HardFloat");
     break;
   case llvm::Triple::loongarch32:
     VersionCondition::addPredefinedGlobalIdent("LoongArch32");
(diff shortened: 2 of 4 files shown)


tests/driver/msp430_float_versions.d
```diff
@@ -0,0 +1,6 @@
+// REQUIRES: target_MSP430
```
kinke: Without actual codegen (`-o-`), the LLVM backend doesn't need to be enabled. So the `REQUIRES` in these test files aren't required.

kinke:
Thanks, makes sense to me. I guess defaulting to hard-float for Xtensa makes sense, but I have no idea about that arch.

tests/driver/msp430_float_versions.d
niy-ati: @kinke thanks , dropped the `REQUIRES` lines from the three lit tests since they only use `-c -o-`. Left Xtensa on the hard-float default via `-float-abi` as discussed.

thewilsonator:
Xtensa typically supports Hard float.

PR #5202, opened 2026-07-19
Pass `LLVM_MAJOR_VERSION` through `ldc-build-runtime`

QuantumSegfault:
Fixing a subtle issue with ldc-developers/ldc#5187 with it using `LLVM_VERSION_MAJOR` but is not being set (treated as 0 in comparisons?) and acting like an old version.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -24,6 +24,9 @@ if(NOT LDC_EXE)
     if(NOT DEFINED DMDFE_MINOR_VERSION OR NOT DEFINED DMDFE_PATCH_VERSION)
         message(FATAL_ERROR "Please define the CMake variables DMDFE_MINOR_VERSION and DMDFE_PATCH_VERSION.")
     endif()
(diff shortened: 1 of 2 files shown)


QuantumSegfault:
@thewilsonator

thewilsonator (approved):

kinke:
Hmm, actually this breaks manual runtime-only CMake builds; the var is only required when targeting WASI. Might be a problem for @the-horo and other package maintainers.

QuantumSegfault:
Do you want me to open another PR to only require `LLVM_VERSION_MAJOR` for WASI?

JohanEngelen:
You could change this PR to error only when targeting WASI. That preserves existing behavior, and fixes WASI build too.

QuantumSegfault:
That's what I'm asking.

Can't change THIS PR because it's already merged, but I can open another to tweak it, if desired.

JohanEngelen:
> That's what I'm asking.
> 
> Can't change THIS PR because it's already merged, but I can open another to tweak it, if desired.

Please do, thanks.

PR #5204, opened 2026-07-19
Only require `LLVM_VERSION_MAJOR` for `runtime` CMake for WASI targets.

QuantumSegfault:
See https://github.com/ldc-developers/ldc/pull/5202#issuecomment-5015308446


QuantumSegfault:
@kinke 
@JohanEngelen 
@thewilsonator

kinke:
Thanks!

PR #5206, opened 2026-07-20
Reject GCC asm memory operands on register variables (ImportC)

### LINKED ISSUE #4967: importC: GCC-style inline asm accepts taking the address of a register variable
```c
void f()
{
    register int ax;
    asm("" : "=m" (ax));
    asm("" :: "m" (ax));
}
```
This is illegal C code, so should be rejected at codegen time  (can't be handled by the frontend parser or semantic, as it doesn't know about backend generic/target asm constraints).

niy-ati:
## Summary

ImportC keeps `register` as `STCregister`, but GCC style inline asm with `"m"` or `"=m"` constraints takes the operand address during LDC codegen. That is illegal C (same as GDC rejecting it) and was not caught in semantic analysis because the frontend does not interpret asm constraint strings.

This rejects indirect memory asm operands that refer to a register variable in `GccAsmStatement_toIR`, using the same diagnostic as other ImportC address of register cases.

Fixes ldc-developers/ldc#4967.

## Test plan

Added `tests/dmd/fail_compilation/issue4967.c` covering output and input `"m"` operands on a `register int`.


gen/asm-gcc.cpp
```diff
@@ -207,6 +235,18 @@ void GccAsmStatement_toIR(GccAsmStatement *stmt, IRState *irs) {
...
+    bool registerMemError = false;
+    for (size_t i = 0; i < stmt->args->length; ++i) {
+      if (constraintsBuilder.isIndirectOperand(i) &&
+          checkRegisterMemoryAsmOperand((*stmt->args)[i]->loc,
+                                        (*stmt->args)[i]))
+        registerMemError = true;
+    }
+    if (registerMemError)
+      return;
+
```
kinke: This can be more easily integrated in the loop below, where we already check indirect operands for lvalue-ness (and where we shouldn't call `fatal()` either), inlining `checkRegisterMemoryAsmOperand()` (no need for an extra function for that).

tests/dmd/fail_compilation/issue4967.c
kinke, line 1: Other LDC-specific file additions to the *DMD* testsuite are named `ldc_github_4967.c`, to make clear that these are LDC-specific (incl. issue numbers).

kinke (changes requested):

kinke:
The other helper was fine, I just meant removing one of the two. And the check can be integrated more nicely, no need for `hasErrors` etc. I'm probably quicker doing the changes myself, but this fix is of very low priority.

PR #5207, opened 2026-07-20
WASI CI [2]

kinke:
Based on #5199.

### DIFF
--- a/.github/actions/5a-wasi/action.yml
+++ b/.github/actions/5a-wasi/action.yml
@@ -24,18 +24,36 @@ runs:
           toolchain_suffix=-`echo "$os" | cut -c 5-`
         fi
 
-        # Normally set by 3-build-cross
-        echo "CROSS_TRIPLE=$triple" >> $GITHUB_ENV
+        absInstallDir="$PWD/install"
 
-        # use bootstrap-ldc, which is guaranteed to be native
+        # cross-compile druntime (incl. test runners) and install
         bootstrap-ldc/bin/ldc-build-runtime \
           --ninja \
           --dFlags="-mtriple=$triple" \
           --ldcSrcDir="$PWD/ldc" \
           --installWithSuffix="-$os-$arch" \
           --testrunners \
-          CMAKE_INSTALL_PREFIX="$PWD/install" \
-          CONF_INST_DIR= \
+          CMAKE_INSTALL_PREFIX="$absInstallDir" \
+          RT_CONF_TRIPLE_REGEX="$arch-.*-$os" \
           PHOBOS2_DIR= \
           CMAKE_TOOLCHAIN_FILE="$PWD/wasi-sdk/share/cmake/wasi-sdk${toolchain_suffix}.cmake" \
           BUILD_LTO_LIBS=ON
+
+        # make the installed .conf portable
+        perl -pi -e "s|$absInstallDir/|%%ldcbinarypath%%/../|g" install/etc/ldc2.conf/*.conf
+
+        # pack install/ dir to a .tar.xz archive
+        artifactsDir='${{ github.workspace }}/artifacts'
+        mkdir $artifactsDir
+        chmod -R go=rX install
+        sudo chown -R 0:0 install
+        cd install
+        tar -cf - * | 7z a $artifactsDir/ldc2-addon-$os-$arch.tar.xz -si -txz -mx9
+        cd ..
+
+    - name: Upload artifact
+      uses: actions/upload-artifact@v6
+      with:
+        name: addon-${{ inputs.os }}-${{ inputs.arch }}
+        path: artifacts/
+        compression-level: 0
--- a/.github/actions/5a-wasi/action.yml
+++ b/.github/actions/5a-wasi/action.yml
@@ -1,4 +1,4 @@
-name: "WASI: Cross-compile & libraries & copy to install dir"
+name: "WASI: Cross-compile druntime incl. test runners, install libs + .conf, run druntime unittests"
 inputs:
   os:
     required: true
@@ -7,7 +7,8 @@ inputs:
 runs:
   using: composite
   steps:
-    - shell: bash
+    - name: Cross-compile druntime incl. test runners & install libs + .conf
+      shell: bash
       run: |
         set -eux
         cd ..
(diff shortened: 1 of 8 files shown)


kinke:
@QuantumSegfault: Right, the druntime test runners are green with LLVM 20.

.github/actions/5a-wasi/action.yml
```diff
@@ -0,0 +1,59 @@
...
+
+        # cross-compile druntime (incl. test runners) and install
+        bootstrap-ldc/bin/ldc-build-runtime \
+          --ninja \
+          --dFlags="-mtriple=$triple" \
+          --ldcSrcDir="$PWD/ldc" \
+          --installWithSuffix="-$os-$arch" \
+          --testrunners \
+          CMAKE_INSTALL_PREFIX="$absInstallDir" \
+          RT_CONF_TRIPLE_REGEX="$arch-.*-$os" \
```
QuantumSegfault, line 36: Make sure to make sure to match string end `^`, so it doesn't confuse `wasm32-unknown-wasip1` for `wasm32-unkown-wasi`

QuantumSegfault:
Thank you for all of this!

.github/actions/5a-wasi/action.yml
kinke: `os` is either `wasip1` or `wasip2` at the moment, so they are distinct. - IIRC, the initial wasm triples started out with a `wasm` environment, something like `wasm32-unknown-unknown-wasm`, so I wanted to avoid *requiring* no environment.

kinke:
Well, thank *you* for pushing the wasm support and initiating CI!

.github/actions/5a-wasi/action.yml
QuantumSegfault: I'm saying do `$arch-.*-$os^`, so the config doesn't try applying settings for an installed `wasip1` target to `wasi` (which is a distinct triple), and vice versa. But I see what you are getting at...maybe we could match an optional environment??? `wasm32-.*-wasi(-.*)?^`, etc.?


In fact, I just realized that this CI can't be working as we intend! We are using old LLVM 20, so the new triples didn't exist.

Since it's not breaking down entirely, it's probably treating any `-wasi` as a WASI triple. Yeah...see https://github.com/llvm/llvm-project/pull/165345/changes#diff-2e0e305aa6a2fae00ca5531201ebeee1369614ad0578835d49cc9a05cf6ba961L741; `startsWith`

So we are running CI twice for no good reason. One way or another we should drop the `wasip2` job for the moment. Though we can still use the `wasip1` triple if you want (or strip it back to `wasi` as was canonical).

kinke: Yeah I don't wanna ship the WASI libs built with LLVM 20 anyway, when all official prebuilt compilers are using LLVM 22. So I hope you find a solution for the LLVM 21-22 problem.

[And I'll probably merge the WASI libs to a single CI job, since the actual cross-compilation and testrunners-run takes a few seconds only, while getting the prerequisites and compiling the compiler takes minutes.]

> so the config doesn't try applying settings for an installed wasip1 target to wasi

A normalized effective `…-wasi` triple wouldn't match the current regexes with their WASI-preview suffix.

QuantumSegfault: Another thought...is there a way you could reuse the "bootstrap" compiler from the appropriate native CI jobs? Or is that not worth the effort and interdependence.

> A normalized effective …-wasi triple wouldn't match the current regexes with their WASI-preview suffix.

Good point.

So it's only the opposite that's a problem.

kinke: >  Or is that not worth the effort and interdependence.

Yeah I don't think it's worth it (currently not doable because of the dependency on LLVM 20). If we took it from the native job, we'd have to wait for a long time until the WASI jobs are finally triggered (until the main job is done). And if the main job failed, we wouldn't have any WASI CI results at all etc.

If we had more build-cross-libs-only jobs (not just WASI), it might pay off to split out a separate build-native-compiler-job only, and use that/its artifact for multiple jobs, so maybe something for the future.

> So it's only the opposite that's a problem.

Yep, so no problem for now. But it'd probably be good to allow the user to specify a `…-wasi` triple without preview suffix, and using the proper libs then automatically. Do you know if the preview level is fluid, i.e., might `…-wasi` at one point mean `-wasip[2-3]`, not stick to `p1`?

QuantumSegfault: #5208 so we can do 22.

------

`-wasi` is currently a legacy alias for WASIp1. It's been deprecated, and it's set to be removed from `wasi-sdk` shortly (no more `-wasi` libc builds; only `-wasip1`). It isn't going to be bumped to mean a different preview.

`-wasi` is being freed up for potential use for the full release of WASI 1.0 in the future.

CMakeLists.txt
```diff
@@ -928,74 +925,51 @@ makeConfSection(NAME "30-compiler" SECTION "default"
...
 )
 
-set(wasi_switches)
-# no Phobos, so can't builds libs just yet
-# list(APPEND wasi_switches -defaultlib=druntime-ldc,phobos2-ldc)
-
-# We don't want to export any bloat. Assume that the entrypoint (i.e. main)
-# is the only export needed in most cases. Additional exports can be marked
-# with an LLVM attribute if needed (normal D `export` won't suffice).
-list (APPEND wasi_switches -L--no-export-dynamic)
```
QuantumSegfault, line 956: What about this one?

QuantumSegfault (approved):
LGTM, other than the `-L--no-export-dynamic` being lost.

I don't know if you want to merge it now, or wait until after #5178 (so GC works as intended).

CMakeLists.txt
kinke: I've added a TODO comment there; we could drop it from wasm in general and require users to add it manually if needed. Or handle `export` specially for wasm, if it really requires adding those symbols to `llvm.used`.

kinke:
Okay, I've added ~~4~~ 5 more commits since your review. IMO ready now; we can tackle the `--[no-]export-dynamic` later.

QuantumSegfault (approved):
Okay, we'll worry about that later.

LGTM.

.github/actions/7-package/action.yml
```diff
@@ -10,13 +10,13 @@ inputs:
...
   is_addon:
     required: false
-    default: false
+    default: ''
 runs:
   using: composite
   steps:
 
     - name: Set DMD environment variable
-      if: '!inputs.is_addon'
```
QuantumSegfault, line 19: To use ! in `if` you have to wrap it as `${{ !inputs.is_addon }}`

Here you just have that in quotes...as a string...which will always evaluate true.

kinke: [I don't think that's the case, as (all) other expressions don't need those curly brackets in `if:` conditions, incl. usages of builtin functions like `startsWith`. I rather think the problem was that the vars aren't retained as bools, but passed as strings internally or so, making `false` a `"false"`, which is truthy.]

**Edit**: And the quotes are YAML-quotes only, IIRC required if the string starts with `!`.

QuantumSegfault: I was just investigating this exact thing the other day. 

> When you use expressions in an if conditional, you can, optionally, omit the `${{ }}` expression syntax because GitHub Actions automatically evaluates the if conditional as an expression. However, this exception does not apply everywhere.
>
> You must always use the `${{ }}` expression syntax or escape with `''`, `""`, or `()` when the expression starts with `!`, since `!` is reserved notation in YAML format.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idif

-------

But apparantly quotes work...weirdly enough. So that still makes my comment moot.

kinke: Ah: https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax?versionId=free-pro-team%40latest&productId=actions&restPage=reference%2Cworkflows-and-actions%2Ccontexts#inputsinput_iddefault

> A `string` representing the default value.

kinke:
The new WASI addon package (to be unpacked in an existing non-system LDC install prefix) is now e.g. available from https://github.com/ldc-developers/ldc/releases/tag/CI.

JohanEngelen:
Wowwww, major milestone!!!

PR #5208, opened 2026-07-20
Fix WASI builds on LLVM 21 & 22 (EH)

QuantumSegfault:
Works around an LLVM bug where EH would silently not properly enable.

Somehow introduced for LLVM 21, present for LLVM 22, but fixed for (what will be) 23 by llvm/llvm-project#177542


QuantumSegfault:
@kinke 

I figured it out. `git bisect` is nice. Rebuilding LLVM 20 times is not. -_-

But good news is we can get LLVM 22 to work

kinke (approved):

kinke:
Nice that you found it, thx!

PR #5211, opened 2026-07-22
Don't apply `pragma(inline, true)` to unittest functions

### LINKED ISSUE #5210: Don't apply module-level `pragma(inline, true)` to unittest functions
Wrt. https://github.com/ldc-developers/ldc/blob/47b886bd0220a5792a61e23d8ca5597ac83ef6dd/.github/actions/5a-wasi/action.yml#L49-L53

There's an (LDC-specific) `pragma(inline, true):` at the top of the `core.int128` module, since all of the normal functions there are supposed to be intrinsics-like. Unfortunately, the pragma then applies to unittest functions too, and inlining the unittest functions into the module-unittest function (which just calls all effective unittest functions in that module) makes no sense at all, just causes that wrapper to explode in size (especially without enabled optimizations).

kinke:
Resolves #5210.

### DIFF
--- a/.github/actions/5a-wasi/action.yml
+++ b/.github/actions/5a-wasi/action.yml
@@ -44,11 +44,6 @@ runs:
         set -eux
         cd ../ldc-build-runtime.${{ inputs.os }}-${{ inputs.arch }}
 
+        export WASMTIME_BACKTRACE_DETAILS=1
         ../wasmtime/wasmtime -W exceptions ./druntime-test-runner
-
-        # the debug runner hits:
-        # Error: failed to compile: wasm[0]::function[676]::_D4core6int12810__unittestZ
-        # Caused by:
-        #     0: WebAssembly translation error
-        #     1: Invalid input WebAssembly code at offset 585213: too many locals: locals exceed maximum
-        #../wasmtime/wasmtime -W exceptions ./druntime-test-runner-debug
+        ../wasmtime/wasmtime -W exceptions ./druntime-test-runner-debug
--- a/.github/actions/5a-wasi/action.yml
+++ b/.github/actions/5a-wasi/action.yml
@@ -46,4 +46,5 @@ runs:
 
         export WASMTIME_BACKTRACE_DETAILS=1
         ../wasmtime/wasmtime -W exceptions ./druntime-test-runner
-        ../wasmtime/wasmtime -W exceptions ./druntime-test-runner-debug
+        # FIXME: hangs at one point after core.internal.container.treap
+        #../wasmtime/wasmtime -W exceptions ./druntime-test-runner-debug
--- a/gen/functions.cpp
+++ b/gen/functions.cpp
@@ -654,9 +654,10 @@ void DtoDeclareFunction(FuncDeclaration *fdecl, const bool willDefine) {
     irFunc->setNeverInline();
   } else {
     if (fdecl->inlining == PINLINE::always) {
-      // If the function contains DMD-style inline assembly.
-      if (fdecl->hasInlineAsm()) {
-        // The presence of DMD-style inline assembly in a function causes that
+      if (fdecl->isUnitTestDeclaration()) {
+        // makes no sense: https://github.com/ldc-developers/ldc/issues/5210
+      } else if (fdecl->hasInlineAsm()) {
+        // The presence of *DMD-style* inline assembly in a function causes that
         // function to become never-inline. So, if this function contains DMD-style
         // inline assembly we'll emit an error as it can't be made always-inline.
         // However, we'll make an exception for C functions, as the C standard doesn't
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -93,7 +93,7 @@ if("${TARGET_SYSTEM}" MATCHES "UNIX|WASI")
     ENABLE_LANGUAGE(ASM)
 endif()
 
-# Enable switches needed for EH on Wasm/WASI
+# Enable switches needed on Wasm/WASI (normally inherited from etc/ldc2.conf/{51-target-wasm-base,52-target-wasi-base}.conf)
 if("${TARGET_SYSTEM}" MATCHES "WASI")
     # Make sure LLVM_VERSION_MAJOR is defined
     # (particularly if this CMake project is NOT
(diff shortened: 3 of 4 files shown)


kinke:
Hmm, so wrt. druntime *debug* unittests for WASI, the testrunner now doesn't run into the too-many-locals error for `core.int128` anymore - but a unittest in there apparently causes a segfault, https://github.com/ldc-developers/ldc/actions/runs/29936986473/job/88981081342?pr=5211#step:6:729:
```
+ ../wasmtime/wasmtime -W exceptions ./druntime-test-runner-debug
0.000s PASS debug32 core.atomic
Error: failed to run main module `./druntime-test-runner-debug`

Caused by:
    0: failed to invoke command default
    1: error while executing at wasm backtrace:
    0:  0x5813e - druntime-test-runner-debug!_D4core6int12819__unittest_L1065_C1FNbNiNfZv
    1:  0xc45fb - druntime-test-runner-debug!_D4core6int12810__unittestZ
    2:   0x3808 - druntime-test-runner-debug!_D11test_runner6doTestFPS6object10ModuleInfoKS4core7runtime14UnitTestResultZv
    3:   0x3a4b - druntime-test-runner-debug!_D11test_runner7testAllFZ20__foreachbody_L63_C5MFPS6object10ModuleInfoZi
    4: 0x25a144 - druntime-test-runner-debug!_D6object10ModuleInfo7opApplyFMDFPSQBhQBdZiZ18__lambda_L2575_C13MFyPSQCpQClZi
    5: 0x2979ad - druntime-test-runner-debug!_D2rt5minfo17moduleinfos_applyFMDFyPS6object10ModuleInfoZiZ21__foreachbody_L585_C5MFKSQDg13sections_wasm12SectionGroupZi
    6: 0x29bc0f - druntime-test-runner-debug!_D2rt13sections_wasm12SectionGroup7opApplyFMDFKSQBuQBuQBiZiZi
    7: 0x2978d7 - druntime-test-runner-debug!_D2rt5minfo17moduleinfos_applyFMDFyPS6object10ModuleInfoZiZi
    8: 0x24e601 - druntime-test-runner-debug!_D6object10ModuleInfo7opApplyFMDFPSQBhQBdZiZi
    9:   0x372e - druntime-test-runner-debug!_D11test_runner7testAllFZS4core7runtime14UnitTestResult
   10:   0x345b - druntime-test-runner-debug!_D11test_runner6testerFZS4core7runtime14UnitTestResult
   11: 0x1c0005 - druntime-test-runner-debug!runModuleUnitTests
   12: 0x2831b2 - druntime-test-runner-debug!_D2rt6dmain212_d_run_main2UAAakPUQgZiZ6runAllMFZv
   13: 0x2833d9 - druntime-test-runner-debug!_D2rt6dmain212_d_run_main2UAAakPUQgZiZ7tryExecMFMDFZvZv
   14: 0x283022 - druntime-test-runner-debug!_d_run_main2
   15: 0x282c72 - druntime-test-runner-debug!_d_run_main
   16:   0x40fe - druntime-test-runner-debug!main
   17: 0x2b17bc - druntime-test-runner-debug!__main_void
   18:   0x2f5b - druntime-test-runner-debug!_start
note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information
    2: memory fault at wasm address 0xffff1e68 in linear memory of size 0x550000
    3: wasm trap: out of bounds memory access
```

FYI @QuantumSegfault.

QuantumSegfault:
My best guess considering it doesn't trigger on release: insufficient stack space (stack grows down and is at beginning of address space; might have underflowed and wrapped).

kinke:
Yeah thanks, I was suspecting that as well (as you know, we link the test runners via CMake as a C project, so don't get the wasm/wasi defaults from the LDC compiler, incl. larger stack; so annoyingly we need to duplicate the WASI switches in the runtime CMakeLists.txt as well as the default wasm/wasi .conf files in the parent CMakeLists.txt).

kinke:
Yep, progress - the first modules pass now with 1MB stack, incl. `core.int128`, but then it seems to hang: https://github.com/ldc-developers/ldc/actions/runs/29940325622/job/88992307955?pr=5211#step:6:754

QuantumSegfault:
Based on the fact that the rest of the tests are in alphabetical order (coincidentally), that would point to one of the GC tests not being happy?

kinke:
The next one in the list  would be `core.internal.convert` (see https://github.com/ldc-developers/ldc/actions/runs/29942111348/job/88998394488#step:6:755), but I don't trust it to be the immediate next one (buffered output?) - e.g., last time when we hit the `core.int128` stack overflow, it only managed to print `core.atomic` as first successful test, but there are 7 modules inbetween that first module and `core.int128`.

I commented-out the debug runner again to unblock this PR, as the hang probably requires *local* investigation.

QuantumSegfault:
I can start looking into it shortly.

QuantumSegfault:
I was right, but for the wrong reason.

`ATTEMPT debug32 core.internal.gc.blkcache`

QuantumSegfault:
`GC.collect` is stalling.

QuantumSegfault:
Riddle me this: what happens when you try to allocate GC memory during a collection? -_-

It's stalling on this falsey assert

```
extern (C) void thread_scanAllType(scope ScanAllThreadsTypeFn scan) nothrow
in
{
    assert(suspendDepth > 0);
}
do
{
    callWithStackShell(sp => scanAllTypeImpl(scan, sp));
}
```

I never call `thread_suspendAll` on WASI. I guess I need to fix that (do a dummy implementation instead of skipping).

kinke:
Okay, thx for investigating! That's exactly why I wanted to get the debug tests to work, as they provide more value than the optimized release ones, where e.g. in this case the GC2Stack pass probably got rid of some allocations.

So let's get this PR in, so that you can get working on a fix and confirm via CI.

---

Btw, unrelated, wrt. `--export-dynamic`: I think it's safe to drop it as default switch for all wasm targets, as we don't default to it for normal Posix builds either (unlike DMD). [And AFAIK, the linker defaults to it for shared libs, but not for executables.]

While converting a testcase to clang, I've seen that clang apparently defaults to `-fvisibility=hidden` for wasm: https://godbolt.org/z/s6dv11jGG. We could do the same, so that only explicit `export`ed symbols get public visibility, and are so exported via an opt-in `-L--export-dynamic`. Or at the very least use that for prebuilt druntime (+ Phobos in the future), so that there's no bloat from static druntime/Phobos at least.

.github/actions/5a-wasi/action.yml
```diff
@@ -44,11 +44,7 @@ runs:
         set -eux
         cd ../ldc-build-runtime.${{ inputs.os }}-${{ inputs.arch }}
 
+        export WASMTIME_BACKTRACE_DETAILS=1
```
QuantumSegfault, line 47: Do we want to leave this in?

QuantumSegfault (approved):

.github/actions/5a-wasi/action.yml
kinke: Yep, it shows file/line numbers (for the debug runner at least), which we didn't have before. For the stack overflow, it pointed to the first line in the function body, which was a good hint.

QuantumSegfault: OH!!! I've been trying to figure out how to get debug info to show up for Wasmtime. I couldn't remember how I did it that once.

Thought that would be default...

Okay, yeah, that makes sense. Keep it

QuantumSegfault:
I have the fix ready (there's no reason to wait for upstream DMD since the implementation needs to be slightly different anyway due to refactoring).

I'll push once this is merged.

PR #5212, opened 2026-07-22
druntime: Allow thread resume/suspend to work on WASI

QuantumSegfault:
Rather than erroring out always, allows "suspending" and "resuming" the main thread (which on POSIX, and now WASI, just marks the stack top).

Prevents issues with code expecting `thread_suspendAll` to be called before hand.


QuantumSegfault:
@kinke

kinke:
LGTM. Please enable the debug runner too now.

kinke (approved):

PR #5214, opened 2026-07-22
GHA: Add a few missing step names

kinke:
As these step names of composite actions are now finally shown in the CI logs (as collapsible sections), incl. nesting - I thought I'd never see that happen. :)

So the names are now more than just documenting comments, and a few were missing. Very bad are anonymous multiline run steps, as the displayed section name is the first line then (`Run set -eux`).

### DIFF
--- a/.github/actions/4a-test-ldc2/action.yml
+++ b/.github/actions/4a-test-ldc2/action.yml
@@ -2,5 +2,6 @@ name: Run LDC D unittests
 runs:
   using: composite
   steps:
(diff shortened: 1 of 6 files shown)


kinke:
E.g., this is how it currently looks like (note the multiple ugly `Run set -eux` sections), from https://github.com/ldc-developers/ldc/actions/runs/29955048628/job/89041864009:

image

thewilsonator (approved):

PR #5215, opened 2026-07-22
Wasm EH: Use `invoke` for `wasm.rethrow` when needed

QuantumSegfault:
Ensures that cleanup pads (e.g. `scope(exit)`) are ran properly when an exception falls through a try-catch statement.


QuantumSegfault:
@kinke

PR #5217, opened 2026-07-23
Fix some missing IR-symbol visibilities

kinke:
Mainly for non-class TypeInfos, by consistently using the `setLinkageAndVisibility()` helper.


kinke:
@QuantumSegfault: This should do it.

QuantumSegfault:
Works as expected. Now we're just down to the ModuleInfo.

Now the difference between --export-dynamic and not is only +3.4 K. Seems reasonable.

QuantumSegfault (approved):

thewilsonator (approved):

PR #5219, opened 2026-07-24
WASI CI: Smoke-test LTO too

kinke:


kinke:
The hello-world executable size reduction is only moderate:
* WASIp1: 536,658 => 489,731
* WASIp2: 654,573 => 607,212

PR #5221, opened 2026-07-24
Merge upstream stable

kinke:


kinke:
I didn't realize the `accumulatePure()` moval was already cherry-picked, the main motivation for this PR at this time. The `std.regex` fix is a bit meager for a stable-merge, so I might want to replace this PR with a bigger one later. Keeping it open, just in case upstream stable doesn't see any more action until we're getting ready for v1.43.0-beta1.

QuantumSegfault:
Yeah, I did it ASAP after it was accepted upstream.

https://github.com/ldc-developers/ldc/pull/5186/changes/7fa8bcf913da3f1f8a1d664175933682799dcdcf

------

How long until `beta` anyway?

kinke:
> How long until beta anyway?

IMO, we're pretty much there; let's get your wasm-spill pass in [if you have any executable-size diffs, e.g., for the druntime test runner, that'd be cool, just to get an idea of the effect] and add a changelog entry wrt. your WASI support (and brief note on how to use the addon artifact). So maybe even beta1 this weekend...

QuantumSegfault:
Wow. Okay, so Phobos will have to wait then. That's fine. I need/want to iron dlang/dmd#23398 out first, so that's going to be awhile; no sense in holding things up.

I'll see if I can get some numbers for that.

kinke:
Yeah it's just to get some early feedback; I'm e.g. interested in some feedback wrt. `-foptimize-nothrow`. We release from master (normally, but can always change to a release branch if a big breaking change makes it into master in the v1.43 release cycle); WASI is highly experimental anyway, plus if you're working on druntime support for WASIp2 and Phobos, that's just additions. And we can release as many betas as we want, a release is basically just a push of a tag.

QuantumSegfault:
DRuntime on WASIp2 already works / is in, but it should be a non-breaking change to move away from POSIX emulation internally in the couple time-related places. But yeah it's mostly additive otherwise.

I though beta would mean feature freeze.

------

https://github.com/ldc-developers/ldc/pull/5178#issuecomment-5073982672

PR #5222, opened 2026-07-24
Add support for LLVM 24 to intrinsics.di

gnavdev28:


QuantumSegfault:
Thanks!

I noticed this, but just dealt with it locally. Wasn't sure if we were going to start supporting LLVM 24 until after the next LDC release or not.

PR #5224, opened 2026-07-24
Add new `-fno-moduleinfo-localclasses` switch

QuantumSegfault:
Adds new `-fno-minfo-localclasses` switch to disable population of `localClasses` in ModuleInfo

Keeping `TypeInfo_Class` references around forces keeping the otherwise unused data (and because of vtbl) code around.

Doing so is a breaking change, breaking `Object.factory` (and user code making use of `localClasses`). And despite the note in the docs about it being possibly removed in the future, this has yet to happen. So for now, the switch is strictly opt-out (though we enable it for WASI CI).

### DIFF
--- a/.github/actions/5a-wasi/action.yml
+++ b/.github/actions/5a-wasi/action.yml
@@ -28,6 +28,7 @@ runs:
         bootstrap-ldc/bin/ldc-build-runtime \
           --ninja \
           --dFlags="-mtriple=$triple" \
+          --dFlags="-fno-minfo-localclasses" \
           --ldcSrcDir="$PWD/ldc" \
           --buildDir="ldc-build-runtime.$os-$arch" \
           --installWithSuffix="-$os-$arch" \
--- a/.github/actions/5a-wasi/action.yml
+++ b/.github/actions/5a-wasi/action.yml
@@ -27,7 +27,7 @@ runs:
 
         bootstrap-ldc/bin/ldc-build-runtime \
           --ninja \
-          --dFlags="-mtriple=$triple;-fno-minfo-localclasses" \
+          --dFlags="-mtriple=$triple;-fno-moduleinfo-localclasses" \
           --ldcSrcDir="$PWD/ldc" \
           --buildDir="ldc-build-runtime.$os-$arch" \
           --installWithSuffix="-$os-$arch" \
--- a/.github/actions/5a-wasi/action.yml
+++ b/.github/actions/5a-wasi/action.yml
@@ -27,8 +27,7 @@ runs:
 
         bootstrap-ldc/bin/ldc-build-runtime \
           --ninja \
-          --dFlags="-mtriple=$triple" \
-          --dFlags="-fno-minfo-localclasses" \
+          --dFlags="-mtriple=$triple;-fno-minfo-localclasses"
           --ldcSrcDir="$PWD/ldc" \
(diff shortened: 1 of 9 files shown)


QuantumSegfault:
@kinke

.github/actions/5a-wasi/action.yml
```diff
@@ -28,6 +28,7 @@ runs:
         bootstrap-ldc/bin/ldc-build-runtime \
           --ninja \
           --dFlags="-mtriple=$triple" \
+          --dFlags="-fno-minfo-localclasses" \
```
kinke: You are overriding the dFlags this way (IIRC). Separate them via semicolons instead (see `--help` docs).

dmd/globals.d
```diff
@@ -353,6 +353,8 @@ version (IN_LLVM)
     // Windows-specific:
     bool dllexport;      // dllexport ~all defined symbols?
     DLLImport dllimport; // dllimport data symbols not defined in any root module?
+
+    bool useMInfoLocalClasses = true;
```
kinke: The frontend doesn't need to know about this flag, it's just for our glue code.

gen/moduleinfo.cpp
```diff
@@ -185,24 +185,26 @@ llvm::Constant *buildLocalClasses(Module *m, size_t &count) {
...
-                             cd->toPrettyChars());
-      continue;
-    }
-
-    if (cd->sizeok != Sizeok::done) {
-      IF_LOG Logger::println(
-          "skipping opaque class declaration '%s' in moduleinfo",
-          cd->toPrettyChars());
-      continue;
+  if (global.params.useMInfoLocalClasses) {
```
kinke: And here you can use `opts::fNoMInfoLocalClasses` directly instead.

dmd/globals.d
QuantumSegfault: What should I do with `fNoMInfoLocalClasses` then? Use it directly? Or is there a better place to put this?

gen/moduleinfo.cpp
QuantumSegfault: Got it.

.github/actions/5a-wasi/action.yml
```diff
@@ -27,7 +27,7 @@ runs:
 
         bootstrap-ldc/bin/ldc-build-runtime \
           --ninja \
-          --dFlags="-mtriple=$triple" \
+          --dFlags="-mtriple=$triple;-fno-minfo-localclasses"
```
kinke: [Missing trailing backslash.]

kinke (approved):
Nice, thanks! I suggest adding a changelog entry, to get some visibility; also briefly explaining the rationale (higher linker stripping potential, at the 'cost' of breaking `Object.factory()`), IMO also in the switch description.

QuantumSegfault:
Yeah. I should probably add a changelog entry for the Wasm spill pointers pass (and WASI in general, as you mentioned).

And Wasm EH.

kinke:
Oh, I've just seen that we have `-fno-moduleinfo` already. So expanding `minfo` to full `moduleinfo` would probably be more consistent.

QuantumSegfault:
More consistent, but makes an already long switch longer...

You want it changed?

kinke:
*My* personal preference is consistency over switch compactness.

QuantumSegfault:
Okay. I'll change it in the morning.

PR #5226, opened 2026-07-25
Changelog entries for Wasm & WASI changes

QuantumSegfault:


QuantumSegfault:
@kinke

CHANGELOG.md
```diff
@@ -10,9 +10,16 @@
...
 - The `@ldc.attributes.restrict` UDA is now supported for dynamic-array parameters too. Builtin array ops (`c[] = a[] * b[]` etc.) use it, enabling auto-vectorization. Note that for such array ops, the dlang spec prescribes that the slice on the left and any slices on the right must not overlap (https://dlang.org/spec/arrays.html#array-operations), and this knowledge is now used by the optimizer. (#5148, #5168, #5176)
 - WebAssembly: The default `-fvisibility` setting was changed to `hidden`, like clang, so that only explicit `export`ed symbols are not hidden. And `-L--export-dynamic` isn't used by default for linking anymore either; add it manually to export all non-hidden symbols, as for normal Posix targets. The net effect is that the size of linked wasm binaries shrinks significantly by default. (#5216)
+- WebAssembly: Exceptions are now supported (#5162). Currently, using D and C++ exceptions are mutually exclusive. If linking in DRuntime, you cannot link in C++ code relying on `libunwind` (exceptions), due to the hacks required to convince LLVM to use the D's EH personality. A future LLVM release may help remove the constraint.
```
kinke, line 13: Do you know how this affects emscripten targets? Also wrt. the wasm EH switches in ldc2.conf only getting injected for WASI targets.

Do you have an idea of where emscripten is headed? What does it offer compared to the WASI SDK? Anything of relevance for D?

QuantumSegfault: I'm not as familiar Emscripten as I am with WASI, but I'll try to answer some of this.

> Do you know how this affects emscripten targets? Also wrt. the wasm EH switches in ldc2.conf only getting injected for WASI targets.

Long before Wasm had it's own EH handling, Emscripten had it's own work around using JS. However it appears that nowadays, using Wasm EH is supported in Emscripten: https://emscripten.org/docs/porting/exceptions.html#webassembly-exception-handling-based-support.

So this new EH should also be appropriate for Emscripten, and if/when LDC gets official support (OpenD does; and given that change that pretends Emscripten is Linux musl, it might not be too hard), and should also be added to the Emscripten config. 

In fact it NEEDS to if the code uses exceptions, otherwise LLVM will complain about features being used but not enabled (as codegen is still rigged to be Wasm EH).

> Do you have an idea of where emscripten is headed? What does it offer compared to the WASI SDK? Anything of relevance for D?

That I'm not so clear on. I do know that if you are specifically targetting the web, and want access to graphics, audio, etc, it's really the only option. WASI provides a standardized CLI environment that works both inside the browser (with shims) and out. However if being tied to the browser and JS glue isn't a problem, and WASI doesn't give you what you need, Emscripten is still the better option for more complex things (especially if you are *specifically* trying to interact specifically with the web/browser; e.g. DOM).

For now, Emscripten is a more complete but browser-centric solution, while WASI is slowly building up a Wasm-native non-browser system for what was only (easily) doable with Emscripten before.

kinke: Alright, thanks for the nice overview!

> In fact it NEEDS to if the code uses exceptions, otherwise LLVM will complain about features being used but not enabled (as codegen is still rigged to be Wasm EH).

Yeah that's what I expected. - I don't think we can set these EH options directly in the compiler itself (generic LLVM `TargetOptions`), so we'll probably have to keep injecting the switches via ldc2.conf, and duplicating that in `runtime/CMakeLists.txt`. But then probably for all wasm targets (`51-target-wasm-base.conf`), not just WASI. I guess that's fine for bare-metal wasm too, without any EH usage? (So e.g. not requiring an extra runtime feature baked into the 'binary', which could limit compatibility.)

> and if/when LDC gets official support (OpenD does; and given that change that pretends Emscripten is Linux musl, it might not be too hard)

Right, so I was actually expecting to see some potentially required druntime contributions for Emscripten next, not your WASI work. :)

Okay, I was hoping WASI would be a kinda cleaner/leaner Emscripten successor; so that's seemingly not the case mid-term at least, and Emscripten still relevant - good to know.

QuantumSegfault: > I guess that's fine for bare-metal wasm too, without any EH usage?
....yeah

That _should_ be fine. It shouldn't be a problem if no EH instructions are actually emitted.

> Okay, I was hoping WASI would be a kinda cleaner/leaner Emscripten successor; so that's seemingly not the case mid-term at least, and Emscripten still relevant - good to know.

Kind of is, kind of isn't (yet). There's all sorts of work being done on multiple fronts. e.g. a WebIDL => WIT converter that would make it easier to interact with Web APIs from "WASI" (though technically that's not WASI as much as it is Wasm components).

Either way, Emscripten still has a place in ecosystem for the foreseeable future.

QuantumSegfault:
Anything else that needs to happen here? I want this merged before I resolve the conflict in #5178

kinke (approved):

PR #5227, opened 2026-07-25
Enable EH for all wasm targets, not just WASI

kinke:
And handle the `--exception-model=wasm` required with LLVM 21-22 directly in the compiler.

### DIFF
--- a/driver/targetmachine.cpp
+++ b/driver/targetmachine.cpp
@@ -627,6 +627,19 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
     targetOptions.EmulatedTLS = false;
   }
 
+  // wasm: enable EH (with the help of 51-target-wasm-base.conf)
+  if (triple.isWasm()) {
+    if (!hasFeature("exception-handling")) {
+      features.push_back("+exception-handling");
+    }
+
+#if LLVM_VERSION_MAJOR >= 21 && LLVM_VERSION_MAJOR <= 22
+    // we need to enforce `--exception-model=wasm`, although it should be
+    // implied by `--wasm-enable-eh`
+    targetOptions.ExceptionModel = llvm::ExceptionHandling::Wasm;
+#endif
+  }
+
   const std::string finalFeaturesString =
       llvm::join(features.begin(), features.end(), ",");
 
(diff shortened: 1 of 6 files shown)


kinke:
@QuantumSegfault

driver/targetmachine.cpp
```diff
@@ -627,6 +627,15 @@ createTargetMachine(const std::string targetTriple, const std::string arch,
     targetOptions.EmulatedTLS = false;
   }
 
+#if LLVM_VERSION_MAJOR >= 21 && LLVM_VERSION_MAJOR <= 22
```
QuantumSegfault: I wonder if we could/should move more of the config here.

E.g. `-mattr=+exception-handling`

kinke: I was considering it very briefly, but then thought it'd be too specific, plus I don't see how we can move the others (tampering with the cmdline args passed to/parsed by LLVM would be too ugly IMO).

QuantumSegfault: Fair.

kinke: Oh well actually we already set some features for other targets, so yeah, we could move the `-mattr` over too. Unfortunately not the LLVM-version-specific switches, those are the real PITA.

QuantumSegfault (approved):
LGTM.

I haven't done any testing with Emscripten, but I did check that passing these switches don't have any adverse with baremetal Wasm.

Wasmtime (46) runs it just fine without `-W exceptions`.

Technically it does get added to `target_features` in the final binary (alongside other unused features):
```
(@custom "target_features" (after data) "\0a+\0bbulk-memory+\0fbulk-memory-opt+\16call-indirect-overlong+\12exception-handling+\0eextended-const+\0amultivalue+\0fmutable-globals+\13nontrapping-fptoint+\0freference-types+\08sign-ext")
```

But one could always pass `-mcpu=-exception-handling` if it causes trouble somehow.

QuantumSegfault:
~~Ah hold on. I think setting `exception-model` without `--enable-wasm-eh` will cause problems~~

Never mind. Confused the DRuntime config with the compiler config. It's all good

QuantumSegfault (approved):
Looks good!

kinke:
> But one could always pass -mcpu=-exception-handling if it causes trouble somehow.

Ah good point, I'm actually enforcing it at the moment (the user `-mattr` ones are injected before).

QuantumSegfault (approved):
(is this redundant?)

PR #5228, opened 2026-07-26
cmake: link libzstd when LLD is built against a shared libLLVM

JohanEngelen:
LLD's ELF backend references zstd (OutputSection::maybeCompress, for compressed output sections). When LLVM is a shared library, libLLVM only loads libzstd privately -- it does not re-export those symbols -- and `llvm-config --system-libs` reports nothing in shared mode. The statically-linked lld archives are therefore left with unresolved ZSTD_* symbols at final link (seen with Homebrew's shared llvm@16 on macOS).

Link libzstd explicitly whenever LLVM is shared. Using the full path from find_package(zstd) also covers installs outside the default linker search path (e.g. Homebrew's /opt/homebrew/opt/zstd). This generalizes and subsumes the previous FreeBSD-only workaround, which addressed the exact same gap (its LLVM port also ships a shared libLLVM).

By @ljmf00

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -574,11 +574,21 @@ if(LDC_WITH_LLD)
     else()
         set(LDC_LINKERFLAG_LIST -llldMinGW -llldCOFF -llldELF -llldMachO -llldWasm -llldCommon -lLLVMSymbolize ${LDC_LINKERFLAG_LIST})
     endif()
-    if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
-        # FreeBSD LLVM port links to zstd, but does not convey this information via CMake
-        # Workaround it here until it is fixed in the port
-        find_package(zstd)
-        list(APPEND LDC_LINKERFLAG_LIST "$<TARGET_LINKER_FILE:zstd::libzstd_shared>")
+    if(LLVM_IS_SHARED)
+        # LLD's ELF backend references zstd (OutputSection::maybeCompress). A shared
+        # libLLVM only loads libzstd privately -- it does not re-export those symbols --
+        # and `llvm-config --system-libs` reports nothing in shared mode, so the
+        # statically-linked lld archives are left with unresolved ZSTD_* symbols.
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

kinke:
Looks good.

Note that I've been avoiding the libzstd dependency for the prebuilt macOS packages via `LLVM_ENABLE_ZSTD=OFF` (https://github.com/ldc-developers/llvm-project/blob/103430347a4780a835bdb6e6d862a8ed4e8ef2d5/.github/workflows/ldc-release.yml#L114). IIRC, because the homebrew-installed lib wasn't found by default, and I didn't want to mess with adding homebrew paths. So this issue might happen on Mac in general, with static LLVM too, if homebrew is the standard way of getting libzstd.

If we wanted to enable libzstd support for the next prebuilt LLVM macOS binaries, we'd probably have to make sure to link it statically for LDC, as done on Linux: https://github.com/ldc-developers/ldc/blob/29d4a7dd34d2466036895b4f07457df88db17f1a/.github/actions/1-setup/action.yml#L39-L40

**Edit**: And `LLVM_USE_STATIC_ZSTD=TRUE` for the LLVM build itself: https://github.com/ldc-developers/llvm-project/blob/103430347a4780a835bdb6e6d862a8ed4e8ef2d5/.github/workflows/ldc-release.yml#L19

ljmf00-wekaio (approved):
Thanks for upstreaming this! It was mostly a pain for me in macOS at work but I was not yet sure that was indeed a generalized problem or some misconfiguration of my environment.

QuantumSegfault:
> Thanks for upstreaming this! It was mostly a pain for me in macOS at work but I was not yet sure that was indeed a generalized problem or some misconfiguration of my environment.

Ditto. I use MacPorts, and trying to build against `llvm-mp-20` always required an extra

`-DCMAKE_EXE_LINKER_FLAGS="-lzstd" -DCMAKE_BUILD_RPATH="/opt/local/libexec/llvm-20/lib"`

At least I can drop one of those switches now...

PR #5229, opened 2026-07-26
Revert "druntime: Use WASIp1 directly instead of POSIX emulation"

QuantumSegfault:
This reverts commit e5150d1a0763f1f9cf8e8e18acc6d4bf87358aa5 (PR dlang/dmd#23397)

Following discussion in https://github.com/dlang/dmd/pull/23398

Downstream mirror of dlang/dmd#23462


QuantumSegfault:
@thewilsonator 
@kinke

kinke (approved):
Thx for remembering the changelog.

JohanEngelen:
What's the harm in keeping sys/wasi/p1.d?  The work has been done already now, pity to lose it?

JohanEngelen:
> What's the harm in keeping sys/wasi/p1.d? The work has been done already now, pity to lose it?

Ah, it's already merged upstream. Well, alas. I guess this may become available (with the `p2` file too) as dub package so it's not lost.

QuantumSegfault:
Yeah, I'll find somewhere else for it eventually.

Though I don't see it finding much use if not by DRT/Phobos. I imagine most will just take the easy route, and use libc or Posix.

PR #5236, opened 2026-07-31
Make missed dynamic-cast lowerings in the frontend a proper ICE

kinke:
Since we had multiple reports of related frontend regressions (IIRC, since D v2.112), so we cannot use an `llvm_unreachable` which gets optimized away and causes weird other error messages about Objective-C.


rainers:
Is this related to this: https://github.com/dlang/dmd/pull/23263 ?

kinke:
Not sure, but definitely to https://github.com/dlang/dmd/issues/22422 and https://github.com/dlang/dmd/issues/22681.

PR #5238, opened 2026-07-31
GHA Main: Bump macOS image from macos-15 to macos-26

kinke:


kinke:
Hmm, this went suspiciously smoothly.

PR #5240, opened 2026-07-31
CI: Build more addon packages for simple cross-compilation

kinke:
Starting with Windows and Android.


kinke:
The added CI job simply processes the existing artifacts and extracts Windows/Android addon ones, to be unpacked into the LDC install directory directly, and then augmenting that install with prebuilt libs and a `55-target-*.conf` file. The user needs a cross-SDK (except for Windows), and ideally has its binaries on PATH, or extends the `55-target-*.conf` file with an appropriate `-gcc=<path to C cross-compiler>` switch.

The job prolongs the GHA Main workflow by 2-3 minutes, that's acceptable I think. The Windows addon artifact size is 43 MB (supporting 3 archs incl. arm64, but no MinGW-based libs for arm64, so not suited for cross-linking yet), the Android one (supporting 4 archs) 35 MB.

kinke:
I've tested yesterday's CI artifacts locally on Linux x86_64, cross-compiling & -linking a Phobos-hello-world successfully to the 2 Windows and 4 Android targets, the latter using a PATH as mentioned in the changelog. Also testing `-link-defaultlib-shared`.

PR #5241, opened 2026-07-31
Fixes for `WasmSpillPointers`

QuantumSegfault:
Fixes a couple of issues with `WasmSpillPointers` and its interaction with EH I found while getting Phobos to compile.

- `catchswitch` does NOT always have one successor. The unwind destination is another. However [`WinEHPrepare::demotePHIsOnFunclets`](https://github.com/llvm/llvm-project/blob/f99ee4f26d51091b805fc12affd19b4497081525/llvm/lib/CodeGen/WinEHPrepare.cpp#L864) converts applicable PHIs into stack slots anyway, so we don't actually need to spill!
- We now correctly split the `Invoke->getNormalDest()` edge to place the `lifetime.end` if the target has multiple predecessors.

### DIFF
--- a/gen/passes/WasmPointersSpill.cpp
+++ b/gen/passes/WasmPointersSpill.cpp
@@ -144,7 +144,9 @@ bool WasmPointersSpill::run(Function &F) {
           continue;
 
         if (TypeContainsPointers(I.getType())) {
-          PotentialPointers.insert(&I);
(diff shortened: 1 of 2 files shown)


QuantumSegfault:
@kinke

thewilsonator (approved):

PR #5245, opened 2026-08-02
GHA Main: Try bumping Ubuntu host image (24 => 26) for Alpine x86_64 container job

kinke:
There are signal-handler related unittest errors recently - for some CI runs only (then the same unittest fails for all 4 debug/release shared/static variants).

Might be container-related; try bumping to a newer kernel.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,7 +51,7 @@ jobs:
             with_pgo: true
 
           - job_name: Alpine musl x86_64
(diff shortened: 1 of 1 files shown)


kinke:
Let's see if this gets rid of the flakiness.

kinke:
Argh, new cannot-disable-ASLR issues for sanitizer lit-tests - obviously only after the first successful CI run here, as if just to piss me off.

PR #5250, opened 2026-08-02
druntime: Fix `rt_term` deadlock when a `Thread.start` fails

QuantumSegfault:
An issue I found while testing WASI.

When `Thread.start` fails, the `incrementAboutToStart` is never undone, as `Thread.add` never gets called.

`rt_term` calls `thread_joinAll`. This imbalance causes `thread_joinAll` to deadlock waiting for `ThreadBase.nAboutToStart` to zero out.

---------

Upstream PR: https://github.com/dlang/dmd/pull/23504


QuantumSegfault:
@thewilsonator

PR #5253, opened 2026-08-04
Fix loop-vectorizations being enabled by default with -Oz and -O1

### LINKED ISSUE #5233: Loop vectorization is never gated by -O level (dead condition in getPipelineTuningOptions)
`gen/optimizer.cpp:392-397` intends to enable loop vectorization only at `-O2`/`-O3`, but the
condition can never be true:

```cpp
  // This is final, unless there is a #pragma vectorize enable
  if (disableLoopVectorization) {
    pto.LoopVectorization = false;
    // If option wasn't forced via cmd line (-vectorize-loops, -loop-vectorize)
  } else if (!pto.LoopVectorization) {
    pto.LoopVectorization = optLevelVal > 1 && sizeLevelVal < 2;
  }
```

`pto` is a freshly default-constructed `PipelineTuningOptions`, and LLVM initialises
`LoopVectorization` to `true`. So `!pto.LoopVectorization` is always `false`, the assignment never
runs, and `LoopVectorization` stays `true` for every `-O` and `-Oz`.

The SLP line immediately below is written differently and *is* correctly gated, which suggests the
asymmetry is unintended:

```cpp
  pto.SLPVectorization =
      disableSLPVectorization ? false : optLevelVal > 1 && sizeLevelVal < 2;
```
**Repro:**

```d
// vec.d
void add(float[] a, float[] b, float[] c) {
    foreach (i; 0 .. a.length) a[i] = b[i] + c[i];
}
```

```
ldc2 -c -O1 -mcpu=haswell -output-s vec.d   # emits vmulps/vaddps on ymm — vectorized
ldc2 -c -Oz -mcpu=haswell -output-s vec.d   # also vectorized, despite -Oz
```

**Expected:** loop vectorization off at `-O1`, and off at `-Oz` (`sizeLevelVal == 2`), matching the
SLP behaviour and the apparent intent of the code.

**Note on intent:** the comment mentions `-vectorize-loops` / `-loop-vectorize`, so the original
intent may have been to check whether the user forced LLVM's own `cl::opt`, i.e. something like
`getNumOccurrences() > 0` on that option rather than reading `pto`. Happy to implement whichever
semantics you prefer.

kinke:
And enable the SLP vectorizations with -Oz. Resolves #5233.

Also drop `-vectorize-{loops,slp}` inherited from LLVM, which conflict with our dedicated switches.

### DIFF
--- a/gen/optimizer.cpp
+++ b/gen/optimizer.cpp
@@ -406,23 +406,12 @@ static std::optional<PGOOptions> getPGOOptions() {
 static PipelineTuningOptions getPipelineTuningOptions(unsigned optLevelVal, unsigned sizeLevelVal) {
   PipelineTuningOptions pto;
 
-  pto.LoopUnrolling = optLevelVal > 0;
-
-  pto.LoopUnrolling = !((disableLoopUnrolling.getNumOccurrences() > 0)
-                                   ? disableLoopUnrolling
-                                   : optLevelVal == 0);
-
-  // This is final, unless there is a #pragma vectorize enable
-  if (disableLoopVectorization) {
-    pto.LoopVectorization = false;
-    // If option wasn't forced via cmd line (-vectorize-loops, -loop-vectorize)
-  } else if (!pto.LoopVectorization) {
-    pto.LoopVectorization = optLevelVal > 1 && sizeLevelVal < 2;
-  }
-
-  // When #pragma vectorize is on for SLP, do the same as above
-  pto.SLPVectorization =
-      disableSLPVectorization ? false : optLevelVal > 1 && sizeLevelVal < 2;
+  pto.LoopUnrolling = !disableLoopUnrolling && optLevelVal > 0;
+  // loop vectorizer disabled with -Oz
+  pto.LoopVectorization =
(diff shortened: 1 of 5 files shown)


gen/optimizer.cpp
```diff
@@ -406,23 +406,12 @@ static std::optional<PGOOptions> getPGOOptions() {
...
-
-  // When #pragma vectorize is on for SLP, do the same as above
-  pto.SLPVectorization =
-      disableSLPVectorization ? false : optLevelVal > 1 && sizeLevelVal < 2;
+  pto.LoopUnrolling = !disableLoopUnrolling && optLevelVal > 0;
+  // loop vectorizer disabled with -Oz
+  pto.LoopVectorization =
+      !disableLoopVectorization && optLevelVal > 1 && sizeLevelVal < 2;
+  // SLP vectorizer enabled with -Oz
+  pto.SLPVectorization = !disableSLPVectorization && optLevelVal > 1;
```
kinke, line 414: Do we want the users to be able to enforce these optimizations, via `-disable-…=false`?

JohanEngelen: seems pretty niche to me...

PR #5255, opened 2026-08-05
Fix build on recent FreeBSD/powerpc64le

pkubaj:
FreeBSD on powerpc64le has recently switched to IEEE 128b long double.
Use proper mangling for that. Since we don't support IBM long double -
we switched from 64-bit long double - don't use glibc's double ABI
mechanism.

### DIFF
--- a/runtime/druntime/src/core/stdc/config.d
+++ b/runtime/druntime/src/core/stdc/config.d
@@ -658,6 +658,11 @@ package(core) template muslRedirTime64Mangle(string name, string redirectedName)
 }
 
 version (PPC64)
-    enum PPCUseIEEE128 = real.mant_dig == 113;
+{
+    version (CRuntime_Glibc)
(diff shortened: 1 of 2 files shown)


runtime/druntime/src/core/stdc/config.d
```diff
@@ -658,6 +658,11 @@ package(core) template muslRedirTime64Mangle(string name, string redirectedName)
 }
 
 version (PPC64)
-    enum PPCUseIEEE128 = real.mant_dig == 113;
+{
+    version (CRuntime_Glibc)
+        enum PPCUseIEEE128 = real.mant_dig == 113;
+    else
+        enum PPCUseIEEE128 = false;
```
kinke:
```suggestion
        enum PPCUseIEEE128 = false; // no dual-ABI mangling for e.g. FreeBSD
```

thewilsonator (approved):

PR #5256, opened 2026-08-05
core/int128.d: udivmod: don't inline (fixes #4916)

### LINKED ISSUE #4916: `std.int128` unittest failure with `-O1 -mcpu=x86-64-v3` since #4892
1.41.0-beta1 is failing one of its unittests for me:
```
$ ctest -V -R std.int128
UpdateCTestConfiguration  from :/home/happy/tmp/ldc-build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/happy/tmp/ldc-build/DartConfiguration.tcl
Test project /home/happy/tmp/ldc-build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 370
    Start 370: std.int128-shared

370: Test command: /home/happy/tmp/ldc-build/runtime/phobos2-test-runner-shared "std.int128"
370: Working Directory: /home/happy/tmp/ldc-build/runtime
370: Test timeout computed to be: 10000000
370: ****** FAIL release64 std.int128
370: core.exception.AssertError@std/int128.d(574): Int128(Cent(14, 0)) != Int128(Cent(15, 0))
370: ----------------
370: ??:? _d_assert_msg [0x7f703c761383]
370: ??:? pure nothrow @nogc @safe void std.int128.__unittest_L521_C1() [0x7f703d7507c7]
370: ??:? [0x7f703d751a8f]
370: ??:? [0x5632a7c2c697]
370: ??:? [0x5632a7c2c53f]
370: ??:? [0x5632a7c2c451]
370: ??:? runModuleUnitTests [0x7f703c784173]
370: ??:? void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).runAll() [0x7f703c79f78a]
370: ??:? _d_run_main2 [0x7f703c79f5a6]
370: ??:? _d_run_main [0x7f703c79f38c]
370: ??:? [0x7f703c4bd16d]
370: ??:? __libc_start_main [0x7f703c4bd228]
370: ??:? [0x5632a7c2c1c4]
1/2 Test #370: std.int128-shared ................***Failed    0.01 sec
test 813
    Start 813: std.int128-debug-shared

813: Test command: /home/happy/tmp/ldc-build/runtime/phobos2-test-runner-debug-shared "std.int128"
813: Working Directory: /home/happy/tmp/ldc-build/runtime
813: Test timeout computed to be: 10000000
813: 0.000s PASS debug64 std.int128
2/2 Test #813: std.int128-debug-shared ..........   Passed    0.01 sec

The following tests passed:
	std.int128-debug-shared

50% tests passed, 1 tests failed out of 2

Total Test time (real) =   0.03 sec

The following tests FAILED:
	370 - std.int128-shared (Failed)
Errors while running CTest
Output from these tests are in: /home/happy/tmp/ldc-build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```

I've configured the project with `-DD_FLAGS_RELEASE="-O1;-mcpu=x86-64-v3"`

the-horo:
This works around an issue that makes the code produce seemingly bad results when all of the following condition are met:
1. The code is compiled with `-O1` (at least)
2. The code is compiled with `-mattr=bmi2`
3. The running processor is AMD (reproduced with *AMD Ryzen 7 5825U* and *AMD Ryzen 3960X*)

The `std.int128` unittests can be used to test if the code was misscompiled.

### DIFF
--- a/runtime/druntime/src/core/int128.d
+++ b/runtime/druntime/src/core/int128.d
@@ -750,6 +750,11 @@ U udivmod(Cent c1, U c2, out U modulus)
     {
         version (GNU_OR_LDC_X86_64)
         {
+            // FIXME: https://github.com/ldc-developers/ldc/issues/4916
+            // An AMD CPU with -mattr=bmi2 fails std.int128 unittests.
+            // As a workaround, not inlining fixes it.
+            static if (__traits(targetHasFeature, "bmi2"))
+                pragma(inline, false);
             U ret = void;
             asm pure @trusted nothrow @nogc
             {


kinke:
I can't say I really like this; inlining a single instruction makes perfect sense, if it wasn't for this weird problem on some CPUs.

the-horo:
Yeah, it's relatively inefficient but I'm also looking for any kind of fix because I don't want to keep this bug in Gentoo, where people do compile their stuff with `-march=native`...

Interestingly that the testsuite fails on aarch64, I did run it on my RPI and it didn't fail, which, from the way the test is written, makes me question if the testsuite actually ran.

kinke:
It just fails because I added an explicit codegen check at the asm level for this, making sure that everything is inlined: https://github.com/ldc-developers/ldc/blob/master/tests/codegen/int128.d (always targeting x86_64, on aarch64 too). So that would need an adjustment if we go through with this pessimization.

the-horo:
Hmm, if the function is no longer inline do you know how we could check its generated assembly? I assume if we don't go through linking then we don't see the code of the non-inline function, but if we go, we no longer have a way to check the instruction.

kinke:
Yeah exactly, so the 2nd `mul_divmod()` test function should probably just be removed to get this to pass, it provides no other value, the multiplication is already checked in the first function.

FWIW, I'm still pretty sure this is an LLVM bug, I can't imagine that's really a CPU bug - I did search briefly for divq and CPU bugs, but couldn't find any. Maybe there is a better workaround, such as inserting some dummy instruction afterwards - if the problem is that somehow a divq cannot follow or be followed by a BMI2 instruction, or something along those lines.

the-horo:
If I had to bet on something, it would definitely be that LLVM is the root cause, but I lack the knowledge to debug it. Perhaps if we were able to get similar C code maybe we would be able to identify what is broken.

runtime/druntime/src/core/int128.d
```diff
@@ -744,6 +744,7 @@ Cent udivmod(Cent c1, Cent c2, out Cent modulus)
 pure
 U udivmod(Cent c1, U c2, out U modulus)
 {
+    pragma(inline, false)
```
kinke:
```suggestion
    pragma(inline, false);
```
I'm actually surprised this works; inside a function body, the pragma needs to be in statement form.

runtime/druntime/src/core/int128.d
kinke, line 758: And we actually want it here (x86_64 only, no need to pessimize the generic software implementation for all other archs), and for LDC only, with an according FIXME comment incl. link to the issue.

runtime/druntime/src/core/int128.d
the-horo: Huh, new feature and not a bug, I guess

runtime/druntime/src/core/int128.d
the-horo: Well pointed.

JohanEngelen:
Perhaps only apply the pessimization when `__traits(targetHasFeature, "bmi2") == true` ?

kinke:
Ah right, that sounds much better! Let me try it.

**Edit**: Yep, works totally fine for me, incl. `-O -mcpu=native` on my Ryzen 3960:
```D
static if (__traits(targetHasFeature, "bmi2"))
    pragma(inline, false);
```

runtime/druntime/src/core/int128.d
```diff
@@ -750,6 +750,11 @@ U udivmod(Cent c1, U c2, out U modulus)
     {
         version (GNU_OR_LDC_X86_64)
         {
+            // FIXME: https://github.com/ldc-developers/ldc/issues/4916
+            // An AMD CPU with -mattr=bim2 produces incorect code.
```
kinke:
```suggestion
            // An AMD CPU with -mattr=bmi2 fails std.int128 unittests.
```

kinke (approved):

thewilsonator (approved):

PR #5259, opened 2026-08-10
Emscripten: Add druntime & Phobos support

kinke:
Based on predefined `CRuntime_Musl`, `WASI` and `WASIp1`.

Also add an according CI job similar to the WASI one, incl. generating an addon package.


kinke:
@QuantumSegfault: This seems like a viable approach - basing Emscripten on your WASI work, after reading https://v8.dev/blog/emscripten-standalone-wasm and noting that they apparently try to use WASI APIs as much as possible, even in non-standalone-wasm mode.

To get the druntime test runners to link, I only had to avoid a single undefined `_CLOCK_MONOTONIC` symbol. And a single module fails its unittests (`core.sync.event`): https://github.com/ldc-developers/ldc/actions/runs/31392028311/job/93465932716?pr=5259

QuantumSegfault:
I'm not sure this makes sense to do like this.

Emscripten use a more vanilla Musl, without all of the differences to the headers WASI-libc made. They opted to stub out functions, rather than remove them, and otherwise use the unmodified types, constants, etc. (other than where architecture specific differences are concerned). A lot of the `CRuntime_WASI` blocks, particularly in Phobos, may be unnecessary.

`WASI` is also iffy, as that's an internal implementation detail. Emscripten also provides fuller emulation of things like threads, that we disable tests for on WASI. Also they provide their own filesystem, which isn't subject to all the quirks of WASI (e.g. the inability to have absolute symlinks; though that's largely a Wasmtime limitation).

kinke:
Yeah the `importc_compare` test shows a number of incompatible declarations, seemingly matching your points. - I'll pursue this a bit further, trying to get it more or less green (**edit**: i.e., the druntime/Phobos parts tested for WASI too). Then different avenues (e.g., based on `CRuntime_Musl` alone instead of `WASI`, `WASIp1` and `CRuntime_WASI`) can be explored as well.

I'd hope that Emscripten would eventually be based on wasi-libc...

kinke:
Okay, this is now based on `CRuntime_Musl` with a few `Emscripten` special-cases (instead of `CRuntime_WASI` earlier), but still uses `WASI[p1]` too (mainly to inherit the disabled parts and unitests). `importc_compare` is truly green; druntime still fails for a single `core.sync.event` unittest (depending on `pthread_cond_timedwait`), and Phobos just fails 2 `std.datetime` modules unittests, apparently just because I haven't found a way to make the node process use the `/usr/share/zoneinfo` host timezone files.

I'm okay with depending/building on predefined `WASI`, at least for the time being; supporting the few potential Posix API parts being stubbed/emulated on Emscripten and not available for WASI are IMO not of big interest. What's of interest to me is that Emscripten support is at a similar level to wasi-libc, so that wasm users can target both platforms with little to no hassle.

QuantumSegfault:
> I'm okay with depending/building on predefined `WASI`, at least for the time being; supporting the few potential Posix API parts being stubbed/emulated on Emscripten and not available for WASI are IMO not of big interest. What's of interest to me is that Emscripten support is at a similar level to wasi-libc, so that wasm users can target both platforms with little to no hassle.

Threading is a notable difference. Emscripten actually DOES support threading (over web workers; with some caveats).

https://emscripten.org/docs/porting/pthreads.html

Signals still not, other than for `pthread_kill`

Though I wouldn't mind holding off on the threading discussion until I figure out WASIp3 (with `wasi-sdk` 34 and LLVM 23 coming up; which will support cooperative threading [like Emscripten]; no preemptive threading yet though).

------

Other than that I think the rest of the WASI route-arounds should be fine, if not always necessary.

kinke:
> Though I wouldn't mind holding off on the threading discussion until I figure out WASIp3

Yeah that sounds good to me. [And note that I've just started with `WASIp1`, no idea if `WASIp2` might be a better match for Emscripten.] - In the long run, if Emscripten is to survive, I can definitely imagine it being based on wasi-libc and its Posix emulation layer in the future. So the distinction might go away, and duplicating all current Emscripten divergences potentially not worth the effort.

The compiler has a special case wrt. TLS globals for wasm in general IIRC, emitting them as thread-global; so the compiler will probably have to be adapted for wasm threading too.

QuantumSegfault:
> Yeah that sounds good to me. [And note that I've just started with `WASIp1`, no idea if `WASIp2` might be a better match for Emscripten.]

I'm fairly certain that WASIp1 is the only valid option for Emscripten.

They provide a subset of WASIp1 for the web (but not p2+ AFAICT)

https://github.com/emscripten-core/emscripten/blob/main/src/lib/libwasi.js

> The compiler has a special case wrt. TLS globals for wasm in general IIRC, emitting them as thread-global; so the compiler will probably have to be adapted for wasm threading too.

Good to know.

runtime/druntime/src/core/stdc/errno.d
```diff
@@ -2328,7 +2328,7 @@ else version (Haiku)
     enum B_NO_TRANSLATOR                 = (B_TRANSLATION_ERROR_BASE + 1);
     enum B_ILLEGAL_DATA                  = (B_TRANSLATION_ERROR_BASE + 2);
 }
-else version (CRuntime_WASI)
```
QuantumSegfault, line 2331: These `errno` are correct, but Emscripten provides additional ones:

https://github.com/emscripten-core/emscripten/blob/main/system/lib/libc/musl/arch/emscripten/bits/errno.h

kinke: Yeah here I was lazy and didn't want to duplicate the existing `CRuntime_WASI` block, after seeing that the Emscripten headers forward the WASI codes. [And I don't really care about the extra ones; if people relied on them, their code wouldn't work with WASI.]

At least a comment wrt. Emscripten would be good though; there's some polishing left to do, hence the draft state and WIP title. :)

QuantumSegfault: Yeah. Just the one thing that stuck out as probably somehow wrong, so I checked.

I'll wait for it to be undrafted before reviewing the changes more thoroughly.

driver/linker-gcc.cpp
```diff
@@ -807,16 +816,8 @@ int linkObjToBinaryGcc(llvm::StringRef outputPath,
...
-  std::unique_ptr<ArgsBuilder> argsBuilder;
-  if (global.params.targetTriple->isOSBinFormatWasm() && !global.params.targetTriple->isOSWASI()) {
-    argsBuilder = std::make_unique<LdArgsBuilder>();
-    tool = getProgram("wasm-ld", &opts::linker);
-  } else {
-    argsBuilder = std::make_unique<ArgsBuilder>();
-    tool = getCC(argsBuilder->args);
-  }
+  auto argsBuilder = std::make_unique<ArgsBuilder>();
+  std::string tool = getCC(argsBuilder->args);
```
kinke, line 820: This is e.g. also still to polish, causing the Alpine CI job failures (no integrated LLD, but external `wasm-ld`).

This all dates back to the very early wasm support, like the wasm-defaults .conf file - we used to only support bare-metal wasm, so no druntime and Phobos to link, and no need for a C compiler as linker driver, since there wasn't a wasm libc support either. So we used to invoke `wasm-ld` (unless we have an integrated LLD) with the ld CLI interface, for all (non-WASI) wasm targets.

We can keep this behavior for backwards-compatibility, but then only when targeting an unknown/none OS. We already do that in `55-target-wasm-naked.conf` nowadays, via the `-link-internally` (=> use integrated lld with `wasm-ld` CLI interface) - but only if LLD integration was enabled at build-time.

So the compiler special case here would only be needed without LLD integration. Not sure that's worth the trouble, incl. the `link_WebAssembly` lit complication (looking for an external `wasm-ld`), which I haven't gotten rid of yet.

QuantumSegfault: In theory, using most modern system `clang` as link driver for bare-metal Wasm should be fine? `wasm-ld` is a standard part of LLVM distributions nowadays.

One notable exception is (at least) macOS 15's clang. So you'll have to have either MacPorts/Homebrew LLVM/Clang available to use as `-gcc`, or use `wasi-sdk`'s (or `emsdk`'s).

But if we use `-link-internally` by default, then you probably won't notice the difference most of the time.

kinke: Oh right - I've just tested this with my system-default clang 18, and it works with an extra `-Xcc=-nostdlib`:
```
$ bin/ldc2 -mtriple=wasm32-unknown-unknown -v ../ldc/tests/codegen/wasm.d -link-internally=false -Xcc=-nostdlib -Xcc=-v
[…]
Did not find C cross-compiler: `wasm32-unknown-unknown-gcc`
Did not find C cross-compiler: `wasm32-unknown-unknown-clang`
Found C cross-compiler: `/usr/bin/clang`
/usr/bin/clang --target=wasm32-unknown-unknown wasm.o -o wasm.wasm -Xlinker -z -Xlinker stack-size=1048576 -Xlinker --stack-first -nostdlib -Wl,--gc-sections
[…]
 "/usr/bin/wasm-ld-18" -m wasm32 -L/usr/lib wasm.o -z stack-size=1048576 --stack-first --gc-sections -o wasm.wasm
```

So yeah, that should be good enough. Noone has requested an `ld` CLI interface for other targets either, so...

kinke:
The druntime unittest failure was apparently due to too simplistic pthread stubs (so I guess one has to explicitly opt into emulated pthreads support): https://github.com/emscripten-core/emscripten/blob/6313a2d67a92c3e3b449f48e2fc9882d7e45ff3c/system/lib/pthread/library_pthread_stub.c#L222-L224

The timezone stuff is a bit of a PITA; I got the testrunners to include and be able to read the host runner's `/usr/share/zoneinfo` directory, but e.g. the TZ env-var support seems to be enabled in standalone-wasm mode only, which causes more Phobos unittest regressions (I/O related): https://github.com/emscripten-core/emscripten/blob/6313a2d67a92c3e3b449f48e2fc9882d7e45ff3c/system/lib/libc/musl/src/time/__tz.c#L136-L276 (`USE_EXTERNAL_ZONEINFO` is defined unless using the standalone-wasm mode)

So it's probably best to follow the `CRuntime_WASI` exclusions wrt. timezone stuff, not worth the trouble; users will hardly bake timezone files into their binaries anyway.

kinke:
Ready for review now.

runtime/CMakeLists.txt
```diff
@@ -104,6 +104,10 @@ if("${TARGET_SYSTEM}" MATCHES "WASI|Emscripten")
 
     append("-Wl,-z,stack-size=1048576 -Wl,--stack-first" LD_FLAGS)
 
+    if("${TARGET_SYSTEM}" MATCHES "Emscripten")
+        append("-sALLOW_MEMORY_GROWTH" LD_FLAGS)
```
QuantumSegfault, line 108: Do we want to enable this by default in `ldc2.conf`?

If not maybe this block should be moved outside of
`(normally inherited from...` `if` block to be less confusing

runtime/druntime/src/core/stdc/limits.d
QuantumSegfault, line 1: Some of these values are incorrect.

I think you based these of the `_POSIX_*` values. Some of these have normal values.
```c
#define NAME_MAX 255
#define PATH_MAX 4096
#define PIPE_BUF 4096
```

runtime/phobos
QuantumSegfault, line 1: New changes look good.

------

Regarding the merging down of the changes made during WASI upstreaming, there was one issue I noticed.

The `version` gate for the unittest in [`typecons.d` on line 3553](https://github.com/ldc-developers/phobos/commit/f3ea448a43c5233120bb3d4960c0b9665c50df47#diff-81bed7f05cbd4e992067b7019125e6a1349ebe5098c6980b64bbbca8d5491e17R3553) should have been deleted, as the versioning was moved inside the unittest.

QuantumSegfault (changes requested):

runtime/CMakeLists.txt
kinke: Yeah I'm not sure how likely this is going to be needed for user apps as well. IIRC, Emscripten defaults to something like 20 MB, which does sound really low for D with GC.

runtime/druntime/src/core/stdc/limits.d
kinke: Oh, thx for checking so thoroughly! Yeah they were all based on the POSIX_ ones; the NAME_MAX of 14 or so did sound weird. Will fix.

runtime/phobos
kinke: Ah alright, missed one of these (unfortunately no merge conflict for these) - thx for spotting!

runtime/CMakeLists.txt
kinke: Do you know how wasmtime/wasi-libc handles this? All dynamic by default?

runtime/druntime/src/core/stdc/limits.d
kinke: Let's make `importc_compare` detect mismatching numerical constants for us: https://github.com/dlang/dmd/pull/23618

Currently yields this for Emscripten here:
```
Collected 270 C types.
Collected 1333 numerical C constants.
Error: Constant core.stdc.stdint.WCHAR_MIN diverges: 0 (D) vs. -2147483648 (ImportC)
Error: Constant core.stdc.stdint.WCHAR_MAX diverges: 1114111 (D) vs. 2147483647 (ImportC)
Error: Constant core.stdc.stdint.WINT_MIN diverges: 0 (D) vs. -2147483648 (ImportC)
Error: Constant core.stdc.signal.SIG_IGN diverges: 1 (D) vs. 4294967294 (ImportC)
Error: Constant core.stdc.limits.MB_LEN_MAX diverges: 2 (D) vs. 4 (ImportC)
Error: Constant core.stdc.math.math_errhandling diverges: 3 (D) vs. 2 (ImportC)
Error: Constant core.stdc.wchar_.WEOF diverges: 65535 (D) vs. 4294967295 (ImportC)
```

runtime/CMakeLists.txt
QuantumSegfault: I believe so.

runtime/druntime/src/core/stdc/math.d
```diff
@@ -214,8 +214,16 @@ else
 enum int MATH_ERRNO       = 1;
 ///
 enum int MATH_ERREXCEPT   = 2;
-///
-enum int math_errhandling = MATH_ERRNO | MATH_ERREXCEPT;
+version (Emscripten)
```
QuantumSegfault, line 217: This also applies to WASI

runtime/druntime/src/core/stdc/stdint.d
```diff
@@ -537,15 +537,30 @@ enum sig_atomic_t SIG_ATOMIC_MAX = sig_atomic_t.max;
...
-///
-enum wchar_t WCHAR_MIN = wchar_t.min;
-///
-enum wchar_t WCHAR_MAX = wchar_t.max;
-
-///
-enum wint_t  WINT_MIN  = wint_t.min;
-///
-enum wint_t  WINT_MAX  = wint_t.max;
+version (Emscripten) // wchar_t/wint_t is actually int
```
QuantumSegfault, line 540: This also applies to WASI

runtime/druntime/src/core/stdc/wchar_.d
```diff
@@ -167,8 +179,16 @@ else
 ///
 alias wint_t = wchar_t;
 
-///
-enum wchar_t WEOF = 0xFFFF;
+version (Emscripten)
```
QuantumSegfault, line 182: This also applies to WASI

runtime/druntime/src/core/sys/posix/sys/socket.d
QuantumSegfault, line 1: To avoid having to open a separate PR for this, alongside fixing the other enum constant issues you are for Emscripten, could you also adjust `SHUT_RD`/`SHUT_WR`/`SHUT_RDWR` for WASI. Should be

```d
    enum
    {
        SHUT_RD = 1,
        SHUT_WR = 2,
        SHUT_RDWR = 3
    }
```

(starts at 1; not 0)

QuantumSegfault (commented):
New fixes look good (haven't re-reviewed anything that hasn't changed).

I ran your `importc_compare` patch against `wasip1` and `wasip2` and found some differences. Mostly the same as with Emscripten, but there was one WASI-specific error I made.

runtime/druntime/src/core/stdc/stdint.d
kinke: I'm not sure this is the way to go though, using `int` for these constants, while Phobos apparently can't deal with `wchar_t` being a non-char type.

kinke: I went a step back, leaving `WCHAR_{MIN,MAX}` untouched (so `wchar_t`-typed again), but defining `wint_t` to `int`, for WASI in general.

runtime/CMakeLists.txt
kinke: It's now enabled by default, in `55-emscripten-wasm32.conf`. For the druntime test runner, it apparently defaults to ~18 MB:
> Either (1) compile with -sINITIAL_MEMORY=X with X higher than the current value 18022400, …

QuantumSegfault (approved):

PR #5262, opened 2026-08-14
Bump DUB and dlang-tools versions

0xEAB:


kinke:
I try to keep the dub version in sync with the one used by bundled reggae (since there were cache-directory incompatibilities at one point etc.). So reggae master would need to bump its dub to v1.42.0 first (it's available on code.dlang.org, I've checked), so that we can bump reggae too. **Edit**: See e.g. https://github.com/atilaneves/reggae/pull/317.

Btw I'm surprised the final has been tagged already; the tag timings used to be in sync with the DMD ones.

0xEAB:
@kinke 
<https://github.com/atilaneves/reggae/commit/447958d92d1aac12eafc4573d3b15883d66ce481>

kinke:
Cool, then please update `packaging/reggae_version` to the latest SHA.

thewilsonator (approved):

PR #5266, opened 2026-08-25
druntime: Update to `wasi-sdk` 34

QuantumSegfault:
Updates the supported version of `wasi-libc` to the one distributed by `wasi-sdk` 34.

Also updates the version of Wasmtime used by CI to 48.0.1

The diff is smaller than it normally would be, as I incorporated some changes between 33 and 34 when I initially implemented (inadvertently).

Phobos is not yet updated to make use of `flockfile`. This will be done alongside other changes needed for WASIp3 later on (i.e. once it actually matters).

### DIFF
--- a/runtime/druntime/src/core/stdc/stdint.d
+++ b/runtime/druntime/src/core/stdc/stdint.d
@@ -400,8 +400,8 @@ else version (WASI)
 
     alias int_fast8_t   = byte;   ///
     alias uint_fast8_t  = ubyte;  ///
-    alias int_fast16_t  = short;  ///
-    alias uint_fast16_t = ushort; ///
+    alias int_fast16_t  = int;    ///
+    alias uint_fast16_t = uint;   ///
     alias int_fast32_t  = int;    ///
     alias uint_fast32_t = uint;   ///
 
--- a/runtime/druntime/src/core/sync/event.d
+++ b/runtime/druntime/src/core/sync/event.d
@@ -280,7 +280,10 @@ nothrow @nogc:
                     timespec t = void;
                     mktspec(t, tmout);
 
-                    result = pthread_cond_timedwait(&m_cond, &m_mutex, &t);
+                    version (WASI)   // wasi-libc traps for single-thread futex
+                        result = -1;
+                    else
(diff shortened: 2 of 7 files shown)


thewilsonator (approved):

thewilsonator (approved):

QuantumSegfault:
@thewilsonator 

All good now. Ready to merge

runtime/druntime/src/core/sync/event.d
```diff
@@ -308,6 +308,8 @@ private:
 }
 
 // Test single-thread (non-shared) use.
+version (WASI) {} // WASI is single-threaded
+else              // (& futex wait w/ timeout traps in single-threaded builds)
```
kinke: I'm wondering if this shouldn't be handled in the implementation then, as I did for Emscripten.

QuantumSegfault: That probably makes more sense. It's just `pthread_cond_timedwait` that's problematic for WASI.

I've updated it.

runtime/druntime/src/core/stdc/stdint.d
```diff
@@ -400,8 +400,8 @@ else version (WASI)
 
     alias int_fast8_t   = byte;   ///
     alias uint_fast8_t  = ubyte;  ///
-    alias int_fast16_t  = short;  ///
-    alias uint_fast16_t = ushort; ///
+    alias int_fast16_t  = int;    ///
```
JohanEngelen, line 403: nitpick: maybe add a comment that this highly unintuitive thing is the correct one?

QuantumSegfault: How come? This is far from the only case. In fact, it seems that `fast16_t` is rarely `short`. The BSDs, Solaris, Linux, (non-MinGW) Windows all have differing `fast16`.

Darwin and MinGW being the exceptions, rather than the rule here.

kinke (approved):

runtime/druntime/src/core/stdc/stdint.d
JohanEngelen: ah sorry, i didn't look further than just the diff snippet

PR #5268, opened 2026-08-26
Undo TLS hack on Wasm

QuantumSegfault:
In preparation for WASIp3.

This hack has not been required since LLVM 9 (this behavior of demoting to standard globals has been integrated into LLVM for some time now).

### DIFF
--- a/gen/llvmhelpers.cpp
+++ b/gen/llvmhelpers.cpp
@@ -1722,11 +1722,10 @@ llvm::GlobalVariable *declareGlobal(Loc loc, llvm::Module &module,
                                     llvm::StringRef mangledName,
                                     bool isConstant, bool isThreadLocal,
                                     bool useDLLImport) {
-  // No TLS support for WebAssembly and AVR; spare users from having to add
+  // No TLS support for AVR; spare users from having to add
   // __gshared everywhere.
   const auto arch = global.params.targetTriple->getArch();
-  if (arch == llvm::Triple::wasm32 || arch == llvm::Triple::wasm64 ||
-      arch == llvm::Triple::avr)
+  if (arch == llvm::Triple::avr)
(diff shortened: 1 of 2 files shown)


thewilsonator (approved):

kinke:
This is for post-1.43, so will have to wait a bit.

QuantumSegfault:
So is all WASIp3 related things going to be delayed? What about the upgrade to LLVM 23?

kinke:
I think I'll release v1.43.0 final when DMD v2.113.0 is tagged; this would IMO be too much of a breaking change since beta1.

kinke:
> this behavior of demoting to standard globals has been integrated into LLVM for some time now

Oh sorry I totally missed this. Okay in that case no reason to delay this. But how would this work for WASIp3 then? No demotion for that target builtin already?! **Edit**: Ideally no change for Emscripten targets, i.e., demotion for those.

QuantumSegfault:
WASIp1 (thus Emscripten), p2 demote

p3 has full TLS (there is just a tiny bit of DRuntime fixing we need to support that; but that's irrealvent right now)

QuantumSegfault:
Okay, I investigated more, and that's not QUITE right.

Seems like Emscripten does make use of TLS, more like p3. So doing this now would end up breaking possibly Emscripten (a newly supported target)...

But LLVM 22 doesn't support Wasm TLS anyway, only starting with 23. So this change would still be an NFC until we upgrade to LLVM 23!

-----

In summary, there should be no harm in merging this now. But also no point as there is zero difference without LLVM 23

kinke:
Heh okay. Please dummy-amend to retrigger CI (incl. checking Emscripten).

QuantumSegfault:
It works. Emscripten and WASI still pass.

PR #5272, opened 2026-08-26
update-dmd-rewrite.sh: Include new v2.* release branches

kinke:


kinke:
[Annoyingly, the upstream v2.113.0 tag is skipped for the rewrite, because the commit just touches the VERSION file.]

PR #5282, opened 2026-09-07
dcompute: Prepare pointer ABI rewrite for Vulkan support

gnavdev28:


runtime/phobos
kinke, line 1: No accidental submodule changes please.

gnavdev28: Sorry about that!

PR #5288, opened 2026-09-10
Finalize support for LLVM 23

kinke:


kinke:
Okay, just a single regression remaining (at least for the current single CI job on Mac), for `codegen/dcompute_dual_targets.d` (https://github.com/ldc-developers/ldc/actions/runs/34488939654/job/102910312558?pr=5288#step:11:144):
```
error: <unknown>:0:0: in function _D6inputs6kernel5k_fooFS3ldc8dcompute__T7PointerVEQBaQz9AddrSpacei1TfZQBeZv void (ptr addrspace(1)): store into a read-only SPIR-V storage class is not allowed
LLVM ERROR: cannot select: G_STORE %18:fid(s64), %15:pid(p0) :: (store (f32) into %ir.3, addrspace 2) (in function: _D6inputs6kernel5k_fooFS3ldc8dcompute__T7PointerVEQBaQz9AddrSpacei1TfZQBeZv)
```

thewilsonator:
I'm guessing its https://github.com/ldc-developers/ldc/blob/master/tests/codegen/inputs/kernel.d#L12 that line.

kinke:
It seems to be https://github.com/ldc-developers/ldc/blob/ac0e420af3fe74de50ff2138f9ffcb464c29873a/tests/codegen/inputs/kernel.d#L10, a store to **shared** memory. It fails for both CUDA and OpenCL.

PR #5290, opened 2026-09-11
Carry the D type alignment on struct init memset

thewilsonator:
Fission of #5279 to see what causes the problems with it


kinke:
I think this is pretty safe, as runtime-initializing an under-aligned variable seems exotic, e.g., `align(1) MyStruct s;` on the stack etc.

I'd strongly suggest a `DtoMemSetZero()` overload taking a frontend `Type*` though, to derive the alignment there, once.

JohanEngelen:
> I think this is pretty safe, as runtime-initializing an under-aligned variable seems exotic, e.g., `align(1) MyStruct s;` on the stack etc.

But if we now start to miscompile such code, we should error on code like that. Otherwise it may lead to very hard to track down bugs.

Full example:
```d
struct S { int a; }  // implicit align(4)

void foo() {
    bool b;
    align(1) S s; // miscompiled with this PR because the _type's_ alignment is used, not the variable's alignment.
}
```

PR #5292, opened 2026-09-16
Use aligned `memcpy` for struct assignment

thewilsonator:


gen/llvmhelpers.cpp
```diff
@@ -400,8 +400,11 @@ void DtoAssign(Loc loc, DValue *lhs, DValue *rhs, EXP op,
       // Check whether source and destination values are the same at compile
       // time as to not emit an invalid (overlapping) memcpy on trivial
       // struct self-assignments like 'A a; a = a;'.
-      if (src != dst)
-        DtoMemCpy(DtoType(lhs->type), dst, src);
+      if (src != dst) {
+        unsigned align = DtoAlignment(lhs->type);
+        DtoMemCpy(DtoType(lhs->type), dst, src, false,
+                  align ? align : 1);
```
kinke, line 406: We should probably add a `DtoMemCpy()` overload taking both alignments.

thewilsonator: I think the problem here (see failing CI) is that the call `gIR->ir->CreateMemCpy(dst, A, src, A, nbytes, false /*isVolatile*/);` (where `A` is an alignment generated from the argument to `DtoMemCpy`) actually needs to derive the alignment from `src`/`dst` (which I presume are going to be alloca's?) through the `Align/MaybeAlign`.

kinke: Assuming the rhs is always of the same (non-)qualified type as the lhs, we wouldn't need the separate alignments. Unless we really have assignments to/from under-aligned structs in the frontend, triggering this error, then we can't simply use the type alignments.

> which I presume are going to be alloca's?

No, they can be anything.

